PR #3508, opened 2018-02-16
[CoreAudioKit] Fix macOS xtro output

dalexsoto:


src/coreaudiokit.cs
```diff
@@ -60,6 +74,81 @@ interface AUAudioUnitViewControllerExtensions {
...
+
+		[Static]
+		[Export ("AUPannerViewWithAudioUnit:")]
+		AUPannerView Create (AudioUnit.AudioUnit au);
+	}
+
+	[NoiOS]
+	[Mac (10,13, onlyOn64: true)]
+	[BaseType (typeof (NSWindowController), Name = "CABTLEMIDIWindowController")]
+	interface CABtleMiDiWindowController {
```
spouliot: `Midi` (for MIDI acronym)

src/coreaudiokit.cs
```diff
@@ -60,6 +74,81 @@ interface AUAudioUnitViewControllerExtensions {
...
+
+		[Export ("initWithWindow:")]
+		IntPtr Constructor ([NullAllowed] NSWindow window);
+	}
+
+	[NoiOS]
+	[Mac (10,13, onlyOn64: true)]
+	[BaseType (typeof (NSViewController))]
+	interface CAInterDeviceAudioViewController {
```
spouliot: base a default `init` make sense ?
or should not another base ctor/init be re-exposed ?

dalexsoto: Ok, found some sample code browsing it is using `[CAInterDeviceAudioViewController new];` so we should be fine

PR #3527, opened 2018-02-19
[tests] Automatically rebuild 'test.config' if it can't be found.

rolfbjarne:
This condition usually happens when I 'git clean -xfd' in tests/ and then want
to run tests from xharness.

The manual step is to run "make" in tests/, but it's better if we can do it automatically.


rolfbjarne:
Test failures are unrelated

* MMP regression tests: [MMP regression tests: random failure](https://github.com/xamarin/maccore/issues/645)
* mscorlib/Mac: [MonoTests.System.Security.Cryptography.DSACryptoServiceProviderTest.Persistence_PersistKeyInCsp_False: System.NullReferenceException : Object reference not set to an instance of an object](https://github.com/xamarin/maccore/issues/640)
* mscorlib/iOS 32-bit: [iOS 32-bit sim: TimeZoneTest.TestCtors (D02) fails](https://github.com/xamarin/maccore/issues/629)

PR #3531, opened 2018-02-19
[mtouch] Rename temporary directories so the order assemblies move is clearer.

rolfbjarne:
I implemented this myself, but I can never remember in which order assemblies
go from one directory to another during the build.

So number these temporary directories, so that even the most forgetful minds
can understand without having to remember anything.


rolfbjarne:
Test failure is unrelated:

PR #3532, opened 2018-02-19
[linker] Take extra care to avoid 32/64 bits specific mscorlib.dll

spouliot:
We want to avoid separate `mscorlib.dll` assemblies for 32/64 bits so
the  architecture specific code for `n[u]int` and `nfloat` must be
preserved in both cases.

Because a single, slightly larger, assembly is much smaller than two
(slightly smaller) ones.

This is not a common situation since the extraneous preserved API are
often used in the application (or 3rd party code) so, in most cases,
a single `mscorlib.dll` was already used.

This merely close the gap for some cases, like our `link all` application
where this happened.

### DIFF
--- a/tools/linker/MonoTouch.Tuner/MonoTouchMarkStep.cs
+++ b/tools/linker/MonoTouch.Tuner/MonoTouchMarkStep.cs
@@ -127,7 +127,7 @@ namespace MonoTouch.Tuner {
 			}
 
 			// we want to avoid separate mscorlib.dll for 32/64 bits so the arch specific code for n[u]int and nfloat must be preserved in both cases
-			// a single, slightly larger, assenmbly is much smaller thn two (slightly smaller) ones
+			// a single, slightly larger, assembly is much smaller thn two (slightly smaller) ones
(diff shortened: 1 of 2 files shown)


tools/linker/MonoTouch.Tuner/MonoTouchMarkStep.cs
```diff
@@ -126,6 +126,27 @@ protected override MethodDefinition MarkMethod (MethodReference reference)
 				pending_serialization_constructors.Clear ();
 			}
 
+			// we want to avoid separate mscorlib.dll for 32/64 bits so the arch specific code for n[u]int and nfloat must be preserved in both cases
+			// a single, slightly larger, assenmbly is much smaller thn two (slightly smaller) ones
```
rolfbjarne: Seems like you have problems with your `n` key 😄

spouliot:
failure `[FAIL] NSDataTest.Https : Expected: True ` due to expired certificate -> https://github.com/xamarin/maccore/issues/644

PR #3538, opened 2018-02-19
[Tests] Change the guid of the test project to allow to use VS.

mandel-macaque:


mandel-macaque:
Unrelated failures, they seem to be Https related and date time related. Will create issues accordingly.

PR #644, opened 2016-08-23
[modelio] Update for iOS 10 beta 2

VincentDondain:
Note: Beta 5 and Beta 6 changes canceled each others.

PR #3549, opened 2018-02-20
[tests] Fix NSDataTest.Https to use a site with a valid and complete certificate chain. Fixes #644.

rolfbjarne:
Fixes https://github.com/xamarin/maccore/issues/644.


spouliot:
Not sure it's needed - the certificate was renewed and now expires in 2020

spouliot:
I'm not sure why it fails only on watchOS now (that the certificate was updated) and the logs are not helpful :(
Since this is not the goal of the test (and the test is already excluded from devices) let's apply the PR

spouliot:
build

PR #3554, opened 2018-02-20
Bump Xcode to 9.3 beta 3

VincentDondain:


VincentDondain:
Known issue: https://github.com/xamarin/maccore/issues/644

tests/xtro-sharpie/macOS-ObjCRuntime.ignore
```diff
@@ -20,3 +20,4 @@
 
 !unknown-pinvoke! kevent bound
 !unknown-pinvoke! kqueue bound
+!unknown-pinvoke! close bound
```
spouliot: that's not in the header changes
it might need to be reverted since it seems the same as https://github.com/xamarin/maccore/issues/638
which I suspect is related to metadata order in the assembly (so the p/invoke can get reported in a different location)

PR #3556, opened 2018-02-21
[tests] Fix NSDataTest.Https to use a site with a valid and complete certificate chain. Fixes #644. (#3549)

rolfbjarne:
Fixes https://github.com/xamarin/maccore/issues/644.


spouliot:
it will also be needed in `xcode9.3` branch

PR #3561, opened 2018-02-21
[metalkit] Add 'MTKTextureLoader' missing bindings (fix xtro)

VincentDondain:


src/metalkit.cs
```diff
@@ -288,6 +297,20 @@ interface MTKTextureLoader {
...
+		[NoTV]
+		[Mac (10,12, onlyOn64 : true)]
+		[Export ("newTextureWithName:scaleFactor:displayGamut:bundle:options:completionHandler:")]
+		[Async]
+		void FromName (string name, nfloat scaleFactor, NSDisplayGamut displayGamut, [NullAllowed] NSBundle bundle, [NullAllowed] NSDictionary options, MTKTextureLoaderCallback completionHandler);
+
+		[NoiOS]
+		[NoTV]
+		[Mac (10,12, onlyOn64 : true)]
+		[Wrap ("FromName (name, scaleFactor, displayGamut, bundle, options == null ? null : options.Dictionary, completionHandler)")]
```
dalexsoto: Use `options?.Dictionary` next time easier to read than `options == null ? null : options.Dictionary` :)

VincentDondain: Copy and pasting the code above was just too easy though :P

PR #3563, opened 2018-02-21
[modelio] MDLVoxelArray should have a base type of MDLObject

VincentDondain:
Moving the base type to `MDLObject` wasn't a breaking change on macOS.
There is no reason for the ifdef and xtro reports the issue correctly.


VincentDondain:
build

VincentDondain:
- xammac tests known failure: https://github.com/xamarin/maccore/issues/647
- Rest of the failures all due to watchOS simulator timing out: https://github.com/xamarin/maccore/issues/648

PR #3565, opened 2018-02-21
[arkit] Update to Xcode 9.3 beta 3

VincentDondain:


src/arkit.cs
```diff
@@ -62,6 +62,7 @@ public enum ARErrorCode : nint {
 		SensorFailed = 102,
 		CameraUnauthorized = 103,
 		WorldTrackingFailed = 200,
+		InvalidReferenceImage = 300
```
spouliot: minor (don't update PR just for that): add a `,` for new enum members
this makes the diff smaller for the next addition :)

VincentDondain: You're right, I forgot it this time

VincentDondain:
Known issue: https://github.com/xamarin/maccore/issues/647

PR #3569, opened 2018-02-22
[BusinessChat] Introduction of the new framework on iOS and Mac.

mandel-macaque:


mandel-macaque:
Diffs:

* [iOS](https://github.com/xamarin/xamarin-macios/wiki/BusinessChat-iOS-xcode9.3-beta2) 
* [Mac](https://github.com/xamarin/xamarin-macios/wiki/BusinessChat-macOS-xcode9.3-beta2)

src/BusinessChat/BCChatAction.cs
```diff
@@ -0,0 +1,15 @@
+#if XAMCORE_2_0
+using System;
+using System.Collections.Generic;
+using XamCore.Foundation;
+
+namespace XamCore.BusinessChat {
+	public partial class BCChatAction {
+		public static void OpenTranscript (string businessIdentifier, Dictionary<BCParameterName, string> intentParameters) {
+			var keys = Array.ConvertAll<string, NSString> (new List<string>(intentParameters.Values).ToArray (), v => new NSString (v));
```
spouliot: that's creating a **lot** of short-lived, managed objects (e.g. `ToArray`)

src/BusinessChat/BCChatAction.cs
```diff
@@ -0,0 +1,15 @@
...
+			var values = Array.ConvertAll<BCParameterName, NSString> (new List<BCParameterName> (intentParameters.Keys).ToArray (), k => k.GetConstant ());
```
spouliot: same

src/BusinessChat/BCChatAction.cs
spouliot: in particular `new List<string>(intentParameters.Values).ToArray ()` is bad
why create a short-lived list if what you want is an array

src/BusinessChat/BCChatAction.cs
```diff
@@ -0,0 +1,15 @@
...
+			OpenTranscript (businessIdentifier, NSDictionary<NSString, NSString>.FromObjectsAndKeys (values, keys));
```
spouliot: I'm not sure how many items are expected in the dictionary...
Maybe create a `MutableDictionary` and iterating/adding the items is better
otherwise having two array is an option (single call to native `Dictionary`) but let's get away from the temporary lists

src/BusinessChat/Enums.cs
```diff
@@ -0,0 +1,27 @@
+using XamCore.ObjCRuntime;
+using XamCore.Foundation;
+
+namespace XamCore.BusinessChat {
+
+	[Mac (10,13, onlyOn64: true), iOS (11,3)]
```
spouliot: pretty sure it did not exists in 10,13
let's put 10,13,4 which is the SDK where it became public

src/BusinessChat/Enums.cs
```diff
@@ -0,0 +1,27 @@
...
+	[Native]
+	public enum BCChatButtonStyle : nint {
+		Light = 0,
+		Dark,
+	}
+
+	[Mac (10,13, onlyOn64: true), iOS (11, 3)]
```
spouliot: same 1013,4

src/BusinessChat/Enums.cs
```diff
@@ -0,0 +1,27 @@
...
+	public enum BCParameterName {
+
+		[Field ("BCParameterNameIntent")]
+		Intent,
+
+		[Mac (10, 13, 4, onlyOn64: true), iOS (11, 3)]
```
spouliot: you can remove this (same as type)

src/BusinessChat/Enums.cs
```diff
@@ -0,0 +1,27 @@
...
+		[Field ("BCParameterNameGroup")]
+		Group,
+
+		[Mac (10, 13, 4, onlyOn64: true), iOS (11, 3)]
```
spouliot: you can remove this (same as type)

src/businesschat.cs
```diff
@@ -0,0 +1,44 @@
...
+using XamCore.Foundation;
+using XamCore.ObjCRuntime;
+#if MONOMAC
+using XamCore.AppKit;
+#else
+using XamCore.UIKit;
+#endif
+
+namespace XamCore.BusinessChat {
+	[Mac (10,13, onlyOn64: true), iOS (11,3)]
```
spouliot: 10,13,4 everywhere
quite sure our intro tests will fail otherwise

src/businesschat.cs
```diff
@@ -0,0 +1,44 @@
...
+#if MONOMAC
+	[BaseType (typeof(NSControl))]
+#else
+	[BaseType (typeof(UIControl))]
+#endif
+	interface BCChatButton {
+		[Mac (10,13,4, onlyOn64: true), iOS (11,3)]
```
spouliot: you can remove it (since it will be identical to the type)

src/businesschat.cs
```diff
@@ -0,0 +1,44 @@
...
+		[Export ("initWithStyle:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (BCChatButtonStyle style);
+	}
+
+
+	[Mac (10,13, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof(NSObject))]
+	interface BCChatAction {
```
spouliot: that looks like `[DisableDefaultCtor]` is missing
there's no point in creating an instance of this type if there's nothing to call

src/businesschat.cs
```diff
@@ -0,0 +1,44 @@
...
+	[Mac (10,13, onlyOn64: true), iOS (11,3)]
+#if MONOMAC
+	[BaseType (typeof(NSControl))]
+#else
+	[BaseType (typeof(UIControl))]
+#endif
+	interface BCChatButton {
+		[Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		[Export ("initWithStyle:")]
+		[DesignatedInitializer]
```
spouliot: if that's the default initializer then the type should have `[DisableDefaultCtor]`

tools/mtouch/Makefile
```diff
@@ -277,7 +277,8 @@ SIMLAUNCHER_FRAMEWORKS =  \
 						\
 	-weak_framework ARKit					\
 	-weak_framework FileProvider			\
-	-weak_framework FileProviderUI
+	-weak_framework FileProviderUI			\
+	-weak_framework BusinessChat
```
spouliot: your missing the code to load the framework for XM
look at `src/Foundation/NSObject.mac.cs`

src/BusinessChat/BCChatAction.cs
dalexsoto: I would go one step further, there is no need for the managed representation of a `NSString` so I would create an array of IntPtrs with`NSString.CreateNative` and create the NSMutableArray from those instead just don't forget to release them with `NSString.ReleaseNative`

src/businesschat.cs
```diff
@@ -0,0 +1,44 @@
...
+using XamCore.ObjCRuntime;
+#if MONOMAC
+using XamCore.AppKit;
+#else
+using XamCore.UIKit;
+#endif
+
+namespace XamCore.BusinessChat {
+	[Mac (10,13, onlyOn64: true), iOS (11,3)]
+#if MONOMAC
```
dalexsoto: You can reduce the number of `#if MONOMAC` here since you are just using one on the top just use

```
#if MONOMAC
using XamCore.AppKit;
using UIControl = XamCore.AppKit.NSControl;
#else
using XamCore.UIKit;
#endif
```

and just have `[BaseType (typeof (UIControl))]` for  `BCChatButton`.

src/BusinessChat/BCChatAction.cs
mandel-macaque: Grab that when I tried to do it via a Wrap, will fix it.

spouliot:
```
make[3]: *** [Xamarin.Mac.registrar.mobile.i386.m] Error 1
14:24:33 make[3]: *** Waiting for unfinished jobs....
14:24:33 error MM4162: The type 'BusinessChat.BCChatButtonStyle' (used as a parameter in BusinessChat.BCChatButton..ctor) is not available in macOS 10.13.2 (it was introduced in macOS 10.13.4). Please build with a newer macOS SDK (usually done by using the most recent version of Xcode).
```

@chamons why is not that check made with the SDK version ? and not the OS version ?

rolfbjarne:
We probably need a new entry for Xcode 9.3 -> macOS 10.13.4 here: https://github.com/xamarin/xamarin-macios/blob/0dd292f3d1e4d67cb2c60eba1dd84e5a613475b4/tools/mmp/driver.cs#L631-L648

spouliot:
Ok, that needs to be added to the TODO on new Xcode releases because it would seem we also miss 9.1/10.13.1 mapping :|

@mandel-macaque please add both missing entries in this PR
@chamons. please check on master/d15-7 and do a PR if needed. 

Thanks!

mandel-macaque:
Failures are to unrelated tests (https://github.com/xamarin/maccore/issues/652).

mandel-macaque:
We also experience the issue https://github.com/xamarin/maccore/issues/645 so merging since it is unrelated.

PR #3596, opened 2018-02-26
[system-dependencies] simplify mono version check

lewurm:


akoeplinger:
It was added with https://github.com/mono/mono/commit/eedeb428d604e37fbe2934c2299be6166e903049 in MOno 5.2

PR #3598, opened 2018-02-26
[tests] Fix ProtocolTest to work with in both monotouch-test and linkall.

rolfbjarne:
The [Protocol] attributes the test checks for are not linked away in
monotouch-test (because monotouch-test only links SDK assemblies), but they
are linked away in link all, so adjust the test accordingly.


rolfbjarne:
Test failure is unrelated:

* apitest/Classic: [apitest hangs](https://github.com/xamarin/maccore/issues/653)

PR #3601, opened 2018-02-26
[xharness] Always build the test libraries before doing anything else.

rolfbjarne:
Fixes an issue where multiple projects would try to build the test libraries simultaneously.

Fixes https://github.com/xamarin/maccore/issues/637.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -874,6 +874,7 @@ namespace xharness
 					populating = false;
 				}).Wait ();
 				GenerateReport ();
+				BuildTestLibraries ();
 				if (!IsServerMode) {
 					foreach (var task in Tasks)
 						tasks.Add (task.RunAsync ());
@@ -890,6 +891,11 @@ namespace xharness
 
 		public bool IsServerMode {
 			get { return Harness.JenkinsConfiguration == "server"; }
+		}
+
+		void BuildTestLibraries ()
+		{
+			ProcessHelper.ExecuteCommandAsync ("make", $"all -j{Environment.ProcessorCount} -C {StringUtils.Quote (Path.Combine (Harness.RootDirectory, "test-libraries"))}", MainLog, TimeSpan.FromMinutes (1)).Wait ();
 		}
 
 		Task RunTestServer ()


spouliot:
related to any known issue on the bots ?

rolfbjarne:
@spouliot yes, it fixes https://github.com/xamarin/maccore/issues/637 (I've updated the PR description).

tests/xharness/Jenkins.cs
```diff
@@ -874,6 +874,7 @@ public int Run ()
 					populating = false;
 				}).Wait ();
 				GenerateReport ();
+				BuildTestLibraries ();
 				if (!IsServerMode) {
 					foreach (var task in Tasks)
 						tasks.Add (task.RunAsync ());
```
mandel-macaque: I might be wrong here, ProcessHelper.ExecuteCommandAsync does return a task, correct? 

```csharp
public static Task<bool> PollForExitAsync (int pid, TimeSpan timeout)
```

The task is ignored in BuildTestLibraries that returns void (which we should only do in events).  So, I think we should be paying attention to the result of the task and, since it is async but needed, we might wait for it to success before we do anything else (there is now guaranteed AFAIK that the build will be done before we needed it).

I might be completely wrong here, but just wanted you to confirm is correct.

tests/xharness/Jenkins.cs
```diff
@@ -892,6 +893,11 @@ public int Run ()
 			get { return Harness.JenkinsConfiguration == "server"; }
 		}
 
+		void BuildTestLibraries ()
+		{
+			ProcessHelper.ExecuteCommandAsync ("make", $"all -j{Environment.ProcessorCount} -C {StringUtils.Quote (Path.Combine (Harness.RootDirectory, "test-libraries"))}", MainLog, TimeSpan.FromMinutes (1)).Wait ();
+		}
```
mandel-macaque: We should return a Task, not void.

rolfbjarne: I call `.Wait ()` at the end to make it synchronous.

rolfbjarne:
Test failure is unrelated:

* MSBuild/iOS: [Xamarin.iOS.Tasks.TargetTests.RebuildExecutable_NoModification ](https://github.com/xamarin/maccore/issues/592)

Issue #3207, opened 2018-01-12
Test failure on bots: Mono.Security/Mac/Modern: MonoTests.Mono.Security.Authenticode.AuthenticodeDeformatterTest.VerifySignedAssembly


rolfbjarne:
This is a consistent test failure on wrench with the Mono.Security tests (but only on modern/macOS):

```
1) VerifySignedAssembly (MonoTests.Mono.Security.Authenticode.AuthenticodeDeformatterTest.VerifySignedAssembly)
     Reason
  Expected: True
  But was:  False

  at MonoTests.Mono.Security.Authenticode.AuthenticodeDeformatterTest.VerifySignedAssembly () [0x0000f] in /Users/builder/data/lanes/1381/c7c356c3/source/xamarin-macios/external/mono/mcs/class/Mono.Security/Test/Mono.Security.Authenticode/AuthenticodeDeformatterTest.cs:703 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in /Library/Frameworks/Xamarin.Mac.framework/Versions/4.3.0.36/src/Xamarin.Mac/mcs/class/corlib/System.Reflection/MonoMethod.cs:305 

```

I'm not able to reproduce it locally, but the test fails if I copy the test .app to the bot and run it there (which rules out build differences).

Being able to change the test app locally also made it possible to get more information:

* This is the assert that fails: https://github.com/mono/mono/blob/9484143e9389dba5feba3aec6f4df25c6f1a00fb/mcs/class/Mono.Security/Test/Mono.Security.Authenticode/AuthenticodeDeformatterTest.cs#L703. Additional logging shows that "ad.Reason = 1" when the assert fails.
* I ran also traced exceptions, and this shows up on the bots, but not locally:

```
[0x700001007000:] EXCEPTION handling: System.NullReferenceException: Object reference not set to an instance of an object

"Threadpool worker" tid=0x0x700001007000 this=0x0x103bd4508 , thread handle : 0x7fbee2f35d10, state : not waiting
[...]

PR #3603, opened 2018-02-26
[tests][macos] Fix AuthenticodeDeformatterTest.VerifySignedAssembly failure on modern. Fixes 3207

spouliot:
Modern does not, by default/design, ship a machine.config file. This
means it only knowns about the default crypto shipped with .NET.

That's a problem for MD2 for which some old certificates might still
exists in the computer store, like our wrench bots.

That's generally not a problem since XM apps delegate trust to macOS
except for one case: Authenticode.

So verifying an authenticode signature can end up, thru X509Chain,
loading a certificate using MD2 without knowing how to create the
digest algorithm - and fail.

The fix is to register the algorithm manually (if not found).

https://github.com/xamarin/xamarin-macios/issues/3207

### DIFF
--- a/tests/common/mac/MacTestMain.cs
+++ b/tests/common/mac/MacTestMain.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using Mono.Security.Cryptography;
 #if XAMCORE_2_0 || __UNIFIED__
 using AppKit;
 using Foundation;
@@ -25,6 +27,8 @@ namespace Xamarin.Mac.Tests
 #if !NO_GUI_TESTING
 			NSApplication.Init();
 #endif
+			if (CryptoConfig.CreateFromName ("MD2") == null)
+				CryptoConfig.AddAlgorithm (typeof (MD2Managed), "MD2");
 			RunTests (args);
 		}
 
--- a/tests/common/mac/MacTestMain.cs
+++ b/tests/common/mac/MacTestMain.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
-using Mono.Security.Cryptography;
 #if XAMCORE_2_0 || __UNIFIED__
 using AppKit;
 using Foundation;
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Is there no way to fix this inside the BCL instead of in only the test?

The annoying part is that the mere presence of a certificate using MD2 makes any authenticode verification fail.

spouliot:
@rolfbjarne no - that would introduce other compatibility issues for something that is a corner case (of Mono.Security, not the BCL).

The crypto mappings are dynamic by design. We break this in XI (full AOT) by making it static (apps can't be extended) but it's not a good option for XM (any profile) since they can be extended (e.g. code download or generation).

XM/full (and desktop mono) complete the BCL mappings set by providing, by default, a `machine.config` which includes this. However that's not linker friendly - but not a problem for XM/full since it's not allowed on that profile.

The last option is to add code (like this PR) to extend the crypto mappings that the BCL ships. This has the advantage of being linker friendly - which a XM/modern app allows and would fail with the `machine.config` option.

rolfbjarne:
@spouliot this would also solve the problem: https://github.com/rolfbjarne/mono/commit/b9a2815e3c5319c6d1f6404e1d5cd7df9d354ef5 (fail to load certificates whose hash function is unavailable). That means the remaining (valid) certificates can still be used (and the test passes, because it doesn't use the MD2-based certificate, but another one). Then if someone really needs the MD2-based certificate, they can execute your code first, and it should work.

spouliot:
That's a (behavioural) breaking change [1] that affects several parts of mono, e.g. not every platform delegates the TLS trust to the OS (like we do in iOS and macOS).

[1] because we can load such certificates today (we just can build a chain with them) and such code would not get exceptions.

rolfbjarne:
My problem with this is that if our customer's customers run into this, it will be pretty much impossible for anybody to figure out the problem, it was hard enough for me to track it down 😞

I have another idea, IIRC this is somewhere in the middle of where things start to go wrong, the NRE is thrown inside `x509.Hash`: https://github.com/mono/mono/blob/b8da9736209427ef49267923e1321c801c65dfd2/mcs/class/Mono.Security/Mono.Security.X509/X509CertificateCollection.cs#L126

Which means this could be a solution (untested):

```csharp
    for (int i=0; i < InnerList.Count; i++) {
        X509Certificate x509 = (X509Certificate) InnerList [i];
        byte[] otherHash;
        try {
            otherHash = x509.Hash;
        } catch {
            // Ignore any exceptions from x509.Hash, if something goes wrong there it's not the same certificate.
            continue;
        }
        if (Compare (otherHash, hash))
            return i;
    }
```

I can't see how this would be a behavioral breaking change.

spouliot:
@rolfbjarne it's incomplete as it cover only half the problem (and make it even harder to diagnose the 2nd half) but I see your point :)

Let's split this into two parts
1. fixing the bots with this PR - which is an immediate issue;
2. let's fix mono and, eventually, we'll bump to include the fix

spouliot:
for the mono/bcl side https://github.com/mono/mono/issues/7322

rolfbjarne:
Test failures are unrelated:

* monotouch-test: [ProtocolTest failures](https://github.com/xamarin/maccore/issues/655)
* MMP regression tests: [random failure](https://github.com/xamarin/maccore/issues/645)

PR #3610, opened 2018-02-27
Bump mono and touchunit

lewurm:
this should fix https://github.com/xamarin/maccore/issues/651

Bump mono

Commit list for mono/mono:

* mono/mono@331f927 Bump nunitlite (#7276)
* mono/mono@da6a45a Check if we are exceeding the worker count limit
* mono/mono@21d6b33 Ios sdk changes (#7258)
* mono/mono@e87d123 Merge mxe-Win32 and mxe-Win64 targets since we are building both in XA anyway (#7280)
* mono/mono@b9fb4d8 Bump bockbuild to get curl error changes
* mono/mono@b7031bc Emit the same instances in hybrid aot mode as in full aot mode.
* mono/mono@3538148 [2017-12] [xbuild] add property that is available in msbuild (#7224)
* mono/mono@947e457 [sdks] Instead of using a XCODE_VERSION variable, let the caller set the XCODE_DIR variable directly, since x.ios is using a full path. (#7111)
* mono/mono@7f0c97a Bump bockbuild for mono/bockbuild#62 (#7207)
* mono/mono@85cfb9c Disable partial sharing during JITting, its only been tested in full aot mode, and doesn't seem to work with JITted methods. Fixes mono/mono#7085.
* mono/mono@22a6ec1 Add hybrid aot tests.
* mono/mono@454bd20 Remove reflection calls for value has to be set

Diff: https://github.com/mono/mono/compare/4adf8229e88ffeed60a9b63d402e9d074323f13b...331f927bfea0c6f3c8dd5fdc091f0b6d0f6aafa9

Bump Touch.Unit

Commit list for spouliot/Touch.Unit:

* spouliot/Touch.Unit@1c41c75 Revert "[TouchRunner] do not use FinallyDelegate in Runner, because it requires remoting API to be available (#49)" (#50)

Diff: Diff: https://github.com/spouliot/Touch.Unit/compare/4ac4e325001e97a4f1874d38183eddf78d4ca334...1c41c75983ed840e9dc7079b5d4b0c31d1565359


rolfbjarne:
Test failures are unrelated:

* monotouch-test: [ProtocolTest failures](https://github.com/xamarin/maccore/issues/655)
* MMP regression tests: [random failure](https://github.com/xamarin/maccore/issues/645)

PR #3616, opened 2018-02-28
[runtime] Fix exception marshalling when the dynamic registrar is removed.

rolfbjarne:
This can be reproduced by running monotouch-test with all optimizations in
Debug mode (because some of the exception marshalling tests are only enabled
in debug mode), so add such a configuration to xharness. To avoid bloating PR
builds, this configuration is only enabled when running all tests (either
manually selecting all tests for a PR, or on Wrench, where everything is
always tested).

### DIFF
--- a/runtime/delegates.t4
+++ b/runtime/delegates.t4
@@ -47,7 +47,7 @@
 			"int", "int", "exc_handle"
 		) {
 			WrappedManagedFunction = "UnwrapNSException",
-			OnlyDynamicUsage = true,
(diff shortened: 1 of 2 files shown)


rolfbjarne:
Test failures are unrelated:

* [FAIL] ProtocolTest: https://github.com/xamarin/maccore/issues/655
* [FAIL] CalendarTest.TestAddingByComponents : DateBySettingUnit - One day from now should be tomorrow: https://github.com/xamarin/maccore/issues/656

PR #655, opened 2016-08-24
[Intents] Fixes bug 43205 - SiriKit API Inconsistencies

dalexsoto:
https://bugzilla.xamarin.com/show_bug.cgi?id=43205

Per [documentation](1) INIntentResolutionResult is an abstract class
and now we honor that, also we now provide the right return type
on each of the properties (NeedsValue, NotRequired and Unsupported)
in all subclasses of INIntentResolutionResult.

Unit tests added for all INIntentResolutionResult subclasses.

[1]: https://developer.apple.com/reference/intents/inintentresolutionresult?language=objc

PR #3620, opened 2018-02-28
[xharness] Propagate custom defines to referenced projects as well. Fixes maccore #655. (#3615)

rolfbjarne:
The ProtocolTest test is located in the bindings-test assembly, and it
requires certain conditional compilation defines to be set in order to behave
properly.

With this fix we correctly set these defines when cloning the bindings-test
project (in addition to when we set the defines for the main project, which is
either monotouch-test or linkall for this test).

Fixes https://github.com/xamarin/maccore/issues/655.


rolfbjarne:
Test failure is unrelated:

* [monotouch-test] [FAIL] [CalendarTest.TestAddingByComponents : DateBySettingUnit - One day from now should be tomorrow fail](https://github.com/xamarin/maccore/issues/656)

PR #3622, opened 2018-03-01
[mtouch] Propagate the computed value for removal of the dynamic registrar to code shared app extensions.

rolfbjarne:
This fixes a startup crash in code shared app extensions due to having the
wrong value set in the runtime (the dynamic registrar was removed, but the
executable didn't know it).


rolfbjarne:
Test failure is unrelated:

PR #3623, opened 2018-03-01
[static registrar] Optimize creation of delegates for blocks.

rolfbjarne:
Optimize creation of delegates for blocks so that it doesn't require the
dynamic registrar.

This is done by getting the metadata token for the Create method that creates
the delegate, and embed that metadata token in the generated code from the
static registrar.

Also add tests, since this scenario was not covered by tests already.


tools/linker/CoreMarkStep.cs
```diff
@@ -271,67 +271,13 @@ protected override MethodDefinition MarkMethod (MethodReference reference)
...
-								// one cannot simply mark the `ca.ConstructorArguments [0].Value` type,
-								// e.g. Trampolines.NIDActionArity1V26
-								// as the relation to *_Extensions will be reflected at runtime
-								MarkMethod (m);
-							}
-						}
-					}
-				}
+			if (LinkContext.Target.StaticRegistrar.MapProtocolMember (method, out var extensionMethod)) {
+				// one cannot simply mark the `ca.ConstructorArguments [0].Value` type,
```
rolfbjarne: This comment isn't quite true anymore, so I created an issue to fix it: #3624. I don't want to do it in this PR, since we'll want this PR in d15-7 and fixing this issue would be an enhancement (with its own inherent risks, since it would end up removing more cod).

tools/common/StaticRegistrar.cs
```diff
@@ -4001,6 +4054,145 @@ void Specialize (AutoIndentStringBuilder sb, ObjCMethod method, List<Exception>
...
+							continue;
+						if (pMethod.Name != method.Name)
+							continue;
+						if (!TypeMatch (pMethod.ReturnType, method.ReturnType))
+							continue;
+						if (pMethod.Parameters.Length != method.Parameters.Count)
+							continue;
+						for (int i = 0; i < pMethod.Parameters.Length; i++)
+							if (!TypeMatch (pMethod.Parameters [i], method.Parameters [i].ParameterType))
+								continue;
```
spouliot: don't we have this matching code somewhere else ? I don't recall it (I know there were helpers methods in gendarme) but we compare methods in many places

tools/linker/MobileExtensions.cs
```diff
@@ -89,7 +89,7 @@ static bool HasGeneratedCodeAttribute (ICustomAttributeProvider provider, Derive
 					return rv;
 			}
 
-			return GetBindingImplAttribute (provider, context?.GetCustomAttributes (provider, Namespaces.ObjCRuntime, "BindingImplOptions"));
+			return GetBindingImplAttribute (provider, context?.GetCustomAttributes (provider, Namespaces.ObjCRuntime, "BindingImplAttribute"));
```
spouliot: that must have cause other issues, right ?

tools/common/StaticRegistrar.cs
rolfbjarne: Yes, we have similar logic in other places, the problem is that here we're comparing instances of `MethodDefinition` and `ObjCMethod`, so those other places can't be used as-is because they're comparing two instances of the same type.

tools/linker/MobileExtensions.cs
rolfbjarne: Should have, but as far as I can tell, didn't 😒

rolfbjarne:
Test failures are unrelated

* mscorlib/Mac: [MonoTests.System.Security.Cryptography.DSACryptoServiceProviderTest.Persistence_PersistKeyInCsp_False](https://github.com/xamarin/maccore/issues/640)
* mscorlib/iOS 32-bit: [TimeZoneTest.TestCtors (D02)](https://github.com/xamarin/maccore/issues/629)

PR #3626, opened 2018-03-01
[mtouch] Bundle duplicated assemblies in the container app. Fixes #58873.

rolfbjarne:
Assemblies that satisfy all of these conditions:

* Are not in the container project.
* Are in multiple app extensions.

can (and should) be bundled in the container app.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=58873


rolfbjarne:
build

PR #3632, opened 2018-03-01
[xtro] Avoid duplicate 'close' p/invoke across frameworks

spouliot:
Depending on the metadata order the symbol will be reported against a
different framework and cause false positives.

Re-using one of the declaration solves this - even if they should,
ideally, be moved to a separated/shared type (like one per fx)

reference:
https://github.com/xamarin/maccore/issues/638

note: a different fix will be needed on master wrt XamCore changes


spouliot:
It's actually already fixed in master (and why it rang a bell - just not hard enough for me to hear)

PR #3645, opened 2018-03-02
[tests] Use NUnit.Runners v2.6.4 for the install sources tests.

rolfbjarne:
Since that's what all the other test suites that use NUnit v2 use, and it's
also what xharness expects (and tries to execute).

This fixes running these tests when other tests haven't already restored the
v2.6.4 nuget.


rolfbjarne:
Test failure is unrelated:

PR #3660, opened 2018-03-05
[xharness] Don't count the skipped tests as failed

VincentDondain:
Currently we consider the failed **and** the skipped tests as failures which is confusing. The tests didn't all "fail" because some were just not ran.

This PR updates the summary message to be a little more clear.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -1367,9 +1367,10 @@ namespace xharness
 					markdown_summary.Write ($"# Test run in progress: ");
 					markdown_summary.Write (string.Join (", ", list));
 					markdown_summary.WriteLine ();
-				} else if (failedTests.Any () || skippedTests.Any ()) {
-					var skipped = skippedTests.Any () ? $" and {skippedTests.Count ()} tests skipped" : "";
-					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {passedTests.Count ()} tests passed{skipped}.");
+				} else if (failedTests.Any ()) {
+					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {skippedTests.Count ()} tests skipped, {passedTests.Count ()} tests passed.");
+				} else if (skippedTests.Any ()) {
+					markdown_summary.WriteLine ($"{skippedTests.Count ()} tests skipped, {passedTests.Count ()} tests passed.");
 				} else if (passedTests.Any ()) {
 					markdown_summary.WriteLine ($"# All {passedTests.Count ()} tests passed");
 				} else {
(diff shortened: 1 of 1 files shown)


tests/xharness/Jenkins.cs
```diff
@@ -1367,7 +1368,8 @@ void GenerateReportImpl (Stream stream, StreamWriter markdown_summary = null)
 					markdown_summary.Write (string.Join (", ", list));
 					markdown_summary.WriteLine ();
 				} else if (failedTests.Any ()) {
-					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {passedTests.Count ()} tests passed.");
+					var skipped = skippedTests.Any () ? $" and {skippedTests.Count ()} tests skipped" : "";
+					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {passedTests.Count ()} tests passed{skipped}.");
```
rolfbjarne: This won't be printed if there are only skipped tests, in which case it'll just show "All ### tests passed" (the next if just below), which is wrong. I think you need to change the condition above to:

```csharp
} else if (failedTests.Any () || skippedTests.Any ()) {
```

VincentDondain:
Known issues: https://github.com/xamarin/maccore/issues/661 and https://github.com/xamarin/maccore/issues/658.

@rolfbjarne can you approve?

tests/xharness/Jenkins.cs
```diff
@@ -1366,8 +1367,9 @@ void GenerateReportImpl (Stream stream, StreamWriter markdown_summary = null)
 					markdown_summary.Write ($"# Test run in progress: ");
 					markdown_summary.Write (string.Join (", ", list));
 					markdown_summary.WriteLine ();
-				} else if (failedTests.Any ()) {
-					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {passedTests.Count ()} tests passed.");
+				} else if (failedTests.Any () || skippedTests.Any ()) {
+					var skipped = skippedTests.Any () ? $" and {skippedTests.Count ()} tests skipped" : "";
+					markdown_summary.WriteLine ($"{failedTests.Count ()} tests failed, {passedTests.Count ()} tests passed{skipped}.");
```
spouliot: The logic does not read well, likely why there's 2 calls to `skippedTests.Any ()`

It looks like what we want is
```
if (fail) { show failures + skipped }
else if (skipped) { show skipped }
else if (pass) { show passed }
...
```

spouliot: not that `Any` followed by `Count` is much better ;-)

VincentDondain:
I also added the same fix for an other part of the file that was identical.

tests/xharness/Jenkins.cs
```diff
@@ -1694,7 +1695,7 @@ function toggleAll (show)
 				var headerColor = "black";
 				if (unfinishedTests.Any ()) {
 					; // default
-				} else if (failedTests.Any ()) {
+				} else if (failedTests.Any () || skippedTests.Any ()) {
```
spouliot: might want to pick another colour (like orange) since the point of the PR is to separate failures from unknowns

VincentDondain: Fair enough.

PR #3673, opened 2018-03-06
[ARKit][Contacts] Update to Xcode 9.3 Beta 4

dalexsoto:
https://github.com/xamarin/xamarin-macios/wiki/Contacts-watchOS-xcode9.3-beta4
https://github.com/xamarin/xamarin-macios/wiki/ARKit-iOS-xcode9.3-beta4

### DIFF
--- a/src/Contacts/CNEnums.cs
+++ b/src/Contacts/CNEnums.cs
@@ -85,7 +85,6 @@ namespace XamCore.Contacts {
 		CommunicationError = 1,
 		DataAccessError = 2,
 		AuthorizationDenied = 100,
-		[iOS (11,3), Mac (10,13,3)]
 		NoAccessableWritableContainers = 101,
 		RecordDoesNotExist = 200,
 		InsertedRecordAlreadyExists = 201,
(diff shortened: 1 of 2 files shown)


src/Contacts/CNEnums.cs
```diff
@@ -85,6 +85,8 @@ public enum CNErrorCode : nint {
 		CommunicationError = 1,
 		DataAccessError = 2,
 		AuthorizationDenied = 100,
+		[iOS (11,3), Mac (10,13,3)]
```
VincentDondain: Interesting it's marked as 10.13.3 instead of 10.13.4 which is the new beta. As long as the tests pass though...

dalexsoto: Yep, I wanted to stick to headers since this particular case there is not much of an issue

spouliot: availability attributes on error codes are not very useful (e.g. introspection can't use it) and we generally ignore them
the reality is that the contract is the (enum) type, not the value(s)

dalexsoto: Yeah, I still think it is a good thing to have because of IDE help on this, at least it raises awareness on when it was introduced

VincentDondain: Oh yea didn't realize when reviewing it was an error code. Should be removed IMHO.

dalexsoto:
Closing it in favor of the next 3 PRs:
* https://github.com/xamarin/xamarin-macios/pull/3678

PR #3683, opened 2018-03-07
[tests] Don't try to compile the Main function in F# extensions.

rolfbjarne:
The F# compiler complains that:

    /Users/xamarinqa/vsts/_work/52/s/tests/fsharp/Main.fs(13,9): error FS0433: A function labeled with the 'EntryPointAttribute' attribute must be the last declaration in the last file in the compilation sequence.

Main.fs is the last file in the project file, but the MSBuild tasks adds
another one at the end:

    obj/iPhone/Debug64-today-extension/Xamarin.iOS,Version=v1.0.AssemblyAttribute.fs

because we're building a library (in which case the MSBuild tasks assume that
there won't be any Main functions in the project, and as such it's safe to
append files to compile).

Work around this by excluding the Main function from F# extensions, it's not
needed anyway.

### DIFF
--- a/tests/fsharp/Main.fs
+++ b/tests/fsharp/Main.fs
@@ -6,7 +6,7 @@ open UIKit
 open MonoTouch.UIKit
 #endif
 
-#if !__WATCHOS__
+#if !__WATCHOS__ && !TODAY_EXTENSION
 
 module Main = 
     [<EntryPoint>]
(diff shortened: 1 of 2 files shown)


rolfbjarne:
Test failure is unrelated:

* xammac tests: https://github.com/xamarin/maccore/issues/658.

PR #3696, opened 2018-03-07
[monotouch-test] Stop MidiThruConnectionTests.FindTest from randomly failing

dalexsoto:
Fixes xamarin/maccore#658

When a MidiThruConnection is created but for some reason is not disposed
the system keeps it alive even between app/simulator restarts, if you want
to clean the connections you must reset contents and settings from simulator.

In order to avoid this test from failing randomly and since the intent of the
test is to check if `MidiThruConnection.Find` works we change the assert to
`>= 2` since this is at least the number of connections we expect.

### DIFF
--- a/tests/monotouch-test/CoreMidi/MidiThruConnectionTests.cs
+++ b/tests/monotouch-test/CoreMidi/MidiThruConnectionTests.cs
@@ -131,11 +131,10 @@ namespace MonoTouchFixtures.CoreMidi {
 
 			using (var connection1 = MidiThruConnection.Create ("com.xamarin.midi", cnnParams1))
 			using (var connection2 = MidiThruConnection.Create ("com.xamarin.midi", cnnParams2)) {
-				MidiError err;
-				var connections = MidiThruConnection.Find ("com.xamarin.midi", out err);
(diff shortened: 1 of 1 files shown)


spouliot:
an alternative would be to use something like:
```
var name = `com.xamarin.midi.{DateTime.UtcNow.Ticks}`;
...
```
and Find them, so older left over connections won't affect the outcome.

dalexsoto:
@spouliot do we really need that change? I mean the intent of the test is to check if you get something out of `MidiThruConnection.Find`, since we use an identifier `com.xamarin.midi` those connections were created by us anyways.

spouliot:
@dalexsoto I disagree - even if I did not write the test ;-)

In specific: the test adds two connections and tries to find them. It obviously wants to find the ones that were *just* added, not ones added in the past executions. Also, even if it was not the original goal, it does test that the creation (and not just finding) worked properly.

In general: the fix dilute the test value (it might not catch a `Create` failure now) because something, that happened once over years, has gone wrong. That's not a good strategy. We should aim for more resilient tests whenever possible. I understand it's not always possible (or can prohibitively too costly to implement) but this is not the case here.

dalexsoto:
Fair enough! Making the suggested change!

PR #3734, opened 2018-03-13
[mtouch] Disable minor collections for file provider extensions.

rolfbjarne:
Works around a mono issue: https://github.com/mono/mono/issues/7597.

This fixes the test case from https://bugzilla.xamarin.com/show_bug.cgi?id=60597.


rolfbjarne:
This is just a fix for the file provider extensions, at some point we should take the time to check all extensions and their memory limits.

rolfbjarne:
I'll come up with something more general.

PR #3735, opened 2018-03-13
[security] Adding new SecAccessControlCreateFlags

### LINKED ISSUE #3732: [security] Deprecate SecAccessControlCreateFlags's 'TouchIDAny' and 'TouchIDCurrentSet' enum values (Xcode 9.3)
Looks like we missed (or skipped) the deprecation of some enum values (they got renamed).

See [API diff](https://gist.github.com/VincentDondain/682467b301bf98583b87db6ae5650251).
Lasted [xcode9.3 code](https://github.com/xamarin/xamarin-macios/blob/xcode9.3/src/Security/SecAccessControl.cs#L23-L53).
[Binding status](https://github.com/xamarin/xamarin-macios/wiki/xcode9.3-Bindings-Status) says the framework is up to date with Xcode 9.3 beta 4.

We should follow Apple on this one, deprecate the old enum values and make the new values equal to the old ones.

Otherwise it could be confusing in the future if documentation points to the new names.

mdbech:
Should fix #3732.

### DIFF
--- a/src/Security/SecAccessControl.cs
+++ b/src/Security/SecAccessControl.cs
@@ -36,14 +36,14 @@ namespace XamCore.Security {
 		[Deprecated (PlatformName.MacOSX, 10, 13, 4, "Use 'BiometryAny' instead.")]
 		[Deprecated (PlatformName.TvOS, 11, 3, message: "Use 'BiometryAny' instead.")]
 		[Deprecated (PlatformName.WatchOS, 4, 3, message: "Use 'BiometryAny' instead.")]
-		TouchIDAny          = 1 << 2,
+		TouchIDAny          = BiometryAny,
 
 		[iOS (9,0)][Mac (10,12,1)]
 		[Deprecated (PlatformName.iOS, 11, 3, message: "Use 'BiometryCurrentSet' instead.")]
 		[Deprecated (PlatformName.MacOSX, 10, 13, 4, "Use 'BiometryCurrentSet' instead.")]
 		[Deprecated (PlatformName.TvOS, 11, 3, message: "Use 'BiometryCurrentSet' instead.")]
 		[Deprecated (PlatformName.WatchOS, 4, 3, message: "Use 'BiometryCurrentSet' instead.")]
-		TouchIDCurrentSet   = 1 << 5,
+		TouchIDCurrentSet   = BiometryCurrentSet,
 
 		[iOS (11, 3), Mac (10, 13, 4), TV (11, 3), Watch (4, 3)]
 		BiometryAny         = 1 << 1,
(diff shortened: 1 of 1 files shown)


VincentDondain:
build

src/Security/SecAccessControl.cs
```diff
@@ -32,10 +32,24 @@ public enum SecAccessControlCreateFlags : nint {
...
 
 		[iOS (9,0)][Mac (10,12,1)]
-		TouchIDCurrentSet   = 1 << 3,
+		[Deprecated (PlatformName.iOS, 11, 3, message: "Use 'BiometryCurrentSet' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 13, 4, "Use 'BiometryCurrentSet' instead.")]
+		[Deprecated (PlatformName.TvOS, 11, 3, message: "Use 'BiometryCurrentSet' instead.")]
+		[Deprecated (PlatformName.WatchOS, 4, 3, message: "Use 'BiometryCurrentSet' instead.")]
+		TouchIDCurrentSet   = BiometryCurrentSet,
+
+		[iOS (11, 3), Mac (10, 13, 4), TV (11, 3), Watch (4, 3)]
```
spouliot: That's literally true but it's not *really* true :)

Once compiled there's no difference between the two enum member names. So `BiometryAny` will work in iOS 9.0 because `2` is something iOS9 will accept (whatever the header files).

So marking the members as available from iOS 11,3 actually induce people in using the older values, e.g. to have their code compatible with earlier version. Developers are not likely to notice the values are identical, so it's the "binders" job to help them avoid code like:

```
var value = ios11_3 ? .BiometryCurrentSet : .TouchIDCurrentSet;
```

or see deprecation warnings in the IDE.

The best thing is to copy the original availability attributes, i.e. `[iOS (9,0)][Mac (10,12,1)]` to the new names and to add a comment about this.

src/Security/SecAccessControl.cs
```diff
@@ -32,10 +32,24 @@ public enum SecAccessControlCreateFlags : nint {
...
+		BiometryAny         = 1 << 1,
+
+		[iOS (11, 3), Mac (10, 13, 4), TV (11, 3), Watch (4, 3)]
+		BiometryCurrentSet  = 1 << 3,
```
spouliot: same

src/Security/SecAccessControl.cs
VincentDondain: Ah yes I totally missed that. Thanks @spouliot for the clarification.

VincentDondain:
So we can use either use `[Advice ("Message")]` or `[Obsolete ("Message")]` where message could be something like: "'BiometryAny' is preferred over 'TouchIDAny' since Xcode 9.3. Touch ID and Face ID together are biometric authentication mechanisms."

`Advice` is of "info" severity (blue) and `Obsolete` is of "warning" severity (orange). Also obsolete will show a message of this format: "'SecAccessControlCreateFlags.TouchIDAny' is obsolete: 'message'".

@spouliot any preference? I personally vote for `Advice`.

dalexsoto:
My vote is on Advice aswell, this is the perfect scenario for it 👍

VincentDondain:
build

VincentDondain:
build

VincentDondain:
build

VincentDondain:
Thanks, merging.

PR #3739, opened 2018-03-13
[monotouch-tests] Adds mono's WeakAttribute tests

dalexsoto:


dalexsoto:
@spouliot good news and bad news

Good news ✅:
* I modified the sample to cope with watchOS.

Bad news 😞:
* It seems `WeakAttribute` does not work on watchOS, is this expected? or should I file an issue? 
* I had to disable the test for watchOS while it gets fixed so we do not break our tests.

dalexsoto:
OTOH this particular test might be incorrect for watchOS's GC?? I have tested on simulator and got the failure results like if `WeakAttribute` isn't really working, I have tried to test on device but I have not been able to make IDE to detect my watch again... What do you think @rolfbjarne?

I have created this smaller watchOS app in case anyone is interested to test (d15-7 is needed): 
[weakattributewatch.zip](https://github.com/xamarin/xamarin-macios/files/1809555/weakattributewatch.zip)

dalexsoto:
Ahh interesting bit of information, 32 bits tests are failing too maybe this is why it did not work on watchOS in the first place...

tests/monotouch-test/mono/WeakReferenceTest.cs
```diff
@@ -0,0 +1,110 @@
...
+
+			GC.Collect ();
+			GC.WaitForPendingFinalizers ();
+			Assert.That (MyParentView.count, Is.LessThan (totalTestObjects), "No retain cycles expected");
+		}
+
+		[Test]
+		public void RetainCyclesExpectedTest ()
+		{
+			for (int i = 0; i < totalTestObjects; i++) {
```
rolfbjarne: You're not setting `MyParentView.count` to zero here, which means that its result depends on the result of the other test.

tests/monotouch-test/mono/WeakReferenceTest.cs
```diff
@@ -0,0 +1,110 @@
...
+				var parent = new MyParentView (useWeak: false);
+				parent.TouchButton ();
+			}
+
+			GC.Collect ();
+			GC.WaitForPendingFinalizers ();
+			Assert.That (MyParentView.count, Is.EqualTo (totalTestObjects), "Retain cycles expected");
```
rolfbjarne: You're leaking the MyParentView instances you created here :smile:

spouliot:
@dalexsoto watchOS uses coop - so maybe things differs, please file and link to issue
c.c. @vargaz

tests/monotouch-test/mono/WeakReferenceTest.cs
spouliot: it would explain the 
```
[FAIL] WeakReferenceTest.RetainCyclesExpectedTest :   Retain cycles expected
  Expected: 100
  But was:  200
```
on 32 bits :)

dalexsoto:
@spouliot / @rolfbjarne Can I get a second look? Thanks!!

PR #3743, opened 2018-03-14
[msbuild] Fix detecting already signed executables.

rolfbjarne:
The MSBuild tasks will codesign an executable if the executable's timestamp is
later than `_CodeSignature/CodeResources`'s timestamp (or if
`_CodeSignature/CodeResources` doesn't exist).

Unfortunately, the codesign executable modifies both of those files, and the
executable last. This means that even just after running codesign, the
executable's timestamp might be later than `_CodeSignature/CodeResources`'s
timestamp (due to HFS+'s one-second timestamp resolution, this might happen
all within the same second, which means that this is a random issue: the
problem only occurs if the executable was modified at least a second later
than `_CodeSignature/CodeResources`.)

So make sure to touch `_CodeSignature/CodeResources` after running codesign,
so that the next time a build occurs (with no modifications), we don't resign
needlessly.

Fixes this (random) test failure when running the MSBuild tests:

    1) Test Failure : Xamarin.iOS.Tasks.TargetTests.RebuildExecutable_NoModifications
         #1: ../MySingleView/bin/iPhoneSimulator/Debug/MySingleView.app/MySingleView
      Expected: 2017-11-30 10:04:20.000
      But was:  2017-11-30 10:04:22.000

Fixes https://github.com/xamarin/maccore/issues/592.

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
@@ -1856,6 +1856,11 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 			Condition="'$(IsMacEnabled)' == 'true' And ('$(_CanOutputAppBundle)' == 'true' And '%(_AppExtensionCodesignProperties.SigningKey)' != '') And Exists ('$(AppBundleDir)\..\%(_AppExtensionCodesignProperties.Identity).dSYM\Contents\Info.plist')"
 			Files="$(AppBundleDir)\..\%(_AppExtensionCodesignProperties.Identity).dSYM\Contents\Info.plist"
 		/>
+		<Touch
+			SessionId="$(BuildSessionId)"
+			Condition="'$(IsMacEnabled)' == 'true' And Exists ('$(_AppBundlePath)PlugIns\%(_AppExtensionCodesignProperties.Identity)\_CodeSignature\CodeResources')"
+			Files="$(_AppBundlePath)PlugIns\%(_AppExtensionCodesignProperties.Identity)\_CodeSignature\CodeResources"
+		/>
 	</Target>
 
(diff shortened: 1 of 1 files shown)


rolfbjarne:
@spouliot not all our bots, but the High Sierra bots should. This is a good question, because this should be 100% reproducible on APFS... yet it isn't.

Test failure from a High Sierra bot ([log](http://xamarin-storage/macios-mac-master/b2/b242afc8aa72092869bfd95c0610fcdcb31bab9d/Itchy-HighSierra/e76fb3ac0f32/jenkins-results/tests/MSBuild%20tests_268/execute-20180309_143001.txt), [wrench revision](https://wrench.internalx.com/Wrench/ViewLane.aspx?lane_id=1381&host_id=434&revision_id=931423)) :

    Errors and Failures:
    1) Test Failure : Xamarin.iOS.Tasks.TargetTests.RebuildExecutable_NoModifications
         #1: ../MySingleView/bin/iPhoneSimulator/Debug/MySingleView.app/MySingleView
      Expected: 2018-03-09 19:48:09.000
      But was:  2018-03-09 19:48:11.000

the timestamps clearly have a one-second resolution.

I logged onto the bot, to confirm its file system:

```shell
$ diskutil info /
   [...]
   File System Personality:  APFS
   Type (Bundle):            apfs
   Name (User Visible):      APFS
```

I tested that mono returns sub-second times:

```shell
$ csharp -e 'System.IO.File.GetLastWriteTimeUtc ("/Users/builder/MonkeyWrench.log").ToString ("HH:mm:ss.fffffff")'
"11:22:45.2788357"
```

so I have no idea why this is random on High Sierra as well.

rolfbjarne:
So I figured out why our High Sierra bots don't fail: mono didn't support sub-second timestamp resolution on non-Windows platforms until mono 5.12 (https://github.com/mono/mono/commit/e0e185d70abd73e82bbde7520edcb3bb0c68f1c8).

I tested for exactly this on the bot, but it turns out that the bot had mono 5.12 installed from another test run when I tested it 😠

PR #3745, opened 2018-03-14
[mtouch] Don't build libmonotouch-fixes.dylib anymore, it's not used.

rolfbjarne:
We haven't used libmonotouch-fixes.dylib for over a year (and it stopped
working before that), so this seems safe to remove.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=48830.


spouliot:
Someone still depend on it ;-)
```
14:49:58 monotouch-main.m:28:10: fatal error: '../tools/mtouch/monotouch-fixes.c' file not found
14:49:58 #include "../tools/mtouch/monotouch-fixes.c"
14:49:58          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
14:49:58 1 error generated.
14:49:58 
make[2]: *** [.libs/ios/monotouch-main.x86_64.o] Error 1
14:49:58 make[2]: *** Waiting for unfinished jobs....
```

rolfbjarne:
Turns out I was somewhat wrong, libmonotouch-fixes.dylib hasn't been working/used for over a year, but for x86_64 the same code was instead put inside libxamarin, and has been used all this time.

Which means that removing the code might very well break something for x86_64.

I'll just remove the libmonotoch-fixes.dylib library then, instead of removing the corresponding code file as well, since the code is still used.

PR #3747, opened 2018-03-14
[msbuild] fix creation ouf output dependency

lewurm:
attempt to fix https://github.com/mono/mono/issues/7518

`ilrepack.txt` should only be created if `ilrepack.exe` was successful.

To be honest, I've no idea what I'm doing, and it might not work for the problem we see on the bots. I had a similar issue _locally_ and this change fixed it for me, but I can't explain it really. My msbuild knowledge is pretty... bad 🙂 

That said, if I undestand the `ILRepack` task correctly, it does something like that translated to `make`:

```Makefile
ilrepack.txt: 1.dll 2.dll 3.dll 4.dll
  ilrepack /out:1.dll $^
  touch $@
```

That is funky and I think we shouldn't do that. Instead we should have a dedicated `.dll` for the merged result. Alas, I'm not able to translate that to `msbuild` words 😅

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Tasks.csproj
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Tasks.csproj
@@ -145,9 +145,9 @@
       <Output TaskParameter="ConsoleOutput" PropertyName="ILRepackOutput" />
       <Output TaskParameter="ExitCode" PropertyName="ExitCode" />
     </Exec>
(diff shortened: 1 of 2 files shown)


lewurm:
Looks like it fixed the issue 🎉  There's another failure, which is already tracked [here](https://github.com/xamarin/maccore/issues/592) and unrelated. The linker failures got fixed in `mono-2018-02` a couple hours ago -> Merging.

PR #3754, opened 2018-03-15
[jenkins] Clean keystore on bots before running tests

spouliot:
Because if invalid data gets into the store then some unit tests
will always fail on that particular bot.

Fix https://github.com/xamarin/maccore/issues/640


spouliot:
Need to be run on xam-mac-mini-1 to confirm the issue is solved

spouliot:
we got lucky `xam-mac-mini-1` picked it up on first try :D

PR #3763, opened 2018-03-16
[generator/registrar] Add support for blocks in static protocol members. Fixes #41226.

rolfbjarne:
Add support for blocks in static protocol members by adding another field to
the [ProtocolMember] attribute that specifies the block proxy type.

Fixes https://bugzilla.xamarin.com/show_bug.cgi?id=41226.


rolfbjarne:
@dalexsoto no, it won't (I have no idea what will happen in that case).

PR #3768, opened 2018-03-16
[mtouch-mmp-errors] Fix missing " for 0132

VincentDondain:


spouliot:
Nice catch :) but let's save wrench 5 hours since it will be overwritten soon, i.e.  I included it inside https://github.com/xamarin/xamarin-macios/pull/3771 (15.7) which will be merged back to master later

Issue #3787, opened 2018-03-20
Update the apidiff tool to export Markdown instead of HTML


bradumbaugh:
- Needed so that API diffs render correctly on docs.microsoft.com

spouliot:
I wrote the original HTML one, let's talk later this week about the details

bradumbaugh:
Sounds good - please include @asb3993 in the discussion.

rolfbjarne:
markdown instead of html would make us lose the diff colors, as well as the ability to hide non-breaking changes

spouliot:
@rolfbjarne sadly yes, but that's a requirement for docs.microsoft.com
However the bots can continue to produce the HTML version, which is useful for our own reviews

spouliot:
@asb3993 here's a first version of the output. https://gist.github.com/spouliot/862ba31ccb865f689f083feb6a4bb853

A few more stuff needs to be changed but it's close to what we can achieve with markdown.

Can you also confirm the header (recopied here) are correct ? it's based on https://github.com/MicrosoftDocs/xamarin-docs/blob/live/contributing-guidelines/template.md#metadata-and-markdown-template

```
---
title: Xamarin.iOS API diff
description: List of API differences for Xamarin.iOS
author: spouliot
ms.author: sepoulio
ms.date: 3/21/2018
ms.topic: article
ms.assetid: 3755637e-75a4-4c0a-be1f-3ddb09d167f2
ms.prod: xamarin
---
```

spouliot:
There's a newer reversion (same URL, might need to refresh) that handle modified API like a `diff`. Still a work in progress

spouliot:
@asb3993 Updated again, looking good now for my eyes.

@rolfbjarne I used `diff` for changes. Any other suggestions ?

Latest headers are:
```
---
title: "Xamarin.iOS SDK API diff: 11.8.0 vs 11.10.0"
description: List of API changes between Xamarin.iOS versions 11.8.0 and 11.10.0.
author: spouliot
ms.author: sepoulio
ms.date: 3/22/2018
ms.topic: article
ms.assetid: 2da20940-eef0-4530-8c2c-102595baae96
ms.prod: xamarin
---
```

rolfbjarne:
@spouliot I don't understand, what are you diffing?

spouliot:
@rolfbjarne see the modified API inside https://gist.github.com/spouliot/862ba31ccb865f689f083feb6a4bb853

[...]

PR #3814, opened 2018-03-25
[tools][api-diff] Add a tool to merge several .md files into a single one

spouliot:
This depends on an updated mono-api-html tool (not yet available). This
PR will test the creation (empty headers) and storage of the files on bots.

Reference https://github.com/xamarin/xamarin-macios/issues/3787


tools/apidiff/merger.cs
```diff
@@ -0,0 +1,90 @@
...
+			foreach (var line in File.ReadAllLines (file)) {
+				if (foundTitle != null) {
+					content.WriteLine (line);
+					if (line == "#### Type Changed: ObjCRuntime.Constants") {
+						lookForVersion = true;
+					} else if (line.StartsWith ("#### ")) {
+						lookForVersion = false;
+					}
+					if (lookForVersion) {
+						if (line.StartsWith ("-public const string Version = "))
```
rolfbjarne: Nitpick: `StringComparison.Ordinal` (below as well).

tools/apidiff/Makefile
```diff
@@ -15,7 +15,7 @@ APIDIFF_DIR=.
 MONO_API_HTML_DIR = $(MONO_PATH)/mcs/tools/mono-api-html
 MONO_API_INFO_DIR = $(MONO_PATH)/mcs/tools/corcompare
 MONO_API_INFO = $(MONO_API_INFO_DIR)/bin/Debug/mono-api-info.exe
-MONO_API_HTML = $(MONO_API_HTML_DIR)/bin/Debug/mono-api-html.exe
+MONO_API_HTML = $(TOP)/external/macios-binaries/mono-api-html.exe
```
rolfbjarne: I assume this will use the source code from mono once we use a branch that has the corresponding code?

tools/apidiff/Makefile
```diff
@@ -245,6 +264,18 @@ endif
 	@echo "@MonkeyWrench: AddFile: $(CURDIR)/api-diff.html"
 	@# remove empty files so they're not uploaded
 	@-find $(CURDIR)/diff -size 0 | xargs rm
+ifdef INCLUDE_IOS
```
rolfbjarne: I think this can be improved to make it more parallelizable, by declaring the *-markdown targets like this:

```makefile
ifdef INCLUDE_IOS
ios-markdown: merger.exe $(foreach file,$(IOS_ASSEMBLIES),$(APIDIFF_DIR)/diff/xi/$(file).md)
    $(Q) $(SYSTEM_MONO) --debug merger.exe Xamarin.iOS $(CURDIR)/diff/xi/Xamarin.iOS/ ios
else
ios-markdown: ; @true
endif

ifdef INCLUDE_TVOS
    ...
endif

...
```

and then here do:

```makefile
    $(MAKE) -j8 ios-markdown tvos-markdown watchos-markdown macos-markdown
```

I'm not sure how much of a difference it will make though.

tools/apidiff/Makefile
spouliot: yes. I suspect there will be some adjustments in the short term and then it will become quite _static_ again. Backporting it (well all adjustments) to several release branches is time consuming.

PR #3817, opened 2018-03-26
[xharness] Use msbuild for some tests

VincentDondain:
- Removed `new string [] { "Import", "Project" },` which is causing issues when using msbuild (hopefully it's not breaking things since we're mostly using msbuild now).

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -324,6 +324,7 @@ namespace xharness
 							Platform = pair.Item2,
 							Ignored = pair.Item3,
 							TestName = project.Name,
+							UseMSBuild = true,
 						};
 						derived.CloneTestProject (pair.Item1);
 						var simTasks = CreateRunSimulatorTaskAsync (derived);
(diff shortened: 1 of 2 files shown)


VincentDondain:
This PR is a tentative at using msbuild for our test. It needs to be tested on the bots since their setup is different than local setup (where those changes work).

VincentDondain:
I hoped this PR might work on the bots as it works locally but it seems to be hitting issues due to lack of msbuild redirect.

In the tests build logs we can see a mix of `/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.iOS/*` and `/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.iOS/*` which breaks.

Might be due in part or in full to `MSBuildExtensionsPath=null` which is defined when using msbuild. However using something else doesn't seem to be supported by msbuild yet (tried locally).

In short, we're hitting things that @radical is trying to address in https://bugzilla.xamarin.com/show_bug.cgi?id=53164.

I'm closing this PR as he'll likely revisit switching the tests to msbuild when his fix lands.

PR #3853, opened 2018-04-03
[ClassKit] Add Xcode 9.4 Beta 1 Bindings

dalexsoto:


src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+		double Duration { get; }
+
+		[Export ("primaryActivityItem", ArgumentSemantic.Assign)]
+		CLSActivityItem PrimaryActivityItem { get; set; }
+
+		[Export ("addProgressRangeFromStart:toEnd:")]
+		void AddProgressRange (double start, double end);
+
+		[Export ("addAdditionalActivityItem:")]
+		void AddAdditional (CLSActivityItem activityItem);
```
VincentDondain: Maybe just `Add`.

I think `AddProgressRange` above is fine because the 2 double parameters are not enough to understand what we're adding but here it's obvious you add an activityItem.

Also "Add" and "Additional" are too similar, it's weird to have them together.

src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+
+	[NoWatch, NoTV, NoMac, iOS (11,3)]
+	[BaseType (typeof (CLSObject))]
+	[DisableDefaultCtor]
+	interface CLSContext {
+
+		[Export ("identifier")]
+		string Identifier { get; }
+
+		[NoWatch, NoTV, NoMac, iOS (11,4)]
```
VincentDondain: NoWatch, NoTV, NoMac are redundant here.

src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+		// From CLSContext (Hierarchy) Category
+
+		[NullAllowed, Export ("parent")]
+		CLSContext Parent { get; }
+
+		[Export ("removeFromParent")]
+		void RemoveFromParent ();
+
+		[Export ("addChildContext:")]
+		void AddChild (CLSContext childContext);
```
VincentDondain: Just `Add`?

src/classkit.cs
dalexsoto: Nice catch

src/classkit.cs
dalexsoto: I originally had `AddAdditionalActivityItem`, I am inclined to go back to it. My main reason to not have just `Add` is because the container type isn't named as a collection, maybe `AddActivity`?

src/classkit.cs
dalexsoto: Same reasoning, `CLSContext` does not refer to be a collection, I think `AddChild` es explicit on its intent

src/classkit.cs
VincentDondain: I see that's a good reason. Then I vote for `AddActivity`. Could be `AddAdditionalActivity` but really... we're adding an activity so it's obviously in "addition" (additional) to the others...

Note: in Swift they went for `addAdditionalActivityItem`...
https://developer.apple.com/documentation/classkit/clsactivity/2953065-addadditionalactivityitem

src/classkit.cs
VincentDondain: Fair enough. Let's leave it as is them.

src/classkit.cs
dalexsoto: > Then I vote for `AddActivity`. Could be `AddAdditionalActivity` but really... we're adding an activity so it's obviously in "addition" (additional) to the others...

We have `CLSActivity` and `CLSActivityItem` objects, so I think `AddActivityItem` makes more sense and it's clearer, agree that `AddAdditional` sounds "recursive" so +1 on omitting the `Additional` part

src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+
+		[Async]
+		[Export ("descendantMatchingIdentifierPath:completion:")]
+		void FindDescendantMatching (string [] identifierPath, Action<CLSContext, NSError> completion);
```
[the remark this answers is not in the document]
dalexsoto: Looks like it isn't according to docs

> The identifier path of the context to find, relative to the current context.

Looks related to the actual ctx identifier which is a string

src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+	interface ICLSDataStoreDelegate { }
+
+	[NoWatch, NoTV, NoMac, iOS (11,3)]
+	[Protocol, Model]
+	[BaseType (typeof (NSObject))]
+	interface CLSDataStoreDelegate {
+
+		[Abstract]
+		[Export ("createContextForIdentifier:parentContext:parentIdentifierPath:")]
+		CLSContext CreateContext (string identifier, CLSContext parentContext, string [] parentIdentifierPath);
```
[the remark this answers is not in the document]
dalexsoto: same, string looks ok here, identifiers can be something like `chapter-1`, `section-1` etc, meaning that the parent identifier path is a collection of the former depending on the hierarchy

src/foundation.cs
```diff
@@ -5101,6 +5101,16 @@ partial interface NSUserActivity {
...
+
+		// From NSUserActivity (CLSDeepLinks)
+
+		[NoWatch, NoTV, NoMac, iOS (11,3)]
+		[Export ("isClassKitDeepLink")]
+		bool IsClassKitDeepLink { get; }
+
+		[NoWatch, NoTV, NoMac, iOS (11,3)]
+		[NullAllowed, Export ("contextIdentifierPath")]
+		string[] ContextIdentifierPath { get; }
```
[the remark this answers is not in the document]
dalexsoto: This is also fine, `ContextIdentifierPath` is ok as is, the enum is used for `FindContextsMatching` but the one that takes a predicate

src/classkit.cs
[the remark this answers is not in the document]
spouliot: interesting, then where are `CLSPredicateKeyPaths` being used ?

src/classkit.cs
```diff
@@ -0,0 +1,329 @@
...
+
+		// From CLSDataStore (Contexts) Category
+
+		[Async]
+		[Export ("contextsMatchingPredicate:completion:")]
+		void FindContextsMatching (NSPredicate predicate, CLSDataStoreMatchingResults completion);
+
+		[Async]
+		[Export ("contextsMatchingIdentifierPath:completion:")]
+		void FindContextsMatching (string [] identifierPath, CLSDataStoreMatchingResults completion);
```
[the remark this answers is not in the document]
dalexsoto: The API is fine, the enum is for the overload that takes a predicate, you would use it as follows

```csharp
var store = CLSDataStore.Shared;
var predicate = NSPredicate.FromFormat ("%K = %@", CLSPredicateKeyPaths.Parent.GetConstant (), store.MainAppContext);
var ctxs = await store.FindContextsMatchingAsync (predicate);
foreach (var ctx in ctxs) {
	Console.WriteLine (ctx.Title);
}
```

`%@` is a substitution for an object value.
`%K` is a var arg substitution for a key path, a `CLSPredicateKeyPaths` member in this case.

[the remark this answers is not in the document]
spouliot: hmm... that does not make a great use of an enum if the only use case requires calling `GetConstant` :|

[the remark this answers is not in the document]
dalexsoto: Do you want me to turn this into a static class instead? I think it would make more sense

dalexsoto:
@spouliot / @VincentDondain I think I have addressed your feedback, thanks a lot!!

spouliot:
@dalexsoto all builds have failed with unit tests failures

src/classkit.cs
```diff
@@ -276,8 +273,12 @@ interface CLSDataStore {
 		[NullAllowed, Export ("runningActivity")]
 		CLSActivity RunningActivity { get; }
 
+		[Wrap ("WeakDelegate"), Protocolize]
```
spouliot: `[Protocolize]` was only needed to maintain compatibility with classic

src/foundation.cs
```diff
@@ -5101,6 +5101,16 @@ partial interface NSUserActivity {
 		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3), NoWatch]
 		[NullAllowed, Export ("detectedBarcodeDescriptor", ArgumentSemantic.Copy)]
 		CIBarcodeDescriptor DetectedBarcodeDescriptor { get; }
+
+		// From NSUserActivity (CLSDeepLinks)
+
+		[NoWatch, NoTV, NoMac, iOS (11,3)]
```
spouliot: should be 11,4

src/foundation.cs
```diff
@@ -5101,6 +5101,16 @@ partial interface NSUserActivity {
...
+		[Export ("isClassKitDeepLink")]
+		bool IsClassKitDeepLink { get; }
+
+		[NoWatch, NoTV, NoMac, iOS (11,3)]
```
spouliot: should be 11,4

src/classkit.cs
```diff
@@ -0,0 +1,336 @@
...
+		[Export ("mainAppContext")]
+		CLSContext MainAppContext { get; }
+
+		[NullAllowed, Export ("activeContext")]
+		CLSContext ActiveContext { get; }
+
+		[NullAllowed, Export ("runningActivity")]
+		CLSActivity RunningActivity { get; }
+
+		[Wrap ("WeakDelegate"), Protocolize]
```
spouliot: `[Protocolize]` was to keep compatibility with classic.

dalexsoto: @spouliot Yeah I know but something weird is happening

If I do 

```csharp
[Wrap ("WeakDelegate")]
[NullAllowed]
CLSDataStoreDelegate Delegate { get; set; }

```

I get a warning `BI1110: bgen: The property ClassKit.CLSDataStore.Delegate exposes a model (ClassKit.CLSDataStoreDelegate). Please expose the corresponding protocol type instead (ClassKit.ICLSDataStoreDelegate).` which is completely expected and correct, if I change the code to 

```csharp
[Wrap ("WeakDelegate")]
[NullAllowed]
ICLSDataStoreDelegate Delegate { get; set; }

```

I get a build error

```
xamarin-macios/src/build/ios/native/ClassKit/CLSDataStore.g.cs(215,20): error CS0266: Cannot implicitly convert type `ClassKit.ICLSDataStoreDelegate' to `Foundation.NSObject'. An explicit conversion exists (are you missing a cast?)
xamarin-macios/src/build/ios/native/ClassKit/CLSDataStore.g.cs(215,20): error CS0266: Cannot implicitly convert type `ClassKit.ICLSDataStoreDelegate' to `Foundation.NSObject'. An explicit conversion exists (are you missing a cast?)
```

Generated code looks like this:

```csharp
[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		WeakDelegate = value;
	}
}
```

The only way to get the right code is using `[Protocolize]` 😞 

```csharp
[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		var rvalue = value as NSObject;
		if (value != null && rvalue == null)
			throw new ArgumentException ("The object passed of type " + value.GetType () + " does not derive from NSObject");
		WeakDelegate = rvalue;
	}
}
```

Or what am I missing here?

dalexsoto:
build

src/classkit.cs
rolfbjarne: I disagree :smile:

Reading through Apple's documentation (https://developer.apple.com/documentation/classkit/recording_additional_metrics_about_a_completed_task), there seems to be a "primary" activity, and then you can add additional activities as well, which makes the `Additional` part slightly less redundant (it also clearly states that it will never be a primary activity).

So I'd go for `AddAdditionalActivity` (there's also an `AdditionalActivityItems` property; and whatever we do those two should match).

src/classkit.cs
rolfbjarne: This might need a generator change 😒

We didn't used to add the `WeakDelegate` property for Delegate properties that were typed as protocols. We have a few, but they're all readonly from what I can see, and don't run into this problem.

I think it's fine to comment out the WeakDelegate property, and file an issue for the generator.

src/classkit.cs
dalexsoto: Good point @rolfbjarne, I think I will go for the [swift name](https://developer.apple.com/documentation/classkit/clsactivity/2953065-addadditionalactivityitem) here `AddAdditionalActivityItem` but yeah makes sense now reading it that way, thanks for having a look.

src/classkit.cs
dalexsoto: @spouliot filed as https://github.com/xamarin/xamarin-macios/issues/3869

dalexsoto:
unrelated test failure (https://github.com/xamarin/maccore/issues/592)

spouliot:
@dalexsoto please add this to the release notes draft
https://github.com/xamarin/xamarin-macios/wiki/xamarin.ios_11.9.2
thanks!

dalexsoto:
@spouliot Already there :D

PR #3850, opened 2018-04-03
Merge xcode9.3 into master

VincentDondain:
- Bump maccore and maciostool for Xcode 9.3 merge.

PR #3855, opened 2018-04-03
API diff update (from stables) and xtro cleanup (unneeded files)

spouliot:


spouliot:
@rolfbjarne the compare API/gen diff started to fail with https://github.com/xamarin/xamarin-macios/pull/3850
It should be made an error (not a green check) since assuming people will scroll down to the status is not very safe.

PR #3799, opened 2018-03-22
[jenkins] Create comment file for PR builds.

rolfbjarne:

PR #3858, opened 2018-04-04
[xcode9.4] [jenkins] Create comment file for PR builds. (#3799)

rolfbjarne:


rolfbjarne:
@spouliot no, because the xcode9.4 branch doesn't have the script to create api diff / generator diff.

PR #3871, opened 2018-04-04
Fix GetMatchingFontDescriptors overload with sort callback

filipnavara:
It incorrectly returned array of arrays instead of array of CTFontDescriptors.


filipnavara:
Unfortunately I won't have time to produce the test case in the next few days. I only stumbled upon it by accident when trying to optimise some code that used the other overload.

spouliot:
@VincentDondain as code owner (lucky you) please add a unit test to the PR and merge it. thanks!

VincentDondain:
build

VincentDondain:
@spouliot (and others) can you approve / re-approve?

PR #3882, opened 2018-04-05
[BusinessChat] Make mtouch and mmp aware of BusinessChat framework

dalexsoto:


spouliot:
please back port to `d15-7`, thanks!

Issue #3869, opened 2018-04-04
[generator] Using WrapAttribute and interface protocols leads to uncompilable code


dalexsoto:
### Steps to Reproduce

1. use the following interface definition in a binding project

```csharp
interface ICLSDataStoreDelegate { }

[Protocol, Model]
[BaseType (typeof (NSObject))]
interface CLSDataStoreDelegate {

	[Abstract]
	[Export ("createContextForIdentifier:parentContext:parentIdentifierPath:")]
	NSObject CreateContext (string identifier, NSObject parentContext, string [] parentIdentifierPath);
}

[BaseType (typeof (NSObject))]
interface CLSDataStore {

	[Wrap ("WeakDelegate")]
	[NullAllowed]
	ICLSDataStoreDelegate Delegate { get; set; }

	[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
	NSObject WeakDelegate { get; set; }
}
```
2. Build and observe error:
> CS0266: Cannot implicitly convert type `ClassKit.ICLSDataStoreDelegate' to `Foundation.NSObject'. An explicit conversion exists (are you missing a cast?)

### Expected Behavior

Generated code builds

```csharp
[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		var rvalue = value as NSObject;
		if (value != null && rvalue == null)
			throw new ArgumentException ("The object passed of type " + value.GetType () + " does not derive from NSObject");
		WeakDelegate = rvalue;
	}
}
```
### Actual Behavior
Generated code fails to build:

```csharp
[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		WeakDelegate = value;
	}
}
```

### Environment

```
=== Visual Studio Enterprise 2017 for Mac ===

Version 7.4.1 (build 48)
Installation UUID: 4e014908-906a-4d2b-9aed-1d598a6d71cf
Runtime:
	Mono 5.10.0.105 (2017-12/4b1745401d5) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)

	Package version: 510000105

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.0.5
	2.0.3
[...]

PR #3885, opened 2018-04-05
[generator] Fixes WrapAttribute usage with interface protocols

dalexsoto:
Fixes xamarin/xamarin-macios#3869

We need to support the scenario where `WrapAttribute` is used
to wrap an interface protocol type declaration like:

```csharp

[Wrap ("WeakDelegate")]
[NullAllowed]
ICLSDataStoreDelegate Delegate { get; set; }

```

We generated invalid code for the setter that needed an explicit cast
and ended up with a CS0266 compiler error:

Bad generated code:

```csharp

[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		WeakDelegate = value;
	}
}

```

Good generated code:

```csharp

[CompilerGenerated]
public ICLSDataStoreDelegate Delegate {
	get {
		return WeakDelegate as ICLSDataStoreDelegate;
	}
	set {
		var rvalue = value as NSObject;
		if (value != null && rvalue == null)
			throw new ArgumentException ("The object passed of type " + value.GetType () + " does not derive from NSObject");
		WeakDelegate = rvalue;
	}
}

```

Build diff: https://gist.github.com/dalexsoto/0899524f285a9fc673f90f4b45431994


rolfbjarne:
Thinking a bit more about this, the generated code is not ideal, we can do better than that:

```csharp
public ICLSDataStoreDelegate Delegate {
	get {
		ICLSDataStoreDelegate ret;
		if (IsDirectBinding) {
			ret = Runtime.GetINativeObject<ICLSDataStoreDelegate> (global::ObjCRuntime.Messaging.IntPtr_objc_msgSend (this.Handle, Selector.GetHandle ("delegate")), false);
		} else {
			ret = Runtime.GetINativeObject<ICLSDataStoreDelegate> (global::ObjCRuntime.Messaging.IntPtr_objc_msgSendSuper (this.SuperHandle, Selector.GetHandle ("delegate")), false);
		}
		return ret;
	}
	set {
		if (IsDirectBinding) {
			global::ObjCRuntime.Messaging.void_objc_msgSend_IntPtr (this.Handle, Selector.GetHandle ("setDelegate:"), value.GetHandle ());
		} else {
			global::ObjCRuntime.Messaging.void_objc_msgSendSuper_IntPtr (this.SuperHandle, Selector.GetHandle ("setDelegate:"), value.GetHandle ());
		}
	}
}
```

this removes the restriction that the Delegate must be an NSObject (for instance our *Wrapper types are not NSObject subclasses).

I'm not sure how useful this would be though, since Delegate instances are usually provided by the app developer (and thus something the app developer controls, as opposed to something Apple gives us, and which can be anything).

@spouliot what do you think?

dalexsoto:
While I agree with the idea, I think this is for a different PR, I am not sure we might want to backport such a generator change into d15-7 at this stage. My vote is on merging the current change as is, backport into d15-7 and file a new issue with said enhancement.

That said :+1: on Rolf’s proposal

spouliot:
@rolfbjarne I like it but, like most of bindings enhancements, it's tricky to measure the impact (but I think we're getting close, e.g. by monitoring app size changes) to be able to judge them more effectively. For the moment please file a RFC so we don't lose it and get additional feedback.

@dalexsoto like we discussed it's too late for 15.7 since it's not really needed (not a regression). If it helps us then we _might_ cherry-pick it (since it's simple) into a branch based on 15.7 (e.g. `xcode10`).

EugeneUr:
As i was upgrading my Xamarin bindings to MAUI i ran into 

`JTCalendarManager.g.cs(176,20): Error CS0266 : Cannot implicitly convert type 'Foundation.NSObject' to 'JTCalendar.IJTCalendarDelegate'. An explicit conversion exists (are you missing a cast?)`

and the way to fix was to replace 

`WeakDelegate = rvalue;`
with
`WeakDelegate = value;`

Which seems more correct.

PR #3891, opened 2018-04-06
[mtouch-errors] Update MT1108 DeveloperDiskImage.dmg full path

VincentDondain:
Path changed and it has /Platforms now. Also added /Applications at the beginning.


VincentDondain:
Known issue: https://github.com/xamarin/maccore/issues/648

PR #3900, opened 2018-04-09
[docs] Document recent changes to ModelAttribute in the generator.

rolfbjarne:


docs/website/binding_types_reference_guide.md
```diff
@@ -720,6 +720,28 @@ class.
 
  <a name="NoDefaultValueAttribute"></a>
```
spouliot: seems this is inserted in the wrong place

dalexsoto: Whoops, yeah nice catch!

rolfbjarne: @spouliot fixed, thanks!

PR #3905, opened 2018-04-09
[msbuild] Catch ArgumentException if IBToolTask fails to load plist o…

jstedfast:
…utput

Fixes the stack trace spew in https://devdiv.visualstudio.com/DevDiv/_workitems/edit/595179


jstedfast:
@spouliot I just verified that ArgumentException and FormatException will catch all the exceptions thrown by the plist parser.

I suppose that it's possible that in the future, others may emerge though. I can modify the PR so that it simply catches all exceptions.

PR #3918, opened 2018-04-10
[xharness] Fix creation of html reports for NUnit tests.

rolfbjarne:
Make sure to override all required TextWriter methods so that the LogFile gets
all the output any writers (such as the XmlWriter that is used to create the
html report for NUnit tests by applying an XSLT transform to the xml results)
writes.


rolfbjarne:
Test failure is a known issue (https://github.com/xamarin/maccore/issues/643).

PR #3919, opened 2018-04-11
[tests] Fix MT4134 test by adding new framework to list of unsupported frameworks when using older Xcode.

rolfbjarne:
Fixes this test failure:

    1) Failed : Xamarin.Registrar.MT4134
      errors
      Expected: 9
      But was:  10

This was not detected by the PR bot because this test requires having Xcode
8.3 installed (otherwise the test is ignored), and the PR bot didn't have that
version of Xcode.

### DIFF
--- a/tests/mtouch/RegistrarTest.cs
+++ b/tests/mtouch/RegistrarTest.cs
@@ -279,9 +279,20 @@ class MyObjectErr : NSObject, IFoo1, IFoo2
 				mtouch.Linker = MTouchLinker.LinkSdk;
 				mtouch.Registrar = MTouchRegistrar.Static;
 				mtouch.AssertExecuteFailure ();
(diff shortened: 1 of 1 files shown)


dalexsoto:
@rolfbjarne so this test needs to be updated every year during the summer, right?

rolfbjarne:
@spouliot that won't work (since mtouch uses `tools/common/Frameworks.cs` to determine which frameworks to link with, mtouch won't try to link with frameworks that are not in that list, and thus won't either error out with the MT4134 error - iow you'd use the same data source for both the the test and the behavior you're testing, and if some data is missing, the test won't catch it).

But we should still figure out a way to test this, it's a frequent problem.

@dalexsoto that would be preferable, yes.

Issue #3930, opened 2018-04-12
[XI]WatchOS Intent sample fail to build with MT4116 / NullReferenceException


GouriKumari:
### Steps to Reproduce

1. Build watch intent sample from IDE or command line

### Expected Behavior
App should build successfully
### Actual Behavior
App fails to build with an error could not register the assembly xamarin.watchOS. This issue occurs only with watch intent sample. Watchkitcatalog is building successfully
### Environment
```
https://gist.github.com/GouriKumari/49a33b1658478361ce0b177808b7fdbb
```

### Build Logs

Build Log: https://gist.github.com/GouriKumari/cca1bfb3a64d77a64e6e164b414e63b5

### Regression:
This is working with d15-6, hence it is a regression in watch intent sample. This issue also occurs with latest XI from d15-7 11.10.1.170

> VS bug [#599867](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/599867)

spouliot:
```
MTOUCH : error MT4116: Could not register the assembly 'Xamarin.WatchOS': System.NullReferenceException: Object reference not set to an instance of an object
```
https://gist.github.com/GouriKumari/cca1bfb3a64d77a64e6e164b414e63b5#file-gistfile1-txt-L8601

@rolfbjarne can you have a look and see if there's a workaround for it ? thanks!

rolfbjarne:
I can reproduce, and I've PR'ed a fix.

This affects watchOS extensions with extensions (i.e. not very common).

It's also possible to work around the bug by passing `--optimize:-all` to mtouch.

rolfbjarne:
Closing according to previous comment.

PR #3937, opened 2018-04-13
[d15-7][mtouch] watchOS extensions are top-level containers. Fixes issue #3930. (#3934)

rolfbjarne:
watchOS extensions are top-level containers in our build (because we ignore
watchOS apps entirely in mtouch), so treat them as such when computing the
protocol member map.

Fixes issue https://github.com/xamarin/xamarin-macios/issues/3930.


spouliot:
Chatted with @AdrianMurphy. Merging for QA verification before P5

PR #3957, opened 2018-04-19
[build] Remove MT.D source build and replace it with a binary

dalexsoto:
This commit removes MonoTouch.Dialog from our source build and
replaces it with a binary from https://github.com/xamarin/macios-binaries/tree/c913506df2d106f9c073f211b7c66b03eefc89f2/MonoTouch.Dialog-Unified

The MT.D hash used in this commit is fixed to migueldeicaza/MonoTouch.Dialog@92c6e14


dalexsoto:
Huh? interesting, I am pretty sure I ran the tvOS tests

dalexsoto:
@rolfbjarne do you have an idea why this is happening? I ran `make jenkins` locally and seems to do the right thing here https://gist.github.com/dalexsoto/c53926dc76c9337455b125a6ecbd4a89

rolfbjarne:
This looked puzzling, so I had a quick look, and just became more puzzled.

I guess I'll have to have a deeper look 👀

rolfbjarne:
This makes no sense 😒

I rebuilt & reran the tests manually on the bot, and it worked just fine.

I guess I'll have to add debug code to xharness to see what's going on.

dalexsoto:
we've been (un?)lucky enough to build 3 times under the same bot...

rolfbjarne:
Yay, I figured out it: it's the code to build the previous revision & calculate the api/generator diff that's interfering.

Now to find a fix...

spouliot:
added `do-not-merge` label since it's green, approved but _now_ has debugging code that should not be merged.

dalexsoto:
Removed the do-not-merge tag since there is no more debug spew and it is now truly green :)

PR #3963, opened 2018-04-20
[tools] Fix commit comparison to not change generated .csproj files.

rolfbjarne:
Fix commit comparison to not change generated .csproj files by adding support
for modifying the directory where the .csproj are generated, and when doing
commit comparison, do exactly that.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,6 +2,7 @@ TOP=..
 include $(TOP)/Make.config
 
 BUILD_DIR=build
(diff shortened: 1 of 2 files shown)


rolfbjarne:
The API/generator diff failed because it detected that unexpected files were modified.

In this particular case that's expected, because this PR fixes some issues, which means that it detects files that are modified when building the "before" code (which does not have the fix and thus touches files to shouldn't).

PR #3979, opened 2018-04-24
[runtime] Don't lock while calling selectors that can up calling managed code. Fixes #3943.

### LINKED ISSUE #3943: Deadlock when dealloc calls managed code and provokes a GC, and the GC has to process togglerefs
Test case: https://github.com/rolfbjarne/xamarin-macios/commit/e990ce0352492e1f8037375d9642eac8eab971d8

Conditions:

* ObjC object overrides dealloc, does something that calls managed code.
* A GC occurs as a result of the ObjC object's dealloc method calling managed code (but not on the same thread).
* The GC processes toggle ref objects.

bt all: https://gist.github.com/rolfbjarne/0c9a0f90cbb7163275288a4a4d6231d2.

Unfortunately the main thread's stack trace is truncated, but it's deallocating the ObjC object, as a result of this: https://github.com/xamarin/xamarin-macios/blob/0a9453c1f296261a2a081232e3519a7a4c5397ba/runtime/runtime.m#L1935

Then a background thread is trying to run the GC, which tries to lock the framework lock already locked before calling `[self release]` above, and deadlocks: https://github.com/xamarin/xamarin-macios/blob/0a9453c1f296261a2a081232e3519a7a4c5397ba/runtime/runtime.m#L859

Some historical references:

* [[newrefcount] Handle explicit dispose correctly.](https://github.com/xamarin/monotouch/commit/09beb274e8f1f094c6ea51d910a058a85b08017f)
* [[toggleref] Make the toggleref mutex recursive, and lock it when returning from a native->managed trampoline.](https://github.com/xamarin/monotouch/commit/abcfd1ff170ec72c6e0b39343b17a4655a1b480f)

rolfbjarne:
Instead of locking the framework peer lock while we call release on dying
object, we just lock and then immediately unlock again before calling release.

This enforces an execution order that still strong enough to not run into race
conditions, while at the same time not running into deadlocks.

Fixes https://github.com/xamarin/xamarin-macios/issues/3943.


runtime/runtime.m
```diff
@@ -1929,12 +1929,77 @@ -(void) xamarinSetGCHandle: (int) gc_handle;
 	} else {
 		/* If we're a wrapper type, we need to unregister here, since we won't enter the release trampoline */
 		xamarin_unregister_nsobject (self, managed_obj, &exception_gchandle);
-		// This lock is needed so that we can safely call retainCount in the toggleref callback.
+
+		//
+		// This lock is needed so that we can safely call retainCount in the
```
VincentDondain: Love this long comment so much <3

runtime/runtime.m
```diff
@@ -1929,12 +1929,77 @@ -(void) xamarinSetGCHandle: (int) gc_handle;
...
+		//      b2 won't do anything
+		//   2) b1-b2-a1-a2: retainCount before release is safe.
+		//   3) b1-a1-b2-a2: retainCount before release is safe.
+		//   4) b1-a1-a2-b2: unsafe; this tries to call retainCount after
+		//      release.
+		//
+		// Order 4 would look like this:
+		//
+		//   * Thread A runs a GC, and starts calling toggleref callbacks.
+		//   * Thread A fetches the handle (H) for object Z in a toggleref
```
VincentDondain: Minor: it's _slightly_ confusing that you're flipping thread A and B in this description and below. "Thread A fetches the handle (H) for object Z in a toggleref" is b1 above.

runtime/runtime.m
```diff
@@ -1929,12 +1929,77 @@ -(void) xamarinSetGCHandle: (int) gc_handle;
...
+		//    managed code), and we can deadlock:
+		//    1) Thread A calls release on a native object.
+		//    2) Thread A executes managed code, which blocks on something
+		//       that's supposed to happen on another thread B.
+		//    3) Thread B causes a garbage collection.
+		//    4) Thread B tries to lock the framework peer lock before running
+		//       the GC, and deadlocks because thread A already has the
+		//       framework peer lock.
+		//
+		//    This is https://github.com/xamarin /xamarin-macios/issues/3943
```
VincentDondain: Minor: Unwanted space in URL here.

tests/bindings-test/RuntimeTest.cs
```diff
@@ -0,0 +1,44 @@
...
+	{
+		[Test]
+		public void EvilDeallocatorTest ()
+		{
+			// Create a few toggle-ref objects
+			for (var i = 0; i < 10; i++) {
+				var ed = new EvilDeallocator ();
+				ed.MarkMeDirty ();
+			}
+			// Now create an object that will call a managed callback it its destructor
```
VincentDondain: Minor typo: **in** its destructor.

PR #3984, opened 2018-04-25
[foundation] Avoid unnecessary native calls for NSNull.Null. Fixes #3544

### LINKED ISSUE #3544: Unnecessary calls to NSNull.Null slows performance
### Steps to Reproduce

1. Use test case from here: https://github.com/xamarin/xamarin-macios/issues/3488#issuecomment-366817032
2. The code ends up converting a native `NSArray` with thousands of elements to a managed `NSMenuItem[]`.
3. Each array element ends up calling `NSArray.UnsafeGetItem<T>`, which calls `NSNull.Null.Handle`: https://github.com/xamarin/xamarin-macios/blob/2a964030a88efc35acfa987778da7950bb32d1f3/src/Foundation/NSArray.cs#L342

The call to `NSNull.Null` will call native code every time, and also spend time finding/creating the managed wrapper, just to get the handle.

I think it would be better to have a static `NSNull.NullHandle` field, which is used instead:

```csharp
static IntPtr null_handle;
internal IntPtr NullHandle {
    get {
        if (null_handle == IntPtr.Zero)
            null_handle = ...;
        return null_handle;
    }
}
```

spouliot:
Repetitive calls to `NSNull.Null.Handle`, e.g. from `NSArray.UnsafeGetItem<T>`,
means several (one by item), costly, calls into ObjC code - but it always
return the same (native) singleton.

This manually cache the managed `NSNull.Null` static value, once it's
initialized the first time, so future calls won't have the performance
penalty.

https://github.com/xamarin/xamarin-macios/issues/3544


rolfbjarne:
Build hung 😒  (https://github.com/xamarin/maccore/issues/762)

rolfbjarne:
build

Issue #3810, opened 2018-03-23
[msbuild] Add msbuild test for #3791: Skip reference assemblies passed to mtouch


VincentDondain:
See https://bugzilla.xamarin.com/show_bug.cgi?id=53164.

Our msbuild tests are in fact still using `xbuild`. Because the test case in https://github.com/xamarin/xamarin-macios/pull/3791 is using NugetPackages added via "PackageReference" it needs msbuild. `xbuild` cannot "restore" nor is using the right `ResolveAssemblyReferences` target (that feeds csc with the right assemblies).

Adding the test itself once our tests are using msbuild should be trivial. It's just a matter of building the project with msbuild which will resolve the packages correctly.

VincentDondain:
There should also be a mmptest to cover the Xamarin.Mac side of the house.

https://github.com/xamarin/xamarin-macios/pull/3991

VincentDondain:
xbuild fix now being tracked here: https://github.com/xamarin/xamarin-macios/issues/4110

rolfbjarne:
This is no longer relevant with .NET.

PR #3991, opened 2018-04-25
[msbuild/mac] Skip reference assemblies passed to mmp

VincentDondain:
- Similar to issue #3199: Could not AOT the assembly System.Runtime.CompilerServices.Unsafe.dll (MT3001)
  (https://github.com/xamarin/xamarin-macios/issues/3199)
- Test case: https://www.dropbox.com/s/49jxl8iftmlymes/Issue3199Test3.zip?dl=0

Problem
=======

Given a Nuget Package added via the "package reference" mechanism and the said package having netstandard `lib` **and** `ref` folders;
`mmp`, if given `--aot:all` was getting reference assemblies and couldn't AOT them as those assemblies are only facades.

Solution
========

Skipping the assemblies that have `/ref/` in their path seems like the simplest and yet most functional solution to the problem.

As it turn out, there is some logic already in place that copies the `lib` assemblies to the destination folder. It seems equivalent to marking them as "Local Copy".
What this does is that it makes those assemblies available to msbuild via `@(ReferenceCopyLocalPaths)`. This gives us the opportunity to safely remove the `ref` assemblies from `@(ReferencePath)`.

*Note: `mmp` is getting the assemblies to reference via a combination of `@(ReferencePath)` and `@(ReferenceCopyLocalPaths)`.*

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
@@ -479,6 +479,11 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		Inputs="$(TargetDir)$(TargetFileName)"
 		Outputs="$(_AppBundlePath)Contents\MacOS\$(_AppBundleName)">
 		<ItemGroup>
+			<!-- Skip the reference assemblies. 'mmp' will fail to AOT them if passed '--aot:all'. Similar to: https://github.com/xamarin/xamarin-macios/issues/3199 -->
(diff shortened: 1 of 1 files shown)


VincentDondain:
@spouliot mmh it's too late to get anything into d15-7 anymore isn't it?

spouliot:
@VincentDondain yes, much too late for d15.7. If it turns out to be a problem (has not so far for XM) we can re-evaluate for a service release (there's still the old workaround that can be used).

VincentDondain:
@chamons I agree with you on the need for a test but unless mmp tests are using msbuild it's the same issue reported here: https://github.com/xamarin/xamarin-macios/issues/3810

I added Mac tags, feel free to subscribe to it.

PR #3998, opened 2018-04-26
Remove more unused/duplicated makefile targets/logic.

rolfbjarne:
It's also defined here: https://github.com/xamarin/xamarin-macios/blob/bf159969f9b5c72abe3d6959cf8c4840495b3f2b/Makefile#L106-L107

Fixes this problem when creating API/Generator diff:

    make: *** No rule to make target `/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tools/comparison/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/Version', needed by `all-ios'.  Stop.


rolfbjarne:
It's expected that the API/Generator diff fails, because it fails building the previous hash, which does obviously not have the fix.

PR #4006, opened 2018-04-27
[tests] Adjust SetupBlockPerfTest to expect at least a 6x speedup instead of 8x. Fixes maccore #649.

rolfbjarne:
Adjust SetupBlockPerfTest to expect at least a 6x speedup instead of 8x, since
the test is randomly hitting the assert. None of the reported failures so far
has had a speedup of less than 6x, so hopefully it will be better now.

Fixes https://github.com/xamarin/maccore/issues/649.


VincentDondain:
I somehow hit that 6x speedup in my xharness device fixes back port branch: https://github.com/xamarin/xamarin-macios/commit/ee28b9e53d9d3ef3a5633609c8a3688623747345

It looks like I have your change and it still failed FYI.

VincentDondain:
Expected: greater than 6
  But was:  4.9317220348827826d

So I'm adjusting the test on that branch to go to 4x

PR #4011, opened 2018-04-27
[mtouch/mmp] Fix deprecation notice for --force-thread-check/--disable-thread-check.

rolfbjarne:
The --linker-optimize option is an early version of the current --optimize
option, and apparently I forgot to update everything when the option changed
name.


rolfbjarne:
Known build failure (https://bugzilla.xamarin.com/show_bug.cgi?id=57959)

rolfbjarne:
build

PR #4015, opened 2018-04-30
[runtime] bump system 2018-02 mono

lewurm:
so that this commit is included:
https://github.com/mono/mono/commit/4a00c501b21cdb2cb81d4ddefd1115e32d56b4f5

It makes roslyn hangs (and crashes?) less likely, so let's use it. Might
helps with https://bugzilla.xamarin.com/show_bug.cgi?id=57959#c20

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -60,9 +60,9 @@ XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_9.3.xip
 XCODE_DEVELOPER_ROOT=/Applications/Xcode93.app/Contents/Developer
 
 # Minimum Mono version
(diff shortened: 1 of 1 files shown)


rolfbjarne:
build

rolfbjarne:
In both test runs the same tests failed (and I haven't seen those tests fail in other PRs), which means it's highly likely that something in this PR breaks those tests.

Since the failing tests are related to detecting when things should be rebuilt (or not), my guess is that something related to subsecond timestamp resolution for files regressed.

lewurm:
@rolfbjarne thanks, the version initially specified didn't include all the timestamp related fixes. let's try with the latest 2018-02 build.

PR #4047, opened 2018-05-07
[msbuild] Updated MetalTaskBase to use the new -m{os}-version-min arg…

jstedfast:
…ument

Fixes issue #4019

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
+++ b/msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
@@ -1,5 +1,6 @@
 ﻿using System.IO;
 
+using Xamarin.MacDev;
 using Xamarin.MacDev.Tasks;
 
 namespace Xamarin.Mac.Tasks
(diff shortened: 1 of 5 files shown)


msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
```diff
@@ -1,13 +1,18 @@
...
 {
 	public class Metal : MetalTaskBase
 	{
 		protected override string OperatingSystem {
-			get { return "osx"; }
+			get { return "macosx"; }
+		}
+
+		protected override string MinimumDeploymentTargetKey {
+			get { return ManifestKeys.LSMinimumSystemVersion; }
```
spouliot: not a new problem but I thought the leaf classes where supposed to be empty so XVS could replace them (for remoting) ?

jstedfast: Yea, and for iOS we do that, but I think the Mac tasks were done before XVS started supporting that.

It's probably something we'll need to refactor into the same structure we have for iOS.

PR #4049, opened 2018-05-08
[msbuild] Updated MetalTaskBase to use the new -m{os}-version-min arg…

jstedfast:
…ument

Fixes issue #4019

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
+++ b/msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
@@ -1,5 +1,6 @@
 ﻿using System.IO;
 
+using Xamarin.MacDev;
 using Xamarin.MacDev.Tasks;
 
 namespace Xamarin.Mac.Tasks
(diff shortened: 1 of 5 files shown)


jstedfast:
This is the same as https://github.com/xamarin/xamarin-macios/pull/4047 but for master instead of the xcode9.4 branch.

msbuild/Xamarin.Mac.Tasks/Tasks/Metal.cs
```diff
@@ -1,13 +1,18 @@
...
 {
 	public class Metal : MetalTaskBase
 	{
 		protected override string OperatingSystem {
-			get { return "osx"; }
+			get { return "macosx"; }
+		}
+
+		protected override string MinimumDeploymentTargetKey {
+			get { return ManifestKeys.LSMinimumSystemVersion; }
```
spouliot: My previous comment/question remains :)
It's not a `Base` type so it should be empty right ?

spouliot: There's already partial support for XM, how does it work ?
@emaf ?

emaf: @spouliot, the partial support for XM we have on VS is just letting the users open the project and building it just on Windows, so we're not running tasks remotely on the Mac.

So for now this change is fine, and in the future if we want to support full builds for XM on VS we'll need to fix that using a base task on Xamarin.Mac.Tasks.Core. As Jeff mentioned, XM tasks are not following this appoach right now.

spouliot: @emaf thanks for clarifying
@jstedfast we can postpone that - it's unrelated, but `\Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Mac/Xamarin.Mac.Tasks.Core.dll` does exists :)

PR #4051, opened 2018-05-08
[mtouch] Take into account the contents of response files when determining whether the cache is valid or not. Fixes #4033. (#4032)

### LINKED ISSUE #4033: The contents of response files aren't taken into account when determining whether the cache is still valid or not
Repro:

0. Enable diagnostic build log.
1. Create a new project.
2. Add something to the additional mtouch arguments (for instance `-gcc_flags -v`).
3. Build
4. Change the additional mtouch arguments (for instance to `-gcc_flags -v -v`).
5. Build again. The build log should state that the cache was cleared:

> A full rebuild has been forced because the cache for ... is not valid.

Support for response files was added in d15-7, and since we changed the MSBuild tasks to write arguments into response files, this is a functional regression.

> VS bug [#611119](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/611119), VS bug [#611121](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/611121)

rolfbjarne:
* [mtouch] Take into account the contents of response files when determining whether the cache is valid or not.

* Correctly ignore new test on macOS.

Fixes #4033.

### DIFF
--- a/tests/common/BundlerTest.cs
+++ b/tests/common/BundlerTest.cs
@@ -21,6 +21,37 @@ namespace Xamarin
 	[TestFixture]
 	public class BundlerTests
 	{
+#if __MACOS__
+		// The cache doesn't work properly in mmp yet.
+		// [TestCase (Profile.macOSMobile)]
+#else
+		[Test]
+		[TestCase (Profile.iOS)]
+#endif
+		public void ModifiedResponseFile (Profile profile)
+		{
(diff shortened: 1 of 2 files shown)


spouliot:
QA verified -> https://github.com/xamarin/xamarin-macios/issues/4033#issuecomment-387546610

PR #4052, opened 2018-05-08
[mmp]: Reflect latest 'FtpRequestCreator' -> 'FtpWebRequestCreator' change.

baulig:


akoeplinger:
btw. the AttributeTest.DebuggerTypeProxy_24203 are expected because we're now using `Dictionary<T>` from corefx which uses [`IDictionaryDebugView<,>`](https://github.com/mono/corefx/blob/34e0d8b15c78140f71a68c1816e97e62a81d709e/src/Common/src/CoreLib/System/Collections/Generic/Dictionary.cs#L33) instead of [`Mscorlib_DictionaryDebugView<,>`](https://github.com/mono/mono/blob/1c91423be383634995b2b0a1484ec9b1b951c4f9/mcs/class/referencesource/mscorlib/system/collections/generic/dictionary.cs#L55).

The test needs to be fixed.

Issue #4046, opened 2018-05-07
VS MacOS build fails with: error MT0000: Unexpected error


dannyatftecu:
### Steps to Reproduce

1. Update to the latest VS Mac as of 5/7/2018 10AM PDT.
2. Build for iPhone [hardware]
3. Fail with "Unexpected error."
4. This build was working fine last week before I upgraded VS Mac.

### Expected Behavior

Build like iOS Simulator without error.

### Actual Behavior

Build error only for iOS device build.

### Environment

```
=== Visual Studio Community 2017 for Mac ===

Version 7.5 (build 1254)
Installation UUID: 0cdcd297-61e0-41a3-80cc-ce8ccd7de743
Runtime:
	Mono 5.10.1.47 (2017-12/8eb8f7d5e74) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.0.36 (master / 0c7c49a6)

	Package version: 510010047

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.0.5
	2.0.0
SDK: /usr/local/share/dotnet/sdk/2.1.4/Sdks
SDK Versions:
	2.1.4
	2.0.0
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.10.1/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.2
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Xamarin.Android ===

Version: 8.3.0.19 (Visual Studio Community)
Android SDK: /Users/nikhilshinde/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		4.4 (API level 19)
		5.0 (API level 21)
		6.0 (API level 23)
		7.0 (API level 24)
		7.1 (API level 25)
		8.1 (API level 27)

SDK Tools Version: 25.2.5
SDK Platform Tools Version: 27.0.1
SDK Build Tools Version: 27.0.3

Java SDK: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

[...]

PR #4055, opened 2018-05-09
[mtouch/mmp] Add proper error for condition encountered in issue #4046.

rolfbjarne:
There is no test case because I don't know how to reproduce the condition.


rolfbjarne:
Superseded by #4073.

PR #4060, opened 2018-05-09
[mk] Put shell code in shell scripts and add support for switching between different remotes for dependencies.

rolfbjarne:
Put the shell code for resetting README dependencies in a shell script instead
of embedded in the Makefile so that it's easier to write, read and debug.

Also add support for switching between different remotes for README
dependencies (this means that `make reset-X` will now work fine if `X`'s
remote changed).

A side effect is that all README dependencies will now end up with a 'xamarin'
remote in addition to the 'origin' remote, but there should be no other
side effects.

### DIFF
--- a/mk/xamarin-reset.sh
+++ b/mk/xamarin-reset.sh
@@ -1,0 +1,51 @@
+#!/bin/bash -e
+
+DEPENDENCY_NAME="$1"
+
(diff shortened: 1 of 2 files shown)


rolfbjarne:
Test failure is a known issue: https://github.com/xamarin/maccore/issues/581

PR #4061, opened 2018-05-09
[xharness] Improve logging of launched processes to not log the same information twice.

rolfbjarne:
Also log any added/changed/removed environment variables for launched
processes (compared to xharness itself).


rolfbjarne:
Test failure is a known issue: https://github.com/xamarin/maccore/issues/581

PR #4080, opened 2018-05-11
[msbuild] Adds minimum support for XM binding projects from VS

emaf:
Minimum support means:
- Letting users open XM binding projects
- Building XM binding projects offline (not connected to the Mac)


spouliot:
build

emaf:
@spouliot do you know why the Wrench check is failing? I don't see any link to check the failure

Issue #4087, opened 2018-05-13
Crash when setting RenderingApi on SCNRenderingOptions


praeclarum:
The key for `RenderingApi` seems to be incorrect on iOS 11.3 and using it results in an ArgumentNull exception.

### Steps to Reproduce

```csharp
new SCNView (UIScreen.MainScreen.Bounds, new SCNRenderingOptions {
            RenderingApi = SCNRenderingApi.Metal,
        })
```

1. Run that code on iOS 11.3
2. Note the crash

### Expected Behavior

SCNView is created with Metal.

### Actual Behavior

Crash

```
System.ArgumentNullException: Value cannot be null.
Parameter name: key
  at Foundation.DictionaryContainer.NullCheckAndRemoveKey (Foundation.NSString key, System.Boolean removeEntry) [0x0001e] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.10.1.174/src/Xamarin.iOS/Foundation/DictionaryContainer.cs:330 
  at Foundation.DictionaryContainer.SetNumberValue (Foundation.NSString key, System.Nullable`1[T] value) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.10.1.174/src/Xamarin.iOS/Foundation/DictionaryContainer.cs:390 
  at SceneKit.SCNRenderingOptions.set_RenderingApi (System.Nullable`1[T] value) [0x00009] in /Library/Frameworks/Xamarin.iOS.framework/Versions/11.10.1.174/src/Xamarin.iOS/SceneKit/SCNRenderingOptions.cs:20 
```

### Environment

```
=== Visual Studio Enterprise 2017 for Mac (Preview) ===

Version 7.5 Preview (7.5 build 1222)
Installation UUID: 4ba352e1-73e8-4dab-a9db-28012ca05af8
Runtime:
	Mono 5.10.1.47 (2017-12/8eb8f7d5e74) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.0.36 (master / 0c7c49a6)

	Package version: 510010047

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.1.0-preview2-26406-04
	2.0.3
SDK: /usr/local/share/dotnet/sdk/2.1.300-preview2-008530/Sdks
SDK Versions:
[...]

PR #4090, opened 2018-05-15
[SceneKit] Makes SCNRenderingOptions.RenderingApi use the right key

dalexsoto:
Fixes https://github.com/xamarin/xamarin-macios/issues/4087

`SCNRenderingOptions.RenderingApi` is using the wrong key to
set its dictionary container value that should come from
`SCNRenderingOptionsKeys.RenderingApiKey`.


rolfbjarne:
A test would have been nice :smile:

PR #4091, opened 2018-05-15
[Compression] Ensure that we do link against libcompression.

mandel-macaque:
We need to pass libcrompession in the linking flags, which was not
present and makes the build fail.


rolfbjarne:
Please link to the issue in the commit subject/message when squashing.

PR #4115, opened 2018-05-23
[msbuild] Log the name of the ibtool/actool log file if we cannot par…

jstedfast:
…se it

Improves diagnostics for https://github.com/xamarin/xamarin-macios/issues/4096

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XcodeCompilerToolTask.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XcodeCompilerToolTask.cs
@@ -191,7 +191,8 @@ namespace Xamarin.MacDev.Tasks
 						var plist = PDictionary.FromFile (manifest.ItemSpec);
 
 						LogWarningsAndErrors (plist, items[0]);
-					} catch {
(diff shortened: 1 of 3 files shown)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XcodeCompilerToolTask.cs
```diff
@@ -191,7 +191,8 @@ protected int Compile (ITaskItem[] items, string output, ITaskItem manifest)
 						var plist = PDictionary.FromFile (manifest.ItemSpec);
 
 						LogWarningsAndErrors (plist, items[0]);
-					} catch {
+					} catch (Exception ex) {
+						Log.LogError ("Failed to load {0} log file `{1}`: {2}", ToolName, manifest.ItemSpec, ex.Message);
```
rolfbjarne: Won't this show an error for something that was ignored before?

jstedfast: That was a good question though because I had to go back and check the code :)

PR #4128, opened 2018-05-24
[ObjCRuntime] Improve MarkDirty speed by caching the IsCustomType value in our flags byte.

rolfbjarne:
This makes calling MarkDirty for custom types ~60x times faster.

Reference: https://xamarinhq.slack.com/archives/C03CCJHCF/p1527194568000197

### DIFF
--- a/src/Foundation/NSObject2.cs
+++ b/src/Foundation/NSObject2.cs
@@ -79,6 +79,8 @@ namespace Foundation {
 			RegisteredToggleRef = 8,
 			InFinalizerQueue = 16,
 			HasManagedRef = 32,
+			KnowsIsCustomType = 64,
+			IsCustomType = 128,
 		}
 
 		bool disposed { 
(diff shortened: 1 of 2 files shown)


src/Foundation/NSObject2.cs
```diff
@@ -79,6 +79,8 @@ enum Flags : byte {
 			RegisteredToggleRef = 8,
 			InFinalizerQueue = 16,
 			HasManagedRef = 32,
+			// 64, // Used by SoM
```
spouliot: `runtime.h` should be updated

src/Foundation/NSObject2.cs
```diff
@@ -103,6 +105,19 @@ enum Flags : byte {
 			get { return ((flags & Flags.InFinalizerQueue) == Flags.InFinalizerQueue); }
 		}
 
+		bool IsCustomType {
+			get {
+				bool value;
+				var value = (flags & Flags.IsCustomType) == Flags.IsCustomType;
+				if (!value) {
```
spouliot: that's only faster for custom types
i.e. it still calls the (costly?) method for non-custom types
not sure if this matters (in other cases than the one provided)

rolfbjarne: If `IsCustomType` returns false, `IsRegisteredToggleRef` will be set to true, and the next time `MarkDirty` is called on the same object it'll exit immediately.

This obviously only matters if `MarkDirty` is called multiple times per instance, which is exactly what's happening.

src/Foundation/NSObject2.cs
rolfbjarne: Fixed.

PR #4138, opened 2018-05-28
[jenkins] Improve api/generator diff reporting to say if there were changes, and if they were breaking or not.

rolfbjarne:
For the api diff for this PR we now show:

* `🔥 breaking changes 🔥`: If there are any breaking changes in the api diff (for this PR).
* `please review changes`: If there are any non-breaking changes in the api diff.
* `no change`: If the api diff is empty.

For the generator diff we show:

* `only version changes`: If there were only changes related to version numbers (since the XI/XM version number is added to the generator, that version number will always show up as a diff when comparing the generated source code)
* `please review changes`: If anything other that version numbers changed in the generated source code.


rolfbjarne:
`do-not-merge` since I'll have to push a few test commits to actually test this properly.

rolfbjarne:
@spouliot this is ready for review now.

The two monojenkins comments above show how it shows up with breaking API changes and generator changes (first comment), and non-breaking API changes (second comment)

PR #4145, opened 2018-05-29
[xharness] Trace exceptions to have better information when something goes very wrong.

rolfbjarne:


rolfbjarne:
> is that not quite verbose ?

It's somewhat verbose (depends on how many exceptions are thrown, there's at least one, usually a few more, but not that many).

They're printed to stdout (Jenkins' console view), which is usually not where we spend time looking for anything (unless I'm trying to figure out why xharness is broken, in which case this can save a lot of time).

PR #4151, opened 2018-05-29
[mtouch][mmp] Cache ReaderParameters instances in CoreResolver

spouliot:
In my test project 969828 instances (67MB) of `ReaderParameters` were
created by `CoreResolver.Resolve`, which always allocated a new one.

Since we already create them, most of the time*, we can reuse the
instances when additional members requires resolving.

* only 14 additional instances are now required

### DIFF
--- a/tools/common/CoreResolver.cs
+++ b/tools/common/CoreResolver.cs
@@ -17,6 +17,7 @@ namespace Xamarin.Bundler {
 	public abstract class CoreResolver : IAssemblyResolver {
 
 		internal Dictionary<string, AssemblyDefinition> cache;
+		Dictionary<string,ReaderParameters> params_cache;
 
 		public string FrameworkDirectory { get; set; }
 		public string RootDirectory { get; set; }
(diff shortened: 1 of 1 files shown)


tools/common/CoreResolver.cs
```diff
@@ -17,6 +17,7 @@ namespace Xamarin.Bundler {
 	public abstract class CoreResolver : IAssemblyResolver {
 
 		internal Dictionary<string, AssemblyDefinition> cache;
+		Dictionary<string,ReaderParameters> params_cache;
```
rolfbjarne: Nitpick: space between `string`and `ReaderParameters`.

tools/common/CoreResolver.cs
```diff
@@ -45,7 +47,12 @@ public void Dispose ()
 
 		public AssemblyDefinition Resolve (AssemblyNameReference name)
 		{
-			return Resolve (name, new ReaderParameters { AssemblyResolver = this });
+			var key = name.ToString ();
+			if (!params_cache.TryGetValue (key, out ReaderParameters parameters)) {
+				parameters = new ReaderParameters { AssemblyResolver = this };
+				params_cache [key] = parameters;
```
rolfbjarne: Why does this need to be keyed using the assembly name? Does anything in the `ReaderParameters` change depending on the assembly?

spouliot: I thought of sharing it across assemblies (one for **in memory**, one for **on disk**).
However it's mutable and it's used by code outside our reach (cecil and linker) making any assumption risky.

PR #4155, opened 2018-05-29
Fix HTML comments in Issue Template

therealjohn:
Added the missing exclamation mark
New lines in the comments seemed to break the comment, so I removed them
This should prevent the template from appearing in the final issue when the user posts it


rolfbjarne:

PR #4161, opened 2018-05-30
Improve IsCustomType performance

ermau:
While NSObject may now flag internally whether it is a custom type, this still has to be looked up each time a new instance of that type is created.

The dynamic registrar has a fast lookup for whether a type is a custom type or not, but `IsCustomType` first checks the static registrar regardless. `GetClassHandle` already caches the class handles, so this extends it to include whether or not the type is custom so then `IsCustomType` can use the cached path.

This test case is improved from `0.27ms` per to `0.09ms` per new view when using the dynamic registrar.

```csharp
public class Button
    : NSView
{
    public Button()
    {
        checkbox = new NSButton
        {
            AllowsMixedState = true,
            ControlSize = NSControlSize.Small,
            Title = string.Empty,
            TranslatesAutoresizingMaskIntoConstraints = false
        };
        checkbox.SetButtonType(NSButtonType.Switch);
        checkbox.Activated += Button_Activated;

        AddSubview(checkbox);
    }

    private NSButton checkbox;

    void Button_Activated (object sender, EventArgs e)
    {
    }
}
```

### DIFF
--- a/src/ObjCRuntime/Class.cs
+++ b/src/ObjCRuntime/Class.cs
@@ -23,18 +23,15 @@ namespace ObjCRuntime {
 #if !COREBUILD
 		public static bool ThrowOnInitFailure = true;
 
-		static Dictionary<Type, ClassCache> type_to_class; // accessed from multiple threads, locking required.
-		struct ClassCache {
-			internal IntPtr handle;
-			internal bool is_custom_type;
-		}
+		// We use the last significant bit of the IntPtr to store if this is a custom class or not.
+		static Dictionary<Type, IntPtr> type_to_class; // accessed from multiple threads, locking required.
 
 		internal IntPtr handle;
 
 		[BindingImpl (BindingImplOptions.Optimizable)]
 		internal unsafe static void Initialize (Runtime.InitializationOptions* options)
 		{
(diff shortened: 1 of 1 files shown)


src/ObjCRuntime/Class.cs
```diff
@@ -23,14 +23,18 @@ public partial class Class : INativeObject
 #if !COREBUILD
 		public static bool ThrowOnInitFailure = true;
 
-		static Dictionary<Type, IntPtr> type_to_class; // accessed from multiple threads, locking required.
+		static Dictionary<Type, ClassCache> type_to_class; // accessed from multiple threads, locking required.
+		struct ClassCache {
+			internal IntPtr handle;
+			internal bool is_custom_type;
+		}
```
spouliot: I'd rather avoid an extra struct / allocation in that code.

@rolfbjarne can we, like we do in `NSObject` use the LSB of the class handle to store this extra data ?

rolfbjarne: @spouliot yes, we could use the LSB of the class handle to store if it's a custom type or not. I'll have a look at that.

spouliot:
link sdk/watchOS - simulator/Debug: Failed -> known issue (junk in XML) https://github.com/xamarin/maccore/issues/827

PR #4162, opened 2018-05-31
Optimize NSActionDispatcher usage

Therzok:
This PR adds the following value:

1. There is no Action wrapper being constructed on async continuations,
thus on every await call we gain: 1 less allocation (lambda capture),
1 less indirected call to the actual continuation, and possibly
System.Action being removed by the linker, as its no longer used in these cases.

2. NSActionDispatcher* classes can now be linked out completely, due to
the static selector no longer being used everywhere

3. One unused class removed

### DIFF
--- a/src/Foundation/NSAction.cs
+++ b/src/Foundation/NSAction.cs
@@ -54,31 +54,6 @@ namespace Foundation {
 		}
 	}
 
-	// Use this for synchronous operations
-	[Register ("__MonoMac_ActionDispatcher")]
-	internal sealed class ActionDispatcher : NSObject {
-		public const string SelectorName = "xamarinApplySelector";
-		public static readonly Selector Selector = new Selector (SelectorName);
-
-		readonly Action action;
-
-		public ActionDispatcher (Action action)
-		{
-			if (action == null)
-				throw new ArgumentNullException ("action");
-
-			this.action = action;
-			IsDirectBinding = false;
-		}
-
-		[Export (SelectorName)]
-		[Preserve (Conditional = true)]
-		public void Apply ()
-		{
-			action ();
-		}
-	}
-
 	// Used this for NSTimer support
 	[Register ("__Xamarin_NSTimerActionDispatcher")]
 	internal sealed class NSTimerActionDispatcher : NSObject {
--- a/src/Foundation/NSAction.cs
+++ b/src/Foundation/NSAction.cs
@@ -34,9 +34,7 @@ namespace Foundation {
 	[Register ("__MonoMac_NSActionDispatcher")]
 	internal sealed class NSActionDispatcher : NSObject {
 		public const string SelectorName = "xamarinApplySelector";
-#if MONOMAC
 		public static readonly Selector Selector = new Selector (SelectorName);
-#endif
 
 		readonly Action action;
 
@@ -91,9 +89,7 @@ namespace Foundation {
 	[Register ("__Xamarin_NSTimerActionDispatcher")]
 	internal sealed class NSTimerActionDispatcher : NSObject {
 		public const string SelectorName = "xamarinFireSelector:";
-#if MONOMAC
 		public static readonly Selector Selector = new Selector (SelectorName);
-#endif
 
 		readonly Action<NSTimer> action;
 
--- a/src/Foundation/NSAction.cs
+++ b/src/Foundation/NSAction.cs
@@ -159,7 +159,7 @@ namespace Foundation {
 
 	// Use this for asynchronous operations
 	[Register ("__MonoMac_NSAsyncSynchronizationContextDispatcher")]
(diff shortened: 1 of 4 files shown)


src/Foundation/NSAction.cs
```diff
@@ -55,27 +56,26 @@ public void Apply ()
...
+		readonly object state;
 
-		public ActionDispatcher (Action action)
+		public NSActionDispatcher2 (SendOrPostCallback d, object state)
 		{
-			if (action == null)
-				throw new ArgumentNullException ("action");
-
-			this.action = action;
-			IsDirectBinding = false;
```
spouliot: you can't remove `IsDirectBinding = false;` since there's managed state in the instance

src/Foundation/NSAction.cs
```diff
@@ -107,6 +107,8 @@ public void Fire (NSTimer timer)
 	// Use this for asynchronous operations
 	[Register ("__MonoMac_NSAsyncActionDispatcher")]
 	internal class NSAsyncActionDispatcher : NSObject {
+		public const string SelectorName = "xamarinApplySelector";
+		public static readonly Selector Selector = new Selector (SelectorName);
```
spouliot: the `Selector` should beside a `#if MONOMAC` since it's not used for XI

src/Foundation/NSAction.cs
```diff
@@ -142,6 +144,52 @@ public void Apply ()
 			}
 		}
 	}
+
+	// Use this for asynchronous operations
+	[Register ("__MonoMac_NSAsyncActionDispatcher2")]
+	internal class NSAsyncActionDispatcher2 : NSObject
+	{
+		public const string SelectorName = "xamarinApplySelector";
+		public static readonly Selector Selector = new Selector (SelectorName);
```
spouliot: same, the `Selector` should beside a `#if MONOMAC` since it's not used for XI

src/Foundation/NSAction.cs
```diff
@@ -55,27 +56,26 @@ public void Apply ()
...
 
-		readonly Action action;
+		readonly SendOrPostCallback d;
+		readonly object state;
 
-		public ActionDispatcher (Action action)
+		public NSActionDispatcher2 (SendOrPostCallback d, object state)
 		{
-			if (action == null)
-				throw new ArgumentNullException ("action");
```
spouliot: I think the null check is still required

src/Foundation/NSAction.cs
Therzok: Ah, my bad.

src/Foundation/NSAction.cs
Therzok: Well, these calls come in from SynchronizationContext. If someone schedules a null task, there's bigger problems.

src/Foundation/NSAction.cs
Therzok: 👍 Will fix others too.

src/Foundation/NSAction.cs
dalexsoto: While I agree with the reasoning of not having the null check, it is always a great idea to be defensive so :+1: on having the null check here

src/Foundation/NSAction.cs
```diff
@@ -55,35 +58,42 @@ public void Apply ()
 	}
 
 	// Use this for synchronous operations
-	[Register ("__MonoMac_ActionDispatcher")]
-	internal sealed class ActionDispatcher : NSObject {
+	[Register ("__MonoMac_NSActionDispatcher2")]
+	internal sealed class NSActionDispatcher2 : NSObject
```
rolfbjarne: Why was the class renamed? It doesn't look necessary, and it makes the diff harder to understand.

src/Foundation/NSAction.cs
```diff
@@ -138,6 +155,57 @@ public void Apply ()
...
 				// Dispose here.
 				Dispose ();
+#endif
+			}
+		}
+	}
+
+	// Use this for asynchronous operations
+	[Register ("__MonoMac_NSAsyncActionDispatcher2")]
+	internal class NSAsyncActionDispatcher2 : NSObject
```
rolfbjarne: This class is almost entirely a copy-paste of `NSAsyncActionDispatcher`, and as such it seems it would be a good idea to introduce a base class for the common code.

src/Foundation/NSAction.cs
Therzok: This wasn't renamed. Check commit-by-commit. It removed an unused class.

src/Foundation/NSAction.cs
Therzok: Sure! I can do that. I just thought about the benefits of not having virtual calls and co, but I can seal the classes. Will fix.

PR #4163, opened 2018-05-31
Update XML references wrt XI 11.12 (d15-7 with xcode9.4 support) and bump to 11.13

spouliot:


tools/apidiff/.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
+{
```
rolfbjarne: Did you mean to include this file?

rolfbjarne:
Looks like I'll be fixing https://github.com/xamarin/maccore/issues/830 quite soon...

tools/apidiff/.vscode/launch.json
spouliot: no :)

spouliot:
build

spouliot:
@rolfbjarne only the XI version number changed, so that should not 🔥 a breaking change warning

spouliot:
and _ideally_ the *please review changes* for the generator should be skipped (since that will always change even if the API are identical).

spouliot:
@rolfbjarne oh, it's a `const` so it make sense (from API diff point of view)

PR #4167, opened 2018-05-31
[xharness] Hide summary of test failures in an expandable block in the markdown summary.

rolfbjarne:
Hide summary of test failures in an expandable block in the markdown summary,
so that when there are many failures it's still a summary and not pages and
pages of text.

Horror example this will fix: https://github.com/xamarin/xamarin-macios/pull/4163#issuecomment-393566282


spouliot:
much nicer output :) issues all look to be from (now fixed) https://github.com/xamarin/maccore/issues/830

spouliot:
build

PR #4169, opened 2018-06-01
[Compression] Ensure we use the correct linking flags for older versions. Fixes #4129.

### LINKED ISSUE #4129: Can't launch monotouch-test on iOS <9
Repro:

* Run monotouch-test on an older (< iOS 9.0) device (in the `Release (all optimizations)` configuration using xharness)

Crashes at launch with

```
Dyld Error Message:
  Library not loaded: /usr/lib/libcompression.dylib
  Referenced from: /private/var/mobile/Containers/Bundle/Application/04F52304-EF42-4491-BE2B-45BD9062C875/monotouchtest.app/monotouchtest
  Reason: image not found
  Dyld Version: 353.10
```

which makes sense since libcompression.dylib was introduced in iOS 9: https://developer.apple.com/documentation/compression?language=objc

Probable fix: if the deployment target is < iOS 9 / macOS 10.11 / tvOS 9.0 / watchOS 2.0, then use `-weak-lcompression` instead of `-lcompression` here: https://github.com/xamarin/xamarin-macios/blob/238eb94caacb79051eae1e28143dc6999d935d78/tools/common/Assembly.cs#L367-L370

mandel-macaque:
Libcompresison was added after iOs 9.0, TvOS 9.0, Mac 10.11 and Watch
2.0. We want to use weak in those older platforms.

Fixes issue https://github.com/xamarin/xamarin-macios/issues/4129

### DIFF
--- a/tests/monotouch-test/Compression/CompressionStreamTest.cs
+++ b/tests/monotouch-test/Compression/CompressionStreamTest.cs
@@ -44,6 +44,8 @@ namespace MonoTouchFixtures.Compression {
 		[ExpectedException (typeof (ArgumentNullException))]
 		public void Constructor_Null ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			DeflateStream ds = new DeflateStream (null, CompressionMode.Compress, CompressionAlgorithm.Zlib);
 		}
 
@@ -51,6 +53,8 @@ namespace MonoTouchFixtures.Compression {
 		[ExpectedException (typeof (ArgumentException))]
 		public void Constructor_InvalidCompressionMode ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			DeflateStream ds = new DeflateStream (new MemoryStream (), (CompressionMode)Int32.MinValue, CompressionAlgorithm.Zlib);
 		}
 
@@ -60,6 +64,8 @@ namespace MonoTouchFixtures.Compression {
 		[TestCase (CompressionAlgorithm.Zlib)]
 		public void CheckCompressDecompress (CompressionAlgorithm algorithm)
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			byte [] data = new byte[100000];
 			for (int i = 0; i < 100000; i++) {
 				data[i] = (byte) i;
@@ -83,6 +89,8 @@ namespace MonoTouchFixtures.Compression {
 		[Test] // Not passing the algorithm because the compressed data is Zlib compressed.
 		public void CheckDecompress ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			MemoryStream backing = new MemoryStream (compressed_data);
 			DeflateStream decompressing = new DeflateStream (backing, CompressionMode.Decompress, CompressionAlgorithm.Zlib);
 			StreamReader reader = new StreamReader (decompressing);
@@ -94,6 +102,8 @@ namespace MonoTouchFixtures.Compression {
 		[ExpectedException (typeof (ArgumentNullException))]
 		public void CheckNullRead ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			MemoryStream backing = new MemoryStream (compressed_data);
 			DeflateStream decompressing = new DeflateStream (backing, CompressionMode.Decompress, CompressionAlgorithm.Zlib);
 			decompressing.Read (null, 0, 20);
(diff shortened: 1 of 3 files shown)


mandel-macaque:
I have no access to and old device. @GouriKumari can you take a look in a real device in the lab? Would really appreciate it!

mandel-macaque:
build

rolfbjarne:
build

tools/common/Assembly.cs
```diff
@@ -336,6 +336,27 @@ void AddFramework (string file)
...
+				return "-weak-lcompression";
+			case ApplePlatform.TVOS:
+			case ApplePlatform.iOS:
+				if (App.SdkVersion >= new Version (9,0))
+					return "-lcompression";
+				return "-weak-lcompression";
+			case ApplePlatform.WatchOS:
+				if (App.SdkVersion >= new Version (2, 0))
+					return "-lcompression";
+				return "-weak-lcompression";
```
spouliot: 2.0 is the minimum for watchOS so it's always supported

tools/common/Assembly.cs
```diff
@@ -336,6 +336,27 @@ void AddFramework (string file)
...
 		}
 
+		public string GetCompressionLinkingFlag ()
+		{
+			switch(App.Platform) {
+			case ApplePlatform.MacOSX:
+				if (App.SdkVersion >= new Version (10, 11, 0))
+					return "-lcompression";
+				return "-weak-lcompression";
+			case ApplePlatform.TVOS:
```
spouliot: 9,0 is the minimum for tvOS so it's always supported

tools/common/Assembly.cs
mandel-macaque: 👍

mandel-macaque:
@spouliot indeed, simplified and updated the PR.

GouriKumari:
@mandel-macaque : Is there a build for this?

tools/common/Assembly.cs
```diff
@@ -336,6 +336,25 @@ void AddFramework (string file)
...
+				return "-weak-lcompression";
+			case ApplePlatform.iOS:
+				if (App.SdkVersion >= new Version (9,0))
+					return "-lcompression";
+				return "-weak-lcompression";
+			case ApplePlatform.TVOS:
+			case ApplePlatform.WatchOS:
+				return "-lcompression";
+			default:
+				return null;
```
spouliot: minor but, for consistency, this should report an error like other places, i.e.
```
				throw ErrorHelper.CreateError (71, "Unknown platform: {0}. This usually indicates a bug in {1}; please file a bug report at http://bugzilla.xamarin.com with a test case.", platform, Application.ProductName);
```

tools/common/Assembly.cs
```diff
@@ -364,11 +383,15 @@ public void ComputeLinkerFlags ()
...
-					case "libcompression":
 						// remove lib prefix
 						LinkerFlags.Add ("-l" + file.Substring (3));
 						Driver.Log (3, "Linking with {0} because it's referenced by a module reference in {1}", file, FileName);
-					break;
+						break;
+					case "libcompression":
+						var compressionLinkingFlag = GetCompressionLinkingFlag ();
+						if (!string.IsNullOrEmpty (compressionLinkingFlag))
+							LinkerFlags.Add (compressionLinkingFlag);
```
spouliot: and you can simplify that, it won't ever be empty

mandel-macaque:
@GouriKumari there is no need for QA to take a look. I have been able to reproduce the bug using an old simulator. With  master, and a iOS 8.4 machine you get the following:

```
objc[76488]: Class AMSupportURLConnectionDelegate is implemented in both /System/Library/PrivateFrameworks/EmbeddedOSInstall.framework/Versions/A/EmbeddedOSInstall (0x7fff85a3b748) and /System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x106888368). One of the two will be used. Which one is undefined.
objc[76488]: Class AMSupportURLSession is implemented in both /System/Library/PrivateFrameworks/OSPersonalization.framework/Versions/A/OSPersonalization (0x7fff86d5c9f0) and /System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x1068883b8). One of the two will be used. Which one is undefined.
dyld: Library not loaded: /usr/lib/libcompression.dylib
  Referenced from: /Users/mandel/Library/Developer/CoreSimulator/Devices/0B14050F-DECB-43EB-97F1-3914A9BC3E83/data/Containers/Bundle/Application/65401855-DB5F-4EC5-836C-199FF91B8BF0/monotouchtest.app/monotouchtest
  Reason: no suitable image found.  Did find:
	/usr/lib/libcompression.dylib: mach-o, but not built for iOS simulat
```or

With the fix, the test application will run as expected in an old device.

File: tests/monotouch-test/Compression/CompressionStreamTest.cs [hunk 5]
```diff
@@ -44,13 +44,17 @@ private static bool compare_buffers (byte[] first, byte[] second, int length)
 		[ExpectedException (typeof (ArgumentNullException))]
 		public void Constructor_Null ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
```
rolfbjarne: _For the future_: there's `TestRuntime.AssertXcodeVersion` as well, which allows you to do this in a single line.

mandel-macaque:
@rolfbjarne I quite do not understand the build result. Did it fail?

spouliot:
@mandel-macaque `Aborted by user Connor Adsit` <- I ping'ed you on slack about it (updating bots to Xcode 10)

spouliot:
build

PR #4172, opened 2018-06-01
[msbuild] Change MtouchHttpClientHandler default

VincentDondain:
If the Http Client value isn't set in the csproj, we should default to `NSUrlSessionHandler` which is also what the Xamarin.iOS Analysis rules try to enforce.


dalexsoto:
❤️❤️❤️❤️❤️❤️

7drhiqr

VincentDondain:
Filed an issue for the unrelated test failure: https://github.com/xamarin/maccore/issues/977.

PR #4179, opened 2018-06-04
Bump to use Xcode 10 beta 1

spouliot:


spouliot:
xtro and tests will be updated / fixed in a later commit

spouliot:
Failure with `make[2]: *** [.stamp-configure-mac32] Error 1`
since 10.14 SDK does not compile 32 bits macOS applications.
Which means we need to disable classic and unified 32bits builds

spouliot:
build

msbuild/Xamarin.iOS.Tasks.Core/Tasks/ResolveNativeWatchAppTaskBase.cs
```diff
@@ -55,6 +55,10 @@ public override bool Execute ()
 				version = currentSdk.GetClosestInstalledSdk (version, false);
 				sdk_path = currentSdk.GetSdkPath (version, SdkIsSimulator);
 			} else {
+				if (AppleSdkSettings.XcodeVersion.Major >= 10) {
+					Log.LogError ("Xcode 10 does not support watchOS 1 apps. Either upgrade to watchOS 2 apps, or use an earlier version of Xcode.");
+					return false;
```
jstedfast: Maybe instead of saying "earlier version", say "older version". I think that might read better in English. At least *I* usually say "older version".

PR #4182, opened 2018-06-05
[msbuild] Fix build with Xcode 10 beta 1.

rolfbjarne:
Many years ago (in Xcode 7 according to code comment)
Developer/Platforms/iPhoneOS.platform/Developer/usr disappeared, and we coped
by looking at Developer/usr instead (and also the subsequent code to locate
the bin directory was based on the location of the usr directory).

Developer/Platforms/iPhoneOS.platform/Developer/usr reappeared in Xcode 10
beta 1, but it seems useless (for one it doesn't contain a bin directory), so
in order to try to keep things sane don't look for this directory in Xcode 10
and instead go directly for Developer/usr (which is what we've been using as
the usr directory for years anyway).

Fixes this problem when building apps with Xcode 10 beta 1:

      /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/iOS/Xamarin.iOS.Common.targets(626,3): error : Could not locate SDK bin directory [/Users/rolf/Projects/TestApp/test-app.csproj]


rolfbjarne:
We might want to try to get this in a d15-7 SR, since as far as my (very limited) testing goes, it's the only thing that prevents d15-7 from working with Xcode 10.

PR #4184, opened 2018-06-05
[Compression] Ensure we use the correct linking flags for older versions. Fixes #4129. (#4169)

### LINKED ISSUE #4129: Can't launch monotouch-test on iOS <9

mandel-macaque:
Libcompresison was added after iOs 9.0, TvOS 9.0, Mac 10.11 and Watch
2.0. We want to use weak in those older platforms.

Fixes issue https://github.com/xamarin/xamarin-macios/issues/4129

This is a cherry-pick of https://github.com/xamarin/xamarin-macios/pull/4169

### DIFF
--- a/tests/monotouch-test/Compression/CompressionStreamTest.cs
+++ b/tests/monotouch-test/Compression/CompressionStreamTest.cs
@@ -44,6 +44,8 @@ namespace MonoTouchFixtures.Compression {
 		[ExpectedException (typeof (ArgumentNullException))]
 		public void Constructor_Null ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
+				Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
 			DeflateStream ds = new DeflateStream (null, CompressionMode.Compress, CompressionAlgorithm.Zlib);
 		}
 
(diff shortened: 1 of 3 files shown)


tests/monotouch-test/Compression/CompressionStreamTest.cs
```diff
@@ -44,13 +44,17 @@ private static bool compare_buffers (byte[] first, byte[] second, int length)
 		[ExpectedException (typeof (ArgumentNullException))]
 		public void Constructor_Null ()
 		{
+			if (!TestRuntime.CheckXcodeVersion (7, 0))
```
dalexsoto: All of those new checks are identical maybe you can do something like this

```csharp
[SetUp]
public void SetUp ()
{
	if (!TestRuntime.CheckXcodeVersion (7, 0))
		Assert.Ignore ("Requires iOS 9.0+ or macOS 10.11+");
}
```

mandel-macaque:
@dalexsoto yeah, it was a good idea, too late :/

PR #4198, opened 2018-06-07
Xcode10 CoreGraphics B1-B5

migueldeicaza:
Additionally, completed CFPropertyList


src/CoreGraphics/CGColorSpace.cs
```diff
@@ -80,6 +80,13 @@ public CGColorSpace (IntPtr handle)
 			CGColorSpaceRetain (handle);
 		}
 
+		public CGColorSpace (CFPropertyList propertyList)
+		{
+			if (propertyList == null)
+				throw new ArgumentNullException (nameof(propertyList));
```
mandel-macaque: Small style issues. Missing space after nameof.

src/CoreFoundation/CFPropertyList.cs
```diff
@@ -45,6 +42,51 @@ public CFPropertyList (IntPtr handle)
...
+			return new CFPropertyList (CFPropertyListCreateDeepCopy (IntPtr.Zero, handle, options));
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		extern static /*CFDataRef*/IntPtr CFPropertyListCreateData(IntPtr allocator, IntPtr propertyList, CFPropertyListFormat format, CFPropertyListMutabilityOptions options, out IntPtr error);
+
+		public NSData AsData (CFPropertyListFormat format = CFPropertyListFormat.BinaryFormat1)
+		{
+			IntPtr error;
+			return new NSData (CFPropertyListCreateData (IntPtr.Zero, handle, format, 0, out error));
```
mandel-macaque: I've got a question here. How is the user going to access the error if it occurrs? I think we should surface that.

src/CoreFoundation/CFPropertyList.cs
```diff
@@ -14,10 +14,7 @@
 
 namespace CoreFoundation
 {
-	// This is currently incomplete and thus marked internal;
-	// it is only used by CFPreferences, and I'm not necessarily
-	// sure we'd want to expose it? -abock
-	internal class CFPropertyList : INativeObject, IDisposable
+	public class CFPropertyList : INativeObject, IDisposable
```
rolfbjarne: This needs;

1. A review of the entire `CFPropertyList` API.
2. Tests for the entire  `CFPropertyList` API.
3. It needs a `CFPropertyList (IntPtr handle, bool owns)` constructor, and the `CFPropertyList (IntPtr)` constructor must call it with `owns=false` so that it matches the behaviour of other INativeObject types (this changes the existing behaviour, which also means all existing (internal) usages of `CFPropertyList` must be audited.

rolfbjarne: WIth 1. I meant review everything _already bound_ (so that we don't make something public that's broken).

spouliot:
build

src/CoreFoundation/CFPropertyList.cs
migueldeicaza: I did (1)

src/CoreGraphics/CGColorSpace.cs
```diff
@@ -506,6 +513,25 @@ public NSData GetIccData ()
...
+		static extern IntPtr CGColorSpaceCreateWithPropertyList (IntPtr plist);
+
+		[iOS(10,0)][Mac(10,12)]
+		[TV(10,0)][Watch(5,0)]
+		public CFPropertyList ToPropertyList ()
+		{
+			var x = CGColorSpaceCopyPropertyList (handle);
+			if (x == IntPtr.Zero)
+				return null;
+			return new CFPropertyList (x);
```
spouliot: calling that `.ctor` will call retain on something coming out of `*Copy*` (already retained)

src/CoreFoundation/CFPropertyList.cs
rolfbjarne: I've now done all of these.

PR #4209, opened 2018-06-11
[arkit] Update to Xcode 10 beta 1 (ARKit 2.0)

VincentDondain:
I obsoleted `GetProjectPoint` in favor of `Project` because of the introduction of `Unproject` and based on the API doc https://developer.apple.com/documentation/arkit/arcamera/2923538-projectpoint?language=objc. The `CGPoint` returned is the projection of a point. An other option would have been to name the method `GetProjectedPoint` but I think `Project` is closer to the original and clear enough.


src/arkit.cs
```diff
@@ -177,9 +210,20 @@ interface ARCamera : NSCopying {
 			get;
 		}
 
+#if !XAMCORE_4_0
+		[Obsolete ("Use 'ProjectPoint' instead.")]
```
dalexsoto: Makes sense but I think you meant `Project` instead of `ProjectPoint` 😉

src/arkit.cs
```diff
@@ -1045,6 +1158,128 @@ interface ARDirectionalLightEstimate {
...
+
+		[Export ("exportObjectToURL:previewImage:error:")]
+		bool ExportObject (NSUrl url, [NullAllowed] UIImage previewImage, [NullAllowed] out NSError error);
+
+		[Export ("referenceObjectByApplyingTransform:")]
+		[MarshalDirective (NativePrefix = "xamarin_simd__", Library = "__Internal")]
+		ARReferenceObject GetReferenceObject (Matrix4 transform);
+	}
+
+	[iOS (12,0)]
```
dalexsoto: just curious why `[NoWatch, NoTV, NoMac]` isn't present here but it is above??

dalexsoto: Same for `ARWorldMap`

src/arkit.cs
spouliot: should not it be `Project` ?

src/arkit.cs
```diff
@@ -1045,6 +1158,128 @@ interface ARDirectionalLightEstimate {
...
+		[Export ("autoFocusEnabled")]
+		bool AutoFocusEnabled { [Bind ("isAutoFocusEnabled")] get; set; }
+
+		[Export ("planeDetection", ArgumentSemantic.Assign)]
+		ARPlaneDetection PlaneDetection { get; set; }
+	}
+
+	[iOS (12,0)]
+	[NoWatch, NoTV, NoMac]
+	[BaseType (typeof(ARAnchor))]
```
spouliot: Considering the `init*` don't accept `nil` is seems unlikely the default constructor should be allowed.

src/arkit.cs
```diff
@@ -1045,6 +1158,128 @@ interface ARDirectionalLightEstimate {
...
+			get;
+		}
+
+		[Export ("rawFeaturePoints", ArgumentSemantic.Strong)]
+		ARPointCloud RawFeaturePoints { get; }
+
+		[Static]
+		[Export ("referenceObjectsInGroupNamed:bundle:")]
+		[return: NullAllowed]
+		NSSet<ARReferenceObject> GetReferenceObjectsInGroup (string name, [NullAllowed] NSBundle bundle);
```
spouliot: last part of the selector is the first parameter name
`GetReferenceObjects (string groupName, ...)` seems better

src/arkit.cs
```diff
@@ -1045,6 +1158,128 @@ interface ARDirectionalLightEstimate {
...
+		[Export ("exportObjectToURL:previewImage:error:")]
+		bool ExportObject (NSUrl url, [NullAllowed] UIImage previewImage, [NullAllowed] out NSError error);
+
+		[Export ("referenceObjectByApplyingTransform:")]
+		[MarshalDirective (NativePrefix = "xamarin_simd__", Library = "__Internal")]
+		ARReferenceObject GetReferenceObject (Matrix4 transform);
+	}
+
+	[iOS (12,0)]
+	[BaseType (typeof(ARAnchor))]
```
spouliot: the base type has it's default actor marked as not available
unless mentioned explicitly in the headers this should not expose it

stephen-hawley:
:eyes:

spouliot:
xtro files were not updated

src/arkit.cs
VincentDondain: Ah yes forgot to update that, thanks.

src/arkit.cs
VincentDondain: Yup read that selector wrong

src/arkit.cs
VincentDondain: Good catch I missed those.

VincentDondain:
@rolfbjarne I have to revisit the things I added to ARKit.ignore, I'm not quite sure why they are reported yet given the typedef (simd_floatX is typedefed to vector_floatX).

Also what tests were you talking about in your comment? The `Simd` monotouch-tests or ARKit specific tests? I don't recall specific tests for `MarshalDirective`.

rolfbjarne:
@VincentDondain 

> Also what tests were you talking about in your comment? The Simd monotouch-tests or ARKit specific tests? I don't recall specific tests for MarshalDirective.

Just monotouch-tests exercising all API with a MarshalDirective.

spouliot:
Some types might not be user-creatable so doing unit tests might not be possible. @rolfbjarne did you have an idea for them ?

If not let's do tests for the types that are (user createable) inside this PR. For others we better have a sample that exercise the API and, having the PR merged, might also give a chance for more people to try it out.

rolfbjarne:
> Some types might not be user-creatable so doing unit tests might not be possible. 

If they're not user-creatable (and there's no other trivial API that creates them), then we should just skip doing monotouch-tests for them, and instead rely on samples.

VincentDondain:
Comments addressed, I added tests for everything that was user-creatable.

_Bonus: the .arobject 3D model was generated from the real world object bellow._

model3

VincentDondain:
@rolfbjarne @dalexsoto please re-review (:

src/arkit.cs
```diff
@@ -177,9 +210,20 @@ interface ARCamera : NSCopying {
 			get;
 		}
 
+#if !XAMCORE_4_0
+		[Obsolete ("Use 'Project' instead.")]
+		[Wrap ("Project (point, orientation, viewportSize)")]
+		CGPoint GetProjectPoint (Vector3 point, UIInterfaceOrientation orientation, CGSize viewportSize);
```
rolfbjarne: This is a breaking change: it's removing the `virtual` specifier (see the api diff for this PR).

PR #4217, opened 2018-06-11
[CarPlay] Adds Xcode 10 Beta 1 Bindings

dalexsoto:
Diff:  https://github.com/xamarin/xamarin-macios/wiki/CarPlay-iOS-xcode10-beta1


src/carplay.cs
```diff
@@ -0,0 +1,719 @@
...
+	interface CPListTemplateDelegate {
+
+		[Abstract]
+		[Export ("listTemplate:didSelectListItem:completionHandler:")]
+		void DidSelectListItem (CPListTemplate listTemplate, CPListItem item, Action completionHandler);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (NSObject))]
+	interface CPManeuver : NSCopying, NSSecureCoding {
```
spouliot: it's not clear to me if this is a user-created object ?
not sure the UI should be adding/modifying maneuvers

spouliot: hmm... there are setters everywhere so I guess it's mutable enough :)

src/carplay.cs
```diff
@@ -0,0 +1,719 @@
...
+		[NullAllowed, Export ("image", ArgumentSemantic.Strong)]
+		UIImage Image { get; set; }
+
+		[NullAllowed, Export ("focusedImage", ArgumentSemantic.Strong)]
+		UIImage FocusedImage { get; set; }
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (CPTemplate))]
+	interface CPMapTemplate : CPBarButtonProviding {
```
spouliot: should not the base .ctor be copied here ?
along with the disable default ctor attribute ?

src/carplay.cs
```diff
@@ -0,0 +1,719 @@
...
+		[Export ("trip", ArgumentSemantic.Strong)]
+		CPTrip Trip { get; }
+
+		[Export ("updateTravelEstimates:forManeuver:")]
+		void UpdateTravelEstimates (CPTravelEstimates estimates, CPManeuver maneuver);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (CPTemplate))]
+	interface CPSearchTemplate {
```
spouliot: same, base ctor ? disable default ctor ?

src/carplay.cs
```diff
@@ -0,0 +1,719 @@
...
+
+		[Abstract]
+		[Export ("sessionConfiguration:limitedUserInterfacesChanged:")]
+		void LimitedUserInterfacesChanged (CPSessionConfiguration sessionConfiguration, CPLimitableUserInterface limitedUserInterfaces);
+	}
+
+	[Abstract]
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (NSObject))]
+	interface CPTemplate : NSSecureCoding {
```
spouliot: seems I read the wrong thing earlier and there's no base ctor to copy :)

src/carplay.cs
```diff
@@ -0,0 +1,719 @@
...
+		[Export ("routeChoices", ArgumentSemantic.Copy)]
+		CPRouteChoice [] RouteChoices { get; }
+
+		[NullAllowed, Export ("userInfo", ArgumentSemantic.Strong)]
+		NSObject UserInfo { get; set; }
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (CPTemplate))]
+	interface CPVoiceControlTemplate {
```
spouliot: still in such case I don't think the default ctor should be included because the `init*` call does not accept null parameters

src/carplay.cs
dalexsoto: Yep, I think this one is fine as is

src/carplay.cs
dalexsoto: @spouliot mmm what base ctor?

```
[Abstract]
[NoWatch, NoTV, NoMac, iOS (12,0)]
interface CPTemplate : NSSecureCoding {
}
```

src/carplay.cs
dalexsoto: Same, I do not think there is other base ctor other than init here

```csharp
[Abstract]
	[NoWatch, NoTV, NoMac, iOS (12,0)]
	interface CPTemplate : NSSecureCoding {

	}
```

src/carplay.cs
dalexsoto: ahh gotcha :)

src/carplay.cs
spouliot: I misread (see later comment) but the default .ctor might not be something to expose in every (or all) subclasses

src/carplay.cs
dalexsoto: ahh dang, I thought I caught them all :pokeball: nice catch!

dalexsoto:
@spouliot Addressed your feedback, thanks for the review 👀

src/carplay.cs
```diff
@@ -0,0 +1,720 @@
+﻿//
+// CarPlay bindings
+//
+// Authors:
+//	Alex Soto  <alexsoto@microsoft.com>
+//
+// Copyright 2018 Xamarin Inc. All rights reserved.
```
mandel-macaque: I've done this before, we are Microsoft Corp now ;)

tools/mtouch/Makefile
```diff
@@ -301,7 +301,9 @@ SIMLAUNCHER_FRAMEWORKS =  \
 						\
 	-weak_framework IdentityLookup			\
 	-weak_framework BusinessChat			\
-	-weak_framework ClassKit
+	-weak_framework ClassKit				\
+						\
+	-weak_framework CarPlay
```
mandel-macaque: Can we add a \ at the end of the list to simplify the next diff?

mandel-macaque:
@dalexsoto Also, might be a good idea to also update the owners document ;)

src/carplay.cs
```diff
@@ -0,0 +1,720 @@
...
+	[NoWatch, NoTV, NoMac, iOS (12,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface CPNavigationAlert : NSSecureCoding {
+
+		// Defined inside CPNavigationAlert.h
+		// static NSTimeInterval const CPNavigationAlertMinimumDuration = 10;
+		[Static]
+		[Wrap ("10D")]
+		double NavigationAlertMinimumDuration { get; }
```
rolfbjarne: Maybe this should be a constant in the code behind instead?

```csharp
partial class CPNavigationAlert {
    public const double MinimumDuration = 10;
}
```

src/carplay.cs
```diff
@@ -0,0 +1,720 @@
...
+
+		[Export ("initWithSummaryVariants:additionalInformationVariants:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string [] summaryVariants, string [] additionalInformationVariants);
+
+		[Export ("summaryVariants", ArgumentSemantic.Copy)]
+		string [] SummaryVariants { get; }
+
+		[NullAllowed, Export ("additionalInformationVariants", ArgumentSemantic.Copy)]
+		string [] AdditionalInformationVariants { get; }
```
rolfbjarne: This is `NullAllowed`, but the corresponding parameter in the constructor isn't? (and the same for `SummaryVariants`)

src/carplay.cs
```diff
@@ -0,0 +1,720 @@
...
+	interface CPVoiceControlTemplate {
+
+		[Export ("initWithTitleVariants:animatedImage:")]
+		IntPtr Constructor (string [] titleVariants, UIImage animatedImage);
+
+		[NullAllowed, Export ("titleVariants", ArgumentSemantic.Copy)]
+		string [] TitleVariants { get; }
+
+		[NullAllowed, Export ("animatedImage", ArgumentSemantic.Strong)]
+		UIImage AnimatedImage { get; }
```
rolfbjarne: Same: `NullAllowed` doesn't match the corresponding parameter in the constructor.

src/carplay.cs
spouliot: good point, that's closer to the original definition

src/carplay.cs
dalexsoto: yep, agreed, according to docs this should not be nil since it is a requirement in the ctor

src/carplay.cs
dalexsoto: yep, agreed, according to docs this should not be nil since it is a requirement in the ctor

VincentDondain:
Was reviewing this while it was merged, bindings look good to me 👍

PR #4231, opened 2018-06-12
[msbuild] Fix detecting already signed executables. (#3743)

spouliot:
**Note: backport from master / d15-8**

The MSBuild tasks will codesign an executable if the executable's timestamp is
later than `_CodeSignature/CodeResources`'s timestamp (or if
`_CodeSignature/CodeResources` doesn't exist).

Unfortunately, the codesign executable modifies both of those files, and the
executable last. This means that even just after running codesign, the
executable's timestamp might be later than `_CodeSignature/CodeResources`'s
timestamp (due to HFS+'s one-second timestamp resolution, this might happen
all within the same second, which means that this is a random issue: the
problem only occurs if the executable was modified at least a second later
than `_CodeSignature/CodeResources`.)

So make sure to touch `_CodeSignature/CodeResources` after running codesign,
so that the next time a build occurs (with no modifications), we don't resign
needlessly.

Fixes this (random) test failure when running the MSBuild tests:

    1) Test Failure : Xamarin.iOS.Tasks.TargetTests.RebuildExecutable_NoModifications
         #1: ../MySingleView/bin/iPhoneSimulator/Debug/MySingleView.app/MySingleView
      Expected: 2017-11-30 10:04:20.000
      But was:  2017-11-30 10:04:22.000

Fixes https://github.com/xamarin/maccore/issues/592.


jstedfast:
Awesome catch, thanks!

PR #4234, opened 2018-06-12
[authenticationservices] New iOS-only framework added in beta 1

spouliot:


src/authenticationservices.cs
```diff
@@ -0,0 +1,195 @@
...
+		[Export ("sharedStore")]
+		ASCredentialIdentityStore SharedStore { get; }
+
+		[Async]
+		[Export ("getCredentialIdentityStoreStateWithCompletion:")]
+		void GetCredentialIdentityStoreState (Action<ASCredentialIdentityStoreState> completion);
+
+		[Async]
+		[Export ("saveCredentialIdentities:completion:")]
+		void SaveCredentialIdentities (ASPasswordCredentialIdentity[] credentialIdentities, [NullAllowed] Action<bool, NSError> completion);
```
dalexsoto: Minor:

I am not sure I like `Action<bool, NSError>` here, what does that `bool` means? also the same delegate can be shared below I guess

src/authenticationservices.cs
```diff
@@ -0,0 +1,195 @@
...
+		[Export ("supportsIncrementalUpdates")]
+		bool SupportsIncrementalUpdates { get; }
+	}
+
+	[iOS (12,0)]
+	[BaseType (typeof (NSExtensionContext))]
+	[DisableDefaultCtor]
+	interface ASCredentialProviderExtensionContext {
+		[Export ("completeRequestWithSelectedCredential:completionHandler:")]
+		void CompleteRequest (ASPasswordCredential credential, [NullAllowed] Action<bool> completionHandler);
```
dalexsoto: Same, I would use a named delegate here for `Action<bool>`, what does that `bool` means?

spouliot:
failure due to https://github.com/xamarin/xamarin-macios/issues/4241 which seems to happen randomly :|

PR #4236, opened 2018-06-12
[NaturalLanguage] Added the new framework for Xcode10.

mandel-macaque:


src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+				}
+				return result;
+			}
+			set {
+				var rKeys = Array.ConvertAll<NLLanguage, NSString> (value.Keys.ToArray (), e => NLLanguageExtensions.GetConstant (e));
+				NativeLanguageHints = NSDictionary<NSString, NSNumber>.FromObjectsAndKeys (value.Values.ToArray (), rKeys, rKeys.Length);
+			}
+		}
+	}
+}
```
mandel-macaque: Not sure about this part, do we really want to add the Dictionary or just use the NSDictionary<NSString, NSNumer>. NSDictionary<K, V> does not lake enums as keys.

src/NaturalLanguage/Enums.cs
```diff
@@ -0,0 +1,270 @@
...
+		OmitPunctuation = 1uL << 1,
+		OmitWhitespace = 1uL << 2,
+		OmitOther = 1uL << 3,
+		JoinNames = 1uL << 4,
+		JoinContractions = 1uL << 5,
+	}
+
+	[Native]
+	[iOS (12,0), Mac (10,14, onlyOn64: true), TV (12,0)]
+	public enum NLModelType : ulong {
```
spouliot: in general when it's not a `[Flag]` it should be `long`

src/NaturalLanguage/Enums.cs
```diff
@@ -0,0 +1,270 @@
...
+		Classifier,
+		Sequence,
+	}
+
+	[Native]
+	[iOS (12,0), Mac (10,14, onlyOn64: true), TV (12,0)]
+	public enum NLTokenUnit : ulong {
```
spouliot: n general when it's not a `[Flag]` it should be `long`

src/NaturalLanguage/Enums.cs
```diff
@@ -0,0 +1,270 @@
...
+		[Field ("NLLanguageThai")]
+		Thai,
+		[Field ("NLLanguageTibetan")]
+		Tibetan,
+		[Field ("NLLanguageTraditionalChinese")]
+		TraditionalChinese,
+		[Field ("NLLanguageTurkish")]
+		Turkish,
+		[Field ("NLLanguageUkrainian")]
+		NLLanguageUkrainian,
```
spouliot: -> `Ukrainian,`

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+				lang = NLLanguageExtensions.GetValue (nslang);
+			NSString.ReleaseNative (nsstring);
+			nslang.Dispose ();
+			return lang;
+		}
+	
+		public Dictionary<NLLanguage, NSNumber> GetLanguageHypotheses (nuint maxHypotheses)
+		{
+			using (var hypo = GetNativeLanguageHypotheses (maxHypotheses))
+			{
```
spouliot: minor: `{` on the same line as `using`

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+			var nslang = _GetDominantLanguage (nsstring);
+			var lang =  NLLanguage.Undetermined;
+			if (nslang.Handle != IntPtr.Zero)
+				lang = NLLanguageExtensions.GetValue (nslang);
+			NSString.ReleaseNative (nsstring);
+			nslang.Dispose ();
+			return lang;
+		}
+	
+		public Dictionary<NLLanguage, NSNumber> GetLanguageHypotheses (nuint maxHypotheses)
```
spouliot: can we avoid `NSNumber` ? i.e. what's the real (value)type being returned ?

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+			{
+				var result = new Dictionary<NLLanguage, NSNumber> ();
+				foreach (var k in hypo.Keys) {
+					result[NLLanguageExtensions.GetValue (k)] = hypo[k];
+				}
+				return result;
+			}
+		}
+
+		public Dictionary<NLLanguage, NSNumber> LanguageHints
```
spouliot: same, try to avoid `NSNumber`
it's only deferring the _"querying the right type"_ to the caller
and it means references instead of value types

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+			NSString.ReleaseNative (nsstring);
+			nslang.Dispose ();
+			return lang;
+		}
+	
+		public Dictionary<NLLanguage, NSNumber> GetLanguageHypotheses (nuint maxHypotheses)
+		{
+			using (var hypo = GetNativeLanguageHypotheses (maxHypotheses))
+			{
+				var result = new Dictionary<NLLanguage, NSNumber> ();
```
spouliot: you already know the size of the dictionary so give it to the .ctor to avoid resizing operations

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+					result[NLLanguageExtensions.GetValue (k)] = hypo[k];
+				}
+				return result;
+			}
+		}
+
+		public Dictionary<NLLanguage, NSNumber> LanguageHints
+		{
+			get {
+				var result = new Dictionary<NLLanguage, NSNumber> ();
```
spouliot: same, size is known

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+				foreach (var k in NativeLanguageHints.Keys) {
+					result[NLLanguageExtensions.GetValue (k)] = NativeLanguageHints[k];
+				}
+				return result;
+			}
+			set {
+				var rKeys = Array.ConvertAll<NLLanguage, NSString> (value.Keys.ToArray (), e => NLLanguageExtensions.GetConstant (e));
```
spouliot: let's avoid `ToArray`
and just loop to avoid the extra allocation(s)

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+				NativeLanguageHints = NSDictionary<NSString, NSNumber>.FromObjectsAndKeys (value.Values.ToArray (), rKeys, rKeys.Length);
```
spouliot: same (ToArray) and it might be better to use a MutableDictionary and initialize it with the (known) size

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+	[BaseType (typeof(NSObject))]
+	interface NLModelConfiguration : NSCopying, NSSecureCoding {
+		[Export ("type")]
+		NLModelType Type { get; }
+
+		[Internal]
+		[NullAllowed, Export ("language")]
+		NSString _Language { get; }
+
+		NLLanguage Language { [Wrap ("(_Language.Handle != IntPtr.Zero)? NLLanguageExtensions.GetValue (_Language) : NLLanguage.Undetermined")] get; }
```
spouliot: what if `_Language` is `null` ? in that case an NRE will happen
there's normally no instance with an `Handle` set of `IntPtr.Zero` since using it would generally crash

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+
+		[Export ("unit")]
+		NLTokenUnit Unit { get; }
+
+		[NullAllowed, Export ("string", ArgumentSemantic.Retain)]
+		string String { get; set; }
+
+		[Internal]
+		[Export ("setLanguage:")]
+		void _SetLanguage (string language);
```
spouliot: use `NSString` here, otherwise you'll cause extra conversions as `GetConstant` returns an `NSString` instance

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+
+		[Export ("tokenRangeAtIndex:")]
+		NSRange GetTokenRange (nuint characterIndex);
+
+		[Export ("tokensForRange:")]
+		NSValue[] GetTokens (NSRange range);
+
+		[Async (ResultTypeName="NLTokenizerEnumerateContinuation")]
+		[Export ("enumerateTokensInRange:usingBlock:")]
+		void EnumerateTokensInRange (NSRange range, Action<NSRange, NLTokenizerAttributes, bool> block);
```
spouliot: `EnumerateTokens` -> the range is the first parameter

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+		[NullAllowed, Export ("dominantLanguage")]
+		NSString _DominantLanguage { get; }
+
+		[Wrap ("NLLanguageExtensions.GetValue (_DominantLanguage)")]
+		NLLanguage DominantLanguage { get; }
+
+		[Async (ResultTypeName="NLTaggerEnumerateTagsContinuation")]
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		[Export ("enumerateTagsInRange:unit:scheme:options:usingBlock:")]
+		void EnumerateTags (NSRange range, NLTokenUnit unit, NSString scheme, NLTaggerOptions options, Action<NSString, NSRange, bool> block);
```
spouliot: `Action<NSString, NSRange, bool>` is too long (one is good, two if the 2nd is `NSError`)
Please use a custom delegate where the parameter names will act as self-documention

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+
+		[Wrap ("EnumerateTags (range, unit, scheme.GetConstant (), options, block)")]
+		void EnumerateTags (NSRange range, NLTokenUnit unit, NLTagScheme scheme, NLTaggerOptions options, Action<NSString, NSRange, bool> block);
```
spouliot: same

src/NaturalLanguage/NLLanguageRecognizer.cs
dalexsoto: Add a whitespace at the end

src/naturallanguage.cs
dalexsoto: Also use a named delegate here, and check if that bool is an out parameter, usually those Enumerate methods uses the bool as a stop flag.

dalexsoto: And if it is the case that the bool is an out flag then remove the Async call

src/naturallanguage.cs
dalexsoto: Also remove `block` as parameter name, use `handler` or something alike. That said check if that bool isn't as a stop flag (i.e. `out bool`) and if so remove the Async here

src/naturallanguage.cs
dalexsoto: same as above

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+		[Export ("setLanguage:range:")]
+		void SetLanguage (NSString language, NSRange range);
+
+		[Wrap ("SetLanguage (language.GetConstant (), range)")]
+		void SetLanguage (NLLanguage language, NSRange range);
+
+		[Export ("setOrthography:range:")]
+		void SetOrthography (NSOrthography orthography, NSRange range);
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
```
dalexsoto: `[EditorBrowsable (EditorBrowsableState.Advanced)]` is default in VS4Mac so maybe `[Protected]` would work better out of the box

src/naturallanguage.cs
```diff
@@ -0,0 +1,246 @@
...
+		[Export ("setModels:forTagScheme:")]
+		void SetModels (NLModel[] models, NSString tagScheme);
+
+		[Wrap ("SetModels (models, tagScheme.GetConstant ())")]
+		void SetModels (NLModel[] models, NLTagScheme tagScheme);
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
```
dalexsoto: Same

src/naturallanguage.cs
dalexsoto: Also `EnumerateTokensInRange ` => `EnumerateTokens`

src/NaturalLanguage/NLLanguageRecognizer.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace NaturalLanguage {
+
+	public partial class NLLanguageRecognizer {
+
+		public static NLLanguage GetDominantLanguage (string @string)
+		{
+			var nsstring = NSString.CreateNative (@string);
+			var nslang = _GetDominantLanguage (nsstring);
+			var lang =  NLLanguage.Undetermined;
+			if (nslang.Handle != IntPtr.Zero)
```
rolfbjarne: If the dominant language is a null handle, `nslang` would be null, and this condition should be `nslang != null`.

Manually bound API should also have tests.

src/naturallanguage.cs
mandel-macaque: It is on VS4Mac, but on windows? Protected would mean that the if the user needs the method it will need to subclass. I want to expose the method in case we ever miss an enum value, so that the user can pass an NSString. I would personally leave it like it is there.

src/naturallanguage.cs
mandel-macaque: Same reply ;)

mandel-macaque:
I have created an issue for the unrelated failing tests: https://github.com/xamarin/xamarin-macios/issues/4241

src/naturallanguage.cs
```diff
@@ -0,0 +1,248 @@
...
+
+		[Export ("predictedLabelForString:")]
+		[return: NullAllowed]
+		string GetPredictedLabel (string @string);
+
+		[Export ("predictedLabelsForTokens:")]
+		string[] GetPredictedLabels (string[] tokens);
+	}
+
+	public delegate void NLTokenizerEnumerateContinuationHandler (NSRange tokenRange, NLTokenizerAttributes flags, out bool stop);
```
dalexsoto: Remove the `public` modifier, this could cause issues in the future if for some reason this is used from another APIDefinition. As a rule always keep the same modifier as in interfaces, the generator will take care of making them public.

src/naturallanguage.cs
```diff
@@ -0,0 +1,248 @@
...
+		NSRange GetTokenRange (nuint characterIndex);
+
+		[Export ("tokensForRange:")]
+		NSValue[] GetTokens (NSRange range);
+
+		[Export ("enumerateTokensInRange:usingBlock:")]
+		void EnumerateTokens (NSRange range, NLTokenizerEnumerateContinuationHandler handler);
+	}
+
+	public delegate void NLTaggerEnumerateTagsContinuationHandler (NSString tag, NSRange tokenRange, out bool stop);
```
dalexsoto: same.

mandel-macaque:
@dalexsoto addressed your comments, please take a look at I'll land this.

PR #4224, opened 2018-06-12
Cherry-pick changes from master to implement support for building on internal Jenkins.

rolfbjarne:
* 9cdaf2b75 [jenkins] Cherry-pick a series of fixes for internal jenkins support from the d15-8 branch (#4195)
* ae7c72cba Add support for building on Jenkins. (#4159)
* fa7f52dab [jenkins] Improve api/generator diff reporting to say if there were changes, and if they were breaking or not. (#4138)
* f7527aaba [jenkins] Revert unintended change that prevents the keychain from unlocking properly. (#4158)
* e0e2f9317 [jenkins] Fix scripts to be shellcheck-happy. (#4148)
* 96d3445cc [tests] Make sure accessing the keychain doesn't make macOS show any dialogs. (#4141)
* bece50d6b [jenkins] Don't treat API/Generator diff failures as errors. (#4127)
* c35adef14 [jenkins] Update the urls for html reports. (#4119)
* ccb2f59ca [jenkins] Make it possible to skip API comparison by applying a label to a pull request. (#3994)
* d7efd346e [tools] Improve error reporting when trying to create API/generator diff for a pull request with conflicts. (#3966)
* fb322c4d1 [jenkins] Clean keystore on bots before running tests (#3754)
* 61ec52009 Bump maccore.

PR #4240, opened 2018-06-13
[jenkins] Use different emojiis depending on the result of the api/generator diffs.

rolfbjarne:
Makes it harder to accidentially skip reviewing changes.


rolfbjarne:
@spouliot 

> note: it might have been fixed by backports but I noticed that (at least xcode10 branch) links do not point to the reports (only jenkins)

This will be fixed in PR #4224.

PR #4243, opened 2018-06-13
[Intents] Adds Xcode 10 Beta 1 bindings

dalexsoto:


src/intents.cs
```diff
@@ -396,7 +414,17 @@ public enum INPhotoAttributeOptions : ulong {
...
+		[Watch (5,0), iOS (12,0)]
+		PortraitPhoto = (1uL << 24),
+		[Watch (5,0), iOS (12,0)]
+		LivePhoto = (1uL << 25),
+		[Watch (5,0), iOS (12,0)]
+		LoopPhoto = (1uL << 26),
+		[Watch (5,0), iOS (12,0)]
+		BouncePhoto = (1uL << 27),
+		[Watch (5,0), iOS (12,0)]
+		LongExposurePhoto = (1uL << 28)
```
mandel-macaque: Following the change, add an extra ',' at the end to simplify the next diff so that it does not happen the same as in this one.

src/intents.cs
```diff
@@ -8528,5 +8702,299 @@ interface INVisualCodeTypeResolutionResult {
...
+
+		[Export ("initWithCode:userActivity:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (INPlayMediaIntentResponseCode code, [NullAllowed] NSUserActivity userActivity);
+
+		[Export ("code")]
+		INPlayMediaIntentResponseCode Code { get; }
+
+		[NullAllowed, Export ("nowPlayingInfo", ArgumentSemantic.Copy)]
+		NSDictionary WeakNowPlayingInfo { get; set; }
```
mandel-macaque: Could we add a strong dictionary here?

src/intents.cs
dalexsoto: ahh missed this one, thanks

src/intents.cs
dalexsoto: Yes, it's on my TODOs to figure out the keys that are available here, the comment is a bit vague and I do not want to add APIs that should not be there

mandel-macaque: Good to me, but lets add an issue so that we do not forget :)

src/intents.cs
```diff
@@ -4538,11 +4692,18 @@ interface INSearchCallHistoryIntentResponse {
 	[BaseType (typeof (INIntent))]
 	interface INSearchForMessagesIntent {
 
+		[Deprecated (PlatformName.MacOSX, 10,14, message: "Use the '[DesignatedInitializer]' overload instead")]
+		[Deprecated (PlatformName.WatchOS, 5,0, message: "Use the '[DesignatedInitializer]' overload instead")]
+		[Deprecated (PlatformName.iOS, 12,0, message: "Use the '[DesignatedInitializer]' overload instead")]
```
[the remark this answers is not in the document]
dalexsoto: Interesting, yeah I kinda remember that was the case funny that bots are also happy.

src/intents.cs
```diff
@@ -8528,5 +8702,299 @@ interface INVisualCodeTypeResolutionResult {
...
+	[Watch (5,0), NoMac, iOS (12,0)]
+	[Protocol]
+	interface INPlayMediaIntentHandling {
+
+		[Abstract]
+		[Export ("handlePlayMedia:completion:")]
+		void HandlePlayMedia (INPlayMediaIntent intent, Action<INPlayMediaIntentResponse> completion);
+
+		[Export ("confirmPlayMedia:completion:")]
+		void Confirm (INPlayMediaIntent intent, Action<INPlayMediaIntentResponse> completion);
```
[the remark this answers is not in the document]
dalexsoto: Yeah, it is `Confirm` because of this https://bugzilla.xamarin.com/show_bug.cgi?id=59164 and `HandlePlayMedia` cannot be just `Handle` because it will clash with our special property.

src/intents.cs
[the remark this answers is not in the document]
dalexsoto: Yes we can and I am working on it :) that said I do not think this should stop the PR from merging 😁

dalexsoto:
@spouliot thanks for your feedback, already addressed it, I'll have the strong dictionary with the IntentsUI PR :)

src/intents.cs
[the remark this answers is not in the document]
spouliot: @VincentDondain any idea why bots did not report it, e.g. missing case in tests ? ^

PR #4246, opened 2018-06-13
[CoreNFC] Update bindings to Xcode 10 beta 1

dalexsoto:


src/corenfc.cs
```diff
@@ -296,5 +296,14 @@ interface NFCTagCommandConfiguration : NSCopying {
 		[Export ("retryInterval")]
 		double RetryInterval { get; set; }
 	}
+
+	[iOS (12,0)]
+	[Category]
+	[BaseType (typeof (NSUserActivity))]
+	interface NSUserActivityCoreNFC {
```
spouliot: The common naming we used so far is like `NSUserActivity_MKMapItem` and `NSUserActivity_IntentsAdditions`, so the original `NSUserActivity_CoreNFC` would be more consistent (even if I dislike underscores, good thing it does not show up in sources).

Also I'm not sure if we should use `Nfc` since it's the suffix, not the prefix. OTOH it's the same as the prefix and fx name... so I guess NFC makes sense.

dalexsoto: Fixed thanks, Yeah I was also inclined to do Nfc but got hit by the same opinion in the end, that it makes sense as is.

Issue #4241, opened 2018-06-13
Xamarin.Linker.SdkTest.NoAssemblyReferenceInAttributes fails with monotouch 


mandel-macaque:
Using the [natural language PR](https://github.com/xamarin/xamarin-macios/pull/4236) as an example. 

The tests ran fine on [initially](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/7342//Test_Report/) but started failing on [later builds](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/7346//Test_Report/). 

Beware that the failing tests (in the natural lang API) where failing because 32 devices were not running the latests iOS and those failures can be considered noise.

rolfbjarne:
This was fixed in the xcode10 branch here: https://github.com/xamarin/xamarin-macios/commit/ddd7d4cc19b2c3429c116d87f62cbb61c32b38a5.

PR #4247, opened 2018-06-13
[CoreMotion] Update bindings to Xcode 10 Beta 1

dalexsoto:


tests/xtro-sharpie/common-CoreMotion.ignore
```diff
@@ -2,3 +2,10 @@
 
 ## no generator support for FastEnumeration - https://bugzilla.xamarin.com/show_bug.cgi?id=4391
 !missing-protocol-conformance! CMSensorDataList should conform to NSFastEnumeration
+
+## For some reason sharpie is not picking up the following classes
```
spouliot: double check the umbrella header, it's likely just missing the new header file
if that's the case file a rdar - if not let's figure out why and file the rdar next ;-)

dalexsoto: Oh wow @spouliot, nice catch, `CMMovementDisorderManager.h` is indeed missing from the umbrella header. TIL ❤️

dalexsoto:
Failure due to https://github.com/xamarin/xamarin-macios/issues/4241

PR #4251, opened 2018-06-14
[Network, Security] Bindings for Network, Security and some classes needed to port samples

migueldeicaza:
Test build


src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+
+		public virtual void Dispose (bool disposing)
+		{
+			if (handle != IntPtr.Zero) {
+				CFObject.CFRelease (handle);
+				handle = IntPtr.Zero;
+			}
+		}
+
+		[Mac(10,14)][iOS(12,0)][Watch(5,0)][TV(12,0)]
```
spouliot: Sadly the framework is not available on watchOS so all `[Watch(5,0)]` should be removed.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+		public string BonjourServiceName => nw_endpoint_get_bonjour_service_name (handle);
+
+
+		
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern string nw_endpoint_get_bonjour_service_type (OS_nw_endpoint endpoint);
+
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		public string BonjourSericeType => nw_endpoint_get_bonjour_service_type (handle);
```
spouliot: typo `Serice`

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+		[DllImport (Constants.NetworkLibrary)]
+		static extern string nw_endpoint_copy_address_string (OS_nw_endpoint endpoint);
+		
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		public string Address => nw_endpoint_copy_address_string (handle);
+		
+		// extern const struct sockaddr * _Nonnull nw_endpoint_get_address (nw_endpoint_t _Nonnull endpoint) __attribute__((availability(tvos, introduced=12.0))) __attribute__((availability(watchos, introduced=5.0))) __attribute__((availability(ios, introduced=12.0))) __attribute__((availability(macos, introduced=10.14)));
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern IntPtr /* struct sockaddr* */ nw_endpoint_get_address (OS_nw_endpoint endpoint);
```
rolfbjarne: This isn't used/exposed, in which case it should probably be commented out.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe OS_nw_endpoint nw_endpoint_create_bonjour_service (string name, string type, string domain);
+
+		public NWEndpoint CreateBonjourService (string name, string serviceType, string domain)
+		{
+			if (serviceType == null)
+				throw new ArgumentNullException (nameof (serviceType));
+			var x = nw_endpoint_create_bonjour_service (name, serviceType, domain);
+			return new NWEndpoint (x, owns: true);
```
rolfbjarne: `nw_endpoint_create_bonjour_service` can return NULL.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+		}
+		
+	
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe string nw_endpoint_get_bonjour_service_name (OS_nw_endpoint endpoint);
```
rolfbjarne: This P/Invoke will free the string returned from `nw_endpoint_get_bonjour_service_name`, but it's defined as `const char *` in the header, so it shouldn't be freed.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		public string BonjourServiceName => nw_endpoint_get_bonjour_service_name (handle);
+
+
+		
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern string nw_endpoint_get_bonjour_service_type (OS_nw_endpoint endpoint);
```
rolfbjarne: Same: this P/Invoke will free the returned string.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,163 @@
...
+
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		public string BonjourSericeType => nw_endpoint_get_bonjour_service_type (handle);
+
+
+		
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern string nw_endpoint_get_bonjour_service_domain (OS_nw_endpoint endpoint);
```
rolfbjarne: Same: this P/Invoke will free the returned string.

src/Network/NWParameters.cs
```diff
@@ -0,0 +1,148 @@
...
+
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		public bool LocalOnly {
+			get => nw_parameters_get_local_only (handle);
+			set => nw_parameters_set_local_only (handle, value);
+		}
+	
+		[Watch (5,0), TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_parameters_set_prefer_no_proxy (nw_parameters_t parameters, bool prefer_no_proxy);
```
rolfbjarne: I think all bools in P/Invokes need a `MarshalAs(UnmanagedType.I1)` attribute, since the native `bool` type is 1 byte, and by default they're 4 bytes in P/Invokes.

src/Network/NWEndpoint.cs
migueldeicaza: That is a work in progress

src/Network/NWEndpoint.cs
migueldeicaza: It is indeed, the header files show that everywhere :-(

src/Network/NWEndpoint.cs
migueldeicaza: good catch!

migueldeicaza:
Status: https://gist.github.com/migueldeicaza/08491ba75fe3037f6a54c01ae6788c12

migueldeicaza:
Work-in-progress netcat port:

https://gist.github.com/migueldeicaza/100e944240e5264d90c6a5452de97b74

migueldeicaza:
NetCat sample is now here, what i use to try this out and explore the API:

https://github.com/migueldeicaza/NetCatNetwork

rolfbjarne:
build

src/Network/NWContentContext.cs
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+
+		// This sending context represents the entire connection
+		// ie: NW_CONNECTION_DEFAULT_STREAM_CONTEXT
+		static NWContentContext defaultStream;
+		public static NWContentContext DefaultStream {
+			get {
+				if (defaultStream == null)
+					defaultStream = MakeGlobal (Marshal.ReadIntPtr (Dlfcn.dlsym (Libraries.Network.Handle, "_nw_content_context_default_stream")));
```
rolfbjarne: Why do you need to make these objects global when they're stored in static variables anyway?

src/Security/SecCertificate2.cs
```diff
@@ -0,0 +1,46 @@
...
+using Foundation;
+using CoreFoundation;
+
+namespace Security {
+	
+	[TV (12,0), Mac (10,14), iOS (12,0)]
+	public class SecCertificate2 : NativeObject {
+		public SecCertificate2 (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
spouliot: not needed in members (when identical to the type-level attributes)

src/Security/SecIdentity2.cs
```diff
@@ -0,0 +1,84 @@
...
+		[DllImport (Constants.SecurityLibrary)]
+		extern static IntPtr sec_identity_create (IntPtr secidentityHandle);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public SecIdentity2 (SecIdentity identity)
+		{
+			if (identity == null)
+				throw new ArgumentNullException (nameof (identity));
+			
+			handle = sec_identity_create (identity == null ? IntPtr.Zero : identity.Handle);
```
spouliot: That API can return `null` so `InitializeHandle` (from base `NativeObject`) should be called to avoid creating an invalid instance (that will crash/throw later).

src/Security/SecIdentity2.cs
```diff
@@ -0,0 +1,84 @@
...
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public SecIdentity2 (SecIdentity identity, SecCertificate [] certificates)
+		{
+			if (identity == null)
+				throw new ArgumentNullException (nameof (identity));
+			if (certificates == null)
+				throw new ArgumentNullException (nameof (certificates));
+			using (var nsarray = NSArray.FromObjects (certificates)){
+				handle = sec_identity_create_with_certificates (identity.Handle, nsarray.Handle);
```
spouliot: same -> `InitializeHandle`

src/Security/SecIdentity2.cs
```diff
@@ -0,0 +1,84 @@
...
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public SecCertificate [] Certificates {
+			get {
+				var certArray = sec_identity_copy_certificates_ref (GetHandle ());
+				var n = (int) NSArray.GetCount (certArray);
+				var ret = new SecCertificate [n];
+				for (int i = 0; i < n; i++){
+					ret [i] = new SecCertificate (NSArray.GetAtIndex (certArray, (nuint) i), owns: false);
+				}
+				return ret;
```
spouliot: `certArray` should be disposed before returning

src/Security/SecProtocolMetadata.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Security {
+	
+	public class SecProtocolMetadata : NativeObject {
+		internal SecProtocolMetadata (IntPtr handle) : base (handle, false) {}
+
+		// This type is only ever surfaced in response to callbacks in TLS/Network and documented as read-only
+		// if this ever changes, make this public
+		internal SecProtocolMetadata (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
spouliot: same, move to type-level

src/Security/SecProtocolOptions.cs
```diff
@@ -0,0 +1,203 @@
...
+                                block_handler.SetupBlockUnsafe (Trampolines.SDSecProtocolKeyUpdate.Handler, keyUpdate);
+
+                                sec_protocol_options_set_key_update_block (handle, (IntPtr)((void*) block_ptr_handler), keyUpdateQueue.Handle);
+                                block_ptr_handler->CleanupBlock ();
+			}
+		}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.SecurityLibrary)]
+		static extern void sec_protocol_options_set_challenge_block(sec_protocol_options_t options, IntPtr challenge_block, dispatch_queue_t challenge_queue);
```
spouliot: better comment this (and the next one), since they are not publicly exposed. That way xtro can report them as missing/ignored

src/Security/SecTrust2.cs
```diff
@@ -0,0 +1,48 @@
...
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.SecurityLibrary)]
+		extern static IntPtr sec_trust_create (IntPtr sectrustHandle);
+		
+		public SecTrust2 (SecTrust trust)
+		{
+			if (trust == null)
+				throw new ArgumentNullException (nameof (trust));
+			
+			handle = sec_trust_create (trust == null ? IntPtr.Zero : trust.Handle);
```
spouliot: can return `null`, should use `InitializeHandle`

src/generator-attributes.cs
```diff
@@ -856,3 +856,14 @@ public DefaultEnumValueAttribute ()
...
 	}
 }
+
+//
+// This prevents the generator from generating the managed proxy to
+// the method being called, this is done when we are interested in
+// the side effects of the binding of the method, rather than the
+// method itself -- for security for example, we generate the
+// proxy code for invoking blocks and callbacks to blocks
+// see
```
spouliot: incomplete comment ?

tests/introspection/ApiCMAttachmentTest.cs
```diff
@@ -416,9 +425,45 @@ protected INativeObject GetINativeInstance (Type t)
 					new CVPixelBufferPoolSettings (),
 					new CVPixelBufferAttributes (CVPixelFormatType.CV24RGB, 100, 50)
 				);
+			case "NWAdvertiseDescriptor":
+				return NWAdvertiseDescriptor.CreateBonjourService ("sampleName" + DateTime.Now);
```
spouliot: The NRE in the tests is because `GetINativeInstance` returns `null` which is likely because of the extra entries being added returns `null`. It could be this one

tests/introspection/ApiCMAttachmentTest.cs
```diff
@@ -416,9 +425,45 @@ protected INativeObject GetINativeInstance (Type t)
...
+			case "NWConnection": {
+				var endpoint = NWEndpoint.Create ("www.microsoft.com", "https");
+				var parameters = NWParameters.CreateTcp (configureTcp: null);
+				return new NWConnection (endpoint, parameters);
+			}
+			case "NWContentContext":
+				return new NWContentContext ("contentContext" + DateTime.Now);
+			case "NWEndpoint":
+				return NWEndpoint.Create ("www.microsoft.com", "https");
```
spouliot: this one

tests/introspection/ApiCMAttachmentTest.cs
```diff
@@ -416,9 +425,45 @@ protected INativeObject GetINativeInstance (Type t)
...
+			case "NWListener":
+				return NWListener.Create (NWParameters.CreateTcp (configureTcp: null));
+			case "NWParameters":
+				return NWParameters.CreateTcp (configureTcp: null);
+			case "NWProtocolDefinition":
+				// Makes a new instance every time
+				return NWProtocolDefinition.TcpDefinition;
+			case "NWProtocolOptions":
+				return NWProtocolOptions.CreateTcp ();
```
spouliot: or one of the above...

src/Network/NWContentContext.cs
migueldeicaza: To prevent someone calling `Dispose()` and ruining the global instance for everyone else.

tools/common/Frameworks.cs
```diff
@@ -410,6 +412,7 @@ public static Frameworks GetwatchOSFrameworks (Application app)
 					{ "Vision", "Vision", 11 },
 
 					{ "NaturalLanguage", "NaturalLanguage", 12,0 },
+					{ "Network", "Network", 14, 0 } ,
```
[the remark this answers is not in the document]
migueldeicaza: Got it

src/CoreFoundation/DispatchData.cs
```diff
@@ -0,0 +1,134 @@
...
+
+		public static DispatchData FromByteBuffer (byte [] buffer, int start, int length)
+		{
+			if (buffer == null)
+				throw new ArgumentNullException (nameof (buffer));
+			if (start < 0 || start >= buffer.Length)
+				throw new ArgumentException (nameof (start));
+			if (length < 0)
+				throw new ArgumentException (nameof (length));
+			if (start+length >= buffer.Length)
```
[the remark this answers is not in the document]
migueldeicaza: Replaced with:

```
			if (start > buffer.Length + length)
```

src/CoreFoundation/DispatchIO.cs
```diff
@@ -0,0 +1,117 @@
...
+// DispatchIO.cs: IO routines for Dispatch
+//
+// Authors:
+//   Miguel de Icaza (miguel@gnome.org)
+//   Marek Safar (marek.safar@gmail.com)
+//
+// Copyright 2010 Novell, Inc.
+// Copyright 2011-2014 Xamarin Inc
+//
+// this contains DispathcObject, Group and Queue
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWAdvertiseDescriptor.cs
```diff
@@ -0,0 +1,66 @@
...
+using nw_advertise_descriptor_t=System.IntPtr;
+using OS_nw_advertise_descriptor=System.IntPtr;
+
+namespace Network {
+	
+	public class NWAdvertiseDescriptor : NativeObject {
+		public NWAdvertiseDescriptor (IntPtr handle, bool owns) : base (handle, owns)
+		{ }
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWAdvertiseDescriptor.cs
```diff
@@ -0,0 +1,66 @@
...
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_advertise_descriptor_set_txt_record (IntPtr handle, string txtRecord, IntPtr txtLen);
+
+		public void SetTxtRecord (string txt)
+		{
+			if (txt == null)
+				throw new ArgumentNullException (nameof (txt));
+			var n = System.Text.Encoding.UTF8.GetByteCount (txt);
```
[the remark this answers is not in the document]
migueldeicaza: Because Mono marshals `System.String` using UTF8 encoding, and this API needs the number of bytes being passed.

src/Network/NWConnection.cs
```diff
@@ -0,0 +1,574 @@
...
+	//
+	// Signature for a method invoked on data received, same as NWConnectionReceiveCompletion,
+	// but they receive DispatchData instead of data + dataSize
+	//
+	public delegate void NWConnectionReceiveDispatchDataCompletion (DispatchData data, NWContentContext context, bool isComplete, NWError error);
+	
+	public class NWConnection : NativeObject {
+		public NWConnection (IntPtr handle, bool owns) : base (handle, owns) {} 
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+		Invalid = 0,
+		Address = 1,
+		Host = 2,
+		BonjourService = 3
+	}
+	
+	public class NWEndpoint : NativeObject {
+		public NWEndpoint (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[Mac(10,14)][iOS(12,0)][TV(12,0)]
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+				return null;
+			return new NWEndpoint (handle, owns: true);
+		}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern IntPtr nw_endpoint_get_hostname (OS_nw_endpoint endpoint);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public string Hostname => Marshal.PtrToStringAnsi (nw_endpoint_get_hostname (GetHandle()));
```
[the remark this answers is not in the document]
migueldeicaza: This one returns a pointer to an internal string, the other one returns a copy.

PInvoke with a string signature releases return values.

src/Network/NWEndpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+		static extern string nw_endpoint_copy_address_string (OS_nw_endpoint endpoint);
+		
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public string Address => nw_endpoint_copy_address_string (GetHandle());
+		
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		static extern IntPtr /* struct sockaddr* */ nw_endpoint_get_address (OS_nw_endpoint endpoint);
+
+		// TODO: same
```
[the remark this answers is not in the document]
migueldeicaza: We want to, but we need to have a discussion about what to do and how to surface it.   This is going to be needed, but we can not use the .NET data type, we are going to need a new data type.

src/Network/NWInterface.cs
```diff
@@ -0,0 +1,49 @@
...
+using CoreFoundation;
+
+using OS_nw_interface=System.IntPtr;
+
+namespace Network {
+	
+	public class NWInterface : NativeObject {
+		public NWInterface (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWListener.cs
```diff
@@ -0,0 +1,253 @@
...
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public static NWListener Create (string port, NWParameters parameters)
+		{
+			IntPtr handle;
+
+			if (parameters == null)
+				throw new ArgumentNullException (nameof (parameters));
+			if (port == null)
+				throw new ArgumentNullException (nameof (parameters));
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWListener.cs
```diff
@@ -0,0 +1,253 @@
...
+
+		[MonoPInvokeCallback (typeof (nw_listener_advertised_endpoint_changed_handler_t))]
+		static unsafe void TrampolineAdvertisedEndpointChangedHandler (IntPtr block, IntPtr endpoint, byte added)
+		{
+			var descriptor = (BlockLiteral *) block;
+			var del = (AdvertisedEndpointChanged) (descriptor->Target);
+			if (del != null){
+				var nwendpoint = new NWEndpoint (endpoint, owns: false);
+			        del (nwendpoint, added != 0 ? true : false);
+				nwendpoint.Dispose ();
```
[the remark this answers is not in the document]
migueldeicaza: Oh thank god.   This bit me hard, and ended up adding fences to track down this bug.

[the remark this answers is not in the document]
migueldeicaza: Updating all the sites that had this problem.

src/Network/NWParameters.cs
```diff
@@ -0,0 +1,516 @@
...
+		Handover = 1,
+		Interactive = 2,
+		Aggregate = 3, 
+	}
+
+	static partial class Libraries  {
+		static public class Network {
+                        static public readonly IntPtr Handle = Dlfcn.dlopen (Constants.NetworkLibrary, 0);
+                }
+	}
```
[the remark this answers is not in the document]
migueldeicaza: It is not being generated, I do not know the reason for this.

src/Network/NWParameters.cs
```diff
@@ -0,0 +1,516 @@
...
+
+		[MonoPInvokeCallback(typeof (nw_parameters_configure_protocol_block_t))]
+		static unsafe void TrampolineConfigureHandler (IntPtr block, IntPtr iface)
+		{
+                        var descriptor = (BlockLiteral *) block;
+                        var del = (Action<NWProtocolOptions>) (descriptor->Target);
+                        if (del != null){
+				var x = new NWProtocolOptions (iface, owns: false);
+				del (x);
+				x.Dispose ();
```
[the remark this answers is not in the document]
migueldeicaza: handled

src/Network/NWParameters.cs
```diff
@@ -0,0 +1,516 @@
...
+                [MonoPInvokeCallback (typeof (nw_parameters_iterate_interfaces_block_t))]
+		[return: MarshalAs (UnmanagedType.I1)]
+		static unsafe bool TrampolineIterateProhibitedHandler (IntPtr block, IntPtr iface)
+		{
+                        var descriptor = (BlockLiteral *) block;
+                        var del = (Func<NWInterface,bool>) (descriptor->Target);
+                        if (del != null){
+				var x = new NWInterface (iface, owns: false);
+                                var ret = del (x);
+				x.Dispose ();
```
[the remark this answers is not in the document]
migueldeicaza: handled

src/Network/NWPath.cs
```diff
@@ -0,0 +1,153 @@
...
+		Invalid = 0,
+		Satisfied = 1,
+		Unsatisfied = 2,
+		Satisfiable = 3
+	}
+	
+	public class NWPath : NativeObject {
+		public NWPath (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWPath.cs
```diff
@@ -0,0 +1,153 @@
...
+		public bool IsExpensive => nw_path_is_expensive (GetHandle());
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		[return: MarshalAs(UnmanagedType.U1)]
+		extern static bool nw_path_has_ipv4 (IntPtr handle);
+		
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public bool HasIpV4 => nw_path_has_ipv4 (GetHandle());
```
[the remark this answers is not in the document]
migueldeicaza: got i

src/Network/NWPath.cs
```diff
@@ -0,0 +1,153 @@
...
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.NetworkLibrary)]
+		[return: MarshalAs(UnmanagedType.U1)]
+		extern static bool nw_path_has_ipv6 (IntPtr handle);
+		
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public bool HasIpV6 => nw_path_has_ipv6 (GetHandle());
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Network/NWPath.cs
```diff
@@ -0,0 +1,153 @@
...
+		delegate void nw_path_enumerate_interfaces_block_t (IntPtr block, IntPtr iface);
+		static nw_path_enumerate_interfaces_block_t static_Enumerator = TrampolineEnumerator;
+		
+		[MonoPInvokeCallback (typeof (nw_path_enumerate_interfaces_block_t))]
+		static unsafe void TrampolineEnumerator (IntPtr block, IntPtr iface)
+		{
+			var descriptor = (BlockLiteral *) block;
+			var del = (Action<NWInterface>) (descriptor->Target);
+			if (del != null){
+				del (new NWInterface (iface, owns: false));
```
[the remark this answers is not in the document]
migueldeicaza: got it

src/Security/SecIdentity2.cs
migueldeicaza: Got it - will use the property version.

src/Security/SecIdentity2.cs
migueldeicaza: Got it - will use the property version.

src/Security/SecIdentity2.cs
migueldeicaza: got it

src/Security/SecProtocolMetadata.cs
migueldeicaza: got it

src/Security/SecProtocolOptions.cs
migueldeicaza: got it

src/Network/NWListener.cs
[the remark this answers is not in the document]
migueldeicaza: Thinking about this a little more - I think that the send/receive stacks probably should document instead that the instance is not usable after the callback, as the GetINativeObject path is slower than wrapping the instance.

migueldeicaza:
Addressed most of the issues here, but two caveats:

* I do not follow your guidance on the null issue being raised, need some help.

* Did not remove the CoreGraphics code, will merge that here.

* I suspect that we do not want GetINativeObject in the send/receive main methods, and rather document that you are not supposed to keep the object, as it will be released (and the check is already in place).      So that would be the NWError for Send, and the NWContentContext/NWError in Receive.

We should revisit the latter.

mandel-macaque:
The failing tests is due to the NWAdvertiseDescriptor.CreateBonjourService returning a null pointer which means that later accessing the handler returns all the errors seen in the report. The fix is quite simple, the method must take the type. A number of example types that Mac OS X uses can be found [here](https://developer.apple.com/library/archive/qa/qa1312/_index.html).

For example, changing the test to use the following fixes it:
```csharp
return NWAdvertiseDescriptor.CreateBonjourService ("sampleName" + DateTime.Now, "_nfs._tcp");
```
But this makes me wonder if we should not make type to be optional, the documentation on the headers only states that domain is optional, but not type (will add a review in the PR). But as you can see in the header docs:

```
*!
+ * @function nw_advertise_descriptor_create_bonjour_service
+ *
+ * @abstract
+ *		Creates a new advertise descriptor object based on a Bonjour service type
+ *		and optional domain. This object can be used with listener objects to
+ *      specify the service the listener should advertise.
+ *
+ *      If the name is unspecified, the default name for the device will be used.
+ *      If the domain is unspecified, the default domains for registration will be
+ *      used. If the listener is local only, the domain 'local.' will be used
+ *      regardless of the parameter passed to domain.
+ *
+ * @param name
+ *		An optional Bonjour service name.
+ *
+ * @param type
+ *		A Bonjour service type.
+ *
+ * @param domain
+ *		An optional Bonjour service domain.
+ *
+ * @result
+ *		An instantiated browse descriptor object.
+ */
```

src/Network/NWAdvertiseDescriptor.cs
```diff
@@ -0,0 +1,62 @@
...
+	
+	[TV (12,0), Mac (10,14), iOS (12,0)]
+	public class NWAdvertiseDescriptor : NativeObject {
+		public NWAdvertiseDescriptor (IntPtr handle, bool owns) : base (handle, owns)
+		{ }
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern IntPtr nw_advertise_descriptor_create_bonjour_service (string name, string type, string domain);
+
+		public static NWAdvertiseDescriptor CreateBonjourService (string name, string type = null, string domain = null)
```
mandel-macaque: @migueldeicaza I would not add type to be optional, since that will most (if not always) result in the nw_advertise_descriptor_create_bonjour_service returning a null pointer. I added a comment on how to fix the issue in the tests.

src/Security/SecProtocolOptions.cs
```diff
@@ -0,0 +1,201 @@
...
+		static extern void sec_protocol_options_add_tls_ciphersuite_group (sec_protocol_options_t handle, SslCipherSuiteGroup cipherSuiteGroup);
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		public void AddTlsCipherSuiteGroup (SslCipherSuiteGroup cipherSuiteGroup) => sec_protocol_options_add_tls_ciphersuite_group (GetCheckedHandle (), cipherSuiteGroup);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.SecurityLibrary)]
+		static extern void sec_protocol_options_set_tls_min_version (sec_protocol_options_t handle, SslProtocol protocol);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		void SetTlsMinVersion (SslProtocol protocol) => sec_protocol_options_set_tls_min_version (GetCheckedHandle (), protocol);
```
spouliot: `public` ?

src/Security/SecProtocolOptions.cs
```diff
@@ -0,0 +1,201 @@
...
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.SecurityLibrary)]
+		static extern void sec_protocol_options_set_tls_max_version (sec_protocol_options_t handle, SslProtocol protocol);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		void SetTlsMaxVersion (SslProtocol protocol) => sec_protocol_options_set_tls_max_version (GetCheckedHandle (), protocol);
```
spouliot: `public` ?

src/Security/SecProtocolOptions.cs
```diff
@@ -0,0 +1,201 @@
...
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		[DllImport (Constants.SecurityLibrary)]
+		static extern void sec_protocol_options_add_tls_application_protocol (sec_protocol_options_t handle, string applicationProtocol);
+
+		[TV (12,0), Mac (10,14), iOS (12,0)]
+		void AddTlsApplicationProtocol (string applicationProtocol)
```
spouliot: `public` ?

src/Security/SecProtocolOptions.cs
migueldeicaza: agreed.

src/Security/SecProtocolOptions.cs
migueldeicaza: agreed.

src/Security/SecProtocolOptions.cs
migueldeicaza: agreed.

migueldeicaza:
The source code for CFRetain shows that it handles NSObject retain as well:

https://opensource.apple.com/source/CF/CF-550/CFRuntime.c

Mike Ash also says that this is allowed:

https://www.mikeash.com/pyblog/friday-qa-2010-01-22-toll-free-bridging-internals.html

This shows that when it is dealing with an NSObject, this invokes the ObjC retain/release.

That said, I do not mind changing that to call NSObject.retain/NSObject release

migueldeicaza:
Hello - and I just validated that calling CFRelease/CFRetain on an NSObject does indeed call the overwritten method.

PR #4253, opened 2018-06-14
[UIKit] Update for Xcode10.

mandel-macaque:
This is an update for the Xcode 10. I am interested in input regarding the addition of the new protocols that have been added with old methods (already present) and new ones:

* UIUserActivityRestoring: "void RestoreUserActivityState (NSUserActivity userActivity)" was already present.
* UIFocusItemScrollableContainer: "ContentOffset" and "ContentSize" already present.
* UIFocusItemContainer: "Frame" already present.

I added the methods in the protocols with the correct signature, left the methods in the old interfaces and added new to ignore errors. Comments are welcome regarding this approach.

Diffs are:

* https://github.com/xamarin/xamarin-macios/wiki/UIKit-iOS-xcode10-beta1
* https://github.com/xamarin/xamarin-macios/wiki/UIKit-tvOS-xcode10-beta1
* https://github.com/xamarin/xamarin-macios/wiki/UIKit-watchOS-xcode10-beta1


src/UIKit/UIEnums.cs
```diff
@@ -2253,5 +2253,24 @@ public enum UIDocumentBrowserErrorCode : long
...
+		Automatic = 0,
+		Extended,
+		Standard,
+	}
+
+	[iOS (12,0), NoTV, NoWatch]
+	[Native]
+	public enum UIPrintErrorCode : long
+	{
+		ingNotAvailableError = 1,
```
rolfbjarne: `ing`?

src/spritekit.cs
```diff
@@ -149,8 +149,13 @@ partial interface SKNode : NSSecureCoding, NSCopying {
 		[return: NullAllowed]
 		SKNode Create (string filename);
 
+		// new is needed after iOS 12 due to the new UIFocustItem protocol.
```
rolfbjarne: I suppose the comment should say `UIFocusItem`?

src/uikit.cs
```diff
@@ -16687,6 +16732,15 @@ interface UIFocusItem : UIFocusEnvironment
 		[Abstract]
 		[Export ("canBecomeFocused")]
 		bool CanBecomeFocused { get; }
+
+		[TV (12, 0), iOS (12, 0), NoWatch]
+		[Abstract]
```
rolfbjarne: This is a breaking change: can't add new `[Abstract]` members to an existing protocol.

It should go into a `#if XAMCORE_4_0` block.

src/uikit.cs
```diff
@@ -16930,7 +17002,30 @@ interface UIFocusEnvironment {
 		[Export ("soundIdentifierForFocusUpdateInContext:")]
 		[return: NullAllowed]
 		NSString GetSoundIdentifier (UIFocusUpdateContext context);
-		
+
+
+		[TV (12, 0), iOS (12, 0)]
+		[Abstract]
```
rolfbjarne: Breaking change.

src/uikit.cs
```diff
@@ -16930,7 +17002,30 @@ interface UIFocusEnvironment {
...
+		[NullAllowed, Export ("parentFocusEnvironment", ArgumentSemantic.Weak)]
+		IUIFocusEnvironment ParentFocusEnvironment { get; }
+
+		[TV (12, 0), iOS (12, 0)]
+		[Abstract]
```
rolfbjarne: Breaking change.

src/uikit.cs
```diff
@@ -18235,6 +18330,52 @@ interface UIDocumentBrowserAction {
...
+		[Export ("focusItemsInRect:")]
+		IUIFocusItem[] GetFocusItems (CGRect rect);
+	}
+
+	[iOS (12,0), TV(12,0), NoWatch]
+	[Protocol]
+	interface UIFocusItemScrollableContainer : UIFocusItemContainer
+	{
+		[Abstract]
+		[TV(9,0), NoWatch]
```
rolfbjarne: The protocol itself is `TV (12,0)`, but there are members with `TV (9,0)`? That sounds strange (but otoh it wouldn't surprise me either...)

src/uikit.cs
mandel-macaque: So XAMCORE_4_0 then

mandel-macaque:
@rolfbjarne changes pushed, added the #is statements and filled the rdars that have been added in the comments and in the trello board.

src/uikit.cs
```diff
@@ -16687,6 +16732,19 @@ interface UIFocusItem : UIFocusEnvironment
 		[Abstract]
 		[Export ("canBecomeFocused")]
 		bool CanBecomeFocused { get; }
+
+		// FIXME: declared as a @required, but this breaks compatibilit
+		// Radar: 41121416
+#if XAMCORE_4_0
```
rolfbjarne: Only the `[Abstract]` need to be in `XAMCORE_4_0`:

```csharp
#if XAMCORE_4_0
    [Abstract]
#endif
    [TV (12, 0), iOS (12, 0), NoWatch]
    [Export ("frame")]
    CGRect Frame { get; }
```

src/uikit.cs
```diff
@@ -5235,7 +5235,7 @@ interface UIDocument : NSFilePresenter, NSProgressReporting {
 
 		[iOS (8,0)]
 		[Export ("restoreUserActivityState:")]
-		void RestoreUserActivityState (NSUserActivity userActivity);
+		new void RestoreUserActivityState (NSUserActivity userActivity);
```
rolfbjarne: Can you try locally to remove this method, so that we get the one from the protocol, and see what happens to the api diff? I _think_ it won't be a breaking change, in which case it's better to remove it.

src/uikit.cs
```diff
@@ -6133,6 +6133,10 @@ interface UIGraphicsImageRendererFormat
 		[Export ("opaque")]
 		bool Opaque { get; set; }
 
+		[Introduced (PlatformName.iOS, 10, 0, message: "Use the 'PreferredRange' property instead.")]
```
spouliot: you don't need the `Introduced` here

src/uikit.cs
```diff
@@ -6133,6 +6133,10 @@ interface UIGraphicsImageRendererFormat
...
+		[Deprecated (PlatformName.iOS, 12, 0, message: "Use the 'PreferredRange' property instead.")]
+		[Introduced (PlatformName.TvOS, 10, 0, message: "Use the 'PreferredRange' property instead.")]
```
spouliot: same

src/uikit.cs
```diff
@@ -18235,6 +18337,48 @@ interface UIDocumentBrowserAction {
...
+		[Abstract]
+		[Export ("contentSize")]
+		CGSize ContentSize { get; }
+
+		[Abstract]
+		[Export ("visibleSize")]
+		CGSize VisibleSize { get; }
+	}
+
+	[iOS (8,0), TV (8,0), NoWatch] // it was added on 8,0, but was not binded and the method was added in 12,0
```
spouliot: tvOS started with 9.0 so remove it's attribute (which means it's always been available)

src/uikit.cs
```diff
@@ -18276,4 +18420,16 @@ interface UIFontMetrics {
...
+	[iOS (12,0), TV (12,0), Watch (5,0)]
+	interface UIKitVersion {
+
+		[Field ("UIKitVersionNumber")]
+		double UIKitVersionNumber { get; }
+
+		[Field ("UIKitVersionString")]
+		NSString UIKitVersionString { get; }
+
+	}
```
spouliot: remove those and put them in `UIKit.ignore`
we're doing the same in other semi-versioned frameworks

src/spritekit.cs
```diff
@@ -149,8 +149,10 @@ partial interface SKNode : NSSecureCoding, NSCopying {
 		[return: NullAllowed]
 		SKNode Create (string filename);
 
+#if MONOMAC || WATCH
```
VincentDondain: I don't understand that change. I believe this is a breaking change. Also I'm confused about a SpriteKit change in this UIKit PR.

mandel-macaque: No, it is not. In iOS and TV we have the IUIFocusItem interface that already contains the [Frame property](https://github.com/xamarin/xamarin-macios/pull/4253/files#diff-1ca9e8e46629c0f8806cee2c8be4aad3R16730). That means that in those platforms, it can be removed. 

On WatchOS and MacOS we do not have that interface, therefore the property does not get generated and you have to add it manually.

mandel-macaque: @VincentDondain once the build is done, take a look at the api diff and the Extrospection results both confirm my statement :)

VincentDondain: Ah got it, didn't realize it was in `IUIFocusItem` 👍

src/uikit.cs
```diff
@@ -10024,15 +10030,12 @@ interface UIScreen : UITraitEnvironment {
...
+	interface UIScrollView : UIFocusItemScrollableContainer {
 		[Export ("initWithFrame:")]
 		IntPtr Constructor (CGRect frame);
 
-		[Export ("contentOffset")]
-		CGPoint ContentOffset { get; set; }
-
 		[Export ("contentSize")]
-		CGSize ContentSize { get; set; }
+		new CGSize ContentSize { get; set; }
```
rolfbjarne: Can this be removed as well, so that we get it from `UIFocusItemScrollableContainer` instead, or will the api diff complain?

src/uikit.cs
```diff
@@ -12995,7 +12998,7 @@ interface UIView : UIAppearance, UIAppearanceContainer, UIAccessibility, UIDynam
 		CoreAnimation.CALayer Layer { get; }
 
 		[Export ("frame")]
-		CGRect Frame { get; set; }
+		new CGRect Frame { get; set; }
```
rolfbjarne: Same: can this be removed as well, so that we get it from `UIFocusItemContainer ` instead, or will the api diff complain?

src/uikit.cs
```diff
@@ -9862,10 +9872,6 @@ interface UIResponder : UIAccessibilityAction, UIAccessibilityFocus
 		[Export ("updateUserActivityState:")]
 		void UpdateUserActivityState (NSUserActivity activity);
 
-		[iOS (8,0)]
-		[Export ("restoreUserActivityState:")]
-		void RestoreUserActivityState (NSUserActivity activity);
```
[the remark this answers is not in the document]
dalexsoto: mmm, we have allowed parameter name changes in the past, but since the `UIUserActivityRestoring` protocol just got added lets just rename  `userActivity` to `activity` to avoid the breaking change.

src/uikit.cs
```diff
@@ -18235,6 +18326,48 @@ interface UIDocumentBrowserAction {
...
+	}
+
+	[iOS (8,0), NoWatch] // it was added on 8,0, but was not binded and the method was added in 12,0
+	[Protocol]
+	interface UIUserActivityRestoring
+	{
+		[Abstract]
+		[iOS (8,0), TV(12,0)]
+		[Export ("restoreUserActivityState:")]
+		void RestoreUserActivityState (NSUserActivity userActivity);
```
dalexsoto: Rename `userActivity` to `activity` to avoid a breaking change.

src/uikit.cs
mandel-macaque: No, in this case the protocol just exposes the get, but not the set.

src/uikit.cs
mandel-macaque: Same, protocol just exposes get, not set.

src/uikit.cs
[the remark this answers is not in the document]
mandel-macaque: Sure, also, braking change is probably a too "strong" adjective. API will change, not break ;)

Doing it nevertheless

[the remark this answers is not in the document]
dalexsoto: It is a breaking change because of named parameters for example if anyone is doing  RestoreUserActivityState (activity: x); it would not compile anymore, but yeah not very likely

PR #4263, opened 2018-06-15
[SpriteKit] Update bindings to Xcode 10 Beta 1

dalexsoto:


src/SpriteKit/SKNode.cs
```diff
@@ -50,6 +50,27 @@ IEnumerator IEnumerable.GetEnumerator ()
 		{
 			return GetEnumerator ();
 		}
+
+		[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		public static SKNode Create (string filename, Type [] types, out NSError error)
+		{
+			// Let's fail early.
+			if (filename is null)
```
spouliot: let's use `==`, not `is`, for consistency and not always knowing if `==` was (or not) overloaded.

src/SpriteKit/SKNode.cs
```diff
@@ -50,6 +50,27 @@ IEnumerator IEnumerable.GetEnumerator ()
...
+				throw new ArgumentNullException (nameof (filename));
+			if (types is null)
+				throw new ArgumentNullException (nameof (filename));
```
spouliot: same

src/SpriteKit/SKNode.cs
```diff
@@ -50,6 +50,27 @@ IEnumerator IEnumerable.GetEnumerator ()
...
+
+			using (var classes = new NSMutableSet<Class> (types.Length)) {
+				foreach (var type in types)
+					classes.Add (new Class (type));
+				return Create (filename, classes.Handle, out error);
+			}
+		}
+
+		[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		public static SKNode Create (string filename, NSSet<Class> classes, out NSError error) => Create (filename, classes.Handle, out error);
```
spouliot: why not bind this signature directly (i.e. avoid the extra internal method) ? and have the one above call it

dalexsoto:
Failure due to https://github.com/xamarin/xamarin-macios/issues/4241

PR #4268, opened 2018-06-15
[UserNotificationsUI] Update the fro Xcode 10.

mandel-macaque:
Diff can be found here: 

https://github.com/xamarin/xamarin-macios/wiki/UserNotificationsUI-iOS-xcode10-beta1

### DIFF
--- a/src/usernotificationsui.cs
+++ b/src/usernotificationsui.cs
@@ -83,6 +83,23 @@ namespace UserNotificationsUI {
 
 		[Export ("mediaPlayingPaused")]
 		void MediaPlayingPaused ();
+
+		[NoWatch, NoTV, NoMac, iOS (12,0)]
+		[Export ("performNotificationDefaultAction")]
+		void PerformNotificationDefaultAction ();
+
+		[NoWatch, NoTV, NoMac, iOS (12,0)]
+		[Export ("dismissNotificationContentExtension")]
+		void DismissNotificationContentExtension ();
+
+		// property, but we have to add the two methods
+		[iOS (12, 0)]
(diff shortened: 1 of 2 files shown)


src/usernotificationsui.cs
```diff
@@ -83,6 +83,23 @@ interface NSExtensionContext_UNNotificationContentExtension {
...
+
+		[NoWatch, NoTV, NoMac, iOS (12,0)]
+		[Export ("performNotificationDefaultAction")]
+		void PerformNotificationDefaultAction ();
+
+		[NoWatch, NoTV, NoMac, iOS (12,0)]
+		[Export ("dismissNotificationContentExtension")]
+		void DismissNotificationContentExtension ();
+
+		// property, but we have to add the two methods
```
rolfbjarne: Maybe add "because this is a category" here.

mandel-macaque: @rolfbjarne valid point, we never know who is going to read the code. Will push a small change with that.

src/usernotificationsui.cs
```diff
@@ -83,6 +83,23 @@ interface NSExtensionContext_UNNotificationContentExtension {
...
+		[iOS (12, 0)]
+		[Export ("notificationActions", ArgumentSemantic.Copy)]
+		UNNotificationAction[] NotificationActions ();
+
+		[iOS (12, 0)]
+		[Export ("setNotificationActions:", ArgumentSemantic.Copy)]
```
dalexsoto: Remove the ArgumentSemantic, not used.

mandel-macaque:
Timeouts are not related to the PR.

PR #4270, opened 2018-06-15
[xharness] Fix publishing xtro's html report.

rolfbjarne:
1. Put xtro's results in a subdirectory of the current test's log directory,
   not a sibling directory of the root log directory (which would prevent it
   from being uploaded after a test run, since only the root log directory is
   uploaded).

2. Just add the existing index.html as the html report to the collection of
   logs, no need to create a new file pointing to it. This fixes a problem
   where the generated html file would redirect to a local file, which would
   not work when served from a web server.


rolfbjarne:
build

PR #4294, opened 2018-06-19
Bump for xcode10 beta 2

spouliot:


tools/common/StaticRegistrar.cs
```diff
@@ -2705,11 +2705,15 @@ void Specialize (AutoIndentStringBuilder sb)
...
 				if (Driver.XcodeVersion == new Version (10, 0) && @class.Type.Namespace == "TVMLKit") {
-					// TVML's headers are broken in Xcode 10 beta 1: rdar://40824310
-					// I'm restricting this to a very specific version of Xcode,
-					// so that we can remove the workaround asap once Apple fixes their headers.
-					if (Driver.XcodeBundleVersion != "14274.16")
-						throw ErrorHelper.CreateError (99, "Verify if the workaround for rdar://40824310 is still needed.");
+					// TVML's headers are broken in Xcode 10 beta: rdar://40824310
+					switch (Driver.XcodeBundleVersion) {
+					case "14274.16": // beta 1
+					case "14274.19": // beta 2
```
VincentDondain: @spouliot just double checking but did you check the workaround is still needed?

spouliot: @VincentDondain are you double-checking ? or am I double-checking ? ;-)

Seriously, yes and this, and the other one, are still needed. It's quite possible the beta 2 build was already frozen when the rdar were filed.

dalexsoto:
@VincentDondain without some changes into CarPlay the PR would not be green (Apple removed some selectors) so I just figured instead of making some changes here and after making an actual PR for the rest; I would just dump everything here 😄

spouliot:
watchOS test failure due to invalid XML results -> https://github.com/xamarin/maccore/issues/827

PR #4300, opened 2018-06-20
[NaturalLanguage] Xcode10-beta2 update.

mandel-macaque:


mandel-macaque:
Test failure is unrelated to the PR since the Framework it is not supported in that platform.

PR #4310, opened 2018-06-21
Fix resetting a README dependency when the hash exists, but the branch does not. Fixes #4308.

### LINKED ISSUE #4308: make reset-X doesn't work for README dependencies if the hash is already present, but the branch is not
`make reset-X` doesn't work for README dependencies if the required hash is already fetched, but the branch is not.

Repro:
* Create & push a new branch from the an existing hash in maccore.
* Change the NEEDED_MACCORE_BRANCH in mk/xamarin.mk to the new branch.
* `make reset-maccore`

rolfbjarne:
Fix this by checking if the branch exists, and if not, fetch the repository.

Also add a better error message if the remote doesn't have the requested branch.

Fixes https://github.com/xamarin/xamarin-macios/issues/4308.


rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/868)

PR #4333, opened 2018-06-25
[jenkins] Add support for completely skipping the public jenkins job using a label.

rolfbjarne:
This can be useful when working on the support for private jenkins, since many
of those changes can only be tested as a pull request, and they also tend to
include a lot of commits (because nothing can be tested locally).

So until whatever needs implementing for private jenkins is complete and
working, there's no need for the public bots to do anything for such pull
requests.


rolfbjarne:
build

PR #4336, opened 2018-06-25
[HealthKit] Bump to Xcode10-beta2

mandel-macaque:


src/healthkit.cs
```diff
@@ -189,6 +189,18 @@ interface HKPredicateKeyPath {
...
 		[Field ("HKPredicateKeyPathWorkoutTotalFlightsClimbed")]
 		NSString TotalFlightsClimbed { get; }
+
+		[Watch (5, 0), iOS (12, 0)]
+		[Field ("HKPredicateKeyPathSum")]
+		NSString PathSum { get; }
+
+		[NoWatch, iOS (12, 0)]
+		[Field ("HKPredicateKeyPathClinicalRecordFHIRResourceIdentifier")]
+		NSString ClinicalRecordFHIRResourceIdentifier { get; }
```
dalexsoto: `ClinicalRecordFHIRResourceIdentifier` -> `ClinicalRecordFhirResourceIdentifier`

src/healthkit.cs
```diff
@@ -189,6 +189,18 @@ interface HKPredicateKeyPath {
...
+
+		[NoWatch, iOS (12, 0)]
+		[Field ("HKPredicateKeyPathClinicalRecordFHIRResourceType")]
```
dalexsoto: `HKPredicateKeyPathClinicalRecordFHIRResourceType ` -> `HKPredicateKeyPathClinicalRecordFhirResourceType `

src/healthkit.cs
```diff
@@ -2442,4 +2557,220 @@ interface HKWorkoutRouteQuery {
...
+
+		[Export ("workoutEvents", ArgumentSemantic.Copy)]
+		HKWorkoutEvent[] WorkoutEvents { get; }
+
+		[Export ("initWithHealthStore:configuration:device:")]
+		IntPtr Constructor (HKHealthStore healthStore, HKWorkoutConfiguration configuration, [NullAllowed] HKDevice device);
+
+		[Async]
+		[Export ("beginCollectionWithStartDate:completion:")]
+		void BeginCollection (NSDate startDate, Action<bool, NSError> completionHandler);
```
dalexsoto: There are a bunch of `Action<bool, NSError>` what does the `bool` means? Can't we have a named delegate here instead?

src/healthkit.cs
```diff
@@ -2442,4 +2557,220 @@ interface HKWorkoutRouteQuery {
...
+
+		[Export ("statisticsForType:")]
+		[return: NullAllowed]
+		HKStatistics Statistics (HKQuantityType quantityType);
+
+		[Export ("seriesBuilderForType:")]
+		HKSeriesBuilder SeriesBuilder (HKSeriesType seriesType);
+	}
+
+	delegate void HKQuantitySeriesSampleQueryQuantityDelegate (HKQuantitySeriesSampleQuery query, HKQuantity quantity, NSDate date, bool done, NSError error);
```
dalexsoto: just make sure that `bool done` isn't an out/ref param

src/healthkit.cs
```diff
@@ -2442,4 +2557,220 @@ interface HKWorkoutRouteQuery {
...
+
+		[NullAllowed, Export ("device", ArgumentSemantic.Copy)]
+		HKDevice Device { get; }
+
+		[Export ("insertQuantity:date:error:")]
+		bool Insert (HKQuantity quantity, NSDate date, [NullAllowed] out NSError error);
+
+		[Async, Internal]
+		[Export ("finishSeriesWithMetadata:completion:")]
+		void FinishSeries ([NullAllowed] NSDictionary metadata, Action<NSArray<HKQuantitySample>, NSError> completionHandler);
```
dalexsoto: This time a delegate is a much better alternative

``` csharp
Action<NSArray<HKQuantitySample>

delegate void Foo (HKQuantitySample [] foo);
```

You get a managed array vs an NSArray

src/healthkit.cs
```diff
@@ -2442,4 +2557,220 @@ interface HKWorkoutRouteQuery {
...
+
+		[Async]
+		[Wrap ("FinishSeries (metadata.Dictionary, completionHandler)")]
+		void FinishSeries ([NullAllowed] HKMetadata metadata, Action<NSArray<HKQuantitySample>, NSError> completionHandler);
```
dalexsoto: same `NSArray<HKQuantitySample>` vs `HKQuantitySample []` I do not recall if Action supports the managed array version, if not just use a named delegate

src/healthkit.cs
```diff
@@ -2442,4 +2557,220 @@ interface HKWorkoutRouteQuery {
...
+	interface HKClinicalRecord : NSSecureCoding, NSCopying
+	{
+		[Export ("clinicalType", ArgumentSemantic.Copy)]
+		HKClinicalType ClinicalType { get; }
+
+		[Export ("displayName")]
+		string DisplayName { get; }
+
+		[NullAllowed, Export ("FHIRResource", ArgumentSemantic.Copy)]
+		HKFHIRResource FHIRResource { get; }
```
dalexsoto: Unsure if we should turn this into `FhirResource `

mandel-macaque: Same opinion, not sure which one would be better.

src/healthkit.cs
mandel-macaque: Same as with the other property. Lets look at other API and be consistent.

src/healthkit.cs
mandel-macaque: @dalexsoto all those, are following the pattern already  found in the file, I personally would follow the same pattern in all possible cases. @spouliot @VincentDondain do you have any comments since you work on this before?

src/healthkit.cs
mandel-macaque: AFAIK the delegate is correct:

void (^)(HKQuantitySeriesSampleQuery *query, HKQuantity *quantity, NSDate *date, BOOL done, NSError *error))quantityHandler;

src/healthkit.cs
mandel-macaque: Changed them in the next commit after looking in the entire file.

src/healthkit.cs
```diff
@@ -492,6 +504,16 @@ interface HKHealthStore {
 		[Notification]
 		[Field ("HKUserPreferencesDidChangeNotification")]
 		NSString UserPreferencesDidChangeNotification { get; }
+
+		[Async]
+		[Watch (5,0), iOS (12,0)]
+		[Export ("getRequestStatusForAuthorizationToShareTypes:readTypes:completion:")]
+		void GetRequestStatus (NSSet<HKSampleType> typesToShare, NSSet<HKObjectType> typesToRead, Action<HKAuthorizationRequestStatus, NSError> completion);
```
rolfbjarne: The method for `requestAuthorizationToShareTypes:readTypes:completion:` is named `RequestAuthorizationToShare`, so I think this would be better named as `GetRequestStatusForAuthorizationToShare`.

Another argument for the same change is that the For* suffix does not refer to the first parameter, but explains what the function does.

src/healthkit.cs
```diff
@@ -1134,6 +1197,28 @@ interface HKQuery {
 		[Static]
 		[Export ("predicateForActivitySummariesBetweenStartDateComponents:endDateComponents:")]
 		NSPredicate GetPredicateForActivitySummariesBetween (NSDateComponents startDateComponents, NSDateComponents endDateComponents);
+
+
+		// @interface HKClinicalRecordPredicates (HKQuery)
+		[NoWatch, iOS (12,0)]
+		[Static, Internal]
+		[Export ("predicateForClinicalRecordsWithFHIRResourceType:")]
+		NSPredicate GetPredicate (NSString resourceType);
```
rolfbjarne: All the other predicateFor* selectors in this class are bound as `GetPredicateFor*`, so I think the same applies here: `GetPredicateForClinicalRecords` (and the overloads below as well).

src/healthkit.cs
```diff
@@ -2432,6 +2538,15 @@ interface HKWorkoutRouteBuilder {
...
 		[Async, Wrap ("FinishRoute (workout, metadata != null ? metadata.Dictionary : null, completion)")]
 		void FinishRoute (HKWorkout workout, HKMetadata metadata, Action<HKWorkoutRoute, NSError> completion);
+
+		[Watch (5, 0), iOS (12, 0)]
+		[Async, Protected]
+		[Export ("addMetadata:completion:")]
+		void AddMetadata (NSDictionary metadata, Action<bool, NSError> completion);
+
+		[Watch (5, 0), iOS (12, 0)]
+		[Async, Wrap ("AddMetadata (metadata != null ? metadata.Dictionary : null, completion)")]
```
rolfbjarne: No need to change anything, but for the future you can use `metadata?.Dictionary` to make the wrap expression smaller 😉

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+
+		[Export ("discardWorkout")]
+		void DiscardWorkout ();
+
+		[Export ("elapsedTimeAtDate:")]
+		double ElapsedTimeAtDate (NSDate date);
+
+		[Export ("statisticsForType:")]
+		[return: NullAllowed]
+		HKStatistics Statistics (HKQuantityType quantityType);
```
rolfbjarne: This needs a verb: `GetStatistics`.

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+
+		[Async]
+		[Export ("finishWorkoutWithCompletion:")]
+		void FinishWorkout (Action<HKWorkout, NSError> completionHandler);
+
+		[Export ("discardWorkout")]
+		void DiscardWorkout ();
+
+		[Export ("elapsedTimeAtDate:")]
+		double ElapsedTimeAtDate (NSDate date);
```
rolfbjarne: This needs a verb: `GetElapsedTime (NSDate atDate)`

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+
+		[Export ("statisticsForType:")]
+		[return: NullAllowed]
+		HKStatistics Statistics (HKQuantityType quantityType);
+
+		[Export ("seriesBuilderForType:")]
+		HKSeriesBuilder SeriesBuilder (HKSeriesType seriesType);
```
rolfbjarne: This needs a verb: `GetSeriesBuilder`.

src/healthkit.cs
```diff
@@ -189,6 +189,18 @@ interface HKPredicateKeyPath {
...
 		[Field ("HKPredicateKeyPathWorkoutTotalFlightsClimbed")]
 		NSString TotalFlightsClimbed { get; }
+
+		[Watch (5, 0), iOS (12, 0)]
+		[Field ("HKPredicateKeyPathSum")]
+		NSString PathSum { get; }
+
+		[NoWatch, iOS (12, 0)]
+		[Field ("HKPredicateKeyPathClinicalRecordFHIRResourceIdentifier")]
+		NSString ClinicalRecordFhiRResourceIdentifier { get; }
```
spouliot: what's `FHIR` and why is it turned into `FhiR` instead of `Fhir` ?

src/healthkit.cs
```diff
@@ -189,6 +189,18 @@ interface HKPredicateKeyPath {
...
+
+		[NoWatch, iOS (12, 0)]
+		[Field ("HKPredicateKeyPathClinicalRecordFHIRResourceType")]
+		NSString ClinicalRecordFhiRResourceType { get; }
```
spouliot: same

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+		void EnableCollection (HKQuantityType quantityType, [NullAllowed] NSPredicate predicate);
+
+		[Export ("disableCollectionForType:")]
+		void DisableCollection (HKQuantityType quantityType);
+	}
+
+	[NoWatch, iOS (12,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface HKFHIRResource : NSSecureCoding, NSCopying
```
rolfbjarne: You've bound `FHIR` as `FhiR` elsewhere.

src/healthkit.cs
```diff
@@ -643,9 +665,17 @@ interface HKMetadata {
 		[Export ("ElevationAscended")]
 		HKQuantity ElevationAscended { get; set; }
 
-		[Watch (4,2), iOS (11,2)]
-		[Export ("ElevationDescended")]
-		HKQuantity ElevationDescended { get; set; }
```
spouliot: why was the above removed in the PR ?

src/HealthKit/Enums.cs
```diff
@@ -564,4 +564,52 @@ public enum HKVO2MaxTestType : long {
 		PredictionSubMaxExercise,
 		PredictionNonExercise,
 	}
+ 
+	[NoWatch, iOS (12, 0)]
+	public enum HKFHIRResourceType {
```
rolfbjarne: You've bound `FHIR` as `FhiR` elsewhere (btw: why isn't that `Fihr`? it seems strange to capitalize the last letter only).

src/healthkit.cs
```diff
@@ -893,6 +935,15 @@ interface HKObjectType : NSSecureCoding, NSCopying {
...
 		[return: NullAllowed]
 		HKSeriesType GetSeriesType (string identifier);
+
+		[Watch (5,0), iOS (12,0)]
+		[Static, Internal]
+		[Export ("clinicalTypeForIdentifier:")]
+		[return: NullAllowed]
+		HKClinicalType GetClinicalType (NSString identifier);
+
+		[Wrap ("GetClinicalType (identifier.GetConstant ())")]
```
spouliot: copy the availability attributes to the developer can see when it's available

src/HealthKit/Enums.cs
spouliot: `HKFHIRResourceType` -> `HKFhirResourceType`

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+
+		[NullAllowed, Export ("endDate", ArgumentSemantic.Copy)]
+		NSDate EndDate { get; }
+
+		[Export ("workoutConfiguration", ArgumentSemantic.Copy)]
+		HKWorkoutConfiguration WorkoutConfiguration { get; }
+
+		[Internal]
+		[Export ("metadata", ArgumentSemantic.Copy)]
+		NSDictionary _Metadata { get; }
```
spouliot: it's not symmetrical with the above `[Protected]` `AddMetadata` API

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+		[Export ("startDate", ArgumentSemantic.Copy)]
+		NSDate StartDate { get; }
+
+		[NullAllowed, Export ("device", ArgumentSemantic.Copy)]
+		HKDevice Device { get; }
+
+		[Export ("insertQuantity:date:error:")]
+		bool Insert (HKQuantity quantity, NSDate date, [NullAllowed] out NSError error);
+
+		[Async, Internal]
```
spouliot: inconsistent accessibility

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+		[Export ("finishSeriesWithMetadata:completion:")]
+		void FinishSeries ([NullAllowed] NSDictionary metadata, HKQuantitySeriesSampleBuilderFinishSeriesDelegate completionHandler);
+
+		[Async]
+		[Wrap ("FinishSeries (metadata.Dictionary, completionHandler)")]
```
spouliot: That will throw an NRE even if null is allowed in the internal method -> `metadata?.Dictionary`

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+	interface HKClinicalRecord : NSSecureCoding, NSCopying
+	{
+		[Export ("clinicalType", ArgumentSemantic.Copy)]
+		HKClinicalType ClinicalType { get; }
+
+		[Export ("displayName")]
+		string DisplayName { get; }
+
+		[NullAllowed, Export ("FHIRResource", ArgumentSemantic.Copy)]
+		HKFHIRResource FhiRResource { get; }
```
spouliot: why not `FhirResource` ?

src/healthkit.cs
```diff
@@ -2442,4 +2557,222 @@ interface HKWorkoutRouteQuery {
...
+	}
+
+	interface IHKLiveWorkoutBuilderDelegate {}
+	[Watch (5,0), NoiOS]
+	[Protocol]
```
spouliot: it's a delegate so we provide the `[Model]`

src/healthkit.cs
mandel-macaque: That was me screwing up sed, sorry

src/healthkit.cs
spouliot: You used `[Protected]` on similar API

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
...
+		void EnableCollection (HKQuantityType quantityType, [NullAllowed] NSPredicate predicate);
+
+		[Export ("disableCollectionForType:")]
+		void DisableCollection (HKQuantityType quantityType);
+	}
+
+	[NoWatch, iOS (12,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface HKFhirResource : NSSecureCoding, NSCopying
```
spouliot: that's not the native name, you need to update `BaseType`

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
...
+
+		[Export ("elapsedTimeAtDate:")]
+		double GetElapsedTimeAtDate (NSDate date);
+
+		[Export ("statisticsForType:")]
+		[return: NullAllowed]
+		HKStatistics GetStatistics (HKQuantityType quantityType);
+
+		[Export ("seriesBuilderForType:")]
+		HKSeriesBuilder SeriesBuilder (HKSeriesType seriesType);
```
rolfbjarne: This needs a verb: `GetSeriesBuilder`

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
...
+
+		[Async]
+		[Export ("finishWorkoutWithCompletion:")]
+		void FinishWorkout (HKWorkoutBuilderCompletionHandler completionHandler);
+
+		[Export ("discardWorkout")]
+		void DiscardWorkout ();
+
+		[Export ("elapsedTimeAtDate:")]
+		double GetElapsedTimeAtDate (NSDate date);
```
rolfbjarne: The `AtDate` suffix is not needed: `double GetElapsedTime (NSDate atDate)` looks better.

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
...
+
+		[NullAllowed, Export ("endDate", ArgumentSemantic.Copy)]
+		NSDate EndDate { get; }
+
+		[Export ("workoutConfiguration", ArgumentSemantic.Copy)]
+		HKWorkoutConfiguration WorkoutConfiguration { get; }
+
+		[Protected]
+		[Export ("metadata", ArgumentSemantic.Copy)]
+		NSDictionary _Metadata { get; }
```
rolfbjarne: API with leading underscores should not be exposed (even if it's protected it's still exposed). Maybe something like `NativeMetadata` instead? IIRC we've used a `Native` suffix for something like this elsewhere.

src/HealthKit/Enums.cs
```diff
@@ -564,4 +564,52 @@ public enum HKVO2MaxTestType : long {
...
+		[Field ("HKClinicalTypeIdentifierProcedureRecord")]
+		ProcedureRecord,
+		[Field ("HKClinicalTypeIdentifierVitalSignRecord")]
+		VitalSignRecord,
+	}
+
+	[Watch (5,0), iOS (12,0)]
+	[Native]
+	public enum HKAuthorizationRequestStatus : long 
+	{
```
VincentDondain: Minor styling issue here.

tests/introspection/ApiTypoTest.cs
```diff
@@ -206,6 +206,7 @@ protected ApiTypoTest ()
 			"Exr",
 			"Felica", // Japanese contactless RFID smart card system
 			"Femtowatts",
+			"Fhir",
```
VincentDondain: I think that means "Fast Healthcare Interoperability Resources" if we care to add a comment.

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
 		IntPtr Constructor (HKWorkoutRoute workoutRoute, HKWorkoutRouteBuilderDataHandler dataHandler);
 	}
 
+	delegate void HKWorkoutBuilderCompletionHandler (bool success, NSError error);
+	[Watch (5,0), iOS (12,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface HKWorkoutBuilder
+	{
```
VincentDondain: Minor: styling, this goes 1 line up (:

VincentDondain: That's repeated everywhere.

src/healthkit.cs
```diff
@@ -2442,4 +2566,225 @@ interface HKWorkoutRouteQuery {
...
+	[NoWatch, iOS (12,0)]
+	[BaseType (typeof (NSObject), Name = "HKFHIRResource")]
+	[DisableDefaultCtor]
+	interface HKFhirResource : NSSecureCoding, NSCopying
+	{
+		[Internal]
+		[Export ("resourceType")]
+		NSString _ResourceType { get; }
+
+		HKFhirResourceType ResourceType { [Wrap ("HKFhirResourceTypeExtensions.GetValue (_ResourceType)")] get; }
```
VincentDondain: Opinion: my personal styling preference for that is to go multilines.

PR #4354, opened 2018-06-27
[CloudKit] Update bindings to Xcode 10 Beta 1, nothing in beta 2

dalexsoto:


src/cloudkit.cs
```diff
@@ -653,16 +669,32 @@ interface CKFetchRecordZoneChangesOperation
 		[Export ("init")]
 		IntPtr Constructor ();
 
+		[Deprecated (PlatformName.WatchOS, 5, 0, message: "Use the 'configurationsByRecordZoneID' overload instead.")]
+		[Deprecated (PlatformName.TvOS, 12, 0, message: "Use the 'configurationsByRecordZoneID' overload instead.")]
+		[Deprecated (PlatformName.iOS, 12, 0, message: "Use the 'configurationsByRecordZoneID' overload instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 14, message: "Use the 'configurationsByRecordZoneID' overload instead.")]
```
spouliot: `configurationsByRecordZoneID` looks like the native, not managed, name

spouliot: Not sure what's best, better ask some native English speakers, @chamons @bradumbaugh ?

We're already using different patterns like
```
"Use the overload that takes 'conversationIdentifiers' instead."
"Use 'DidAuthorizePayment' overload with the 'Action<PKPaymentAuthorizationResult>' parameter instead."
"Use overload accepting an 'UNNotification' parameter."
```

I find them less confusing but I'm not the real target for them :)

bradumbaugh: Hmm... I don't have a strong opinion here, but I'd probably choose the style of the second one from Sebastien's list of options, since it's the most explicit:

`"Use 'DidAuthorizePayment' overload with the 'Action<PKPaymentAuthorizationResult>' parameter instead."`

dalexsoto:
@spouliot fixed, thanks!

Issue #4189, opened 2018-06-05
Xcode 10: list of changes to be reviewed/reverted before final release


rolfbjarne:
- [x] Workaround for rdar://40824310: [[mtouch] Work around build problem with Apple's simd headers in Objective-C++ mode.](https://github.com/xamarin/xamarin-macios/pull/4179/commits/390052703f254c9ebdf6ae5007d5e87bedd915c5) **Fixed: [Use libc++ instead of stdlibc++ when compiling with Xcode 10.](https://github.com/xamarin/xamarin-macios/commit/0d317207702d33cf78cf10d9a56adcda18d68329)**
- [x] Workaround for rdar://40824697: [[mtouch] Work around broken tvOS headers in Xcode 10 beta 1](https://github.com/xamarin/xamarin-macios/pull/4179/commits/663eaeb3898c8a0be87ee0797f96504a01cee740) **Fixed: [Revert "[mtouch] Work around broken tvOS headers in Xcode 10 beta 1."](https://github.com/xamarin/xamarin-macios/pull/4502)**
- [x] monotouch-tests disabled test cases (to review) https://github.com/xamarin/xamarin-macios/pull/4179/commits/07f79ac57d7f9746063ac2303a548c81d3a51ee9
    - [x] AVFoundation/CaptureDeviceTest.cs **Fixed: https://github.com/xamarin/xamarin-macios/pull/4655**
    - [x] AVKit/PlayerViewControllerTest.cs **Fixed: https://github.com/xamarin/xamarin-macios/pull/4656**
    - [x] CoreGraphics/FunctionTest.cs **Fixed: [[CoreGraphics] Fix CGFunction to work after being disposed.](https://github.com/xamarin/xamarin-macios/pull/4644)**
    - [x] CoreImage/FilterTest.cs **Fixed: https://github.com/xamarin/xamarin-macios/pull/4651**
- [x] new core image filters are only stubbed right now - [up to date with beta 4](https://github.com/xamarin/xamarin-macios/pull/4513) (minus open issues)
- [x] some old core image filters were removed - fixed in b4, tests re-enabled with c19710269fbb76c82f6e8eac19f1f0c283cb17fd
- [x] some GameKit selector don't respond anymore - need verification (then radar) **Fixed: [[Gamekit] Re-enable introspection tests in the skipped selectors.](https://github.com/xamarin/xamarin-macios/pull/4682)**
[...]

PR #4359, opened 2018-06-28
[coreml] Update for Xcode 10 beta 2

VincentDondain:


src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+	interface MLArrayBatchProvider : MLBatchProvider {
+
+		[Export ("array")]
+		IMLFeatureProvider[] Array { get; }
+
+		[Export ("initWithFeatureProviderArray:")]
+		IntPtr Constructor (IMLFeatureProvider[] array);
+
+		[Export ("initWithDictionary:error:")]
+		IntPtr Constructor (NSDictionary<NSString, NSArray> dictionary, [NullAllowed] out NSError error);
```
spouliot: since 
* it's not mutable - no setter on `Array`; and
* none of the `.ctor` accept a `null` input 

then it seems unlikely that `init` is allowed - so the type should be decorated with `[DisableDefaultCtor]`

src/coreml.cs
```diff
@@ -191,6 +222,11 @@ interface MLModel {
 		[return: NullAllowed]
 		IMLFeatureProvider GetPrediction (IMLFeatureProvider input, MLPredictionOptions options, out NSError error);
 
+		[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		[Export ("predictionsFromBatch:options:error:")]
+		[return: NullAllowed]
+		IMLBatchProvider GetPredictions (IMLBatchProvider inputBatch, MLPredictionOptions options, [NullAllowed] out NSError error);
```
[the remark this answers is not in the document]
VincentDondain: Minor or I should actually fix that? (:

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+		IMLFeatureProvider GetFeatures (nint index);
+	}
+
+	[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+	[Protocol]
+	interface MLCustomModel {
+
+		[Abstract]
+		[Export ("initWithModelDescription:parameterDictionary:error:")]
+		IntPtr Constructor (MLModelDescription modelDescription, NSDictionary<NSString, NSObject> parameters, [NullAllowed] out NSError error);
```
spouliot: that does not work, we can't enforce constructors in .net interfaces

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+
+	[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+	[BaseType (typeof(NSObject))]
+	interface MLImageSize {
+
+		[Export ("pixelsWide")]
+		nint PixelsWide { get; }
+
+		[Export ("pixelsHigh")]
+		nint PixelsHigh { get; }
```
spouliot: same, it's immutable and without a `init*`
allowing to create a default instance seems incorrect -> `[DisableDefaultCtor]`

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+		MLImageSizeConstraintType Type { get; }
+
+		[Export ("pixelsWideRange")]
+		NSRange PixelsWideRange { get; }
+
+		[Export ("pixelsHighRange")]
+		NSRange PixelsHighRange { get; }
+
+		[Export ("enumeratedImageSizes")]
+		MLImageSize[] EnumeratedImageSizes { get; }
```
spouliot: same, the type is only exposed as a read-only property
so there's no point in allowing it to be created, with `init`, with empty values

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+	interface MLMultiArrayShapeConstraint {
+
+		[Export ("type")]
+		MLMultiArrayShapeConstraintType Type { get; }
+
+		[Export ("sizeRangeForDimension")]
+		NSValue[] SizeRangeForDimension { get; }
+
+		[Export ("enumeratedShapes")]
+		NSArray<NSNumber>[] EnumeratedShapes { get; }
```
spouliot: same -> `[DisableDefaultCtor]`

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+
+		[Export ("stringValues")]
+		string[] StringValues { get; }
+
+		[Static]
+		[Export ("sequenceWithInt64Array:")]
+		MLSequence Create (NSNumber[] int64Values);
+
+		[Export ("int64Values")]
+		NSNumber[] Int64Values { get; }
```
spouliot: the design seems to force people to use static factory method to create instances
No point in allowing a default instance to be created -> `[DisableDefaultCtor]`

src/coreml.cs
```diff
@@ -381,5 +425,131 @@ interface MLCustomLayer {
...
+
+	[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+	[BaseType (typeof(NSObject))]
+	interface MLSequenceConstraint : NSCopying {
+
+		[Export ("valueDescription")]
+		MLFeatureDescription ValueDescription { get; }
+
+		[Export ("countRange")]
+		NSRange CountRange { get; }
```
spouliot: same as before, only exposed thru a read-only property
no point in having a default ctor

src/coreml.cs
VincentDondain: Whoops

VincentDondain:
@spouliot I addressed your feedback, I think the PR is ready to land (:

src/coreml.cs
```diff
@@ -456,25 +457,27 @@ interface MLBatchProvider {
 	}
 
 	[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
-	[Protocol]
+	[BaseType (typeof(NSObject))]
+	[Protocol, Model]
```
spouliot: nice workaround (does not always work but it's good here)
@rolfbjarne I suspect that might run into naming conflict in the future, was your fix only in master (or 15.8) ?

dalexsoto: oh nice catch @spouliot, @VincentDondain  if `[Model (AutoGeneratedName = true)]` is not available please add it here https://github.com/xamarin/xamarin-macios/issues/4189 so we do not forget.

rolfbjarne: The AutoGeneratedName fix is in d15-7 as well.

Issue #4254, opened 2018-06-14
Crash linked to becomeFirstResponder : Failed to marshal the Objective-C object 0x1124818b0 (type: UIView)


Bouncaaa:
### Steps to Reproduce

1. Have  a uipopover and a textfield on the same screen
2. make sure becomeFirstResponder is called on the textfield
3. open the uipopover -> crash.

### Expected Behavior

UIPopover opens without a crash

### Actual Behavior

App crashes immediately after the popover opens with following error:

Failed to marshal the Objective-C object 0x1124818b0 (type: UIView). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'UIKit.UIView&' does not have a constructor that takes one IntPtr argument).

### Environment

```
=== Visual Studio Community 2017 for Mac ===

Version 7.5.2 (build 40)
Installation UUID: f20e1583-d3a8-4309-a11b-11e33062b785
Runtime:
	Mono 5.10.1.57 (2017-12/ea8a24b1bbf) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.1.178 (master / eeaeb7e6)

	Package version: 510010057

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.0.7
	2.0.5
	2.0.0
	1.1.1
	1.0.4
SDK: /usr/local/share/dotnet/sdk/2.1.105/Sdks
SDK Versions:
	2.1.105
	2.1.4
	2.0.0
	1.0.3
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.10.1/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.2
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Apple Developer Tools ===

Xcode 9.4.1 (14161)
Build 9F2000

=== Xamarin.Mac ===

Version: 4.4.1.193 (Visual Studio Community)

=== Xamarin.iOS ===

Version: 11.12.0.4 (Visual Studio Community)
Hash: 64fece5f
Branch: d15-7
Build date: 2018-05-29 20:00:44-0400

=== Xamarin.Android ===

Version: 8.3.3.2 (Visual Studio Community)
[...]

PR #4360, opened 2018-06-28
[registrar] Improve error message when failing to create a managed wrapper for a native handle.

rolfbjarne:
* Convert it into a MM8027/MT8027 error (with documentation).
* Add information about the selector and managed method that triggered the error.

Now the problem reported in issue #4254 shows up as:

> ObjCRuntime.RuntimeException: Failed to marshal the Objective-C object 0x7f8080412810 (type: UIView). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'UIKit.UIView&' does not have a constructor that takes one IntPtr argument).
> Additional information:
> 	Selector: popoverController:willRepositionPopoverToRect:inView:
> 	Method: UIKit.UIPopoverController+_UIPopoverControllerDelegate.WillReposition(UIKit.UIPopoverController, CoreGraphics.CGRect ByRef, UIKit.UIView ByRef)

instead of just:

> ObjCRuntime.RuntimeException: Failed to marshal the Objective-C object 0x7f8080412810 (type: UIView). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'UIKit.UIView&' does not have a constructor that takes one IntPtr argument).

which makes it much easier to understand, track down, and fix/work around,
both for customers and ourselves.


rolfbjarne:
Test failures are known (https://github.com/xamarin/maccore/issues/868)

PR #4367, opened 2018-06-29
[StoreKit] Update bindings to Xcode 10 Beta 1, nothing in beta2

dalexsoto:


src/StoreKit/SKAdNetwork.cs
```diff
@@ -0,0 +1,25 @@
...
+	[Obsolete ("Not usable from tvOS and will be removed in the future.")]
+	[Unavailable (PlatformName.TvOS)]
+	public class SKAdNetwork : NSObject {
+
+		[Obsolete ("Throws a 'NotSupportedException'.")]
+		[Unavailable (PlatformName.TvOS)]
+		public static void RegisterAppForAdNetworkAttribution () => throw new NotSupportedException ();
+	}
+}
+#endif
```
spouliot: minor (no need to update unless other changes are required). We generally create `{prefix}Compat.cs` files for those _hopefully_ short(er)-lived compatibility stubs. It should ease moving to a new profile (git rm) without messing to much with history. It also minimize the number of place to look for them.

dalexsoto: Ahh you are completely right, I forgot about it :( sorry about that.

PR #4278, opened 2018-06-15
[tests] Run Xamarin.Mac tests on Mojave, and add more Xamarin.Mac tests.

rolfbjarne:
* Add more Xamarin.Mac tests: introspection, link sdk, link all and xammac_tests.
* Fix TextureAtlasTest.Empty to not crash due to Apple not liking null callbacks. (#4003)
* Run Xamarin.Mac tests on Mojave as well, even though the build OS is an earlier OS (High Sierra).

PR #4368, opened 2018-06-29
[CoreText] Update bindings to Xcode 10 Beta 1 & 2

dalexsoto:


src/CoreText/CTFontCollection.cs
```diff
@@ -178,6 +178,22 @@ public CTFontDescriptor[] GetMatchingFontDescriptors ()
 			return matches;
 		}
 
+		[Mac (10,7), iOS (12,0), TV (12,0), Watch (5,0)]
+		[DllImport (Constants.CoreTextLibrary)]
+		static extern IntPtr CTFontCollectionCreateMatchingFontDescriptorsWithOptions (IntPtr collection, IntPtr options);
+
+		[Mac (10,7), iOS (12,0), TV (12,0), Watch (5,0)]
+		public CTFontDescriptor [] GetMatchingFontDescriptors (CTFontCollectionOptions options)
```
rolfbjarne: These are manual bindings, so they should have manual tests too.

src/CoreText/CTFramesetter.cs
```diff
@@ -121,6 +121,22 @@ public CGSize SuggestFrameSize (NSRange stringRange, CTFrameAttributes frameAttr
 					constraints, out fitRange);
 		}
 #endregion
+		[Mac (10,14, onlyOn64: true), iOS (12,0), TV (12,0), Watch (5,0)]
+		[DllImport (Constants.CoreTextLibrary)]
+		static extern IntPtr CTFramesetterCreateWithTypesetter (IntPtr typesetter);
+
+		[Mac (10,14, onlyOn64: true), iOS (12,0), TV (12,0), Watch (5,0)]
+		public static CTFramesetter Create (CTTypesetter typesetter)
```
rolfbjarne: Same: tests please :smile:

src/CoreText/CTFontCollection.cs
spouliot: @rolfbjarne can we run the tests on the old-macOS from the PR ?

src/CoreText/CTFontCollection.cs
```diff
@@ -178,6 +178,22 @@ public CTFontDescriptor[] GetMatchingFontDescriptors ()
...
+		{
+			var cfArrayRef = CTFontCollectionCreateMatchingFontDescriptorsWithOptions (handle, options == null ? IntPtr.Zero : options.Dictionary.Handle);
+			if (cfArrayRef == IntPtr.Zero)
+				return new CTFontDescriptor [0];
```
spouliot: `Array.Empty<CTFontDescriptor>`

src/CoreText/CTFontCollection.cs
```diff
@@ -178,6 +178,22 @@ public CTFontDescriptor[] GetMatchingFontDescriptors ()
...
 		}
 
+		[Mac (10,7), iOS (12,0), TV (12,0), Watch (5,0)]
+		[DllImport (Constants.CoreTextLibrary)]
+		static extern IntPtr CTFontCollectionCreateMatchingFontDescriptorsWithOptions (IntPtr collection, IntPtr options);
+
+		[Mac (10,7), iOS (12,0), TV (12,0), Watch (5,0)]
+		public CTFontDescriptor [] GetMatchingFontDescriptors (CTFontCollectionOptions options)
+		{
+			var cfArrayRef = CTFontCollectionCreateMatchingFontDescriptorsWithOptions (handle, options == null ? IntPtr.Zero : options.Dictionary.Handle);
```
spouliot: minor: we have an extension method `GetHandle ()` that does the null check (to return `IntPtr.Zero`)

src/CoreText/CTTypesetter.cs
```diff
@@ -94,6 +71,15 @@ public CTTypesetterOptions (NSDictionary dictionary)
 			get {return Adapter.GetInt32Value (Dictionary, CTTypesetterOptionKey.ForceEmbeddingLevel);}
 			set {Adapter.SetValue (Dictionary, CTTypesetterOptionKey.ForceEmbeddingLevel, value);}
 		}
+
+		[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		public bool AllowUnboundedLayout {
+			get => CFDictionary.GetBooleanValue (Dictionary.Handle, CTTypesetterOptionKey.AllowUnboundedLayout.Handle);
```
spouliot: that will NRE before the fields are available :|
was the `Adapter.*` code above checking for null values ?

src/CoreText/CTFontCollection.cs
rolfbjarne: That doesn't work here because the code is not doing `options.Handle`, but `options.Dictionary.Handle`.

src/CoreText/CTFontCollection.cs
rolfbjarne: @spouliot no, but what I can do is create a new PR that has both the changes from this PR and #4278.

spouliot: k, will catch it at merge time then :)

dalexsoto:
@spouliot / @rolfbjarne  implemented feedback thanks!

src/CoreText/CTTypesetter.cs
dalexsoto: Yes, adapter also has this issue, we'll need to audit this even for StrongDictionaries as we discussed, will check and prepare an RFC

PR #4374, opened 2018-06-29
[gamekit] Update for beta2 (largely deprecations)

spouliot:


src/GameKit/GameKit.cs
```diff
@@ -153,8 +153,17 @@ public enum GKTransportType : long {
...
 
+	[Deprecated (PlatformName.MacOSX, 10,14)]
+	[Deprecated (PlatformName.TvOS, 12,0)]
+	[Deprecated (PlatformName.iOS, 12,0)]
 	[Native]
+#if WATCH && !XAMCORE_4_0
+	// removed in Xcode 10 but a breaking change (for us) to remove
+	[Obsolete ("Not used in watchOS.")]
+#else
+	[Unavailable (PlatformName.WatchOS)]
```
VincentDondain: Mmh why not `[NoWatch]`?

spouliot: `[NoWatch]` is only usable in binding files, i.e. where the generator process it.
`GameKit.cs` is not processed by the generator so it must be the final (long form) attribute

VincentDondain: Oh yea I missed that we were not in a binding file.

src/gamekit.cs
```diff
@@ -2349,6 +2357,9 @@ interface GKTurnBasedEventListener
 
 	[NoWatch]
 	[iOS (10,0)][Mac (10,12)][TV (10,0)]
+	[Deprecated (PlatformName.MacOSX, 10,14, message: "Use 'GKMatchmakerViewController' (real-time) or 'GKTurnBasedMatchmakerViewController' (turn based) instead.")]
```
VincentDondain: Minor: I think you can add a - for "turn-based" similar to "real-time".

PR #4378, opened 2018-07-02
[modelio] Update for Xcode 10 beta 2

VincentDondain:
Renamed and obsoleted 'FromBundle' in favor of 'FromName' for consistency with new API and because I believe it was a mistake anyway.
For instance 'textureNamed:' was 'FromBundle', there's no mention of the bundle there.

### DIFF
--- a/src/modelio.cs
+++ b/src/modelio.cs
@@ -1638,13 +1638,33 @@ namespace ModelIO {
 		[Export ("init")]
 		IntPtr Constructor ();
 
+#if !XAMCORE_4_0
+		[Static]
+		[Obsolete ("Use 'FromName' instead.")]
+		[Wrap ("FromName (name)")]
+		MDLTexture FromBundle (string name);
+#endif
+
 		[Static]
 		[Export ("textureNamed:")]
-		MDLTexture FromBundle (string name);
+		MDLTexture FromName (string name);
+
+#if !XAMCORE_4_0
+		[Static]
+		[Obsolete ("Use 'FromName' instead.")]
+		[Wrap ("FromName (name, bundleOrNil)")]
+		MDLTexture FromBundle (string name, [NullAllowed] NSBundle bundleOrNil);
+#endif
 
 		[Static]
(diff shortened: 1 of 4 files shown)


src/modelio.cs
```diff
@@ -1771,8 +1791,13 @@ interface MDLTextureSampler
 
 	[iOS (9,0), Mac(10,11, onlyOn64 : true)]
 	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor] // designated
 	interface MDLTransform : MDLTransformComponent, NSCopying {
 
+		[DesignatedInitializer]
+		[Export ("init")]
+		IntPtr Constructor ();
```
spouliot: `[DesignatedDefaultCtor]` was added last year

src/modelio.cs
```diff
@@ -1638,13 +1638,33 @@ interface MDLTexture : MDLNamed
...
+#if !XAMCORE_4_0
+		[Static]
+		[Obsolete ("Use 'FromName' instead.")]
+		[Wrap ("FromName (name, bundleOrNil)")]
 		MDLTexture FromBundle (string name, [NullAllowed] NSBundle bundleOrNil);
+#endif
+
+		[Static]
+		[Export ("textureNamed:bundle:")]
+		MDLTexture FromName (string name, [NullAllowed] NSBundle bundleOrNil);
```
rolfbjarne: While I agree that `FromName` sounds more logical than `FromBundle`, neither has a verb.

If we're renaming, my suggestion would be `CreateTexture` or `CreateNamedTexture` (which is also more similar to the `CreateTextureCube` overloads further below).

src/modelio.cs
VincentDondain: Missed it, thanks (:

src/modelio.cs
VincentDondain: I agree, thanks for the suggestion, didn't scroll far enough (;

spouliot:
Nothing new in beta3

PR #4388, opened 2018-07-04
[AVFoundation] Xcode 10 beta 3 support.

mandel-macaque:


src/avfoundation.cs
```diff
@@ -9160,6 +9160,19 @@ interface AVCapturePhotoSettings : NSCopying
...
 		bool EmbedsPortraitEffectsMatteInPhoto { get; set; }
+
+		[Internal]
+		[NoWatch, NoTV, NoMac, iOS (12, 0)]
+		[Export ("availableRawEmbeddedThumbnailPhotoCodecTypes")]
+		NSString[] _GetAvailableRawEmbeddedThumbnailPhotoCodecTypes { get; }
+
+		[NoWatch, NoTV, NoMac, iOS (12, 0)]
+		[Wrap ("Array.ConvertAll (_GetAvailableRawEmbeddedThumbnailPhotoCodecTypes, s => AVVideoCodecTypeExtensions.GetValue (s))")]
+		AVVideoCodecType[] GetAvailableRawEmbeddedThumbnailPhotoCodecTypes { get; }
```
spouliot: remove `Get` prefix from property

src/avfoundation.cs
```diff
@@ -9160,6 +9160,19 @@ interface AVCapturePhotoSettings : NSCopying
...
+
+		[NoWatch, NoTV, NoMac, iOS (12, 0)]
+		[NullAllowed, Export ("rawEmbeddedThumbnailPhotoFormat", ArgumentSemantic.Copy)]
+		NSDictionary RawEmbeddedThumbnailPhotoFormat { get; set; }
```
spouliot: looks like an opportunity for a better (strong typed) return type

spouliot: even if we did not provide it for the non-`Raw` API previously :|

mandel-macaque: @spouliot that is precisely the reason I did not add it. I can do it, no problem.

src/avfoundation.cs
mandel-macaque: @spouliot I'm following the pattern used here: https://github.com/xamarin/xamarin-macios/blob/master/src/avfoundation.cs#L9073

spouliot: bad one, we missed it :(
let's obsolete it and provide a _properly-property-named_ version

src/avfoundation.cs
spouliot: not sure about this one - do we know the format ? might have been the reason it was not done last year

mandel-macaque: I'll take a look and comment here.

src/avfoundation.cs
```diff
@@ -9145,13 +9167,27 @@ interface AVCapturePhotoSettings : NSCopying
...
+		[iOS (11, 0)]
+		[Wrap ("Array.ConvertAll (_GetAvailableEmbeddedThumbnailPhotoCodecTypes, s => AVVideoCodecTypeExtensions.GetValue (s))")]
+		AVVideoCodecType[] AvailableEmbeddedThumbnailPhotoCodecTypes { get; }
+
+#if XAMCORE_4_0
+		[iOS (11, 0)]
+		[NullAllowed, Export ("embeddedThumbnailPhotoFormat", ArgumentSemantic.Copy)]
+		NSDictionary NativeEmbeddedThumbnailPhotoFormat { get; set; }
+
+		[Warap ("NativeEmbeddedThumbnailPhotoFormat")]
```
spouliot: `Warap` won't compile

src/avfoundation.cs
```diff
@@ -9039,6 +9039,28 @@ interface AVCaptureMetadataOutputObjectsDelegate {
...
+		NSString WidthKey { get; }
+
+		[Field ("AVVideoHeightKey")]
+		NSString HeightKey { get; }
+	}
+
+	[NoTV, NoMac, NoWatch, iOS (12,0)]
+	[StrongDictionary ("AVCapturePhotoSettingsThumbnailFormatKeys")]
+	interface AVCapturePhotoSettingsThumbnailFormat {
+		NSString Codec { get; set; }
```
spouliot: are those map to constants ?

src/avfoundation.cs
```diff
@@ -9039,6 +9039,28 @@ interface AVCaptureMetadataOutputObjectsDelegate {
...
+		NSNumber Width { get; set; }
+		NSNumber Height { get; set; }
```
spouliot: What's the _real_ underlying type ? [u]integers ?
IIRC we added support to expose those as value types, c.c. @dalexsoto

src/avfoundation.cs
```diff
@@ -9164,15 +9200,19 @@ interface AVCapturePhotoSettings : NSCopying
...
 		[NoWatch, NoTV, NoMac, iOS (12, 0)]
-		[Wrap ("Array.ConvertAll (_GetAvailableRawEmbeddedThumbnailPhotoCodecTypes, s => AVVideoCodecTypeExtensions.GetValue (s))")]
-		AVVideoCodecType[] GetAvailableRawEmbeddedThumbnailPhotoCodecTypes { get; }
+		[Wrap ("Array.ConvertAll (_AvailableRawEmbeddedThumbnailPhotoCodecTypes, s => AVVideoCodecTypeExtensions.GetValue (s))")]
+		AVVideoCodecType[] AvailableRawEmbeddedThumbnailPhotoCodecTypes { get; }
 
 		[NoWatch, NoTV, NoMac, iOS (12, 0)]
 		[NullAllowed, Export ("rawEmbeddedThumbnailPhotoFormat", ArgumentSemantic.Copy)]
-		NSDictionary RawEmbeddedThumbnailPhotoFormat { get; set; }
+		NSDictionary NativeRawEmbeddedThumbnailPhotoFormat { get; set; }
```
spouliot: note sure about the `Native` prefix...
the current one is `Weak` and, while it can be confusing we better select one that will fit everywhere (i.e. replace it once)

src/avfoundation.cs
dalexsoto: Yes, you can use the following types https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#strong-dictionary-types

src/avfoundation.cs
dalexsoto: If so I do not think we have generator support for smart enums in strong dictionaries  but he might be able to use a Wrap property instead.

src/avfoundation.cs
```diff
@@ -9160,6 +9196,23 @@ interface AVCapturePhotoSettings : NSCopying
 		[NoWatch, NoTV, NoMac, iOS (12, 0)]
 		[Export ("embedsPortraitEffectsMatteInPhoto")]
 		bool EmbedsPortraitEffectsMatteInPhoto { get; set; }
+
+		[Internal]
+		[NoWatch, NoTV, NoMac, iOS (12, 0)]
+		[Export ("availableRawEmbeddedThumbnailPhotoCodecTypes")]
+		NSString[] _AvailableRawEmbeddedThumbnailPhotoCodecTypes { get; }
```
dalexsoto: You might be able to just use `[BindAs (typeof (AVVideoCodecType []))]` here instead of the Wrap

src/avfoundation.cs
mandel-macaque: Ah! Good catch, I compiled the project but did not see the error due to the XAMCORE if statement.

src/avfoundation.cs
mandel-macaque: I'll double check the expected value range. Although the docs were not very clear.

src/avfoundation.cs
mandel-macaque: Hmmmm let me see if I can find the values.

src/avfoundation.cs
mandel-macaque: Weak sounds good. And it is a pattern we use in other places.

src/avfoundation.cs
mandel-macaque: Wrapping won't work. The generator will expect a key for the added property. But I can deal with it via a partial class (I think).

src/avfoundation.cs
mandel-macaque: There is nothing stated in the documentation regarding the value number to add in those fields. I prefer not to take the risk in this case.

src/avfoundation.cs
mandel-macaque: API will stay the same, right? It is a nice to have, so I'll prefer to reduce the number of changes :) I'll use BindAs the next time.

dalexsoto: Kind of, we will actually generate the strong API with the export so not really the same but it is the same for the end user perspective unless you subclass heh

src/avfoundation.cs
```diff
@@ -9145,13 +9168,27 @@ interface AVCapturePhotoSettings : NSCopying
 		[Export ("availableEmbeddedThumbnailPhotoCodecTypes")]
 		NSString[] _GetAvailableEmbeddedThumbnailPhotoCodecTypes { get; }
 
+		[Obsoleted (PlatformName.iOS, 12, 0, message: "Use AvailableEmbeddedThumbnailPhotoCodecTypes instead.")]
```
spouliot: that's us (not Apple) deprecating the API right ?
if so use `[Obsolete ("...")]` and put the API under `#if !XAMCORE_4_0` so it will be removed in a future profile without any additional work on our part

spouliot: and that means the new, correct API should be `virtual` (it's not below), so you need to reverse your `[Wrap]` ( both for `virtual` and to allow it to compile once `XAMCORE_4_0` is defined).

src/avfoundation.cs
spouliot: @mandel-macaque ^ might was well pick that fix up

src/avfoundation.cs
VincentDondain: Also please use '' for 'AvailableEmbeddedThumbnailPhotoCodecTypes'.

PR #4395, opened 2018-07-04
[UIKit] Added support for Xcode 10 beta 3.

mandel-macaque:


src/UIKit/UIDocumentBrowserViewController.cs
```diff
@@ -0,0 +1,53 @@
...
+
+	public partial class UIDocumentBrowserViewController {
+
+		static bool CheckSystemVersion ()
+		{
+#if IOS
+			return UIKit.UIDevice.CurrentDevice.CheckSystemVersion (12, 0);
+#else
+			#error Unknown platform
+#endif
```
spouliot: it's already under a global `#if IOS`

src/UIKit/UIDocumentBrowserViewController.cs
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		public UIDocumentBrowserTransitionController GetTransitionController (NSUrl documentUrl) {
```
spouliot: that's not `virtual` so it's a breaking change

also (minor) `{` on the next line for methods ;-)

src/uikit.cs
```diff
@@ -18268,6 +18274,7 @@ interface IUIDocumentBrowserViewControllerDelegate {}
 	[Protocol, Model]
 	[BaseType (typeof(NSObject))]
 	interface UIDocumentBrowserViewControllerDelegate {
+		[Deprecated (PlatformName.iOS, 12, 0, message: "Use  DidPickDocumentAtUrls (UIDocumentBrowserViewController, NSUrl[]) instead.")]
```
spouliot: API should be between single quote `'`

tests/xtro-sharpie/iOS-UIKit.todo
```diff
@@ -1,3 +0,0 @@
-!missing-protocol-conformance! NSShadow should conform to NSSecureCoding
```
spouliot: if **Intro** did not spot that it's because it's been added (in beta 2) as an exception
please updated `tests/introspection/ApiProtocolTest.cs` to remove it

mandel-macaque: Looking at the diff, it was added in beta 3 => https://github.com/xamarin/xamarin-macios/wiki/UIKit-iOS-xcode10-beta3

src/UIKit/UIDocumentBrowserViewController.cs
mandel-macaque: I know, I did it on purpose :)

If someone just changes the top #if, I want them to be sure that they update the check. To avoid something to be landed that is wrong.

tests/xtro-sharpie/iOS-UIKit.todo
spouliot: @mandel-macaque yes, the header was updated in beta 3, but that was not my point :)

The fact that **intro** (not **xtro**) did not spot it means it was excluded earlier - because it was conforming but **not** in the headers. Now that the headers are updated then that special case, in **Intro**, needs to be removed.

mandel-macaque: Oh, got you.

src/UIKit/UIDocumentBrowserViewController.cs
```diff
@@ -31,7 +31,8 @@
 
 namespace UIKit {
 
-	public partial class UIDocumentBrowserViewController {
+	public partial class UIDocumentBrowserViewController
+	{
```
spouliot: minor: `{` on same line for classes and on next line for methods

src/uikit.cs
spouliot: @mandel-macaque that 👍 did not result in a commit ;-)

src/uikit.cs
```diff
@@ -18285,6 +18292,10 @@ interface UIDocumentBrowserViewControllerDelegate {
 
 		[Export ("documentBrowser:willPresentActivityViewController:")]
 		void WillPresent (UIDocumentBrowserViewController controller, UIActivityViewController activityViewController);
+
+		[iOS (12,0)]
+		[Export ("documentBrowser:didPickDocumentsAtURLs:")]
+		void DidPickDocumentAtUrls (UIDocumentBrowserViewController controller, NSUrl[] documentUrls);
```
dalexsoto: `DidPickDocumentAtUrls` -> `DidPickDocument`

VincentDondain: And it's `DidPickDocuments` with an `s` (:

mandel-macaque: @dalexsoto we talked about this online, DidPickDocumentUrls was deprecated over this new method, which is a rename, but is exactly the same one. Since it is a protocol, I have to create the new method, and the Urls is needed in the name, since is not a document what is pick, but the urls. I'll address the 's' from vincent. Will leave Urls.

PR #4398, opened 2018-07-05
[corevideo] Update for beta 3

spouliot:


spouliot:
@dalexsoto agreed, but it's a breaking change. Most, but not all, of them were changed in `XAMCORE_2_0`

PR #4399, opened 2018-07-05
[macos][appkit] Update for beta 3

spouliot:
Also convert `NSSharingServiceName` to a smart-enum, add some missing
values and adapt helper method to use the smart-enum (future-proofing it)

### DIFF
--- a/src/AppKit/Enums.cs
+++ b/src/AppKit/Enums.cs
@@ -2450,26 +2450,6 @@ namespace AppKit {
 		Full
 	}
 
-	// Convenience enum for string values in ObjC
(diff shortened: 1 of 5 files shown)


dalexsoto:
Looks like there is an extra entry in 'macOS-AppKit.todo' so xtro went insane 🤪 about it

```
?dupe-todo? Entry '!missing-field! NSMenuItemImportFromDeviceIdentifier not bound' in both 'macOS-AppKit.ignore' and 'macOS-AppKit.todo' files
Sanity check failed (1)
```

PR #4405, opened 2018-07-05
[healthkit] Update for Xcode 10 beta 3

VincentDondain:
Somehow some of the APIs added there weren't reported as missing by xtro.

### DIFF
--- a/src/healthkit.cs
+++ b/src/healthkit.cs
@@ -2409,17 +2409,46 @@ namespace HealthKit {
 		[NullAllowed, Export ("endDate")]
 		NSDate EndDate { get; }
 
-		[Deprecated (PlatformName.WatchOS, 3, 0, message: "Use HKWorkoutSession (HKWorkoutConfiguration, out NSError)")]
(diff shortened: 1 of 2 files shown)


dalexsoto:
I did spot those but I assumed they were coming from the beta 1 diff

spouliot:
https://github.com/xamarin/xamarin-macios/wiki/HealthKit-watchOS-xcode10-beta1 (which was not updated) shows those API being added

VincentDondain:
Yea and xtro did not report them (and it's not complaining after I added them). I filed an issue: https://github.com/xamarin/xamarin-macios/issues/4407.

@spouliot I'm not sure why you're requesting changes (:

spouliot:
```
-- (void)startActivity API_AVAILABLE(watchos(5.0));
+- (void)startActivityWithDate:(nullable NSDate *)date API_AVAILABLE(watchos(5.0));
```

For some reason **xtro** did not see the first one/line until it was modified to be the 2nd. It's likely for the same reason that the others were not picked up.

**Reminder** if the **xtro** changes differs from your bindings change then there is something to investigate. Ping on slack if unsure and file an issue if the no _good/known_ reason exists for the difference. Also clean up the diff as you go and commit it (once the binding PR merged) to spot those early.

c.c. @mandel-macaque (for beta1)

PR #4406, opened 2018-07-05
[arkit] Update to Xcode 10 beta 3

VincentDondain:
- Updated some ARReferenceObject APIs based on their (better) Swift names. Breaking changes but on new APIs.
- Update ARReferenceObjectTest for device (center and extent have real values on device).
- Reuploaded an arobject file from Beta 3 just in case because of: "ARReferenceObject and ARWorldMap data generated using iOS 12 beta 2 or earlier isn’t compatible with beta 3 or later. Please rescan your objects to generate new ARReferenceObject and ARWorldMap data."


src/arkit.cs
```diff
@@ -1245,11 +1246,15 @@ interface ARReferenceObject : NSSecureCoding {
 		NSSet<ARReferenceObject> GetReferenceObjects (string resourceGroupName, [NullAllowed] NSBundle bundle);
 
 		[Export ("exportObjectToURL:previewImage:error:")]
-		bool ExportObject (NSUrl url, [NullAllowed] UIImage previewImage, [NullAllowed] out NSError error);
+		bool Export (NSUrl url, [NullAllowed] UIImage previewImage, [NullAllowed] out NSError error);
 
 		[Export ("referenceObjectByApplyingTransform:")]
 		[MarshalDirective (NativePrefix = "xamarin_simd__", Library = "__Internal")]
-		ARReferenceObject GetReferenceObject (Matrix4 transform);
+		ARReferenceObject ApplyingTransform (Matrix4 transform);
```
spouliot: much better, but I think `ApplyTransform` is more .nutty :)

src/arkit.cs
```diff
@@ -1245,11 +1246,15 @@ interface ARReferenceObject : NSSecureCoding {
...
+
+		[Export ("referenceObjectByMergingObject:error:")]
+		[return: NullAllowed]
+		ARReferenceObject Merging (ARReferenceObject @object, [NullAllowed] out NSError error);
```
spouliot: -> `Merge` (again using a *verb* for methods)

src/arkit.cs
VincentDondain: Yeah I was hesitating with that and couldn't really make up my mind (:

PR #4432, opened 2018-07-11
[scenekit] SCNPhysicsShape's Create should take a SCNMatrix4 array, not SCNVector3 [Backport]

dalexsoto:
This is needed by Xcode 10 sample porting


dalexsoto:
This is a backport from master so no mods at this time. Thanks for the review

PR #4439, opened 2018-07-12
[PDFKit] Update bindings to Xcode 10 Beta 1, 2 & 3

dalexsoto:


src/pdfkit.cs
```diff
@@ -508,9 +508,13 @@ interface PdfActionRemoteGoTo {
...
 
 	[iOS (11,0)]
+	[DisableDefaultCtor]
 	[BaseType (typeof (PdfAction), Name = "PDFActionResetForm")]
 	interface PdfActionResetForm {
-		//has a public Init ???
+		// - (instancetype)init NS_DESIGNATED_INITIALIZER;
+		[Export ("init")]
+		[DesignatedInitializer]
+		IntPtr Constructor ();
```
spouliot: @dalexsoto we added a type-level `[DesignatedDefaultCtor]` last year to cover that case

PR #4482, opened 2018-07-20
[Intents] Update bindings to Xcode 10 Beta 4

dalexsoto:


dalexsoto:
Of course, the one I do not build locally  because it’s just an attribute change fails...

PR #4484, opened 2018-07-20
[Photos] Update bindings to Xcode 10 Beta 4

dalexsoto:


dalexsoto:
Unrelated test network failure

src/photos.cs
```diff
@@ -1278,6 +1278,10 @@ interface PHProjectChangeRequest {
 		[Mac (10,14, onlyOn64: true)]
 		[Export ("setProjectPreviewImage:")]
 		void SetProjectPreviewImage (NSImage previewImage);
+
+		[Mac (10,14, onlyOn64: true)]
+		[Export ("removeAssets:")]
+		void RemoveAssets (PHAsset [] /*id<NSFastEnumeration>*/ assets); //a collection of PHAsset objects
```
rolfbjarne: @dalexsoto Are you sure that binding an `id<NSFastEnumeration>` as a managed array works? This looks suspect to me.

dalexsoto: I am very confident that it should work since it expects an `NSObject` that implements `NSFastEnumeration` and our managed array ends up being an `NSArray` which implements said protocol, also docs mentions that it expects `a collection of PHAsset objects`.

I'll try to create a test/sample that exercises said API.

PR #4495, opened 2018-07-24
[msbuild] Don't put gcc/linker flags in the response file, since Mono.Options doesn't support escaping quotes. Fixes vsts#649776.

rolfbjarne:
Mono.Options doesn't (yet) support escaped quotes when parsing response files,
which becomes a problem because escaped quotes are necessary when passing
paths with spaces as gcc/linker flags.

So don't write gcc/linker flags in the response file, and instead pass them as
normal command line arguments, and to be on the safe side, do the same thing
for all extra arguments passed to mmp/mtouch.

Also add tests.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/649776.


jstedfast:
Just for documentation purposes, a long-term solution for this problem in Mono.Options can be found here: https://github.com/mono/mono/pull/9641

PR #4501, opened 2018-07-25
Use libc++ instead of stdlibc++ for Xamarin.Mac too.

rolfbjarne:
See 0d317207702d33cf78cf10d9a56adcda18d68329 for more info.

Also revert a series of commits that bumps min macOS version from 10.7 to 10.9


VincentDondain:
Ah... That wasn't a hardcore fix, had to know where to look at (:

PR #4521, opened 2018-07-27
[arkit] Introduce runWithConfiguration:

VincentDondain:
The comment: "'runWithConfiguration:' selector marked as unavailable in Xcode 9 beta 5"
was *wrongly* referencing 'NS_SWIFT_UNAVAILABLE("Use run(_:options:) instead")'. However that applies to swift and not objective-c, this selector is valid and lets users skip the options.

Note: tested with real-world ARKit app.

### DIFF
--- a/src/arkit.cs
+++ b/src/arkit.cs
@@ -576,7 +576,9 @@ namespace ARKit {
 		[NullAllowed, Export ("configuration", ArgumentSemantic.Copy)]
 		ARConfiguration Configuration { get; }
 
-		// 'runWithConfiguration:' selector marked as unavailable in Xcode 9 beta 5. Use 'Run (ARConfiguration configuration, ARSessionRunOptions options)' instead.
+		[Export ("runWithConfiguration:")]
+		void Run (ARConfiguration configuration);
+
 		[Export ("runWithConfiguration:options:")]
 		void Run (ARConfiguration configuration, ARSessionRunOptions options);
 
--- a/tests/xtro-sharpie/iOS-ARKit.ignore
+++ b/tests/xtro-sharpie/iOS-ARKit.ignore
@@ -1,6 +1,3 @@
-## NS_SWIFT_UNAVAILABLE("Use run(_:options:) instead")
-!missing-selector! ARSession::runWithConfiguration: not bound
-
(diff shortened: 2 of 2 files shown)


dalexsoto:
> Xcode 9 beta 5

Do you come from the future? how is it? 😄 </sarcasm>

VincentDondain:
That's not a new API. It was introduced in Xcode 9 beta 5.

dalexsoto:
haha I am so used to read Xcode 10 now, but yeah makes sense your PR 👍

src/arkit.cs
```diff
@@ -576,7 +576,9 @@ interface ARSession {
 		[NullAllowed, Export ("configuration", ArgumentSemantic.Copy)]
 		ARConfiguration Configuration { get; }
 
-		// 'runWithConfiguration:' selector marked as unavailable in Xcode 9 beta 5. Use 'Run (ARConfiguration configuration, ARSessionRunOptions options)' instead.
+		[Export ("runWithConfiguration:")]
+		void Run (ARConfiguration configuration);
```
spouliot: That's fine since `NS_SWIFT_UNAVAILABLE` is not the same as being unavailable (like in ObjC)

VincentDondain:
@spouliot well IMHO it does solve the issue. You couldn't pass a "null" `ARSessionRunOptions` and now you still can't but you can use the appropriate overload that doesn't have `ARSessionRunOptions` (it's actually one of the suggested solutions).

You want me to add `None = 0,` to `ARSessionRunOptions`? I didn't want to do that as it's not an official value.

spouliot:
@VincentDondain it does not totally solve the issue, just the given example. E.g. if you want to create a method that returns a `ARSessionRunOptions` or add flags to an empty one based on some condition then you need to start from a clean/empty state. 

Swift has this built-in by using the type name (a bit like `new ARSessionRunOptions ()` but with a few less characters) and [Obj]C will accept integral values (ugly but shorter).

In C#/.net there's no good reason for flag-enums not to define `0`, see https://msdn.microsoft.com/en-us/library/ms182149.aspx

VincentDondain:
Ah, interesting article!
Ok I'll add `None`, thanks for the feedback.

PR #4522, opened 2018-07-27
[msbuild] Make PropertyListEditor.Value a property like it should be

jstedfast:


lassana:
Hi everyone,
Which version will include this PR?

rolfbjarne:
@lassana this fix will be included in our `d15-9` release, which is currently in the Alpha and Beta channels.

PR #4523, opened 2018-07-27
[arkit] Add ARAnchorCopying protocol

VincentDondain:
Even if empty this protocol will help with tests and subclasses.


VincentDondain:
This PR fails on device with the following error:

```
/Users/vidondai/Documents/xi-xcode10/xamarin-macios/tests/introspection/iOS/obj/iPhone/Debug-unified/device-builds/iphone10.6-12.0/mtouch-cache/registrar.m(50381,4): error G464A8611: cannot find protocol declaration for 'ARAnchorCopying'
MTOUCH : error MT4109: Failed to compile the generated registrar code. Please file a bug report at http://bugzilla.xamarin.com
```

Let's brainstorm here on how to solve this (:

rolfbjarne:
Does a simple Objective-C sample using `ARAnchorCopying` compile? If it does, what's the significant difference between the generated registrar code and the sample code?

spouliot:
@VincentDondain the bots failures are:

`[FAIL] [Rule 3] Missing '.' in attribute's message: "To conform to 'ARAnchorCopying' you need to implement: `

and xtro

```
19:40:23.7417620 ?unknown-entry? !missing-protocol! ARAnchorCopying not bound in 'iOS-ARKit.ignore'
19:40:23.7418430 ?unknown-entry? !missing-protocol-conformance! ARAnchor should conform to ARAnchorCopying in 'iOS-ARKit.ignore'
```

As you have not enabled device builds we can't see any details (and you did not link to a gist of the registrar generated code)

VincentDondain:
Registrar generated code: https://gist.github.com/VincentDondain/80c3262ae2b135336fe77d9d2bcedd95

spouliot:
build

spouliot:
@VincentDondain not sure if it's beta5 related (no related header changes) but applying your change locally worked for device builds. Is it possible you were building with an older Xcode10 beta ?

Adding label to re-build (with beta 4) with device builds to see if bots have the issue.

VincentDondain:
It is possible. I switched from my laptop to desktop (after working remotely) and I installed beta 4 on the desktop but maybe I was out of sync there. Will retry now that I'm up to date with beta 5 (and did a full rebuild).

VincentDondain:
Ah stupid me, it's because I was trying to run the tests from Visual Studio Mac and didn't not have the right Xcode selected there (was beta 2 on this computer).

Now that's an annoying problem people can face, they update their Xcode, update their XI but forget to change the Xcode path in VSMac. The only way I can see to fix that would be to add the Xcode version code in our versions.plist and warn that customers are not using the exact right recommended Xcode.

spouliot:
@VincentDondain I don't think it's common - besides us working with several Xcode versions and betas. Most people only use the default location and, for a very few, a single beta location. 

Also a specific version check would affect (people hate warnings they can't solve) everyone using a newer version for which we don't update, eg. Xcode 9.4.1

FWIW that should not happen, even to us, very often since we do have such a check in our build system (so rebuilding with that fix should have warned/stopped you).

VincentDondain:
@spouliot mmh not sure which check in our build system you're talking about. I think you're talking about the `xcode-select` check, that check I hit it and I fixed my `xcode-select` path but our build system doesn't force you to update your VSMac Xcode path (which was my issue here).

An other thing we could do (not sure it's possible) is to use the `xcode-select` path in VSMac. In Xcode when you switch the command line tool in the IDE you effectively do an `xcode-select -s`.

screenshot 2018-07-31 11 49 57

VincentDondain:
@spouliot yes I'm now working on that. I initially wanted to push this because I ran into an issue locally (:

PR #4526, opened 2018-07-30
[tests] Change the MT4134 test to only expect device builds to fail.

rolfbjarne:
As an unintended side effect of 215ab7fc1aa2f2fd64e944a8115d300630f6f32b, we
stopped reporting MT4134 errors ("Your app is using framework X, which means
you must update your Xcode") for simulator builds.

This can be either good (people's simulator builds now succeed when the
previously didn't) or bad (people's simulator builds don't always match their
device builds, since they may still get the MT4134 error for device builds).

This patch assumes we want the improved simulator builds, and adjusts the
corresponding test accordingly.


rolfbjarne:
The alternative fix for this issue (the unindented side effect of 215ab7fc1aa2f2fd64e944a8115d300630f6f32b) is to fix the side effect to get back the previous behavior.

PR #4527, opened 2018-07-30
Update for xcode 10 beta 5

spouliot:


tests/xtro-sharpie/watchOS-UIKit.todo
```diff
@@ -0,0 +1,2 @@
+!missing-protocol-conformance! NSAttributedString should conform to NSItemProviderReading (defined in 'NSAttributedString_ItemProvider' category)
```
VincentDondain: Well it's `UINSItemProvider` > `NSAttributedString_ItemProvider`, they only changed the category. Arguably `missing-protocol-conformance` shouldn't print "defined in ...".

I don't think we should move that to `.todo` we should update the `.ignore`. Or we move all the `missing-protocol-conformance` for NSItemProviderReading/Writing in the todo.

tests/xtro-sharpie/macOS-StoreKit.ignore
```diff
@@ -1,7 +1,3 @@
 # Fixed in XAMCORE_4_0
 !unknown-native-enum! SKCloudServiceAuthorizationStatus bound
 !unknown-native-enum! SKCloudServiceCapability bound
-
-# Breaking change by Apple in Xcode 10, so we don't follow suit.
-# https://trello.com/c/A34S0kLY/125-41782055-skproductdownloadable-crashes-on-macos-1011-when-compiled-with-xcode-10
```
VincentDondain: Minor but I would have also updated the binding file: https://github.com/xamarin/xamarin-macios/pull/4528/files#diff-e9aebf3ee5f3c844607eb2da9d2feb14L218 (I updated the Trello card).

tests/xtro-sharpie/watchOS-UIKit.todo
spouliot: It does not matter, everything in the `.todo` will be processed before releasing.
What matters is to have a green bump PR since it holds further beta 5 PR updating the various frameworks.

PR #4534, opened 2018-07-31
[coreml] Update for Xcode 10 beta 5

VincentDondain:


src/coreml.cs
```diff
@@ -74,6 +74,13 @@ public enum MLMultiArrayShapeConstraintType : long {
 		Range = 3,
 	}
 
+	[Watch (5,0), TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+	[Native]
+	public enum MLComputeUnits : long {
+		Only = 0,
+		AndGpu = 1,
```
spouliot: +    MLComputeUnitsCPUOnly = 0,
+    MLComputeUnitsCPUAndGPU = 1

you're missing the `Cpu` prefix (which sharpie likely removed since it was common, but it's not in the type name)

VincentDondain: Ah nice catch!

PR #4537, opened 2018-07-31
[passkit] Update for beta 5

spouliot:


tests/introspection/ApiTypoTest.cs
```diff
@@ -200,6 +200,7 @@ protected ApiTypoTest ()
 			"Embd",
 			"Enc",
 			"Eppc",
+			"Eftpos", // Electronic funds transfer at point of sale
```
VincentDondain: :trollface: what a good acronym!

PR #4542, opened 2018-07-31
[homekit] Remove export for 'RemoveUser' (iOS prohibited)

VincentDondain:
We want to remove the native API because it's prohibited and Apple could reject apps with it. Therefore we exposed empty stubs to avoid breaking the API.

### DIFF
--- a/src/HomeKit/HMHome.cs
+++ b/src/HomeKit/HMHome.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using ObjCRuntime;
 using Foundation;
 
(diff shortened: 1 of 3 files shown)


src/HomeKit/HMHome.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
```
spouliot: ?

src/HomeKit/HMHome.cs
```diff
@@ -80,5 +81,21 @@ public partial class HMHome
 
 			return GetServices (arr.ToArray ());
 		}
+
+		[NoTV]
+		[NoWatch]
+		[Introduced (PlatformName.iOS, 8,0, PlatformArchitecture.All, message: "This API in now prohibited on iOS. Use 'ManageUsers' instead.")]
```
spouliot: likely minor (ot sure if/how affects the display of the message): no need to duplicate the message in `[Introduced]`

src/HomeKit/HMHome.cs
```diff
@@ -80,5 +81,21 @@ public partial class HMHome
...
+		[Obsoleted (PlatformName.iOS, 9,0, PlatformArchitecture.All, message: "This API in now prohibited on iOS. Use 'ManageUsers' instead.")]
+		public virtual void RemoveUser (HMUser user, Action<NSError> completion) {
```
rolfbjarne: Style: opening brace for functions is on the next line.

src/HomeKit/HMHome.cs
```diff
@@ -80,5 +81,21 @@ public partial class HMHome
...
+			throw new NotSupportedException ();
+		}
+
+		[NoTV]
+		[NoWatch]
+		[Introduced (PlatformName.iOS, 8,0, PlatformArchitecture.All, message: "This API in now prohibited on iOS. Use 'ManageUsers' instead.")]
+		[Obsoleted (PlatformName.iOS, 9,0, PlatformArchitecture.All, message: "This API in now prohibited on iOS. Use 'ManageUsers' instead.")]
+		public virtual Task RemoveUserAsync (HMUser user) {
```
rolfbjarne: Style: opening brace for functions is on the next line.

src/HomeKit/HMHome.cs
VincentDondain: https://github.com/xamarin/xamarin-macios/pull/4542/files#diff-9916e0779d658478980c68f22131ae6fR97

`public virtual Task RemoveUserAsync (HMUser user)`

PR #4544, opened 2018-08-01
[iads] Update for beta 5

spouliot:
Apple removed `MPMoviePlayerController_iAdPreroll.h` in beta 5.

The deprecation existed, but indirectly, on `MPMoviePlayerController`,
the type on which the category is based. IOW it's became obsoleted so
we're adding the attributes to match this.

### DIFF
--- a/src/iad.cs
+++ b/src/iad.cs
@@ -170,6 +170,8 @@ namespace iAd {
 
 	[Category (allowStaticMembers: true)] // Classic isn't internal so we need this
 	[BaseType (typeof (MPMoviePlayerController))]
+	[Deprecated (PlatformName.iOS, 9,0, message: "Use 'iAdPreroll_AVPlayerViewController' instead.")]
+	[Obsoleted (PlatformName.iOS, 12,0)] // header removed in xcode10 beta5
 	partial interface IAdPreroll {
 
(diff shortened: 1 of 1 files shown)


src/iad.cs
```diff
@@ -170,6 +170,8 @@ interface ADInterstitialAdDelegate {
 
 	[Category (allowStaticMembers: true)] // Classic isn't internal so we need this
 	[BaseType (typeof (MPMoviePlayerController))]
+	[Deprecated (PlatformName.iOS, 9,0, message: "Use 'iAdPreroll_AVPlayerViewController' instead.")]
+	[Obsoleted (PlatformName.iOS, 12,0)] // header removed in xcode10 beta5
```
VincentDondain: You can have messages for obsoleted. Not sure we want to talk about header removal but we could say that we don't expect the API to work since Xcode 10 anymore.

spouliot: Not quite sure what (or how) to say - an ambiguous message probably won't help. It _works_ today and will probably continue to work because of existing applications (will it continue to show ads is another story).

The fact that Apple removed the API is pretty much the "default" meaning for obsoleted and the solution is covered in the deprecated message. Any better ideas ?

PR #4547, opened 2018-08-01
[CoreGraphics] Add missing API added in Xcode 10 beta 1.

rolfbjarne:


src/CoreGraphics/CGPDFArray.cs
```diff
@@ -142,5 +142,41 @@ public bool GetString (int idx, out string result)
...
+		}
+
+		public delegate bool ApplyBlockCallback (nint index, CGPDFObject value, IntPtr info);
+
+		[DllImport (Constants.CoreGraphicsLibrary)]
+		[iOS (12, 0)][Mac (10, 14)][TV (12, 0)][Watch (5, 0)]
+		extern static bool CGPDFArrayApplyBlock (/* CGPDFArrayRef */ IntPtr array, /* CGPDFArrayApplierBlock */ ref BlockLiteral block, /* void* */ IntPtr info);
+
+		[iOS (12, 0)][Mac (10, 14)][TV (12, 0)][Watch (5, 0)]
+		[BindingImpl (BindingImplOptions.Optimizable)]
```
spouliot: Only semi-unrelated to PR but it seems we could/should ensure/test `BindingImplOptions.Optimizable` works (when `BindingImplOptions.GeneratedCode` is not set) inside one of our test suite (MTouch?). Otherwise we risk finding out it's not _optimizable_ at build time (and that coverage will vary depending if the linker keeps, or not, that manually bound method).

src/CoreGraphics/CGPDFArray.cs
```diff
@@ -142,5 +142,41 @@ public bool GetString (int idx, out string result)
...
+		public bool ApplyBlock (ApplyBlockCallback callback, IntPtr info = default (IntPtr))
```
spouliot: having an `IntPtr` makes it harder to pass state to the callback, than an `System.Object`

rolfbjarne: That's a good point, we could wrap an `object` with a GCHandle.

src/CoreGraphics/CGPDFArray.cs
spouliot: I'm worried about the non-generated (manual) code, like this one, that the linker can remove (if unused) when running monotouch-test (as @chamons note there's no tests for this API so the linker will remove it). Once removed nothing checks if that manual code is (or not) optimizable right ? (I might be forgetting a different test that covers it)

rolfbjarne: I've just added tests for this.

PR #4559, opened 2018-08-02
[AudioToolbox] Add Xcode 10 beta 5 support.

mandel-macaque:


tests/xtro-sharpie/tvOS-AudioToolbox.ignore
```diff
@@ -0,0 +1,2 @@
+# already obsoleted and pointed to use ReadPacketData
+!unknown-pinvoke! AudioFileReadPackets bound
```
spouliot: that might still be a problem because of bitcode and the fact the symbol might not be present anymore

Note: the `[Obsolete]` was not in `XAMCORE_2_0` so it's not in `Xamarin.TVOS.dll`, except that the p/invoke is not under the same condition (and bitcode -> direct call -> symbol presence ).

Considering that (managed) linking should always remove it (it's never unused) then it's not critical and can be fixed in a separate PR.

mandel-macaque: I'll make a diff PR with that fix then.

PR #4564, opened 2018-08-03
[CoreImage] Add Xcode 10 beta 5 support.

mandel-macaque:


src/coreimage.cs
```diff
@@ -1841,6 +1880,62 @@ interface CIImage : NSSecureCoding, NSCopying {
...
+		CIImage CreateByInsertingIntermediate (bool cache);
+
+		// CIImage_AVPortraitEffectsMatte category
+
+		[TV (12, 0), iOS (12, 0), Mac (10, 14, onlyOn64: true)]
+		[NullAllowed, Export ("portraitEffectsMatte")]
+		AVPortraitEffectsMatte PortraitEffectsMatte { get; }
+
+		[TV (12,0), iOS (12,0), Mac (10, 14, onlyOn64: true)]
+		[Export ("initWithPortaitEffectsMatte:options:")] // selector typo, rdar filled 42894821
```
spouliot: 👍 for rdar

src/coreimage.cs
```diff
@@ -1841,6 +1880,62 @@ interface CIImage : NSSecureCoding, NSCopying {
...
+
+		[TV (11,0), iOS (11,0), Mac (10, 14, onlyOn64: true)]
+		[Export ("initWithPortaitEffectsMatte:")] // selector typo, rdar filled 42894821
+		IntPtr Constructor (AVPortraitEffectsMatte matte);
+
+		[TV (12,0), iOS (12,0), Mac (10, 14, onlyOn64: true)]
+		[Static]
+		[Export ("imageWithPortaitEffectsMatte:options:")] // selector typo, rdar filled 42894821
+		[return: NullAllowed]
+		CIImage FromPortraitEffectsMatte (AVPortraitEffectsMatte matte, [NullAllowed] NSDictionary<NSString, NSObject> options);
```
spouliot: use the non-generic `NSDictionary` since that signature is not very helpful and it will make it harder to create strong dictionaries in the future

src/coreimage.cs
```diff
@@ -1841,6 +1880,62 @@ interface CIImage : NSSecureCoding, NSCopying {
...
+		[Static]
+		[Export ("imageWithPortaitEffectsMatte:")] // selector typo, rdar filled 42894821
+		[return: NullAllowed]
+		CIImage FromPortraitEffectsMatte (AVPortraitEffectsMatte matte);
+
+		// CIImage_AVDepthData category
+
+		[TV (11, 0), iOS (11, 0), Mac (10,13)]
+		[Export ("initWithDepthData:options:")]
+		IntPtr Constructor (AVDepthData data, [NullAllowed] NSDictionary<NSString, NSObject> options);
```
spouliot: same

src/coreimage.cs
```diff
@@ -1841,6 +1880,62 @@ interface CIImage : NSSecureCoding, NSCopying {
...
+
+		[TV (11, 0), iOS (11, 0), Mac (10,13)]
+		[Export ("initWithDepthData:")]
+		IntPtr Constructor (AVDepthData data);
+
+		[TV (11, 0), iOS (11, 0), Mac (10,13)]
+		[Static]
+		[Export ("imageWithDepthData:options:")]
+		[return: NullAllowed]
+		CIImage FromDepthData (AVDepthData data, [NullAllowed] NSDictionary<NSString, NSObject> options);
```
spouliot: same

src/coreimage.cs
```diff
@@ -5409,7 +5512,11 @@ interface CIImageRepresentationOptions {
 		bool DepthImage { get; set; }
 
 		bool DisparityImage { get; set; }
+
+		bool PortraitEffectsMatteImage { get; set}
```
spouliot: missing `;` I thought that would be a compilation error

PR #4567, opened 2018-08-03
[CoreMedia] Add Xcode 10 beta 5 support.

mandel-macaque:


src/CoreMedia/CMTime.cs
```diff
@@ -308,6 +308,16 @@ public static CMTime GetMinimum (CMTime time1, CMTime time2)
 			return CMTimeMinimum (time1, time2);
 		}
 
+		[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		[DllImport (Constants.CoreMediaLibrary)]
+		extern static CMTime CMTimeFoldIntoRange (CMTime time, CMTimeRange foldRange);
+
+		[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		public static CMTime FoldIntoRange (CMTime time, CMTimeRange foldRange)
```
spouliot: I'm not sure if the `IntoRange` is needed as `Fold` (verb) is already describing the action and swift is using
```
+- Name: CMTimeFoldIntoRange
+  SwiftName: "CMTime.fold(self:intoRange:)"
```

tests/xtro-sharpie/macOS-CoreMedia.ignore
```diff
@@ -3,3 +3,11 @@
 !missing-pinvoke! CMAudioDeviceClockGetAudioDevice is not bound
 !missing-pinvoke! CMAudioDeviceClockSetAudioDeviceID is not bound
 !missing-pinvoke! CMAudioDeviceClockSetAudioDeviceUID is not bound
+
+# as per headers: API_AVAILABLE(ios(9.0))
```
spouliot: It's unclear wrt docs 
https://developer.apple.com/documentation/coremedia/kcmsamplebufferlensstabilizationinfo_active?language=objc
but it's safer to exclude them until we're sure 👍

PR #4590, opened 2018-08-07
[coreservices] Update up to beta 5

spouliot:
Also avoid creating NSString instances for hidden constants where
only the handle is every used.


mandel-macaque:
@spouliot xtro complains about kCFHTTPAuthenticationSchemeOAuth1, should that be added in the ignore?

spouliot:
@mandel-macaque yup, forgot to add the new file :|

PR #4593, opened 2018-08-08
Bump XI/XM versions after 15.9 branch

spouliot:


spouliot:
API diff -> version change

PR #4601, opened 2018-08-08
[msbuild] Update 'metal' binary path for Xcode 10

### LINKED ISSUE #4576: [xcode10] 'Metal Game' fails to build
### Steps to Reproduce

1. Open VSfM
2. Create macOS `Metal Game` app
3. Build it

OR

1. Download a [MetalKitEssentials sample](https://github.com/xamarin/mac-ios-samples/tree/master/MetalKitEssentials) 
2. Open and build it

### Expected Behavior
Build is successful

### Actual Behavior
Build is failed

```
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Mac/Xamarin.Mac.Common.targets(3,3): Error MSB6003: The specified task executable "metal" could not be run. ApplicationName='metal', CommandLine=' -arch air64 -emit-llvm -c -gline-tables-only -ffast-math -std=osx-metal1.0 -serialize-diagnostics obj/x86/Debug/metal/Shaders.dia -o obj/x86/Debug/metal/Shaders.air Shaders.metal', CurrentDirectory='', Native error= Cannot find the specified file (MSB6003) (Test)
```

### Environment

1. VSfM 7.6 Preview (7.6 build 2135)
https://gist.github.com/mykyta-bondarenko-gl/ef16ac39ec37d627700a39960f179d05

### Build Logs

1. VSfM output
https://gist.github.com/mykyta-bondarenko-gl/0ac2f4b6f194b6f548ccad9cf0267d08
2. msbuild output
https://gist.github.com/mykyta-bondarenko-gl/0b7d79dbb3e6461c12edc34f2814d97f

### Example Project 
- [Example](https://github.com/xamarin/xamarin-macios/files/2262136/Metal_Game_Sample.zip)
- [MetalKitEssentials](https://github.com/xamarin/mac-ios-samples/tree/master/MetalKitEssentials)

VincentDondain:
- Fixes #4576: [xcode10] 'Metal Game' fails to build. (https://github.com/xamarin/xamarin-macios/issues/4576)

In Xcode 10 Apple moved the "metal" binary from `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/usr/bin/metal` to `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/metal`.

### DIFF
--- a/tests/mtouch/ToolTasksBinPathTest.cs
+++ b/tests/mtouch/ToolTasksBinPathTest.cs
@@ -58,50 +58,36 @@ namespace Xamarin.Mac.Tasks {
 	[TestFixture]
 	public class ToolTasksBinPathTest {
 
-		[SetUp]
-		public void Setup ()
-		{
-			Environment.SetEnvironmentVariable ("DEVELOPER_DIR", Configuration.xcode_root);
-		}
-
 		[Test]
 		public void MetalBinPathTest ()
 		{
 			var metalTask = new MetalPoker ();
-
-			var psi = new ProcessStartInfo ("xcrun") {
-				Arguments = "-f metal",
-				UseShellExecute = false,
-				CreateNoWindow = true,
-				RedirectStandardOutput = true,
-				RedirectStandardError = true,
-			};
-			var proc = Process.Start (psi);
-
-			string output = proc.StandardOutput.ReadToEnd ();
-			string err = proc.StandardError.ReadToEnd ();
-
-			Assert.True (output.Contains (metalTask.DevicePlatformBinDir), err);
+			CheckToolBinDir ("metal", metalTask.DevicePlatformBinDir);
 		}
 
 		[Test]
 		public void MetalLibBinPathTest ()
 		{
 			var metalLibTask = new MetalLibPoker ();
+			CheckToolBinDir ("metallib", metalLibTask.DevicePlatformBinDir);
+		}
 
+		public void CheckToolBinDir (string taskName, string binDirToCheck)
+		{
 			var psi = new ProcessStartInfo ("xcrun") {
-				Arguments = "-f metallib",
+				Arguments = $"-f {taskName}",
 				UseShellExecute = false,
 				CreateNoWindow = true,
 				RedirectStandardOutput = true,
 				RedirectStandardError = true,
 			};
+			psi.EnvironmentVariables.Add ("DEVELOPER_DIR", Configuration.xcode_root);
 			var proc = Process.Start (psi);
 
 			string output = proc.StandardOutput.ReadToEnd ();
 			string err = proc.StandardError.ReadToEnd ();
 
-			Assert.True (output.Contains (metalLibTask.DevicePlatformBinDir), err);
+			Assert.True (output.Contains (binDirToCheck), err);
 		}
 	}
 }


VincentDondain:
Ah! Nice nice nice.

Yeah about the unit test..., it's not trivial to add msbuild mac unit tests (it is for iOS...)

spouliot:
@VincentDondain it does not have to be an MSBuild unit test. A normal one (e.g. mtouch tests) can verify that condition and (ideally) embed some, not all, of the logic (to avoid code duplication).

VincentDondain:
Oh wow even better than `xcrun -f` I offer you `xcrun -r` where you can execute whatever tool with arguments. So effectively we should replace all our direct tools paths with `xcrun`.

spouliot:
@VincentDondain you don't need `-r` since it's the default behaviour. However you need to make sure `xcrun` is using the _right_ Xcode, e.g. setting `DEVELOPER_DIR`, or you won't get the right/expected tool.

As for the test(s) they should be added unless `xcrun` is used. That's what I meant by _at least_ - if we need our own logic (not the first time we talked about `xcrun` and I don't recall all the ins-and-outs) then we should have (xcrun-based) tests to ensure that logic remains _true_.

VincentDondain:
> you don't need -r since it's the default behavior

Not sure where you get that from. `-f` just prints the path.

E.g:

```
xcrun -f actool --errors --warnings --notices --output-format xml1 --output-partial-info-plist /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/partial-info.plist --app-icon AppIcon --minimum-deployment-target 10.12 --platform macosx --compile /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/bundle /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/cloned-assets/Assets.xcassets
xcrun: error: unrecognized option: --errors
```

However if you want `xcrun` to find + execute the found binary then `-r` is your friend:

```
xcrun -r actool --errors --warnings --notices --output-format xml1 --output-partial-info-plist /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/partial-info.plist --app-icon AppIcon --minimum-deployment-target 10.12 --platform macosx --compile /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/bundle /Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/cloned-assets/Assets.xcassets
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>com.apple.actool.compilation-results</key>
	<dict>
		<key>output-files</key>
		<array>
			<string>/Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/bundle/AppIcon.icns</string>
			<string>/Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/bundle/Assets.car</string>
			<string>/Users/vidondai/Projects/MacMetal/MacMetal/obj/x86/Debug/actool/partial-info.plist</string>
		</array>
	</dict>
</dict>
</plist>
```

VincentDondain:
I mean yes `-f` is what would be useful for the tests. But `-r` is what's useful long term if we want all the tasks to be updated to not deal with paths.

(maybe that's the confusion in this discussion).

VincentDondain:
Ah ok got it. I thought you meant I should forget `-r` because it's the default behavior for `-f` (to also execute). #communication

tests/mtouch/ToolTasksBinPathTest.cs
```diff
@@ -0,0 +1,105 @@
...
+			var psi = new ProcessStartInfo ("xcrun") {
+				Arguments = "-f metal",
+				UseShellExecute = false,
+				CreateNoWindow = true,
+				RedirectStandardOutput = true,
+			};
+			var proc = Process.Start (psi);
+
+			string output = proc.StandardOutput.ReadToEnd ();
+			Console.WriteLine (output); // Print the output in case something wrong happens when executing the command
```
mandel-macaque: Do we get anything in StandardError?

tests/mtouch/ToolTasksBinPathTest.cs
```diff
@@ -0,0 +1,105 @@
...
+			var psi = new ProcessStartInfo ("xcrun") {
+				Arguments = "-f metallib",
+				UseShellExecute = false,
+				CreateNoWindow = true,
+				RedirectStandardOutput = true,
+			};
+			var proc = Process.Start (psi);
+
+			string output = proc.StandardOutput.ReadToEnd ();
+			Console.WriteLine (output); // Print the output in case something wrong happens when executing the command
```
mandel-macaque: Same stderr comment

tests/mtouch/ToolTasksBinPathTest.cs
VincentDondain: Probably but I'm not sure it's critical to get it. The standard output tells you what happened. The test will fail also.

VincentDondain:
@mandel-macaque @spouliot please re-review (:

tests/mtouch/ToolTasksBinPathTest.cs
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+
+	[TestFixture]
+	public class ToolTasksBinPathTest {
+
+		[SetUp]
+		public void Setup ()
+		{
+			Environment.SetEnvironmentVariable ("DEVELOPER_DIR", Configuration.xcode_root);
```
spouliot: that change the execution environment, i.e. it could affect other (existing or future) tests randomly (since the execution order is not guaranteed) and that will be hard to diagnose

To avoid this you can move this into a 
```
var dd = getenv ("DEVELOPER_DIR");
try {
   setenv ("DEVELOPER_DIR", Configuration.xcode_root);
   xcrun (...);
}
finally {
   setenv ("DEVELOPER_DIR", dd);
}
```

tests/mtouch/ToolTasksBinPathTest.cs
```diff
@@ -0,0 +1,107 @@
...
+				Arguments = "-f metal",
+				UseShellExecute = false,
+				CreateNoWindow = true,
+				RedirectStandardOutput = true,
+				RedirectStandardError = true,
+			};
+			var proc = Process.Start (psi);
+
+			string output = proc.StandardOutput.ReadToEnd ();
+			string err = proc.StandardError.ReadToEnd ();
```
spouliot: a large part of the above could be shared across both (likely more in the future) tests - in addition to the try/catch

tests/mtouch/ToolTasksBinPathTest.cs
rolfbjarne: Or you could set the environment variable in the `ProcessStartInfo` instead, then you wouldn't need any cleanup.

VincentDondain:
@rolfbjarne so the idea if that this PR contains the simplest fix (call updated metal/lib path) and we'll revisit all the msbuild tasks to make them all use `xcrun` (a bit more work). https://github.com/xamarin/xamarin-macios/issues/4634

spouliot:
@rolfbjarne the test is just a canary in case the location changes (again) before the tasks are updated (or in case they can't be)

PR #4604, opened 2018-08-09
[Generator] Ensure the correct error is raised when an Export is missing on a Property.

mandel-macaque:
The generator had a small bug in which we would get a NRE when a
property missed a Export attribute and had no WrapAttr. The issue is due
to the fact that an || is being used and does not shortcut when the
attribute is missing. In that case, the first Get check would pass and a
second attempt would happen with the set. In the case the set is missing,
we would get a NRE. The correct way is to ensure that we do have the get
and the set BEFORE the attr is checked.

The added test shows an exaple of the issue.

### DIFF
--- a/src/generator.cs
+++ b/src/generator.cs
@@ -2300,7 +2300,10 @@ public partial class Generator : IMemberGatherer {
 					if (AttributeManager.HasAttribute<CoreImageFilterPropertyAttribute> (pi))
 						continue;
 
-					if (AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ()) || AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ()))
+					// we ensure that we do not get a NRE in the case we forgot and ExportAttr in a property
+					var hasWrapGet = pi.GetGetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ());
+					var hasWrapSet = pi.GetSetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ());
+					if (hasWrapGet || hasWrapSet)
 						continue;
 
 					throw new BindingException (1018, true, "No [Export] attribute on property {0}.{1}", t.FullName, pi.Name);
--- a/src/generator.cs
+++ b/src/generator.cs
@@ -2300,9 +2300,11 @@ public partial class Generator : IMemberGatherer {
 					if (AttributeManager.HasAttribute<CoreImageFilterPropertyAttribute> (pi))
 						continue;
 
-					// we ensure that we do not get a NRE in the case we forgot and ExportAttr in a property
-					var hasWrapGet = pi.GetGetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ());
-					var hasWrapSet = pi.GetSetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ());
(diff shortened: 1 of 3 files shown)


src/generator.cs
```diff
@@ -2300,7 +2300,10 @@ public void Go ()
 					if (AttributeManager.HasAttribute<CoreImageFilterPropertyAttribute> (pi))
 						continue;
 
-					if (AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ()) || AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ()))
+					// we ensure that we do not get a NRE in the case we forgot and ExportAttr in a property
+					var hasWrapGet = pi.GetGetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ());
+					var hasWrapSet = pi.GetSetMethod () != null && AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ());
```
mandel-macaque: This two could have been added in the if, but made the code 'ugly' and hard to understand. I think it is nicer to be readable.

src/generator.cs
```diff
@@ -2300,7 +2300,10 @@ public void Go ()
 					if (AttributeManager.HasAttribute<CoreImageFilterPropertyAttribute> (pi))
 						continue;
 
-					if (AttributeManager.HasAttribute<WrapAttribute> (pi.GetGetMethod ()) || AttributeManager.HasAttribute<WrapAttribute> (pi.GetSetMethod ()))
+					// we ensure that we do not get a NRE in the case we forgot and ExportAttr in a property
```
spouliot: `in the case we forgot and ExportAttr in a property`
comment is unclear, you mean:
* `an` instead of `and` ?
* `[Wrap] attribute` instead of `ExportAttr` ?

src/generator.cs
mandel-macaque: Believe it or not, I had that and though it was a little noise. Will do :)

src/generator.cs
mandel-macaque: The issue happens when we forget an ExportAttr and we do not have a WrapAttr. I'll try to clarify it.

spouliot: maybe
``// Ensure there's a [Wrap] on either (or both) the getter and setter - since we already know there's no [Export]

xamarin-release-manager:
❌ [Build was (probably) aborted](http://xamarin-jenkins.guest.corp.microsoft.com:8080/job/macios/job/PR-4604/1/display/redirect)

🔥 [Jenkins job](http://xamarin-jenkins.guest.corp.microsoft.com:8080/job/macios/job/PR-4604/1/display/redirect) (on internal Jenkins) failed in stage 'Checkout' 🔥 : hudson.AbortException: script returned exit code 128

xamarin-release-manager:
❌ [Build was (probably) aborted](http://xamarin-jenkins.guest.corp.microsoft.com:8080/job/macios/job/PR-4604/2/display/redirect)

🔥 [Jenkins job](http://xamarin-jenkins.guest.corp.microsoft.com:8080/job/macios/job/PR-4604/2/display/redirect) (on internal Jenkins) failed in stage 'Checkout' 🔥 : hudson.AbortException: script returned exit code 128

xamarin-release-manager:
❌ [Build was (probably) aborted](https://xamarin-jenkins.internalx.com/job/macios/job/PR-4604/2/display/redirect)

🔥 [Jenkins job](https://xamarin-jenkins.internalx.com/job/macios/job/PR-4604/2/display/redirect) (on internal Jenkins) failed in stage 'Checkout' 🔥 : hudson.AbortException: script returned exit code 128

PR #4653, opened 2018-08-20
[AppKit] Xcode 10 beta 6 support.

mandel-macaque:


src/AppKit/Enums.cs
```diff
@@ -2898,4 +2898,20 @@ public enum NSColorSystemEffect : long {
...
+		SetAttributes,
+		ReplaceFile,
+	}
+
+	[Mac (10,14, onlyOn64: true)]
+	[Native]
+	public enum NSWorkspaceAuthorizationError : ulong  {
+		Invalid = 67328,
+		Minimum = 67328,
+		Maximum = 67455,
```
spouliot: Don't add `Minimum` and `Maximum` since they could change in future versions - and that's a breaking change for us

src/appkit.cs
```diff
@@ -21879,8 +21879,17 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Mac (10,14, onlyOn64: true)]
+		[Export ("requestAuthorizationOfType:completionHandler:")]
+		void RequestAuthorization (NSWorkspaceAuthorizationType type, Action<NSWorkspaceAuthorization, NSError> completionHandler);
 	}
 	
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface NSWorkspaceAuthorization {
+	}
```
spouliot: missing availability attribute ^ it's new in 10.14

src/foundation.cs
```diff
@@ -13480,6 +13480,13 @@ partial interface NSFileManager {
 
 		[Mac (10, 8), Export ("trashItemAtURL:resultingItemURL:error:")]
 		bool TrashItem (NSUrl url, out NSUrl resultingItemUrl, out NSError error);
+
+#if MONOMAC
+		[Mac (10,14)]
```
spouliot: ^ missing `onlyOn64: true`

src/appkit.cs
```diff
@@ -21879,8 +21879,18 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
 		[Field ("NSWorkspaceAccessibilityDisplayOptionsDidChangeNotification")]
 		[Notification]
 		NSString DisplayOptionsDidChangeNotification { get; }
+
+		[Mac (10,14, onlyOn64: true)]
+		[Export ("requestAuthorizationOfType:completionHandler:")]
+		void RequestAuthorization (NSWorkspaceAuthorizationType type, Action<NSWorkspaceAuthorization, NSError> completionHandler);
```
rolfbjarne: Would `[Async]` make sense here?

PR #4671, opened 2018-08-21
[apidiff] Update XML references with 15.8 as the new base

spouliot:


spouliot:
@rolfbjarne XM _should_ be identical to XI results. I _think_ this happened because we update XI more often (`Xcode*` branches) so that change was _done_ for XI earlier. I'll look at the filtering options the tool has

rolfbjarne:
@spouliot you're updating the XM reference files, but the api diff done by the bot still lists those differences. If the api diff should ever be empty, it should be when we're updating the reference files...

This feels like one of those differences we had in the past when we were getting differences because of different mono versions used to generate the references vs when creating the api diff.

spouliot:
@rolfbjarne the same mono was used to create references for both XI and XM in the PR. If the XML differs then it's because the assemblies also differ

spouliot:
I can reduce the diff size, removing assemblies with only those change, to help readability. Hand editing I'll avoid :)

PR #4683, opened 2018-08-23
[mtouch/mmp/generator] Build using project file instead of manually creating compiler command line. Fixes #4384.

### LINKED ISSUE #4384: [XI]mtouch commands fail with a missing method exception with Xamarin.iOS master and mono 5.12
### Steps to Reproduce

1.  Install Xamarin.iOS master, Xamarin.iOS
Version: 11.15.0.6 (Visual Studio Community)
Hash: e1106121
2. Install desktop mono, Mono 5.12.0.285 (2018-02/cae8c548849) (64-bit)
3. Execute any mtouch  command, /Developer/MonoTouch/usr/bin/mtouch --listdev

### Expected Behavior
Command should list device info

### Actual Behavior
Fails with a missing method exception. https://gist.github.com/GouriKumari/0465cc5ff0c42166de496fc654d934dc. 

### Environment
```
Xamarin.iOS
Version: 11.15.0.6 (Visual Studio Community)
Hash: e1106121
Mono 5.12.0.285 (2018-02/cae8c548849) (64-bit)
```

### Build Logs

Log: https://gist.github.com/GouriKumari/0465cc5ff0c42166de496fc654d934dc

###Additional Info
Issue is not observed when builds from following branches are installed:
1. XI from master branch after mono2018-04 bump and Desktop mono from 2018-04
2. XI from master branch before mono2018-04 bump and Desktop mono from 2018-02
3. XI from d15-8 branch and Desktop mono from 2018-02

I think bumping XI master to mono2018-04 caused the issue.

rolfbjarne:
Build using project files, so that we can target a specific .NET Framework
version, and not just use the latest available on the system that compiles the
tools.

This makes sure that the tools don't accidentally use API that might not be
available on the system where they will be executed.

Also add logic to extract the dependencies for each tool from the project
files, and use those to create make targets that will only invoke msbuild to
build the tools when needed (since make is much faster at determining that
nothing needs to be done).

Fixes https://github.com/xamarin/xamarin-macios/issues/4384.


tools/mmp/mmp.csproj
```diff
@@ -5,47 +5,31 @@
...
-    <UseXamMacFullFramework>true</UseXamMacFullFramework>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>True</DebugSymbols>
-    <DebugType>full</DebugType>
+    <DebugType>portable</DebugType>
     <Optimize>False</Optimize>
     <OutputPath>bin\Debug</OutputPath>
-    <DefineConstants>MONOMAC;MMP;DEBUG</DefineConstants>
+    <DefineConstants>MONOMAC;MMP;XAMARIN_APPLETLS;DEBUG</DefineConstants>
```
spouliot: I thought we killed `XAMARIN_APPLETLS` ? it's the only option today, why is `mmp` need it ?

rolfbjarne: I didn't check if it was needed, I just made sure the csproj does what the Makefile did.

PR #4686, opened 2018-08-23
[Network] Update to Xcode 10 beta 3.

mandel-macaque:
Adds the new functions added in beta 3 and cleans some of the xtro
files.


mandel-macaque:
No all the functions have been done, we need to deal with sockaddr and the two callbacks, which I'll do in a different PR.

tests/xtro-sharpie/common-Network.ignore
```diff
@@ -0,0 +1,28 @@
+# the following protocols ae binded manualy as NativeObject
```
spouliot: incorrect comment, they are **not** bound as protocol, but types

I thought the protocols were in addition tp the types...

tests/xtro-sharpie/common-Network.ignore
```diff
@@ -0,0 +1,28 @@
...
+!missing-protocol! OS_nw_path not bound
+!missing-protocol! OS_nw_path_monitor not bound
+!missing-protocol! OS_nw_protocol_definition not bound
+!missing-protocol! OS_nw_protocol_metadata not bound
+!missing-protocol! OS_nw_protocol_options not bound
+!missing-protocol! OS_nw_protocol_stack not bound
+
+# binded but with a name alias changes and xtro gets confused.
+!missing-pinvoke! nw_ip_create_metadata is not bound
+!missing-pinvoke! nw_udp_create_metadata is not bound
```
spouliot: I don't follow your comment ?!? where's xtro being confused.
AFAICS that code is under `#if false` (in `src/Network/NWProtocolMetadata.cs`) so xtro is right to report them as missing

tests/xtro-sharpie/common-Network.ignore
```diff
@@ -0,0 +1,28 @@
...
+
+# binded via a Marshal.ReadIntPtr (Dlfcn.dlsym ..
+!missing-field! _nw_parameters_configure_protocol_default_configuration not bound
+!missing-field! _nw_parameters_configure_protocol_disable not bound
+!missing-field! _nw_content_context_default_message not bound
+!missing-field! _nw_content_context_default_stream not bound
+!missing-field! _nw_content_context_final_send not bound
+!missing-field! _nw_connection_send_idempotent_content not bound
```
spouliot: hmm... maybe we should move them to the binding file (with the enum) so we can have our existing tests run on them

tests/xtro-sharpie/iOS-Network.todo
```diff
@@ -1,37 +1,6 @@
...
-!missing-field! kNWErrorDomainDNS not bound
-!missing-field! kNWErrorDomainPOSIX not bound
-!missing-field! kNWErrorDomainTLS not bound
 !missing-pinvoke! nw_endpoint_get_address is not bound
-!missing-pinvoke! nw_ip_create_metadata is not bound
 !missing-pinvoke! nw_path_monitor_set_cancel_handler is not bound
 !missing-pinvoke! nw_path_monitor_set_update_handler is not bound
-!missing-pinvoke! nw_protocol_metadata_is_tcp is not bound
 !missing-pinvoke! nw_release is not bound
 !missing-pinvoke! nw_retain is not bound
```
spouliot: those 2 can be moved to .ignore (see comments in original PR)

tests/xtro-sharpie/common-Network.ignore
mandel-macaque: True, sorry, I did not see the #if, I'll update the comment to state the rdar.

tests/xtro-sharpie/common-Network.ignore
mandel-macaque: _nw_parameters_configure_protocol_default_configuration and _nw_parameters_configure_protocol_disable are BlockLiterals, the others are NWContentContext, does Field support that?

mandel-macaque:
build

tests/xtro-sharpie/common-Network.ignore
spouliot: no, but you can replace
```
... Dlfcn.dlsym (Libraries.Network.Handle, "_nw_content_context_default_message") ...
```
with
```
Network.Constants._nw_content_context_default_message
```
and have this defined in the binding files (under a static `Constants` type).

src/Network/NWProtocolMetadata.cs
```diff
@@ -38,7 +38,6 @@ public enum NWIPEcnFlag {
 	[TV (12,0), Mac (10,14), iOS (12,0)]
 	public class NWProtocolMetadata : NativeObject {
 
-#if false
 		// Officially listed on header files, but seems to not work on Mac/iOS
 		// https://bugreport.apple.com/web/?problemID=42443077
```
spouliot: if working, remove the comment
they will be part of git history

src/network.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using ObjCRuntime;
+using Foundation;
+using CoreFoundation;
+
+namespace Network {
+
+	[TV (12,0), Mac (10,14), iOS (12,0)]
```
spouliot: `,onlyOn64: true` for `[Mac]`
or the intro tests will _likely_ fail

mandel-macaque:
@spouliot I have tested the framework with the NetCatNetwork and everything works as expected. We are going to have two xtro failures (I know before the build in the bots, I wanted to get the PR there for discussion).

The problem is that we have two enums NWPathStatus on in the [Network namespace](https://developer.apple.com/documentation/network/nw_path_status_t?language=objc), and another one in the[ NetworkExtensions framework](https://developer.apple.com/documentation/networkextension/nwpathstatus). Both have the same values (for now) and xtro complains about the Network namespace as follows:

`
?unknown-entry? !missing-enum-native! NWPathStatus in 'macOS-NetworkExtension.ignore'
?unknown-entry? !wrong-enum-size! NWPathStatus managed 4 vs native 8 in 'macOS-NetworkExtension.ignore'
`

Which I think it refers to the enum in network and not network extension. Is this an xtro bug? Should I change the way we deal with the enum in the Network namespace (adding extensions to it might not be the right thing to do).

src/Network/NWPath.cs
```diff
@@ -14,7 +14,9 @@
 using CoreFoundation;
 
 namespace Network {
-	public enum NWPathStatus {
+
+	[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+	public enum NWPathStatus : long {
```
spouliot: `: long` is not correct for Network.framework

It's defined as
```
typedef enum {
	/*! @const nw_path_status_invalid The path is not valid */
	nw_path_status_invalid = 0,
	/*! @const nw_path_status_satisfied The path has a usable route upon which to send and receive data */
	nw_path_status_satisfied = 1,
	/*! @const nw_path_status_unsatisfied The path does not have a usable route. This may be due to a network interface being down, or due to system policy.  */
	nw_path_status_unsatisfied = 2,
	/*! @const nw_path_status_satisfiable The path does not currently have a usable route, but a connection attempt will trigger network attachment */
	nw_path_status_satisfiable = 3,
} nw_path_status_t;
```

spouliot: since it's confusing let's add a comment like:
```
// this maps to `nw_path_status_t` in Network/Headers/path.h (and not the enum from NetworkExtension)
```

tests/xtro-sharpie/tvOS-Network.todo
```diff
@@ -1,37 +1,4 @@
...
-!missing-pinvoke! nw_retain is not bound
-!missing-pinvoke! nw_udp_create_metadata is not bound
-!missing-protocol! OS_nw_advertise_descriptor not bound
-!missing-protocol! OS_nw_connection not bound
-!missing-protocol! OS_nw_content_context not bound
-!missing-protocol! OS_nw_endpoint not bound
-!missing-protocol! OS_nw_error not bound
-!missing-protocol! OS_nw_interface not bound
-!missing-protocol! OS_nw_listener not bound
 !missing-protocol! OS_nw_object not bound
```
spouliot: that one can also be moved to `.ignore`

PR #2588, opened 2017-09-04
[mmp] Set DEVELOPER_DIR to our Xcode.

rolfbjarne:
This makes it possible to build mmp (the partial static registrar code) with
the wrong system Xcode (which is not supported, but that doesn't mean we can't
try to make it mostly work to ease our lives).

PR #4691, opened 2018-08-24
[mmp] Set DEVELOPER_DIR to our Xcode.

rolfbjarne:
This makes it possible to build mmp (the partial static registrar code) with
the wrong system Xcode (which is not supported, but that doesn't mean we can't
try to make it mostly work to ease our lives).


rolfbjarne:
This is a repeat of PR #2588, which was for xcode9, but was never merged to master so it got lost.

rolfbjarne:
Jenkins failed:

> FATAL: java.nio.file.NoSuchFileException: /var/lib/jenkins/global-pre-script/remove-problematic-ghprb-parameters.groovy

rolfbjarne:
build

PR #4700, opened 2018-08-27
Make the bgen build respect the BUILD_DIR variable. Fixes maccore#959.

rolfbjarne:
Make the bgen build respect the BUILD_DIR variable, so that the API comparison
can redirect the output as expected/desired.

Fixes https://github.com/xamarin/maccore/issues/959.


rolfbjarne:
The API comparison will still fail, because what matters is the _previous_ code, and since this is the fix, the previous code does not have the fix yet.

rolfbjarne:
API comparison failure is a known issue (https://github.com/xamarin/maccore/issues/959)

PR #4662, opened 2018-08-21
[system-dependencies] Use 'xcodebuild -runFirstLaunch' to install first-launch packages instead of manually installing each package.

rolfbjarne:
This is more future-proof, since the list of packages may change, or there may
be other tasks that need doing in addition to installing packages.

This might help/fix https://github.com/xamarin/maccore/issues/952.

PR #4710, opened 2018-08-28
[system-dependencies] Use 'xcodebuild -runFirstLaunch' to install first-launch packages instead of manually installing each package. (#4707)

rolfbjarne:
This is more future-proof, since the list of packages may change, or there may
be other tasks that need doing in addition to installing packages.

Fixes https://github.com/xamarin/maccore/issues/952.

This is a backport of the PRs #4662 and #4681.

### DIFF
--- a/system-dependencies.sh
+++ b/system-dependencies.sh
@@ -231,6 +231,45 @@ function install_visual_studio () {
 	rm -f $VS_DMG
 }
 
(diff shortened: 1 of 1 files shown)


rolfbjarne:
API comparison failure is known (https://github.com/xamarin/maccore/issues/959)

PR #4714, opened 2018-08-28
Add NullAllowed for GetContentSizeForFrame and GetFrameSizeForContent parameters

cwensley:
From https://developer.apple.com/documentation/appkit/nsscrollview/1403471-contentsizeforframesize:
> horizontalScrollerClass: The class used as the horizontal scroller. A value of nil specifies that no horizontal scroller is used.


VincentDondain:
build

PR #4722, opened 2018-08-29
[Network] Add missing handlers on NWPathMonitor.

mandel-macaque:


src/Network/NWPathMonitor.cs
```diff
@@ -54,5 +54,78 @@ public void SetQueue (DispatchQueue queue)
...
 		}
+
+		delegate void nw_path_monitor_update_handler_t (IntPtr block, IntPtr path);
+		static nw_path_monitor_update_handler_t static_UpdateSnapshot = TrampolineUpdatedSnapshot;
+
+		[MonoPInvokeCallback (typeof (nw_path_monitor_update_handler_t))]
+		static void TrampolineUpdatedSnapshot (IntPtr block, IntPtr path)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWPath>> (block);
+			if (del != null){
```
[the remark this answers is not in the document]
mandel-macaque: My vim does not have the right setup, will clean those.

src/Network/NWPathMonitor.cs
```diff
@@ -54,5 +54,78 @@ public void SetQueue (DispatchQueue queue)
...
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe void nw_path_monitor_set_cancel_handler (IntPtr handle, void *callback);
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void SetMonitorCancelledHandler (Action callback)
+		{
+			unsafe {
+				if (callback == null) {
+					nw_path_monitor_set_update_handler (GetCheckedHandle (), null);
```
spouliot: should not it be `nw_path_monitor_set_cancel_handler` ?

src/Network/NWPathMonitor.cs
```diff
@@ -54,5 +54,78 @@ public void SetQueue (DispatchQueue queue)
...
+			if (del != null) {
+				del ();
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe void nw_path_monitor_set_cancel_handler (IntPtr handle, void *callback);
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void SetMonitorCancelledHandler (Action callback)
```
spouliot: never sure if it's `Cancelled` or `Canceled` 
iirc the FXDG had something for those names

src/Network/NWPathMonitor.cs
```diff
@@ -54,5 +54,78 @@ public void SetQueue (DispatchQueue queue)
...
+				var nwPath = new NWPath (path, owns: false);
+				del (nwPath);
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe void nw_path_monitor_set_update_handler (IntPtr handle, void *callback);
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void SetUpdatedSnapshotHandler (Action<NWPath> callback)
```
spouliot: why `Snapshot` in the name ?

mandel-macaque: From the documentation of the headers

`
+/*!
+ * @typedef nw_path_monitor_update_handler_t
+ * @abstract
+ *		A block handler to be called whenever a path monitor has a new snapshot.
+ *		This will be called one or more times after calling nw_path_monitor_start.
+ */
+typedef void (^nw_path_monitor_update_handler_t) (nw_path_t path);
+
+/*!
+ * @function nw_path_monitor_set_update_handler
+ * @abstract
+ *		Sets the client update handler. This block will be called with the
+ *		current path when start is called and any time the path changes.
+ *
+ * @param monitor
+ *		The path monitor object.
+ *
+ * @param update_handler
+ *		The block to call when the path changes.
+ */
+API_AVAILABLE(macos(10.14), ios(12.0), watchos(5.0), tvos(12.0))
+void
+nw_path_monitor_set_update_handler(nw_path_monitor_t monitor,
+								   nw_path_monitor_update_handler_t update_handler);
`

src/Network/NWPathMonitor.cs
mandel-macaque: Good catch

src/Network/NWPathMonitor.cs
spouliot: https://www.mono-project.com/docs/tools+libraries/tools/gendarme/rules/naming/#usepreferredtermsrule

use one `l` instead of two `ll` since it's the American way ;-)

PR #4724, opened 2018-08-29
Make the bgen build respect the BUILD_DIR variable. Fixes maccore#959. (#4700)

rolfbjarne:
* [src] Fix bgen build to support custom output directory for api comparison. Fixes maccore#959.

This broke in 87c27e0c3f3582f105446b88179616d78b08793f, which made bgen
compile using a project file (I forgot to verify that it wouldn't affect the
API comparison, and the PR build didn't complain because problems with the API
comparison typically show up in subsequent PRs).

https://github.com/xamarin/maccore/issues/959

* [src] Fix response file usage to use BUILD_DIR, so that API comparison can redirect as expected.

* [src] Fix generation of generator.csproj's dependency file to use BUILD_DIR, so that API comparison can redirect as expected.

* [src] Fix bgen.exe's dependencies.

* Use full paths to create-makefile-fragment.sh.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -96,10 +96,10 @@ $(IOS_BUILD_DIR)/native/core.dll: $(IOS_CORE_SOURCES) frameworks.sources
 		$(IOS_CORE_SOURCES)
 
 # generated_sources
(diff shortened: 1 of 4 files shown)


rolfbjarne:
The API comparison failure is expected (this is the fix, which will start working for any subsequent PRs/commits).

PR #4727, opened 2018-08-29
[Security] Add missing SecProtocolMetadata missing API.

mandel-macaque:
This is a partial commit that contains most of the missing methods for
the handlers API on SecProtocolMetadata. The not added ones are more
complicated and will be added in a later commit.

### DIFF
--- a/src/Security/SecProtocolMetadata.cs
+++ b/src/Security/SecProtocolMetadata.cs
@@ -80,7 +80,7 @@ namespace Security {
 #if !COREBUILD
 
 		delegate void sec_protocol_metadata_access_distinguished_names_handler_t (IntPtr block, IntPtr dispatchData);
- 		static sec_protocol_metadata_access_distinguished_names_handler_t static_DistinguishedNamesForPeer  = TrampolineDistinguishedNamesForPeer;
+ 		static sec_protocol_metadata_access_distinguished_names_handler_t static_DistinguishedNamesForPeer = TrampolineDistinguishedNamesForPeer;
 
  		[MonoPInvokeCallback (typeof (sec_protocol_metadata_access_distinguished_names_handler_t))]
  		static void TrampolineDistinguishedNamesForPeer (IntPtr block, IntPtr data)
@@ -93,31 +93,29 @@ namespace Security {
  		}
 
  		[DllImport (Constants.SecurityLibrary)]
- 		static extern unsafe void sec_protocol_metadata_access_distinguished_names (IntPtr handle, void *callback);
+		[return: MarshalAs (UnmanagedType.I1)]
+ 		static extern bool sec_protocol_metadata_access_distinguished_names (IntPtr handle, ref BlockLiteral callback);
 
  		[BindingImpl (BindingImplOptions.Optimizable)]
  		public void SetDistinguishedNamesForPeerHandler (Action<DispatchData> callback)
  		{
- 			unsafe {
- 				if (callback == null) {
- 					sec_protocol_metadata_access_distinguished_names (GetCheckedHandle (), null);
- 					return;
- 				}
+			if (callback == null)
+				throw new ArgumentNullException (nameof (callback));
 
- 				BlockLiteral block_handler = new BlockLiteral ();
- 				BlockLiteral *block_ptr_handler = &block_handler;
- 				block_handler.SetupBlockUnsafe (static_DistinguishedNamesForPeer, callback);
+			var block_handler = new BlockLiteral ();
+			block_handler.SetupBlockUnsafe (static_DistinguishedNamesForPeer, callback);
 
- 				try {
- 					sec_protocol_metadata_access_distinguished_names (GetCheckedHandle (), (void*) block_ptr_handler);
- 				} finally {
- 					block_handler.CleanupBlock ();
- 				}
- 			}
+			try {
+				if (!sec_protocol_metadata_access_distinguished_names (GetCheckedHandle (), ref block_handler)) {
+					throw new InvalidOperationException ("Distinguished names are not accessible.");
+				}
+			} finally {
+				block_handler.CleanupBlock ();
+			}
  		}
 
 		delegate void sec_protocol_metadata_access_ocsp_response_handler_t (IntPtr block, IntPtr dispatchData);
- 		static sec_protocol_metadata_access_ocsp_response_handler_t static_OcspReposeForPeer  = TrampolineOcspReposeForPeer;
+ 		static sec_protocol_metadata_access_ocsp_response_handler_t static_OcspReposeForPeer = TrampolineOcspReposeForPeer;
 
  		[MonoPInvokeCallback (typeof (sec_protocol_metadata_access_ocsp_response_handler_t))]
  		static void TrampolineOcspReposeForPeer (IntPtr block, IntPtr data)
@@ -130,31 +128,29 @@ namespace Security {
  		}
 
  		[DllImport (Constants.SecurityLibrary)]
- 		static extern unsafe void sec_protocol_metadata_access_ocsp_response (IntPtr handle, void *callback);
+		[return: MarshalAs (UnmanagedType.I1)]
+ 		static extern bool sec_protocol_metadata_access_ocsp_response (IntPtr handle, ref BlockLiteral callback);
 
  		[BindingImpl (BindingImplOptions.Optimizable)]
  		public void SetOcspResponseForPeerHandler (Action<DispatchData> callback)
  		{
- 			unsafe {
- 				if (callback == null) {
- 					sec_protocol_metadata_access_ocsp_response (GetCheckedHandle (), null);
- 					return;
- 				}
+			if (callback == null)
+				throw new ArgumentNullException (nameof (callback));
 
- 				BlockLiteral block_handler = new BlockLiteral ();
- 				BlockLiteral *block_ptr_handler = &block_handler;
- 				block_handler.SetupBlockUnsafe (static_OcspReposeForPeer, callback);
+			var block_handler = new BlockLiteral ();
+			block_handler.SetupBlockUnsafe (static_OcspReposeForPeer, callback);
 
- 				try {
- 					sec_protocol_metadata_access_ocsp_response (GetCheckedHandle (), (void*) block_ptr_handler);
- 				} finally {
- 					block_handler.CleanupBlock ();
- 				}
- 			}
+			try {
+				if (!sec_protocol_metadata_access_ocsp_response (GetCheckedHandle (), ref block_handler)) {
+					throw new InvalidOperationException ("The OSCP response is not accessible.");
+				}
+			} finally {
+				block_handler.CleanupBlock ();
+			}
  		}
 
 		delegate void sec_protocol_metadata_access_peer_certificate_chain_handler_t (IntPtr block, IntPtr certificate);
- 		static sec_protocol_metadata_access_peer_certificate_chain_handler_t static_CertificateChainForPeer  = TrampolineCertificateChainForPeer;
+ 		static sec_protocol_metadata_access_peer_certificate_chain_handler_t static_CertificateChainForPeer = TrampolineCertificateChainForPeer;
 
  		[MonoPInvokeCallback (typeof (sec_protocol_metadata_access_peer_certificate_chain_handler_t))]
  		static void TrampolineCertificateChainForPeer (IntPtr block, IntPtr certificate)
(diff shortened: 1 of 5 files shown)


src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,173 @@ public class SecProtocolMetadata : NativeObject {
...
-		//
-		// MISSING: all the block APIs
-		//
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_challenge_parameters_are_equal (IntPtr metadataA, IntPtr metadataB);
 
+		public static bool ParametersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
+		{
+			return sec_protocol_metadata_challenge_parameters_are_equal (metadataA.GetCheckedHandle (), metadataB.GetCheckedHandle ());
```
spouliot: there's no null check on parameters -> NRE
I think a better name would be `ChallengeParametersAreEqual`

even if I'd prefer `CompareChallengeParameters` but that suggest returning an `int` :|

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,173 @@ public class SecProtocolMetadata : NativeObject {
...
+		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_peers_are_equal (IntPtr metadataA, IntPtr metadataB);
+
+		public static bool PeersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
+		{
+			return sec_protocol_metadata_peers_are_equal (metadataA.GetCheckedHandle (), metadataB.GetCheckedHandle ());
```
spouliot: missing null check - it's not the same as `GetHandle` which is an extension method, maybe it's the one you're looking for (if the API accept `nil`)

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,173 @@ public class SecProtocolMetadata : NativeObject {
...
+ 					block_handler.CleanupBlock ();
+ 				}
+ 			}
+ 		}
+
+		delegate void sec_protocol_metadata_access_ocsp_response_handler_t (IntPtr block, IntPtr dispatchData);
+ 		static sec_protocol_metadata_access_ocsp_response_handler_t static_OCSPReposeForPeer  = TrampolineOCSPReposeForPeer;
+
+ 		[MonoPInvokeCallback (typeof (sec_protocol_metadata_access_ocsp_response_handler_t))]
+ 		static void TrampolineOCSPReposeForPeer (IntPtr block, IntPtr data)
```
spouliot: -> `Ocsp`

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,173 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				var dispatchData = new DispatchData (data, owns: false);
+ 				del (dispatchData);
+ 			}
+ 		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+ 		static extern unsafe void sec_protocol_metadata_access_ocsp_response (IntPtr handle, void *callback);
+
+ 		[BindingImpl (BindingImplOptions.Optimizable)]
+ 		public void SetOCSPResponseForPeerHandler (Action<DispatchData> callback)
```
spouliot: -> `Ocsp`

src/Security/SecProtocolMetadata.cs
mandel-macaque: The headers do not mention anything about null, so I'll add the null check.

src/Security/SecProtocolMetadata.cs
mandel-macaque: The headers do not mention anything about null, so I'll add the null check.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
-		//
-		// MISSING: all the block APIs
-		//
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_challenge_parameters_are_equal (IntPtr metadataA, IntPtr metadataB);
 
+		public static bool ChallengeParametersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
+		{
+			if (metadataA == null)
```
rolfbjarne: Should we return `true` if both `metadataA` and `metadataB` are null? What does `sec_protocol_metadata_challenge_parameters_are_equal` return if both inputs are `IntPtr.Zero`?

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
 		public bool EarlyDataAccepted => sec_protocol_metadata_get_early_data_accepted (GetCheckedHandle ()) != 0;
 
-		//
-		// MISSING: all the block APIs
-		//
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_challenge_parameters_are_equal (IntPtr metadataA, IntPtr metadataB);
 
+		public static bool ChallengeParametersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
```
rolfbjarne: Apple doesn't win any naming price for this... I think something like `CompareChallengeParameters` would be a better name, but maybe it's too different to the native name.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+			return sec_protocol_metadata_challenge_parameters_are_equal (metadataA.GetCheckedHandle (), metadataB.GetCheckedHandle ());
+		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_peers_are_equal (IntPtr metadataA, IntPtr metadataB);
+
+		public static bool PeersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
+		{
+			if (metadataA == null)
```
rolfbjarne: Same: what if both inputs are null?

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+				throw new ArgumentNullException (nameof (metadataA));
+			if (metadataB == null)
+				throw new ArgumentNullException (nameof (metadataB));
+			return sec_protocol_metadata_peers_are_equal (metadataA.GetCheckedHandle (), metadataB.GetCheckedHandle ());
+		}
+
+#if !COREBUILD
+
+		delegate void sec_protocol_metadata_access_distinguished_names_handler_t (IntPtr block, IntPtr dispatchData);
+ 		static sec_protocol_metadata_access_distinguished_names_handler_t static_DistinguishedNamesForPeer  = TrampolineDistinguishedNamesForPeer;
```
rolfbjarne: Style: only one space before the equals sign.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+ 		static extern unsafe void sec_protocol_metadata_access_distinguished_names (IntPtr handle, void *callback);
+
+ 		[BindingImpl (BindingImplOptions.Optimizable)]
+ 		public void SetDistinguishedNamesForPeerHandler (Action<DispatchData> callback)
+ 		{
+ 			unsafe {
+ 				if (callback == null) {
```
rolfbjarne: The native function is an iterator, which means that a null callback isn't valid.

This means a few things:

1. The managed function name could be better. `AccessDistinguishedNames` is what would most match the native name, but someone else might have a better suggestion.
2. Throw ArgumentNullException if callback is null.
3. The P/Invoke can use `ref BlockLiteral` (and not be unsafe):

```csharp
static extern void sec_protocol_metadata_access_distinguished_names (IntPtr handle, ref BlockLiteral callback);
```

4. The method that calls the P/Invoke doesn't have to be unsafe either:

```csharp
BlockLiteral block_handler = new BlockLiteral ();
block_handler.SetupBlockUnsafe (static_DistinguishedNamesForPeer, callback);

try {
	sec_protocol_metadata_access_distinguished_names (GetCheckedHandle (), ref block_handler);
} finally {
	block_handler.CleanupBlock ();
}
```

5. I wonder if we should also add an overload that does not take a block, but returns an array of all the iterated elements (@spouliot?):

```csharp
public IList<DispatchData> GetDistinguishedNames ()
{
    var rv = new List<DispatchData> ();
    AccessDistinguishedNames ((v) => rv.Add (v));
    return rv;
}
```

It seems a bit overkill to have an iterator function when there's no way to stop iterating in the middle.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				var dispatchData = new DispatchData (data, owns: false);
+ 				del (dispatchData);
+ 			}
+ 		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+ 		static extern unsafe void sec_protocol_metadata_access_ocsp_response (IntPtr handle, void *callback);
+
+ 		[BindingImpl (BindingImplOptions.Optimizable)]
+ 		public void SetOcspResponseForPeerHandler (Action<DispatchData> callback)
```
rolfbjarne: Same: iterator function.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				try {
+ 					sec_protocol_metadata_access_distinguished_names (GetCheckedHandle (), (void*) block_ptr_handler);
+ 				} finally {
+ 					block_handler.CleanupBlock ();
+ 				}
+ 			}
+ 		}
+
+		delegate void sec_protocol_metadata_access_ocsp_response_handler_t (IntPtr block, IntPtr dispatchData);
+ 		static sec_protocol_metadata_access_ocsp_response_handler_t static_OcspReposeForPeer  = TrampolineOcspReposeForPeer;
```
rolfbjarne: Same: two spaces before equals.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				try {
+ 					sec_protocol_metadata_access_ocsp_response (GetCheckedHandle (), (void*) block_ptr_handler);
+ 				} finally {
+ 					block_handler.CleanupBlock ();
+ 				}
+ 			}
+ 		}
+
+		delegate void sec_protocol_metadata_access_peer_certificate_chain_handler_t (IntPtr block, IntPtr certificate);
+ 		static sec_protocol_metadata_access_peer_certificate_chain_handler_t static_CertificateChainForPeer  = TrampolineCertificateChainForPeer;
```
rolfbjarne: Same: two spaces before equals.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				var secCertificate = new SecCertificate (certificate, owns: false);
+ 				del (secCertificate);
+ 			}
+ 		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+ 		static extern unsafe void sec_protocol_metadata_access_peer_certificate_chain (IntPtr handle, void *callback);
+
+ 		[BindingImpl (BindingImplOptions.Optimizable)]
+ 		public void SetCertificateChainForPeerHandler (Action<SecCertificate> callback)
```
rolfbjarne: Same: iterator function.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 				try {
+ 					sec_protocol_metadata_access_peer_certificate_chain (GetCheckedHandle (), (void*) block_ptr_handler);
+ 				} finally {
+ 					block_handler.CleanupBlock ();
+ 				}
+ 			}
+ 		}
+
+		delegate void sec_protocol_metadata_access_supported_signature_algorithms_handler_t (IntPtr block, ushort signatureAlgorithm);
+ 		static sec_protocol_metadata_access_supported_signature_algorithms_handler_t static_SignatureAlgorithmsForPeer  = TrampolineSignatureAlgorithmsForPeer;
```
rolfbjarne: Same: two spaces before equals.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,181 @@ public class SecProtocolMetadata : NativeObject {
...
+ 			if (del != null) {
+ 				del (signatureAlgorithm);
+ 			}
+ 		}
+
+ 		[DllImport (Constants.SecurityLibrary)]
+ 		static extern unsafe void sec_protocol_metadata_access_supported_signature_algorithms (IntPtr handle, void *callback);
+
+ 		[BindingImpl (BindingImplOptions.Optimizable)]
+ 		public void SetSignatureAlgorithmsForPeerHandler (Action<ushort> callback)
```
rolfbjarne: Same: iterator function.

src/Security/SecProtocolMetadata.cs
rolfbjarne: @spouliot what about just `Equals` (like `string.Equals`)? The input parameter types varies, so there can be multiple overloads with that name.

src/Security/SecProtocolMetadata.cs
mandel-macaque: Thanks for this! I'll update the iterator ones. I personally like the idea of the overload, but lets wait for sebs input. I'll fix the other comments and will push. We can add the overload later.

src/Security/SecProtocolMetadata.cs
mandel-macaque: I checked the API and when both are NULL the API returns false. I'll copy the behaviour.

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,177 @@ public class SecProtocolMetadata : NativeObject {
...
-		// MISSING: all the block APIs
-		//
+ 		[DllImport (Constants.SecurityLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool sec_protocol_metadata_challenge_parameters_are_equal (IntPtr metadataA, IntPtr metadataB);
 
+		public static bool ChallengeParametersAreEqual (SecProtocolMetadata metadataA, SecProtocolMetadata metadataB)
+		{
+			if (metadataA == null && metadataB == null)
+				return false;  // This was tested in a native app. We do copy the behaviour.
```
spouliot: weird it's not true...

src/Security/SecProtocolMetadata.cs
```diff
@@ -51,9 +51,177 @@ public class SecProtocolMetadata : NativeObject {
...
+			if (metadataA == null)
+				throw new ArgumentNullException (nameof (metadataA));
+			if (metadataB == null)
+				throw new ArgumentNullException (nameof (metadataB));
```
spouliot: does the API allows `nil` ? the previous comment makes it sound like it does,  if that's the case then we should too, so no throwing `ANE`. The common pattern is:
```
if (a == null)
   return (b == null);
else if (b == null)
   return false;
return sec_...(a, b);
```

spouliot: or, if `nil` is allowed then it might be better to use `GetHandle` extension method, to avoid any exception to be thrown (not really needed since the decision should be based on the `bool` result), e.g.
```
return sec_protocol_metadata_challenge_parameters_are_equal (metadataA.GetHandle (), metadataB.GetHandle ());
```

mandel-macaque: @spouliot the API does allow the null, but a warning is raised (as in passing NULL, NULL) will try with a one null and the other one present.

spouliot: If the headers don't mention null is fine and the OS shows warning in the logs then let's assume null is not fine (at least to give to the API). In that case the first approach (exception less) is better than the `GetHandle` one.

mandel-macaque: :+1:

mandel-macaque:
@rolfbjarne as we talked on slack, lets land this and I will provide tests for the manual code as soon as we have tests for Network. We need Network to be tested and work as the entry point for the tests in the security framework.

Issue #4190, opened 2018-06-05
UIGestureRecognizer needs a way of surfacing the token for the default action


migueldeicaza:
The UIGestureRecognizer can easily create strong cycles when the action is either capturing the context, or is the instance of a class.   And we currently do not have a way of removing the object.

We need a way to surfacing the "Recognizers" or on first creation the first Token, so users have a way of manually unregistering the getsture recognizer.

There is just no way of getting the "Token" for the first instance, only for any manually added targets (when we return the `Token` value from `AddTarget`).

migueldeicaza:
Probably should also support Dispose().

ailen0ada:
Hello! thank you for your kind!

migueldeicaza:
It might be enough to make the fields WeakReferences.

migueldeicaza:
Some additional notes:

* We can use `[Dispose ("OnDispose")]` as a decoration on `UIGestuzeRecognizer` to have some code run on the `Dispose()` method that will clear the objects in `recognizers`.

* The API that we need to call to unregisters handlers requires both the `NSObject` target and the selector.   Currently `recognizers` only stores the target `NSObject` for the single case, or holds a hashtable of the `NSObjects` in there.   We need to track both targets and selectors (because we use different selectors).

* We likely need to make `recognizers` a Dictionary to track these.

* There is a constructor we surface that takes an NSObject/Selector that we currently do not intercept, so we would not be able to Dispose() that particular one, the user would have to do that.

rolfbjarne:
Fixed with PR #4729.

PR #4729, opened 2018-08-30
[UIKit] UIGestureRecognizer, support a way of unsubscribing without creating cycles

migueldeicaza:
This now tracks all the uses of AddTarget with delegates by recording
the Token/Selector pair and making `Dispose()` release all the linkage
as well as providing an enumerator that can be used to get all the
registered Action handlers - this can then be used with .First() and
then passed to `RemoveTarget`.

This addresses https://github.com/xamarin/xamarin-macios/issues/4190

This initial patch is here for discussion of the approach, want to
review and test this before we merge.


src/UIKit/UIGestureRecognizer.cs
```diff
@@ -130,12 +132,22 @@ public void RemoveTarget (Token token)
 				throw new ArgumentNullException ("token");
 			if (recognizers == null)
 				return;
-			if (recognizers == token)
-				recognizers = null;
-			Hashtable asHash = recognizers as Hashtable;
-			if (asHash != null)
-				asHash.Remove (token);
-			RemoveTarget (token, token is ParametrizedDispatch ? Selector.GetHandle (parametrized_selector) : Selector.GetHandle (tsel));
+			if (recognizers.ContainsKey (token)){
```
rolfbjarne: Minor code improvement:

```csharp
if (recognizers.TryGetValue (token, out var sel)) {
    recognizers.Remove (token);
    RemoveTarget (token, sel);
}
```

src/UIKit/UIGestureRecognizer.cs
```diff
@@ -130,12 +132,22 @@ public void RemoveTarget (Token token)
...
+				RemoveTarget (token, sel);
+			}
+		}
+
+		//
+		// Used to enumerate all the registered handlers for this UIGestureRecognizer
+		//
+		public IEnumerable<Token> GetTargets ()
+		{
+			if (recognizers == null)
```
rolfbjarne: Can't this just be:

```csharp
public IEnumerable<Token> GetTargets ()
{
    return recognizers?.Keys;
}
```

src/UIKit/UIGestureRecognizer.cs
```diff
@@ -30,18 +34,26 @@ public UIGestureRecognizer (Action action) : this (Selector.GetHandle (tsel), ne
...
+			recognizers = null;
+		}
+		
 		//
 		// Signature swapped, this is only used so we can store the "token" in recognizers
 		//
 		public UIGestureRecognizer (Selector sel, Token token) : this (token, sel)
 		{
-			recognizers = token;
+			recognizers [token] = sel.Handle;
```
spouliot: that can cause an NRE here since the check (in `this`) would be done later.

src/UIKit/UIGestureRecognizer.cs
migueldeicaza: Good idea to avoid the for loop, but I will keep the split yield break/yield return to avoid returning a null for empty cases, instead, by catching this early we would return an empty collection.

src/UIKit/UIGestureRecognizer.cs
migueldeicaza: Love it

src/UIKit/UIGestureRecognizer.cs
migueldeicaza: You mean because the selector might be null?   I do not follow the "would be done later" part

migueldeicaza: This would call the constructor generated for the init: method, and that one should throw already if either sel or token are null in this case.

src/UIKit/UIGestureRecognizer.cs
Therzok: `recognizers?.Keys ?? Array.Empty<Token>`? That would avoid allocating an Iterator for the method.

src/UIKit/UIGestureRecognizer.cs
Therzok: Given this targets [modern mono](https://github.com/mono/mono/blob/master/mcs/class/corlib/corlib.dll.sources#L1693) with [corefx APIs](https://github.com/mono/corefx/blob/master/src/Common/src/CoreLib/System/Collections/Generic/Dictionary.cs#L829), can't it just be:
```
if (recognizers.Remove (token, out var sel)) {
    RemoveTarget (token, sel);
}
```

src/UIKit/UIGestureRecognizer.cs
spouliot: yes, the input can be `null`, directly or indirectly, e.g. `Selector.FromHandle` can return `null`

`new UIGestureRecognizer (null, token);` would throw the `NullReferenceException` before it calls `this (token, sel)` which has the null check (and would throw an `ArgumentNullException`)

rolfbjarne: @spouliot the other (`this`) constructor is called first, not last.

rolfbjarne:
`do-not-merge` until it's been tested (which is pending according to the original description).

Unit tests would also be nice to have.

PR #4737, opened 2018-08-31
[src] Strip reference assemblies now that the generator doesn't need to load them anymore.

rolfbjarne:


spouliot:
most failures seems to happen on tvOS
```
MTOUCH: error MT2091: Inliner failed processing `System.Boolean Photos.PHFetchResult/<GetEnumerator>d__2::MoveNext()`. 
```
might be a missing resolve - but weird it happens only on tvOS

rolfbjarne:
Yes, I need to investigate a bit.

PR #4738, opened 2018-08-31
[tests] Skip LinkSdkRegressionTest.ForeignKey_650402 on watchOS because it uses features not supported with bitcode.

rolfbjarne:


rolfbjarne:
Test failures are due to https://github.com/xamarin/maccore/issues/977.

PR #4739, opened 2018-08-31
[xharness] Add project configurations to run & debug Makefile generation.

rolfbjarne:
This also required a minor fix to allow relative directories as the root
directory.


rolfbjarne:
Test failures are due to https://github.com/xamarin/maccore/issues/977.

PR #4750, opened 2018-09-04
mono-2018-08: use coop transitions for monomac

### LINKED ISSUE #4723: Race condition with GC stopping the world and framework_peer_release_lock
Deadlock: https://gist.github.com/Therzok/6132ea63a97eb2bcc7b9d7282e973f72

Thread 29 is [trying to stop the world](https://gist.github.com/Therzok/6132ea63a97eb2bcc7b9d7282e973f72#file-monobt-all-L597-L604): 

```
  thread #29
    frame #0: 0x00007fff6efd8246 libsystem_kernel.dylib`semaphore_wait_trap + 10
    frame #1: 0x000000010c27b978 libmonosgen-2.0.dylib`mono_os_sem_timedwait [inlined] mono_os_sem_wait at mono-os-semaphore.h:90 [opt]
    frame #2: 0x000000010c27b970 libmonosgen-2.0.dylib`mono_os_sem_timedwait(sem=<unavailable>, timeout_ms=<unavailable>, flags=MONO_SEM_FLAGS_NONE) at mono-os-semaphore.h:109 [opt]
    frame #3: 0x000000010c27b1f5 libmonosgen-2.0.dylib`mono_threads_wait_pending_operations at mono-threads.c:310 [opt]
    frame #4: 0x000000010c21ec33 libmonosgen-2.0.dylib`sgen_client_stop_world at sgen-stw.c:297 [opt]
    frame #5: 0x000000010c21ec1f libmonosgen-2.0.dylib`sgen_client_stop_world(generation=1, serial_collection=0) at sgen-stw.c:124 [opt]
    frame #6: 0x000000010c231f76 libmonosgen-2.0.dylib`sgen_stop_world(generation=1, serial_collection=0) at sgen-gc.c:3821 [opt]
```

Thread 1 is [trying to lock `framework_peer_release_lock`](https://gist.github.com/Therzok/6132ea63a97eb2bcc7b9d7282e973f72#file-bt-all-L2-L7):

```
* thread #1, name = 'GUI Thread', queue = 'com.apple.main-thread'
  * frame #0: 0x00007fff6efe1a46 libsystem_kernel.dylib`__psynch_mutexwait + 10
    frame #1: 0x00007fff6f1a9b9d libsystem_pthread.dylib`_pthread_mutex_lock_wait + 83
    frame #2: 0x00007fff6f1a74c8 libsystem_pthread.dylib`_pthread_mutex_lock_slow + 253
    frame #3: 0x00000001094b75e4 libxammac.dylib`xamarin_framework_peer_lock + 20
    frame #4: 0x00000001094b9a66 libxammac.dylib`xamarin_release_managed_ref + 166
```

Thread 29 has the `framework_peer_release_lock`:

```
(lldb) x/10gx &framework_peer_release_lock
0x1094d9ae0: 0x000000004d555458 0x0001206800000000
0x1094d9af0: 0x4d55545800000000 0x000000000001623a
0x1094d9b00: 0x000ede00000ee003 0xffffffffffffffff
0x1094d9b10: 0xfffffffef6b2651f 0x4d5554584d555458
0x1094d9b20: 0x00007fe2a1aaf5e0 0x00007fe2a1aaf4e0
```

This means the lock is held by the thread with id 0x1623a, [which is thread #29](https://gist.github.com/Therzok/6132ea63a97eb2bcc7b9d7282e973f72#file-thread-info-all-L56):

```
thread #29: tid = 0x1623a, 0x00007fff6efd8246 libsystem_kernel.dylib`semaphore_wait_trap + 10, name = 'Thread Pool Worker'
```

The deadlock goes like this:

1. Thread 29 starts stopping the world, and calls XM's `gc_event_callback`, which will lock `framework_peer_release_lock`: https://github.com/xamarin/xamarin-macios/blob/2ee29f38ce955aff07a5025fde296b5fea1a967f/runtime/runtime.m#L855-L870.
2. Thread 29 takes the `framework_peer_release_lock`, but has not gotten around to stopping the entire world quite yet.
3. Thread 1 tries to take the `framework_peer_release_lock`, can't, and starts waiting for it to be released.
4. Thread 29 now tries to suspend all threads, and keeps waiting for thread 1 to do so.
5. Deadlock between threads 1 & 29, they're waiting for each other.

lambdageek:
With `2018-08` (nee 5.18) Mono uses hybrid suspend by default.

Hybrid suspend use cooperative suspend for threads that are running managed code or native Mono runtime code (or other code such as XamMac code marked with GC Unsafe regions) and preemptive suspend for threads that are running other code (such as P/Invokes, blocking system calls, and other code marked by GC Safe regions).

Every mono API function called from XM (listed in the `runtime/exports.t4` file) has been wrapped inside to perform GC Unsafe transitions when called.

This PR changes XM to use the coop attach/detach functions.  If hybrid suspend is being used, these will ensure that when threads detach they will be left in GC Safe mode.  (If preemptive suspend is used, the coop attach/detach just amount to the same thing as the JIT attach functions from before).

Also bump the required minimum system mono version for running XM apps to 5.18.0.68.

Fixes #4723

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -62,13 +62,13 @@ XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_9.4.xip
 XCODE_DEVELOPER_ROOT=/Applications/Xcode94.app/Contents/Developer
 
 # Minimum Mono version
-MIN_MONO_VERSION=5.18.0.15
+MIN_MONO_VERSION=5.18.0.68
 MAX_MONO_VERSION=5.18.99
-MIN_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-08/7/f123d67e048465fdcc980059e7ec70daf9e0f7cc/MonoFramework-MDK-5.18.0.15.macos10.xamarin.universal.pkg
+MIN_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-08/46/ad0854917ee4a71cbdb24f5507793d5a863bb893/MonoFramework-MDK-5.18.0.68.macos10.xamarin.universal.pkg
 
 # Minimum Mono version for Xamarin.Mac apps using the system mono
-MIN_XM_MONO_VERSION=5.14.0.136
-MIN_XM_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-04/111/07c8f25fe536dbb7e244f965aa3f9a871f41e953/MonoFramework-MDK-5.14.0.136.macos10.xamarin.universal.pkg
+MIN_XM_MONO_VERSION=5.18.0.68
+MIN_XM_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-08/46/ad0854917ee4a71cbdb24f5507793d5a863bb893/MonoFramework-MDK-5.18.0.68.macos10.xamarin.universal.pkg
 
 # Minimum Visual Studio version
 MIN_VISUAL_STUDIO_URL=https://download.visualstudio.microsoft.com/download/pr/11550896/783d2219a348f93b6988fb415951788a/VisualStudioForMac-Preview-7.4.0.985.dmg
--- a/runtime/xamarin/runtime.h
+++ b/runtime/xamarin/runtime.h
@@ -342,9 +342,8 @@ public:
 #endif /* !TARGET_OS_WATCH */
 
 // Once we have one mono clone again the TARGET_OS_WATCH
-// condition should be removed (DYNAMIC_MONO_RUNTIME should still
-// be here though).
-#if TARGET_OS_WATCH && !defined (DYNAMIC_MONO_RUNTIME)
+// condition should be removed.
+#if TARGET_OS_WATCH || MONOMAC
 #define MONO_THREAD_ATTACH \
 	do { \
 		gpointer __thread_dummy; \
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -72,7 +72,7 @@ const char *xamarin_arch_name = "x86_64";
 const char *xamarin_arch_name = NULL;
 #endif
 
-#if TARGET_OS_WATCH
+#if TARGET_OS_WATCH || MONOMAC
 bool xamarin_is_gc_coop = true;
 #else
 bool xamarin_is_gc_coop = false;


lambdageek:
In particular the attach/detach stuff is needed so that Grand Central Dispatch worker threads are left in GC Safe mode when they're idle.  Otherwise they were incorrectly left in GC Unsafe, and Mono would deadlock while waiting for them to cooperatively suspend (which they would never do since they weren't running any code).

Make.config
```diff
@@ -67,8 +67,8 @@ MAX_MONO_VERSION=5.18.99
 MIN_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-08/7/f123d67e048465fdcc980059e7ec70daf9e0f7cc/MonoFramework-MDK-5.18.0.15.macos10.xamarin.universal.pkg
 
 # Minimum Mono version for Xamarin.Mac apps using the system mono
-MIN_XM_MONO_VERSION=5.14.0.136
-MIN_XM_MONO_URL=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-04/111/07c8f25fe536dbb7e244f965aa3f9a871f41e953/MonoFramework-MDK-5.14.0.136.macos10.xamarin.universal.pkg
+MIN_XM_MONO_VERSION=5.18.0.68
```
rolfbjarne: You need to bump `MIN_MONO_VERSION` (and `MIN_MONO_URL`) as well, otherwise the bots won't install the system mono required to build/run XM tests.

runtime/xamarin/runtime.h
```diff
@@ -344,7 +344,7 @@ class XamarinObject {
 // Once we have one mono clone again the TARGET_OS_WATCH
 // condition should be removed (DYNAMIC_MONO_RUNTIME should still
 // be here though).
-#if TARGET_OS_WATCH && !defined (DYNAMIC_MONO_RUNTIME)
+#if (TARGET_OS_WATCH && !defined (DYNAMIC_MONO_RUNTIME)) || MONOMAC
```
rolfbjarne: This can be simplified to:

```c
#if TARGET_OS_WATCH || MONOMAC
```

that said, is there a reason we should continue using the jit versions for iOS and tvOS? Are they faster than the coop versions, or do the coop versions not work correctly?

rolfbjarne:
> Also bump the required minimum system mono version for running XM apps to 5.18.0.68.

What happens for existing Xamarin.Mac apps built with earlier versions of XM/mono, that now happen to run with a 5.18 system mono? Will mono default to the old suspend logic in that case?

rolfbjarne:
And last question: isn't this the fix for https://github.com/xamarin/xamarin-macios/issues/4723?

runtime/xamarin/runtime.h
lambdageek: The coop versions will work correctly with preemptive suspend.

There are a few reasons not to change iOS and tvOS:
1. Caution.  There's no particular reason motivating a change on those devices at this time.
2. Stack space usage: the coop functions need to store a cookie on the stack (in preemptive it'll always be junk) so there's a stack penalty for deeply recursive code that goes through many layers of managed->runtime->managed->native.  Not sure if any real code behaves like that, however.

lambdageek:
> What happens for existing Xamarin.Mac apps built with earlier versions of XM/mono, that now happen to run with a 5.18 system mono? Will mono default to the old suspend logic in that case?

That's a good question - I don't know if we have any way to detect this situation.

> isn't this the fix for #4723?

yes

lambdageek:
I also need to set `xamarin_is_gc_coop` -  we want native unwinding via objC exceptions

rolfbjarne:
> I also need to set xamarin_is_gc_coop - we want native unwinding via objC exceptions

This means you need to update this logic as well: https://github.com/xamarin/xamarin-macios/blob/bc492585d137d8c3d3a2ffc827db3cdaae3cc869/tools/common/Application.cs#L456

Do we want customers to be able to disable coop in Xamarin.Mac? If so, the `xamarin_is_gc_coop` value should be set here (according to the `EnableCoopGC` value) instead of making it a constant `true`: https://github.com/xamarin/xamarin-macios/blob/bc492585d137d8c3d3a2ffc827db3cdaae3cc869/tools/mmp/driver.cs#L1080

lambdageek:
> Do we want customers to be able to disable coop in Xamarin.Mac?

I don't know.

Hold on. Maybe I'm on the wrong track here.  For now we only want hybrid suspend on system mono - maybe `TARGET_OS_WATCH || MONOMAC` should really be `TARGET_OS_WATCH || DYNAMIC_MONO_RUNTIME` ?

lambdageek:
I had another idea which I'm going to try out locally and likely replace this PR:

1. Mono will not use `mono_jit_thread_attach` internally for anything (it's not used under coop or hybrid suspend - the only remaining uses are in previously AOTd assemblies and in embedders using the Mono API).  So bump the AOT format version to force a re-compile and use another symbol.
2. Since the only remaining uses are in embedders, and the hybrid suspend position is that all embedder code is GC Safe until it calls a Mono API, if there's a call to `mono_jit_thread_attach`, it can only come from embedding code so therefore attach threads as GC Safe.
3. Don't do any thread state transitions or managed exception marshalling in Xamarin.Mac.  (We could keep some of what's in this PR, but it should also be correct to do nothing)

I think this will work.  And also it will be compatible with old versions of XM running on newer Mono.

rolfbjarne:
> I had another idea which I'm going to try out locally and likely replace this PR:

This idea sounds better :smile:

> managed exception marshalling

Managed exception marshalling is the right thing to do, but for different reasons (not doing it is always broken because it results in undefined behavior).

On watchOS we were able to make it mandatory, because it was a new platform (and due to coop the undefined behavior was more likely to actually result in problems, and maybe most importantly: any problems are _difficult_ to track down).

On other platforms it's currently opt-in due to backwards compatibility concerns (one day we'll make it opt-out, and hopefully even mandatory at some later point in the future).

lambdageek:
This PR isn't going forward.  Will try the approach from my previous comment.

PR #4751, opened 2018-09-05
[mtouch] add mixed-mode support

lewurm:
Note: Requires `mono-2018-06` to be merged, thus opened against it.

When enabling this option, mtouch will AOT compile `mscorlib.dll`.  At runtime that means every method that wasn't AOT'd will be executed by the runtime interpreter.

https://github.com/mono/mono/issues/6860

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
@@ -52,7 +52,7 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		<MtouchProjectDirectory>$(MSBuildProjectDirectory)</MtouchProjectDirectory>
 		<MtouchEnableSGenConc Condition="'$(MtouchEnableSGenConc)' == ''">False</MtouchEnableSGenConc>
 		<MtouchUseInterpreter Condition="'$(MtouchUseInterpreter)' == ''">False</MtouchUseInterpreter>
-		<MtouchUseInterpreterMixed Condition="'$(MtouchUseInterpreterMixed)' == ''">False</MtouchUseInterpreterMixed>
+		<MtouchInterpreter Condition="'$(MtouchInterpreter)' == ''">all</MtouchInterpreter>
 		<MtouchVerbosity Condition="$(MtouchVerbosity) == ''">2</MtouchVerbosity>
 
 		<IsMacEnabled>true</IsMacEnabled>
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
@@ -51,8 +51,6 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		<MtouchUseThumb Condition="'$(MtouchUseThumb)' == ''">False</MtouchUseThumb>
 		<MtouchProjectDirectory>$(MSBuildProjectDirectory)</MtouchProjectDirectory>
 		<MtouchEnableSGenConc Condition="'$(MtouchEnableSGenConc)' == ''">False</MtouchEnableSGenConc>
-		<MtouchUseInterpreter Condition="'$(MtouchUseInterpreter)' == ''">False</MtouchUseInterpreter>
-		<MtouchInterpreter Condition="'$(MtouchInterpreter)' == ''">all</MtouchInterpreter>
 		<MtouchVerbosity Condition="$(MtouchVerbosity) == ''">2</MtouchVerbosity>
 
 		<IsMacEnabled>true</IsMacEnabled>
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
@@ -52,6 +52,7 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		<MtouchProjectDirectory>$(MSBuildProjectDirectory)</MtouchProjectDirectory>
 		<MtouchEnableSGenConc Condition="'$(MtouchEnableSGenConc)' == ''">False</MtouchEnableSGenConc>
 		<MtouchUseInterpreter Condition="'$(MtouchUseInterpreter)' == ''">False</MtouchUseInterpreter>
+		<MtouchUseInterpreterMixed Condition="'$(MtouchUseInterpreterMixed)' == ''">False</MtouchUseInterpreterMixed>
 		<MtouchVerbosity Condition="$(MtouchVerbosity) == ''">2</MtouchVerbosity>
 
 		<IsMacEnabled>true</IsMacEnabled>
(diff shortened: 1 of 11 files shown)


lewurm:
@rolfbjarne correct. `mscorlib.dll` is special handled regarding the interpreter, since it's used to attach AOT compiled wrappers/trampolines:
* `--aot=interp mscorlib.dll`: emits a `.dylib` that only contains wrappers/trampolines for the interpreter, but no methods of the assembly are AOT compiled.
* `--aot=full,interp mscorlib.dll`: emits a `.dylib` that contains the same as `--aot=full mscorlib.dll` plus the wrappers/trampolines for the interpreter mentioned above.

Note that `interp` option in the AOT compiler is ignored for any other assembly than `mscorlib.dll`. So for example, `--aot=full System.dll` equals `--aot=full,interp System.dll`.

rolfbjarne:
OK, I think I understand, and I propose the following command-line syntax for mtouch:

* `--interpreter`: Enables the interpreter. Optionally takes a comma-separated list of assemblies to interpret (if prefixed with a minus sign, the assembly will be AOT-compiled instead). `all` can be used to specify all assemblies. Can be specified multiple times.

Examples:

* `--interpreter=all`: Interpret all assemblies. This is the same as `--interpreter`.
* `--interpreter=-all,System.dll`: AOT-compile all assemblies except System.dll
* `--interpreter=-mscorlib.dll`: Interpret all assemblies except mscorlib.dll, which is AOT-compiled.

This means that your `--interp-mixed` would instead be `--interpreter=-mscorlib.dll`.

Does that sound like it could work (and be flexible enough)?

lewurm:
I like your proposal, looks good.

Do you mind to implement your suggestion yourself? For example, I don't know how the suggested option would propagate to msbuild task.

rolfbjarne:
> Do you mind to implement your suggestion yourself?

Yes, I can have a look at this.

rolfbjarne:
@lewurm have a look now to see if what I did will work for you.

tools/common/Application.cs
```diff
@@ -64,6 +64,7 @@ public partial class Application
 		public bool? EnableCoopGC;
 		public bool EnableSGenConc;
 		public bool UseInterpreter;
+		public List<string> InterpretedAssemblies = new List<string> ();
```
spouliot: hmm... it's inside `common` which is shared with XM
not sure it's the best place, Maybe under `#if MTOUCH` ?

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
```diff
@@ -52,6 +52,7 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		<MtouchProjectDirectory>$(MSBuildProjectDirectory)</MtouchProjectDirectory>
 		<MtouchEnableSGenConc Condition="'$(MtouchEnableSGenConc)' == ''">False</MtouchEnableSGenConc>
 		<MtouchUseInterpreter Condition="'$(MtouchUseInterpreter)' == ''">False</MtouchUseInterpreter>
+		<MtouchInterpreter Condition="'$(MtouchInterpreter)' == ''">all</MtouchInterpreter>
```
spouliot: maybe both variables could be shared ? otherwise we end up with inconsistent state - and the tooling might start doing weird things (if the same duplicated conditions are not also _duplicated_ in code)

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
```diff
@@ -832,6 +832,7 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 			EnableBitcode="$(MtouchEnableBitcode)"
 			EnableSGenConc="$(MtouchEnableSGenConc)"
 			UseInterpreter="$(MtouchUseInterpreter)"
+			Interpreter="$(MtouchInterpreter)"
```
spouliot: same

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.props
rolfbjarne: Fixed to use a single variable.

tools/common/Application.cs
rolfbjarne: Fixed by moving the code to an mtouch-specific file.

tools/mtouch/mtouch.cs
```diff
@@ -476,6 +477,8 @@ public static string GetAotArguments (Application app, string filename, Abi abi,
 				args.Append ("llvmonly,");
 			else if (interp)
```
lewurm: I suggest to add an assert here, `interp` should only be true of `fname == "mscorlib.dll"` is true.

rolfbjarne: Don't you mean that `interp_full` should only be true if `name == "mscorlib.dll"`?

lewurm: No, other assemblies than `mscorlib.dll` are vaild with `--aot=interp,full` too, as it will add wrappers specific to signatures seen during AOT compilation that are necessary for the AOT<>interpreter transitions.

However, `--aot=interp` only makes sense for `mscorlib.dll`. At least that's how it's supposed to work, I should double check that on the mono side: https://github.com/mono/mono/issues/10543

ysmoradi:
> * `--interpreter=all`: Interpret all assemblies. This is the same as `--interpreter`.
> * `--interpreter=-all,System.dll`: AOT-compile all assemblies except System.dll
> * `--interpreter=-mscorlib.dll`: Interpret all assemblies except mscorlib.dll, which is AOT-compiled.

I passed --interpreter to additional mtouch args and aot limitations went away on physical device. am I able to use other examples in d16? for example `--interpreter=-all,SomeLibrary.dll`?
Thanks

lewurm:
@ysmoradi sorry I missed your message.  How did you verify that a specific assembly is interpreted or not?

rolfbjarne:
@ysmoradi 

> I provided some invalid names to it, and it's didn't show any kind of error to me.

We show a warning in that case, not an error: https://github.com/xamarin/xamarin-macios/blob/a2c80ed8e4b1e23f8d48ea4bccf8c00b2e036832/tools/mtouch/Target.cs#L305

You can make it an error by adding `/warnaserror:138` to the additional mtouch arguments in the project's iOS Build options.

ysmoradi:
> @ysmoradi sorry I missed your message. How did you verify that a specific assembly is interpreted or not?

I have an issue which results into JIT fallback & crash on iOS devices only.
I provided --interpreter and problem become solved. I decided to run only part of my app in an interpreted mode, because AOT mode has better performance for sure. So, I provided --interpreter=-all,SomeLibrary.dll
Based on what I've got understand, that means all assemblies should run in AOT mode, but SomeLibrary.dll should run in interpreted mode. I've no SomeLibrary.dll at all (!). Based on what @rolfbjarne said, there is no exception is expcted. That's fine. But I except my app to gets crashed again, but it doesn't crash. If all assemblies are running in AOT mode, I expect my app to get crashed.
In a nutshell, can I say --interpreter=-all is equal to no interpreter at all?

Issue #4689, opened 2018-08-24
dontlink/Debug/iOS 32-bit/device crashes at launch


rolfbjarne:
Stack trace:

```
2018-08-23 01:32:31.684 dont link[1179:2728925] critical: Stacktrace:

2018-08-23 01:32:31.685 dont link[1179:2728925] critical: 
Native stacktrace:

2018-08-23 01:32:31.886 dont link[1179:2728925] critical: 	0   libmonosgen-2.0.dylib               0x06c952bb mono_handle_native_crash + 196
2018-08-23 01:32:31.887 dont link[1179:2728925] critical: 	1   libmonosgen-2.0.dylib               0x06c9f46d mono_sigsegv_signal_handler + 226
2018-08-23 01:32:31.887 dont link[1179:2728925] critical: 	2   libsystem_platform.dylib            0x1e018259 _sigtramp + 24
2018-08-23 01:32:31.887 dont link[1179:2728925] critical: 	3   libXamarin.iOS.dll.dylib            0x0249e500 wrapper_native_to_managed_ObjCRuntime_Runtime_register_assembly_intptr_int_ + 236
2018-08-23 01:32:31.888 dont link[1179:2728925] critical: 	4   libxamarin-debug.dylib              0x06f728ec xamarin_register_assembly + 64
2018-08-23 01:32:31.888 dont link[1179:2728925] critical: 	5   libxamarin-debug.dylib              0x06f74804 _ZL17register_assemblyP13_MonoAssemblyPj + 152
2018-08-23 01:32:31.889 dont link[1179:2728925] critical: 	6   libxamarin-debug.dylib              0x06f7578c xamarin_initialize + 992
2018-08-23 01:32:31.889 dont link[1179:2728925] critical: 	7   libxamarin-debug.dylib              0x06f89758 xamarin_main + 2004
2018-08-23 01:32:31.890 dont link[1179:2728925] critical: 	8   dont link                           0x000a6937 dont link + 563511
2018-08-23 01:32:31.890 dont link[1179:2728925] critical: 	9   libdyld.dylib                       0x1de924eb <redacted> + 2
2018-08-23 01:32:31.890 dont link[1179:2728925] critical: 
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

```

The executable seems to hang, so there's no crash report.

[...]

PR #4757, opened 2018-09-05
[xharness] Automatically pass -Os to the native linker when building watchOS extensions with the linker disabled.

rolfbjarne:
This works around https://github.com/xamarin/xamarin-macios/issues/4689 for our own unit tests.

This also requires bumping the build timeout, because unfortunately using -Os
makes builds *much, much* slower.


rolfbjarne:
> This (or another PR) should make sure only dont link (and not introspection) is being run without the linker enabled.

https://github.com/xamarin/maccore/issues/1003

PR #4775, opened 2018-09-07
[msbuild] Added support for Siri Intents

jstedfast:
This PR implements the MSBuild portion of the new Siri Intents feature. There are 2 parts:

1. `IntentsDefinitionCompiler` - this maps to Xcode's `intentbuilderc` which (mostly) just copies the `Intents.intentdefinition` file into the app bundle (or, in our case, copies it into an intermediate output directory and adds it to the list of bundle resources with a LogicalName set). The other thing it seems to do is set some default values for some version fields and such. We let `intentbuilderc` do its thing.
2. `IntentsDefinitionCodegen` - this duplicates Xcode's codegen step which generates Objective-C code and compiles them into object files for linking into the main app during the later build phase. For us, we generate C# source files (one per intent, just like Xcode does) and add the C# source files to the `Compile` items list.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/SiriIntentsKeys.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/SiriIntentsKeys.cs
@@ -1,0 +1,69 @@
+﻿//
+// SiriIntentsKeys.cs
+//
+// Author: Jeffrey Stedfast <jestedfa@microsoft.com>
+//
+// Copyright (c) 2018 Microsoft Corp.
+//
+
+namespace Xamarin.MacDev.Tasks
+{
+	static class SiriIntentsKeys
+	{
+		public const string INIntentDefinitionModelVersion = "INIntentDefinitionModelVersion";
+		public const string INIntentDefinitionSystemVersion = "INIntentDefinitionSystemVersion";
+		public const string INIntentDefinitionToolsBuildVersion = "INIntentDefinitionToolsBuildVersion";
+		public const string INIntentDefinitionToolsVersion = "INIntentDefinitionToolsVersion";
+
+		public const string INEnums = "INEnums";
+		public const string INEnumName = "INEnumName";
+		public const string INEnumType = "INEnumType";
+		public const string INEnumValues = "INEnumValues";
+		public const string INEnumValueDisplayName = "INEnumValueDisplayName";
+		public const string INEnumValueDisplayNameID = "INEnumValueDisplayNameID";
+		public const string INEnumValueIndex = "INEnumValueIndex";
+		public const string INEnumValueName = "INEnumValueName";
+
+		public const string INIntents = "INIntents";
+		public const string INIntentCategory = "INIntentCategory";
+		public const string INIntentDefaultImageName = "INIntentDefaultImageName";
+		public const string INIntentDescription = "INIntentDescription";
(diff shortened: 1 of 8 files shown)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/IntentDefinitionCodegenTaskBase.cs
```diff
@@ -0,0 +1,439 @@
...
+		void WriteIntentDefinition (StreamWriter writer, string prefix, PDictionary intentDefinition)
+		{
+			writer.WriteLine ("\t[Register (\"{0}Intent\")]", prefix);
+			writer.WriteLine ("\tclass {0}Intent : INIntent", prefix);
+			writer.WriteLine ("\t{");
+
+			// write out the .ctors
+			writer.WriteLine ("\t\tpublic {0}Intent () : base (NSObjectFlag.Empty) {{ }}", prefix);
+			writer.WriteLine ();
+			writer.WriteLine ("\t\tpublic {0}Intent (IntPtr handle) : base (handle) {{ }}", prefix);
```
rolfbjarne: This should be `protected`.

jstedfast:
@rolfbjarne - I've just fixed the IntPtr .ctor to be `protected` as you suggested.  Thanks!

jstedfast:
This PR is still missing *some* code that might need to get generated. For example, it doesn't currently generate the "Handler" class. This is mostly because I don't know how to do that correctly. It seems that the user needs to actually add custom logic to the methods in this class.

msbuild/Xamarin.MacDev.Tasks.Core/SiriIntentsKeys.cs
```diff
@@ -0,0 +1,69 @@
...
+// Author: Jeffrey Stedfast <jestedfa@microsoft.com>
+//
+// Copyright (c) 2018 Microsoft Corp.
+//
+
+namespace Xamarin.MacDev.Tasks
+{
+	static class SiriIntentsKeys
+	{
+		public const string INIntentDefinitionModelVersion = "INIntentDefinitionModelVersion";
```
stephen-hawley: Minor - if the name of the C# symbol and the value of the string are the same, you can do this:
`public const string INIntentDefinitionModelVersion = nameof (INIntentDefinitionModelVersion);`

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/IntentDefinitionCodegenTaskBase.cs
```diff
@@ -0,0 +1,439 @@
...
+		{
+			if (name.Length > 1)
+				return char.ToUpperInvariant (name[0]) + name.Substring (1);
+
+			return name.ToUpperInvariant ();
+		}
+
+		static string GetCSharpTypeName (string type)
+		{
+			switch (type) {
```
stephen-hawley: similarly, you can you `nameof()` on type names here: `return nameof (NSString);`

jstedfast:
Screenshot of Intents editor:

intents-editor

VincentDondain:
@jstedfast sweeeeeet that looks good, though it's probably more relevant to this PR I think: https://github.com/xamarin/md-addins/pull/3399

jstedfast:
D'oh, wrong PR :p

PR #4784, opened 2018-09-11
[security] Add new SecCertificateCopyKey and SecTrustEvaluateWithError API with unit tests

spouliot:


spouliot:
timeout seems unrelated -> https://github.com/xamarin/maccore/issues/1021

PR #4803, opened 2018-09-13
[ios][coretelephony] Update for GM

spouliot:


src/coretelephony.cs
```diff
@@ -80,8 +82,29 @@ interface CTTelephonyNetworkInfo {
...
+
+		[iOS (12,0)]
+		[NullAllowed]
+		[Export ("serviceCurrentRadioAccessTechnology", ArgumentSemantic.Retain)]
+		NSDictionary<NSString, NSString> ServiceCurrentRadioAccessTechnology { get; }
+
+		[iOS (12,0)]
+		[NullAllowed]
+		[Export ("serviceSubscriberCellularProvidersDidUpdateNotifier", ArgumentSemantic.Copy)]
+		Action<NSString> CellularProvidersDidUpdateNotifier { get; set; }
```
VincentDondain: This is inconsistent with other APIs. It should be `ServiceSubscriberCellularProvidersDidUpdateNotifier`.

src/coretelephony.cs
```diff
@@ -80,8 +82,29 @@ interface CTTelephonyNetworkInfo {
...
+
+		[iOS (12,0)]
+		[Notification]
+		[Field ("CTServiceRadioAccessTechnologyDidChangeNotification")]
+		NSString RadioAccessTechnologyDidChangeNotification { get; }
```
VincentDondain: Same it's either everything has `Service` or nothing does (:

src/coretelephony.cs
```diff
@@ -80,8 +82,29 @@ interface CTTelephonyNetworkInfo {
 		CTCarrierEventHandler CellularProviderUpdatedEventHandler { get; set; }
```
VincentDondain: Good thing it's deprecated because we should probably have kept the `Subscriber` prefix.

src/coretelephony.cs
VincentDondain: _Note: don't forget to update the deprecated message if you change this._

VincentDondain: The swift signature, if it's of any help, is long: https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/3024512-servicesubscribercellularprovide

spouliot: the name has not been swift-inized - but you're right about consistency :)

src/coretelephony.cs
```diff
@@ -69,9 +69,11 @@ interface CTRadioAccessTechnology {
 
 	[BaseType (typeof (NSObject))]
 	interface CTTelephonyNetworkInfo {
+		[Deprecated (PlatformName.iOS, 12,0, message: "Use 'ServiceSubscriberCellularProviders' instead.")]
 		[Export ("subscriberCellularProvider", ArgumentSemantic.Retain)]
 		CTCarrier SubscriberCellularProvider { get; }
 
+		[Deprecated (PlatformName.iOS, 12,0, message: "Use 'CellularProvidersDidUpdateNotifier' instead.")]
```
VincentDondain: This is now: `ServiceSubscriberCellularProvidersDidUpdateNotifier`.

👀  https://github.com/xamarin/xamarin-macios/pull/4803#discussion_r217233430 👀

PR #4812, opened 2018-09-13
[tests][introspection] Re-enable 'RPSystemBroadcastPickerView' in tests

spouliot:
The type was not available in the simulator(s) before GM

ref: https://github.com/xamarin/xamarin-macios/issues/4189


spouliot:
hmm... it did work locally or I messed something :|
Need to double check

PR #4838, opened 2018-09-19
[Metal] Fixe the size of structures. Fixes #4611

### LINKED ISSUE #4611: [metal] Incorrect structs in Metal/Defs.cs
These structs are incorrect: https://github.com/xamarin/xamarin-macios/blob/master/src/Metal/Defs.cs#L278-L305

They do not match the native struct's size and potentially alignment.

```
typedef struct {
    /* NOTE: edgeTessellationFactor and insideTessellationFactor are interpreted as half (16-bit floats) */
    uint16_t edgeTessellationFactor[4];
    uint16_t insideTessellationFactor[2];
} MTLQuadTessellationFactorsHalf;

typedef struct {
    /* NOTE: edgeTessellationFactor and insideTessellationFactor are interpreted as half (16-bit floats) */
    uint16_t edgeTessellationFactor[3];
    uint16_t insideTessellationFactor;
} MTLTriangleTessellationFactorsHalf;
```

We cannot have `ushort[]` the ushort need to be inlined, fixed size.

In addition we should have tests that ensure the size of the managed struct is equal to the size of the native struct.

mandel-macaque:
The structs MTLQuadTessellationFactorsHalf and
MTLTriangleTessellationFactorsHalf are wrong as per the header
definition:

typedef struct {
    /* NOTE: edgeTessellationFactor and insideTessellationFactor are interpreted as half (16-bit floats) */
    uint16_t edgeTessellationFactor[4];
    uint16_t insideTessellationFactor[2];
} MTLQuadTessellationFactorsHalf;

typedef struct {
    /* NOTE: edgeTessellationFactor and insideTessellationFactor are interpreted as half (16-bit floats) */
    uint16_t edgeTessellationFactor[3];
    uint16_t insideTessellationFactor;
} MTLTriangleTessellationFactorsHalf;

We set the arrays size to be the correct one.

Fixes https://github.com/xamarin/xamarin-macios/issues/4611

### DIFF
--- a/src/Metal/Defs.cs
+++ b/src/Metal/Defs.cs
@@ -279,7 +279,9 @@ namespace Metal {
 	[StructLayout (LayoutKind.Sequential)]
 	public struct MTLQuadTessellationFactorsHalf
 	{
+		[MarshalAs (UnmanagedType.ByValArray, SizeConst = 4)]
 		public ushort[] EdgeTessellationFactor;
+		[MarshalAs (UnmanagedType.ByValArray, SizeConst = 2)]
 		public ushort[] InsideTessellationFactor;
 		
 		public MTLQuadTessellationFactorsHalf (ushort[] edgeTessellationFactor, ushort[] insideTessellationFactor)
@@ -294,6 +296,7 @@ namespace Metal {
 	[StructLayout (LayoutKind.Sequential)]
 	public struct MTLTriangleTessellationFactorsHalf
 	{
+		[MarshalAs (UnmanagedType.ByValArray, SizeConst = 3)]
 		public ushort[] EdgeTessellationFactor;
 		public ushort InsideTessellationFactor;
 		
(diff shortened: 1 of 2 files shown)


mandel-macaque:
I'm not sure if we have a pattern to test the size of the structs, or if we simply should do an assert on the expected size. Feedback welcome. (As mentioned in the issue, an intro/xtro test for this errors will be amazing).

mandel-macaque:
@spouliot your comments? I don't mind giving it a go to the implementation from @rolfbjarne

spouliot:
Let's not delay the fix for what is known to be broken.

Instead let's discuss how we can automate (as much as possible) in the [RFC](https://github.com/xamarin/xamarin-macios/issues/4842). I like @rolfbjarne approach but I think it can be done with less human intervention.

p.s. @mandel-macaque please add the 👍 and 👎 on the RFC :)

PR #4881, opened 2018-09-26
Report updateinfo for xamarin.ios and xamarin.mac in artifacts.json

rolfbjarne:
We'd like to expose the productId and versionId in artifacts.json to autofill this information in while publishing so it's less of a manual process


rolfbjarne:
Not needed.

PR #4897, opened 2018-10-01
[runtime] Only write to NSLog in chunks of max 4096 characters. Fixes maccore#1014.

rolfbjarne:
It seems older iOS versions have a bug where big(ish) chunks of string can
cause it to hang.

Fix this by not writing big strings to NSLog, instead split them up in chunks.

Fixes https://github.com/xamarin/maccore/issues/1014.

### DIFF
--- a/runtime/xamarin-support.m
+++ b/runtime/xamarin-support.m
@@ -35,7 +35,30 @@ xamarin_log (const unsigned short *unicodeMessage)
 		fwrite ("\n", 1, 1, stdout);
 	fflush (stdout);
 #else
-	NSLog (@"%@", msg);
+	if (length > 4096) {
+		// Write in chunks of max 4096 characters; older versions of iOS seems to have a bug where NSLog may hang with long strings (!).
+		// https://github.com/xamarin/maccore/issues/1014
+		const char *utf8 = [msg UTF8String];
+		int len = strlen (utf8);
+		const int max_size = 4096;
+		while (len > 0) {
+			int chunk_size = len > max_size ? max_size : len;
+
+			// Try to not break in the middle of a line, by looking backwards for a newline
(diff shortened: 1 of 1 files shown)


runtime/xamarin-support.m
```diff
@@ -35,7 +35,30 @@
 		fwrite ("\n", 1, 1, stdout);
 	fflush (stdout);
 #else
-	NSLog (@"%@", msg);
+	if (length > 4096) {
+		// Write in chunks of max 4096 characters; older versions of iOS seems to have a bug where NSLog may hang with long strings (!).
+		// https://github.com/xamarin/maccore/issues/1014
+		const char *utf8 = [msg UTF8String];
+		int len = strlen (utf8);
+		const int max_size = 4096;
```
spouliot: Why `4096` ? that's based on testing or on some indication (bug#) that mention it's safe ?

Splitting the output can have some consequences. E.g. from `os/log.h` in iOS 10 (which switched to `os_log` from `asl`) 

```
 * There is a physical cap of 1024 bytes per log line for dynamic content,
 * such as %s and %@, that can be written to the persistence store.
 * All content exceeding the limit will be truncated before it is
 * written to disk.
```

so a very large message would persist the first 1024 of every 4096 bytes. Not a huge problem, just a bit different than the actual behaviour. Some people split their output into 1024 chunks not to lose data [1] - also a different behaviour (but _maybe_ more useful).

[1] https://forums.developer.apple.com/thread/63537

rolfbjarne: See https://github.com/xamarin/maccore/issues/1014#issuecomment-426562706 - the problem seems to appear after ~8150 characters.

4096 seemed like the biggest (logical) value that would still work (I chose _biggest_ to change current behavior as little as possible).

Interesting data about the 1024 limit: seems like another max value that could make sense to use (either value is fine with me).

PR #4898, opened 2018-10-01
[msbuild] Added support for Siri Intents

jstedfast:
This PR implements the MSBuild portion of the new Siri Intents feature. There are 2 parts:

1. `IntentsDefinitionCompiler` - this maps to Xcode's `intentbuilderc` which (mostly) just copies the `Intents.intentdefinition` file into the app bundle (or, in our case, copies it into an intermediate output directory and adds it to the list of bundle resources with a LogicalName set). The other thing it seems to do is set some default values for some version fields and such. We let `intentbuilderc` do its thing.
2. `IntentsDefinitionCodegen` - this duplicates Xcode's codegen step which generates Objective-C code and compiles them into object files for linking into the main app during the later build phase. For us, we generate C# source files (one per intent, just like Xcode does) and add the C# source files to the `Compile` items list.

This PR supersedes https://github.com/xamarin/xamarin-macios/pull/4775

The only changes between this PR and the previous PR is the code generator now includes logic to generate the "Handling" class which makes use of some new XI APIs introduced in master.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/SiriIntentsKeys.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/SiriIntentsKeys.cs
@@ -1,0 +1,69 @@
+﻿//
+// SiriIntentsKeys.cs
+//
+// Author: Jeffrey Stedfast <jestedfa@microsoft.com>
+//
+// Copyright (c) 2018 Microsoft Corp.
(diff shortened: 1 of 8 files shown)


jstedfast:
Closing this one since we'll use the PR from the xamarin branch instead.

PR #4916, opened 2018-10-03
Enforce that the commit distances for Xamarin.iOS and Xamarin.Mac are identical.

rolfbjarne:
Example failure:

screenshot 2018-10-03 15 42 33

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,14 @@ world: check-system
 
 .PHONY: check-system
 check-system:
+	@if [[ "x$(IOS_COMMIT_DISTANCE)" != "x$(MAC_COMMIT_DISTANCE)" ]]; then \
+		echo "$(COLOR_RED)*** The commit distance for Xamarin.iOS ($(IOS_COMMIT_DISTANCE)) and Xamarin.Mac ($(MAC_COMMIT_DISTANCE)) are different.$(COLOR_CLEAR)"; \
+		echo "$(COLOR_RED)*** To fix this problem, the lines in Make.config defining IOS_PACKAGE_VERSION and MAC_PACKAGE_VERSION must be changed in the same commit.$(COLOR_CLEAR)"; \
+		echo "$(COLOR_RED)*** Once fixed, you need to commit the changes for them to pass this check.$(COLOR_CLEAR)"; \
+		exit 1; \
+	fi
 	@./system-dependencies.sh
+	@echo "Building Xamarin.iOS $(IOS_PACKAGE_VERSION) and Xamarin.Mac $(MAC_PACKAGE_VERSION)"
 
 $(DIRECTORIES):
 	$(Q) mkdir -p $@


spouliot:
hmm... I'm having second thoughts, e.g. merging an `xcode*` branch often only change XI version.

The check itself is likely good (a consistent distance) but it's not enough if our merges breaks it. We'll end up doing some hacks (or disabling the check) that will cause issue (e.g. a 2nd merge for the next Xcode).

rolfbjarne:
@spouliot what about this: https://github.com/rolfbjarne/xamarin-macios/commit/version-numbers?

spouliot:
@rolfbjarne I like it better :)

* the arrangement makes it less error prone (e.g. variables are closer);

* using 3 digits is simpler (again less error prone);

* much better comments 😄 

but

* it does not make it easier to fix a mistake (like in 15.9) with a reset to the commit distance (which is what triggered my 2nd thoughts yesterday). If I change many lines then it's more likely to cause conflicts and problems on next merge (e.g. `xcode10.1`);

* it does not ensure a consistent commit distance (nor warn/fail) - that's less of an real issue but it's confusing many times ☹️ 

Here's what I came up last night (incomplete) https://gist.github.com/spouliot/6821a7687e67e1bec7195f25bd62866d

* use a separate file to ease tracking version-only changes (only file RE has to touch). Git history should be shorter and only include version related changes;

* a single distance calculation, i.e. if we release the same commit/branch then it always has the same distance (we know it's the same revision looking at the version). I lacked something existing :( so I made my own variable to track;

* a workaround (offset) to fix the build number (distance) if we have to change anything (like the logic) or fix a mistake (like in 15.9). Not pretty but localized.

rolfbjarne:
> * it does not make it easier to fix a mistake (like in 15.9) with a reset to the commit distance

Bump the revision.

I can also implement your offset solution, but while it provides more control over the resulting version number, it's also more complex (merging a branch back into master could end up bringing non-zero offsets as well for instance - or the reverse could happen, where a branch with zero offset is merged into a branch that ).

> * it does not ensure a consistent commit distance (nor warn/fail)

I could modify the check in this PR to tell the developer to bump the revision (for both XI and XM).

> Here's what I came up last night

I see a couple of potential problems here:

1. `COMMIT_DISTANCE` isn't bumped regularly (seems like an easy thing to forget). This can become a problem if COMMIT_DISTANCE becomes > 999, because we only support 3 digits for this number (we can/should of course add a check for this).
2. `COMMIT_DISTANCE` is bumped when nothing else changes. This will cause repeated version numbers. This will probably happen if the future check for 1. triggers, because someone will think "ok, let's just bump COMMIT_DISTANCE then!" and think that's it.
3. `COMMIT_DISTANCE` is bumped and either `IOS_PACKAGE_VERSION_MAJOR_MINOR` or `MAC_PACKAGE_VERSION_MAJOR_MINOR` (but not both) are bumped. I'm assuming this will only happen if a branch is used to release either XI or XM (but not both), in which case it's not a problem - but when branches merge back I'm not sure what will happen. Let's say a branch where `IOS_PACKAGE_VERSION_MAJOR_MINOR` and `COMMIT_DISTANCE` were bumped is merged into a branch where `MAC_PACKAGE_VERSION_MAJOR_MINOR` was *not* bumped: I think this will end up with repeating version numbers for XM.

It seems there are a few pitfalls... I'm slightly in favor of an implementation where every problem can be solved with just "bump the revision" 😄

New version: https://github.com/rolfbjarne/xamarin-macios/commit/version-numbers-take-2 (where I've documented that [Any other problem can also usually be solved by bumping the revision](https://github.com/rolfbjarne/xamarin-macios/commit/version-numbers-take-2#diff-02f811b5b7fbabd1b10c6e970a944633R34)) 😉

spouliot:
@rolfbjarne I like that one even better. Let's go with it and try the bump-revision approach to fix errors.

rolfbjarne:
Closing in favor of #4927.

PR #4924, opened 2018-10-04
[CoreFoundation] Fetch a few static values lazily.

rolfbjarne:
This avoids using static constructors, and also avoids fetching the values
unless they're needed.


src/CoreFoundation/CFArray.cs
```diff
@@ -85,19 +85,12 @@ protected virtual void Dispose (bool disposing)
...
-				kCFTypeArrayCallbacks_ptr = Dlfcn.GetIndirect (handle, "kCFTypeArrayCallBacks");
-			}
-			finally {
-				Dlfcn.dlclose (handle);
+		static IntPtr kCFTypeArrayCallbacks_ptr_value;
+		static IntPtr kCFTypeArrayCallbacks_ptr {
+			get {
+				if (kCFTypeArrayCallbacks_ptr_value == IntPtr.Zero)
+					kCFTypeArrayCallbacks_ptr_value = Dlfcn.GetIndirect (Libraries.CoreFoundation.Handle, "kCFTypeArrayCallBacks");
+				return kCFTypeArrayCallbacks_ptr_value;
```
spouliot: Why not move this into `corefoundation.cs` as `[Field]` ?
That will give us some *free** introspection and xtro tests and less manual code to maintain.

src/CoreFoundation/CFBoolean.cs
```diff
@@ -38,20 +38,21 @@ namespace CoreFoundation {
...
-		public static readonly CFBoolean False;
-
-		static CFBoolean ()
-		{
-			var handle = Dlfcn.dlopen (Constants.CoreFoundationLibrary, 0);
-			if (handle == IntPtr.Zero)
-				return;
-			try {
-				True  = new CFBoolean (Dlfcn.GetIntPtr (handle, "kCFBooleanTrue"), false);
-				False = new CFBoolean (Dlfcn.GetIntPtr (handle, "kCFBooleanFalse"), false);
```
spouliot: same, if we touch that code then I'd rather use the bindings file

src/CoreFoundation/DispatchData.cs
```diff
@@ -43,10 +43,13 @@ public DispatchData (IntPtr handle) : base (handle, false)
...
-		static IntPtr free;
-		static DispatchData ()
-		{
-			free = Marshal.ReadIntPtr (Dlfcn.dlsym (Libraries.LibC.Handle, "_dispatch_data_destructor_free"));
+		static IntPtr free_ptr;
+		static IntPtr free {
+			get {
+				if (free_ptr == IntPtr.Zero)
+					free_ptr = Marshal.ReadIntPtr (Dlfcn.dlsym (Libraries.LibC.Handle, "_dispatch_data_destructor_free"));
+				return free_ptr;
```
spouliot: same

PR #4925, opened 2018-10-04
[CoreFoundation] Simplify DispatchQueue.MainQueue.

rolfbjarne:
* The dispatch_get_main_queue function doesn't exist anywhere (tested on iOS 5.1.1 - iOS 12, macOS 10.7 - 10.14), and when called from native code, it's always an inlined function, so just remove the call completely.
    
* Getting the _dispatch_main_q symbol from either the current address space, libSystem or libdispatch works fine everywhere. Looking up something in the current address space is costly (according to 'man dlsym'), so stop doing that: only look in libdispatch (since that's where the symbol actually is according to 'nm').

* I find no reason for the lock in DispatchQueue.MainQueue, nor does history reveal anything helpful, so I removed the lock.

### DIFF
--- a/src/CoreFoundation/Dispatch.cs
+++ b/src/CoreFoundation/Dispatch.cs
@@ -265,34 +265,22 @@ namespace CoreFoundation {
 				return new DispatchQueue (dispatch_get_global_queue ((nint) (int) DispatchQueuePriority.Default, 0), false);
 			}
 		}
-#if MONOMAC
-		static DispatchQueue PInvokeDispatchGetMainQueue ()
-		{
-			return new DispatchQueue (dispatch_get_main_queue (), false);
-		}
 
-#endif
 		static IntPtr main_q;
-		static object lockobj = new object ();
 
 		public static DispatchQueue MainQueue {
 			get {
-				lock (lockobj) {
-					if (main_q == IntPtr.Zero) {
-						// Try loading the symbol from our address space first, should work everywhere
-						main_q = Dlfcn.dlsym ((IntPtr) (-2), "_dispatch_main_q");
+				if (main_q == IntPtr.Zero) {
+					// Try loading the symbol from our address space first, should work everywhere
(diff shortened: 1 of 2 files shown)


src/CoreFoundation/Dispatch.cs
```diff
@@ -427,11 +415,8 @@ public void Submit (Action<int> action, long times)
 		// dispatch_queue_t dispatch_get_global_queue (long priority, unsigned long flags);
 		extern static IntPtr dispatch_get_global_queue (nint priority, nuint flags);
 
-#if MONOMAC
-		[Obsoleted (PlatformName.MacOSX, 10, 7)]
 		[DllImport (Constants.libcLibrary)]
 		extern static IntPtr dispatch_get_main_queue ();
```
spouliot: according to web docs it's new in iOS 8 and macOS 10.10
weird that it would have been deprecated in 10.7 ?!?
let's run this into old sim/devices

spouliot: https://developer.apple.com/documentation/dispatch/1452921-dispatch_get_main_queue

rolfbjarne: It seems my grepping was flawed: `dispatch_get_main_queue` shows up in earlier iOS SDKs, but it's a define, not an actual function (which it didn't become until iOS 8).

I'll have a look and see if we can remove the call to `dispatch_get_main_queue` entirely, and just rely on the `_dispatch_main_q` variable instead.

rolfbjarne: Some history:

https://github.com/xamarin/maccore/commit/25097444a9eacab88fe9b0d7f3ae640ce1768d14
https://github.com/xamarin/maccore/commit/04ea3ac8642812a94f401c286e62e67e2f1920dc

rolfbjarne:
I've done some testing, and this can be simplified further. Fix will be pushed after #4924 is merged, since the changes depends on that PR.

PR #4927, opened 2018-10-04
Try to make version numbers clearer, and establish some ground rules.

rolfbjarne:

PR #4931, opened 2018-10-04
[d15-9] Try to make version numbers clearer, and establish some ground rules.

spouliot:
Backport of https://github.com/xamarin/xamarin-macios/pull/4927
with 15.9 numbers

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -5,22 +5,20 @@ include $(TOP)/mk/subdirs.mk
 -include $(TOP)/Make.config.inc
 $(TOP)/Make.config.inc: $(TOP)/Make.config
 	@rm -f $@
-	@printf "IOS_COMMIT_DISTANCE:=$(shell LANG=C; export LANG && git --git-dir $(TOP)/.git log `git --git-dir $(TOP)/.git blame -- ./Make.config HEAD | grep IOS_PACKAGE_VERSION= | sed 's/ .*//' `..HEAD --oneline | wc -l | sed 's/ //g')\n" >> $@
-	@printf "MAC_COMMIT_DISTANCE:=$(shell LANG=C; export LANG && git --git-dir $(TOP)/.git log `git --git-dir $(TOP)/.git blame -- ./Make.config HEAD | grep MAC_PACKAGE_VERSION= | sed 's/ .*//' `..HEAD --oneline | wc -l | sed 's/ //g')\n" >> $@
+	@printf "IOS_COMMIT_DISTANCE:=$(shell LANG=C; export LANG && git --git-dir $(TOP)/.git log `git --git-dir $(TOP)/.git blame -- ./Make.versions HEAD | grep IOS_PACKAGE_VERSION= | sed 's/ .*//' `..HEAD --oneline | wc -l | sed 's/ //g')\n" >> $@
+	@printf "MAC_COMMIT_DISTANCE:=$(shell LANG=C; export LANG && git --git-dir $(TOP)/.git log `git --git-dir $(TOP)/.git blame -- ./Make.versions HEAD | grep MAC_PACKAGE_VERSION= | sed 's/ .*//' `..HEAD --oneline | wc -l | sed 's/ //g')\n" >> $@
 	@if which ccache > /dev/null 2>&1; then printf "ENABLE_CCACHE=1\nexport CCACHE_BASEDIR=$(abspath $(TOP)/..)\n" >> $@; echo "Found ccache on the system, enabling it"; fi
 	@if test -d $(TOP)/../maccore; then printf "ENABLE_XAMARIN=1\n" >> $@; echo "Detected the maccore repository, automatically enabled the Xamarin build"; fi
+
+include $(TOP)/Make.versions
 
 PACKAGE_HEAD_REV=$(shell git rev-parse HEAD)
 
 #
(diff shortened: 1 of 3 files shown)


Make.config
```diff
@@ -30,18 +28,12 @@ else
...
-IOS_PACKAGE_VERSION_REV=$(PACKAGE_VERSION_REV)
-
 IOS_PRODUCT=Xamarin.iOS
 IOS_PACKAGE_NAME=Xamarin.iOS
 IOS_PACKAGE_NAME_LOWER=$(shell echo $(IOS_PACKAGE_NAME) | tr "[:upper:]" "[:lower:]")
-# NEVER customize IOS_PACKAGE_VERSION itself, other parts (mtouch, web updater) are using the IOS_PACKAGE_VERSION_* variables
-IOS_PACKAGE_VERSION=12.2.$(IOS_PACKAGE_VERSION_REV).$(IOS_COMMIT_DISTANCE)
 IOS_PACKAGE_VERSION_MAJOR=$(word 1, $(subst ., ,$(IOS_PACKAGE_VERSION)))
 IOS_PACKAGE_VERSION_MINOR=$(word 2, $(subst ., ,$(IOS_PACKAGE_VERSION)))
+IOS_PACKAGE_VERSION_REV:=$(word 3, $(subst ., ,$(IOS_PACKAGE_VERSION)))
```
VincentDondain: `:=$`?

spouliot: it can be overridden=, e.g. `IOS_PACKAGE_VERSION_REV=9 make`
I don't think it's required here, but it's not harmful

Make.versions
```diff
@@ -0,0 +1,47 @@
...
+# Update version numbers on master as well, to the next version
+#
+
+#
+# ** Version numbers **
+#
+# Versions are defined as: Major.Minor.Revison.Build
+#
+# Major/minor (first/second numbers - max 2 digits each):
+# - Bump for major/minor releases.
```
VincentDondain: Might want to explain that for the minor version, even is for release and odd is for preview?

Make.versions
```diff
@@ -0,0 +1,47 @@
...
+# - Reset to 0 after a major or minor bump (do not use 99 for Xcode preview
+#   branches (use 0 instead), because otherwise we can't bump it further if
+#   needed).
+# - Bump for service releases and previews.
+# - Bump if commit distance becomes > 999.
+# - Can also be bumped for other reasons (in particular there's no correlation
+#   between Preview/Service Release #X and Revision #Y).
+#   - Bumping revision to a high enough number to make it clear that there's
+#     no correlation is a valid reason to bump.
+# - The revision must be bumped at the same time for both iOS and Mac
```
VincentDondain: So shouldn't we have one `rev` variable shared between iOS and Mac?

Make.versions
```diff
@@ -0,0 +1,47 @@
...
+#
+# Revision (third number - max 2 digits):
+# - Reset to 0 after a major or minor bump (do not use 99 for Xcode preview
+#   branches (use 0 instead), because otherwise we can't bump it further if
+#   needed).
+# - Bump for service releases and previews.
+# - Bump if commit distance becomes > 999.
+# - Can also be bumped for other reasons (in particular there's no correlation
+#   between Preview/Service Release #X and Revision #Y).
+#   - Bumping revision to a high enough number to make it clear that there's
```
VincentDondain: I don't understand this one. Past examples?

spouliot: we use a revision of `99` for Xcode previews

VincentDondain: But you said above we shouldn't do that? (:

Make.versions
spouliot: they don't have to be identical, e.g. if we merge `xcodeX` then we might only change XI revision (back to 0) but keep a high(er) revision for XM

Make.versions
spouliot: good idea, but that's needed in master (not in the backport) so it will need a different PR :)

Make.versions
VincentDondain: Oh ok so if we bump XI back to 0 the XM rev needs to be +1.

That's important. IMHO the comment should be divided in 2:

- Explanations of the various numbers and would _can_ be done
- TODO: things to check when changing any numbers

VincentDondain: Cause that one line for instance doesn't jump out but every time we update the package version we should re-read it to not forget (:

Make.versions
spouliot: It's generally a **minor** `99` (for major Xcode) or an odd minor number (for minor Xcode).
But there were cases where we're stuck in _odd_ positions and used `99` in revisions (e.g. when the next _even_ number was already used)- but it's not common and there are alternatives (so many numbers)

Make.versions
spouliot: We needed this _quickly_ since 15.9 numbers got broken/reset with xcode10 merge (and missing `0`) but that should not stop adding more documentation in master. Please prepare a PR :)

Make.config
rolfbjarne: `:=` means "calculate value once and store it in the variable"
`=` means "store string as-is, and evaluate it every time the variable is used"

The latter can be slow if the variable is used many times, in particular if it calls an external process.

PR #4937, opened 2018-10-05
[xharness] Verify that there's at least 1GB of space left before trying to build or run each test.

rolfbjarne:
If there's not enough space, then fail the test.

This might help https://github.com/xamarin/maccore/issues/965 (test run hangs).


rolfbjarne:
Test failure is known (https://github.com/xamarin/maccore/issues/701)

PR #4939, opened 2018-10-05
[CoreFoundation] Refactor Dispatch code to take advantage of recent code improvements.

rolfbjarne:
Make DispatchObject inherit from NativeObject, and change existing code to
adapt to and reuse code from NativeObject.


rolfbjarne:
Test failures are known (all due to https://github.com/xamarin/maccore/issues/1080).

API diff complain about acceptable changes: a new base class was introduced, and some members were moved there. API diff detects these changes as breaking changes, when they're not.

PR #4871, opened 2018-09-24
Report updateinfo for xamarin.ios and xamarin.mac in artifacts.json

cadsit:
We'd like to expose the `productId` and `versionId` in artifacts.json to autofill this information in while publishing so it's less of a manual process

PR #4943, opened 2018-10-05
Report updateinfo for xamarin.ios and xamarin.mac in artifacts.json (…

cadsit:
…#4871)

Expose the productId and versionId in artifacts.json to autofill this information in while publishing so it's less of a manual process.


rolfbjarne:
build

PR #4952, opened 2018-10-09
[xharness] Improve logging a bit.

rolfbjarne:
* Use timestamp in more log paths.
* Create numbered log subdirectories to make things nicer to look at in a
  terminal (thousands of subdirectories can be annoying to shift through; this
  way there's an additional subdirectory level).

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -81,7 +81,7 @@ namespace xharness
 			Devices.Harness = Harness;
 
 			if (SimulatorLoadLog == null)
-				SimulatorLoadLog = Logs.Create ("simulator-list.log", "Simulator Listing");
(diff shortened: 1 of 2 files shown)


tests/xharness/Jenkins.cs
```diff
@@ -2345,7 +2345,7 @@ public void CloneTestProject (TestProject project)
 
 		public string LogDirectory {
 			get {
-				var rv = Path.Combine (Jenkins.LogDirectory, $"{TestName}_{ID}");
+				var rv = Path.Combine (Jenkins.LogDirectory, $"{TestName}", $"{ID}");
```
spouliot: Idk if the compiler is smart enough to do it (or not) but `${TestName}` and `${ID}` are potentially more costly than using `TestName` and `ID` directly.

PR #4961, opened 2018-10-10
[tests] improve MonoWeakReferenceTest to avoid false pinning

lewurm:
Fixes an issue when executed with interpreter. Now, the test is closer
to what we do in the runtime test:

https://github.com/mono/mono/blob/831e73abcd5b9e10b4aaff7872b8d68e8e0f60fd/mono/tests/weak-fields.cs
https://github.com/mono/mono/blob/831e73abcd5b9e10b4aaff7872b8d68e8e0f60fd/mono/mini/TestHelpers.cs

### DIFF
--- a/tests/monotouch-test/mono/MonoWeakReferenceTest.cs
+++ b/tests/monotouch-test/mono/MonoWeakReferenceTest.cs
@@ -27,51 +27,85 @@ namespace MonoTouchFixtures {
 	[TestFixture]
 	[Preserve (AllMembers = true)]
 	public partial class WeakReferenceTest {
+		public static class FinalizerHelpers {
+			private static IntPtr aptr;
+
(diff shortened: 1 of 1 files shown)


lewurm:
Can someone verify if this still works on watchOS? This test has some problematic nature.

spouliot:
@rolfbjarne can you check (and if required fix) this on watchOS hardware ? thanks!

rolfbjarne:
The test (with these changes, I haven't tested without) work fine on watchOS device:

```
WeakReferenceTest
	[PASS] WeakReferenceTest.NoRetainCyclesExpectedTest
	[PASS] WeakReferenceTest.RetainCyclesExpectedTest
	[PASS] WeakReferenceTest.WeakTest
WeakReferenceTest : 24412 ms
```

PR #4968, opened 2018-10-11
[xharness] Generate a system variant of dont link, and run it using the oldest mono version we support in Jenkins. Fixes #4121.

### LINKED ISSUE #4121: Run XM tests with older system monos
Running an XM app using system mono _usually_ works if the system mono is older than the one used at build time.

Unfortunately right now we have no way of knowing _which_ older system mono's might work.

We should add tests (maybe reusing those running on older macOS versions) and run them with older system monos.

And then if we find that a particular mono version is required, we should bump the dependency as described here: #4120.

Reference: https://xamarinhq.slack.com/archives/C07MKNX1T/p1527156257000264
Reference: #4116

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/issues/4121.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -309,6 +309,24 @@ def runXamarinMacTests (url, macOS, maccore_hash, xamarin_macios_hash)
                     failedTests.add (t)
                 }
             }
+            // Run dontlink using the oldest system mono we support
+            def t = "dontlink (system)"
+            try {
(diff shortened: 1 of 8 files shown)


rolfbjarne:
> Just to clarify - this will be run _in addition_ on the extra _old_ macOS bots running XM tests right ? (and not on our main or PR bots)

Correct on all accounts.

PR #4973, opened 2018-10-12
[tests] make Symbols.FunctionNames aware of interpreter

lewurm:
Not sure what the original indent of this test case was. Maybe we need to do something else.

### DIFF
--- a/tests/monotouch-test/mono/Symbols.cs
+++ b/tests/monotouch-test/mono/Symbols.cs
@@ -26,7 +26,13 @@ namespace MonoTouchFixtures {
 				Assert.Ignore ("This is a device-only test.");
 			
 			Collect ();
-			Assert.That (symbols [1], Contains.Substring ("MonoTouchFixtures_Symbols_Collect"), "#1");
+			bool aot = symbols [1].Contains ("MonoTouchFixtures_Symbols_Collect");
+			/* ves_pinvoke_method (slow path) and do_icall (fast path) are
+			 * MONO_NEVER_INLINE, so they should show up in the backtrace
+			 * reliably */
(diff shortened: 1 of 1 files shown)


lewurm:
This the commit in the private repro that introduced it back then: https://github.com/xamarin/maccore/commit/773cc73f0ee52b3f2561a50006fea0109cb1941e

lewurm:
@rolfbjarne mono is supposed to print a managed stack trace in its crash handler. Alas this isn't part of the native crash handler, but the managed stack trace should be printed by mono right before it is taken over by the native crash handler if I understand correctly.

That said, maybe it's still a good idea to check for the interpreter function names, as it verifies that the runtime binary isn't stripped. Otherwise, what would be a good way to disable the test for the interpreter (keep in mind that we also have to handle mixed mode at some point)?

rolfbjarne:
@lewurm printing the managed stack trace to the console is useless for released apps, because app developers don't have access to their customers' consoles. The way the Apple ecosystem works is that app developers get the crash report as generated by iOS, and then that crash report is symbolicated using the dSYM created at build time.

I guess this is one (probably the) major downside of the interpreter: it'll be much more difficult to diagnose crashes that only occur on customer devices, because the crash report won't contain much information for managed code.

OTOH I don't see a way to fix this, so it's probably something we'll just have to document.

Regarding this specific test, I'd just check if the interpreter is enabled at all, and in that case disable the test (is it possible to somehow determine at runtime if the interpreter is enabled? By the presence of a file in the app, or a symbol in memory, for instance?)

lewurm:
@rolfbjarne I didn't know about crash reports not containing console output. This is unfortunate and indeed a major drawback 🙁 

There is no API in the BCL that can be used to detect if the interpreter is enabled from managed code. Given that, I think the proposed change in this PR would be okay to detect the interpreters presence.

PR #4982, opened 2018-10-12
[uikit] Duplicate bindings for 'UIScrollView.ContentOffset' to get correct availability information. Fixes #4893

### LINKED ISSUE #4893: ContentOffset only available in iOS 12 or newer
### Steps to Reproduce

1. Create a `UIScrollView` and access `ContentOffset` property.

### Expected Behavior

Don't show any IDE hint.

### Actual Behavior
Shows IDE hint saying that `ContentOffset` property is only available in iOS 12 or newer.

screen shot 2018-09-29 at 12 23 15 am

### Environment

VS for mac Version 7.6.7 (build 49)

Xamarin.iOS
Version: 12.0.0.15 (Visual Studio Community)
Hash: 84552a46
Branch: xcode10
Build date: 2018-09-17 21:54:33-0400

screen shot 2018-09-29 at 12 18 23 am

spouliot:
`ContentOffset` was moved to `UIFocusItemScrollableContainer` in iOS 12.
Since it's a new protocol it's decorated as introduced in iOS 12 - but
the API was actually already available for a long time.

The _trick_ is to keep the original binding inside `UIScrollView`
so it overrides the protocol and keeps the correct availability info.
It requires adding `new` to bindings to avoid compiler warnings.

reference: https://github.com/xamarin/xamarin-macios/issues/4893

### DIFF
--- a/src/uikit.cs
+++ b/src/uikit.cs
@@ -10040,6 +10040,10 @@ namespace UIKit {
 		[Export ("initWithFrame:")]
 		IntPtr Constructor (CGRect frame);
 
+		// moved to UIFocusItemScrollableContainer in iOS 12 - but that makes the availability information incorrect (so `new` is used to avoid compiler warnings)
+		[Export ("contentOffset")]
+		new CGPoint ContentOffset { get; set; }
+
 		[Export ("contentSize")]
 		new CGSize ContentSize { get; set; }
 


rolfbjarne:
@monojenkins backport master

Issue #4975, opened 2018-10-12
Port mmp regression tests to "normal" unit tests for XM


spouliot:
Current mmp regressions tests are old and using the deprecated classic profile - so no 64 bits support (which means no Xcode10+ or macOS 10.14 can build/run them).

Most of the tests can be moved into common unit tests in the projects created in 
https://github.com/xamarin/xamarin-macios/pull/3234

- [x] ~~link-gdiplus-1 #6497~~
- [x] ~~link-gdiplus-2 #6497~~
- [x] link-posix-1 #6509
- [x] link-posix-2 #6509
- [x] link-system.web-icalls #6500
- [x] link-keep-resources-1 #6527
- [x] link-keep-resources-2 #6533
- [x] link-preserve-assembly #6541
- [x] link-uithread-1 #6546
- [x] link-uithread-2 #6546
- [x] link-safe-1 #6556
- [x] link-frameworks-1 #6543
- [x] link-aes-1 #6501
- [x] link-blocks-1 #6536
- [x] ~~link-preserve-calendar-1 https://github.com/xamarin/xamarin-macios/pull/4983~~
- [x] ~~link-preserve-calendar-2 https://github.com/xamarin/xamarin-macios/pull/4983~~
- [x] ~~link-typedescriptor-1 https://github.com/xamarin/xamarin-macios/pull/4994~~
- [x] ~~link-typedescriptor-2 https://github.com/xamarin/xamarin-macios/pull/4994~~
- [x] embedded-mono
- [x] system-mono
- [x]	embedded-mono-profile
- [x] system-mono-profile

DISABLED_TESTS_MONO_4_2

- [x] link-data-contract-1 #6535
- [x] link-webclient-2 #6557
- [x] link-xml-serialization #6555

DISABLED_TESTS_MONO_4_4

- [x] link-webclient-1 #6557
- [x] link-webclient-3 #6557
- [x]	link-remove-attributes-1 #6547

TESTS_UNIFIED

- [x] custom-bundle-name

TESTS_4_5

- [x] link-callfilepath-45 #6539

DISABLED_4_5_TESTS_MONO_4_2

- [x] link-httpclient-async #6554

spouliot:
ref: 
https://github.com/xamarin/xamarin-macios/pull/4596/commits/d063c531f60e064fed5493ca774284963471338c

rolfbjarne:
All the unit tests have been ported to Unified projects or added to the LinkAll and/or LinkSdk tests.

PR #4983, opened 2018-10-12
[tests][mmp] Convert old mmp calendar regression tests to unit tests

spouliot:
part of https://github.com/xamarin/xamarin-macios/issues/4975


mandel-macaque:
@spouliot seems that the failure is with one of the tests: LinkSdkTests.LinkSdkTest.Calendars: UmAlQuraCalendar

PR #4985, opened 2018-10-15
Fix the llvm32 build by using Xcode 9.4.

rolfbjarne:
Since Xcode 10 doesn't support building 32-bit macOS binaries.

The targets don't quite make sense (`build-llvm32` builds both 32-bit and 64-bit llvm for instance), but this seemed like the simplest change that wouldn't run into a long test-fail-fix cycle of make changes.

Ideally this wouldn't be that critical, since llvm is not supposed to be built but provisioned, but that's also broken.


rolfbjarne:
CC @luhenry

baulig:
I ran into the same issue while doing the 2018-10 integration.

rolfbjarne:
> does not seem related

it's not, and if the build got this far, it means the llvm brokenness is fixed.

PR #4996, opened 2018-10-16
[gamecontroller] Update for Xcode 10.1 beta 2

VincentDondain:


src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -47,6 +47,11 @@ public struct GCExtendedGamepadSnapShotDataV100 {
 		public float /* float_t = float */ LeftTrigger;
 		public float /* float_t = float */ RightTrigger;
 
+		[TV (12, 0), Mac (10, 14, onlyOn64: true), iOS (12, 0)]
```
rolfbjarne: Shouldn't the version in the attributes be 12.1 and not 12.0 (and 10.14.1 for macOS)?

VincentDondain: Mmh indeed it looks like intro isn't happy about those selectors on iOS 12.0. Sometimes it works though (private APIs made public).

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -47,6 +47,11 @@ public struct GCExtendedGamepadSnapShotDataV100 {
 		public float /* float_t = float */ LeftTrigger;
 		public float /* float_t = float */ RightTrigger;
 
+		[TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
+		public bool LeftThumbstickButton;
+		[TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
+		public bool RightThumbstickButton;
```
spouliot: That's a `struct` so it's memory representation must be identical between managed and native. 

That's not the case here. That's an Apple issue (please file rdar) since the fields where added in the **middle** of the structure, i.e. code compiled with the 12.1 headers won't match the runtime memory representation on iOS 12 (some fields will be shifted).

VincentDondain: Yea that makes total sense. Will try to keep size/alignment concerns in mind next time I update a struct (:

VincentDondain: https://trello.com/c/7FoGTORD

spouliot:
@VincentDondain you'll need to bump the max macOS version supported
```
[FAIL] 10.14.1 > 10.14 (Max) on 'GameController.GCControllerButtonInput get_LeftThumbstickButton()'.
```

Issue #4936, opened 2018-10-05
Class.LookupClass is slow


rolfbjarne:
See profile here: https://github.com/xamarin/xamarin-macios/issues/4923#issuecomment-427175183, where Class.LookupClass is a significant portion of the timing.

The problem is that it ends up calling Assembly.GetName (), which recomputes the name every time, and is quite expensive to compute (as can be seen from the profile).

Ideas:

* Cache the IntPtr -> Type lookup in a dictionary.
* Make Assembly.GetName () cache the value (or cache it ourselves somehow).

spouliot:
IIRC there was (before ref / corefx import) some caching in `Assembly.GetName ()` but it might have been wrong in some cases (ref emit). Still that might have caused a slow down.

It might make sense to cache it ourselves, even more if we're calling it inside other time sensitive code paths.

Still using a _lookup_ inside `Class.LookupClass` does make a lot of sense 😸

rolfbjarne:
I've implemented 4 different optimizations (PRs #5009, #5013, #5016, #5017), which improved object creation by 69-86% (3-7x faster).

I also implemented a fifth optimization (create a factory class that can create object instances without using reflection), but it's much more invasive (and would bloat the assemblies a lot). The POC (with numbers) is [here](https://github.com/rolfbjarne/xamarin-macios/commit/c6c9e2a8ba2fe5ae8ca8b38027c84b902473fa7b). This optimization has good results (40% faster object creation), but due to its invasiveness I've discarded it for now.

rolfbjarne:
I've implemented 4 different optimizations (PRs #5009, #5013, #5016, #5017), which improved object creation by 69-86% (3-7x faster).

I also implemented a fifth optimization (create a factory class that can create object instances without using reflection), but it's much more invasive (and would bloat the assemblies a lot). The POC (with numbers) is [here](https://github.com/rolfbjarne/xamarin-macios/commit/c6c9e2a8ba2fe5ae8ca8b38027c84b902473fa7b). This optimization has good results (40% faster object creation), but due to its invasiveness I've discarded it for now.

spouliot:
@rolfbjarne awesome :) please open an RFC on your fifth optimization that we don't forget about it so before we have a chance to discuss it (and possible variations)

rolfbjarne:
RFC: https://github.com/xamarin/xamarin-macios/issues/5025

PR #5009, opened 2018-10-18
[Class] Sort our array of Class -> token references so that we can do binary instead of linear searches in it.

rolfbjarne:
Our type map has two sections: first come all the wrapper types, then all the
custom types. This means we need to sort these two sections separately, since
code elsewhere depends on this split in order to determine if a type is a
custom type or not.

We also need a minor modification in the array of skipped types, since it
contained indexes into the type map, which won't be valid after is has been
sorted. Instead store a type reference for the actual type in the array, and
use that to search the type map for the corresponding Class. This is a little
bit slower, but the results are cached in a dictionary, so it'll only happen
once for each type.

The performance is slightly slower when the type map has very few entries, but
that is repaid many times over when the number of entries go up.

Numbers
=======

Test case: https://github.com/rolfbjarne/TestApp/commit/004283d7b628a29fcf711d98d8842bfd4ef4393b

iPad Air 2
----------

| Configuration       | Before | After  | Improvement   |
| ------------------- | ------ | ------ | ------------: |
| Release (link all)  | 477 ms | 481 ms | -4 ms (-0,8%) |
| Release (dont link) | 738 ms | 656 ms |   82 ms (11%) |

iPhone X
--------

| Configuration       | Before | After  | Improvement |
| ------------------- | ------ | ------ | ----------: |
| Release (link all)  |  98 ms |  99 ms | -1 ms (-1%) |
| Release (dont link) | 197 ms | 153 ms | 44 ms (22%) |

When linking all assemblies, the type map has 24 entries, and when not linking
at all it has 2993 entries.

This is part 1 of multiple fixes for #4936.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -1001,11 +1001,27 @@ object_queued_for_finalization (MonoObject *object)
  * Registration map
  */ 
 
+static int
+compare_mtclassmap (const void *a, const void *b)
+{
+	MTClassMap *mapa = (MTClassMap *) a;
+	MTClassMap *mapb = (MTClassMap *) b;
+	if (mapa->handle == mapb->handle)
+		return 0;
+	if ((intptr_t) mapa->handle < (intptr_t) mapb->handle)
+		return -1;
+	return 1;
+}
+
 void
 xamarin_add_registration_map (struct MTRegistrationMap *map)
 {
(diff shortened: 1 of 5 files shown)


spouliot:
I think it's fair to mention the **link all** scenario is with a very minimal app - and that anything useful would be a net gain in performance.

PR #5021, opened 2018-10-22
Use the zlib-helper.o library built by mono, instead of building our own.

rolfbjarne:
This also makes it possible to simplify/remove some of our build logic.


spouliot:
build

rolfbjarne:
build

PR #5030, opened 2018-10-23
[2018-08][watchos] Use mono_dangerous_add_raw_internal_call for watchOS icalls

lambdageek:
With mono/mono@5756ba4 we added a new `mono_dangerous_add_raw_internal_call` function to register new icalls - icalls added with `mono_add_internal_call` run in GC Safe mode; icalls added with the new API run in GC Unsafe mode.

For hybrid suspend this means that icalls added with `mono_add_internal_call` can continue to be oblivious to the thread state (GC Safe / GC Unsafe) and can continue to make blocking calls, run for unbounded amounts of time without safepointing, etc.

For coop and hybrid suspend icalls added with `mono_dangerous_add_raw_internal_call` have to be thread state aware - they start in GC Unsafe mode and must switch to GC Safe around OS calls, blocking operations, or loops that run for a long time without safepointing.

This PR ensures that on watchOS icalls are registered using the new function so that they continue to run in GC Unsafe mode.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -194,6 +194,25 @@ struct Managed_NSObject {
 	uint8_t flags;
 };
 
+static void
+xamarin_add_internal_call (const char *name, const void *method)
+{
+	/* COOP: With cooperative GC, icalls will run, like manageed methods,
+	 * in GC Unsafe mode, avoiding a thread state trandition.  In return
+	 * the icalls must guarantee that they won't block, or run indefinitely
+	 * without a safepoint, by manually performing a transition to GC Safe
+	 * mode.  With backward-compatible hybrid GC, icalls run in GC Safe
+	 * mode and the Mono API functions take care of thread state
+	 * transitions, so don't need to perform GC thread state transitions
+	 * themselves.
+	 *
+	 */
+	if (xamarin_is_gc_coop)
+		mono_dangerous_add_raw_internal_call (name, method);
+	else
+		mono_add_internal_call (name, method);
+}
+
 id
 xamarin_get_nsobject_handle (MonoObject *obj)
 {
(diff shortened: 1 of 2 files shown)


runtime/runtime.m
```diff
@@ -194,6 +194,27 @@
...
+	 * without a safepoint, by manually performing a transition to GC Safe
+	 * mode.  With backward-compatible hybrid GC, icalls run in GC Safe
+	 * mode and the Mono API functions take care of thread state
+	 * transitions, so don't need to perform GC thread state transitions
+	 * themselves.
+	 *
+	 * Currently cooperative GC is only used by watchOS targets.
+	 */
+#ifdef TARGET_OS_WATCH
+	mono_dangerous_add_raw_internal_call (name, method);
```
rolfbjarne: Any reason we can't just call `mono_dangerous_add_raw_internal_call` always, and not only for watchOS? The icalls are already thread state aware.

This would make it easier when more platforms move to coop (one less place to fix).

lambdageek: Up to you.
Unconditionally turning on `MONO_ENTER_GC_SAFE` et al. on all platforms seemed like a bigger change.

rolfbjarne: What happens if we don't turn on `MONO_ENTER_GC_SAFE` on a platform and still call `mono_dangerous_add_raw_internal_call`?

lambdageek: Hybrid suspend will hang - this was an observed hange in VSfM, for example, with hybrid suspend https://github.com/mono/mono/issues/11138:
- suppose thread A takes `xamarin_framework_peer_lock()` and is suspended
- now suppose another thread B is at https://github.com/xamarin/xamarin-macios/blob/f4dbb6840ae6d908523ae69f5b5a01c75228fc88/runtime/runtime.m#L1998-L1999 - it is in GC Unsafe so the runtime wants to cooperatively suspend it, but B is waiting for the lock
- there's now a three-way deadlock: A holds the peer lock and is suspended; B is waiting for the peer lock; thread suspend initiator (the GC) is waiting for B (and won't resume A)

rolfbjarne: OK, so that's bad idea then.

In that case I'd prefer to use the `xamarin_is_gc_coop` variable to determine which method to call instead: https://github.com/xamarin/xamarin-macios/blob/b403da0d85c83558eabb326cef3621190059bf66/runtime/runtime.m#L74

marek-safar:
@rolfbjarne could you review/approve the change?

lambdageek:
rebased on top of latest changes on the `mono-2018-08` branch

PR #5034, opened 2018-10-24
[introspection] Fix a few issues found on iPad Air 2 with iOS 12.

rolfbjarne:


tests/introspection/iOS/iOSApiCtorInitTest.cs
```diff
@@ -73,7 +73,10 @@ protected override bool Skip (Type type)
 					return true;
 				break;
 #endif // !__WATCHOS__
-			case "CoreNFC": // Only available on device
+			case "CoreNFC": // Only available on devices that support NFC, so check if NFCNDEFReaderSession is present.
+				if (Class.GetHandle ("NFCNDEFReaderSession") == IntPtr.Zero)
```
spouliot: I wonder if we should expose this as a `static public bool IsAvailable { get; }` inside CoreNFC (and re-use it here)

rolfbjarne: Apple already thought about that, and helpfully added [NFCNDEFReaderSession.readingAvailable](https://developer.apple.com/documentation/corenfc/nfcndefreadersession/2915854-readingavailable).

Then they made it useless by _not including the framework at all_ on some devices 🤦‍♂️ [1] 

I propose we add `NFCNdefReaderSession.IsReallyAvailable`.

[1] See also https://stackoverflow.com/a/52855961/183422.

spouliot: hmm... maybe we can merge both ?
```
if (Class.GetHandle ("NFCNDEFReaderSession") == IntPtr.Zero)
    return false;
return internalReadingAvailable;
```
so our version is more usable than Apple's one. Adding a new API would make it's discovery harder (and confusing)

rolfbjarne: So `NFCNdefReaderSession.ReadingAvailable` isn't usable by monotouch-test, because not only does it check if the device support NFC, it also checks if the app is allowed to use it, which requires a specific entitlement, which requires a custom provisioning profile, which becomes a pain for monotouch-test. So let's not do that.

Then it turns out `NFCNdefReaderSession.ReadingAvailable` actually works _even if the framework doesn't exist_, because of an Objective-C "feature": sending a message to NULL returns NULL, i.e. 0, i.e. false.

So this code successfully returns false:

```csharp
return Messaging.bool_objc_msgSend (Class.GetHandle ("NFCNDEFReaderSession"), Selector.GetHandle ("readingAvailable"));
```

because `Class.GetHandle ("NFCNDEFReaderSession")` returns IntPtr.Zero, which means sending the message to NULL, which returns NULL/0/false.

This means there's no need to modify our binding, because it already works.

spouliot: 👍

PR #5035, opened 2018-10-24
[msbuild] Always set IOSDebugOverWiFi to true for WatchOS apps/extens…

jstedfast:
…ions

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/712824

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.WatchOS.AppExtension.Common.targets
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.WatchOS.AppExtension.Common.targets
@@ -18,6 +18,7 @@ Copyright (C) 2015-2016 Xamarin. All rights reserved.
 	<PropertyGroup>
 		<IsAppExtension>True</IsAppExtension>
 		<IsWatchExtension>True</IsWatchExtension>
+		<IOSDebugOverWiFi>True</IOSDebugOverWiFi>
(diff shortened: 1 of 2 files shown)


msbuild/Xamarin.iOS.Tasks.Core/Xamarin.WatchOS.AppExtension.Common.targets
```diff
@@ -18,6 +18,7 @@ Copyright (C) 2015-2016 Xamarin. All rights reserved.
 	<PropertyGroup>
 		<IsAppExtension>True</IsAppExtension>
 		<IsWatchExtension>True</IsWatchExtension>
+		<IOSDebugOverWiFi>True</IOSDebugOverWiFi>
```
spouliot: the extension runs on the phone so the normal (USB) debugging should be fine right ?

jstedfast: Are you sure? I thought the extensions ran on the Watch now.

WatchOS1 extensions ran on the iPhone, but I thought starting with 2.0, they ran on the watch.

spouliot: you're right, the app is a shell

PR #5052, opened 2018-10-29
NSUrlSessionHandler - basic auth and multiple auth headers

mrward:
If the challenge is for basic authentication then try to get the
credentials for this authentication type.

Also handle multiple WWW-Authenticate headers, such as:

    WWW-Authenticate: Negotiate,  NTLM

If the credentials provided on the NSUrlSessionHandler are not
available for a specific authentication type, or if the authentication
method is not supported, then reject the protection space. This allows
the next authentication challenge to be sent to the DidReceiveChallenge
method. Also handle a null reference exception that could occur if
there were no credentials available for the authentication type.

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -380,17 +380,46 @@ namespace Foundation {
 					}
 				}
 
-				if (challenge.ProtectionSpace.AuthenticationMethod == NSUrlProtectionSpace.AuthenticationMethodNTLM && sessionHandler.Credentials != null) {
-					var credentialsToUse = sessionHandler.Credentials as NetworkCredential;
-					if (credentialsToUse == null) {
+				if (sessionHandler.Credentials != null && TryGetAuthenticationType (challenge.ProtectionSpace, out string authType)) {
+					NetworkCredential credentialsToUse = null;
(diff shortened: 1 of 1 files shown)


mrward:
Mono supports NTLM, Basic and Digest. Digest seems to be less used than Basic auth from what I have read. Not sure if we need Xamarin.Mac to support Digest too. Not managed to get Digest enabled on a Windows server website, but did not spend a lot of time looking into this.

spouliot:
@mandel-macaque can you test those changes ? thanks!

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -380,18 +380,47 @@ public override void DidReceiveChallenge (NSUrlSession session, NSUrlSessionTask
...
+					// Want to reject this authentication type to allow the next authentication method in the request to
+					// be used.
+					authenticationType = RejectProtectionSpaceAuthType;
+				} else {
+					// ServerTrust, ClientCertificate or Default.
+					authenticationType = null;
+					return false;
+				}
+				return true;
+			}
```
mandel-macaque: Can't you use a switch here? I will be easier to read, right?

mandel-macaque:
@spouliot for I was able to test agains a real server. Everything works ok.

PR #5053, opened 2018-10-29
[msbuild/tests] Add tests for issue 3199

### LINKED ISSUE #3810: [msbuild] Add msbuild test for #3791: Skip reference assemblies passed to mtouch

VincentDondain:
- Fixes #3810: [msbuild] Add msbuild test for #3791: Skip reference assemblies passed to mtouch
  (https://github.com/xamarin/xamarin-macios/issues/3810)
- MSBuild tests now build the iOS and Mac test cases for issue 3199 that's using a nuget package as "PackageReference" which was causing some issues.
  With the fix disabled we get:

```
Errors and Failures:
1) Test Failure : Xamarin.iOS.Tasks.Issue3199.BuildTest
     #RunTarget-ErrorCount
	Could not AOT the assembly '/Users/vidondai/Documents/xi-master/xamarin-macios/msbuild/tests/Issue3199/obj/iPhone/Debug/mtouch-cache/3-Build/System.Runtime.CompilerServices.Unsafe.dll'
  Expected: 0
  But was:  1
```

and

```
"/Users/vidondai/Documents/xi-master/xamarin-macios/tests/common/mac/TestProjects/Issue3199/Issue3199.csproj" (default target) (1) ->
(_CompileToNative target) ->
MMP : error MM3001: Could not AOT the assembly '/Users/vidondai/Documents/xi-master/xamarin-macios/tests/common/mac/TestProjects/Issue3199/bin/Debug/Issue3199.app/Contents/MonoBundle/System.Runtime.CompilerServices.Unsafe.dll' [/Users/vidondai/Documents/xi-master/xamarin-macios/tests/common/mac/TestProjects/Issue3199/Issue3199.csproj]

63 Warning(s)
1 Error(s)
```


VincentDondain:
**TL;DR;** I hear you but I believe what you're asking for is an _enhancement request_ on its own (I'm happy to investigate) and this PR, as is, is how those things have always been tested. So I think it's worth merging it IMHO.

> Adding full projects (both iOS and Mac) won't scale, e.g. .if we have to change a single file (like Contents.json) for a new Xcode we'll end up with dozens of files to update.

@spouliot I agree but you're questioning almost all msbuild tests with that reasoning. This is how things are done _today_ (not that I like it), you add full projects and build them and yes it requires adding a bunch of redundant files.
I also made the argument in #mac today that the msbuild tests for iOS and Mac should be merged and that would avoid some of the duplication.

There might be a slightly better approach possible based on some Mac msbuild tests that are using a shared csproj but the fact we need to restore nugget packages might be a deal breaker. See @chamons comment in https://xamarinhq.slack.com/archives/C03CCJHCF/p1540827269024100.

_Note: The only msbuild tests that don't require full projects are the **tasks** tests that are more "classic" unit tests._

> it hides what's the real content of the test

I could add a "test comment" to explain what it does or even rename it so it's clearer.

spouliot:
@VincentDondain nothing personal :) Build and test times were not priorities until recently. Now that we're working on (both fixing and measuring) them it means further work must not undo our gains - which could be easy since testing msbuild has been unblocked.

PR #5055, opened 2018-10-30
Update for Xcode 10.1 final release, including last minute Pen/Tap API

spouliot:


src/uikit.cs
```diff
@@ -18433,4 +18433,45 @@ interface UIFontMetrics {
 #endif // !WATCH
 	}
 
+#if !WATCH
```
VincentDondain: Minor but could we not have avoided that with empty skeletons like we usually do?

spouliot: yes, but the errors were on an existing type and I did not want to PR to touch unrelated code.

spouliot:
build

spouliot:
* watchOS - simulator (Crashed; Succeeded; Ignored) -> HE0038 known random issue -> 
https://github.com/xamarin/maccore/issues/581

spouliot:
Most of the failures are now also reported on `master` - which is to be expected since they are related to the bot upgrade to 10.14. I'm merging this one to unblock other release tasks.

PR #5056, opened 2018-10-30
added ScreenImage overload for Xamarin Mac allowing for more options

msioen:
Added an extra overload to the Xamarin.Mac ScreenShot functionality allowing to pass additional options

### DIFF
--- a/src/CoreGraphics/CGImage.cs
+++ b/src/CoreGraphics/CGImage.cs
@@ -217,13 +217,14 @@ namespace CoreGraphics {
         
 		public static CGImage ScreenImage (int windownumber, CGRect bounds)
 		{
-			return ScreenImage (windownumber, bounds, CGWindowListOption.IncludingWindow);
+			return ScreenImage (windownumber, bounds, CGWindowListOption.IncludingWindow, CGWindowImageOption.Default);
 		}
 
-		public static CGImage ScreenImage (int windownumber, CGRect bounds, CGWindowListOption windowOption)
+		public static CGImage ScreenImage (int windownumber, CGRect bounds, CGWindowListOption windowOption,
+			CGWindowImageOption imageOption)
 		{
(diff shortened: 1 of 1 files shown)


src/CoreGraphics/CGImage.cs
```diff
@@ -216,12 +216,17 @@ protected virtual void Dispose (bool disposing)
 		static extern IntPtr CGWindowListCreateImage(CGRect screenBounds, CGWindowListOption windowOption, uint windowID, CGWindowImageOption imageOption);
         
 		public static CGImage ScreenImage (int windownumber, CGRect bounds)
-		{                    
-			IntPtr imageRef = CGWindowListCreateImage(bounds, CGWindowListOption.IncludingWindow, (uint)windownumber,
+		{
+			return ScreenImage (windownumber, bounds, CGWindowListOption.IncludingWindow);
+		}
+
+		public static CGImage ScreenImage (int windownumber, CGRect bounds, CGWindowListOption windowOption)
```
spouliot: I wonder if the `imageOption` should not be exposed too ?

spouliot:
build

src/CoreGraphics/CGImage.cs
msioen: Initially just added the window options because that's what I'm using :-) but I guess it would indeed be cleaner to expose both options in this overload. I'll update the pr if you want

spouliot: I think it make sense but you can wait for more comments and let the bot complete this first around of testing :)

msioen: pushed an update including the imageOption

msioen:
Not used to pull request comments on github, apparently resolving the comment removes it all together?
=> pushed an update with imageOption added as well

spouliot:
@msioen _Resolving_ is a new option and we have not really used much it since we want feedback to be visible to other reviews. But it does not matter much for smaller contributions so it's fine :)

PR #5059, opened 2018-10-31
[jenkins] Remove any system versions of XI/XM to make absolutely sure nothing in our build requires them.

rolfbjarne:
Completely remove XI/XM from the system when provisioning a bot, so that we can be absolutely sure nothing in our build and/or test setup tries to use them.


rolfbjarne:
> Is this related to any issue / event ?

Yes: https://github.com/xamarin/xamarin-macios/pull/4596#issuecomment-434644864

PR #5054, opened 2018-10-30
[msbuild] VS Incremental build fixes

emaf:
Most of these changes are needed from VS to fix incremental build issues.

The problem here is VS runs MSBuild on Windows and remotes (most of) the task executions to the Mac. Since MSBuild is running on Windows the inputs and outputs are checked there, but the output files won't be created on Windows unless those are explicitly declared as output ITaskItems of a task. VS don't copy every file created on the Mac back to Windows because that will increase the build time unnecessarily.

For instance, the _GenerateFrameworkDebugSymbols target was using the Info.plist file created by the dsymutil tool as output, but that file was not declared as ITaskItem output of the DsymUtil task so VS didn't know that file should be created on Windows.

This doesn't mean every task should have an output property declaring ITaskItems, but if you're writing a target that will use certain files as output those files should be output of one of the tasks that target is running.

Partial fix for https://dev.azure.com/devdiv/DevDiv/_workitems/edit/710309.

PR #5064, opened 2018-10-31
[msbuild] VS Incremental build fixes

emaf:
Backport of PR #5054

Most of these changes are needed from VS to make incremental builds work.

The problem here is VS runs MSBuild on Windows and remotes (most of) the task executions to the Mac. Since MSBuild is running on Windows the inputs and outputs are checked there, but the output files won't be created on Windows unless those are explicitly declared as output ITaskItems of a task. VS don't copy every file created on the Mac back to Windows because that will increase the build time unnecessarily.

For instance, the _GenerateFrameworkDebugSymbols target was using the Info.plist file created by the dsymutil tool as output, but that file was not declared as ITaskItem output of the DsymUtil task so VS didn't know that file should be created on Windows.

This doesn't mean every task should have an output property declaring ITaskItems, but if you're writing a target that will use certain files as output those files should be output of one of the tasks that target is running.

Partial fix for https://dev.azure.com/devdiv/DevDiv/_workitems/edit/710309.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
@@ -1,6 +1,7 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Text;
+using System.Linq;
 using System.Diagnostics;
 
 using Parallel = System.Threading.Tasks.Parallel;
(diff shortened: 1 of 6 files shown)


spouliot:
The errors are _fine_ and only happens on 10.14 bots (after upgrade). We have a fix for them but `d16-0-p1` is frozen and requires QB approval at this time.

spouliot:
@emaf without QB approval this won't be in P1 (only P2, re-branched from master). There's not much time remaining if you want a freeze exception on this one.

PR #5074, opened 2018-11-02
[ImageCaptureCore] Add ImageCaptureCore Bindings

dalexsoto:
Fixes xamarin/xamarin-macios#4273

Sample available @ https://github.com/dalexsoto/ScannerSample


src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		void RequestOpenSession ();
+
+		[Export ("requestCloseSession")]
+		void RequestCloseSession ();
+
+		[Export ("requestYield")]
+		void RequestYield ();
+
+		[Export ("requestSendMessage:outData:maxReturnedDataSize:sendMessageDelegate:didSendMessageSelector:contextInfo:")]
+		unsafe void RequestSendMessage (uint messageCode, NSData data, uint maxReturnedDataSize, NSObject sendMessageDelegate, Selector selector, [NullAllowed] IntPtr contextInfo);
```
rolfbjarne:
```suggestion
	void RequestSendMessage (uint messageCode, NSData data, uint maxReturnedDataSize, NSObject sendMessageDelegate, Selector selector, [NullAllowed] IntPtr contextInfo);
```

`unsafe` is not needed.

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		void CancelDownload ();
+
+		[Export ("requestUploadFile:options:uploadDelegate:didUploadSelector:contextInfo:")]
+		void RequestUploadFile (NSUrl fileUrl, [NullAllowed] NSDictionary<NSString, NSObject> options, NSObject uploadDelegate, Selector didUploadSelector, [NullAllowed] IntPtr contextInfo);
+
+		[Export ("requestReadDataFromFile:atOffset:length:readDelegate:didReadDataSelector:contextInfo:")]
+		void RequestReadDataFromFile (ICCameraFile file, long offset, long length, NSObject readDelegate, Selector didReadDataSelector, [NullAllowed] IntPtr contextInfo);
+
+		[Export ("requestSendPTPCommand:outData:sendCommandDelegate:didSendCommandSelector:contextInfo:")]
+		void RequestSendPtpCommand (NSData command, NSData outData, NSObject sendCommandDelegate, Selector didSendCommandSelector, [NullAllowed] IntPtr contextInfo);
```
rolfbjarne: For the Request* methods that take a Delegate + Selector + Context Info we might want to provide a different overload that takes a single delegate parameter instead:

```csharp
delegate void DidSendPtpCommand (NSData command, NSData inData, NSData response, NSError error);
public void RequestSendPtpCommand (NSData command, NSData outData, DidSendPtpCommand callback)
{
    // Magic goes here
}
```

src/imagekit.cs
```diff
@@ -199,19 +197,17 @@ interface IKDeviceBrowserView {
...
 
 	[BaseType (typeof (NSObject))]
 	[Model]
 	[Protocol]
 	interface IKDeviceBrowserViewDelegate {
-//		FIXME need ImageCaptureCore;
-//		[Abstract]
-//		[Export ("deviceBrowserView:selectionDidChange:"), EventArgs ("IKDeviceBrowserViewICDevice")]
-//		void SelectionDidChange (IKDeviceBrowserView deviceBrowserView, ICDevice device);
+		[Abstract]
```
rolfbjarne:
```suggestion
#if XAMCORE_4_0
		[Abstract]
#endif
```

New members in existing protocols can't be `[Abstract]`, it's a breaking change.

src/ImageCaptureCore/Defs.cs
```diff
@@ -0,0 +1,287 @@
...
+		Local = 256,
+		Shared = 512,
+		Bonjour = 1024,
+		Bluetooth = 2048,
+	}
+
+	// Combination of ICDeviceTypeMask and ICDeviceLocationTypeMask
+	[Native]
+	[Flags]
+	public enum ICBrowsedDeviceType : ulong {
```
spouliot: add `None = 0,`
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1008-enums-should-have-zero-value?view=vs-2017

src/ImageCaptureCore/Defs.cs
```diff
@@ -0,0 +1,287 @@
...
+
+	[Native]
+	public enum ICScannerColorDataFormatType : ulong {
+		Chunky = 0,
+		Planar = 1,
+	}
+
+	[Native]
+	public enum ICScannerPixelDataType : ulong {
+		Bw = 0,
```
spouliot: Two letter acronyms should be in capitalized.
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1709-identifiers-should-be-cased-correctly?view=vs-2017

src/ImageCaptureCore/Defs.cs
```diff
@@ -0,0 +1,287 @@
...
+		ApsP = 75,
+		Standard35 = 76,
+		Mf = 77,
+		Lf = 78,
+	}
+
+	public enum ICScannerFunctionalUnitState : uint {
+		Ready = (1u << 0),
+		ScanInProgress = (1u << 1),
+		OverviewScanInProgress = (1u << 2),
```
spouliot: that looks like a `[Flag]`

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		NSString NotificationKey { get; }
+
+		[Field ("ICStatusCodeKey")]
+		NSString CodeKey { get; }
+
+		[Field ("ICLocalizedStatusNotificationKey")]
+		NSString LocalizedNotificationKey { get; }
+	}
+
+	[Static]
```
spouliot: maybe a comment (for other present/future reviewers) that this is used in a delegate so an enum is not possible

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		ICDeviceType Type { get; }
+
+		[NullAllowed, Export ("name")]
+		string Name { get; }
+
+		[NullAllowed, Export ("icon")]
+		CGImage Icon { get; }
+
+		[Export ("capabilities")]
+		string[] Capabilities { get; }
```
spouliot: if that's related to `ICDeviceCapabilities` then it should be exposed as `NSString[]` otherwise if will force conversions between the types.

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		int UsbProductId { get; }
+
+		[Export ("usbVendorID")]
+		int UsbVendorId { get; }
+
+		[Export ("fwGUID")]
+		long FwGuid { get; }
+
+		[NullAllowed, Export ("serialNumberString")]
+		string SerialNumberString { get; }
```
spouliot: not sure the `String` suffix is needed

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		int UsbLocationId { get; }
+
+		[Export ("usbProductID")]
+		int UsbProductId { get; }
+
+		[Export ("usbVendorID")]
+		int UsbVendorId { get; }
+
+		[Export ("fwGUID")]
+		long FwGuid { get; }
```
spouliot: any clue what `Fw` means ? (forward?)

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+		string SerialNumberString { get; }
+
+		[NullAllowed, Export ("locationDescription")]
+		string LocationDescription { get; }
+
+		[Export ("hasOpenSession")]
+		bool HasOpenSession { get; }
+
+		[NullAllowed, Export ("UUIDString")]
+		string UuidString { get; }
```
spouliot: same (`String` suffix)

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+
+		[NullAllowed, Export ("persistentIDString")]
+		string PersistentIdString { get; }
```
spouliot: same (`String` suffix)

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+
+		[Export ("ptpObjectHandle")]
+		uint PtpObjectHandle { get; }
+
+		[Export ("addedAfterContentCatalogCompleted")]
+		bool AddedAfterContentCatalogCompleted { [Bind ("wasAddedAfterContentCatalogCompleted")] get; }
+	}
+
+	[BaseType (typeof(ICCameraItem))]
+	interface ICCameraFolder {
```
spouliot: double check if this is user-creatable (for .ctor)
maybe it's `ICCameraItem` that is abstract ?

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+
+		[Export ("contents")]
+		ICCameraItem[] Contents { get; }
+	}
+
+	[BaseType (typeof(ICCameraItem))]
+	interface ICCameraFile {
```
spouliot: same

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+
+		[Export ("requestReadDataFromFile:atOffset:length:readDelegate:didReadDataSelector:contextInfo:")]
+		void RequestReadDataFromFile (ICCameraFile file, long offset, long length, NSObject readDelegate, Selector didReadDataSelector, [NullAllowed] IntPtr contextInfo);
+
+		[Export ("requestSendPTPCommand:outData:sendCommandDelegate:didSendCommandSelector:contextInfo:")]
+		void RequestSendPtpCommand (NSData command, NSData outData, NSObject sendCommandDelegate, Selector didSendCommandSelector, [NullAllowed] IntPtr contextInfo);
+	}
+
+	[BaseType (typeof(NSObject))]
+	interface ICScannerFeature {
```
spouliot: user creatable ?

src/imagecapturecore.cs
```diff
@@ -0,0 +1,900 @@
...
+
+	[BaseType (typeof(ICScannerFeature))]
+	interface ICScannerFeatureTemplate {
+
+		[Export ("targets")]
+		NSMutableArray[] Targets { get; }
+	}
+
+	[BaseType (typeof(NSObject))]
+	interface ICScannerFunctionalUnit {
```
spouliot: user creatable ?

src/imagekit.cs
spouliot: IIRC that was discussed in the original PR

dalexsoto: Yes, here it is

https://github.com/xamarin/xamarin-macios/pull/4404#issuecomment-407481964

> I have undone the XAMCORE_4_0 change, as this API while breaking change could have never been used by anyone before, so the regular policy does not apply to this.

src/imagecapturecore.cs
```diff
@@ -276,7 +276,7 @@ interface ICDevice {
 		void RequestYield ();
 
 		[Export ("requestSendMessage:outData:maxReturnedDataSize:sendMessageDelegate:didSendMessageSelector:contextInfo:")]
-		unsafe void RequestSendMessage (uint messageCode, NSData data, uint maxReturnedDataSize, NSObject sendMessageDelegate, Selector selector, [NullAllowed] IntPtr contextInfo);
+	void RequestSendMessage (uint messageCode, NSData data, uint maxReturnedDataSize, NSObject sendMessageDelegate, Selector selector, [NullAllowed] IntPtr contextInfo);
```
spouliot: minor: update (`unsafe` removal) broke tab alignment

src/imagecapturecore.cs
spouliot: @dalexsoto is that API used in your sample ?

It's nice to hide (or provide better API) than exposing `Selector` and `NSObject` in signatures - but without tests it's also easy too easy to get it wrong.

Since it's not a very common API (or framework) on macOS (only) I think it's fine to file an enhancement on it (would be a nice and not _too_ hard contribution from anyone consuming the API).

dalexsoto: @spouliot Unfortunately no, it does not touch any of those APIs, I have opened this issue https://github.com/xamarin/xamarin-macios/issues/5093 per your request

PR #5080, opened 2018-11-02
Fix min macOS version confusion.

rolfbjarne:
* MIN_OSX_SDK_VERSION: the minimum macOS version we support for running macOS apps.
* MIN_OSX_VERSION_FOR_MAC: the minimum macOS version we support for running XM/XI themselves.

Change our build to make sure the above is respected, and in particular:

* The mac32 and mac64 builds must use MIN_OSX_SDK_VERSION.
* The runtime/ build must use MIN_OSX_SDK_VERSION.
* The tools64 build should use MIN_OSX_VERSION_FOR_MAC.

Also document a bit better the various version variables.


Make.config
```diff
@@ -110,11 +112,12 @@ endif
 XCODE_MAC_SDKROOT=$(XCODE_DEVELOPER_ROOT)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
 XCODE94_MAC_SDKROOT=$(XCODE94_DEVELOPER_ROOT)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
 
-MAC_CC=$(CCACHE)$(XCODE_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -isysroot $(XCODE_MAC_SDKROOT) -mmacosx-version-min=$(MIN_OSX_VERSION_FOR_MAC) -stdlib=libc++
-MAC_CXX=$(CCACHE)$(XCODE_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -isysroot $(XCODE_MAC_SDKROOT) -mmacosx-version-min=$(MIN_OSX_VERSION_FOR_MAC) -stdlib=libc++
+# THe MAC_* variables do not contain the -mmacosx-version-min flag on purpose: each usage must specify it separately.
```
spouliot: minor: `THe` typo

Make.config
```diff
@@ -110,11 +112,12 @@ endif
...
+MAC_CC=$(CCACHE)$(XCODE_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -isysroot $(XCODE_MAC_SDKROOT) -stdlib=libc++
+MAC_CXX=$(CCACHE)$(XCODE_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -isysroot $(XCODE_MAC_SDKROOT) -stdlib=libc++
 
-MAC32_CC=$(CCACHE)$(XCODE94_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -isysroot $(XCODE94_MAC_SDKROOT) -mmacosx-version-min=$(MIN_OSX_VERSION_FOR_MAC) -stdlib=libc++
-MAC32_CXX=$(CCACHE)$(XCODE94_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -isysroot $(XCODE94_MAC_SDKROOT) -mmacosx-version-min=$(MIN_OSX_VERSION_FOR_MAC) -stdlib=libc++
+MAC32_CC=$(CCACHE)$(XCODE94_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -isysroot $(XCODE94_MAC_SDKROOT) -stdlib=libc++
+MAC32_CXX=$(CCACHE)$(XCODE94_DEVELOPER_ROOT)/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -isysroot $(XCODE94_MAC_SDKROOT) -stdlib=libc++
```
spouliot: > each usage must specify it separately.

That's hard to ensure it will be (and remain) correct over time.

Maybe we should add a test that read the value from the binary we ship and compare it with what we expect ?

rolfbjarne: Good idea, filed it as https://github.com/xamarin/xamarin-macios/issues/5082.

Make.config
rolfbjarne: Fixed

rolfbjarne:
I forgot that Jenkins hasn't built this yet 😒 I'll keep an eye out and see if it breaks something.

PR #5106, opened 2018-11-08
[jenkins] Run the 'Test docs' step even if the 'Run tests' step failed.

rolfbjarne:
This is done by preventing the 'Run tests' failure from throwing exceptions,
and instead manually set failure modes.


rolfbjarne:
I'm skipping the build on public Jenkins because this has already been tested on private Jenkins: the results shown above unfortunately didn't test this scenario (because the 'Run tests' step succeeded [1]), but another build was done where I forced the 'Run tests' step to fail, and that build shows that the 'Test docs' code executed as expected: https://jenkins.internalx.com/blue/organizations/jenkins/macios/detail/jenkins-test-docs-even-if-tests-failed/2/pipeline/170.

[1] It's funny how the tests succeed when you want them to fail...

PR #5114, opened 2018-11-08
[xtro] Report incorrect 'ArgumentSemantic' enum value usage

### LINKED ISSUE #4018: [xtro] Report incorrect 'ArgumentSemantic' enum value usage
Xtro should report incorrect 'ArgumentSemantic' enum value usage.

E.g: ClassKit in Xcode 9.4 beta 2 had a bunch of:

``` Diff
-@property (nullable, nonatomic) __kindof CLSActivityItem *primaryActivityItem;
+@property (nullable, nonatomic, strong) __kindof CLSActivityItem *primaryActivityItem;
```

It'd be very helpful if Xtro reported those.

VincentDondain:
- Fixes #4018: [xtro] Report incorrect 'ArgumentSemantic' enum value usage
  (https://github.com/xamarin/xamarin-macios/issues/4018).

### DIFF
--- a/tests/xtro-sharpie/SelectorCheck.cs
+++ b/tests/xtro-sharpie/SelectorCheck.cs
@@ -64,12 +64,11 @@ namespace Extrospection {
 				return;
 
 			var nativeArgumentSemantic = decl.Attributes.ToArgumentSemantic ();
-
 			var nativeMethodDefinition = decl.QualifiedName;
 
 			bool found = qualified_properties.TryGetValue (nativeMethodDefinition, out var managedArgumentSemantic);
 			if (found && managedArgumentSemantic != nativeArgumentSemantic)
-				Log.On (framework).Add ($"!incorrect-argument-semantic! Native '{nativeMethodDefinition}' has ({nativeArgumentSemantic.ToUsableString ().ToLowerInvariant ()}) instead of 'ArgumentSemantic.{managedArgumentSemantic.ToUsableString ()}'");
+				Log.On (framework).Add ($"!incorrect-argument-semantic! Native '{nativeMethodDefinition}' is declared as ({nativeArgumentSemantic.ToUsableString ().ToLowerInvariant ()}) but mapped to 'ArgumentSemantic.{managedArgumentSemantic.ToUsableString ()}'");
 		}
 
 		public override void VisitObjCMethodDecl (ObjCMethodDecl decl, VisitKind visitKind)
--- a/tests/xtro-sharpie/SelectorCheck.cs
+++ b/tests/xtro-sharpie/SelectorCheck.cs
@@ -11,7 +11,6 @@ using System.Collections.Generic;
 using Mono.Cecil;
 
 using Clang.Ast;
-using System.Linq;
 
 namespace Extrospection {
 
@@ -36,7 +35,7 @@ namespace Extrospection {
 				case "ExportAttribute":
 					var methodDefinition = method.GetName ();
 					if (!string.IsNullOrEmpty (methodDefinition)) {
-						var argumentSemantic = Helpers.ArgumentSemantic.None;
+						var argumentSemantic = Helpers.ArgumentSemantic.Assign; // Default
 						if (ca.ConstructorArguments.Count > 1) {
 							argumentSemantic = (Helpers.ArgumentSemantic)ca.ConstructorArguments [1].Value;
 							qualified_properties.Add (methodDefinition, argumentSemantic);
@@ -66,15 +65,11 @@ namespace Extrospection {
 
 			var nativeArgumentSemantic = decl.Attributes.ToArgumentSemantic ();
 
-			// Ignore Copy, UnsafeUnretained and Weak for now
-			if (nativeArgumentSemantic == Helpers.ArgumentSemantic.Copy || nativeArgumentSemantic == Helpers.ArgumentSemantic.UnsafeUnretained || nativeArgumentSemantic == Helpers.ArgumentSemantic.Weak)
-				return;
-
 			var nativeMethodDefinition = decl.QualifiedName;
 
 			bool found = qualified_properties.TryGetValue (nativeMethodDefinition, out var managedArgumentSemantic);
 			if (found && managedArgumentSemantic != nativeArgumentSemantic)
-				Log.On (framework).Add ($"!incorrect-argument-semantic! '{nativeMethodDefinition}' (native) has ({nativeArgumentSemantic.ToUsableString ().ToLower ()}) instead of 'ArgumentSemantic.{managedArgumentSemantic.ToUsableString ()}'");
+				Log.On (framework).Add ($"!incorrect-argument-semantic! Native '{nativeMethodDefinition}' has ({nativeArgumentSemantic.ToUsableString ().ToLowerInvariant ()}) instead of 'ArgumentSemantic.{managedArgumentSemantic.ToUsableString ()}'");
 		}
 
 		public override void VisitObjCMethodDecl (ObjCMethodDecl decl, VisitKind visitKind)
@@ -105,7 +100,6 @@ namespace Extrospection {
 					return;
 				name = "+" + name;
 			}
-
 			bool found = qualified_selectors.Contains (name);
 			if (!found) {
 				// a category could be inlined into the type it extend
--- a/tests/xtro-sharpie/SelectorCheck.cs
+++ b/tests/xtro-sharpie/SelectorCheck.cs
@@ -17,7 +17,8 @@ namespace Extrospection {
 
 	public class SelectorCheck : BaseVisitor {
 
-		HashSet<(string MethodDefinition, Helpers.ArgumentSemantic ArgumentSemantic)> qualified_export_arguments = new HashSet<(string, Helpers.ArgumentSemantic)> ();
+		HashSet<string> qualified_selectors = new HashSet<string> ();
+		Dictionary<string, Helpers.ArgumentSemantic> qualified_properties = new Dictionary<string, Helpers.ArgumentSemantic> ();
 
 		// most selectors will be found in [Export] attribtues
 		public override void VisitManagedMethod (MethodDefinition method)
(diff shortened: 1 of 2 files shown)


VincentDondain:
Time of `SelectorCheck` with new test:

```
Extrospection.SelectorCheck: Elapsed=00:00:02.0330457
Extrospection.SelectorCheck: Elapsed=00:00:05.7196766
Extrospection.SelectorCheck: Elapsed=00:00:10.8809971
Extrospection.SelectorCheck: Elapsed=00:00:14.4416889
```

(x4 because ran on 4 platforms)

Before:
```
Extrospection.SelectorCheck: Elapsed=00:00:00.5158108
Extrospection.SelectorCheck: Elapsed=00:00:00.8786211
Extrospection.SelectorCheck: Elapsed=00:00:01.0491429
Extrospection.SelectorCheck: Elapsed=00:00:01.2374137
```

Now I realize moving to `LINQ` might not have been such a good idea (will investigate) (:

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -4,38 +4,20 @@
...
-
-		// duplicate
-		//	-> the selector is defined more than once for the same type
 
-		HashSet<string> known_selectors = new HashSet<string> ();
-
-		HashSet<string> qualified_selectors = new HashSet<string> ();
-
-		//Dictionary<TypeDefinition,HashSet<string>> type_exports = new Dictionary<TypeDefinition,HashSet<string>> ();
+		HashSet<(string MethodDefinition, Helpers.ArgumentSemantic ArgumentSemantic)> qualified_export_arguments = new HashSet<(string, Helpers.ArgumentSemantic)> ();
```
spouliot: you're now allocating a tuple for each entry - even if only properties actually needs the extra argument semantics

spouliot: depending on the numbers it might be cheaper to have a dictionary where you store argument semantics (and only when available) along with the `MethodDefinition`

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -51,24 +33,53 @@ public override void VisitManagedMethod (MethodDefinition method)
...
+			// Ignore Copy, UnsafeUnretained and Weak for now
+			if (nativeArgumentSemantic == Helpers.ArgumentSemantic.Copy || nativeArgumentSemantic == Helpers.ArgumentSemantic.UnsafeUnretained || nativeArgumentSemantic == Helpers.ArgumentSemantic.Weak)
+				return;
+
+			var nativeMethodDefinition = decl.QualifiedName;
+
+			var exportArgs = qualified_export_arguments.FirstOrDefault (sel => sel.MethodDefinition.Contains (nativeMethodDefinition));
+
+			if (!string.IsNullOrEmpty (exportArgs.MethodDefinition) && exportArgs.ArgumentSemantic != nativeArgumentSemantic)
+				Log.On (framework).Add ($"!incorrect-argument-semantic! {nativeMethodDefinition} has ArgumentSemantic.{nativeArgumentSemantic.ToUsableString ()} instead of ArgumentSemantic.{exportArgs.ArgumentSemantic.ToUsableString ()}");
```
spouliot:
```
!incorrect-argument-semantic! SKUniform::textureValue has ArgumentSemantic.Strong|Retain instead of ArgumentSemantic.None
```
it's a bit confusing when reading (without our bindings sources or xtro sources). 

Either (best) you use the `ArgumentSemantic` for the managed version and the ObjC syntax (for clang/headers); or you mention which one is managed and native.

tests/xtro-sharpie/SelectorCheck.cs
rolfbjarne: Can't this be `Dictionary<string, ArgumentSemantic>`? Or can you have multiple methods with the same name, but different argument semantics?

That would mean any lookup would be a hashed lookup, and not a sequential lookup.

tests/xtro-sharpie/SelectorCheck.cs
VincentDondain: @spouliot `!incorrect-argument-semantic! 'SKUniform::textureValue' (native) has (strong|retain) instead of 'ArgumentSemantic.None'`?

spouliot: @VincentDondain yes, that would remove the ambiguity :)

VincentDondain:
**UPDATED**

Time of SelectorCheck with **new** test:

```
Extrospection.SelectorCheck: Elapsed=00:00:00.7263742
Extrospection.SelectorCheck: Elapsed=00:00:01.2136911
Extrospection.SelectorCheck: Elapsed=00:00:01.2747602
Extrospection.SelectorCheck: Elapsed=00:00:01.7494063
```
(x4 because ran on 4 platforms)

Before:

```
Extrospection.SelectorCheck: Elapsed=00:00:00.5158108
Extrospection.SelectorCheck: Elapsed=00:00:00.8786211
Extrospection.SelectorCheck: Elapsed=00:00:01.0491429
Extrospection.SelectorCheck: Elapsed=00:00:01.2374137
```

It is now only _slightly slower_ but I think that's acceptable (:

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -4,38 +4,21 @@
...
-//             if we have a selector that is not part of the header files
-//
 
 using System;
 using System.Collections.Generic;
 
 using Mono.Cecil;
 
 using Clang.Ast;
+using System.Linq;
```
spouliot: are you still using any linq ?

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -51,24 +34,49 @@ public override void VisitManagedMethod (MethodDefinition method)
...
+
+			// Ignore Copy, UnsafeUnretained and Weak for now
+			if (nativeArgumentSemantic == Helpers.ArgumentSemantic.Copy || nativeArgumentSemantic == Helpers.ArgumentSemantic.UnsafeUnretained || nativeArgumentSemantic == Helpers.ArgumentSemantic.Weak)
+				return;
+
+			var nativeMethodDefinition = decl.QualifiedName;
+
+			bool found = qualified_properties.TryGetValue (nativeMethodDefinition, out var managedArgumentSemantic);
+			if (found && managedArgumentSemantic != nativeArgumentSemantic)
+				Log.On (framework).Add ($"!incorrect-argument-semantic! '{nativeMethodDefinition}' (native) has ({nativeArgumentSemantic.ToUsableString ().ToLower ()}) instead of 'ArgumentSemantic.{managedArgumentSemantic.ToUsableString ()}'");
```
spouliot: use `ToLowerInvariant`

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -97,6 +105,7 @@ public override void VisitObjCMethodDecl (ObjCMethodDecl decl, VisitKind visitKi
 					return;
 				name = "+" + name;
 			}
+
```
spouliot: minor: noise

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -51,24 +34,49 @@ public override void VisitManagedMethod (MethodDefinition method)
...
+			if (!decl.IsAvailable ())
+				return;
+
+			var framework = Helpers.GetFramework (decl);
+			if (framework == null)
+				return;
+
+			var nativeArgumentSemantic = decl.Attributes.ToArgumentSemantic ();
+
+			// Ignore Copy, UnsafeUnretained and Weak for now
```
spouliot: why ?

tests/xtro-sharpie/SelectorCheck.cs
```diff
@@ -51,24 +34,49 @@ public override void VisitManagedMethod (MethodDefinition method)
 			foreach (var ca in method.CustomAttributes) {
 				switch (ca.Constructor.DeclaringType.Name) {
 				case "ExportAttribute":
-					string selector = ca.ConstructorArguments [0].Value as string;
-					if (!known_selectors.Contains (selector))
-						known_selectors.Add (selector);
+					var methodDefinition = method.GetName ();
+					if (!string.IsNullOrEmpty (methodDefinition)) {
+						var argumentSemantic = Helpers.ArgumentSemantic.None;
```
spouliot: `None` is not a value - it means the default (pre-ARC it's `Assign`)
The comparison you make later might not be correct.

spouliot: Here's an example
```
!incorrect-argument-semantic! 'CNContainer::type' (native) has (none) instead of 'ArgumentSemantic.UnsafeUnretained|Assign'
```
and the default (when nothing is specified) is `assign` - that's a false positive.

spouliot:
```
!incorrect-argument-semantic! 'AVRoutePickerView::activeTintColor' (native) has (strong|retain) instead of 'ArgumentSemantic.UnsafeUnretained|Assign'
```

Might just be me but the `instead of` is confusing, since it sounds like the managed version is right - while it's the native one that we can't change. What about:

```
'X' (native) is declared as (Y) but mapped to 'Z' (managed).
```

spouliot:
Not sure if it's worth cleaning up `weak` vs `ArgumentSemantic.UnsafeUnretained|Assign` since it's close and IIRC we don't care. It's nicer (and less confusing) to have matching declarations, but it's noisy to update (for future merges, e.g. xcode10.2 on 15.9).

@rolfbjarne can you look at the xtro results ? thanks!

rolfbjarne:
The results look correct to me from what I can see.

What we really care about is if native code uses 'copy', and our binding don't match (we can end up with invalid pointers). The reverse has a performance hit (if we declare something as 'ArgumentSemantic.Copy', while native doesn't), because we'd copy values when not needed, but _in theory_ it shouldn't cause trouble.

I think we should fix any mismatched 'copy' semantics, and the rest can ignore and/or delay until sometime later to fix.

rolfbjarne:
@VincentDondain correct.

spouliot:
@VincentDondain add a comment `FIXME: only Copy mistakes are reported now` and the related logic. That will tell us how many fixes are needed. You can run it locally (and report the numbers here), no need to push that yet.

Next disable the rule (so it does not report anything) and we can merge it into `master` (along with above changes).

Finally we'll back port the commit to `xcode10.2` (once we create it - and we can do it earlier than Apple) and have it fixed there to avoid merge conflicts (since doing large binding changes in `master`, while `xcode10.2` will be based on `d15-9` will likely be painful). It also means the fixes will become available sooner (than `d16-0`).

VincentDondain:
The result is 142 `!incorrect-argument-semantic!` found.

[report.zip](https://github.com/xamarin/xamarin-macios/files/2625650/report.zip)

rolfbjarne:
build

PR #5128, opened 2018-11-14
[DO_NOT_MERGE] Fix tests to run with msbuild

radical:
//commit messages: TODO


radical:
Testing this.

tools/xibuild/Options.cs
```diff
@@ -0,0 +1,2000 @@
+//
```
mandel-macaque: We no not need this file, do we?

radical:
Added more fixes and fixed other ways that the tests might be run.

rolfbjarne:
@radical thanks for this! I'll have a look at this next week and fix whatever needs fixing.

marek-safar:
@rolfbjarne @spouliot could we fast-track this and review/merge ASAP ?

rolfbjarne:
build

tests/xharness/Jenkins.cs
```diff
@@ -2872,8 +2874,9 @@ protected override async Task RunTestAsync ()
 				using (var proc = new Process ()) {
 
 					proc.StartInfo.WorkingDirectory = WorkingDirectory;
-					proc.StartInfo.FileName = "/Library/Frameworks/Mono.framework/Commands/mono";
+					proc.StartInfo.FileName = "xibuild";
 					var args = new StringBuilder ();
+					args.Append ("-t -- ");
```
rolfbjarne: Same: won't this try to write to `/Library/Frameworks/Mono.framework/Commands/mono.config`?

radical: This is running `TestExecutable` with `xibuild`, which should be nunit from the packages directory.

marek-safar:
@rolfbjarne @spouliot please have a second look

spouliot:
@radical did you force push to your branch ? because I don't see a newer commit. Doing that makes it a lot harder to review changes and ensure feedback was applied.

radical:
I did. I wanted to have clean micro commits. I can try to do separate commits for changes in future PRs.

spouliot:
@radical please do, I don't mind merge commit or squash (depending if history is important) but reviewing, again, changes inside 20 files makes the process heavier and longer than it needs to be :)

tools/xibuild/Main.cs
```diff
@@ -0,0 +1,220 @@
...
+				Console.WriteLine ("\t-t: Path to the managed tool to run. If this and `-c` are not used, then this runs msbuild");
+				Console.WriteLine ("\t-m <base config file>: Config file to merge with the one from msbuild.dll.config");
+				Console.WriteLine ("\t-h: help");
+				Console.WriteLine ("\nNote: Adds the path from the environment variable MSBuildExtensionsPathFallbackPathsOverride to the list of fallback paths in the generated app.config");
+				Console.WriteLine ("\nExamples:");
+				Console.WriteLine ("xibuild -c /path/to/foo.exe : Generate /path/to/foo.exe.config");
+				Console.WriteLine ("xibuild -- /v:diag /path/to/project.csproj : Run msbuild with a custom app.config and the arguments passed to msbuild.");
+				Console.WriteLine ("xibuild -t -- /path/to/managed_tool.exe args : Run managed_tool.exe with the arguments, and a custom app.config");
+				Console.WriteLine ("");
+				Console.WriteLine ("Add `-m /path/to/base.exe.config` to merge the generated app.config with base.exe.config .");
```
mandel-macaque: Minor comment, but I think you could have used the options print method, so that the help and the options are in sync. Something like

```csharp
options.WriteOptionDescriptions (Console.Out);
```

tools/xibuild/Main.cs
```diff
@@ -0,0 +1,220 @@
...
+			dstXml.AppendChild (dstConfigNode);
+
+			if (!String.IsNullOrEmpty (baseConfigFile) && File.Exists (baseConfigFile)) {
+				var baseXml = new XmlDocument ();
+				baseXml.Load (baseConfigFile);
+				CopyConfigNode (baseXml, dstConfigNode);
+			}
+
+			// Copy over msbuild.dll.config
+			{
```
mandel-macaque: Why the {} ?

tools/xibuild/Main.cs
```diff
@@ -0,0 +1,220 @@
...
+				}
+			}
+
+			/// <summary>
+			/// If the value exists, sets value attribute, else creates the element
+			/// </summary>
+			void SetToolsetProperty (string name, string value)
+			{
+				if (string.IsNullOrEmpty (value))
+					return;
```
mandel-macaque: WE check for the value, but what about if the name is null or empty?

tests/common/mac/ProjectTestHelpers.cs
```diff
@@ -573,7 +574,14 @@ static string GenerateOutputCommand (string tmpDir, Guid guid)
 
 		public static void NugetRestore (string project)
 		{
-			var rv = ExecutionHelper.Execute ("nuget", $"restore {StringUtils.Quote (project)}", out var output);
+			string rootDirectory = FindRootDirectory ();
+
+			Environment.SetEnvironmentVariable ("TargetFrameworkFallbackSearchPaths", rootDirectory + "/Library/Frameworks/Mono.framework/External/xbuild-frameworks");
+			Environment.SetEnvironmentVariable ("MSBuildExtensionsPathFallbackPathsOverride", rootDirectory + "/Library/Frameworks/Mono.framework/External/xbuild");
+			Environment.SetEnvironmentVariable ("XAMMAC_FRAMEWORK_PATH", rootDirectory + "/Library/Frameworks/Xamarin.Mac.framework/Versions/Current");
+			Environment.SetEnvironmentVariable ("XamarinMacFrameworkRoot", rootDirectory + "/Library/Frameworks/Xamarin.Mac.framework/Versions/Current");
```
mandel-macaque: Path.Combine over the concatenation?

tools/xibuild/Main.cs
radical: There are multiple *Xml variables, and this one is required only for this specific part, so scoped it to prevent accidental use.

tools/xibuild/Main.cs
radical: That would be a bug and IMHO letting the exception be thrown would be good. We could check for null/empty name and explicitly show an error+exit, but not sure if that is worth doing here.

tests/common/mac/ProjectTestHelpers.cs
radical: This is very specifically for macOS, so I don't see a value in using `Path.Combine` here, unless I break the fixed path into it's parts. Also, this is copy pasted from existing code (not necessarily a good reason for sticking to this though). If more people feel this is required, then I can make the change.

radical:
Updated to address the review feedback, with a separate commit.

tests/common/mac/ProjectTestHelpers.cs
mandel-macaque: Nor a big deal. Just doing a review from a general point of view :)

Issue #5129, opened 2018-11-14
Build error for library project with SceneKit asset after updating to Xamarin.iOS 12.2.1.10


mikebluestein:
### Steps to Reproduce

1. Create an iOS library project
2. Add a file with SceneKitAsset as the build action
3. Build project

### Expected Behavior
Project builds

### Actual Behavior
Build failure: 

/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/iOS/Xamarin.iOS.Common.targets(1281,3): error MSB4044: The "CompileSceneKitAssets" task was not given a value for the required parameter "AppBundleDir".

### Environment

```
=== Visual Studio Enterprise 2017 for Mac ===

Version 7.6.11 (build 9)
Installation UUID: 0cc4e5d4-daf4-4e4e-bf89-0c02c659a044
Runtime:
	Mono 5.12.0.309 (2018-02/39d89a335c8) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.1.178 (master / eeaeb7e6)

	Package version: 512000309

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.1.2
	2.1.1
	2.0.5
SDK: /usr/local/share/dotnet/sdk/2.1.302/Sdks
SDK Versions:
	2.1.302
	2.1.301
	2.1.4
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.12.0/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.3
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Xamarin.Android ===

Version: 9.0.0.20 (Visual Studio Enterprise)
Android SDK: /Users/mikeb/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		7.1 (API level 25)
		8.0 (API level 26)
		8.1 (API level 27)

SDK Tools Version: 26.1.1
SDK Platform Tools Version: 27.0.1
SDK Build Tools Version: 26.0.3

Java SDK: /usr
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

Android Designer EPL code available here:
[...]

PR #4851, opened 2018-09-20
[msbuild] Fix SceneKit asset compilation. Fixes #3766.

rolfbjarne:
It seems SceneKit assets must be compiled into a directory whose parent
directory is named like the app.

In addition the `--resource-folder-path` argument is required.

Fixes https://github.com/xamarin/xamarin-macios/issues/3766.

PR #5130, opened 2018-11-15
[msbuild] Fixed compilation of SceneKit assets in library projects

jstedfast:
The problem is that AppBundleDir never gets defined for library projects.

Luckily, _AppBundleName and AppBundleExtension are always defined and
all we need is the directory name (i.e. MyApp.app), so this will work
just fine.

Fixes issue #5129


jstedfast:
Turns out I gave bad advice to @rolfbjarne in https://github.com/xamarin/xamarin-macios/pull/4851 - I advised him to use `$(AppBundleDir)` instead of `$(_AppBundleName).app`, but the problem is that `$(AppBundleDir)` does not get defined for library projects.

Luckily we have `$(AppBundleExtension)` that we can use instead of blindly assuming ".app", so we can use: `$(_AppBundleName)$(AppBundleExtension)`.

Both of these variables are ALWAYS defined regardless of whether the project is a library project, app extension, or an actual app project.

jstedfast:
Forgot to update the Mac targets :(

spouliot:
@monojenkins backport to d16-0-p2

PR #5139, opened 2018-11-19
[xharness] Only build the BCL tests when we need them.

rolfbjarne:
Only build the BCL tests when we need them, because we can't build them when
we're running the device tests (because we don't have a built source tree in
that case).

Since the test assemblies are required when generating the BCL projects, this
means we also delay creating the BCL projects until we need them.


rolfbjarne:
Test failure is unrelated:

* monotouch-test/watchOS: https://github.com/xamarin/maccore/issues/964

PR #5154, opened 2018-11-20
[Harness] Reduce the noise in the make files.

mandel-macaque:
The Makefile.cs code generates a lot of noise becuase it tries to build
a make rule for the new BCL tests, these rules do not work and can be
ignored.

Fixes issue: https://github.com/xamarin/maccore/issues/1156

### DIFF
--- a/tests/xharness/MakefileGenerator.cs
+++ b/tests/xharness/MakefileGenerator.cs
@@ -297,7 +297,7 @@ namespace xharness
 
 				// build/[install/]run targets for specific test projects.
 				foreach (var target in allTargets) {
-					if (!target.IsExe)
+					if (!target.IsExe || target.Name.Contains ("BCLTests"))
 						continue;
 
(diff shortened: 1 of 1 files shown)


tests/xharness/MakefileGenerator.cs
```diff
@@ -297,7 +297,7 @@ public static void CreateMakefile (Harness harness, IEnumerable<UnifiedTarget> u
 
 				// build/[install/]run targets for specific test projects.
 				foreach (var target in allTargets) {
-					if (!target.IsExe)
+					if (!target.IsExe || target.Name.Contains ("BCLTests"))
```
spouliot: add `,StringComparison.Ordinal`

less an issue for testing code (than the SDK) but it's faster and requires less memory - which is always nice inside a loop

mandel-macaque:
Ignore the failure from the bot, that was due to an abort after a push.

PR #5175, opened 2018-11-23
Revert "Mono 5.16.0.87+ breaks us, so make sure we don't use it for now. (#5088)"

spouliot:
This reverts commit 1ca7da537da1e81d56bcdeb215c63e16d2a08bdc.

This is fixed so the restriction is not needed anymore.


spouliot:
known random issues
* apitest/Mac -> https://github.com/xamarin/maccore/issues/643
* monotouch-test/watchOS -> https://github.com/xamarin/maccore/issues/581

PR #5181, opened 2018-11-26
[macos] Fix Constants.mac.cs.in for backward compatibility

spouliot:
On macOS the path changes are forward but not backward compatible.

This means we need to keep older values, as they will work on newer OS,
while the new path won't work (on older OS).

The test are updated to add exceptions for such cases - in order to
give us a better chance to spot typos (since they were some).

This should fix
https://github.com/xamarin/maccore/issues/1174
https://github.com/xamarin/maccore/issues/1175
https://github.com/xamarin/maccore/issues/1176
https://github.com/xamarin/maccore/issues/1177


rolfbjarne:
Breaking changes are expected.

The test failures on internal jenkins are unrelated:

* XM tests on '10.10': https://github.com/xamarin/maccore/issues/1098

Issue #4107, opened 2018-05-21
Detect invalid delegate names in AdoptsAttribute and show a better error.


rolfbjarne:
See #4094 for how a customer ran into this.

Test case:

```csharp
[Adopts ("Foo.Bar")]
class Foo : NSObject
{
}
```

results in:

```
/.../registrar.m(35722,30): error G65037BDE: expected '>'
/.../registrar.m(35722,18): error GC919C233: type arguments cannot be applied to non-parameterized class 'NSObject'
/.../registrar.m(35722,30): error G9772CD21: cannot use dot operator on a type
MTOUCH : error MT4109: Failed to compile the generated registrar code. Please file a bug report at http://bugzilla.xamarin.com
```

rolfbjarne:
A potential fix would probably be to implement something similar to how we verify `[Register]` attributes: https://github.com/xamarin/xamarin-macios/blob/07356b8391eecf1dabe11ba3e662f9ca8684a4cf/src/ObjCRuntime/Registrar.cs#L162

PR #5191, opened 2018-11-28
[registrar] Validate input to 'Adopts' attribute

dalexsoto:
Fixes xamarin/xamarin-macios#4107

We currently generate invalid registrar code when we get invalid input
in 'Adopts' attribute, this can happen because 'Adopts' can take a
plain string. We now generate a better error MT4177 instead of
generating invalid registrar.m/h code.

This might seem like a "breaking change" because it "errors out" a build
but since we generated bad `registrar.m/h` there is no way this is shipping
out there.


dalexsoto:
Test Failure unrelated:

* [apitest] Assertion failure in void assertRunningOnAppKitThread(void) -> https://github.com/xamarin/maccore/issues/643

PR #5193, opened 2018-11-28
[xharness] Write intermediate html reports to disk when running in server mode too.

rolfbjarne:
That way the report is still available even if xharness dies.

Also improve report generation to write directly to a temporary file on disk,
instead of writing to memory and then writing to disk (improves memory usage).


rolfbjarne:
Test failure is unrelated:

PR #5194, opened 2018-11-28
[monotouch-test] Require iOS 9+ for the ActionTest.TimingFunction_* tests.

rolfbjarne:
Require iOS 9+ for the ActionTest.TimingFunction_* tests because Apple's
implementation was broken in iOS 8 and causes us to crash.


rolfbjarne:
API comparison failure is unrelated: https://github.com/xamarin/maccore/issues/1173

PR #5198, opened 2018-11-29
[xharness] Don't recursively print diagnostics for processes when printing diagnostics times out in the first place. Fixes maccore#1163.

rolfbjarne:
Recursively printing diagnostics can leave us with _many_ lldb processes, each
pretty much hanging, and the system overloaded, which, in the worst case
scenario, would require a hard reboot.

Most commonly the "only" thing that happens is an OutOfMemoryException when
the OS eventually tells xharness NO to launching new processes.

Fixes https://github.com/xamarin/maccore/issues/1163.

### DIFF
--- a/tests/xharness/Process_Extensions.cs
+++ b/tests/xharness/Process_Extensions.cs
@@ -56,17 +56,17 @@ namespace xharness
 
 	public static class Process_Extensions
 	{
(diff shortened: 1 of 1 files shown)


rolfbjarne:
Test failures are unrelated, they're all caused by https://github.com/xamarin/maccore/issues/1080 (failure to edit TCC.db which causes permission dialogs, making tests time out).

PR #5199, opened 2018-11-29
[xharness] No need to spawn to kill, when we can just P/Invoke the corresponding function.

rolfbjarne:


rolfbjarne:
xharness OOM'ed: https://github.com/xamarin/maccore/issues/1163

rolfbjarne:
build

PR #5209, opened 2018-11-30
[xibuild] Don't redirect standard input.

rolfbjarne:
It's not needed, and it prevents subprocesses from detecting a terminal (and
thus no colors in the output).


rolfbjarne:
build

PR #5211, opened 2018-11-30
[xibuild] Auto-make if not already made.

rolfbjarne:
Automatically make xibuild if it doesn't already exist.

Also fix the script:

* Use bash instead of sh, since it references BASH_SOURCE.
* Don't hide errors (-e).
* Fix shellcheck warnings
	* Use $(..) instead of backticks, because backticks are strongly discouraged [1].
	* Sprinkle quotes.

[1] For the curious: https://stackoverflow.com/a/4708569/183422


rolfbjarne:
Test failure is unrelated:

* apitest/Mac Unified 32-bit: https://github.com/xamarin/maccore/issues/643

PR #5223, opened 2018-12-04
[xibuild] Don't redirect standard input.

rolfbjarne:
It's not needed, and it prevents subprocesses from detecting a terminal (and
thus no colors in the output).


rolfbjarne:
> looks identical to #5209 ?!?

Oops... I keep forgetting what I've done and what I haven't :smile:

Strange that github would show a diff though, you'd think it would turn out empty.

Issue #5233, opened 2018-12-05
NSFont FromFontName should not check for UI thread


praeclarum:
NSFont FromFontName/fontWithName:size: should not check for the UI thread since NSFonts are used throughout the APIs for non-UI tasks. Such things include text rendering with NSGraphics, SceneKit, and just general font playing around.

### Steps to Reproduce

1. Create a background thread
2. Call NSFont.FromFontName ("Helvetica", 10);

### Expected Behavior

Everything is OK, fonts are allowed on background threads.

### Actual Behavior

EnsureUIThread throws an exception

### Environment

```
=== Visual Studio Enterprise 2017 for Mac ===

Version 7.7 (build 1868)
Installation UUID: 4ba352e1-73e8-4dab-a9db-28012ca05af8
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.1.178 (master / eeaeb7e6)

	Package version: 516000221

=== Mono Framework MDK ===

Runtime:
	Mono 5.16.0.221 (2018-06/b63e5378e38) (64-bit)
	Package version: 516000221

=== NuGet ===

Version: 4.7.0.5148

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.1.2
	2.1.0-preview2-26406-04
	2.0.3
SDK: /usr/local/share/dotnet/sdk/2.1.302/Sdks
SDK Versions:
	2.1.302
	2.1.300-preview2-008530
	2.0.3
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.4
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Apple Developer Tools ===

Xcode 10.1 (14460.46)
Build 10B61

=== Xamarin.Mac ===

Version: 5.2.1.11 (Visual Studio Enterprise)
Hash: 5ef73d34
Branch: 
Build date: 2018-11-21 00:03:25-0500

=== Xamarin.iOS ===

Version: 12.2.1.11 (Visual Studio Enterprise)
Hash: 5ef73d34
Branch: d15-9
Build date: 2018-11-21 00:03:24-0500

=== Xamarin.Android ===

Version: 9.1.0.38 (Visual Studio Enterprise)
[...]

Issue #5237, opened 2018-12-06
[generator|tests] Leverage Apple's Main Thread Checker to get correct [ThreadSafe] checks 


dalexsoto:
### Description

We currently try to guess what is thread safe based on documentation, experience and reports like [5235](https://github.com/xamarin/xamarin-macios/issues/5235), [5234](https://github.com/xamarin/xamarin-macios/issues/5234), [5233](https://github.com/xamarin/xamarin-macios/issues/5233), etc. In Xcode 9, Apple [introduced Main Thread Checker](https://developer.apple.com/documentation/code_diagnostics/main_thread_checker?language=objc) we should find a way to leverage it in our advantage maybe as part of one of our test suites (intro) or as a brand new test.

We might need to adjust the generator once we find a way to do the above.

rolfbjarne:
See also: https://trello.com/c/n8cY8Tbs

spouliot:
some (old) documents about thread safety

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaDrawingGuide/Introduction/Introduction.html#//apple_ref/doc/uid/TP40003290
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaDrawingGuide/GraphicsContexts/GraphicsContexts.html#//apple_ref/doc/uid/TP40003290-CH203-BCIJFBJJ

jurganson:
I know this thread issue is about 5 years old, but here goes nothing! Since we need XCode for working with Xamarin on iOS devices and simulators, is there any way to enable the Main Thread Checker through either through some arguments or on the paired mac whenever I debug my Xamarin.Forms application on iOS?

rolfbjarne:
Here's Apple's current documentation:

https://developer.apple.com/documentation/xcode/diagnosing-memory-thread-and-crash-issues-early#Detect-improper-UI-updates-on-background-threads

and they say you can enable the Main Thread Checker by injecting a dylib into your executable on macOS.

That library doesn't work on iOS, but it seems there's a dylib already on device, ready to be used. I looked a little bit at how Xcode loads the library for iOS, but it wasn't obvious how they do it, so we might have to debug Xcode to figure it out, and that's a much bigger task.

[...]

PR #5238, opened 2018-12-06
[AppKit] Adds some [ThreadSafe] to reported cases

dalexsoto:
Fixes xamarin/xamarin-macios#5233
Fixes xamarin/xamarin-macios#5234
Fixes xamarin/xamarin-macios#5235

The following selectors are thread safe:

* NSFont fontWithName:size:
* NSBezierPath bezierPathWithOvalInRect:
* NSString drawAtPoint:withAttributes:

Fixed reported cases and created an enhancement request for us https://github.com/xamarin/xamarin-macios/issues/5237

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -1159,6 +1159,7 @@ namespace AppKit {
 		[Export ("bezierPathWithRect:")]
 		NSBezierPath FromRect (CGRect rect);
 
+		[ThreadSafe]
 		[Static]
 		[Export ("bezierPathWithOvalInRect:")]
 		NSBezierPath FromOvalInRect (CGRect rect);
@@ -6187,6 +6188,7 @@ namespace AppKit {
 	[BaseType (typeof (NSObject))]
 	[DisableDefaultCtor] // crash at runtime (e.g. description). Documentation state: "You don’t create NSFont objects using the alloc and init methods."
 	partial interface NSFont : NSSecureCoding, NSCopying {
+		[ThreadSafe]
 		[Static]
 		[Export ("fontWithName:size:")]
 		NSFont FromFontName (string fontName, nfloat fontSize);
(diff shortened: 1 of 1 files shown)


src/appkit.cs
```diff
@@ -1159,6 +1159,7 @@ interface NSBezierPath : NSSecureCoding, NSCopying {
 		[Export ("bezierPathWithRect:")]
 		NSBezierPath FromRect (CGRect rect);
 
+		[ThreadSafe]
```
spouliot: I don't want to over generalize but I think it's safe to assume other similar methods will be thread safe on move the attribute on the class level.

Even more for this case and the `NString` extension since I could find:

> Drawing Restrictions
> The Application Kit is generally thread-safe when drawing with its graphics functions and classes, including the NSBezierPath and NSString classes.
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html

src/appkit.cs
```diff
@@ -6187,6 +6188,7 @@ interface NSFontChanging
 	[BaseType (typeof (NSObject))]
 	[DisableDefaultCtor] // crash at runtime (e.g. description). Documentation state: "You don’t create NSFont objects using the alloc and init methods."
 	partial interface NSFont : NSSecureCoding, NSCopying {
+		[ThreadSafe]
```
spouliot: same

> Font objects are immutable, so it is safe to use them from multiple threads in your app.
https://developer.apple.com/library/archive/documentation/TextFonts/Conceptual/CocoaTextArchitecture/FontHandling/FontHandling.html

and `NSFont` is mention in the same paragraph.

dalexsoto:
@spouliot makes sense, fixed! Thank you!

dalexsoto:
Test failure is unrelated:

* apitest/Mac Classic/NonSystem: https://github.com/xamarin/maccore/issues/626

PR #5254, opened 2018-12-11
[builds] Consume llvm + cross compilers from mono's package.

rolfbjarne:
This PR might be best reviewed commit-by-commit.

### DIFF
--- a/builds/Makefile
+++ b/builds/Makefile
@@ -1447,7 +1447,7 @@ clean-llvm: $(SDK_CONFIG)
 	$(MAKE) -C $(MONO_SDK_BUILDDIR) clean-llvm36-llvm32 clean-llvm-llvm64 $(SDK_ARGS)
 	$(RM) .stamp-*-llvm*
 
-.PHONY: install-llvm64 llvm32 llvm64
+.PHONY: install-llvm64 install-llvm32 llvm32 llvm64
 
 LLVM64_TARGETS = \
 	$(PREFIX)/LLVM/bin/opt \
--- a/builds/Makefile
+++ b/builds/Makefile
@@ -8,7 +8,7 @@ download: downloads/$(basename $(MONO_FILENAME))
 downloads/$(MONO_FILENAME):
 	$(Q) mkdir -p downloads
 	$(Q) echo "Downloading $(MONO_URL)..."
-	$(Q) curl -s -f -L $(MONO_URL) --output $@.tmp
+	$(Q) curl -f -L $(if $(V),-v,-s) $(MONO_URL) --output $@.tmp
 	$(Q) mv $@.tmp $@
 	$(Q) echo "Downloaded $(MONO_URL)"
 
--- a/builds/Makefile
+++ b/builds/Makefile
@@ -1509,9 +1509,11 @@ build:: build-$(1)
 install-local:: install-$(1)
 clean-local:: clean-$(1)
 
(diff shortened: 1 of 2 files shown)


builds/Makefile
```diff
@@ -1429,48 +1429,53 @@ ifdef INCLUDE_DEVICE
...
+	$(MAKE) -C $(MONO_SDK_BUILDDIR) provision-llvm-llvm64 $(SDK_ARGS)
 	$(Q) touch $@
 
-.stamp-build-llvm32: $(SDK_CONFIG)
-	$(MAKE) -C $(SDK_BUILDDIR) provision-llvm36-llvm32 $(SDK32_ARGS)
+.stamp-compile-llvm32: $(SDK_CONFIG)
+	$(MAKE) -C $(MONO_SDK_BUILDDIR) provision-llvm36-llvm32 $(SDK32_ARGS)
 	$(Q) touch $@
 
+.stamp-build-llvm64: .stamp-$(MONO_BUILD_MODE)-llvm64
```
luhenry: Where is `.stamp-download-llvm64`.

builds/Makefile
```diff
@@ -1429,48 +1429,53 @@ ifdef INCLUDE_DEVICE
...
 	$(PREFIX)/LLVM/bin/llc \
+
+LLVM32_TARGETS = \
 	$(PREFIX)/LLVM36/bin/opt \
-	$(PREFIX)/LLVM36/bin/llc
+	$(PREFIX)/LLVM36/bin/llc \
 
-$(PREFIX)/LLVM/bin/%: $(SDK_DESTDIR)/llvm-llvm64/bin/% | $(PREFIX)/LLVM/bin
-	$(call Q_2,INSTALL ,[LLVM64]) install -c -m 0755 $(INSTALL_STRIP_FLAG) $^ $@
+$(MONO_SDK_DESTDIR)/llvm-llvm64/bin/%: .stamp-build-llvm64; @true
```
luhenry: Why do you need `; @true`?

builds/Makefile
```diff
@@ -1429,48 +1429,53 @@ ifdef INCLUDE_DEVICE
...
+
 clean-llvm: $(SDK_CONFIG)
-	$(MAKE) -C $(SDK_BUILDDIR) clean-llvm36-llvm32 clean-llvm-llvm64 $(SDK_ARGS)
+	$(MAKE) -C $(MONO_SDK_BUILDDIR) clean-llvm36-llvm32 clean-llvm-llvm64 $(SDK_ARGS)
 	$(RM) .stamp-*-llvm*
 
-.PHONY: install-llvm64 llvm llvm64
-
-install-llvm: install-llvm32 install-llvm64
+.PHONY: install-llvm64 llvm32 llvm64
```
luhenry: Missing `install-llvm32`?

builds/Makefile
rolfbjarne: It's handled by a pattern rule: https://github.com/xamarin/xamarin-macios/blob/a4fa62a9d28b36942c0500c9efb6c761111c17bd/builds/Makefile#L25-L26

builds/Makefile
rolfbjarne: This is a pattern rule, and pattern rules are a bit special.

What's happening here is that in `$(MONO_SDK_DESTDIR)/llvm-llvm64/bin/*` is made by the dependency `.stamp-build-llvm64`, and not by any recipe for pattern rule. This means that in theory I could have made this rule simpler, by not including a recipe.

The problem is that make doesn't consider pattern rules without recipes. So without the `@true`, make  complains that it doesn't know how to build anything that this rule is supposed to match.

BTW this is identical to:

```make
$(MONO_SDK_DESTDIR)/llvm-llvm64/bin/%: .stamp-build-llvm64
    @true
```

it's just in one line instead of two.

builds/Makefile
rolfbjarne: Fixed.

PR #5262, opened 2018-12-12
[XHarness] Add filtering support to the NUnit and xUnit runners.

mandel-macaque:
First step to get to the point in which we can filter the execution of
the bcl tests via filters. This commit adds the filtering support to the
runners, which later will be used by xharness to pass those tests
filtered.

The end goal is to not skip a complete assembly but just those failing
tests.


mandel-macaque:
A small comment, number of tests and test results should not change since we are just adding the possibility to filter, not yet using it.

mandel-macaque:
@grendello @marek-safar please take a look :)

spouliot:
@luhenry has been working on the SDK integration (and his approval could replace the one from Marek)

PR #5264, opened 2018-12-12
[tests] fix symbols test for interpreter

lewurm:
`ves_pinvoke_method` / `do_icall` moved a frame up, so make the check
more flexible.


rolfbjarne:
build

PR #5270, opened 2018-12-13
[XHarness] Do not build the test assemblies for the iOS platforms.

mandel-macaque:
We now download the ios Mono sdk which means that there is no need for
us to build the test assemblies on iOS, WatchOS and tvOS BUT we need to
do so for the Mac tests.

The change includes:
  1. Use fo the downloaded test assemblies.
  2. Removal of the dependency of building them in the tests.
  3. Move back to using reflection for the project generation.
  4. Remove the cached project details added in the workaround.

PS: Assemblies use lower case in the SDK and some tests had to be updated.

### DIFF
--- a/tests/xharness/Harness.cs
+++ b/tests/xharness/Harness.cs
@@ -34,11 +34,6 @@ namespace xharness
 			get { return Path.GetFullPath (Path.Combine (RootDirectory, "..", "tools", "xibuild", "xibuild")); }
 		}
 
-		public string DownloadsPath
-		{
-			get { return Path.GetFullPath (Path.Combine (RootDirectory, "..", "builds", "downloads")); }
-		}
-
 		public static string Timestamp {
 			get {
 				return $"{DateTime.Now:yyyyMMdd_HHmmss}";
@@ -91,6 +86,7 @@ namespace xharness
 		public string JENKINS_RESULTS_DIRECTORY { get; set; } // Use same name as in Makefiles, so that a grep finds it.
 		public string MAC_DESTDIR { get; set; }
 		public string IOS_DESTDIR { get; set; }
+		public string MONO_SDK_DESTDIR { get; set; }
 		public bool IncludeMac32 { get; set; }
 
 		// Run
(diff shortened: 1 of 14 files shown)


mandel-macaque:
@rolfbjarne AFAIK we are not yet downloading the Mac SDK, correct?

tests/xharness/Harness.cs
```diff
@@ -34,6 +34,11 @@ public class Harness
 			get { return Path.GetFullPath (Path.Combine (RootDirectory, "..", "tools", "xibuild", "xibuild")); }
 		}
 
+		public string DownloadsPath
+		{
+			get { return Path.GetFullPath (Path.Combine (RootDirectory, "..", "builds", "downloads")); }
```
rolfbjarne: You can get the location of the downloaded SDK like this: https://gist.github.com/rolfbjarne/f5108a574f46dd2fcbe639d84c02e3d1

That should simplify some code (and it'll work even if you end up with multiple downloads after mono is bumped).

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -260,6 +258,24 @@ public BCLTestProjectGenerator (string outputDirectory, string monoRootPath, str
 			RegisterTypesTemplatePath = registerTypesTemplatePath ?? throw new ArgumentNullException (nameof (registerTypesTemplatePath));
 		}
 
+		string GetReleaseDownload (Platform platform)
+		{
+			if (string.IsNullOrEmpty (DownloadsPath))
+				return null;
+			switch(platform) {
```
rolfbjarne: Style: space before opening parenthesis.

tests/xharness/Harness.cs
mandel-macaque: Ah, perfect!

mandel-macaque: Updated after the chat we had in slack.

mandel-macaque:
Jenkins is down for maintenance. We might have to retry this later when ever is up.

PR #5288, opened 2018-12-18
[XHarness] Add support for ignore files in the blc tests

mandel-macaque:
This PR adds support to the use of *.ignore files for iOS/tvOS/WatchOS. The files are very similar to those that are use in xtro. We have:

* common-{TestProjectName}.ignore: For those tests that fail in all platforms.
* tvOS-{TestProjectName}.ignore: For tvOS specific failures.
* watchOS-{TestProjectName}.ignore: For watchOS specific failures.
* iOS-{TestProjectName}.ignore: for iOS specific projects.

Two test projects that were ignored previously have been added to show that the code does as advertised:

*  SystemDataXunit: xUnit test project example
* SystemServiceModelWebTests: NUnit test project example.

The files take the name of the failing tests, although we do support ignoring by class/namespace and other combinations, that would make the *.ignore files format more complicated and it is not worth it.


tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/NUnitTestRunner.cs
```diff
@@ -169,7 +169,6 @@ public void TestFinished (ITestResult result)
 					sb.Append ("\t[INFO] ");
 					break;
 				}
-				TotalTests++;
```
mandel-macaque: Issue found will testing the feature, we were reporting two times the real number of tests.

tests/bcl-test/BCLTests/templates/common/IgnoreFileParser.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace BCLTests {
+
+	/// <summary>
+	/// Class that can parser a file/stream with the ignored tests and will
+	/// return a list of the ignored tests.
+	/// </summary>
+	public static class IgnoreFileParser {
+
+		public static async Task<string[]> ParseStream (TextReader textReader)
```
rolfbjarne: Minor: async methods are typically suffixed `Async`: `ParseStreamAsync`.

spouliot: can't you use a `IEnumerable<strig>` here ?
that way you can return the `List<string>` itself, without a (allocating) call to `ToArray`

tests/bcl-test/BCLTests/templates/iOSApp/ViewController.cs
```diff
@@ -56,7 +58,26 @@ protected virtual void TerminateWithSuccess ()
 		}
 #endif
 
-		public override void ViewDidLoad ()
+		public async Task<string[]> GetSkippedTests ()
+		{
+			var ignoredFiles = new List<string> ();
+			// the project generator added the required resources,
+			// we extract them, parse them and add the result
+			var executingAssembly = Assembly.GetExecutingAssembly();
```
rolfbjarne: Style: space before opening parenthesis.

tests/bcl-test/BCLTests/templates/iOSApp/ViewController.cs
```diff
@@ -56,7 +58,26 @@ protected virtual void TerminateWithSuccess ()
...
+			foreach (var resourceName in executingAssembly.GetManifestResourceNames ()) {
+				if (resourceName.EndsWith (".ignore", StringComparison.Ordinal)) {
+					using (var stream = executingAssembly.GetManifestResourceStream(resourceName))
```
rolfbjarne: Style: space before opening parenthesis.

tests/bcl-test/BCLTests/templates/iOSApp/ViewController.cs
```diff
@@ -56,7 +58,26 @@ protected virtual void TerminateWithSuccess ()
...
+					using (var reader = new StreamReader(stream)) {
```
rolfbjarne: Style: space before opening parenthesis.

mandel-macaque:
We should see the new assemblies being tested in the test result page with no failures since the failing ones have been ignored.

tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/TestMethodFilter.cs
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		public TestMethodFilter (IEnumerable<string> methods)
+		{
+			if (methods == null)
+				throw new ArgumentNullException (nameof (methods));
+
+			foreach (string m in methods) {
+				AddMethod (m);
+			}
```
spouliot: use `AddRange` ?

tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/TestMethodFilter.cs
```diff
@@ -0,0 +1,66 @@
...
+
+		public void AddMethod (string method)
+		{
+			if (string.IsNullOrEmpty (method))
+				throw new ArgumentException ("must not be null or empty", nameof (method));
+
+			if (methods == null)
+				methods = new List <string> ();
+			if (methods.Contains (method))
+				return;
```
spouliot: checking for dupe could be doe while loading the entries (from the file)

tests/bcl-test/BCLTests/templates/tvOSApp/ViewController.cs
```diff
@@ -54,7 +56,26 @@ protected virtual void TerminateWithSuccess ()
 		}
 #endif
 
-		public override void ViewDidLoad ()
+		public async Task<string[]> GetSkippedTests ()
```
rolfbjarne: It looks like the code in this method is repeated several times, maybe add an `IgnoreFileParser.ParseAssemblyResource (Assembly)` method:

```csharp
class IgnoreFileParser {
    public async Task<string[]> ParseAssemblyResourcesAsync (Assembly asm)
    {
        var ignoredFiles = new List<string> ();
        // the project generator added the required resources,
        // we extract them, parse them and add the result
        foreach (var resourceName in asm.GetManifestResourceNames ()) {
            if (resourceName.EndsWith (".ignore", StringComparison.Ordinal)) {
                using (var stream = executingAssembly.GetManifestResourceStream(resourceName))
                    using (var reader = new StreamReader (stream)) {
                        var ignored = await ParseStream (reader);
                        // we could have more than one file, lets add them
                        ignoredFiles.AddRange (ignored);
                    }
                }
            }
        return ignoredFiles.ToArray ();
    }
}
```

tests/bcl-test/BCLTests/templates/iOSApp/ViewController.cs
```diff
@@ -56,7 +58,26 @@ protected virtual void TerminateWithSuccess ()
...
 #endif
 
-		public override void ViewDidLoad ()
+		public async Task<string[]> GetSkippedTests ()
+		{
+			var ignoredFiles = new List<string> ();
+			// the project generator added the required resources,
+			// we extract them, parse them and add the result
+			var executingAssembly = Assembly.GetExecutingAssembly();
+			foreach (var resourceName in executingAssembly.GetManifestResourceNames ()) {
```
spouliot: would not it be simpler to add files as "content" in the project ? and probe for `*.ignore` at runtime
that way a change to the ignore file won't trigger an assembly re-compilation (much slower for AOT)

tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/TestMethodFilter.cs
mandel-macaque: Yes, and no. In the sense, we want to share later this with android, and we do not know if they will use *.ignore files as we do or not. I'll ping them about it. If they use *ignore files, well, we can do it over there.

spouliot: good point

tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/TestMethodFilter.cs
```diff
@@ -0,0 +1,66 @@
...
+			}
+		}
+
+		public void AddMethod (string method)
+		{
+			if (string.IsNullOrEmpty (method))
+				throw new ArgumentException ("must not be null or empty", nameof (method));
+
+			if (methods == null)
+				methods = new List <string> ();
```
jonpryor: Should `methods` be a `HashSet<string>` if you don't allow duplicates?

tests/bcl-test/BCLTests/templates/common/TestRunner.NUnit/TestMethodFilter.cs
jonpryor: As your `.ignore` files are `@(EmbeddedResource)` entries, Android *can* use them.  (If they were `@(Content)`, that wouldn't work, as Android doesn't support `@(Content)`.)

Android already uses a *similar* mechanism, looking specifically for the [`nunit-excluded-tests.txt`](https://github.com/xamarin/xamarin-android/blob/f5132708f0652a0f29eef6e832ae7f21cebd71dc/tests/BCL-Tests/Xamarin.Android.Bcl-Tests/NUnitInstrumentation.cs#L63) or [`xunit-excluded-tests.txt`](https://github.com/xamarin/xamarin-android/blob/master/tests/BCL-Tests/Xamarin.Android.Bcl-Tests/XUnitInstrumentation.cs#L52) resources.

A question though: do we want/need *separate* lists for NUnit vs. xUnit?  Android currently has separate lists, but I don't know why, and @grendello is currently on vacation.

If we *don't* need separate lists, then `*.ignore` resources should be fine.

If we *do* want separate lists, we may want a different naming convention to separate them from each other.

A different issue with `*.ignore` isn't the mechanics in this PR, but the mechanics of git, and mixed macOS/Windows use.  By default, `*.ignore` won't be treated as a "text" file (unless you update `.gitattributes`), while `.txt` *will* be treated as text.

As such, perhaps instead of `*.ignore` it might be better to use `*.ignore.txt`?

mandel-macaque: @jonpryor we use *.ignore because we have the same ignoring pattern in other tests from iOS, but it we want to share this, I personally have no issue in changing the naming to *.ignore.txt (as long as @spouliot is ok with that).

We do not make a difference in the ignore files names between nunit and xunit because, when we generate the test projects, we know in advance if we are dealing with a NUnit test project or a xUnit one, since we know that:

* "*_test.dll" -> NUnit
* "*_xunit-test.dll" -> xUnit

So we have a list per testing project (to make it easier to maintain). I suppose that the difference in your side resides in the fact that you do not AFAIK have a filter that works on methods for the NUnit runner (something that I am adding in this PR). 

About resource files, I can add a method to load from @(Content) and other for resource files, each platform can decide how the files are added. We could use content to speed our rebuild times and leave the resources file approach there for the android team.

mandel-macaque: Let me know your opinion and I'll do push regarding the use or resource. All other comments have been addressed :)

spouliot: The extension name is not important, it just needs to be consistent (at least across unit tests).

tests/bcl-test/BCLTests/templates/iOSApp/ViewController.cs
spouliot: Maybe there could be more code sharing with XA if they already have this working ?

mandel-macaque: Resolving since I moved it to content files.

Issue #5289, opened 2018-12-18
CFNetworkHandler hangs of no data is received


spouliot:
### Steps to Reproduce

1.  Use test case from 
https://github.com/xamarin/xamarin-macios/issues/5190#issuecomment-448285868
2. Supply an instance of `CFNetworkHandler` to the test code
3. Run

### Expected Behavior

`PASS` after 120 seconds

### Actual Behavior

Hangs

### Environment

Stable and master

### Notes

This does not seems related to the timeout itself. The URL used simply returns `200` after the specified time (120 seconds). The handler receive an *Open* and the a *Close* event - but never a *DataAvailable* - and that seems to be avoiding some _required_ code to avoid the hang.

spouliot:
The issue is not related to the timeout. I have a fix (for mono), but it will need to be updated after https://github.com/mono/mono/pull/12047 gets merged.

PR #5290, opened 2018-12-18
[foundation] Respect HttpClient.Timeout for NSUrlSessionHandler. Fix #5190

### LINKED ISSUE #5190: [Foundation]NSUrlSessionHandler does not respect the timeout that is set in the HttpClient.
Following the description found in https://developercommunity.visualstudio.com/content/problem/388530/httpclienttimeout-values-set-above-default-100-sec.html I have been able to confirm that when using the NSUrlSessionHandler as the message handler for the HttpClient, we do not respect the given timeout. The following code can be used as an example:

```CSharp
 HttpClient client = new HttpClient();
 client.Timeout = TimeSpan.FromSeconds (150);
 client.BaseAddress = new Uri("http://slowwly.robertomurray.co.uk");
 client.DefaultRequestHeaders.Accept.Clear();
 client.DefaultRequestHeaders.Accept.Add(new System.Net.Http.Headers.MediaTypeWithQualityHeaderValue("application/json"));
 HttpResponseMessage response = await client.GetAsync("http://slowwly.robertomurray.co.uk/delay/160000/url/http://www.google.co.uk");
response.EnsureSuccessStatusCode();
```

This happens because, when the message handler is instantiated the NSUrlSessionConfiguration does not has its TimeoutIntervalForRequest set to the correct value and uses the default, which is 60 seconds. 

The solution would be to set the timeout used by the handler so that when the configuration is created the correct timeout is used.

spouliot:
When `HttpClient` is used it might not be possible to set custom
properties on the handler.

This PR avoids a fight between the `HttpClient.Timeout` and the ones that
`NSURLSession` provides - making its use, as default, working as expected

It is still possible to set those custom properties when creating the
`NSUrlSessionHandler` manually, so there's no loss of functionalities.

Adding a unit test would be tricky since it depends on external sites and
requires "enough" delays to trigger (both leading to false positives over
time).

Notes
* HttpClientHandler timeout is broken -> https://github.com/mono/mono/issues/12100
* CFNetworkHandler is broken when no data is received -> https://github.com/xamarin/xamarin-macios/issues/5289

Fixes https://github.com/xamarin/xamarin-macios/issues/5190

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -121,7 +121,18 @@ namespace Foundation {
 		readonly Dictionary<NSUrlSessionTask, InflightData> inflightRequests;
 		readonly object inflightRequestsLock = new object ();
 
-		public NSUrlSessionHandler () : this (NSUrlSessionConfiguration.DefaultSessionConfiguration)
+		static NSUrlSessionConfiguration CreateConfig ()
+		{
+			// we want the current (at the moment the instance creation is done) configuration defaults
+			var config = (NSUrlSessionConfiguration) NSUrlSessionConfiguration.DefaultSessionConfiguration.Copy ();
+			// but we want, by default, the timeout from HttpClient to have precedence over the one from NSUrlSession
+			// Double.MaxValue does not work, so default to 24 hours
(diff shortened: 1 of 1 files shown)


spouliot:
@monojenkins backport to d16-0

PR #5321, opened 2018-12-21
[coretext] Simplify CTRunDelegate code. Fixes #5303

### LINKED ISSUE #5303: CTRunDelegate performance enhancement
As mentioned in https://github.com/xamarin/xamarin-macios/issues/5132#issuecomment-439123600 there is potential for much better performance (like removing reflection) inside CTRunDelegate

This is a split off from #5132 since it covered a bug (originally) and it should not be mixed up (or delayed) to investigate performance enhancements.

spouliot:
Reflection is simply not needed.

Existing unit tests confirms the virtual method are called properly.

reference: https://github.com/xamarin/xamarin-macios/issues/5303

### DIFF
--- a/src/CoreText/CTRunDelegate.cs
+++ b/src/CoreText/CTRunDelegate.cs
@@ -25,7 +25,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 using System;
-using System.Reflection;
 using System.Runtime.InteropServices;
 
 using ObjCRuntime;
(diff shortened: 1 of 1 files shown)


src/CoreText/CTRunDelegate.cs
```diff
@@ -125,24 +124,10 @@ internal CTRunDelegateCallbacks GetCallbacks ()
 			var callbacks = new CTRunDelegateCallbacks () {
 				version = CTRunDelegateVersion.Version1,
 				dealloc = Deallocate,
+				getAscent = GetAscent,
```
rolfbjarne: I think the GC can collect these delegates early (this is not something this PR changes though).

I guess we could store the callbacks in the CTRunDelegate instance.

PR #5348, opened 2019-01-04
[arkit] Fix crash in ARConfiguration.SupportedVideoFormats. Fixes #5347

### LINKED ISSUE #5347: Calling ARConfiguration.SupportedVideoFormats crash
### Steps to Reproduce

1.  call `ARConfiguration.SupportedVideoFormats`

### Expected Behavior

no crash

### Actual Behavior

Native exception `NSInvalidArgumentException Reason: Supported video formats should be called on individual configuration class` leading to crash

spouliot:
Even if 'supportedVideoFormats' is static the type is abstract.

> Important
> ARConfiguration is an abstract base class, so its implementation of
> this property always returns an empty array. Read this property from
> the configuration subclass you plan to use for your AR session, such
> as ARWorldTrackingConfiguration or ARFaceTrackingConfiguration.
https://developer.apple.com/documentation/arkit/arconfiguration/2942261-supportedvideoformats?language=objc

and this behave differently in Objective-C (than .net) as every (static)
method will be different (and not a single implementation like C#).

The existing implementation (as a property) calls `ARConfiguration`
implementation which simply throws a (native) exception

> NSInvalidArgumentException Reason: Supported video formats should be called on individual configuration class

The solution is to obsolete the property (can't be subclassed in .net
since it's static) and add, only on the subclasses, a method that
call the 'supportedVideoFormats' selector on the current (not base)
type.

Added unit test to detect the addition of newer subclasses - since they
will also need to expose this method.

reference: https://github.com/xamarin/xamarin-macios/issues/5347

### DIFF
--- a/tests/monotouch-test/ARKit/ARConfigurationTest.cs
+++ b/tests/monotouch-test/ARKit/ARConfigurationTest.cs
@@ -1,0 +1,47 @@
+﻿#if XAMCORE_2_0 && __IOS__
+
+using System;
+using System.Reflection;
+using ARKit;
+using Foundation;
+using NUnit.Framework;
+
+namespace MonoTouchFixtures.ARKit {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class ARCondigurationTest {
+
+		[Test]
+		public void SupportedVideoFormats ()
+		{
+			var svf = ARConfiguration.SupportedVideoFormats;
+			Assert.That (svf, Is.Empty, "empty");
+		}
+
+		[Test]
+		public void GetSupportedVideoFormats ()
+		{
+			Assert.NotNull (ARWorldTrackingConfiguration.GetSupportedVideoFormats (), "ARWorldTrackingConfiguration");
+			Assert.NotNull (AROrientationTrackingConfiguration.GetSupportedVideoFormats (), "AROrientationTrackingConfiguration");
+			Assert.NotNull (ARFaceTrackingConfiguration.GetSupportedVideoFormats (), "ARFaceTrackingConfiguration");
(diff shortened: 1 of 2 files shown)


tests/monotouch-test/ARKit/ARConfigurationTest.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace MonoTouchFixtures.ARKit {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class ARCondigurationTest {
+
+		[SetUp]
+		public void Setup ()
+		{
+			TestRuntime.AssertXcodeVersion (10,0);
```
VincentDondain:
```suggestion
			TestRuntime.AssertXcodeVersion (9,3);
```

Given that those APIs are `[iOS (11,3)]`

spouliot:
@VincentDondain 

A1. Our existing tests did not caught it because there was no test before I added mine ;-) 

Do not assume **intro** runs the generated binding code - most of the generated code has too much prerequisite (e.g. arguments) to automate it. So in most cases **intro** checks if selectors respond (and it does in this case) and that the signature match (and that was also fine).

Now running the code for static properties is possible (no arguments, no instance to create) and worth trying (given time) but it will likely cause a lot of false positives (that will need to be audited and ignored - like many intro tests).

Even then (abstract or not) the base (overridden) code might not throw a native exception (first case I see) so it might not find any such case by doing so...


A2. it's not frequent - but it happened a few times before (e.g. Intents).

It's not really possible to detect from code (at least from a case like this one) since the non-override case is common (and works just like C#). So it falls to the _human_ binder realize the bindings are (or need to be) duplicated in subclasses (either from docs, headers comments...).

Also having an attribute _work_ on subclasses is difficult since there's no ordering in generator (i.e. the subclass could already have been generated).

PR #5360, opened 2019-01-09
[compare-commits] Only validate files when validating if files have unexpectedly changed.

rolfbjarne:
Fixes this problem:

    The following files were modified, and they shouldn't have been:
        /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/src

which shouldn't be reported, since the "file" in question is a directory.

Reference: https://github.com/xamarin/xamarin-macios/pull/5355#issuecomment-452400517


rolfbjarne:
> make sense but any clue why this was not an issue before ?

I have a suspicion that something like this might be the cause: 0772f6a9c634f633139bbd9d91043420f8729f0a, but I haven't confirmed it.

PR #5372, opened 2019-01-10
[tests] Run tests-docs using xharness.

rolfbjarne:
Jenkins has a limitation where you can't mark a step a failure, it has to
*fail* to be reported as such.

This means that running multiple tests, and reporting a failure if any of
those tests fail is not possible.

We run into this with the normal test run + the docs tests; where we currently
don't show a red result in the UI if any of those fail.

So incorporate the test-docs step into xharness, so that we only have one
thing to in the Jenkinsfile, which makes it possible for us to fail the test
run step properly.

This also required a few upgrades to xharness to get more info for pull
requests, since the logic to enable the docs tests is a bit more complicated
than anything else we have (if the current branch (or the target branch for a
PR) is 'master' AND xamarin mode is enabled).

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -15,6 +15,7 @@ errorMessage = null
 currentStage = null
 failedStages = []
 workspace = null
+manualException = false
 
 xiPackageFilename = null
 xmPackageFilename = null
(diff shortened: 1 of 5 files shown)


rolfbjarne:
The above test run didn't run the docs tests (because the conditions to automatically enable them weren't fulfilled), but I made a [commit](
https://github.com/xamarin/xamarin-macios/commit/7d1921699057825b4dc2374c84d8b5ccfa821dfa#commitcomment-31888628) to force them to be executed, which passed ([Html Report](http://xamarin-storage/jenkins/xamarin-macios/jenkins-improve-docs-tests-rendering4/7d1921699057825b4dc2374c84d8b5ccfa821dfa/1/jenkins-results/tests/index.html)).

PR #5373, opened 2019-01-10
[XHarness] Reneable the System.IO.Compression tests.

mandel-macaque:
The ignored tests are because there are files missing needed that are not included as resources.

Fixes https://github.com/xamarin/maccore/issues/1146


mandel-macaque:
Mono issue https://github.com/mono/mono/issues/12356

PR #5378, opened 2019-01-11
[xharness] Move javascript and css to external files.

rolfbjarne:
This makes editing them a bit simpler.

There are no functional changes.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -1076,6 +1076,7 @@ namespace xharness
 							rv.Add (v);
 							return rv;
 						});
+						string serveFile = null;
 						switch (request.Url.LocalPath) {
 						case "/":
 							response.ContentType = System.Net.Mime.MediaTypeNames.Text.Html;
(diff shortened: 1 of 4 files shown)


tests/xharness/Jenkins.cs
```diff
@@ -1425,6 +1438,10 @@ public void GenerateReport ()
 							File.Delete (Harness.MarkdownSummaryPath);
 						File.Move (tmpmarkdown, Harness.MarkdownSummaryPath);
 					}
+					var dependentFileLocation = Path.GetDirectoryName (System.Reflection.Assembly.GetExecutingAssembly ().Location);
+					foreach (var file in new string [] { "xharness.js", "xharness.css" }) {
```
mandel-macaque: Not something that should be changed in the PR, but what about, since we have a dir name, use a regex to get all *.js and *.css and do the  same operation?

tests/xharness/Jenkins.cs
```diff
@@ -1546,290 +1563,15 @@ void GenerateReportImpl (Stream stream, StreamWriter markdown_summary = null)
...
-#nav li:hover ul a { 
-	text-decoration:none;
-}
-#nav li:hover ul li a:hover {
-	text-decoration: underline;
-}
-
-</style>");
+				writer.WriteLine ("<head>");
+				writer.WriteLine ("<link rel='stylesheet' href='xharness.css'>");
```
mandel-macaque: Following the previous, maybe doing a more general thing will be nice so that we just have to add new css files in a dir and forget about it.

PR #5379, opened 2019-01-11
[XHarness] Reenable the System.Data tests from the bcl.

mandel-macaque:
The reason why some tests fail:

1. Some files are missing that are needed by the tests.
2. The mono sdk does not provide watch specific tests, we run the iOS and ignore those that are of classes not supported by the platform.

Fixes https://github.com/xamarin/maccore/issues/1149


mandel-macaque:
mono issue: https://github.com/mono/mono/issues/12384

PR #5382, opened 2019-01-11
[XHarness] Remove not longer present bcl tests.

mandel-macaque:
As of Mono SDK 25a8be91115255341dd80ce4a225415feec46f56 we do not longer
have the test dll and therefore the test project can be removed.

Set the issues to be fixed because we not longer ignore them, they are not longer present.

Fixes https://github.com/xamarin/maccore/issues/1150
Fixes https://github.com/xamarin/maccore/issues/1151


mandel-macaque:
If you want to double check, you can do:

```bash
tree -f builds/downloads/ios-release-Darwin-25a8be91115255341dd80ce4a225415feec46f56/ios-bcl/monotouch/tests/ | grep monotouch_System.Data.DataSetExtensions_test.dll
```

mandel-macaque:
Also FYI whenever this happens, tests disappear, we will have an xharness exception, so we are ready for this situations in the monthly merge branches.

mandel-macaque:
Setting to "do not merge" until we get feedback from mono.

akoeplinger:
The System.Data.DataSetExtensions.dll was never shipped with XI or XM so this should be fine 👍

PR #5384, opened 2019-01-11
[XHarness] Reenable the System.ServiceModel tests from the bcl.

mandel-macaque:
Failing tests are because:

1. Files are missing for the tests.
2. We yet do not have tests for the watch platform, we execute the iOS ones and ignore those that are not supported in the platform.


mandel-macaque:
Mono issue: https://github.com/mono/mono/issues/12417 I'll add it too in the commit msg.

PR #5392, opened 2019-01-14
[Foundation] Add a couple of missing NSDate properties and methods

migueldeicaza:


src/foundation.cs
```diff
@@ -3586,6 +3586,18 @@ interface NSDate : NSSecureCoding, NSCopying, CKRecordValue {
...
+		[Export ("timeIntervalSinceDate:")]
+		double TimeIntervalSince (NSDate anotherDate);
+
+		[Export ("timeIntervalSinceNow")]
+		double TimeIntervalSinceNow { get; } 
+
+		[Export ("timeIntervalSince1970")]
+		double TimeIntervalSince1970 { get; } 
+
+		[Export ("timeIntervalSinceReferenceDate")]
```
rolfbjarne: `timeIntervalSinceReferenceDate` is already bound as `SecondsSinceReferenceDate`: https://github.com/xamarin/xamarin-macios/blob/4741884e68e2dd294070baeb13ed3977ebd56cc6/src/foundation.cs#L3586-L3587

src/foundation.cs
```diff
@@ -3586,6 +3586,18 @@ interface NSDate : NSSecureCoding, NSCopying, CKRecordValue {
 		[Export ("timeIntervalSinceReferenceDate")]
 		double SecondsSinceReferenceDate { get; }
 
+		[Export ("timeIntervalSinceDate:")]
+		double TimeIntervalSince (NSDate anotherDate);
```
spouliot: it's a method so [it should have a verb](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods), e.g. `GetTimeIntervalSince (...)`

src/foundation.cs
spouliot: Might be worth being consistent and having the new members follow the same name `SecondsSince...`
which is a bit more helpful than `TimeInterval` since that name requires reading the doc to know the scale unit.

mandel-macaque:
API issues can be ignored (I need to fix this error).

PR #5397, opened 2019-01-14
[XHarness] Reenable System.Core tests on ios/tvos/watchos

mandel-macaque:
Fixes https://github.com/xamarin/maccore/issues/1143


mandel-macaque:
@spouliot don't worry, I'll always add the mono issue. I just want to get the green build before I add them so that the issue points to a working PR.

mandel-macaque:
Mono issue: https://github.com/mono/mono/issues/12419

Will also be added in the commit msg.

PR #5400, opened 2019-01-14
[XHarness] Remove not needed bcl tests for Mono.Runtime.

mandel-macaque:
Remove tests that are not provided in the Mono iOS SDK for Mono.Runtime.

Fixes https://github.com/xamarin/maccore/issues/1141


mandel-macaque:
As with a previous PR, the test assembly was added when we built the tests locally, but this dll is not present in the iOS Mono SDK.

You can double check with:

```bash
tree builds/downloads/ios-release-Darwin-25a8be91115255341dd80ce4a225415feec46f56/ios-bcl/ | grep monotouch_Mono.Runtime.Tests_test.dll
```

PR #5410, opened 2019-01-15
[XHarness] Reneable the SystemSecurity tests on iOS, TvOS and WatchOS

mandel-macaque:
Some of the test classes are abstract, and we needed to change the
generator to ignore that fact. Before the change, we would have 0 test
cases.

Failing tests are ignored.

Fixes https://github.com/xamarin/maccore/issues/1128


mandel-macaque:
Merging after adding a mono issue regarding the failing tests: https://github.com/mono/mono/issues/12446

PR #5412, opened 2019-01-15
[XHarness] Reneable Mono.Security tests on WatchOS.

mandel-macaque:
Fixes https://github.com/xamarin/maccore/issues/1142


mandel-macaque:
Had a conversation with the mono team, the ignored test is due to the fact that we are running the iOS tests, but this is an issue only in 2018-08, 2018-10 does have the correct dlls.

PR #5413, opened 2019-01-15
[d16-0 P3][msbuild/tests] Ensure our FrameworkList files have correct assemblies (#5367)

VincentDondain:
One mono bump (https://github.com/xamarin/xamarin-macios/commit/6f2ebedb74fe8da250c8f53b61ffae2eb86fc711#diff-e801bb766cbaad95b50b1487b865f971) changed our `Xamarin.iOS-FrameworkList.xml.in` (and the tvOS and watchOS ones) and added 2 files

```
<File AssemblyName="System.Buffers" Version="4.0.99.0" />
<File AssemblyName="System.Memory" Version="4.0.99.0" />
```

Around the same time this change was made in the IDE: https://github.com/mono/monodevelop/blob/e355f65870da4847d77ed1f3f2d07d1f78889cec/main/src/core/MonoDevelop.Core/MonoDevelop.Core.Assemblies/TargetFramework.cs#L260
That means that VSMac, when creating a new cache file listing the assemblies, won’t scan the assemblies in those directories `/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.iOS`.
This resulted in an incomplete cache file missing most assemblies leading to the IDE showing red checkmarks for known and installed assemblies.

Ideally the FrameworkList files we ship should have a list of all the assemblies with PublicKeyToken and friends (the framework shouldn't be lazy and expect the IDEs to generate that file for us).

This commit ships a test using the same logic as the IDE to populate the FrameworkList file. The test will report any missing assembly and print the xml element that should be added to the file.
E.g: `<File AssemblyName="FSharp.Core" Version="3.98.4.0" PublicKeyToken="b03f5f7f11d50a3a" ProcessorArchitecture="MSIL" />`
Then it's only a matter of copy/pasting that (:

Note: even though this is the msbuild tests for iOS land I'm also scanning the Xamarin.Mac files here for simplicity.

We also scan all assemblies in `/Facades`.

Discussed in slack but adding all the facades assemblies doesn't compromise the project in the IDE (e.g: we still can't add incorrect assemblies) and it's required for some NuGet conflict resolution.
See: https://github.com/xamarin/xamarin-macios/commit/ac8fd9e60aba0fae174eb7175a54d42a3c6b61f1


GouriKumari:
QA validation for this PR passed. Comments from GL team below.

 Observations:

 - I can successfully able to add references.

 - Cross checked & confirmed that System.Buffer or System.Memory is not displayed under add reference.

 - Red checkmarks not displayed in an iOS project's references.

 - iOS project build & deployed successfully.

VSFM : 

Build gist: https://gist.github.com/GLjackyvaswani/c996fa3deba432b45d01b4a5164cbaf5

XI build: https://bosstoragemirror.blob.core.windows.net/wrench/jenkins/master/9d37f9583f083e7d431dae763d729d818d54d4a6/332/package/xamarin.ios-12.7.0.53.pkg

All other builds taken from: https://github.com/xamarin/release-manifests/blob/vsmac-8.0/manifests/vsmac-8.0-p2.json

Windows:

Build gist: https://gist.github.com/GLjackyvaswani/8f2ae8a4bb3f2086b223ce5ee8a9fcbe
Installed provided VSIX on top of the 16.0.0 preview 1.1 Market build.
VSIX Installed: https://bosstoragemirror.blob.core.windows.net/xvs-core/c9/c920f0940017e34c534fe89898bbf15d4600b1af/artifacts.json

VincentDondain:
@chamons it's approved by QA so it sounds like you only have the power to merge it (d16-0).
#ivegotthepower

Issue #5416, opened 2019-01-15
[NullAllowed] can be applied on methods leading to easy (review) mistakes


spouliot:
Right now this case is silently ignored (both by the compiler and by the generator) which can lead to mistakes that are easy to miss while reviewing bindings.

### Steps to Reproduce

1. Apply diff

```diff
diff --git a/src/generator-attributes.cs b/src/generator-attributes.cs
index 965d8469..77162253 100644
--- a/src/generator-attributes.cs
+++ b/src/generator-attributes.cs
@@ -250,6 +250,7 @@ public class IsThreadStaticAttribute : Attribute {
 
 // When applied to a member, generates the member as static
 // and passes IntPtr.Zero or null if the parameter is null
+[AttributeUsage (AttributeTargets.Property | AttributeTargets.ReturnValue | AttributeTargets.Parameter)]
 public class NullAllowedAttribute : Attribute {
        public NullAllowedAttribute () {}
 }
```

2. Build

### Expected Behavior

no error

### Actual Behavior

Many errors (needs to be reviewed)

Some case are "semi-valid" like putting `[NullAllowed]` on a `set` since the intent is kind of obvious. However if there's many parameters the intent is unclear.

### Environment

master

spouliot:
E.g. inside `audiounit.cs` we have

```csharp
		[NullAllowed, Export ("setInputHandler:")]
		void SetInputHandler (AUInputHandler handler);
```

which presently generates a null check

```csharp
if (handler == null)
	throw new ArgumentNullException ("handler");
```

which is unlikely the intent (nor what people think they reviewed)

spouliot:
re-opening PR #5417 is not a complete fix (only bindings, not tooling)

PR #5417, opened 2019-01-16
Clarify usage of `[NullAllowed]` in our bindings

spouliot:
First part to fix https://github.com/xamarin/xamarin-macios/issues/5416

We currently allow the `[NullAllowed]` attribute anywhere an attribute
can be used on metadata (i.e. no `AttributeUsage` is used).

However the generator only process the attribute in some specific cases,
*silently* ignoring others. This leads to situations such as

```csharp
[NullAllowed, Export ("setInputHandler:")]
void SetInputHandler (AUInputHandler handler);
```

where a `null` argument will throw an `ArgumentNullException` because
`[NullAllowed]` does not mean anything on a method declaration.

To avoid such confusion `[NullAllowed]` should be added on each parameter
(even if all of them requires it) and if a `null` return value is
possible then use `[return: NullAllowed]`.

This PR allows XI/XM bindings to be built, without error, when this
patch is applied.

```diff
diff --git a/src/generator-attributes.cs b/src/generator-attributes.cs
index 965d8469..77162253 100644
--- a/src/generator-attributes.cs
+++ b/src/generator-attributes.cs
@@ -250,6 +250,7 @@ public class IsThreadStaticAttribute : Attribute {

 // When applied to a member, generates the member as static
 // and passes IntPtr.Zero or null if the parameter is null
+[AttributeUsage (AttributeTargets.Property | AttributeTargets.ReturnValue | AttributeTargets.Parameter)]
 public class NullAllowedAttribute : Attribute {
        public NullAllowedAttribute () {}
 }
```

Note that it's unlikely we'll apply this patch _as-is_ to avoid breaking
existing projects. A better approach (in a future PR) is to have the
generator (instead of the C# compiler) issue a warning (instead of an
error) when a `NullAllowed` is ignored.

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -608,7 +608,8 @@ namespace AppKit {
 		NSMenu MainMenu { get; set; }
 	
 		[Export ("helpMenu", ArgumentSemantic.Retain)]
-		NSMenu HelpMenu { get; [NullAllowed] set; }
+		[NullAllowed]
+		NSMenu HelpMenu { get; set; }
 	
 		[Export ("applicationIconImage", ArgumentSemantic.Retain)]
 		NSImage ApplicationIconImage { get; set; }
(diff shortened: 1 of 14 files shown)


spouliot:
known / unrelated issues
* monotouch-test/iOS Unified 32-bits - simulator/Debug (static registrar): Failed -> https://github.com/xamarin/maccore/issues/1345
monotouch-test/watchOS - simulator/Debug: Failed -> https://github.com/xamarin/maccore/issues/1347

PR #5418, opened 2019-01-16
[do-not-merge] ARM64_32 changes

lewurm:
Consider this a preview of the changes required to get ARM64_32 going, the architecture introduced with the Watch Series 4. This approach uses the armv7k AOT compiler, and then links the runtime built for arm64_32. This works because the AOT compiler emits bitcode and the ABI of armv7k and arm64_32 are "offset compatible". The caveat is that it only works for the Release mode, but that should be enough to unblock us for the March 2019 deadline.

I'm planning to land the relevant mono changes in `2018-10` (see https://github.com/lewurm/mono/commits/2018-08-watch4). For the changes on the `xamarin-macios`, I'll cherry-pick some of them to `master` and some I'll open against the `mono-2018-10` integration branch. The changes in `build/Makefile` will need some more polishing. The xharness changes are quite a hack, it should rather be a different platform target (similar to target32/target64 on iOS), but I didn't get that to work.

Context: https://github.com/mono/mono/issues/10641

Edit: "pics or it didn't happen"

screenshot 2019-01-15 at 23 18 38


spouliot:
You mention both `2018-10` and `2018-08-watch4` which seems (or should) differ ?

tools/mtouch/mtouch.cs
```diff
@@ -394,9 +395,11 @@ public static string GetAotArguments (Application app, string filename, Abi abi,
...
-			if (llvm_only)
+			if (llvm_only && interp) {
+				args.Append ("interp,llvmonly,");
+			} else if (llvm_only) {
 				args.Append ("llvmonly,");
-			else if (interp) {
+			} else if (interp) {
 				if (fname != "mscorlib.dll")
 					throw ErrorHelper.CreateError (99, $"Internal error: can only enable the interpreter for mscorlib.dll when AOT-compiling assemblies (tried to interpret {fname}). Please file an issue at https://github.com/xamarin/xamarin-macios/issues/new.");
 				args.Append ("interp,");
```
spouliot: adding the same value in multiple place makes things harder to understand (and easier to miss when updating).
I think you can have a single `if (interp) { args.Append ("interp,"); }`

tools/mtouch/mtouch.cs
```diff
@@ -583,9 +586,16 @@ public static string GenerateMain (Application app, IEnumerable<Assembly> assemb
...
 						sw.WriteLine ("\txamarin_profiler_symbol = mono_profiler_init_log;");
 
-					if (app.EnableLLVMOnlyBitCode)
+					if (app.EnableLLVMOnlyBitCode && app.UseInterpreter) {
+						sw.WriteLine ("\tmono_icall_table_init ();");
+						sw.WriteLine ("\tmono_marshal_ilgen_init ();");
+						sw.WriteLine ("\tmono_method_builder_ilgen_init ();");
+						sw.WriteLine ("\tmono_sgen_mono_ilgen_init ();");
+						sw.WriteLine ("\tmono_ee_interp_init (NULL);");
+						sw.WriteLine ("\tmono_jit_set_aot_mode (MONO_AOT_MODE_INTERP_LLVMONLY);");
```
spouliot: same, too easy to forget to update the (mostly) identical generated code.

lewurm:
@spouliot my prototype is based on `2018-08` (hence the branch name `2018-08-watch4`), but I'm going to land the changes for `2018-10` because it's a bit late for `2018-08`.

spouliot:
@lewurm talk with @chamons first. We do not have any date (including preview dates) for 16.1 release - i.e. next time we can take a newer mono.

lewurm:
closing in favor of https://github.com/xamarin/xamarin-macios/pull/5465

PR #5419, opened 2019-01-16
[xharness] recognize newer watches

lewurm:
Context: https://github.com/mono/mono/issues/10641

### DIFF
--- a/tests/xharness/Simulators.cs
+++ b/tests/xharness/Simulators.cs
@@ -734,6 +734,7 @@ namespace xharness
 		ARMv7k,
 		ARMv7s,
 		ARM64,
+		ARM64_32,
 		i386,
 		x86_64,
 	}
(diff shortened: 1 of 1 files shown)


tests/xharness/Simulators.cs
```diff
@@ -879,8 +880,20 @@ public class Device : IDevice
 				}
 
 				// https://www.theiphonewiki.com/wiki/List_of_Apple_Watches
-				if (model.StartsWith ("Watch", StringComparison.Ordinal))
-					return Architecture.ARMv7k;
+				if (model.StartsWith ("Watch", StringComparison.Ordinal)) {
+					var identifier = model.Substring ("iPod".Length);
```
rolfbjarne: `.Substring ("iPod")`? I think you're hitting the `default` case below when you've been testing :)

lewurm: d'oh 🙂 I actually do not have an older watch, so chances are good I'm breaking stuff...

PR #5428, opened 2019-01-17
[XHarness] Report issues with the TestFixtureSetup

mandel-macaque:
There are a number of tests that use a TestFixtureSetup method that
fails. While the xml results report these errors, if a user runs xharness locally,
he will not get the errors reported. This issue meant that we had
different errors reported in jenkis and locally.


mandel-macaque:
Known test issue (GestureRecognizerTest.NoStrongCycles): xamarin/maccore#1345

Issue #5408, opened 2019-01-15
generator does not handle `[NullAllowed]` correctly on protocols `_Extensions`


spouliot:
ref: https://github.com/xamarin/xamarin-macios/issues/5403

### Steps to Reproduce

1. Apply

```diff
diff --git a/src/appkit.cs b/src/appkit.cs
index daf22b75..48390cb8 100644
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -431,6 +431,7 @@ namespace AppKit {
        interface NSAppearanceCustomization {
 
                [Mac (10,9)]
+               [NullAllowed]
                [Export ("appearance", ArgumentSemantic.Strong)]
                NSAppearance Appearance { get; set; }
```

2. Build and check `src/build/mac/mobile/AppKit/NSAppearanceCustomization.g.cs`

### Expected Behavior

No null check (it's disabled) on the generated `Set*` method

### Actual Behavior

```csharp
		[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
		public static void SetAppearance (this INSAppearanceCustomization This, NSAppearance value)
		{
			global::AppKit.NSApplication.EnsureUIThread ();
			if (value == null)
				throw new ArgumentNullException ("value");
			global::ObjCRuntime.Messaging.void_objc_msgSend_IntPtr (This.Handle, Selector.GetHandle ("setAppearance:"), value.Handle);
		}
```

### Workaround

Use 
```diff
diff --git a/src/appkit.cs b/src/appkit.cs
index daf22b75..02f908f0 100644
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -432,7 +432,7 @@ namespace AppKit {
 
                [Mac (10,9)]
                [Export ("appearance", ArgumentSemantic.Strong)]
-               NSAppearance Appearance { get; set; }
+               NSAppearance Appearance { get; [NullAllowed] set; }
 
                [Mac (10,9)]
                [Export ("effectiveAppearance", ArgumentSemantic.Strong)]

```

to generate to appropriate code
```csharp
		[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
		public static void SetAppearance (this INSAppearanceCustomization This, NSAppearance value)
		{
			global::AppKit.NSApplication.EnsureUIThread ();
[...]

PR #5431, opened 2019-01-17
[appkit] Allow `null` on `NSAppearanceCustomization.Appearance`. Fixes #5403

### LINKED ISSUE #5403: NSAppearanceCustomization.Appearance should allow for null
`NSAppearanceCustomization.Appearance` (and `NSAppearanceCustomization_Extensions.SetAppearance`) should be able to receive `null`/`NIL` to indicate that appearance (`NSAppearanceCustomization.EffectiveAppearance`) should be inherited from its appearance source.

Our bindings are currently throwing `ArgumentNullException` when set to `null`.

https://developer.apple.com/documentation/appkit/nsappearancecustomization/1533925-appearance?language=objc

spouliot:
> The default value for this property is nil, which means that the
> receiver uses the appearance it inherits from the nearest ancestor that
> has set an appearance. When you set appearance to a non-nil value, the
> receiver and the views it contains use the specified appearance.
> https://developer.apple.com/documentation/appkit/nsappearancecustomization/1533925-appearance?language=objc

reference: https://github.com/xamarin/xamarin-macios/issues/5403

Unit test added. Without it we would not have found it did not work until
it was reported again (since it would have been silently ignored).

reference: https://github.com/xamarin/xamarin-macios/issues/5408

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -431,6 +431,7 @@ namespace AppKit {
 	interface NSAppearanceCustomization {
 
 		[Mac (10,9)]
+		[NullAllowed]
 		[Export ("appearance", ArgumentSemantic.Strong)]
 		NSAppearance Appearance { get; set; }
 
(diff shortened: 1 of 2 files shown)


spouliot:
test failure is a known issue -> https://github.com/xamarin/maccore/issues/1345

PR #5436, opened 2019-01-18
[XHarness] Allow to ignore files per arch or runtime.

mandel-macaque:
Some tests migth fail on a specific arch or runtime, this commit allows
to create a ignore file per arch or runtime.

The new ignore files are:

* iOS32-ProjectName.ignore - for ios 32.
* iOS64-ProjectName.ignore - for ios 64.
* macOSFull-ProjectName.ignore - for Mac OS Full.
* macOSModern-ProjectName.ignore - for MAc OS Modern.

Fixes https://github.com/xamarin/maccore/issues/1351

### DIFF
--- a/tests/bcl-test/BCLTests/iOS32-CorlibTests.ignore
+++ b/tests/bcl-test/BCLTests/iOS32-CorlibTests.ignore
@@ -1,0 +1,1 @@
+MonoTests.System.IntPtrTest.Test64on32
--- a/tests/bcl-test/BCLTests/templates/common/IgnoreFileParser.cs
+++ b/tests/bcl-test/BCLTests/templates/common/IgnoreFileParser.cs
@@ -60,11 +60,38 @@ namespace BCLTests {
 		
 		public static async Task<IEnumerable<string>> ParseContentFilesAsync (string contentDir)
 		{
+#if XAMMAC_4_5
+			bool isFull = true;
+#else
+			bool isFull = false;
+#endif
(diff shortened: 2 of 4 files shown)


mandel-macaque:
@spouliot it is very very uncommon, to the point that this is the only test that has that issue. I agree 100% that the right way is to fix the test. But this gives us a tool to get or builds green super fast, while we wait for the mono test fix and do a bump in the Mono SDK that we download. 

The differences between Full and Modern are more, but they are not an issue at this time. 

In this case, I would get the our master builds green and add an issue in mono, then we can remove the file. I see it as a "lets get green fast, move the issue to mono and be happy until we bump."

mandel-macaque:
Aslo, once I have provided the filters per categories, it will be even less common.

spouliot:
> But this gives us a tool to get or builds green super fast, 

my point is that the existing ignore files already give us that quick-fix :)

adding more options can lead to more mistakes (e.g. if issues only occurs randomly they can be added to the incorrect file, requiring an extra commit to fix an ignore...) so it's not clear if we'll be saving any time

mandel-macaque:
@spouliot perfectly valid point, more options more chances to get it wrong. Lets hear someone else, adding the single line to the include file is a no brainer :) :)

mandel-macaque:
Ok! closing, I have added a PR with the single ignore: https://github.com/xamarin/xamarin-macios/pull/5440

PR #5439, opened 2019-01-18
[mono-2018-10] add build rules for arm64_32 watch

lewurm:
NOTE1: tested with this hacky patch for xharness: https://gist.github.com/lewurm/b4243bad2b4407bdd7f056bbdbeafd5b

NOTE2: depends on https://github.com/mono/mono/pull/12498

In terms of `mono-native*` the handling for the new watch is a bit different, due to watchOS 5.0 doesn't require compat/unified as far as I understand.

Verified by running mini regression suite. It's green 🙂


builds/Makefile
```diff
@@ -1402,31 +1459,45 @@ $(IOS_DESTDIR)$(XAMARIN_WATCHOS_SDK)/usr/lib/libmono-ilgen.a: $(WATCHOS_TARGET_L
 	$(Q) $(WATCHOS_BIN_PATH)/lipo $(WATCHOS_TARGET_LIBMONOILGEN) -create -output $@
 
 $(IOS_DESTDIR)$(XAMARIN_WATCHOS_SDK)/usr/lib/libmono-profiler-log.dylib: $(WATCHOS_TARGET_SHAREDLIBLOGPROFILER) | $(IOS_DESTDIR)$(XAMARIN_WATCHOS_SDK)/usr/lib
-	$(Q) $(WATCHOS_BIN_PATH)/bitcode_strip $(WATCHOS_TARGET_SHAREDLIBLOGPROFILER) -m -o $@
-	$(Q) $(WATCHOS_BIN_PATH)/install_name_tool -id @rpath/libmono-profiler-log.dylib -change $(SDK_DESTDIR)/ios-targetwatch-release/lib/libmonosgen-2.0.1.dylib @rpath/libmonosgen-2.0.dylib $@
+	$(Q) $(WATCHOS_BIN_PATH)/bitcode_strip $(BUILD_DESTDIR)/targetwatch/lib/libmono-profiler-log.0.dylib -m -o $(BUILD_DESTDIR)/targetwatch/tmp-lib/libmono-profiler-log.0.dylib
+	$(Q) $(WATCHOS_BIN_PATH)/bitcode_strip $(BUILD_DESTDIR)/targetwatch64_32/lib/libmono-profiler-log.0.dylib -m -o $(BUILD_DESTDIR)/targetwatch64_32/tmp-lib/libmono-profiler-log.0.dylib
```
lewurm: @rolfbjarne do you have a suggestion to make that a bit nicer?

rolfbjarne:
This looks good, but I'm wondering if we should keep the arm64_32 development on a different branch than `mono-2018-10` so that we don't add too much stuff into `mono-2018-10` (which would probably make the `mono-2018-10` PR easier to complete/merge).

@chamons / @spouliot ^?

spouliot:
IMO this work should be based on (not in) `2018-08` since we do not have any dates for `16.1` which is the earliest date where mono `2018-10` changes could be released. 

Right now if we need a quick release (even a preview) to support this new architecture we would be blocking ourselves (and our customers) on completing mono `2018-10` integration first.

luhenry:
Following Slack discussion (https://xamarinhq.slack.com/archives/C03CFKFGL/p1548259145038500) a custom `xamarin-macios` branch will be created tracking a custom `mono:2018-08-watchos` branch, avoiding the need for 2 `mono` submodules. This `mono:2018-08-watchos` branch will follow closely `2018-08` and will contain any fixes needed for watchOS.

PR #5456, opened 2019-01-22
[XHarness] Remove not needed ignore file for System.Core.

mandel-macaque:
Since we support filtering via categories, we do not longer need the
ignore file in these tests.


mandel-macaque:
build

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1345

PR #5458, opened 2019-01-22
[XHarness] Clean the System.Data ignore.

mandel-macaque:
Since we support filtering there are a number of tests that were ignore
that are part of a ignored category, therefore, they are not needed in
the file.


mandel-macaque:
Crashes are unrelated to the PR. The other issue is known: https://github.com/xamarin/maccore/issues/1345

PR #5459, opened 2019-01-22
[XHarness] Clean the System.Net.Http ignore file.

mandel-macaque:
Since we support categories, there is a test that does not longer needs
to be ignored in ios and tvos but needs to be in watchos (we do not yet
have the correct test assembly for the platform).


mandel-macaque:
build

mandel-macaque:
build

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1345

PR #5460, opened 2019-01-22
[Xharness] Remove not longer needed System.Service.Model.Web ignore file

mandel-macaque:
Since we support categories, the file is not longer needed since the
failing tests are part of an ignored category.


rolfbjarne:
build

Issue #5333, opened 2018-12-28
iOS HttpClient GetAsync hangs indefinitely


jonathanantoine:
### Steps to Reproduce

In certain case when the app is suspended, thread / http requests are no more created and hangs forever.

Steps to reproduce the issue :
1. Launch the app.
2. Wait for this log to be displayed 'Batch X created'.
3. Wait 1 second.
4. Put the app in background ( press the home button for example).
5. Wait 10 seconds and put the app in foreground.
6. Notice that no more thread / requests are made.

(I sometimes have to do it several times for the bug to happen).

### Expected Behavior
Threads and requests are created / send. 
### Actual Behavior
No more threads are created and no more http requests are send.
Example of error log : [logs.zip](https://github.com/xamarin/Xamarin.Forms/files/2714686/logs.zip)

### Environment

```
=== Visual Studio Enterprise 2017 for Mac ===

Version 7.7.2 (build 21)
Installation UUID: 37cc8c40-d076-4974-8bf8-344f707f4364
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.1.178 (master / eeaeb7e6)

	Package version: 516000221

=== Mono Framework MDK ===

Runtime:
	Mono 5.16.0.221 (2018-06/b63e5378e38) (64-bit)
	Package version: 516000221

=== NuGet ===

Version: 4.8.0.5385

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.2.0
	2.1.2
	2.1.1
	2.0.5
SDK: /usr/local/share/dotnet/sdk/2.2.101/Sdks
SDK Versions:
	2.2.101
	2.1.302
	2.1.301
	2.1.4
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.16.0/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.4
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Apple Developer Tools ===

Xcode 10.1 (14460.46)
Build 10B61

=== Xamarin.Mac ===

Xamarin.Mac not installed. Can't find /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/Version.

=== Xamarin.iOS ===

Version: 12.2.1.12 (Visual Studio Enterprise)
Hash: 65ec520f
Branch: d15-9
[...]

PR #5463, opened 2019-01-23
[Foundation] Ensure that we do not block when in the background without a background session.

mandel-macaque:
The mono threadpool gets into an unknown state when the application goes
into the background. This fix allows the task that are inflight to be
canceled when the app goes to the background allowing the application
not to hang and letting the developer retyr the request.

If a developer needs to work with the app in the background, he should
be using a background session, this fix just ensures that we are left in
a known state but does not mean that developers should use this kind of
sessions.

The MessageHandler class is just used in Mac OS X and does not have the
idea of the app going to the background, therefore the fix is not needed
in that handler.

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -51,7 +51,7 @@ using nint = System.Int32;
 using nuint = System.UInt32;
 #endif
 
-#if !MONOMAC && !WATCH
+#if !MONOMAC
 using UIKit;
 #endif
 
@@ -124,7 +124,7 @@ namespace Foundation {
 		readonly NSUrlSession session;
 		readonly Dictionary<NSUrlSessionTask, InflightData> inflightRequests;
 		readonly object inflightRequestsLock = new object ();
-#if !MONOMAC && !WATCH
+#if !MONOMAC && !MONOTOUCH_WATCH
 		readonly bool isBackgroundSession = false;
 		NSObject notificationToken;  // needed to make sure we do not hang if not using a background session
 #endif
@@ -150,7 +150,7 @@ namespace Foundation {
 			if (configuration == null)
 				throw new ArgumentNullException (nameof (configuration));
 
-#if !MONOMAC && !WATCH
+#if !MONOMAC  && !MONOTOUCH_WATCH 
 			// if the configuration has an identifier, we are dealing with a background session, 
 			// therefore, we do not have to listen to the notifications.
 			isBackgroundSession = !string.IsNullOrEmpty (configuration.Identifier);
@@ -173,13 +173,12 @@ namespace Foundation {
 			inflightRequests = new Dictionary<NSUrlSessionTask, InflightData> ();
 		}
 
-#if !MONOMAC && !WATCH
+#if !MONOMAC  && !MONOTOUCH_WATCH
 
 		void AddNotification ()
 		{
 			if (!isBackgroundSession && notificationToken == null)
-				using (var notification = new NSString ("UIApplicationWillResignActiveNotification"))
-					notificationToken = NSNotificationCenter.DefaultCenter.AddObserver (notification, BackgroundNotificationCb);
+				notificationToken = NSNotificationCenter.DefaultCenter.AddObserver (UIApplication.WillResignActiveNotification, BackgroundNotificationCb);
 		}
 
 		void RemoveNotification ()
@@ -206,7 +205,7 @@ namespace Foundation {
 		{
 			lock (inflightRequestsLock) {
 				inflightRequests.Remove (task);
-#if !MONOMAC && !WATCH
+#if !MONOMAC  && !MONOTOUCH_WATCH
 				// do we need to be notified? If we have not inflightData, we do not
 				if (inflightRequests.Count == 0)
 					RemoveNotification ();
@@ -221,7 +220,7 @@ namespace Foundation {
 
 		protected override void Dispose (bool disposing)
 		{
-#if !MONOMAC && !WATCH
+#if !MONOMAC  && !MONOTOUCH_WATCH
 			// remove the notification if present, method checks against null
 			RemoveNotification ();
 #endif
@@ -305,7 +304,7 @@ namespace Foundation {
 			});
 
 			lock (inflightRequestsLock) {
-#if !MONOMAC && !WATCH
+#if !MONOMAC  && !MONOTOUCH_WATCH
 				// Add the notification whenever needed
 				AddNotification ();
 #endif
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -51,6 +51,10 @@ using nint = System.Int32;
 using nuint = System.UInt32;
 #endif
 
+#if !MONOMAC && !WATCH
+using UIKit;
+#endif
+
 #if SYSTEM_NET_HTTP
 namespace System.Net.Http {
 #else
@@ -120,8 +124,9 @@ namespace Foundation {
 		readonly NSUrlSession session;
 		readonly Dictionary<NSUrlSessionTask, InflightData> inflightRequests;
 		readonly object inflightRequestsLock = new object ();
-#if !MONOMAC
-		readonly NSObject notificationToken;  // needed to make sure we do not hang if not using a background session
+#if !MONOMAC && !WATCH
+		readonly bool isBackgroundSession = false;
+		NSObject notificationToken;  // needed to make sure we do not hang if not using a background session
 #endif
 
 		static NSUrlSessionConfiguration CreateConfig ()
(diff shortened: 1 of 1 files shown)


mandel-macaque:
For background info about the fix, please read https://github.com/xamarin/xamarin-macios/pull/5334

mandel-macaque:
weird, this compiled locally with no issues..

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -142,6 +145,15 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
 			if (configuration == null)
 				throw new ArgumentNullException (nameof (configuration));
 
+#if !MONOMAC
+			// if the configuration has an identifier, we are dealing with a background session, 
+			// therefore, we do not have to listen to the notifications.
+			if (string.IsNullOrEmpty (configuration.Identifier)) {
+				using (var notification = new NSString ("UIApplicationWillResignActiveNotification"))
+					notificationToken = NSNotificationCenter.DefaultCenter.AddObserver (notification, BackgoundNotificationCb);
```
rolfbjarne: The observer is never removed, which means the BackgroundNotificationOn delegate contained in the observer will keep a permanent reference to the NSUrlSessionHandler instance, and it will never be collected by the garbage collector.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -159,6 +171,17 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
 			inflightRequests = new Dictionary<NSUrlSessionTask, InflightData> ();
 		}
 
+#if !MONOMAC
+		void BackgoundNotificationCb (NSNotification obj)
```
rolfbjarne: Typo: `BackgoundNotificationCb` => `BackgroundNotificationCb` (missing r)

src/Foundation/NSUrlSessionHandler.cs
spouliot: that string constant is already a field/notification inside `UIApplication`, `WillResignActiveNotification`
any reason why you can't use it ?

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -159,6 +171,17 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
...
+		{
+			// we do not need to call the lock, we call cancel on the source, that will trigger all the needed code to 
+			// clean the resources and such
+			foreach (var pair in inflightRequests) {
```
spouliot: You can iterate the values, should be more efficiently than asking the `.Value` property for each `pair`

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: Yes, we are not compiling the class at a point in time in which we have access to it. Please take a look at the first commit build error in the bots :/

spouliot: That error looks like it was because it's compiling for watchOS where `UIApplication` is not available. 

https://developer.apple.com/documentation/uikit/uiapplicationwillresignactivenotification?language=objc does not mention watchOS has this constant.

Using the name manually you're _working_ around that - but it's unlikely the code works one watchOS (the code just make it looks like it should)

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -197,6 +197,9 @@ void RemoveInflightData (NSUrlSessionTask task, bool cancel = true)
 
 		protected override void Dispose (bool disposing)
 		{
+#if !MONOMAC
+			NSNotificationCenter.DefaultCenter.RemoveObserver (notificationToken);
```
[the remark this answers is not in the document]
rolfbjarne: Yes, I think it's worth it to make sure we introduce new leaks. Someone will eventually complain about it otherwise.

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: I made that change locally and got the following compilation error:

`
System.Net.Http/HttpClientHandler.platformnotsupported.cs(129,63): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
HttpClientEx.cs(83,8): warning CS0219: The variable 'webExceptionStatus' is assigned but its value is never used
/Users/mandel/Xamarin/xamarin-macios/master/xamarin-macios/src/Foundation/NSUrlSessionHandler.cs(181,87): error CS0117: 'UIApplication' does not contain a definition for 'WillResignActiveNotification'
make[3]: *** [../../class/lib/monotouch_watch/reference/System.Net.Http.dll] Error 1
`

Nevertheless, I'll add the !WATCH to show that it is not done on WatchOS.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -171,7 +173,21 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
...
+		{
+			if (!isBackgroundSession && notificationToken == null)
+				using (var notification = new NSString ("UIApplicationWillResignActiveNotification"))
+					notificationToken = NSNotificationCenter.DefaultCenter.AddObserver (notification, BackgroundNotificationCb);
+		}
+
+		void RemoveNotification ()
+		{
+			if (notificationToken != null)
+				NSNotificationCenter.DefaultCenter.RemoveObserver (notificationToken);
```
rolfbjarne: You should set `notificationToken` to null so that it can be re-added if more requests come in.

mandel-macaque: God.. what a stupid mistake! thx!

src/Foundation/NSUrlSessionHandler.cs
spouliot: the error just means it's not available on watchOS - your code avoid that (at compile time) but it won't change runtime (and you're also exposing a constant string that is not part of the public API on the platform which is, at best, risky).

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -159,12 +173,45 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
 			inflightRequests = new Dictionary<NSUrlSessionTask, InflightData> ();
 		}
 
+#if !MONOMAC && !WATCH
+
+		void AddNotification ()
+		{
+			if (!isBackgroundSession && notificationToken == null)
+				using (var notification = new NSString ("UIApplicationWillResignActiveNotification"))
```
spouliot: same comments as earlier:

* why not use `UIApplication.WillResignActiveNotification` instead of your own constant ? Minimally it avoid the memory allocation and any chance of a typo. Also if it ever gets obsoleted we'll not get a compiler warning to update this code.

* why not follow the advice on the field ? `"Use UIApplication.Notifications.ObserveWillResignActive helper method instead."` ?

mandel-macaque:
@GouriKumari The following steps can be used to reproduce the issue and the fix:

1. Get the sample application from https://github.com/xamarin/xamarin-macios/issues/5333
2. Launch the application and wait a few seconds to see that requests are performed.
3. Set the application to the background and wait LONGER than 10 seconds.
4. Get back to the application and ensure that requests continue to happen.

We should check setting the app to the background by:
1. Switching an other app to the foreground.
2. Go to the home screen.

The testes should see that the requests, when the app goes to the background fail, and that when we get back to the app, it is not stuck and requests continue happening.

mandel-macaque:
As soon as we have a package I'll post the link here for QA.

mandel-macaque:
You can ignore the VSTS issues :)

p-lad:
Hello @mandel-macaque @GouriKumari ,

Following is the status of the pr : - 
As mentioned on slack I have reproduced issue on Stable XI build which are as follows:-
Xamarin.Mac Version: 5.2.1.13 (Visual Studio Enterprise)
Xamarin.iOS Version: 12.2.1.13 (Visual Studio Enterprise)

Application output logs:-
https://gist.github.com/p-lad/50ee0c2329d11ce1747e9d0600fe603f

Screen cast link:-
http://recordit.co/N1Yz5Vq2Fc

===============================================================
Following is the status on Fixed build mentioned in pr:-
Build taken : - 
XI :- 
https://bosstoragemirror.blob.core.windows.net/wrench/jenkins/handler-cancel-background/ec0c6e740d0eaffd838c3f52dd943a2811dceb4e/1/package/xamarin.mac-5.7.0.128.pkg

XM:-
https://bosstoragemirror.blob.core.windows.net/wrench/jenkins/handler-cancel-background/ec0c6e740d0eaffd838c3f52dd943a2811dceb4e/1/package/xamarin.ios-12.7.0.128.pkg

Build info:
Xamarin.Mac Version: 5.7.0.128 (Visual Studio Enterprise)
Xamarin.iOS Version: 12.7.0.128 (Visual Studio Enterprise)

Application output logs:-
https://gist.github.com/p-lad/7aabb998e19f6d91e4b812a1485011c7

Screen cast link:-
http://recordit.co/Z4twf2VqDf

As discussed with @mandel-macaque on slack,  app went to the background and got back, all threads started normally with the logs stating that new ones were created.

Hence from QA side it is QA-Approved.

Let us  know if anything is required from our side.

Thanks.

Issue #5471, opened 2019-01-24
[tests] Introspection iOS Unified 64-bits - simulator failing on 5 selector not found for UIKit.UIPreviewInteraction


VincentDondain:
### Steps to Reproduce

1. Run iOS introspection tests for iOS Unified 64-bits 

### Expected Behavior

Tests shouldn't fail with "selector not found for UIKit.UIPreviewInteraction"

### Actual Behavior

5 selector not found for UIKit.UIPreviewInteraction

```
[FAIL] Selector not found for UIKit.UIPreviewInteraction : cancelInteraction 
[FAIL] Selector not found for UIKit.UIPreviewInteraction : locationInCoordinateSpace: 
[FAIL] Selector not found for UIKit.UIPreviewInteraction : delegate 
[FAIL] Selector not found for UIKit.UIPreviewInteraction : setDelegate: 
[FAIL] Selector not found for UIKit.UIPreviewInteraction : view 
```

### Environment

Latest builds from https://github.com/xamarin/xamarin-macios/tree/xcode10.2. After https://github.com/xamarin/xamarin-macios/pull/5470 is merged.

spouliot:
That was fixed (as in ignored in tests) in both `xcode10.2` and `master` branches - 
https://github.com/xamarin/xamarin-macios/pull/5573

PR #5473, opened 2019-01-25
[WatchKit] Update bindings to Xcode 10.2 Beta 1

dalexsoto:
diff: https://github.com/xamarin/xamarin-macios/wiki/WatchKit-watchOS-xcode10.2-beta1

And yup [Watch (5,1)] xtro seems to agree

### DIFF
--- a/src/watchkit.cs
+++ b/src/watchkit.cs
@@ -648,6 +648,14 @@ namespace WatchKit {
 		[Watch (3,0)][NoiOS]
 		[Export ("performSegueForRow:")]
 		void PerformSegue (nint row);
+
+		[Watch (5,1), NoiOS]
+		[Export ("curvesAtTop")]
+		bool CurvesAtTop { get; set; }
+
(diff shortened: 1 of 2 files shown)


dalexsoto:
Expected test failure: https://github.com/xamarin/xamarin-macios/issues/5471

src/watchkit.cs
```diff
@@ -648,6 +648,14 @@ interface WKInterfaceTable {
 		[Watch (3,0)][NoiOS]
 		[Export ("performSegueForRow:")]
 		void PerformSegue (nint row);
+
+		[Watch (5,1), NoiOS]
```
spouliot: @dalexsoto `xtro` is based on sharpie/headers so it will always match
it does not mean it's in 5.1 - you have to run intro on a 5.1 simulator (even better device) to be sure it was present (and not something postponed)

dalexsoto: I am sorry I meant intro in the description

PR #5497, opened 2019-01-28
[healthkit] Update for Xcode 10.2 beta 1

VincentDondain:


tests/monotouch-test/HealthKit/CategoryTypeIdentifierTest.cs
```diff
@@ -42,6 +42,12 @@ public void EnumValues_22351 ()
 					if (!TestRuntime.CheckXcodeVersion (8, 0))
 						continue;
 					break;
+				case HKCategoryTypeIdentifier.HighHeartRateEvent:
+				case HKCategoryTypeIdentifier.LowHeartRateEvent:
+				case HKCategoryTypeIdentifier.IrregularHeartRhythmEvent:
+					if (!TestRuntime.CheckXcodeVersion (10, 2))
```
spouliot: that's not ok - you should be checking only on 64 bits (not the Xcode version)

VincentDondain: @spouliot isn't Xcode 10.2 an _implicit_ 64 bit check given that iOS is 64 bit only there? Also those APIs are 10.2+ and the previous checks do the same Xcode asserts so I'm slightly confused.
+ the test is fixed (:

spouliot: yeah, it's fine - it's not the test I thought it was :)

PR #5500, opened 2019-01-28
[Metal] Update bindings to Xcode 10.2 Beta 1

dalexsoto:


dalexsoto:
Yep likely coming in a new beta or maybe used internally and it leaked but still it has availability info so who knows the way of Apple but Apple

PR #5513, opened 2019-01-29
[coremedia] Update for xcode 10.2 beta 1

spouliot:
Most additions are block-based alternatives to callback API. However,
the way we bind them, it produce identical API so they are ignored.

### DIFF
--- a/src/CoreMedia/CMSampleBuffer.cs
+++ b/src/CoreMedia/CMSampleBuffer.cs
@@ -50,11 +50,6 @@ namespace CoreMedia {
 		Invalidated						= -12744,
 	}
 
-	[iOS (12,2)]
-	[TV (12,2)]
-	[Mac (10,14,4, onlyOn64: true)]
-	public delegate CMSampleBufferError CMSampleBufferMakeDataReadyHandler (CMSampleBuffer sbuf);
-
 	public class CMSampleBuffer : ICMAttachmentBearer 
 #if !COREBUILD
 	, IDisposable
--- a/src/CoreMedia/CMSampleBuffer.cs
+++ b/src/CoreMedia/CMSampleBuffer.cs
@@ -50,6 +50,11 @@ namespace CoreMedia {
 		Invalidated						= -12744,
 	}
 
+	[iOS (12,2)]
+	[TV (12,2)]
+	[Mac (10,14,4, onlyOn64: true)]
+	public delegate CMSampleBufferError CMSampleBufferMakeDataReadyHandler (CMSampleBuffer sbuf);
+
 	public class CMSampleBuffer : ICMAttachmentBearer 
 #if !COREBUILD
 	, IDisposable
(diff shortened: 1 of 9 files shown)


src/CoreMedia/CMSampleBuffer.cs
```diff
@@ -834,6 +841,70 @@ public CMSampleBufferError CopyPCMDataIntoAudioBufferList (int frameOffset, int
...
+
+		static internal class DCMSampleBufferMakeDataReadyHandlerTrampoline {
+			static internal readonly DCMSampleBufferMakeDataReadyHandler Handler = Invoke;
+
+			[MonoPInvokeCallback (typeof (DCMSampleBufferMakeDataReadyHandler))]
+			static unsafe CMSampleBufferError Invoke (IntPtr block, IntPtr sbuf)
+			{
+				var descriptor = (BlockLiteral *) block;
+				var del = (CMSampleBufferMakeDataReadyHandler) descriptor->Target;
+				return del (new CMSampleBuffer (sbuf));
```
dalexsoto: I wonder if we should do something like this instead

```csharp
using (var b = new CMSampleBuffer (sbuf)) {
	return del (b);
}
```

You can eventually run out of sample buffers because the system tries to be clever and reuse them and our GC is not fast enough to Dispose them so it needs a little help by disposing it right after it is not used anymore. This has got some of our users in the past.

src/CoreMedia/CMSampleBuffer.cs
```diff
@@ -834,6 +841,72 @@ public CMSampleBufferError CopyPCMDataIntoAudioBufferList (int frameOffset, int
...
+			}
+		}
+
+		[iOS (12,2)]
+		[TV (12,2)]
+		[Mac (10,14,4, onlyOn64: true)]
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public static CMSampleBuffer CreateReadyWithImageBuffer (CVImageBuffer imageBuffer,
+			bool dataReady, CMSampleBufferMakeDataReadyHandler makeDataReadyHandler,
+			CMFormatDescription formatDescription, CMSampleTimingInfo [] sampleTiming, out CMSampleBufferError error)
```
rolfbjarne: `sampleTiming` must contain (at least) one element, so that should be validated. In fact, the input should probably be just `CMSampleTimingInfo`, not an array of them.

src/CoreMedia/CMSampleBuffer.cs
```diff
@@ -834,6 +841,72 @@ public CMSampleBufferError CopyPCMDataIntoAudioBufferList (int frameOffset, int
 
 			return new CMSampleBuffer (buffer, true);
 		}
+
+		[iOS (12,2)]
+		[TV (12,2)]
+		[Mac (10,14,4, onlyOn64: true)]
+		[DllImport (Constants.CoreMediaLibrary)]
+		static unsafe extern /* OSStatus */ CMSampleBufferError CMSampleBufferCreateForImageBufferWithMakeDataReadyHandler (
```
rolfbjarne: `CMSampleBufferCreateForImageBufferWithMakeDataReadyHandler` is the block based version of `CMSampleBufferCreateForImageBuffer`, which we've already bound: https://github.com/xamarin/xamarin-macios/blob/8e321095063c2b5e3370b39c78bdbc373e7be5f3/src/CoreMedia/CMSampleBuffer.cs#L237-L268

tests/xtro-sharpie/common-CoreMedia.ignore
```diff
@@ -355,3 +371,9 @@
 !missing-pinvoke! CMVideoFormatDescriptionCopyAsBigEndianImageDescriptionBlockBuffer is not bound
 !missing-pinvoke! CMVideoFormatDescriptionCreateFromBigEndianImageDescriptionBlockBuffer is not bound
 !missing-pinvoke! CMVideoFormatDescriptionCreateFromBigEndianImageDescriptionData is not bound
+
+## block based version of CMBufferQueueCreate (which use callbacks and is already bound)
+!missing-pinvoke! CMBufferQueueCreateWithHandlers is not bound
+
+## block based version of CMSampleBufferCreateForImageBuffer (which use callbacks and is already bound)
```
rolfbjarne:
```suggestion
## block based version of CMAudioSampleBufferCreateWithPacketDescriptions (which use callbacks and is already bound)
```

src/CoreMedia/CMSampleBuffer.cs
spouliot: I think it's an array - but I've read too many variations... I'll double check :)
good point on empty - won't be better than `null`

src/CoreMedia/CMSampleBuffer.cs
spouliot: good catch, I missed it :(
better remove it

PR #5517, opened 2019-01-30
[NSString] Add overloads which are optimized for substring handling

### LINKED ISSUE #5134: [API] NSString constructors for substrings without allocating a new managed string
Currently, `NSString` APIs allow only to pass a managed string.

This forces APIs which would typically just use a substring be forced to allocate a string for the substring itself.

```
var myString = "asdf";
var myNSString = new NSString (myString.Subtring (0, 2));
```

Proposed APIs:
```
public NSString (string str, int start, int length) {}
```

The implementation would then just marshal part of the string, avoiding an intermediate string allocation.

NetStandard 2.0 thoughts:
Maybe the APIs could take a `ReadOnlySpan<char>` which would automatically handle slicing and offset calculation.

Therzok:
Fixes #5134

@rolfbjarne any guidance on where new tests would go?

### DIFF
--- a/src/Foundation/NSString.cs
+++ b/src/Foundation/NSString.cs
@@ -92,14 +92,14 @@ namespace Foundation {
 		{
 		}
 
-		static IntPtr CreateWithCharacters (IntPtr handle, string str, bool autorelease = false)
+		static IntPtr CreateWithCharacters (IntPtr handle, string str, int offset, int length, bool autorelease = false)
 		{
 			unsafe {
-				fixed (char *ptrFirstChar = str) {
+				fixed (char *ptrFirstChar = str + offset) {
 	#if MONOMAC
-					handle = Messaging.IntPtr_objc_msgSend_IntPtr_IntPtr (handle, selInitWithCharactersLengthHandle, (IntPtr) ptrFirstChar, (IntPtr) str.Length);
+					handle = Messaging.IntPtr_objc_msgSend_IntPtr_IntPtr (handle, selInitWithCharactersLengthHandle, (IntPtr) ptrFirstChar, (IntPtr) length);
 	#else
-					handle = Messaging.IntPtr_objc_msgSend_IntPtr_IntPtr (handle, Selector.GetHandle (selInitWithCharactersLength), (IntPtr) ptrFirstChar, (IntPtr) str.Length);
+					handle = Messaging.IntPtr_objc_msgSend_IntPtr_IntPtr (handle, Selector.GetHandle (selInitWithCharactersLength), (IntPtr) ptrFirstChar, (IntPtr) length);
 	#endif
 
 					if (autorelease)
@@ -117,8 +117,24 @@ namespace Foundation {
 
 		public static IntPtr CreateNative (string str, bool autorelease)
 		{
+			return CreateNative (str, 0, str.Length, autorelease);
+		}
+
+		public static IntPtr CreateNative (string str, int start, int length)
+		{
+			return CreateNative (str, start, length, false);
+		}
+
+		public static IntPtr CreateNative (string str, int start, int length, bool autorelease)
+		{
 			if (str == null)
 				return IntPtr.Zero;
+
+			if (start < 0 || start > str.Length)
+				throw new ArgumentOutOfRangeException(nameof(start));
+
+			if (length < 0 || start > str.Length - length)
+				throw new ArgumentOutOfRangeException(nameof(length));
 
 #if MONOMAC
 			var handle = Messaging.IntPtr_objc_msgSend (class_ptr, Selector.AllocHandle);
@@ -126,7 +142,7 @@ namespace Foundation {
 			var handle = Messaging.IntPtr_objc_msgSend (class_ptr, Selector.GetHandle (Selector.Alloc));
 #endif
 
-			return CreateWithCharacters (handle, str, autorelease);
+			return CreateWithCharacters (handle, str, start, length, autorelease);
 		}
 
 		public static void ReleaseNative (IntPtr handle)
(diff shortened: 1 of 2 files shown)


rolfbjarne:
@Therzok here: https://github.com/xamarin/xamarin-macios/blob/742f0df45971ed8837bd65c76e5ad7ada572bd36/tests/monotouch-test/Foundation/NSStringTest.cs#L17

src/Foundation/NSString.cs
```diff
@@ -116,17 +116,33 @@ public static IntPtr CreateNative (string str)
 		}
 
 		public static IntPtr CreateNative (string str, bool autorelease)
+		{
+			return CreateNative (str, 0, str.Length, autorelease);
```
spouliot: `str` can be `null`, `str.Length` -> NRE

src/Foundation/NSString.cs
```diff
@@ -116,17 +116,33 @@ public static IntPtr CreateNative (string str)
...
+		}
+
+		public static IntPtr CreateNative (string str, int start, int length)
```
rolfbjarne: We try to avoid abbreviations for parameter names in public API, so `string value` instead of `string str` (yes, I know the existing overload is using `str`, but that's no reason we can't make new overloads better).

src/Foundation/NSString.cs
```diff
@@ -116,17 +116,33 @@ public static IntPtr CreateNative (string str)
...
+		{
+			return CreateNative (str, start, length, false);
+		}
+
+		public static IntPtr CreateNative (string str, int start, int length, bool autorelease)
```
rolfbjarne: Same: `str` -> `value`.

src/Foundation/NSString.cs
```diff
@@ -116,17 +116,33 @@ public static IntPtr CreateNative (string str)
...
 		{
 			if (str == null)
 				return IntPtr.Zero;
 
+			if (start < 0 || start > str.Length)
+				throw new ArgumentOutOfRangeException(nameof(start));
```
rolfbjarne: Style: 
```suggestion
				throw new ArgumentOutOfRangeException (nameof (start));
```

src/Foundation/NSString.cs
```diff
@@ -116,17 +116,33 @@ public static IntPtr CreateNative (string str)
...
+
+			if (length < 0 || start > str.Length - length)
+				throw new ArgumentOutOfRangeException(nameof(length));
```
rolfbjarne: Style: 
```suggestion
				throw new ArgumentOutOfRangeException (nameof (length));
```

src/Foundation/NSString.cs
```diff
@@ -138,7 +154,20 @@ public static void ReleaseNative (IntPtr handle)
 			if (str == null)
 				throw new ArgumentNullException ("str");
 
-			Handle = CreateWithCharacters (Handle, str);
+			Handle = CreateWithCharacters (Handle, str, 0, str.Length);
+		}
+
+		public NSString (string str, int start, int length) {
+			if (str == null)
+				throw new ArgumentNullException ("str");
```
rolfbjarne:
```suggestion
				throw new ArgumentNullException (nameof (str));
```

src/Foundation/NSString.cs
```diff
@@ -138,7 +154,20 @@ public static void ReleaseNative (IntPtr handle)
...
+
+			if (start < 0 || start > str.Length)
+				throw new ArgumentOutOfRangeException(nameof(start));
```
rolfbjarne:
```suggestion
				throw new ArgumentOutOfRangeException (nameof (start));
```

src/Foundation/NSString.cs
```diff
@@ -138,7 +154,20 @@ public static void ReleaseNative (IntPtr handle)
...
+
+			if (length < 0 || start > str.Length - length)
+				throw new ArgumentOutOfRangeException(nameof(length));
```
rolfbjarne:
```suggestion
				throw new ArgumentOutOfRangeException (nameof (length));
```

tests/monotouch-test/Foundation/NSStringTest.cs
```diff
@@ -35,6 +35,37 @@ public void LocalizedFormatTest ()
...
+		public void TestNSStringSubstrings (string input, int start, int length)
+		{
+			var str = new NSString (input.Substring (start, length));
+			var substring = new NSString (input, start, length);
+			var substringHandle = NSString.CreateNative (input, start, length);
+
+			Assert.AreEqual (str, substring);
+
+			substring = NSString.FromHandle (substringHandle);
+			Assert.AreEqual (str, substring);
```
rolfbjarne: `substringHandle` must be freed:

```csharp
NSString.ReleaseNative (substringHandle);
```

src/Foundation/NSString.cs
Therzok: Fixed

tests/monotouch-test/Foundation/NSStringTest.cs
```diff
@@ -35,6 +35,40 @@ public void LocalizedFormatTest ()
...
+		{
+			var str = new NSString (input.Substring (start, length));
+			var substring = new NSString (input, start, length);
+			var substringHandle = NSString.CreateNative (input, start, length);
+			try {
+				Assert.AreEqual (str, substring);
+
+				substring = NSString.FromHandle (substringHandle);
+				Assert.AreEqual (str, substring);
+			finally {
```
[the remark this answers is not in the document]
Therzok: Fixed

Therzok:
@spouliot I tried tackling this problem for both NSAttributedString and NSMutableAttributedString, but it seems like it is not doable without changes to the generator.

Luckily, NSString was manually bound, so it did not need a higher level change.

The generator should probably be made to know that `NSString.CreateNative` now has a substring overload, and new constructors should be offered to handle this scenario.

I think this could evolve into general purpose solution for everything else, with maybe some definition like:

```csharp
partial interface NSAttributedString
{
		[Export ("initWithString:")]
		IntPtr Constructor ([WithSubstring] string str);
}
```

And `[WithSubstring]` will make the code generator generate overloads with start and offset and proper argument checks.

`WithSubstringAttribute` would most likely have to be applied _just_ to string parameters via `[System.AttributeUsage(System.AttributeTargets.Parameter)]`

I'm not sure I have enough generator knowledge to tackle that. If this is the way to go forward, I'll open another issue split up from #5134 with the proposal.

spouliot:
@Therzok interesting but I'm not sure if it's the right/best approach, e.g. it's still possible to bind stuff manually (and it's often best to do so for a while before adding new features to the generator).

> If this is the way to go forward, I'll open another issue split up from #5134 with the proposal.

You read my mind :) 

My earlier comment was more about the PR description, mentioning fixing an issue that covers `NS[Mutable][Attributed]String`.

It's best to update the description / split the issue and go ahead with what you have (for `NSString`) while continuing the work (and get feedback) on the other types.

Therzok:
Filed https://github.com/xamarin/xamarin-macios/issues/5523 and edited 5134

Therzok:
I'll move the discussion into the new issue.

PR #5540, opened 2019-02-04
[xharness] Rework BCL test importation to not bail out completely if something went wrong.

rolfbjarne:
Rework BCL test importation to generate projects that won't compile instead of
giving up completely in case of generation failure.

This means that xharness can still be used for non-BCL tests even if the
generation of the BCL tests fail.

### DIFF
--- a/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectDefinition.cs
+++ b/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectDefinition.cs
@@ -37,9 +37,11 @@ namespace BCLTestImporter {
 			TestAssemblies = assemblies;
 		}
 		
-		static IEnumerable<string> GetAssemblyReferences (string assemblyPath) {
+		static (string FailureMessage, IEnumerable<string> References) GetAssemblyReferences (string assemblyPath) {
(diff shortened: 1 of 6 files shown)


tools/bcl-test-importer/BCLTestImporter/BCLTestProjectDefinition.cs
```diff
@@ -65,24 +67,30 @@ public bool Validate ()
...
+		(string FailureMessage, IEnumerable<string> References) GetProjectAssemblyReferences (string rootPath, Platform platform, bool wasDownloaded)
 		{
 			var set = new HashSet<string> ();
+			string failureMessage = null;
 			foreach (var definition in TestAssemblies) {
-				foreach (var reference in GetAssemblyReferences (definition.GetPath (rootPath, platform, wasDownloaded))) {
-					set.Add (reference);
+				var references = GetAssemblyReferences (definition.GetPath (rootPath, platform, wasDownloaded));
+				if (references.FailureMessage != null) {
+					failureMessage = references.FailureMessage;
```
spouliot: you only get the latest error right ?
concat ?

Issue #4763, opened 2018-09-06
[RFC] Improve handling of filter clauses with LLVM/Bitcode


rolfbjarne:
Mono/LLVM doesn't support filter clauses when compiling to bitcode [1] (and the runtime team has said it won't be fixed for quite a while, if ever).

See also #[58209 LLVM/bitcode can't handle filter clauses](https://bugzilla.xamarin.com/show_bug.cgi?id=58209).

Current behavior:

* Everything works fine in Debug mode, since we're only compiling to bitcode in Release mode.
* In Release mode, methods containing filter clauses will either cause the app to assert or throw an exception when called at runtime.

Suggestions:

* [ ] Document this limitation here: https://docs.microsoft.com/en-us/xamarin/ios/internals/limitations. We might also want to list any BCL assemblies with methods with filter clauses (for instance System.Data has many such methods, see #[59987 System.Data is close to unusable on watchOS](https://bugzilla.xamarin.com/show_bug.cgi?id=59987)). A list of the methods can be found [here](https://github.com/xamarin/xamarin-macios/blob/510bb0b1cd19e12023a9b8c8ab8e62c1ee1df1ed/tests/mtouch/SdkTest.cs#L231-L316).
* [x] Detect filter clauses in mtouch, and produce a warning. We can also replace the entire contents of the method with an exception. This would make the behavior identical for both Debug & Release, as well as avoiding any potential mono asserts. Some time ago I created a POC for this, implemented as a linker step: https://github.com/rolfbjarne/xamarin-macios/commit/f1d46b4b1c94fcea92718d1a4d1321f371ed2482 (see also #[58209#c1](https://bugzilla.xamarin.com/show_bug.cgi?id=58209#c1)). [PR #5551]
* [x] Add tests so that we're notified if new BCL API tries to use filter clauses. It would be bad if something vital in mscorlib.dll suddenly started failing because of this... [PR #5295]

[1] For watchOS only, not tvOS, since tvOS is using "fake" bitcode, where we compile to normal machine code, and then embed that as a special "assembler" section in the bitcode.

spouliot:
Both documentation (c.c. @bradumbaugh) and canary tests (so we know if the situation change) should be done asap.

The System.Data case was discussed a few months back and IIRC @marek-safar said the BCL could be changed since (at least in this case) it did not offer any critical feature where used.

spouliot:
[...]

PR #5551, opened 2019-02-05
[linker] Remove non-bitcode compatible code, and show a warning.

rolfbjarne:
Remove code not currently compatible with bitcode and replace it with an exception instead (otherwise we may assert at runtime).

Also show a warning when we detect this.

This is quite helpful when looking at watch device test runs to filter out failures we already know about, since this exception message is much more specific than the one mono throws ("Attempting to JIT compile method ...").

This fixes point #2 in #4763.

### DIFF
--- a/docs/website/mtouch-errors.md
+++ b/docs/website/mtouch-errors.md
@@ -1787,6 +1787,12 @@ Mixed-mode assemblies can not be processed by the linker.
 
 See https://msdn.microsoft.com/en-us/library/x0w2664k.aspx for more information on mixed-mode assemblies.
 
+### MT2105: The method {method} contains a '{handlerType}' exception clause, which is currently not supported when compiling for bitcode. This method will throw an exception if called.
+
+Currently Xamarin.iOS does not support the 'filter' exception clauses when
+compiling to bitcode. Any methods containing such code will throw a
+NotSupportedException exception.
+
 ## MT3xxx: AOT error messages
 
 <!--
--- a/docs/website/mtouch-errors.md
+++ b/docs/website/mtouch-errors.md
@@ -1787,6 +1787,8 @@ Mixed-mode assemblies can not be processed by the linker.
 
 See https://msdn.microsoft.com/en-us/library/x0w2664k.aspx for more information on mixed-mode assemblies.
 
+<a name="MT2105" />
+
 ### MT2105: The method {method} contains a '{handlerType}' exception clause, which is currently not supported when compiling for bitcode. This method will throw an exception if called.
 
 Currently Xamarin.iOS does not support the 'filter' exception clauses when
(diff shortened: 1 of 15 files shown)


docs/website/mtouch-errors.md
```diff
@@ -1787,6 +1787,12 @@ Mixed-mode assemblies can not be processed by the linker.
 
 See https://msdn.microsoft.com/en-us/library/x0w2664k.aspx for more information on mixed-mode assemblies.
 
+### MT2105: The method {method} contains a '{handlerType}' exception clause, which is currently not supported when compiling for bitcode. This method will throw an exception if called.
+
+Currently Xamarin.iOS does not support the 'filter' exception clauses when
+compiling to bitcode. Any methods containing such code will throw a
+NotSupportedException exception.
```
spouliot: The description is not clear **when** the exception is thrown.

Is the current behaviour to assert when executing the method ? or 
only when an exception occurs ? or
only when an exception with a filter occurs ?

docs/website/optimizations.md
```diff
@@ -604,3 +604,47 @@ It is always enabled by default (as long as the static registrar is enabled).
...
+	}
+}
+```
+
+into the following:
+
+```csharp
+void Method ()
+{
+	throw new NotSupportedException ("This method contains IL not supported when compiled to bitcode.");
```
spouliot: That example should be copied in the main documentation (or be linked from).

```
	try {
			ThrowsAnExceptionOnlyOnTuesday ();
		} catch (Exception e) {
			throw new NotSupportedException ("This method contains IL not supported when compiled to bitcode.");
		}
	}
```

since it's not clear (from the warning) which of the above code is generated as a replacement.

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
```diff
@@ -0,0 +1,108 @@
...
+using Mono.Linker.Steps;
+
+using Xamarin.Bundler;
+using Xamarin.Linker;
+using Mono.Linker;
+using System;
+
+namespace MonoTouch.Tuner {
+	
+	public class RemoveBitcodeIncompatibleCodeStep : BaseStep {
```
spouliot: maybe a `SubStep` would be better (less code, since it shares the assembly/types/methods iterations) ?

docs/website/optimizations.md
rolfbjarne: By "main documentation" you mean mtouch-errors.md?

spouliot: yes, that's wait double-clicking on the error/warning will give to developers

docs/website/mtouch-errors.md
rolfbjarne: The current behavior is to throw an exception when the method is executed (LLVM completely fails for the method if it contains a filter clause, which means that there's no native code for the method at all). I've seen asserts too, but I'm not sure when it asserts and when an exception is thrown.

I've updated the text to be more descriptive.

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
rolfbjarne: I've changed it to use a SubStep.

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
```diff
@@ -0,0 +1,113 @@
...
+using Xamarin.Bundler;
+using Xamarin.Linker;
+using Mono.Linker;
+using System;
+
+using Xamarin.Tuner;
+using Xamarin.Linker;
+
+namespace MonoTouch.Tuner {
+	public class RemoveBitcodeIncompatibleCodeStep : BaseSubStep {
```
spouliot: oops, I guess I should have fully completed my train of thoughts... :)
`ExceptionalSubStep` makes support a lot easier if an exception is thrown during processing (which was the main drawback of substeps), otherwise there's no way to know (from logs) what caused the issue

PR #5557, opened 2019-02-06
[Foundation] Remove possible race condition with the NSUrlSessionHandler. Fixes #5511

### LINKED ISSUE #5511: Token cancellation bug in NSUrlSession http handler in Xamarin.iOS
### Steps to Reproduce
1. Get sample project from: https://github.com/jsauve/PossibleNSUrlSessionHandlerBug
3. Run the app in the iOS simulator.
4. Click "Post data" button in the app.
5. Watch the output window, and click "Cancel" button in the app just as new threads start spinning up when `SendAsync()` is about to complete, but has not quite yet completed (line 87, MainPage.xaml.cs). It's tricky to get it timed just right, but is possible with persistence. (You may need to restart the app if you don't get it on the first try, because the response seems to be cached after the first try, making the behavior more difficult to catch. You could also potentially insert an awaited `Task.Delay()` right before `SendAsync()` to give yourself more time to prepare to click Cancel.)

Screen capture video included for clarity: https://www.youtube.com/watch?v=TYLUjhxgG4Y

The managed http handler does not appear to have this issue; only the NSUrlSession handler.

### Expected Behavior
The awaited Task gets cancelled (via the token), regardless of whether or not NSUrlSession or the managed handler are selected in the iOS project.

### Actual Behavior
When using the NSUrlSession http handler, the awaited `SendAsync()` Task will sometimes not complete, leaving the app hanging. It seems that if the token is cancelled at just the right moment (possibly after the native network call has completed, but the data has not yet been returned to the managed layer), that the cancellation will leave HttpClient's `SendAsync()` method hanging indefinitely.

### Environment
https://gist.github.com/jsauve/cb5fcbddcdc7e3f1508e5308eca37497

### Build Logs
Irrelevant

### Example Project
https://github.com/jsauve/PossibleNSUrlSessionHandlerBug

mandel-macaque:
The SendAsync method in the NSUrlSessionHandler is not handling the
cancellation token correctly. Moving the Register to nearly the end of
the method (just before returning the task) ensures that if we got a
already cancelled token, we execute in a sync manner the cleanup.

Using the IsCancellationRequested in this method is less optimal
because we would have to check several times:

1. When SencAsync receives the token and before it creates the inflight
data.
2. When we are going to resume the data. Since we might have created the
inflight data and at that point an other thread cancelled the request.

Fixes: https://github.com/xamarin/xamarin-macios/issues/5511

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -249,11 +249,6 @@ namespace Foundation {
 
 			var tcs = new TaskCompletionSource<HttpResponseMessage> ();
 
-			cancellationToken.Register (() => {
-				RemoveInflightData (dataTask);
-				tcs.TrySetCanceled ();
-			});
-
 			lock (inflightRequestsLock)
 				inflightRequests.Add (dataTask, new InflightData {
 					RequestUrl = request.RequestUri.AbsoluteUri,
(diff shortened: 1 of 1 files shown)


mandel-macaque:
This PR provides the similar behaviour to PR https://github.com/xamarin/xamarin-macios/pull/5334 as well as it ensures that if the token is cancelled at some other point in time while executing SendAsync it is respected. 

I believe this is a nicer fix.

PR #5567, opened 2019-02-07
[generator] Code simplification.

rolfbjarne:
No need for an if statement if the code is equal in both the true and false path.


rolfbjarne:
build

PR #5569, opened 2019-02-07
Add a Runtime.IsARM64CallingConvention property.

rolfbjarne:
Determining whether we should use the ARM64 calling convention in P/Invokes
gets more complicated with ARM64_32 (which for our purposes is a 32-bit
architecture).

So add a property on the Runtime class to avoid code duplication, and teach
the linker to optimize any calls to this property to a constant value whenever
possible (and the method is marked as optimizable).

Also change our code to use this new property, and make the corresponding
methods optimizable.

Some shuffling in mmp was required, which meant a little bit more code is now
shared between mtouch and mmp.


src/ObjCRuntime/CpuArchitecture.cs
```diff
@@ -0,0 +1,15 @@
+namespace ObjCRuntime
+{
```
VincentDondain: Minor: but wrong coding guideline (should be same line). That's for this entire PR so it sounds like your settings are incorrect (:

VincentDondain: Clarification: for at least your `enum` and `struct`, those should be same line.

tools/common/Optimizations.cs
```diff
@@ -249,6 +256,9 @@ public void Initialize (Application app)
 				RemoveUnsupportedILForBitcode = app.Platform == Utils.ApplePlatform.WatchOS && app.IsDeviceBuild;
 			}
 #endif
+			// Always enable Runtime.CpuArchitecture inlining.
```
spouliot: That's more enabling _by default_ than _always_ enabled :)

tools/linker/CoreOptimizeGeneratedCode.cs
```diff
@@ -44,9 +45,34 @@ public abstract class CoreOptimizeGeneratedCode : ExceptionalSubStep {
...
 
+		bool? is_arm64_calling_convention;
+
 		public CoreOptimizeGeneratedCode (LinkerOptions options)
 		{
 			Options = options;
+
+			if (Optimizations.InlineIsARM64CallingConvention == true) { 
+				if (options.Target.Abis.Count == 1) {
+					// We can only inline Runtime.InlineIsARM64CallingConvention if the generated code will execute on a single architecture (not if building for armv7+armv7s for instance).
```
spouliot: why ? is `armv7+armv7s` then we know arm64 calling conversion is always `false` right ?

tests/linker/mac/link all/link all-mac.csproj
```diff
@@ -81,6 +81,9 @@
     <Compile Include="..\..\CommonLinkAllTest.cs">
       <Link>CommonLinkAllTest.cs</Link>
     </Compile>
+    <Compile Include="..\..\ios\link all\ILReader.cs">
+      <Link>ILReader.cs</Link>
+    </Compile>
```
spouliot: should be moved to `common` if used for the Mac

src/ObjCRuntime/CpuArchitecture.cs
rolfbjarne: Fixed.

tools/common/Optimizations.cs
rolfbjarne: I've updated the comment.

tools/linker/CoreOptimizeGeneratedCode.cs
rolfbjarne: You're correct, I've changed the logic accordingly.

tests/linker/mac/link all/link all-mac.csproj
rolfbjarne: Done

PR #5571, opened 2019-02-07
Skip CoreNFCLibrary in ApiTypoTest.ConstantsCheck on iPad

akoeplinger:
The test was complaining about CoreNFCLibrary on my iPad.

NFC is not supported on any iPad devices at the moment: https://stackoverflow.com/questions/51532544/is-it-possible-to-use-nfc-on-the-ipad

### DIFF
--- a/tests/introspection/ApiTypoTest.cs
+++ b/tests/introspection/ApiTypoTest.cs
@@ -1105,6 +1105,10 @@ namespace Introspection
 #endif
 				default:
 					if (fi.Name.EndsWith ("Library", StringComparison.Ordinal)) {
+						// NFC is currently not available on iPad
+						if (fi.Name == "CoreNFCLibrary" && UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad)
+							continue;
+
 						Assert.True (CheckLibrary (s), fi.Name);
 					} else {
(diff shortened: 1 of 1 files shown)


tests/introspection/ApiTypoTest.cs
```diff
@@ -1105,6 +1105,10 @@ public void ConstantsCheck ()
 #endif
 				default:
 					if (fi.Name.EndsWith ("Library", StringComparison.Ordinal)) {
+						// NFC is currently not available on iPad
+						if (fi.Name == "CoreNFCLibrary" && UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad)
```
spouliot: That file is shared with XM so `UIDevice` won't build for Mac tests. You either use the product subclass, e.g. `iOSApiTypoTest.cs`, or use an `#if` around the non-portable code

Also for performance that test is only run on simulators and when the latest OS version is selected. That normally means (makefiles & xharness) that the iPad is not used - not that it makes it incorrect to fix :)

spouliot: 🤣 now that the bots have reported back, `UIDevice` is also a bit more limited in the watch profile

akoeplinger:
@monojenkins build

PR #5577, opened 2019-02-08
[tests|mmp] Bump mmp's WarningTests to use Xcode 9.4

dalexsoto:
Fixes xamarin/maccore#1390

Apple removed `NSTextViewIvars.sharedData` in mojave which is needed to link
using Xcode 8.3 so we bumped WarningTests to use Xcode 9.4 which also reproduces
the behaviour the test expects and we ensure that the test is run because we
also require Xcode 9.4 to be there.


dalexsoto:
The mmp test is now passing on the bots, the mtouch failure is expected since the fix just got merged.

PR #5592, opened 2019-02-12
[xharness] Ignore duplicated device pairs.

rolfbjarne:
According to simctl this shouldn't be possible (you get an error if you try),
but it happened on my system somehow, so cope with it.

### DIFF
--- a/tests/xharness/Extensions.cs
+++ b/tests/xharness/Extensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace xharness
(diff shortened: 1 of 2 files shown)


tests/xharness/Extensions.cs
```diff
@@ -118,5 +119,23 @@ public static void DoNotAwait (this Task task)
 			// 
 			// This makes it abundantly clear that the intention is to not await 'DoSomething', and no warnings will be shown either.
 		}
+
+		public static IEnumerable<T> Distinct<T> (this IEnumerable<T> collection, Func<T, T, bool> comparer)
```
VincentDondain: Mmh [Distinct<TSource>(IEnumerable<TSource>, IEqualityComparer<TSource>)](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=netframework-4.7.2#System_Linq_Enumerable_Distinct__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEqualityComparer___0__)?

rolfbjarne: That requires creating a class that implements `IEqualityComparer<XmNode>`, which felt a bit more complicated than this approach.

PR #5605, opened 2019-02-14
[xharness] Fix bug in Simulators.LoadAsync()

akoeplinger:
The "Gizmo" and "Companion" are child elements, not attributes on the SimDevicePair.

Also replaced the custom Distinct() implementation with a comparer which can be used with standard LINQ.


rolfbjarne:
build

PR #5606, opened 2019-02-14
[tests] Don't put command output in exception message.

rolfbjarne:
It can make it very hard to get an overview of test reports, if failing
commands have a lot of output.


rolfbjarne:
Test failures are unrelated.

* CorlibTests/watchOS: 
* Mtouch tests: will be fixed with PR #5605.

PR #5607, opened 2019-02-14
[mtouch] Quote MONO_PATH if needed in debug spew.

rolfbjarne:
Makes c&p commands easier.


rolfbjarne:
Test failure is unrelated, and will be fixed with PR #5605.

PR #5610, opened 2019-02-14
Fix NUnit.ConsoleRunner version in Makefiles

akoeplinger:
It was updated to 3.9.0 in https://github.com/xamarin/xamarin-macios/pull/5538 but a few hardcoded 3.5.0 in the Makefiles were missed


rolfbjarne:
Test failures are unrelated, and were fixed with PR #5605.

PR #5614, opened 2019-02-14
[OpenGL] Fix MonoMacGameView .ctor when hardware acceleration isn't available

### LINKED ISSUE #4417: [XM] Add a `Create` method for the 'AppKit.NSOpenGLPixelFormat' type
### Steps to Reproduce

1. Download a [sample](https://github.com/xamarin/mac-samples/blob/master/GLSLShader)
2. Build and run it on VM

### Expected Behavior
The sample fails on the [line](https://github.com/xamarin/mac-samples/blob/master/GLSLShader/MyOpenGLView.cs#L54):
```
Could not initialize an instance of the type 'AppKit.NSOpenGLPixelFormat': the native 'init*' method returned nil.
```

If I remove `NSOpenGLPixelFormatAttribute.Accelerated` from atrributes collection it will work properly. This attribute means that an app should use hardware-accelaration only. But it doesn't exist for VM.

It also affects creating of `OpenTK.Platform.MacOS.MonoMacGameView`.

### Actual Behavior
Add a `Create` method that can return null. 

### Environment

1. VSfM
https://gist.github.com/mykyta-bondarenko-gl/abc538c727d0bd6d78a338c2f45d3f3f

### Build Logs

1. Exception
https://gist.github.com/mykyta-bondarenko-gl/ec66e97deada8ea29a5ef5170e144069

### Example Project 
[Download](https://github.com/xamarin/mac-samples/blob/master/GLSLShader)

VincentDondain:
- Fixes #4417: [XM] Add a `Create` method for the 'AppKit.NSOpenGLPixelFormat' type
  (https://github.com/xamarin/xamarin-macios/issues/4417)
- Doing `new MonoMacGameView (ContentView.Frame)` on CI fails because hardware acceleration isn't available.
  One of our samples had that code: https://github.com/xamarin/mac-samples/blob/a41f4387dc28617e4207b3759617e788aad383ed/MacOpenTK/MacOpenTK/MainWindow.cs#L40
  This made some QA sample tests fail.

### DIFF
--- a/src/OpenGL/MonoMacGameView.cs
+++ b/src/OpenGL/MonoMacGameView.cs
@@ -29,6 +29,7 @@
 using System;
 using System.Drawing;
 using System.ComponentModel;
+using System.Linq;
 
 using Foundation;
 using AppKit;
@@ -73,7 +74,14 @@ namespace OpenTK.Platform.MacOS
 				NSOpenGLPixelFormatAttribute.ColorSize, 24,
 				NSOpenGLPixelFormatAttribute.DepthSize, 16 };
 
-			pixelFormat = new NSOpenGLPixelFormat (attribs);
+			try {
+				pixelFormat = new NSOpenGLPixelFormat (attribs);
+			} catch (Exception) {
+				// Fails on VM because there is no hardware-acceleration
+				// https://github.com/xamarin/xamarin-macios/issues/4417
+				attribs = attribs.Skip (1).ToArray ();
+				pixelFormat = new NSOpenGLPixelFormat (attribs);
+			}
 
 			if (pixelFormat == null)
 				Console.WriteLine ("No OpenGL pixel format");
--- a/src/OpenGL/MonoMacGameView.cs
+++ b/src/OpenGL/MonoMacGameView.cs
@@ -29,7 +29,6 @@
 using System;
 using System.Drawing;
 using System.ComponentModel;
-using System.Linq;
 
 using Foundation;
 using AppKit;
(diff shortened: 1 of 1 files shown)


src/OpenGL/MonoMacGameView.cs
```diff
@@ -73,7 +74,14 @@ public MonoMacGameView (CGRect frame, NSOpenGLContext context) : base(frame)
 				NSOpenGLPixelFormatAttribute.ColorSize, 24,
 				NSOpenGLPixelFormatAttribute.DepthSize, 16 };
 
-			pixelFormat = new NSOpenGLPixelFormat (attribs);
+			try {
+				pixelFormat = new NSOpenGLPixelFormat (attribs);
+			} catch (Exception) {
```
spouliot: can you be more specific ? i.e. you're not handling `OutOfMemoryException` with that

src/OpenGL/MonoMacGameView.cs
```diff
@@ -73,7 +74,14 @@ public MonoMacGameView (CGRect frame, NSOpenGLContext context) : base(frame)
...
+				// Fails on VM because there is no hardware-acceleration
+				// https://github.com/xamarin/xamarin-macios/issues/4417
+				attribs = attribs.Skip (1).ToArray ();
```
spouliot: beside bringing LINQ code, it's not really efficient (even less since there's already allocations inside the `NSOpenGLPixelFormat ` binding code).

You can probably avoid that by
1. moving `NSOpenGLPixelFormatAttribute.Accelerated` at the end (I don't think the order matter);
2. if it fails, replace the last element with `0` (which is like `nil` and will mark the end of the list)
3. try again

Should work - without LINQ code and extraneous memory allocations :)

src/OpenGL/MonoMacGameView.cs
VincentDondain: It doesn't look like we're getting a more specific exception when the issue happens: https://gist.github.com/VincentDondain/2fb0b74548a390a25a33b81ad2fa172e

```
Unhandled Exception:
System.Exception: Could not initialize an instance of the type 'AppKit.NSOpenGLPixelFormat': the native 'init*' method returned nil.
```

src/OpenGL/MonoMacGameView.cs
```diff
@@ -29,6 +29,7 @@
 using System;
 using System.Drawing;
 using System.ComponentModel;
+using System.Linq;
```
dalexsoto: No linq in our source 😄

src/OpenGL/MonoMacGameView.cs
spouliot: hmm... that's something we should change - but not in the PR :)

src/OpenGL/MonoMacGameView.cs
VincentDondain: Well don't do a `git grep System.Linq` in `src/` 🙈

dalexsoto: @spouliot let the git blame games begin :P

VincentDondain:
@monojenkins backport xcode10.2

PR #5628, opened 2019-02-19
[xharness] Improve app installation on device.

rolfbjarne:
* Instead of calculating a timeout based on the app size and an estimated
  transfer speed, keep transferring as long as mlaunch outputs something to
  stdout. The actual transfer speed can vary a lot, in particular if
  installing to multiple nearby watches simultaneously.

* Improve progress text in the html report during installation to include more
  statistics for the impatient.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -264,7 +264,7 @@ namespace xharness
 			}
 		}
 
-		public async Task<ProcessExecutionResult> InstallAsync ()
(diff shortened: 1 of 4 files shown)


rolfbjarne:
I even had a screenshot for this one, but forgot to add it 😞. Anyways, here it is:

screenshot 2019-02-14 12 58 18

A couple of tweaks were done after I made the screenshot, so it's not 100% correct, but it's close enough 😄

PR #5629, opened 2019-02-19
[mtouch] Don't use the native linker to create fat executables.

rolfbjarne:
Don't use the native linker to create fat executables, instead link each
architecture separately, and then manually lipo everything together at the
end. This requires a few changes since we need to keep track of the linker
flags per architecture.

The problem is that bitcode files (.bc) do not correspond with a particular
architecture, so the linker can't distinguish between .bc files for armv7k and
.bc files for arm64_32. So if we pass all together to the linker, the linker
will add all .bc files to both architectures, thus duplicating everything (and
the linking fails with duplicate symbols errors).

### DIFF
--- a/tools/mtouch/Application.cs
+++ b/tools/mtouch/Application.cs
@@ -199,12 +199,15 @@ namespace Xamarin.Bundler {
 		public string UserGccFlags;
 
 		// If we didn't link the final executable because the existing binary is up-to-date.
-		bool cached_executable; 
+		bool cached_executable {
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Test failure are unrelated.

* */iOS Unified 32-bits: https://github.com/xamarin/maccore/issues/1412
* xammac tests/Mac/Release: https://github.com/xamarin/maccore/issues/1332

Issue #3553, opened 2018-02-20
[Xcode 9.3] MetalPerformanceShaders Bindings


dalexsoto:
### Bindings of MetalPerformanceShaders and MPS* sub frameworks

Changes introduced in `MetalPerformanceShaders` in Xcode 9.3 are marked as **experimental** inside `MetalPerformanceShaders.h` so we have no way to know which APIs are going to stay and which ones are going to change.

In order to avoid breaking changes in our binaries we will postpone the bindings until we get a more clear picture where the API is going.

https://github.com/xamarin/xamarin-macios/wiki/xcode9.3-Bindings-Status

dalexsoto:
Part 1 of MPS is up https://github.com/xamarin/xamarin-macios/pull/5466

This includes updates for:

MetalPerformanceShaders.framework
MPSCore.framework
MPSImage.framework
MPSMatrix.framework
MPSRayIntersector.framework

Pending:

MPSNeuralNetwork.framework

praeclarum:
Referring to all of these apis: https://github.com/xamarin/xamarin-macios/blob/49ef93670171ef4e11656a8ed0f3e99f20079c65/tests/xtro-sharpie/iOS-MetalPerformanceShaders.ignore#L201

dalexsoto:
Hello @praeclarum we hear you!

I am actively working on `MetalPerformanceShaders` that includes `MPSNeuralNetwork.framework` right now, we could not do it earlier (back on Xcode 9.3) because it was an API that was marked as experimental and just promoted to stable in Xcode 10 and we have a hard policy to maintain binary compatibility between our releases. Also the task to bind it has been pushed back in the past in favor of other tasks mainly because we have not had any previous request for it.

I expect to have this done within the Xcode 10.2 timeframe so hopefully all of this should go out with the next stable release that includes [Xcode 10.2](https://github.com/xamarin/xamarin-macios/commits/xcode10.2), we are tracking our binding progress [here](https://github.com/xamarin/xamarin-macios/wiki/xcode10.2-Bindings-Status).

Now if you want to get your hands dirty what I would do is create a binding project with the classes you need, if you are using other MPS classes that comes from Xamarin.iOS.dll you should automatically link against MPS framework, else you can give `-framework MetalPerformanceShaders` to `mtouch` in your main project.

Or

[...]

PR #5490, opened 2019-01-25
[MetalPerformanceShaders] Update bindings from Xcode 9.2 to 10.2 - Part 1 of 2

dalexsoto:
Fixes: xamarin/xamarin-macios#5362

This includes updates for:

* MetalPerformanceShaders.framework
* MPSCore.framework
* MPSImage.framework
* MPSMatrix.framework
* MPSRayIntersector.framework

Pending:

* MPSNeuralNetwork.framework

This also includes a couple of cherry-picked commits that are needed so the metal performance shaders work applies cleanly and it does not create more merge hell when this branch is merged into master.

PR #5630, opened 2019-02-19
[MetalPerformanceShaders] Update bindings from Xcode 9.2 to 10.2 - Part 2 of 2

dalexsoto:
Fixes xamarin/xamarin-macios#3553

This is the second part of MPS bindings.

Bindings for:

- MPSNeuralNetwork.framework


dalexsoto:
Both breaking changes are expected, `GetDestinationImageDescriptor` was introduced in https://github.com/xamarin/xamarin-macios/commit/3e0bcf2b32985c745f5144d28a718a191145970a#diff-d05bb29f8320ca898f34fbaeb9d0385dR1735 which is [Part 1](https://github.com/xamarin/xamarin-macios/pull/5490) and the change was done to match other API. 

As for `Count` Apple added more values to the enum, and that particular member must always be the last (documented) and its value is expected to change over time as more members are added, it seems just to be a sugar control value.

src/MetalPerformanceShaders/MPSDefs.cs
```diff
@@ -199,7 +199,13 @@ public enum MPSCnnNeuronType {
...
 		PReLU,
 		ReLun,
-		Count,
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		Power,
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		Exponential,
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		Logarithm,
+		Count, // must always be last
```
spouliot: That's problematic. Maybe we can replace it with an extension method that will always provide the correct value (length - 1).

src/metalperformanceshaders.cs
```diff
@@ -3179,9 +4231,228 @@ interface MPSCnnPoolingNode {
...
+		float Delta { get; set; }
+
+		[Export ("p0")]
+		float P0 { get; set; }
+
+		[Export ("pm")]
+		float Pm { get; set; }
+
+		[Export ("ps")]
+		float Ps { get; set; }
```
spouliot: same about those 2 letters properties ^

src/metalperformanceshaders.cs
```diff
@@ -3340,91 +4611,526 @@ interface MPSNNBinaryArithmeticNode {
...
+		[Export ("initWithSourceGradient:sourceImage:gradientState:")]
+		IntPtr Constructor (MPSNNImageNode sourceGradient, MPSNNImageNode sourceImage, MPSNNGradientStateNode gradientState);
+	}
+
+
+
+
+
+
+
```
spouliot: sticky enter key ? ;-)

src/metalperformanceshaders.cs
```diff
@@ -3340,91 +4611,526 @@ interface MPSNNBinaryArithmeticNode {
...
+		// [TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		// [Export ("initWithDevice:resultImages:resultsAreNeeded:")]
+		// [DesignatedInitializer]
+		// IntPtr Constructor (MTLDevice device, MPSNNImageNode[] resultImages, [NullAllowed] bool* areResultsNeeded);
+
+		// [TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		// [Static]
+		// [Export ("graphWithDevice:resultImages:resultsAreNeeded:")]
+		// [return: NullAllowed]
+		// MPSNNGraph GraphWithDevice (MTLDevice device, MPSNNImageNode[] resultImages, [NullAllowed] bool* areResultsNeeded);
```
spouliot: ^ ???

src/metalperformanceshaders.cs
```diff
@@ -3340,91 +4611,526 @@ interface MPSNNBinaryArithmeticNode {
...
+
+		[Deprecated (PlatformName.TvOS, 11, 3)]
+		[Deprecated (PlatformName.iOS, 11, 3)]
+		[Deprecated (PlatformName.MacOSX, 10, 13, 4)]
+		[Static]
+		[Export ("graphWithDevice:resultImage:")]
+		[return: NullAllowed]
+		MPSNNGraph Create (IMTLDevice device, MPSNNImageNode resultImage);
```
spouliot: again, not clear is we should add deprecated API

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+
+		[Export ("temporaryResultStateForCommandBuffer:sourceImage:sourceStates:destinationImage:")]
+		[return: NullAllowed]
+		MPSCnnBatchNormalizationState GetTemporaryResultState (IMTLCommandBuffer commandBuffer, MPSImage sourceImage, [NullAllowed] NSArray<MPSState> sourceStates, MPSImage destinationImage);
+
+		[Deprecated (PlatformName.TvOS, 12, 0, message: "Please use 'ReloadGammaAndBetaFromDataSource' and/or 'ReloadMeanAndVarianceFromDataSource' instead.")]
+		[Deprecated (PlatformName.iOS, 12, 0, message: "Please use 'ReloadGammaAndBetaFromDataSource' and/or 'ReloadMeanAndVarianceFromDataSource' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 14, message: "Please use 'ReloadGammaAndBetaFromDataSource' and/or 'ReloadMeanAndVarianceFromDataSource' instead.")]
+		[Export ("reloadDataSource:")]
+		void ReloadDataSource (IMPSCnnBatchNormalizationDataSource dataSource);
```
spouliot: same

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+		IMPSCnnConvolutionDataSource DataSource { get; }
+
+		[Export ("gradientOption", ArgumentSemantic.Assign)]
+		MPSCnnConvolutionGradientOption GradientOption { get; set; }
+
+		[Deprecated (PlatformName.TvOS, 12, 0, message: "This doesn't do anything. It is here for backward compatibility. 'MPSCnnConvolutionGradient doesn't serialize weights. It gets weight from 'State.Convolution.DataSource' on first use i.e. first 'Encode' call.")]
+		[Deprecated (PlatformName.iOS, 12, 0, message: "This doesn't do anything. It is here for backward compatibility. 'MPSCnnConvolutionGradient doesn't serialize weights. It gets weight from 'State.Convolution.DataSource' on first use i.e. first 'Encode' call.")]
+		[Deprecated (PlatformName.MacOSX, 10, 14, message: "This doesn't do anything. It is here for backward compatibility. 'MPSCnnConvolutionGradient doesn't serialize weights. It gets weight from 'State.Convolution.DataSource' on first use i.e. first 'Encode' call.")]
+		[Export ("serializeWeightsAndBiases")]
+		bool SerializeWeightsAndBiases { get; set; }
```
spouliot: same - even more in this case since it seems it's not useful (and we don't have backward compatibility to consider since it's new bindings)

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+
+		[Export ("encodeWithCoder:")]
+		void Encode (NSCoder coder);
+
+		[Export ("initWithCoder:")]
+		IntPtr Constructor (NSCoder decoder);
+
+		[Static]
+		[Export ("supportsSecureCoding")]
+		bool SupportsSecureCoding { get; }
```
spouliot: `[Static]` here makes me think the binding won't work.
It's likely `static-per-type`, so if `Foo` conforms to the protocol then `Foo.SupportsSecureCoding` might be true, while `Bar`, also conforming, might be `false`

src/metalperformanceshaders.cs
```diff
@@ -3509,14 +5244,31 @@ interface MPSCnnConvolutionDataSource : NSCopying {
...
+		[return: NullAllowed]
+		MPSCnnConvolutionWeightsAndBiasesState Update (IMTLCommandBuffer commandBuffer, MPSCnnConvolutionGradientState gradientState, MPSCnnConvolutionWeightsAndBiasesState sourceState);
+
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		[Export ("updateWithGradientState:sourceState:")]
+		bool Update (MPSCnnConvolutionGradientState gradientState, MPSCnnConvolutionWeightsAndBiasesState sourceState);
+
+		[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		[Export ("copyWithZone:device:")]
+		IMPSCnnConvolutionDataSource Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: This needs a `[return: Release()]`

src/metalperformanceshaders.cs
```diff
@@ -3826,6 +5607,88 @@ interface MPSMatrixNeuron {
...
+
+		[Export ("encodeToCommandBuffer:gradientMatrix:inputMatrix:biasVector:resultGradientForDataMatrix:resultGradientForBiasVector:")]
+		void Encode (IMTLCommandBuffer commandBuffer, MPSMatrix gradientMatrix, MPSMatrix inputMatrix, [NullAllowed] MPSVector biasVector, MPSMatrix resultGradientForDataMatrix, [NullAllowed] MPSVector resultGradientForBiasVector);
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder decoder, IMTLDevice device);
+
+		[Export ("copyWithZone:device:")]
+		MPSMatrixNeuronGradient Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release ()]`

src/metalperformanceshaders.cs
```diff
@@ -3826,6 +5607,88 @@ interface MPSMatrixNeuron {
...
+
+		[Export ("encodeGradientForWeightsAndBiasToCommandBuffer:gradientMatrix:inputMatrix:resultGradientForWeightMatrix:resultGradientForBiasVector:")]
+		void EncodeGradientForWeightsAndBias (IMTLCommandBuffer commandBuffer, MPSMatrix gradientMatrix, MPSMatrix inputMatrix, MPSMatrix resultGradientForWeightMatrix, [NullAllowed] MPSVector resultGradientForBiasVector);
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder aDecoder, IMTLDevice device);
+
+		[Export ("copyWithZone:device:")]
+		MPSMatrixFullyConnectedGradient Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release ()]`

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+		// Keeping the same name as in the parent class so it ends up in an overload
+		[Export ("encodeToCommandBuffer:gradientMatrix:forwardOutputMatrix:resultMatrix:")]
+		void EncodeToCommandBuffer (IMTLCommandBuffer commandBuffer, MPSMatrix gradientMatrix, MPSMatrix forwardOutputMatrix, MPSMatrix resultMatrix);
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder aDecoder, IMTLDevice device);
+
+		[Export ("copyWithZone:device:")]
+		[return: Release]
```
rolfbjarne: 👍

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+		[Export ("initWithCoder:")]
+		IntPtr Constructor (NSCoder decoder);
+
+		[Static]
+		[Export ("supportsSecureCoding")]
+		bool SupportsSecureCoding { get; }
+
+		[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		[Export ("copyWithZone:device:")]
+		IMPSCnnBatchNormalizationDataSource Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release()]`

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+
+		[Export ("initWithWeights:biases:")]
+		IntPtr Constructor (IMTLBuffer weights, [NullAllowed] IMTLBuffer biases);
+
+		[Export ("initWithDevice:cnnConvolutionDescriptor:")]
+		IntPtr Constructor (IMTLDevice device, MPSCnnConvolutionDescriptor descriptor);
+
+		[Static]
+		[Export ("temporaryCNNConvolutionWeightsAndBiasesStateWithCommandBuffer:cnnConvolutionDescriptor:")]
+		MPSCnnConvolutionWeightsAndBiasesState GetTemporaryCNNConvolutionWeightsAndBiasesState (IMTLCommandBuffer commandBuffer, MPSCnnConvolutionDescriptor descriptor);
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+		[Export ("initWithCoder:")]
+		IntPtr Constructor (NSCoder decoder);
+
+		[Static]
+		[Export ("supportsSecureCoding")]
+		bool SupportsSecureCoding { get; }
+
+		[TV (12,0), Mac (10,14, onlyOn64: true), iOS (12,0)]
+		[Export ("copyWithZone:device:")]
+		IMPSCnnInstanceNormalizationDataSource Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release()]`

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+	interface MPSCnnArithmeticGradientState {
+
+	}
+
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnBinaryKernel), Name = "MPSCNNArithmetic")]
+	[DisableDefaultCtor]
+	interface MPSCnnArithmetic {
+
+		// Inlined but you are not supposed to use this class, only its subclasses
```
rolfbjarne: You could make the constructors protected. That way only subclasses would be able to use them.

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+		[DesignatedInitializer]
+		IntPtr Constructor (IMTLDevice device);
+	}
+
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnGradientKernel), Name = "MPSCNNArithmeticGradient")]
+	[DisableDefaultCtor]
+	interface MPSCnnArithmeticGradient {
+
+		// Inlined but you are not supposed to use this class, only its subclasses
```
rolfbjarne: Same: protected.

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+
+		[Export ("encodeToCommandBuffer:inputMatrix:meanVector:varianceVector:gammaVector:betaVector:resultMatrix:")]
+		void Encode (IMTLCommandBuffer commandBuffer, MPSMatrix inputMatrix, MPSVector meanVector, MPSVector varianceVector, [NullAllowed] MPSVector gammaVector, [NullAllowed] MPSVector betaVector, MPSMatrix resultMatrix);
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder decoder, IMTLDevice device);
+
+		[Export ("copyWithZone:device:")]
+		MPSMatrixBatchNormalization Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release ()]`

src/metalperformanceshaders.cs
```diff
@@ -4203,159 +6066,2016 @@ interface MPSImageReduceUnary {
...
+
+		[Export ("encodeToCommandBuffer:gradientMatrix:inputMatrix:meanVector:varianceVector:gammaVector:betaVector:resultGradientForDataMatrix:resultGradientForGammaVector:resultGradientForBetaVector:")]
+		void Encode (IMTLCommandBuffer commandBuffer, MPSMatrix gradientMatrix, MPSMatrix inputMatrix, MPSVector meanVector, MPSVector varianceVector, [NullAllowed] MPSVector gammaVector, [NullAllowed] MPSVector betaVector, MPSMatrix resultGradientForDataMatrix, [NullAllowed] MPSVector resultGradientForGammaVector, [NullAllowed] MPSVector resultGradientForBetaVector);
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder decoder, IMTLDevice device);
+
+		[Export ("copyWithZone:device:")]
+		MPSMatrixBatchNormalizationGradient Copy ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release ()]`

src/metalperformanceshaders.cs
```diff
@@ -4364,192 +8084,170 @@ interface MPSRayIntersector : NSSecureCoding, NSCopying {
...
-		[Export ("initWithGroup:")]
-		[DesignatedInitializer]
-		IntPtr Constructor (MPSAccelerationStructureGroup group);
+		IntPtr Constructor (NSCoder decoder, IMTLDevice device);
 
-		[Export ("initWithCoder:group:")]
-		[DesignatedInitializer]
-		IntPtr Constructor (NSCoder aDecoder, MPSAccelerationStructureGroup group);
+		[Export ("copyWithZone:device:")]
+		MPSRnnMatrixTrainingLayer CopyWithZone ([NullAllowed] NSZone zone, [NullAllowed] IMTLDevice device);
```
rolfbjarne: `[return: Release ()]`

src/metalperformanceshaders.cs
```diff
@@ -1272,6 +1500,32 @@ interface MPSCnnCrossChannelNormalization {
 		IntPtr Constructor (NSCoder aDecoder, IMTLDevice device);
 	}
 
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnGradientKernel), Name = "MPSCNNCrossChannelNormalizationGradient")]
+	[DisableDefaultCtor]
+	interface MPSCNNCrossChannelNormalizationGradient {
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -1284,6 +1538,20 @@ interface MPSCnnSoftMax {
 		IntPtr Constructor (IMTLDevice device);
 	}
 
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnGradientKernel), Name = "MPSCNNSoftMaxGradient")]
+	[DisableDefaultCtor]
+	interface MPSCNNSoftMaxGradient {
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -2560,6 +3071,88 @@ interface MPSCnnDilatedPoolingMax {
...
+
+		[Export ("initWithCoder:device:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSCoder decoder, IMTLDevice device);
+	}
+
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnPoolingGradient), Name = "MPSCNNPoolingL2NormGradient")]
+	[DisableDefaultCtor]
+	interface MPSCNNPoolingL2NormGradient {
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -3179,9 +4231,228 @@ interface MPSCnnPoolingNode {
...
+		nuint StrideInPixelsX { get; }
+
+		[Export ("strideInPixelsY")]
+		nuint StrideInPixelsY { get; }
+	}
+
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSCnnPoolingGradientNode), Name = "MPSCNNPoolingMaxGradientNode")]
+	[DisableDefaultCtor]
+	interface MPSCNNPoolingMaxGradientNode {
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -3179,9 +4231,228 @@ interface MPSCnnPoolingNode {
...
+		float Beta { get; set; }
+
+		[Export ("delta")]
+		float Delta { get; set; }
+	}
+
+	[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+	[BaseType (typeof (MPSNNGradientFilterNode), Name = "MPSCNNLocalContrastNormalizationGradientNode")]
+	[DisableDefaultCtor]
+	interface MPSCNNLocalContrastNormalizationGradientNode {
```
rolfbjarne: `CNN` -> `Cnn`

src/metalperformanceshaders.cs
```diff
@@ -710,15 +710,35 @@ interface MPSCnnKernel {
...
 		[Export ("encodeToCommandBuffer:sourceImage:destinationImage:")]
 		void EncodeToCommandBuffer (IMTLCommandBuffer commandBuffer, MPSImage sourceImage, MPSImage destinationImage);
 
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)] // keeping same name to match existing member
+		[Export ("encodeToCommandBuffer:sourceImage:destinationState:destinationImage:")]
+		void EncodeToCommandBuffer (IMTLCommandBuffer commandBuffer, MPSImage sourceImage, MPSState destinationState, MPSImage destinationImage);
+
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		[Export ("encodeBatchToCommandBuffer:sourceImages:destinationImages:")]
+		void EncodeBatch (IMTLCommandBuffer commandBuffer, NSArray<MPSImage> sourceImages, NSArray<MPSImage> destinationImages);
```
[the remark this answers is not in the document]
dalexsoto: My reasoning behind using `NSArray<T>` in MPS is because both `NSArray<MPSImage>` and `NSArray<MPSState>` are extensively used trough the API and it seems that they will be reused several times so instead of going back and forth between managed array and `NSArray` on each API I figured just to keep everything in NSArray, also sometimes they are used as an out container, meaning that you just give it an `NSArray` of a fixed size and the API will fill its contents.

src/metalperformanceshaders.cs
```diff
@@ -710,15 +710,35 @@ interface MPSCnnKernel {
...
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
 		[Export ("destinationFeatureChannelOffset")]
 		nuint DestinationFeatureChannelOffset { get; set; }
 
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		[Export ("sourceFeatureChannelMaxCount")]
+		nuint SourceFeatureChannelMaxCount { get; set; }
+
+		[Export ("sourceFeatureChannelOffset")]
+		nuint SourceFeatureChannelOffset { get; set; }
```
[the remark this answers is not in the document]
dalexsoto: Yes, you are correct, my bad

src/metalperformanceshaders.cs
```diff
@@ -1243,6 +1439,38 @@ interface MPSCnnLocalContrastNormalization {
...
+		float Beta { get; set; }
+
+		[Export ("delta")]
+		float Delta { get; set; }
+
+		[Export ("p0")]
+		float P0 { get; set; }
+
+		[Export ("pm")]
+		float Pm { get; set; }
```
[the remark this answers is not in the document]
dalexsoto: Unfortunately we already have other API that way checked in I am not sure we want to be inconsistent, see 

https://github.com/xamarin/xamarin-macios/blob/5f33d73077050c3d06610c978a8812d9724ddb10/src/metalperformanceshaders.cs#L1217-L1224

src/metalperformanceshaders.cs
```diff
@@ -1732,7 +2028,13 @@ interface MPSState {
...
 		[iOS (11,3), TV (11,3), Mac (10,13,4, onlyOn64: true)]
 		[Export ("destinationImageDescriptorForSourceImages:sourceStates:forKernel:suggestedDescriptor:")]
-		MPSImageDescriptor GetDestinationImageDescriptor (MPSImage [] sourceImages, [NullAllowed] MPSState [] sourceStates, MPSKernel kernel, MPSImageDescriptor inDescriptor);
+		MPSImageDescriptor GetDestinationImageDescriptor (NSArray<MPSImage> sourceImages, [NullAllowed] NSArray<MPSState> sourceStates, MPSKernel kernel, MPSImageDescriptor inDescriptor);
+
+		[Deprecated (PlatformName.TvOS, 12, 0, message: "Please use 'GetResource (nuint, bool)' instead.")]
+		[Deprecated (PlatformName.iOS, 12, 0, message: "Please use 'GetResource (nuint, bool)' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 13, 4, message: "Please use 'GetResource (nuint, bool)' instead.")]
+		[NullAllowed, Export ("resource", ArgumentSemantic.Retain)]
+		IMTLResource Resource { get; }
```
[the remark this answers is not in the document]
dalexsoto: Yeah, for this one it wasn't too short lived (like just a minor version like others) since it got introduced in iOS 11 so I would vote to keep it.

src/metalperformanceshaders.cs
dalexsoto: Yes those two seemed to be leaked into this beta release so just a heads up for our future selves

src/metalperformanceshaders.cs
dalexsoto: Agree on this one, it did not make it fully in iOS 11 to 11.3 and there are alternatives to it.

src/metalperformanceshaders.cs
dalexsoto: Agree, super short lived 11,3 -> 12,0

src/metalperformanceshaders.cs
dalexsoto: Agreed, super short lived 11.3 -> 12.0

src/metalperformanceshaders.cs
dalexsoto: Yes you are right, this needs to be exposed into implementors of the protocol, does not make much sense exposing it here. 👍

src/metalperformanceshaders.cs
dalexsoto: I would normally agree but subclasses already inline the constructor that is available for them and it is not always both of these.

src/metalperformanceshaders.cs
dalexsoto: I would normally agree but subclasses already inline the constructor that is available for them and it is not always both of these.

src/MetalPerformanceShaders/MPSKernel.cs
```diff
@@ -320,6 +320,58 @@ public MPSImageLaplacianPyramidAdd (IMTLDevice device, nuint kernelWidth, nuint
 			}
 		}
 	}
+
+	public partial class MPSCnnBinaryConvolutionNode {
+		[TV (11,3), Mac (10,13,4, onlyOn64: true), iOS (11,3)]
+		public static MPSCnnBinaryConvolutionNode Create (MPSNNImageNode sourceNode, IMPSCnnConvolutionDataSource weights, float [] outputBiasTerms, float [] outputScaleTerms, float [] inputBiasTerms, float [] inputScaleTerms, MPSCnnBinaryConvolutionType type, MPSCnnBinaryConvolutionFlags flags)
+		{
+			unsafe {
```
rolfbjarne: Can we validate the size of the arrays before passing pointers to native code?

dalexsoto: I don’t think there is an easy way to check for parameters sizes on those methods, they can even be omitted according to docs.

PR #5635, opened 2019-02-19
[uikit] Update for xcode 10.2 beta 3

spouliot:


src/uikit.cs
```diff
@@ -5482,7 +5482,20 @@ partial interface UIFontDescriptor : NSSecureCoding, NSCopying {
...
 		[Export ("fontDescriptorWithSymbolicTraits:")]
 		UIFontDescriptor CreateWithTraits (UIFontDescriptorSymbolicTraits symbolicTraits);
-		
+
+		[NoiOS][NoTV]
+		[Watch (5,2)]
+		[Export ("fontDescriptorWithDesign:")]
+		[return: NullAllowed]
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		UIFontDescriptor CreateWithDesign (NSString design);
```
dalexsoto: Any reason for not using `BindAs` here?

```csharp
UIFontDescriptor CreateWithDesign ([BindAs (typeof (UIFontDescriptorSystemDesign))] NSString design);
```

spouliot: That would _hide_ the `NSString` overload, right ?

PR #5637, opened 2019-02-20
[mtouch] Automatically install when building the csproj.

rolfbjarne:
I keep forgetting to run 'make mtouch' after changing mtouch, and then
expecting a test project to pick up the changes.


rolfbjarne:
Test failure is unrelated

* monotouch-test/tvOS/Debug: https://github.com/xamarin/maccore/issues/1332

Issue #5618, opened 2019-02-15
assertion/abort running System.Core unit tests with `--interpreter=-System.Core`


spouliot:
### Steps to Reproduce

1. Execute System.Core unit tests on devices using `--interpreter=-System.Core`

### Expected Behavior

works, no failure

### Actual Behavior

abort

```
019-02-15 14:55:22.781 SystemCoreTests[31549:16758495] error: * Assertion at ../../../../../mono/mini/interp/interp.c:1965, condition `is_ok (error)' not met, function:do_jit_call, Attempting to JIT compile method '(wrapper unknown) object:gsharedvt_out_sig (intptr,int&,intptr)' while running in aot-only mode. See https://docs.microsoft.com/xamarin/ios/internals/limitations for more information.
 assembly:<unknown assembly> type:<unknown type> member:(null)
```

### Environment

```
=== Visual Studio Enterprise 2019 Preview for Mac ===

Version 8.0 Preview (8.0 build 2182)
Installation UUID: a8ba93d0-79b6-41af-a590-4edb9c2095b4
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.6.0.2 (d16-0 / 040682909)

	Package version: 518000263

=== Mono Framework MDK ===

Runtime:
	Mono 5.18.0.263 (2018-08/b0416692580) (64-bit)
	Package version: 518000263

=== NuGet ===

Version: 4.8.0.5385

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.1.2
	2.0.5
	2.0.3
SDK: /usr/local/share/dotnet/sdk/2.1.302/Sdks
SDK Versions:
	2.1.302
	2.1.4
	2.0.3
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.18.0/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.9
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Updater ===

Version: 11

=== Apple Developer Tools ===

Xcode 10.1 (14460.46)
Build 10B61

=== Xamarin.Mac ===

Version: 5.7.1.29 (Visual Studio Enterprise)
Hash: 9753c371
Branch: xi-interpreter
Build date: 2019-02-12 14:42:24-0500

=== Xamarin.iOS ===

[...]

PR #5655, opened 2019-02-21
[mtouch] use "--aot=interp,full,[...]" for every assembly if interpreter mixed mode is enabled

lewurm:
The Mono AOT compiler maintains a set of signatures of compiled methods.
Those signatures are necessary to emit wrappers to enable the transition
from interpreter->AOT code. Thus, they must be collected for each
assembly.

Contributes to https://github.com/xamarin/xamarin-macios/issues/5618


rolfbjarne:
build

dalexsoto:
build

spouliot:
build

PR #5661, opened 2019-02-22
Simplify ARM64-based CPU detection logic.

rolfbjarne:
* We only need to do a runtime check on watchOS.
* On watchOS assume we're on a ARM64-based CPU as long as we're not on a
  ARMv7k CPU.

Fixes an issue where we failed to detect a 64-bit iOS CPU as ARM64-based (in
particular iPhone X's ARM64v8).

Draft PR for now since I need to test this on a lot of devices first.


src/ObjCRuntime/Runtime.cs
```diff
@@ -1739,39 +1739,24 @@ struct NXArchInfo {
...
-			// say a 'arm64e2' cpu. This code should continue to work in that case.
-			var arch = CpuArchitecture;
-			var cputype = ((ulong) arch) >> 32;
-			switch (cputype) {
-			case 12 | 0x01000000: // CPU_TYPE_ARM | CPU_ARCH_ABI64. This includes both arm64 and arm64e.
-			case 12 | 0x02000000: // CPU_TYPE_ARM | CPU_ARCH_ABI64_32. This is arm64_32
-				return true;
-			default:
+#elif __IOS__ || __TVOS__
+			return IntPtr.Size == 8 && Arch == Arch.DEVICE;
```
spouliot: not a big deal (it's optimized whenever the linker is used) but you can assume 64 bits for tvOS.

src/ObjCRuntime/Runtime.cs
```diff
@@ -1739,39 +1739,24 @@ struct NXArchInfo {
...
+#elif __WATCHOS__
+			if (Arch != Arch.DEVICE)
 				return false;
```
spouliot: that check could be moved for all `!MONOMAC`

src/ObjCRuntime/Runtime.cs
rolfbjarne: Yes, I thought about it, but figured this would be fewer lines of code :smile:

akoeplinger:
I tried this on my iPad mini 4 and it does fix the crash I was seeing 👍

rolfbjarne:
OK, this looks good on devices as well.

PR #5662, opened 2019-02-22
[siminstaller] Improve failure scenarios.

rolfbjarne:
* Improve the behavior when we fail to download the .dvtdownloadableindex file
  to report a "Not Found" if we get a "Forbidden" (403) response, since that's
  what it really is.
* Improve the behavior when we fail to download the .dvtdownloadableindex file
  to check if the requested simulators are already installed instead of always
  failing. This makes it possible to manually install the required simulators
  if needed (Apple might not have published simulators for the Xcode version
  we're using), and this would allow any checks to pass if the required
  simulators are already installed.
* Improve the code to modify the PackageInfo XML to not use simple string
  replacement (it's too fragile), use proper XML parsing instead.
* Fix "time left" math, I have no idea what my previous math was doing.

### DIFF
--- a/tools/siminstaller/Program.cs
+++ b/tools/siminstaller/Program.cs
@@ -130,9 +130,32 @@ namespace xsiminstaller {
 			if (!File.Exists (tmpfile)) {
 				var wc = new WebClient ();
 				try {
+					if (verbose > 0)
+						Console.WriteLine ($"Downloading '{uri}'");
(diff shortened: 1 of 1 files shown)


rolfbjarne:
Test failures are unrelated:

* apitest: https://github.com/xamarin/maccore/issues/1446
* monotouch-test/watchOS: https://github.com/xamarin/maccore/issues/1332

PR #5663, opened 2019-02-22
[api-diff] Remove apidiff XML reference files. Fixes #4891

### LINKED ISSUE #4891: Remove API diff XML references from repository
Our API diff is based on XML definitions. They are huge (about 327MB today) and while bots needs them (all the time) many people can live happy lives without them.

Removing them from the repo would have some advantages, e.g.
- faster clone of the repository;
- less conflicts when merging branches;
- less branches to update when a new stable is released;

The downside is that we need to store them elsewhere and download (cache) them on-demand (being text they should compress well - *update* a simple, default zip is 14.2MB). 

We are already doing similar caching for other things so it's nothing really new.

Notes:
* `xi` and `xm` directories should be split (to download) since they can diverge

spouliot:
Instead download the assemblies (part of existing bundle.zip) of the
current release and generate the XML locally (no storage in github)
before doing the diff against the newer, just built, assemblies.

reference: https://github.com/xamarin/xamarin-macios/issues/4891


tools/apidiff/Makefile
```diff
@@ -157,32 +157,30 @@ endif
...
 
-all-local:: $(APIDIFF_DIR)/api-diff.html
+all-local:: bundle.zip $(APIDIFF_DIR)/api-diff.html
 
 # Rules to re-create the reference infos. The rules target the updated-references
 # directory, so that they're only invoked manually (otherwise the reference infos
 # would be updated every time the assemblies change, which is not what we want).
 
+bundle.zip:
+	curl $(APIDIFF_REFERENCES) > bundle.zip
```
rolfbjarne: A couple of issues:

* The *.xml files don't depend on bundle.zip, so it won't necessarily be downloaded before running mono-api-info.
* `bundle.zip` won't be re-downloaded after changing the url.
* If a previous download is incomplete (interrupted by ctrl-c for instance), it won't be automatically detected and re-downloaded.
* We can't change the url to a redirect, because curl won't follow it.

So, suggested improvements:

```make
# split the URL in words based on the path separator, and then chose the 6th word (the hash) in the bundle zip filename
BUNDLE_ZIP=bundle-$(word 6,$(subst /, ,$(APIDIFF_REFERENCES))).zip
$(BUNDLE_ZIP):
        # download to a temporary filename so interrupted downloads won't prevent re-downloads.
	$(Q_GEN) curl -L $(APIDIFF_REFERENCES) > $@.tmp
        $(Q) mv $@.tmp $@

.unzip.stamp: $(BUNDLE_ZIP)
    $(Q) rm -Rf temp
    $(Q_GEN) unzip -d temp $(BUNDLE_ZIP)
    $(Q) touch $@

# the semi-colon at the end means an empty recipe, and is required for make to consider pattern rules
temp/%.dll: .unzip.stamp ;
```

PR #5668, opened 2019-02-26
[XHarness] Remove duplicated runs of certain bcl tests.

mandel-macaque:
Most of the tests are comming from the downloaded SDK, but some of them
have to remain because they are running/passing more tests. This
differences are due to how the tests that come from mono are written
(missing resources etc..)

Please take a look at
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/794210/ for the
list of bcl tests and test number differences.


mandel-macaque:
@rolfbjarne that is the plan, but in a different PRs, since it is removing tests I want to make changes as small as possible.

PR #5674, opened 2019-02-26
[xm][msbuild] Fix full profile FrameworkList file

VincentDondain:
- Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801867
  System.Memory.dll not being copied anywhere when Roslyn package is referenced
  The issue was that the XM FrameworkList was shared for the 2 profiles and there are differences. The Full profile for instance doesn't have `System.Memory` and that causes some issues with Nuget package references.
- Updated tests to report issues against the XM Full and Mobile Framework lists
- Now using 2 separate XM FrameworkList files (updated makefile) since the list of assemblies is different and that's expected by the mono team.


VincentDondain:
I didn't test that locally as it's still building and I have to leave my current machine. It should be fine though (:

VincentDondain:
Finally was able to try locally after building and could confirm the fix indeeds **works** and System.Memory is now **present** in both the app bundle and output directory therefore fixing https://devdiv.visualstudio.com/DevDiv/_workitems/edit/801867

VincentDondain:
@monojenkins backport d16-0

PR #5681, opened 2019-02-27
[Tests] Fix failing mtouch tests after bcl test were removed.

mandel-macaque:
After commit 2619d8f5ca1ce6c1d82eac5c796f52d18c1b8ca1 the System.Core
tests does not longer exists. Which makes the mtouch test that depends
on it, fail too. This changes the test to use a present test.


mandel-macaque:
API diff issue can be ignored.

PR #5465, opened 2019-01-23
[watch] add build rules for arm64_32 based on 2018-08-watchos branch

lewurm:
NOTE: tested with this hacky patch for xharness: https://gist.github.com/lewurm/b4243bad2b4407bdd7f056bbdbeafd5b

Verified by running mini regression suite. It's green 🙂

Context: mono/mono#10641

PR #5683, opened 2019-02-28
Fix ARM64-based CPU detection logic; take #3.

rolfbjarne:
The previous implementation was broken, because:

* We called GetIsARM64CallingConvention in Runtime's static constructor.
* Runtime.Arch hadn't been set yet in the static constructor, which means it
  was Arch.DEVICE (0).
* This meant GetIsARM64CallingConvention would return true on iOS/tvOS x86-64
  simulator.

So delay the initialization of the IsARM64CallingConvention field until we've
been initialized.

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -259,6 +259,7 @@ namespace ObjCRuntime {
 #if !XAMMAC_SYSTEM_MONO
 			UseAutoreleasePoolInThreadPool = true;
 #endif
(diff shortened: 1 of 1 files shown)


rolfbjarne:
> Did that fail an existing* unit test ?

Kind of: monotouch-test started crashing when I put the previous fix in the branch with my other arm64_32 changes (PR #5465).

rolfbjarne:
API diff failure is due to https://github.com/xamarin/maccore/issues/1452

PR #5684, opened 2019-02-28
[Tests] Make test inconclusive if httpbin does not return the headers.

mandel-macaque:
In some cases, httpbin might return errors or other results. In that
case, we cannot assert that the auth headers are not present. We set the
test to inconclusive to ensure that we do not fail due to the external
page.

Fixes https://github.com/xamarin/maccore/issues/1453


mandel-macaque:
API diff issues are known: https://github.com/xamarin/maccore/issues/1452

PR #5685, opened 2019-02-28
Improve MachO support to read LC_BUILD_VERSION and static libraries.

rolfbjarne:
* Add support for reading LC_BUILD_VERSION. This is needed to properly process
  Mach-O files with a recent minOS version.
* Add support for reading Mach-O files inside ar archives.
* Augment ProductTests.MinOSVersion to test static libraries as well.

### DIFF
--- a/tools/common/MachO.cs
+++ b/tools/common/MachO.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net;
+using System.Text;
 
 #if MLAUNCH
 using Xamarin.Launcher;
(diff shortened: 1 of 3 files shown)


tools/common/MachO.cs
```diff
@@ -133,8 +134,18 @@ public enum LoadCommands : uint
...
 
+		public enum Platform : uint {
+			PLATFORM_MACOS = 1,
+			PLATFORM_IOS = 2,
+			PLATFORM_TVOS = 3,
+			PLATFORM_WATCHOS = 4,
+			PLATFORM_BRIDGEOS = 5,
+			PLATFORM_IOSSIMULATOR = 7,
+			PLATFORM_TVOSSIMULATOR = 8,
+			PLATFORM_WATCHOSSIMULATOR = 9,
```
spouliot: minor (we don't ship it) but the `PLATFORM*` prefix and casing is not the one we normally use.

rolfbjarne:
API comparison failure is due to https://github.com/xamarin/maccore/issues/1452.

PR #5690, opened 2019-02-28
[apidiff] Fix the comparison with the previous commit

spouliot:
The removal of the XML files broke the comparison with the previous
commit. It did NOT fail the original PR because the targets were
in the previous commit.

And it will fail this PR because the previous commit still does not
have the targets - but it _should_ be fine, once merged, for all
future PR.

Fixes https://github.com/xamarin/maccore/issues/1452


spouliot:
Failure was expected, next commit should be fine

PR #5695, opened 2019-03-01
[XHarness] Fix makefile to take into account the importer sources.

mandel-macaque:
Ensure that when the importer sources are changed, xharness is
recompiled.


mandel-macaque:
API diff failures can be ignored.

PR #5706, opened 2019-03-03
Fix build failure when dsym generation is disabled

msioen:
If your iOS project has the -dsym=false modifier in the MTouchExtraArgs the build fails during the GenerateDebugSymbols step because it tries to touch a file that doesn't exist.

=> added an additional Exists step to ensure the build doesn't fail on this


spouliot:
build

PR #5709, opened 2019-03-04
[Xharness] Reenable the xunit based System.Data tests on mac os x.

mandel-macaque:
Fixes https://github.com/xamarin/maccore/issues/1208


mandel-macaque:
WatchOS crash is a known issue: https://github.com/xamarin/maccore/issues/581

PR #5712, opened 2019-03-04
[corefoundation] Add missing CFStringTransform API

spouliot:
Since there's not NSString or .NET equivalent. This requires a mutable
`CFString` so the new type is also added.

Draft until unit tests are added.

### DIFF
--- a/src/CoreFoundation/CFMutableString.cs
+++ b/src/CoreFoundation/CFMutableString.cs
@@ -1,39 +1,60 @@
+// Copyright 2019 Microsoft Corporation
+
 #if !COREBUILD
 
 using System;
 using System.Runtime.InteropServices;
 
+using Foundation;
 using ObjCRuntime;
 
 namespace CoreFoundation {
 	
 	public class CFMutableString : CFString {
 
-		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
-		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutableWithExternalCharactersNoCopy (/* CFAllocatorRef* */ IntPtr alloc, string chars, nint numChars, nint capacity, /* CFAllocatorRef* */ IntPtr externalCharactersAllocator);
-
-		public CFMutableString (string theString, nint capacity)
+		public CFMutableString (IntPtr handle)
+			: this (handle, false)
 		{
-			if (theString == null)
-				throw new ArgumentNullException (nameof (theString));
-			
-			handle = CFStringCreateMutableWithExternalCharactersNoCopy (IntPtr.Zero, theString, theString.Length, capacity, IntPtr.Zero);
+		}
+		
+		[Preserve (Conditional = true)]
+		protected CFMutableString (IntPtr handle, bool owns)
+			: base (handle, owns)
+		{
 		}
 
 		[DllImport (Constants.CoreFoundationLibrary)]
 		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutable (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength);
 
-		public CFMutableString (nint maxLength)
+		public CFMutableString (string @string = "", nint maxLength = default (nint))
 		{
-			handle = CFStringCreateMutable (IntPtr.Zero, maxLength);
+			// not really needed - but it's consistant with other .ctor
+			if (@string == null)
+				throw new ArgumentNullException (nameof (@string));
+			Handle = CFStringCreateMutable (IntPtr.Zero, maxLength);
+			if (@string != null)
+				CFStringAppendCharacters (Handle, @string, @string.Length);
 		}
 
 		[DllImport (Constants.CoreFoundationLibrary)]
 		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutableCopy (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength, /* CFStringRef* */ IntPtr theString);
 
-		public CFMutableString (CFString theString, nint maxLength)
+		public CFMutableString (CFString theString, nint maxLength = default (nint))
 		{
-			handle = CFStringCreateMutableCopy (IntPtr.Zero, maxLength, theString.GetHandle ());
+			if (theString == null)
+				throw new ArgumentNullException (nameof (theString));
+			Handle = CFStringCreateMutableCopy (IntPtr.Zero, maxLength, theString.GetHandle ());
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
+		static extern unsafe void CFStringAppendCharacters (/* CFMutableStringRef* */ IntPtr theString, string chars, nint numChars);
+
+		public void Append (string @string)
+		{
+			if (@string == null)
+				throw new ArgumentNullException (nameof (@string));
+			str = null; // destroy any cached value
+			CFStringAppendCharacters (Handle, @string, @string.Length);
 		}
 
 		[DllImport (Constants.CoreFoundationLibrary)]
@@ -42,20 +63,70 @@ namespace CoreFoundation {
 
 		public bool Transform (ref CFRange range, CFStringTransform transform, bool reverse)
 		{
-			str = null; // destroy any cached value
-			return Transform (Handle, ref range, transform.GetConstant ().GetHandle (), reverse);
+			return Transform (ref range, transform.GetConstant ().GetHandle (), reverse);
 		}
 
 		// constant documentation mention it also accept any ICT transform
 		public bool Transform (ref CFRange range, CFString transform, bool reverse)
 		{
-			str = null; // destroy any cached value
-			return CFStringTransform (Handle, ref range, transform.GetHandle (), reverse);
+			return Transform (ref range, transform.GetHandle (), reverse);
 		}
 
-		internal bool Transform (IntPtr @string, ref CFRange range, IntPtr transform, bool reverse)
+		public bool Transform (ref CFRange range, NSString transform, bool reverse)
 		{
-			return CFStringTransform (@string, ref range, transform, reverse);
+			return Transform (ref range, transform.GetHandle (), reverse);
+		}
+
+		public bool Transform (ref CFRange range, string transform, bool reverse)
+		{
+			var t = NSString.CreateNative (transform);
+			var ret = Transform (ref range, t, reverse);
+			NSString.ReleaseNative (t);
+			return ret;
+		}
+
+		bool Transform (ref CFRange range, IntPtr transform, bool reverse)
+		{
+			if (transform == IntPtr.Zero)
+				throw new ArgumentNullException (nameof (transform));
+			str = null; // destroy any cached value
+			return CFStringTransform (Handle, ref range, transform, reverse);
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		static extern internal bool /* Boolean */ CFStringTransform (/* CFMutableStringRef* */ IntPtr @string, /* CFRange* */ IntPtr range, /* CFStringRef* */ IntPtr transform, [MarshalAs (UnmanagedType.I1)] /* Boolean */ bool reverse);
+
+		public bool Transform (CFStringTransform transform, bool reverse)
+		{
+			return Transform (transform.GetConstant ().GetHandle (), reverse);
+		}
+
+		// constant documentation mention it also accept any ICT transform
+		public bool Transform (CFString transform, bool reverse)
+		{
+			return Transform (transform.GetHandle (), reverse);
+		}
+
+		public bool Transform (NSString transform, bool reverse)
+		{
+			return Transform (transform.GetHandle (), reverse);
+		}
+
+		public bool Transform (string transform, bool reverse)
+		{
+			var t = NSString.CreateNative (transform);
+			var ret = Transform (t, reverse);
+			NSString.ReleaseNative (t);
+			return ret;
+		}
+
+		bool Transform (IntPtr transform, bool reverse)
+		{
+			if (transform == IntPtr.Zero)
+				throw new ArgumentNullException (nameof (transform));
+			str = null; // destroy any cached value
+			return CFStringTransform (Handle, IntPtr.Zero, transform, reverse);
 		}
 	}
 }
--- a/src/CoreFoundation/CFMutableString.cs
+++ b/src/CoreFoundation/CFMutableString.cs
@@ -12,7 +12,7 @@ namespace CoreFoundation {
 	
 	public class CFMutableString : CFString {
 
-		public CFMutableString (IntPtr handle)
+		protected CFMutableString (IntPtr handle)
 			: this (handle, false)
 		{
 		}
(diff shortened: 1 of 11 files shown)


src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+		static extern internal bool /* Boolean */ CFStringTransform (/* CFMutableStringRef* */ IntPtr @string, /* CFRange* */ ref CFRange range, /* CFStringRef* */ IntPtr transform, [MarshalAs (UnmanagedType.I1)] /* Boolean */ bool reverse);
+
+		public bool Transform (ref CFRange range, CFStringTransform transform, bool reverse)
+		{
+			str = null; // destroy any cached value
+			return Transform (Handle, ref range, transform.GetConstant ().GetHandle (), reverse);
+		}
+
+		// constant documentation mention it also accept any ICT transform
+		public bool Transform (ref CFRange range, CFString transform, bool reverse)
```
dalexsoto: What about this API to ease the ICT stuff? `NSString` and `CFString` are toll-free bridged.

```csharp
public bool Transform (ref CFRange range, string transform, bool reverse)
{
	str = null; // destroy any cached value
	var t = NSString.CreateNative (tagClass);
	var ret = CFStringTransform (Handle, ref range, t, reverse);
	NSString.ReleaseNative (t);
	return ret;
}
```

dalexsoto:
Looks good so far 👍

src/CoreFoundation/CFMutableString.cs
spouliot: I'd need to find real world examples first (for non-constants). It's never 100% clear if the comparison is done on strings (if should) or addresses (should not, but it happens).

Still the next version will be different and use`NSString` because that's how smart enums are exposed :-)

And from here it's not hard to use type`(NSString) String.Empty` if you have a `string` value.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutableWithExternalCharactersNoCopy (/* CFAllocatorRef* */ IntPtr alloc, string chars, nint numChars, nint capacity, /* CFAllocatorRef* */ IntPtr externalCharactersAllocator);
+
+		public CFMutableString (string theString, nint capacity)
+		{
+			if (theString == null)
+				throw new ArgumentNullException (nameof (theString));
+			
+			handle = CFStringCreateMutableWithExternalCharactersNoCopy (IntPtr.Zero, theString, theString.Length, capacity, IntPtr.Zero);
```
rolfbjarne: Is this really the native function you want to use? `This function permits you to create a CFMutableString object whose backing store is an external Unicode character buffer—that is, a buffer that you control (or can control) entirely.` It seems that if we want to use this function we'll have to copy the string to a buffer we control.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutable (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength);
+
+		public CFMutableString (nint maxLength)
```
rolfbjarne: Default value? `nint maxLength = 0`

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+		{
+			handle = CFStringCreateMutable (IntPtr.Zero, maxLength);
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutableCopy (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength, /* CFStringRef* */ IntPtr theString);
+
+		public CFMutableString (CFString theString, nint maxLength)
```
rolfbjarne: Default value? `nint maxLength = 0`

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+		{
+			handle = CFStringCreateMutableCopy (IntPtr.Zero, maxLength, theString.GetHandle ());
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		static extern internal bool /* Boolean */ CFStringTransform (/* CFMutableStringRef* */ IntPtr @string, /* CFRange* */ ref CFRange range, /* CFStringRef* */ IntPtr transform, [MarshalAs (UnmanagedType.I1)] /* Boolean */ bool reverse);
+
+		public bool Transform (ref CFRange range, CFStringTransform transform, bool reverse)
```
rolfbjarne: `range` can be NULL, so maybe add an overload that doesn't take a `range`.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,63 @@
...
+			
+			handle = CFStringCreateMutableWithExternalCharactersNoCopy (IntPtr.Zero, theString, theString.Length, capacity, IntPtr.Zero);
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutable (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength);
+
+		public CFMutableString (nint maxLength)
+		{
+			handle = CFStringCreateMutable (IntPtr.Zero, maxLength);
```
rolfbjarne: Maybe throw an exception if `handle` ends up being IntPtr.Zero?

src/CoreFoundation/CFMutableString.cs
spouliot: I tried that first, but does not work with our `nint` :(
Maybe I'll use `int` and typecast, not ideal :|

src/CoreFoundation/CFMutableString.cs
spouliot: good idea, I did not like a `ref CFRange?`

src/CoreFoundation/CFMutableString.cs
rolfbjarne: You can do this: `nint maxLength = default (nint)`

src/CoreFoundation/CFMutableString.cs
spouliot: I need to test what happens if you allocate a `nil` string and append to it
It's also something that we should do (and share) on the base `CFString` type

src/CoreFoundation/CFMutableString.cs
spouliot: missed that, I'll see what's easier (make CFString mutable or copying)

src/CoreFoundation/CFMutableString.cs
rolfbjarne: CFString should inherit from NativeObject, then we'd get much of this logic for free :smile:

src/CoreFoundation/CFMutableString.cs
spouliot: oh, that's much nicer :) thanks!

src/CoreFoundation/CFMutableString.cs
spouliot: yeah, it's getting farther and father away from the single missing API
but I guess it's a good time to bite the bullet on this one :)

src/CoreFoundation/CFMutableString.cs
spouliot: done

src/CoreFoundation/CFMutableString.cs
spouliot: done

src/CoreFoundation/CFMutableString.cs
spouliot: done

src/CoreFoundation/CFMutableString.cs
spouliot: done

src/CoreFoundation/CFMutableString.cs
spouliot: removed

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+using Foundation;
+using ObjCRuntime;
+
+namespace CoreFoundation {
+	
+	public class CFMutableString : CFString {
```
rolfbjarne: `sealed`?

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+
+		public CFMutableString (CFString theString, nint maxLength = default (nint))
+		{
+			if (theString == null)
+				throw new ArgumentNullException (nameof (theString));
+			Handle = CFStringCreateMutableCopy (IntPtr.Zero, maxLength, theString.GetHandle ());
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
+		static extern unsafe void CFStringAppendCharacters (/* CFMutableStringRef* */ IntPtr theString, string chars, nint numChars);
```
rolfbjarne: `unsafe` isn't needed.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+				throw new ArgumentNullException (nameof (@string));
+			Handle = CFStringCreateMutable (IntPtr.Zero, maxLength);
+			if (@string != null)
+				CFStringAppendCharacters (Handle, @string, @string.Length);
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutableCopy (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength, /* CFStringRef* */ IntPtr theString);
+
+		public CFMutableString (CFString theString, nint maxLength = default (nint))
```
rolfbjarne: `theString` -> `string` to match all the other API.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+		[Preserve (Conditional = true)]
+		protected CFMutableString (IntPtr handle, bool owns)
+			: base (handle, owns)
+		{
+		}
+
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern unsafe /* CFMutableStringRef* */ IntPtr CFStringCreateMutable (/* CFAllocatorRef* */ IntPtr alloc, nint maxLength);
+
+		public CFMutableString (string @string = "", nint maxLength = default (nint))
```
rolfbjarne: What happens if `maxLength < 0`?

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+
+		public bool Transform (ref CFRange range, NSString transform, bool reverse)
+		{
+			return Transform (ref range, transform.GetHandle (), reverse);
+		}
+
+		public bool Transform (ref CFRange range, string transform, bool reverse)
+		{
+			var t = NSString.CreateNative (transform);
+			var ret = Transform (ref range, t, reverse);
```
rolfbjarne:
```csharp
try {
    return Transform (ref range, t, reverse);
} finally {
    NSString.ReleaseNative (t);
}
```

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+		public bool Transform (NSString transform, bool reverse)
+		{
+			return Transform (transform.GetHandle (), reverse);
+		}
+
+		public bool Transform (string transform, bool reverse)
+		{
+			var t = NSString.CreateNative (transform);
+			var ret = Transform (t, reverse);
+			NSString.ReleaseNative (t);
```
rolfbjarne: Same: try/finally.

tests/monotouch-test/CoreFoundation/MutableString.cs
```diff
@@ -0,0 +1,124 @@
...
+				Assert.That (s.Handle, Is.Not.EqualTo (IntPtr.Zero), "Handle");
+				Assert.That (s.ToString (), Is.Empty, "ToString");
+			}
+		}
+
+		[Test]
+		public void CreateString1 ()
+		{
+			Assert.Throws<ArgumentNullException> (() => new CFMutableString ((string)null), "null");
+			using (var s = new CFMutableString ("bonjour!")) {
```
rolfbjarne: More complex test strings would be nice too: emojis, bidi characters, unicode characters with multiple code points, etc.

src/CoreFoundation/CFMutableString.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Runtime.InteropServices;
+
+using Foundation;
+using ObjCRuntime;
+
+namespace CoreFoundation {
+	
+	public class CFMutableString : CFString {
+
+		public CFMutableString (IntPtr handle)
```
rolfbjarne: `internal`? I know the CFString version is public, but that doesn't mean we have to add new public ones.

src/CoreFoundation/CFMutableString.cs
spouliot: That's the same names as the native API, which makes it easier to understand (when the documentation merger is used).

src/CoreFoundation/CFMutableString.cs
spouliot: why ? it's not common in our bindings (for direct, public bindings) and the linker can do that optimization.

src/CoreFoundation/CFMutableString.cs
spouliot: worth a test

src/CoreFoundation/CFMutableString.cs
spouliot: changing to `protected` like the 2 argument `.ctor`

The swift bindings do not seems to make to limit subclassing this type - and I don't see why we should impose our own arbitrary limits (there's enough _non-arbitrary _ of them already)

src/CoreFoundation/CFMutableString.cs
spouliot: good point, `Transform` can now throw

tests/monotouch-test/CoreFoundation/MutableString.cs
spouliot: Tests are to validate the bindings, not the API. That historically caused a lot of troubles (false positives) when run on devices with different OS versions or installed/default languages.

Is there something specific you think would cause problems ?

spouliot:
bot/network issue
```
09:05:37  > git submodule update --init --recursive external/mono # timeout=30
09:35:37 ERROR: Timeout after 30 minutes
09:35:37 Also:   hudson.remoting.Channel$CallSiteStackTrace: Remote call to JNLP4-connect connection from 167.220.149.94/167.220.149.94:15339
```

spouliot:
build

tests/monotouch-test/CoreFoundation/MutableString.cs
rolfbjarne: Yes, I'd like to confirm if Mono's understanding of Unicode is the same as Apple's, to ensure that a roundtrip string -> CFString -> string doesn't end up with something corrupted.

For instance: `string.Length` is "the number of Char objects in this instance, not the number of Unicode characters.", while `CFStringAppendCharacters` takes `numChars:
The number of Unicode characters in chars`. I don't trust documentation, but it seems there is a possibility that `CFStringAppendCharacters` can go off into random memory looking for more characters than we actually provided.

Unicode is hard 😒

src/CoreFoundation/CFMutableString.cs
rolfbjarne: It seems that subclassing CFMutableString is something nobody should ever do, and besides it's always possible unseal it later if we want to.

spouliot: Swift does not seal it and I can't see any positive outcome out of sealing it. At best it does nothing (linker will seal it), at worse someone will need that (for good or bad reasons) and it will takes weeks/months to get a fix out.

spouliot:

PR #5720, opened 2019-03-05
[Tests] Fix the tests.sln to include the new test projects.

mandel-macaque:
Update the test.sln file:

* Remove old not present projects.
* Add new test projects.

This required updating the test project generator to no use the same GUID for all projects. This is done generating the GUID based on the MD5 hash of the project name. This will always generate the same GUID for the project.


tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -390,6 +389,16 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
 			}
 		}
 
+		internal static string GetProjectGuid (string projectName) 
+		{
+			//use MD5 hash to get a 16-byte hash of the string: 
+			using (var provider = new MD5CryptoServiceProvider ()) {
+				var inputBytes = Encoding.Default.GetBytes (projectName);
+				var hashBytes = provider.ComputeHash (inputBytes);
+				return new Guid (hashBytes).ToString ().ToUpper ();
```
spouliot: minor (since it's not in the platform assembly) but `ToUpperInvariant` is better

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -390,6 +389,16 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
 			}
 		}
 
+		internal static string GetProjectGuid (string projectName) 
+		{
+			//use MD5 hash to get a 16-byte hash of the string: 
+			using (var provider = new MD5CryptoServiceProvider ()) {
+				var inputBytes = Encoding.Default.GetBytes (projectName);
```
spouliot: minor (since it's not in the platform assembly) but `Default` can change, using `UTF8` and would satisfy the need

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -390,6 +389,16 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
 			}
 		}
 
+		internal static string GetProjectGuid (string projectName) 
+		{
+			//use MD5 hash to get a 16-byte hash of the string: 
+			using (var provider = new MD5CryptoServiceProvider ()) {
```
spouliot: minor (since it's not in the platform assembly) but `MD5.Create ()` is better (since it will pick up the best implementation of MD5 available).

mandel-macaque:
Meh, network issues cloning in jenkins :/

mandel-macaque:
build

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -725,6 +734,7 @@ async Task<string> GenerateAsync (string projectName, string registerPath, (stri
 			using (var reader = new StreamReader(templatePath)) {
 				var result = await reader.ReadToEndAsync ();
 				result = result.Replace (DownloadPathKey, downloadPath);
+				result = result.Replace (ProjectGuidKey, GetProjectGuid (projectName));
```
rolfbjarne: There's already `Harness.GetStableGuid`: https://github.com/xamarin/xamarin-macios/blob/9168a63698aaeb3d32488a32b4f4168c11cebd00/tests/xharness/Harness.cs#L747-L758

mandel-macaque: @rolfbjarne is that method going to generate the Guid based on the project name. I fear that this method will depend on the order of the call, is that correct?

rolfbjarne: Yes, `NewStableGuid` will return different values in subsequent calls. However, it will always return the same value for call #X (since the initial random seed is fixed). If csproj generation is always done in the same order, the returned GUIDs won't change.

If you really want to use the project name as input to get a stable GUID, I think you should augment the existing `NewStableGuid` method (maybe add a `string seed = null` argument, which uses your implementation if provided).

mandel-macaque: @rolfbjarne I prefer to be able to pass the seed, since I'm sure people will edit the project definitions more ofter than other parts.. I might need to do some refactoring since the importer code does not know about Harness, but I prefer to reuse code.

mandel-macaque:
WatchOS crashes can be ignored.

mandel-macaque:

PR #5721, opened 2019-03-05
[apidiff] Remove the `_*` noise in XM diffs

spouliot:
I was, kind of, hoping using binaries from the bots would _indirectly_
solve it - but it did not.

Only the XM profile reports _inexistent_ changes in `_Attribute`,
`_EventInfo` and `_Exception`. That noise makes it harder to review
the diffs, so we now ignore those specific cases.

### DIFF
--- a/tools/apidiff/Makefile
+++ b/tools/apidiff/Makefile
@@ -46,6 +46,8 @@ TVOS_ASSEMBLIES    = $(foreach file,$(MONO_ASSEMBLIES),Xamarin.TVOS/$(file))    
 IOS_ARCH_ASSEMBLIES = native-32/Xamarin.iOS native-64/Xamarin.iOS
 MAC_ARCH_ASSEMBLIES = native-32/Xamarin.Mac native-64/Xamarin.Mac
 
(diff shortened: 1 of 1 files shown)


spouliot:
Depends on how you define the problem(s)
- customer wise our release notes are inaccurate for XM
- extra noise made it harder to review (and more likely to hide other mistakes)

When I looked at why, a while ago, I did not find the difference with XI (both do the same thing) and I _guessed_ it had to do with XM full-vs-modern (or a different mono - but that's excluded now that everything is done on the same bots)

rolfbjarne:
The problem being noise about `_Exception` and other interfaces in the api diff, and those are still present in the diff:

screenshot 2019-03-05 17 52 51

spouliot:
Not quite, the diffs we publish (from stable) don't have them anymore.
It's the _from previous commit_ that is using another script that still shows them. I'll look at that one too (since it affects reviews)

spouliot:
or it's another _using the previous commit_ script that created the delay...
modifying that part of the build requires a time traveller diploma :)

spouliot:
The _(from PR only)_ should work fine🤞(did locally) once merged (at least when another commit is made, since it goes back in time)

spouliot:

PR #5729, opened 2019-03-06
[uikit] Simplify UIApplication.Main wrt NSString usage

spouliot:
The lifespan of `new NSString (x).Handle` can be shorter than expected
and there's a better (lighter) alternative to get that handle


spouliot:

PR #5730, opened 2019-03-06
Bump XI and XM versions after branching for d16-1

spouliot:


spouliot:
failure is unrelated (to version bump) and filed as https://github.com/xamarin/maccore/issues/1466

Issue #4170, opened 2018-06-01
Certificate pinning using NSUrlSessionHandler


Cheesebaron:
### Steps to Reproduce

1. Attempt to implement certificate pinning for NSUrlSessionHandler
2. ??
3. Cry 😭 

### Expected Behavior
I expected to be able to implement a delegate, such as NSUrlSessionDataDelegate, or event like `ServicePointManager.ServerCertificateValidationCallback` to check the certificate. Both for trusting self signed certificates and for pinning certificates. I am mostly interested in the latter.

### Actual Behavior
[NSUrlSessionHandlerDelegate][nsurlshd] is private and contains a bunch of the necessary logic for NSUrlSessionHandler to even function.

[NSUrlSession][nsurlsession] which uses the delegate is a private readonly member variable, so it cannot easily be set either.

The only way to implement this for NSUrlSessionHandler is to yank the code from this repository and modify it. Which isn't really an optimal solution.

It is a similar story with Android, no out of the box hooks into the events for validating certificates, just like you get with the managed implementation and ServicePointManager.

### Environment

```
=== Visual Studio Enterprise 2017 for Mac ===

Version 7.5.1 (build 22)
Installation UUID: f274f3d1-0b76-4491-bdf0-60aa84226638
Runtime:
	Mono 5.10.1.47 (2017-12/8eb8f7d5e74) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 4.4.1.178 (master / eeaeb7e6)

	Package version: 510010047

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Version: 2.0.5
SDK: /usr/local/share/dotnet/sdk/2.1.4/Sdks
SDK Version: 2.1.4
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.10.1/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.2
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Xamarin.Android ===

Version: 8.3.0.19 (Visual Studio Enterprise)
Android SDK: /Users/tomasci/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		6.0 (API level 23)
		7.0 (API level 24)
		7.1 (API level 25)
[...]

PR #5733, opened 2019-03-06
[Foundation] Add support to the ServerCertificateValidationCallback on the NSUrlSessionHandler

mandel-macaque:
This changes allows to use the service point manager callback within the
NSUrlSessionHanlder.

The logic is as follows:

1. Check that the challange is about the ssl cert.
2. Convert the SecTrust info to the managed objects.
3. Call the users cb.
4. If user returns true, continue, else, cancel and create an exception
similar to the one used in the managed code.

Fixes https://github.com/xamarin/xamarin-macios/issues/4170

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -33,10 +33,12 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Security;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Text;
+using System.Security.Cryptography.X509Certificates;
 
 #if UNIFIED
 using CoreFoundation;
@@ -451,7 +453,9 @@ namespace Foundation {
 					if (error != null) {
 						inflight.Errored = true;
 
-						var exc = createExceptionForNSError (error);
+						// if the exception was set by one of the delefate methods, used it, else use the one
+						// from the error.
+						var exc = inflight.Exception ?? createExceptionForNSError (error);
 						inflight.CompletionSource.TrySetException (exc);
 						inflight.Stream.TrySetException (exc);
 					} else {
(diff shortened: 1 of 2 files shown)


mandel-macaque:
Awesome feedback! Thanks for or the concerns!

mandel-macaque:
@baulig I don't mind you taking over the implementation, but if you are happy with it, I would not mind taking a look on how to hook it up to the new way. I used the ServicePointManager because it was mentioned in the issue, I am more than happy to be ready to the new way to do things, and the more code we reuse the better.

@spouliot I was expecting something of the kind, I created the PR to start the conversation and try to move this forward.. The only issues we have to be worried about are the current way in which we are dealing with the authentication. But this can be quite simple since we just need to check the ProtectionSpace.AuthenticationMethod and branch there. We can expose the SecTrust to the users, and then do the correct thing with the completion handler. So we can either expose a callback to the NSUrlSessionHandler (similar to what is done). But I would wait for that approach until we see what @baulig  is planning, since having a central way to do it and shared between the implementations would me a nicer approach, meaning that the users won't have to add too much custom code per implementation.

PS: I should have used a draft PR, I keep forgetting about that feature, again, I wanted to get the conversation started so that users can see that we do have planes for this.

baulig:
Let me clarify one thing here about what is meant by "the new code".

There are two pieces to it: the code in the `System.Net.Http` and `System.Net.Http.Headers` namespace that we're switching over to the CoreFX implementation - and the actual `SocketsHttpHandler`.

All profiles will receive the former (so that we could completely nuke the `System.Net.Http` and `System.Net.Http.Headers` directories in `mcs/class/System.Net.Http`) - but you can still use the native handlers _without_ the sockets handler (which can either be removed by the linker or "physically" omitted in the first place like on web assembly).

As part of that work, I will go over all the native handlers (like `CFNetworkHandler`, `NSUrlSessionHandler` etc.) and hook up the new certificate validator.  Let's also keep in mind that the new sockets handler does not have any dependencies on the web stack anymore - so ideally, the linker should be able to completely eliminate the web stack.

In addition to what @spouliot pointed out regarding `X509Certificate2`, this is also an expensive operation if the `X509Certificate2` instance is not properly constructed from a native handle.

@mandel-macaque No worries, I have already started work on this last week, but then got distracted by build issues in the `mono-2019-02` branch (related to `Mono.Native`).  There is also a draft PR: https://github.com/mono/mono/pull/13339.

spouliot:
Superseded by https://github.com/xamarin/xamarin-macios/pull/6103

PR #5752, opened 2019-03-08
[XHarness] Fix the hostname parsing on devices.

mandel-macaque:
Following this
http://xamarin-storage/jenkins/xamarin-macios/mono-2018-10/221f3e4a375d565d24087afbfcb1d6bc824c1ca9/2478649/device-tests/jenkins-results/tests/index.html
we need to ensure that we have a split to get the host. Just happens on
iOs and tvOS since they are the only ones using tcp.


mandel-macaque:

PR #5761, opened 2019-03-13
[XHarness] Add an extra configuration option for the tests matrix to match those in QA.

mandel-macaque:


tests/xharness/Jenkins.cs
```diff
@@ -209,6 +210,8 @@ IEnumerable<TestData> GetTestData (RunTestTask test)
 				yield return new TestData { Variation = "AssemblyBuildTarget: SDK framework (debug, profiling)", MTouchExtraArgs = "--assembly-build-target=@sdk=framework=Xamarin.Sdk --assembly-build-target=@all=staticobject", Debug = true, Profiling = true, MonoNativeLinkMode = MonoNativeLinkMode.Static, MonoNativeFlavor = flavor };
 
 				yield return new TestData { Variation = "Release", MTouchExtraArgs = "", Debug = false, Profiling = false, MonoNativeLinkMode = MonoNativeLinkMode.Static };
+				yield return new TestData { Variation = "Release: SGenConc", MTouchExtraArgs = "", Debug = false, Profiling = false, MonoNativeLinkMode = MonoNativeLinkMode.Static, EnableSGenConc = true};
```
rolfbjarne: Do the QA tests try all projects with the concurrent GC, or just some of them?

mandel-macaque: It is tested in the following one:


- mini
- monotouch-test
- don’t link monotouch-test
- link sdk monotouch-test
- link all monotouch-test

- mscorlib
- System
- System.Core
- System.ServiceModel
- System.WebServices
- System.Runtime.Serialization
- System.Transactions
- System.Xml
- System.Xml.Linq
- Mono.Security
- Mono.Data.SqlLite
- System.Json
- System.Component.ModelDataAnnotations
- Mono.CSharp
- Mono.Data.Tds
- System.IO.Compression
- System.IO.Compression.FileSystem

And they do not make a diff in the tvOS Smoke + BCL Tests

tests/xharness/Jenkins.cs
```diff
@@ -209,6 +210,8 @@ IEnumerable<TestData> GetTestData (RunTestTask test)
...
+				yield return new TestData { Variation = "Debug: SGenConc", MTouchExtraArgs = "", Debug = true, Profiling = false, MonoNativeLinkMode = MonoNativeLinkMode.Static, EnableSGenConc = true};
```
spouliot: I don't think there's much (if any) interaction between Debug and Release vs sgen in concurrent mode.

Beside the non-trivial extra time to add both, it makes the test run more susceptible to random errors - leading to spending more (manual) time reviewing and green'ifying builds.

mandel-macaque: @spouliot I'm trying to match the matrix tests in QA, if we are ok of not adding it because we do not need it, I'm more than happy to drop it and we document the reason :)

@GouriKumari any feedback on this?

spouliot: @mandel-macaque it's not about adding it (or not) to the matrix - it's adding it twice when there's nothing suggesting it would give use more than 1x more coverage. That does not scale - no pros (e.g. extra issues found) & lots of cons (e.g. manual work, delay in results)

GouriKumari: In my opinion, enabling sgen concurrent with Debug will suffice. I haven't found any additional issues or differences in results recently by adding this configuration. These configs were added when the sgen concurrent were not public to users.

mandel-macaque: @spouliot @GouriKumari ok, so we have an agreement, I'll remove the extra one

VincentDondain:
Like maybe `monotouch-test` without no concurrent GC and call it a day? How do people feel about that?

mandel-macaque:
I have no strong opinion. As long as we do test both options in those tests that we fill we give the biggest return for the run, I am happy.

@spouliot @rolfbjarne I'll let you choose :)

rolfbjarne:
I think it would be enough to run these tests with the concurrent GC:

* monotouch-test
* mini
* mscorlib

and only in Debug.

mandel-macaque:
@rolfbjarne I'll do the appropriate changes then

mandel-macaque:
build

PR #5762, opened 2019-03-13
[XHarness] Fix crash in the bcl tests.

mandel-macaque:


mandel-macaque:
Watch crash is known: https://github.com/xamarin/maccore/issues/581

PR #5763, opened 2019-03-13
[XHarness] Add an extra configuration option for the tests matrix to match those in QA. (Thumb)

mandel-macaque:


tests/xharness/Jenkins.cs
```diff
@@ -209,6 +210,7 @@ IEnumerable<TestData> GetTestData (RunTestTask test)
 				yield return new TestData { Variation = "AssemblyBuildTarget: SDK framework (debug, profiling)", MTouchExtraArgs = "--assembly-build-target=@sdk=framework=Xamarin.Sdk --assembly-build-target=@all=staticobject", Debug = true, Profiling = true, MonoNativeLinkMode = MonoNativeLinkMode.Static, MonoNativeFlavor = flavor };
 
 				yield return new TestData { Variation = "Release", MTouchExtraArgs = "", Debug = false, Profiling = false, MonoNativeLinkMode = MonoNativeLinkMode.Static };
+				yield return new TestData { Variation = "Release: UseThum", MTouchExtraArgs = "", Debug = false, Profiling = false, MonoNativeLinkMode = MonoNativeLinkMode.Static, UseThumb = true };
```
rolfbjarne: I think it makes more sense to add the 32-bit check here:

```csharp
if (test.Platform == TestPlatform.iOS_Unified32)
    yield return ...
```

that way we keep the test data logic closer together.

spouliot: typo `UseThumb` (missing `b`)

PR #5768, opened 2019-03-14
[XHarness] Reduce the number of test applications for the BCL tests.

mandel-macaque:
This commit improves the state of the BCL testing in the following ways:

1. Improve the device tests running. Less apps, faster results.
2. Reduces the chances of getting issue
    https://github.com/xamarin/maccore/issues/581 when running the tests
    since we have less apps to run.

We now support the ignore files per assembly name to simplify the
tracking of the ignored tests.


mandel-macaque:
This moves to have just 4 apps for ALL the BCL tests.

rolfbjarne:
build

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -76,44 +76,43 @@ public enum WatchAppType {
...
-			(name:"MicrosoftCSharpXunit", assemblies: new [] {"monotouch_Microsoft.CSharp_xunit-test.dll"}),
+			(name:"BCL tests group 3", assemblies: new[] { "monotouch_System.Data_xunit-test.dll",
+									"monotouch_System.Json_xunit-test.dll",
+									"monotouch_System.Xml_xunit-test.dll",
+									"monotouch_System.ComponentModel.Composition_xunit-test.dll",
+									"monotouch_System.Xml.Linq_xunit-test.dll",
+									"monotouch_System.Security_xunit-test.dll",
+									"monotouch_System.Runtime.Serialization_xunit-test.dll",
+									"monotouch_System.Runtime.CompilerServices.Unsafe_xunit-test.dll", }),
+			(name:"BCL tests group 4", assemblies: new[] { "monotouch_System.Core_xunit-test.dll",
```
rolfbjarne: A couple of really minor naming points:

* Maybe name the XUnit and Unit tests differently: `BCL tests group X (NUnit)` and `BCL tests group X (XUnit)`.
* I'd intuitively expect the most basic/important assemblies to use a lower number: that the mscorlib/System tests would be in group 1, and more peripheral assemblies in group >1.

rolfbjarne:
build

mandel-macaque:
@rolfbjarne got a nice idea on how to do it so that locally they are not grouped (better for the developer to test and debug) and group on jenkins). You might or not like it ;)

akoeplinger:
@mandel-macaque it'd be nice if this was an option, I'd like to use the grouped version locally too.

rolfbjarne:
build

mandel-macaque:
@spouliot @akoeplinger after the last push, it is an option. You can make xharness think you are on Jenkins setting the env var https://github.com/xamarin/xamarin-macios/blob/master/tests/xharness/Harness.cs#L692 else, if that env var does not have the right value, you will get one app per assembly as we used. 

I have not look much into the group naming, so ideas are welcome (simple string replacement).

akoeplinger:
@mandel-macaque that's pretty non-discoverable and reusing the Jenkins variable probably has other implications? why not add a separate option?

mandel-macaque:
@akoeplinger code/var reuse, but add a new one in the last commit: https://github.com/xamarin/xamarin-macios/pull/5768/commits/343efd77c83b3ecb4a62f0e635d558937334eed3#diff-8f44b7ce671e35190b27341716ad7e8cR699

mandel-macaque:
Looks like the combination of tests assemblies on mac os x makes the tests block. Will provide a sample to @akoeplinger 

Current split works. Also, if an assembly is ignore in a platform, it will not be included in the grouped app.

mandel-macaque:
As per las commit, iOS/tvOS and MacOS apps are grouped, watchOS are not due to small issues with the deployment to the device and their execution.

mandel-macaque:
The failing issues is known: https://github.com/xamarin/maccore/issues/619

PR #5772, opened 2019-03-14
[mtouch] Update validations around --interpreter

spouliot:
This also centralize other interpreter checks and options in the same
location (making it easier to read / update).

* Warn and switch the REPL if the interpreter is enabled on simulator

Why ? It's confusing to build the same code using different options for
simulator and devices. This is what happens if you try to use features
like `dynamic` or `System.Reflection.Emit`.

So instead of an error, we warn that the interpreter is not supported
and switch to the existing REPL mode.

The JIT remains the only option for the simulator but it allows testing
features without a device.

* Fail early if the interpreter is used on 32bits [1]

The current interpreter only works on 64 bits (so ARM64). However the
error won't be reported, back to the developer, until deployment time.

This temporary [1] fix spot the condition very early and report an error

```
error MT0099 : Internal error : The interpreter is currently only available for 64 bits.
```

instead of the current one at deploy time

```
IncorrectArchitecture: Failed to find matching arch for 32-bit Mach-O input file /private/var/installd/Library/Caches/com.apple.mobile.installd.staging/temp.tNKDlx/extracted/X.app/X
error MT1006: Could not install the application 'X.app' on the device 'Mercure': AMDeviceSecureInstallApplicationBundle returned: 0xe8000087 (kAMDIncorrectArchitectureError).

Application could not be uploaded to the device.
```

[1] https://github.com/mono/mono/issues/9871

### DIFF
--- a/tools/mtouch/mtouch.cs
+++ b/tools/mtouch/mtouch.cs
@@ -1069,7 +1069,7 @@ namespace Xamarin.Bundler
 			{ "abi=", "Comma-separated list of ABIs to target. Currently supported: armv7, armv7+llvm, armv7+llvm+thumb2, armv7s, armv7s+llvm, armv7s+llvm+thumb2, arm64, arm64+llvm, i386, x86_64", v => app.ParseAbi (v) },
 			{ "override-abi=", "Override any previous abi. Only used for testing.", v => { app.ClearAbi (); app.ParseAbi (v); }, true }, // Temporary command line arg until XS has better support for 64bit architectures.
 			{ "cxx", "Enable C++ support", v => { app.EnableCxx = true; }},
-			{ "enable-repl:", "Enable REPL support (simulator and not linking only)", v => { app.EnableRepl = ParseBool (v, "enable-repl"); }, true /* this is a hidden option until we've actually used it and made sure it works as expected */ },
+			{ "enable-repl:", "Enable REPL support. For simulator only and disabling linking is recommanded.", v => { app.EnableRepl = ParseBool (v, "enable-repl"); } },
 			{ "pie:", "Enable (default) or disable PIE (Position Independent Executable).", v => { app.EnablePie = ParseBool (v, "pie"); }},
 			{ "compiler=", "Specify the Objective-C compiler to use (valid values are gcc, g++, clang, clang++ or the full path to a GCC-compatible compiler).", v => { app.Compiler = v; }},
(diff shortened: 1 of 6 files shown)


tools/mtouch/mtouch.cs
```diff
@@ -1069,7 +1069,7 @@ static Application ParseArguments (string [] args, out Action a)
 			{ "abi=", "Comma-separated list of ABIs to target. Currently supported: armv7, armv7+llvm, armv7+llvm+thumb2, armv7s, armv7s+llvm, armv7s+llvm+thumb2, arm64, arm64+llvm, i386, x86_64", v => app.ParseAbi (v) },
 			{ "override-abi=", "Override any previous abi. Only used for testing.", v => { app.ClearAbi (); app.ParseAbi (v); }, true }, // Temporary command line arg until XS has better support for 64bit architectures.
 			{ "cxx", "Enable C++ support", v => { app.EnableCxx = true; }},
-			{ "enable-repl:", "Enable REPL support (simulator and not linking only)", v => { app.EnableRepl = ParseBool (v, "enable-repl"); }, true /* this is a hidden option until we've actually used it and made sure it works as expected */ },
+			{ "enable-repl:", "Enable REPL support. For simulator only and disabling linking is recommanded.", v => { app.EnableRepl = ParseBool (v, "enable-repl"); } },
```
rolfbjarne:
```suggestion
			{ "enable-repl:", "Enable REPL support. For simulator only and disabling linking is recommended.", v => { app.EnableRepl = ParseBool (v, "enable-repl"); } },
```

tools/mtouch/mtouch.cs
```diff
@@ -1323,10 +1323,22 @@ static int Main2 (string[] args)
...
+				// it's confusing to use different options to get a feature to work (e.g. dynamic, SRE...) on both simulator and device
+				if (app.IsSimulatorBuild) {
+					ErrorHelper.Show (ErrorHelper.CreateWarning (141, "The interpreter is not supported in the simulator. Switching to REPL which provide the same extra features on the simulator."));
+					app.UseInterpreter = false;
+				}
+
+				// FIXME: the interpreter only supports ARM64 right now
+				// temporary - without a check here the error happens when deploying
+				if (!app.IsArchEnabled (Abi.ARM64))
+					throw ErrorHelper.CreateError (99, "Internal error: The interpreter is currently only available for 64 bits.");
```
rolfbjarne: Doesn't this error trigger when passing `--interpreter` when building for the simulator?

spouliot: yup, made that one before enabling sim :)
there's no point in limiting the sim to 64 bits (since it's not interpreted)
Fix incoming...

spouliot:
build

PR #5782, opened 2019-03-18
[runtime] Use mono_array_setref instead of mono_array_set.

rolfbjarne:
Otherwise the GC won't know about the assignment, and the assigned value can
be freed if it's no longer referenced anywhere else.

This also includes an update to NSArray to make it possible to create an
NSArray from an array of INativeObjects (for the tests).

### DIFF
--- a/runtime/exports.t4
+++ b/runtime/exports.t4
@@ -192,6 +192,12 @@
 			"void *", "val"
 		),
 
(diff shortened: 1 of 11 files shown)


rolfbjarne:
@monojenkins backport d16-1

PR #5794, opened 2019-03-20
[XHarness] Make TvOS and WatchOS use their test dlls.

mandel-macaque:
Up to this commit the test dlls used for the bcl were from iOS. Since
the Mono SDK provides the dlls for both missing platforms (TvOS and
WatchOS) we can now use the correct path for the dlls.

There is a small trick to minimize the project generation, since there
is a simple string.Replace, the logic now checks the platform under
test and does:

* TvOS - Goes from monotouch_TEST_NAME to monotouch_tv_TEST_NAME
* WatchOS - Gores from monotouch_TEST_NAME to monotouch_watch_TEST_NAME

### DIFF
--- a/tools/bcl-test-importer/BCLTestImporter/BCLTestAssemblyDefinition.cs
+++ b/tools/bcl-test-importer/BCLTestImporter/BCLTestAssemblyDefinition.cs
@@ -10,8 +10,8 @@ namespace BCLTestImporter {
 		static string partialPath = "mcs/class/lib";
 		static Dictionary <Platform, string> platformPathMatch = new Dictionary <Platform, string> {
 			{Platform.iOS, "monotouch"},
-			{Platform.WatchOS, "monotouch"},
-			{Platform.TvOS, "monotouch"},
+			{Platform.WatchOS, "monotouch_watch"},
+			{Platform.TvOS, "monotouch_tv"},
 			{Platform.MacOSFull, "xammac_net_4_5"},
 			{Platform.MacOSModern, "xammac_net_4_5"},
 		};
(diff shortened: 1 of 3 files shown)


tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -550,7 +550,7 @@ bool IsIgnored (string a, Platform platform)
 				if (!projectDefinition.Validate ())
 					throw new InvalidOperationException ("xUnit and NUnit assemblies cannot be mixed in a test project.");
 				// generate the required type registration info
-				var generatedCodeDir = Path.Combine (generatedDir, projectDefinition.Name);
+				var generatedCodeDir = Path.Combine (generatedDir, projectDefinition.Name, (platform == Platform.iOS)?"ios": "tv");
```
mandel-macaque: Needed to make sure we do not step on other platform Register.cs code.

tools/bcl-test-importer/BCLTestImporter/BCLTestAssemblyDefinition.cs
```diff
@@ -31,6 +31,18 @@ public BCLTestAssemblyDefinition (string name)
...
+		{
+			switch (platform) {
+			case Platform.WatchOS:
+				return Name.Replace ("monotouch_", "monotouch_watch_");
+			case Platform.TvOS:
+				return Name.Replace ("monotouch_", "monotouch_tv_");
+			default:
+				return Name;
+			}
+		}
```
mandel-macaque: Needed so that we do not have to modify the project definitions. Use iOS as the default.

mandel-macaque:
build

mandel-macaque:
build

mandel-macaque:
build

PR #5806, opened 2019-03-25
[tests] Remove confusing values for old Xcode versions

akoeplinger:
Xcode 5 to 7 are obsolete, we don't need those values in the test config anymore.


tests/mtouch/MTouch.cs
```diff
@@ -913,10 +913,10 @@ public void MT0051 ()
 				}, "error MT0051: Xamarin.iOS .* requires Xcode 6.0 or later. The current Xcode version [(]found in /Applications/Xcode44.app/Contents/Developer[)] is 4.*");
 			}
 
-			if (Directory.Exists (Configuration.xcode5_root)) {
+			if (Directory.Exists ("/Applications/Xcode511.app/Contents/Developer")) {
 				Asserts.ThrowsPattern<TestExecutionException> (() => {
 					ExecutionHelper.Execute (TestTarget.ToolPath, "-sdkroot /Applications/Xcode511.app/Contents/Developer -sim /tmp/foo");
-				}, "error MT0051: Xamarin.iOS .* requires Xcode 6.0 or later. The current Xcode version [(]found in " + Configuration.xcode5_root + "[)] is 6.0");
+				}, "error MT0051: Xamarin.iOS .* requires Xcode 6.0 or later. The current Xcode version [(]found in /Applications/Xcode511.app/Contents/Developer[)] is 6.0");
```
akoeplinger: I left this one in since the actual test uses a hardcoded `-sdkroot /Applications/Xcode511.app/Contents/Developer` so it wasn't using the variable. Also the assert above this one uses a hardcoded `/Applications/Xcode44.app/Contents/Developer` so this makes it consistent.

akoeplinger: Actually that test doesn't even make sense since it checks for `The current Xcode version ... is 6.0` but then passes in the Xcode 5 path...

PR #5807, opened 2019-03-25
Ignore tests for devices

mandel-macaque:
There are a number of tests that fail on device. We have an issue for the corlib tests (https://github.com/mono/mono/issues/13641 ) and we do not have one for the icrosoft.CSharp_xunit-test.dll because it does not happen on 2019-02 yet we have issue 
https://github.com/mono/mono/issues/13639 that is related with that dll.


mandel-macaque:
Failure is a known issue: https://github.com/xamarin/maccore/issues/581

PR #5817, opened 2019-03-28
[Jenkins] When running tests on VSTS do specify a device type.

mandel-macaque:
We are going to split the work between different types of devices:

- iOS
- iOS Beta
- AppleTv
- AppleTv Beta

So the idea is that before we have the beta devices and the split execution, we should be able to pass the device type so that when the person doing the monitoring can state which of the devices failed.

We add a default value so that the device test works.

### DIFF
--- a/jenkins/vsts-device-tests-set-status.sh
+++ b/jenkins/vsts-device-tests-set-status.sh
@@ -41,7 +41,7 @@ fi
 
 if test -z "$DEVICE_TYPE"; then
 	# set a default value
-	DEVICE_TYPE="iOS/AppleTv"
+	DEVICE_TYPE="iOS/tvOS"
 fi
 
 P=$(cat tmp.p)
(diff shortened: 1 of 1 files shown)


jenkins/vsts-device-tests-set-status.sh
```diff
@@ -30,6 +39,11 @@ if test -z "$TOKEN"; then
 	exit 1
 fi
 
+if test -z "$DEVICE_TYPE"; then
+	# set a default value
+	DEVICE_TYPE="iOS/AppleTv"
```
rolfbjarne: `iOS/tvOS` sounds better, since those are the official platform names.

mandel-macaque:
Failures can be ignored.

mandel-macaque:

mandel-macaque:
@monojenkins backport to 10.2

PR #5821, opened 2019-03-29
[msbuild] Add test case for #5129.

rolfbjarne:


rolfbjarne:
Test failures are regressions in 10.14.4 which will fixed once we merge the xcode10.2 branch.

PR #5823, opened 2019-03-29
[linker] Teach linker to accept old (before optimization) bindings

spouliot:
Existing binding binaries won't have the `[Preserve]` attribute on the `Handler` field and, with the new optimization, would not work properly.

This tweak make sure that older, already linker-safe, bindings will remain this way (safe) in this (and future) versions of both iOS and macOS SDK.


spouliot:
test failures are also on `mono-2019-02` and not related to the PR

PR #5830, opened 2019-04-02
[mk] Support https urls for maccore.

rolfbjarne:
This can make checking out maccore easier on the bots, if https authentication
is configured when git/ssh is not.


rolfbjarne:
Test failures are due to 10.14.4 regressions, and those will be fixed when we merge xcode10.2.

PR #5833, opened 2019-04-02
[Foundation] Make adding/removing observers in NSNotificationCenter thread safe.

rolfbjarne:
NSNotificationCenter is thread-safe, so that observers can be added and
removed in multiple threads simultaneously.

Unfortunately our own helper code wasn't, so make sure to fix that by locking
around all accesses to the observer list.

And add a test case as well.


rolfbjarne:
Test failures are due to 10.14.4 regressions, and those will be fixed when we merge xcode10.2.

PR #5844, opened 2019-04-03
Small fixlets to select better overloads / unneeded parameters

spouliot:


spouliot:
known failures
* pre-xcode 10.2 merge
* network connectivity (checked with @mandel-macaque )

PR #5848, opened 2019-04-04
[XHarness] Ignore a failing test on devices.

mandel-macaque:


mandel-macaque:
To be reverted once we have https://github.com/mono/mono/pull/13693 in master.

mandel-macaque:
Failures are due to network issues :/

PR #5849, opened 2019-04-04
[TVMLKit] Work around Apple breaking API for enum values in TVElementUpdateType.

rolfbjarne:
Reference: https://trello.com/c/Eyox7vlH/118-41123682-tvelementupdatetype-enum-values-changed
Reference: rdar://41123682 (closed as by design)

### DIFF
--- a/src/tvmlkit.cs
+++ b/src/tvmlkit.cs
@@ -86,10 +86,17 @@ namespace TVMLKit {
 	public enum TVElementUpdateType : long {
 		None,
 		Subtree,
+#if XAMCORE_4_0
+		[TV (10,0)]
+		Styles,
+		Children,
+		Self,
+#else
 		Children,
 		Self,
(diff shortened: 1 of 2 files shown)


src/tvmlkit.cs
```diff
@@ -819,8 +826,9 @@ interface TVViewElement : NSCopying {
 			set;
 		}
 
+		[Internal][Sealed]
 		[Export ("updateType")]
-		TVElementUpdateType UpdateType { get; }
+		TVElementUpdateType _UpdateType { get; }
```
spouliot: only for `!XAMCORE_4_0` ?

rolfbjarne: Actually the workaround should continue to exist in XAMCORE_4_0 (theoretically until tvOS 12 is the earliest tvOS version we support). I've updated the manually bound property accordingly.

PR #5850, opened 2019-04-04
[Intents] Update according to breaking changes from Apple.

rolfbjarne:
Reference: https://trello.com/c/dtVv2a6R/131-43425168-inworkoutintentresponsecode-enum-values-changed
Reference: rdar://43425168

Working around this breaking change becomes quite complicates, so I just
exposed it instead. Customers can still work around it if needed by casting
the expected numeric value to the enum type.


rolfbjarne:
The breaking changes are expected (the entire point of this PR).

PR #5858, opened 2019-04-05
[XHarness] Ignore tests that fail on device.

mandel-macaque:


mandel-macaque:
@spouliot I talked with @akoeplinger , do we have a mono issue? I know we talked about these tests.

akoeplinger:
I don't think we have a mono bug as these tests never ran on devices before so we need to make a general pass over them anyway.

mandel-macaque:
@akoeplinger do we create a mono bug, to track this? that way once the bug is closed, we can remove the tests/file

akoeplinger:
@mandel-macaque yeah I'll do that on monday.

spouliot:
known HE0038 issue on watchOS sim -> https://github.com/xamarin/maccore/issues/581

PR #5862, opened 2019-04-07
Add macOS Xcode archive support

wjk:
Does mostly what the title says. Note that this code is totally untested (mostly due to how hard it is to do an end-to-end build of Xamarin.Mac; I have tried and failed several times in the past). I am mainly opening the PR now both for feedback, and so the CI bot can tell me what I did wrong. Note also the TODO comments in my added code; I would appreciate some guidance on how to deal with those. Thanks so much!

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
+++ b/msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.IO;
+using System.Diagnostics;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+using Xamarin.MacDev;
+using Xamarin.MacDev.Tasks;
+
+namespace Xamarin.Mac.Tasks.Tasks
+{
+    public abstract class ArchiveTaskBase : Xamarin.MacDev.Tasks.ArchiveTaskBase
+    {
+        public override bool Execute ()
+        {
+            var archiveDir = CreateArchiveDirectory ();
+
+            try {
+                var plist = PDictionary.FromFile (Path.Combine (AppBundleDir.ItemSpec, "Contents", "Info.plist"));
+                var productsDir = Path.Combine (archiveDir, "Products");
+
+                // TODO: Archive the OnDemandResources.
+                // I don't have any on hand, so I can't use an Xcode-created archive as reference.
+
+                // Archive the Applications...
+                var appDestDir = Path.Combine (productsDir, "Applications", Path.GetFileName (AppBundleDir.ItemSpec));
+                Ditto (AppBundleDir.ItemSpec, appDestDir);
+
+                // Archive the dSYMs...
+                if (Directory.Exists (DSYMDir)) {
+                    var destDir = Path.Combine (archiveDir, "dSYMs", Path.GetFileName (DSYMDir));
+                    Ditto (DSYMDir, destDir);
+                }
+
+                // Generate an archive Info.plist
+                var arInfo = new PDictionary ();
+                var props = new PDictionary ();
+                props.Add ("ApplicationPath", new PString (string.Format ("Applications/{0}", Path.GetFileName (AppBundleDir.ItemSpec))));
+                props.Add ("CFBundleIdentifier", new PString (plist.GetCFBundleIdentifier ()));
+
+                var version = plist.GetCFBundleShortVersionString ();
+                var build = plist.GetCFBundleVersion ();
+                props.Add ("CFBundleShortVersionString", new PString (version ?? (build ?? "1.0")));
+                props.Add ("CFBundleVersion", new PString (build ?? "1.0"));
+                props.Add ("SigningIdentity", SigningKey);
+                // TODO: Add "Team" key (Apple Team ID) here. Extract it from the SigningKey somehow?
+
+                arInfo.Add ("ArchiveVersion", new PNumber (2));
+                arInfo.Add ("CreationDate", Now);
+                arInfo.Add ("Name", new PString (plist.GetCFBundleName () ?? plist.GetCFBundleDisplayName ()));
+                arInfo.Add ("ApplicationProperties", props);
+
+                arInfo.Save (Path.Combine (archiveDir, "Info.plist"));
+            } catch (Exception ex) {
+                Log.LogErrorFromException (ex);
+                Directory.Delete (archiveDir, true);
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+    }
+}
--- a/msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
+++ b/msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
@@ -47,6 +47,20 @@ namespace Xamarin.Mac.Tasks
                 arInfo.Add ("Name", new PString (plist.GetCFBundleName () ?? plist.GetCFBundleDisplayName ()));
                 arInfo.Add ("ApplicationProperties", props);
 
+                if (!string.IsNullOrEmpty (ProjectGuid))
+                    arInfo.Add ("ProjectGuid", new PString (ProjectGuid));
+
+                if (!string.IsNullOrEmpty (ProjectTypeGuids))
+                    arInfo.Add ("ProjectTypeGuids", new PString (ProjectTypeGuids));
+
+                if (!string.IsNullOrEmpty (SolutionPath)) {
+                    arInfo.Add ("SolutionName", new PString (Path.GetFileNameWithoutExtension (SolutionPath)));
+                    arInfo.Add ("SolutionPath", new PString (SolutionPath));
+                }
+
+                if (!string.IsNullOrEmpty (InsightsApiKey))
+                    arInfo.Add ("InsightsApiKey", new PString (InsightsApiKey));
+
                 arInfo.Save (Path.Combine (archiveDir, "Info.plist"));
             } catch (Exception ex) {
                 Log.LogErrorFromException (ex);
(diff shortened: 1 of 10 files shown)


rolfbjarne:
build

chamons:
👋 Thanks for the contribution! 

I think the biggest thing this is missing is a test. I don't personally know archive well enough to know if this works \ continues working. Maybe we can write a test like [this one](https://github.com/xamarin/xamarin-macios/blob/master/tests/mmptest/src/MMPTest.cs#L679)? Generate a project, run the archive target, and compare the output to expected files.

Would that be something you'd be up for trying @wjk ?

@jstedfast - Could you review the task parts?

wjk:
@chamons I can certainly try, but the main way of testing if an Xcode archive is valid is to import it into Xcode and see if it complains, and that can only be done by hand.

Doing this may take me a while, though, due to how hard it is for me to build Xamarin.Mac. (My university has set up their internet service to momentarily interrupt the connections of users who download more than X bytes in Y period of time to "deter illegal file sharing" — which, by the way, is a nonsense excuse. As such, if I try to recursively clone the xamarin-macios repo, my connection gets hiccuped and Git crashes; I must then retry the download from the beginning, only for my connection to hiccup again…) Thanks for bearing with!

chamons:
@wjk You could look into [trickle](https://github.com/sjmulder/trickle) or other bandwidth shaping tools. They let you invoke a command and set a bandwidth "cap", which might allow you to more easily stay within your universities constraints. I've never personally used it, however.

Please let me know if you think you'll be unable to continue on this, I'm happy to pick some some of the work if necessary to make sure we can land it.

Thanks again!

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -0,0 +1,63 @@
...
+        public override bool Execute ()
+        {
+            var archiveDir = CreateArchiveDirectory ();
+
+            try {
+                var plist = PDictionary.FromFile (Path.Combine (AppBundleDir.ItemSpec, "Contents", "Info.plist"));
+                var productsDir = Path.Combine (archiveDir, "Products");
+
+                // TODO: Archive the OnDemandResources.
+                // I don't have any on hand, so I can't use an Xcode-created archive as reference.
```
jstedfast: No need for OnDemandResources for macOS. If macOS apps *do* support this, we don't currently support it in the IDE's archiving logic, so don't worry about it.

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -0,0 +1,63 @@
...
+                var props = new PDictionary ();
+                props.Add ("ApplicationPath", new PString (string.Format ("Applications/{0}", Path.GetFileName (AppBundleDir.ItemSpec))));
+                props.Add ("CFBundleIdentifier", new PString (plist.GetCFBundleIdentifier ()));
+
+                var version = plist.GetCFBundleShortVersionString ();
+                var build = plist.GetCFBundleVersion ();
+                props.Add ("CFBundleShortVersionString", new PString (version ?? (build ?? "1.0")));
+                props.Add ("CFBundleVersion", new PString (build ?? "1.0"));
+                props.Add ("SigningIdentity", SigningKey);
+                // TODO: Add "Team" key (Apple Team ID) here. Extract it from the SigningKey somehow?
```
jstedfast: You can't get the Apple Team ID from the SigningKey. The only way to get this value is to get it from the Provisioning Profile (if the user selected one for this build).

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -0,0 +1,63 @@
...
+
+                arInfo.Add ("ArchiveVersion", new PNumber (2));
+                arInfo.Add ("CreationDate", Now);
+                arInfo.Add ("Name", new PString (plist.GetCFBundleName () ?? plist.GetCFBundleDisplayName ()));
+                arInfo.Add ("ApplicationProperties", props);
+
```
jstedfast: Ideally we'd also add:

```csharp
if (!string.IsNullOrEmpty (ProjectGuid))
	arInfo.Add ("ProjectGuid", new PString (ProjectGuid));

if (!string.IsNullOrEmpty (ProjectTypeGuids))
	arInfo.Add ("ProjectTypeGuids", new PString (ProjectTypeGuids));

if (!string.IsNullOrEmpty (SolutionPath)) {
	arInfo.Add ("SolutionName", new PString (Path.GetFileNameWithoutExtension (SolutionPath)));
	arInfo.Add ("SolutionPath", new PString (SolutionPath));
}

if (!string.IsNullOrEmpty (InsightsApiKey))
	arInfo.Add ("InsightsApiKey", new PString (InsightsApiKey));
```

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
jstedfast: FWIW, we don't currently get that value for iOS either, so if you want to skip it, go ahead.

wjk:
@jstedfast The ArchiveTaskBase class in Xamarin.MacDev.Tasks.Core has marked its ITunesSourceFiles parameter as `[Required]`, yet this parameter is meaningless on macOS AFAICT. As such, this will cause errors when attempting to create a macOS archive. Should I just get rid of the `[Required]` attribute?

jstedfast:
I think the better approach would be to move that property & attribute over to the Xamarin.iOS.Tasks.Core.ArchiveTaskBase class.

wjk:
@chamons Would it be possible for you to test this as you offered above? Download cap or no download cap, building Xamarin.Mac is still way too much of a pain for me; I managed to get reset-submodules to succeed, but then ran into problems with provisioning. (I really don't like having multiple versions of Xcode around; it tends to confuse Launchpad.) If you could take over this task I'd be very appreciative. Thanks!

(Also: That trickle program you linked doesn't throttle bandwidth; it is purely an amusement that rate-limits the passage of characters to the terminal.)

chamons:
That's unfortunate about trickle, that would have been useful. What I get for suggesting without detail reading.

I'll take a look at this later this week.

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -0,0 +1,74 @@
...
+                if (!string.IsNullOrEmpty (ProjectTypeGuids))
+                    arInfo.Add ("ProjectTypeGuids", new PString (ProjectTypeGuids));
+
+                if (!string.IsNullOrEmpty (SolutionPath)) {
+                    arInfo.Add ("SolutionName", new PString (Path.GetFileNameWithoutExtension (SolutionPath)));
+                    arInfo.Add ("SolutionPath", new PString (SolutionPath));
+                }
+
+                if (!string.IsNullOrEmpty (InsightsApiKey))
```
chamons: @wjk is this an Apple thing or something else? Who's key is it?

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -0,0 +1,74 @@
...
+                props.Add ("CFBundleShortVersionString", new PString (version ?? (build ?? "1.0")));
+                props.Add ("CFBundleVersion", new PString (build ?? "1.0"));
+                props.Add ("SigningIdentity", SigningKey);
+
+                arInfo.Add ("ArchiveVersion", new PNumber (2));
+                arInfo.Add ("CreationDate", Now);
+                arInfo.Add ("Name", new PString (plist.GetCFBundleName () ?? plist.GetCFBundleDisplayName ()));
+                arInfo.Add ("ApplicationProperties", props);
+
+                if (!string.IsNullOrEmpty (ProjectGuid))
```
chamons: Were these just things you found useful or needed for archive?

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
wjk: I simply copied it from the iOS version. The targets file implies it's the Xamarin Insights API key. I don't think it's used anywhere, honestly.

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
wjk: @jstedfast [requested](https://github.com/xamarin/xamarin-macios/pull/5862#discussion_r273066536) I add it.

jstedfast: They are for the IDE to associate projects/solutions with the archive

chamons:
I've hacked on it enough that a build with `/p:ArchiveOnBuild=true ` does not error and generates an xcarchive.

However, they aren't loadable by Xcode.

@wjk - Were you able to load load generated xcarchive in Xcode using your generated bundles?

wjk:
@chamons I’ve found the cause of the problem, but I don't know how to solve it. The issue is that the plist-writing framework is using an invalid `<date>` format in the XML plist, as follows:

```xml
<!-- This was written natively by Xcode. -->
<key>CreationDate</key>
<date>2019-04-07T00:29:58Z</date>

<!-- This was written by VSMac. -->
<key>CreationDate</key>
<date>2019-04-11T18:02:00-04:00</date>
```

This appears to be a bug in PlistObject. I suspect that the date needs to be converted to UTC before being written out, but I’ve tried calling `DateTime.ToUniversalTime()`; it doesn’t help. I’ve also tried wrapping the date in `new PDate(...)` explicitly; it doesn’t help either. Any pointers? Thanks!

chamons:
That appears to be it. 
Screen Shot 2019-04-12 at 11 02 49 AM

Works for both full and modern.

rolfbjarne:
build

chamons:
@wjk Thanks for the contribution!

jay8t6:
@wjk @chamons is it possible to create an unsigned xcarchive file?

PR #5872, opened 2019-04-08
[XHarness] Ignore System.Data tests until the correct fix lands in mono.

mandel-macaque:
The tests fail on device, they can be reneabled after we get PR
https://github.com/mono/mono/pull/13935 landed and backported.


akoeplinger:
I just merged the PR down to 2018-10 so this should no longer be needed :)

rolfbjarne:
> I just merged the PR down to 2018-10 so this should no longer be needed :)

OK, superseded by #5875.

PR #5873, opened 2019-04-08
[devops] Add azure devops build support

dalexsoto:


dalexsoto:
This requires a maccore bump, but I removed the bump from this PR in favor of https://github.com/xamarin/xamarin-macios/pull/5874

PR #5887, opened 2019-04-11
[XHarness] Update ignore for some failing tests.

mandel-macaque:
The mono bump added some new tests that fail on device + release
configurations. This is a known issue by mono and they'll fix it in the
next version but will not be backported to 2018-10.


akoeplinger:
The commit to fix this is in progress on our end, just need to work through upstreaming it to corefx etc.

mandel-macaque:
The simulator on watch os crashing is known: https://github.com/xamarin/maccore/issues/581

PR #5891, opened 2019-04-11
[bgen] Add support for sequential and parallel execution of the generator within the same process.

rolfbjarne:
This has two main advantages:

* It'll be possible to step into the generator's code when debugging tests,
  since we won't be launching an external process to run bgen.

* The tests run slightly faster (32s vs 39s on my machine), since we spawn
  fewer subprocesses.

The required changes basically consist of making sure we don't have static
state anywhere. This is mostly accomplished by passing instances around
instead of relying on static API. In many cases I was able to create an
instance field with the same name as the type, which made existing code work
without change.

The only exception is error reporting: in that case we rely on a ThreadStatic
field to store the warnings-as-errors list, since otherwise the modifications
were both ugly and numerous. This will probably break if we one day start
using async in the generator, but for now it should be safe enough.

Most of the other changes were fairly mechanical: just add a parameter or a
field, and then use it.

### DIFF
--- a/src/btouch.cs
+++ b/src/btouch.cs
@@ -52,6 +52,7 @@ public class BindingTouch {
 
 	public Universe universe;
 	public Frameworks Frameworks;
+	public AttributeManager AttributeManager;
 
 	public TargetFramework TargetFramework {
 		get { return target_framework.Value; }
@@ -514,7 +515,7 @@ public class BindingTouch {
 				return 1;
 			}
 
-			AttributeManager.BindingTouch = this;
+			AttributeManager = new AttributeManager (this);
 			Stret.BindingTouch = this;
 			Frameworks = new Frameworks (CurrentPlatform);
 
--- a/src/btouch.cs
+++ b/src/btouch.cs
@@ -51,6 +51,7 @@ public class BindingTouch {
 	List<string> libs = new List<string> ();
 
 	public Universe universe;
+	public TypeManager TypeManager = new TypeManager ();
 	public Frameworks Frameworks;
 	public AttributeManager AttributeManager;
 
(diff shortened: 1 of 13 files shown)


rolfbjarne:
> any reason for a mono bump in the PR ?

No, that's a big oops. No idea how that happened 😒

src/btouch.cs
```diff
@@ -212,6 +215,12 @@ static void SetTargetFramework (string fx)
 	}
 
 	static int Main2 (string [] args)
+	{
+		var touch = new BindingTouch ();
+		return touch.Main3 (args);
+	}
+
+	int Main3 (string [] args)
```
spouliot: can't stop progress :)

tests/generator/BGenTool.cs
```diff
@@ -88,54 +90,78 @@ string BuildArguments ()
 			}
 
 			if (!string.IsNullOrEmpty (targetFramework))
-				sb.Append (" --target-framework=").Append (targetFramework);
+				sb.Add ($"--target-framework={targetFramework}");
 
 			foreach (var ad in ApiDefinitions)
-				sb.Append (" --api=").Append (StringUtils.Quote (ad));
+				sb.Add ($"--api={ad}");
```
spouliot: are all those `StringUtils.Quote` removal intentional ? because even if we don't use them today we might need them inn future tests

rolfbjarne: Yes, the code is now adding all the arguments (unquoted) to an array. If we're calling Main, we pass the array (without quoting). If we call an out-of-process bgen, we join each element of the array to a string, quoting when needed.

PR #5900, opened 2019-04-15
[CFNetwork] Ensure we do not block on redirects.

mandel-macaque:
The changes in mono are changing threading. This means that the code is
getting deadlock in the ReadStreamData.

This happens because we were not removing the handlers from the first request events.
Initially, before we had to take care about the auth headers, this was needed to init
certain internal state for the HttpContent that was exposed to the
user. With the fix of the CVE, the first request, in case of a redirect,
is garbage and is ignored. The second request is the one returned to the
user and contains all the required info.

Removing the event handlers, ensure that HandleCloseEvent is not called and
therefore ReadStream is never called.


akoeplinger:
I ran the monotouch-tests on simulator on two macs, and on my iPad. All passed 👍

PR #5908, opened 2019-04-16
[XHarness] Ignore device tests for platforms in certain bots.

mandel-macaque:
If a bot does not have the required devices, such as a tvOS, we do not
want to generate the device variations for that platform. This allows to
ensure that we do not have Skipped tests for devices we do know that are
not present.

### DIFF
--- a/tests/xharness/Harness.cs
+++ b/tests/xharness/Harness.cs
@@ -704,6 +704,27 @@ namespace xharness
 				return string.IsNullOrEmpty (groupApps) || groupApps == "grouped";
 			}
 		}
+		
+		public bool SkipTvOSDeviceVariations {
+			get {
+				var skipTv = Environment.GetEnvironmentVariable ("SKIP_TVOS_VARIATIONS");
+				return !string.IsNullOrEmpty (skipTv) && skipTv == "true";
+			}
+		}
+		
+		public bool SkipiOSDeviceVariations {
+			get {
+				var skipiOs = Environment.GetEnvironmentVariable ("SKIP_IOS_VARIATIONS");
+				return !string.IsNullOrEmpty (skipiOs) && skipiOs == "true";
+			}
+		}
+		
+		public bool SkipWatchOSDeviceVariations {
+			get {
+				var skipWatchOS = Environment.GetEnvironmentVariable ("SKIP_WATCHOS_VARIATIONS");
+				return !string.IsNullOrEmpty (skipWatchOS) && skipWatchOS == "true";
+			}
+		}
 
 		public int Execute ()
 		{
--- a/tests/xharness/Harness.cs
+++ b/tests/xharness/Harness.cs
@@ -704,27 +704,6 @@ namespace xharness
 				return string.IsNullOrEmpty (groupApps) || groupApps == "grouped";
 			}
 		}
-		
-		public bool SkipTvOSDeviceVariations {
-			get {
-				var skipTv = Environment.GetEnvironmentVariable ("SKIP_TVOS_VARIATIONS");
-				return !string.IsNullOrEmpty (skipTv) && skipTv == "true";
-			}
-		}
-		
-		public bool SkipiOSDeviceVariations {
-			get {
-				var skipiOs = Environment.GetEnvironmentVariable ("SKIP_IOS_VARIATIONS");
-				return !string.IsNullOrEmpty (skipiOs) && skipiOs == "true";
-			}
-		}
-		
-		public bool SkipWatchOSDeviceVariations {
-			get {
-				var skipWatchOS = Environment.GetEnvironmentVariable ("SKIP_WATCHOS_VARIATIONS");
-				return !string.IsNullOrEmpty (skipWatchOS) && skipWatchOS == "true";
-			}
-		}
 
 		public int Execute ()
 		{
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -25,6 +25,7 @@ namespace xharness
 		public bool IncludeMac = true;
 		public bool IncludeMac32 = true;
 		public bool IncludeiOS = true;
+		public bool IncludeiOS32 = true;
 		public bool IncludeiOSExtensions;
 		public bool ForceExtensionBuildOnly;
 		public bool IncludetvOS = true;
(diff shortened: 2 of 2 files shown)


tests/xharness/Harness.cs
```diff
@@ -704,6 +704,27 @@ public void LogWrench (string message)
 				return string.IsNullOrEmpty (groupApps) || groupApps == "grouped";
 			}
 		}
+		
+		public bool SkipTvOSDeviceVariations {
+			get {
+				var skipTv = Environment.GetEnvironmentVariable ("SKIP_TVOS_VARIATIONS");
+				return !string.IsNullOrEmpty (skipTv) && skipTv == "true";
```
stephen-hawley: Why the `IsNullOrEmpty` check? if `skipTv` is null, the test will be false, as you'd like.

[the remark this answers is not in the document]
spouliot:
```
ools/mtouch/mtouch.cs:         internal static bool TryParseBool (string value, out bool result)
tools/mtouch/mtouch.cs:         internal static bool ParseBool (string value, string name, bool show_error = true)
```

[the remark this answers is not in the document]
mandel-macaque: Let me check, could be in the one used in mono options.

mandel-macaque:
@spouliot On the test result HTML you will not see the option/variation ignored.

For example, in a normal run with all variations you have:

Screenshot 2019-04-16 at 17 25 38

With the env var set on a bot you have:

Screenshot 2019-04-16 at 17 24 06

tests/xharness/Harness.cs
mandel-macaque: @stephen-hawley I followed the pattern used in the rest of the properties. In my case, just following the style already present.. don't know if that is a good reason enough.

spouliot:
@mandel-macaque I'd rather have a note *why* than simply remove it (since I can forget things I don't see), e.g.

+ tvOS - device (**none present**) (Ignore) - Run all Build all

VincentDondain:
@mandel-macaque upon further investigation, this does not totally address https://github.com/xamarin/xamarin-macios/commit/d4e41445a20d8b1655c2db07e71bbadb40fbeb4b#commitcomment-33183309 does it? If we have a "skipped" test for any other reason we'll still fail... For instance you wouldn't skip the iOS 32 bit device variation with that logic.

So I don't know if I agree with "skipping variations", as Sebastien said we want to know why things were skipped. I think we should update the GitHub reporting to say things passed and here's the list of things that were skipped and here's why.

mandel-macaque:
@VincentDondain it does. We **DO WANT** to have failures on skips, why? Because a skip may happen if a device died in the middle of a test run, or between runs. If we consider skips to be successful, we could believe we are running the tests and are successful, but in reality we had device issues.

What we really want to do is not run those. There are two ways:

1. The env var (I will push an update)
2. Modify 'make vsts-device-tests' to be able to pass an extra param for it to be ignored.

I decided to go to the first option, since it is easier to later change the behaviour and modify the VSTS jobs (is harder to go job by job modifying a shell script than adding a simple env var).

spouliot:
@mandel-macaque my main issue is how to find it was not run and why. e.g. environment variables can be set on bots without a commit.

How can I know it's the reason the test was skipped ? Ideally without too much clicking to open more web pages...

Now there might already be a way to detect this but it's not clear by reading the PR :)

Throwing ideas... what about updating `Device Listing (ok)` at the top ? right now `ok` can be no device... maybe change `ok` for `ok - no device` or `ok - 2 iOS and 0 tvOS device(s)` ? That would give some _context_ why tvOS are ignored later (not very _local_ but better than clicks)

mandel-macaque:
@spouliot 10000% on showing the device list. I'll do that!

tests/xharness/Harness.cs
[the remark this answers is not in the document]
akoeplinger: I'd use only the `string.IsNullOrEmpty` check, that's what's done elsewhere:

https://github.com/xamarin/xamarin-macios/blob/cf8d42e4c0de067664ddc770de24199c884566e8/tests/xharness/Harness.cs#L821

https://github.com/xamarin/xamarin-macios/blob/cf8d42e4c0de067664ddc770de24199c884566e8/tests/xharness/AppRunner.cs#L322

mandel-macaque:
@spouliot Added the required change, you will not see the tests AND you will see the devices available in the bot as follows:

Screenshot 2019-04-22 at 13 46 51

mandel-macaque:
@rolfbjarne I have removed the changes regarding the env, will pass the correct params to the VSTS jobs.

tests/xharness/Jenkins.cs
```diff
@@ -101,8 +102,24 @@ Task LoadAsync (ref Log log, ILoadAsync loadable, string name)
 					capturedLog.WriteLine ("Failed to load:");
 					capturedLog.WriteLine (v.Exception);
 					capturedLog.Description = $"{name} Listing {v.Exception.Message})";
-				} else if (v.IsCompleted) {
-					capturedLog.Description = $"{name} Listing (ok)";
+				} else if (v.IsCompleted) { // TODO
```
rolfbjarne: TODO?

tests/xharness/Jenkins.cs
```diff
@@ -101,8 +102,24 @@ Task LoadAsync (ref Log log, ILoadAsync loadable, string name)
...
+					var devices = loadable as Devices;
+					var devicesTypes = new StringBuilder ();
+					if (devices != null) {
+						if (devices.Connected32BitIOS.Any()) {
```
rolfbjarne: Style: space before opening parenthesis.

tests/xharness/Jenkins.cs
```diff
@@ -134,7 +151,10 @@ IEnumerable<RunSimulatorTask> CreateRunSimulatorTaskAsync (XBuildTask buildTask)
 				break;
 			case TestPlatform.iOS_Unified:
 				targets = new AppRunnerTarget [] { AppRunnerTarget.Simulator_iOS32, AppRunnerTarget.Simulator_iOS64 };
-				platforms = new TestPlatform [] { TestPlatform.iOS_Unified32, TestPlatform.iOS_Unified64 };
+				if (IncludeiOS32)
```
rolfbjarne: This will prevent the 32-bit tests from showing up. It's better to ignore them instead:

```csharp
bool [] ignored;
...
case TestPlatform.tvOS:
    ignored = new bool [] { false };
    ...
case TestPlatform.watchOS:
    ignored = new bool [] { false };
    ...
case TestPlatform.iOS_Unified:
    ignored = new bool [] { !IncludeiOS32, false };
    ...
...
    runtasks.Add (new RunSimulatorTask (..., Ignored = ignored [i] });
```

tests/xharness/Jenkins.cs
mandel-macaque: I was using it to navigate, will remove.

mandel-macaque:
Was aborted due a timeout because of the 32b dialog :/

mandel-macaque:
Looks like that time in the month in which we are getting the 32b dialog again.. retrying

PR #5919, opened 2019-04-19
[mtouch] Enable dsym even for aot'ed images when interpreter is enabled

BrzVlad:
Fixes https://github.com/mono/mono/issues/13975

### DIFF
--- a/tools/mtouch/Application.cs
+++ b/tools/mtouch/Application.cs
@@ -514,7 +514,12 @@ namespace Xamarin.Bundler {
 			if (EnableLLVMOnlyBitCode)
 				return false;
 
-			if (IsInterpreted (Assembly.GetIdentity (assembly)))
+			// Even if this assembly is aot'ed, if we are using the interpreter we can't yet
+			// guarantee that code in this assembly won't be executed in interpreted mode,
+			// which can happen for virtual calls between assemblies, during exception handling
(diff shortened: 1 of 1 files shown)


tools/mtouch/Application.cs
```diff
@@ -514,7 +514,11 @@ public bool UseDlsym (string assembly)
 			if (EnableLLVMOnlyBitCode)
 				return false;
 
-			if (IsInterpreted (Assembly.GetIdentity (assembly)))
+			// Even if this assembly is aot'ed, if we are using the interpreter we can't yet
+			// guarantee that code in this assembly won't be executed in interpreted mode,
+			// which can happen for virtual calls between assemblies, during exception handling
+			// etc. We make sure we don't strip away symbols needed for pinvoke calls.
+			if (UseInterpreter)
```
spouliot: where is that limitation tracked ?
please add link so we can revert this later

lewurm: I don't think this limitation is fixable

BrzVlad: Created issue on mono side. If only the virtual calls produce this issue, it might be fixable.

spouliot: thanks, at least we know where to discuss (or find discussions) about this 😄

PR #5922, opened 2019-04-22
[ObjCRuntime] Make Class IEquatable.

rolfbjarne:
Makes it easier for tests to compare arrays of Classes, since existing NUnit
API supports comparing arrays of IEquatables.


rolfbjarne:
build

PR #5925, opened 2019-04-22
[runtime] Get the domain at the top of xamarin_invoke_trampoline.

rolfbjarne:
This avoids calling it multiple times in a loop further down in the method.


Therzok:
Ah, damn, it is not an `external approve` :(

Therzok:
> We'll never have app domains?

VSMac has appdomains, but not sure we ever use anything xammac related in an appdomain

rolfbjarne:
> I assume the domain shouldn't change. We'll never have app domains

The current domain won't change randomly, it can only happen from (user) managed code (or from inside mono).

PR #5932, opened 2019-04-23
[tests] Fix bgen test to compile by adding missing using.

rolfbjarne:


tests/generator/BGenTests.cs
```diff
@@ -336,7 +336,6 @@ public void StrongDictSupportTemplatedDicts ()
 		}
 
 		[Test]
-		[Ignore ("Fails with: api.cs(10,17): error CS0246: The type or namespace name `CBUUID' could not be found. Are you missing `CoreBluetooth' using directive?")]
```
spouliot: looks like an unrelated change / not mentioned in the commit message

rolfbjarne: Nope, it's on purpose, this is the test that's being fixed. Since it's fixed, we don't have to ignore it anymore.

PR #5934, opened 2019-04-23
[registrar] Add token type validation.

rolfbjarne:
This should never happen, but an MT0099 at build time is much better than a
runtime exception.


rolfbjarne:
Test failure is unrelated and will be fixed with PR #5940.

PR #5944, opened 2019-04-24
[runtime] Improve exception creation.

rolfbjarne:
First add an inner exception parameter to Runtime.CreateProductException.

This allows us to simplify code by using nested exceptions as a means to
provide information instead of passing extra information around in order to
create a single exception with all we know.

One example is how we pass the selector and method name to the method that
converts from a native id to a managed NSObject instance: passing this
information is not necessary anymore if we can use two exceptions, one for
the failure to convert from an id to a NSObject instance, wrapped in a
second that tells which method/selector call ran into this conversion
problem.

While not a big problem in the current code, upcoming changes ran into this
and this solution turned out to make the code much nicer.

Also slightly improve the exception we throw when failing to marshal something
in the dynamic registrar.


rolfbjarne:
Failure is kind of related, but will be fixed in a different PR: #5945. I'll merge master here once that PR is merged to verify it works.

PR #5952, opened 2019-04-26
[jenkins] Add support to build.sh for timing out the build.

rolfbjarne:


jenkins/build.sh
```diff
@@ -76,6 +120,7 @@ echo "Configuring the build with: $CONFIGURE_FLAGS"
 # shellcheck disable=SC2086
 ./configure $CONFIGURE_FLAGS
 
+sleep 3600 # test timeout support
```
rolfbjarne: I've modified Jenkins to pass --timeout=3600 to build.sh, this line is to test that it times out as expected. It will be removed before merging the PR.

PR #5960, opened 2019-04-29
[builds] Fix race condition when building device runtimes.

rolfbjarne:
We need to complete the corresponding cross builds first, because the cross
builds will also try to configure device builds, so we might end up with two
make processes trying to configure the same target simulatenously.

Also add support for building mono from source in pull requests by using
labels.


rolfbjarne:
build

PR #5962, opened 2019-04-29
[builds] Make the tools build use mono's packaged logic instead of our own.

rolfbjarne:
Make the 'tools64' build use mono's packaged build logic instead of our own.
This is the first step to consuming the BCL from the mono archive.

Also completely refactor the 'tools64' build by removing everything we don't
need and renaming it to 'bcl' (since that's more representative of what it
does).

This also means we're now consuming the BCL from the mono archive.

### DIFF
--- a/builds/Makefile
+++ b/builds/Makefile
@@ -141,7 +141,7 @@ $(MONO_PATH)/sdks/Make.config:
 ##
 
 # We only build the mono runtimes here (mac32 and mac64)
-# The xammac classlibs are built as a part of the 'tools64' build.
+# The xammac classlibs are built as a part of the 'bcl' build.
 
 MAC_ASSEMBLY_NAMES = \
 	I18N \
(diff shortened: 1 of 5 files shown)


builds/Makefile
```diff
@@ -328,8 +326,8 @@ $(MAC_DIRECTORIES) $(BUILD_DESTDIR):
 mac-facade-check: $(MAC_BCL_TARGETS)
```
akoeplinger: We'll need to get rid of the `FACADE_SUBDIRS_MK` which is included directly from mono sources in https://github.com/xamarin/xamarin-macios/blob/03669d4c030b2313d2037493307e3cb2d4f3bd88/builds/Makefile#L188

There's two options: 1) just include all *.dll in the ios/mac-bcl folders and trust mono to not put unwanted files there 2) hardcode list of assemblies/facades here. Any preferences?

rolfbjarne:
The api comparison failed because the src build is too different from the previous commit: i.e. the failure is expected/can be ignored.

builds/Makefile
rolfbjarne: I think the [Xamarin.Linker.SdkTest](https://github.com/xamarin/xamarin-macios/blob/9945186b9e22334e044e6cb4537c3031b710a402/tests/mtouch/SdkTest.cs#L32) should cover this scenario, so I think we can just remove the whole facade check from this makefile.

@spouliot what do you think?

builds/Makefile
```diff
@@ -618,96 +550,116 @@ IOS_BCL_TARGETS +=
...
+	$(BUILD_DESTDIR)/ios-bcl/monotouch_watch/Facades \
+	$(BUILD_DESTDIR)/ios-bcl/monotouch_watch_runtime \
+	$(BUILD_DESTDIR)/mac-bcl/xammac \
+	$(BUILD_DESTDIR)/mac-bcl/xammac/Facades \
+	$(BUILD_DESTDIR)/mac-bcl/xammac_net_4_5 \
+	$(BUILD_DESTDIR)/mac-bcl/xammac_net_4_5/Facades \
+
+$(BUILD_DESTDIR)/ios-bcl/%.dll: $(MONO_IOS_SDK_DESTDIR)/ios-bcl/%.dll | $(TMP_BCL_TARGET_DIRS)
+	$(Q) $(CP) $< $@
+	$(Q_SN) MONO_CFG_DIR=$(TOP) $(SYSTEM_SN) -q -R $@ $(PRODUCT_KEY_PATH)
```
akoeplinger: do we need to re-sign here? the assemblies from the mono archive should already have the correct signature (if they don't I'd consider that a bug).

rolfbjarne: I don't think we do need to re-sign. I'll have a look and put up another PR if that's the case.

builds/Makefile
rolfbjarne: Actually the facade check is just one part of the problem, we'd still need a list of the facade assemblies. Solution 1) won't work when building from source, so I think we need a dual solution: keep the existing code when building from source, and then use solution 1) when we're using the mono archive.

We still have a hard-coded list of facade assemblies elsewhere in our code (in tests), so we'll be notified if new facade assemblies are added to the archive (the corresponding test will start to fail).

PR #5967, opened 2019-04-30
[builds] package-ios-bcl and package-mac-bcl aren't parallel safe.

rolfbjarne:
They're not even safe to run sequentially without some manual intervention.

This will be fixed in mono at some point, and we make this parallelized again.


rolfbjarne:
build

rolfbjarne:
I'll incorporate this into the next attempt at using the BCL from the mono archive.

PR #5970, opened 2019-04-30
[builds] Don't reference makefiles from the mono repository when using the mono archive.

rolfbjarne:
Instead of using a makefile in the mono repository to get the list facades to
ship, we list the facade assemblies present in the mono archive instead.

We still have a hardcoded list of facade assemblies elsewhere, and tests that
checks that list against what's shipped, so any changes in the list of
assemblies (unwanted or not) won't go unnoticed.


rolfbjarne:
Let's wait with this until we're consuming the BCL assemblies from the mono archive again (i.e. #5971 is resolved).

akoeplinger:
Superseded by my https://github.com/xamarin/xamarin-macios/pull/6097

PR #5976, opened 2019-05-01
[apidiff] Keep the same guid when updating markdown documents

spouliot:
We often (e.g. previews, service releases) update the API diff during
a release cycle. The current code generated a new GUID every time, which
is not what correct since it's the same document.

This uses an MD5 digest of the filename as the source of the GUID so
it will remain constant once created (and updated).


spouliot:

PR #5984, opened 2019-05-03
Add more details to MSBuild errors for provisioning profiles

SamuelDebruyn:
A lot of apps get renamed during the build process. E.g. most of our apps have an alpha or a beta version and they have separate bundle names so that you can install them side-by-side.

I usually don't know if there is an error in my build script or if the provisioning profile is really missing on the machine I'm building on. By displaying the name, the user might notice incorrect bundle names.


rolfbjarne:
build

spouliot:

rolfbjarne:
@SamuelDebruyn thanks a lot for the contribution!

PR #5989, opened 2019-05-03
[runtime] Fix crash in the x86_64 dynamic registrar.

rolfbjarne:
In [this][1] innocuous-looking commit I made the x86_64 dynamic registrar
crash when TRACE is defined. The underlying problem is that CallState's rsi
field is re-used as rax upon exit of the trampoline. This means that the
CallState.sel () function that gets the selector, will instead get the return
value upon exit. When TRACE is defined, we try to print the selector upon
exiting the trampoline, and since the rsi field is now the rax field, we end
up trying to print the return value as if it were a selector, and things go
slightly sideways.

[1]: https://github.com/xamarin/xamarin-macios/commit/464882d14a83

### DIFF
--- a/runtime/trampolines-x86_64-asm.s
+++ b/runtime/trampolines-x86_64-asm.s
@@ -4,8 +4,9 @@
 #   %rdi, %rsi, %rdx, %rcx, %r8, %r9
 #   %xmm0-%xmm7
 #   an unknown amount of stack space, but we can pass a pointer to the start of this area.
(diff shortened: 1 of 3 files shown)


rolfbjarne:
No bug report, I ran into it myself, since it requires uncommenting this line to hit the crash: https://github.com/xamarin/xamarin-macios/blob/b8b4435a0f1ab1508282689873c21f32013d53a4/runtime/trampolines-x86_64.m#L18

PR #5990, opened 2019-05-03
[apidiff] Use mono-api-[info|html].exe from the mono archive.

rolfbjarne:


rolfbjarne:
@spouliot we always download both archives. Also these files are only in the XI archive.

PR #5996, opened 2019-05-06
[generator] The --no-mono-path option doesn't do anything anymore, so remove it.

rolfbjarne:


src/btouch.cs
```diff
@@ -275,7 +275,7 @@ int Main3 (string [] args)
 			{ "attributelib=", "Sets the attribute library", v => attributedll = v },
 			{ "use-zero-copy", v=> zero_copy = true },
 			{ "nostdlib", "Does not reference mscorlib.dll library", l => nostdlib = true },
-			{ "no-mono-path", "Launches compiler with empty MONO_PATH", l => { } },
+			{ "no-mono-path", "Launches compiler with empty MONO_PATH", l => { }, true },
```
spouliot: minor: maybe change the description to mention it's now a no-op ? or add a warning ?

rolfbjarne: The `true` I added at the end makes the whole argument disappear from --help. I don't think it's worth it to add a warning (nobody should have used this argument in the first place).

Issue #3668, opened 2018-03-06
Run simulator tests in older simulators


rolfbjarne:
We should run simulator tests in older simulators, probably the oldest the current version of Xcode supports (currently iOS 8.1, tvOS 9.0 and watchOS 2.0).

This will help catch missing availability attributes, missing version checks in tests, and probably a lot of other issues as well.

It's probably enough to do it for the monotouch-test and introspection tests.

VincentDondain:
@rolfbjarne +1 on this. Did you take inspiration somewhere? :P

https://trello.com/c/1x2pGClP/710-xcode-9-post-mortem#comment-59a5d86e913dddd334b14be4 (oldest comment)

rolfbjarne:
@VincentDondain hehe, I've had it in mind for quite a while, and decided to just document it :smile: I didn't remember seeing it before, but that might very we'll be the inspiration 😄

rolfbjarne:
Martin asked us to run mscorlib and System tests on the earliest iOS version we can as well: 
https://github.com/xamarin/xamarin-macios/pull/4914#issuecomment-429969978

rolfbjarne:
We're doing this, and we have for quite a while already.

PR #6004, opened 2019-05-06
Add support for running tests with the earliest possible simulator, and use it for introspection tests.

rolfbjarne:
* Add support for running tests with the earliest possible simulator
  (currently iOS 8.1, tvOS 9.1 and watchOS 2.0).
* Make the introspection tests run with the earliest possible simulator.
* Fix several binding issues (mostly missing availability attributes), and a
  couple of issues in the introspection tests themselves.

Reference: https://github.com/xamarin/xamarin-macios/issues/3668.


src/HealthKit/EnumsCompat.cs
```diff
@@ -15,6 +15,8 @@ public unsafe static partial class HKQuantityTypeIdentifierKey  {
 			}
 		}
 
+		[Introduced (PlatformName.iOS, 9,3, PlatformArchitecture.All)]
+		[Introduced (PlatformName.WatchOS, 2,2, PlatformArchitecture.All)]
```
spouliot: it's already on the `getter` ?
looks like the test code is wrong (and should be fixed)
but we should not duplicate it on both the property and the getter (too much metadata if the linker is not enabled)

rolfbjarne: The generator puts the attributes on both the getter and the property:

```csharp
[Introduced (PlatformName.iOS, 9,0, PlatformArchitecture.All)]
[Introduced (PlatformName.MacOSX, 10,11, PlatformArchitecture.All)]
public static NSString ColorPrimaries_DCI_P3 {
	[Introduced (PlatformName.iOS, 9,0, PlatformArchitecture.All)]
	[Introduced (PlatformName.MacOSX, 10,11, PlatformArchitecture.All)]
	get {
		return CVImageBufferColorPrimaries.DciP3.GetConstant () as NSString;
	}
}
```

I agree that it should be redundant to put the attributes on both, but I think that's best fixed in a different PR.

rolfbjarne:
> I wonder if those conflicts with devices

I ran introspection on my iOS 8.2 device and the 64-bit version worked (the 32-bit version failed to build due to the 32-bit size issue).

rolfbjarne:
build

src/HealthKit/EnumsCompat.cs
spouliot: ok, there must be a reason (what the IDE display) for the duplication. Better find it (again) before fixing it :)

rolfbjarne:
The binary mlaunch must be updated: https://github.com/xamarin/macios-binaries/pull/18.

rolfbjarne:
Test failure is unrelated:

* mmptest: https://github.com/xamarin/maccore/issues/1597

PR #6011, opened 2019-05-08
[msbuild] Add reference to `System.Drawing.Common.dll` to XI projects.

radical:
Fixes https://github.com/mono/mono/issues/13483 :

```
@akoeplinger: Since we moved types from Mono.Android.dll and
Xamarin.iOS/WatchOS/TVOS.dll to System.Drawing.Common.dll user projects
would fail to compile. We need to add some msbuild logic to add a
reference to the assembly automatically.
```


radical:
> @radical is there any way to for customers to undo/ignore this fix locally (in their csproj) if it ends up causing problems for someone?

@rolfbjarne Not as is, but this could be made conditional on some new property that the user can set to disable this behavior.

spouliot:
bot issues

```
Java NullPointerException in logThis likely indicates an error in the Jenkins slave processIndication 1
--
  | MicrosoftAzureStorage - Upload ErrorMicrosoftAzureStorage - Error occurred while uploading to AzureIndication 2
```

spouliot:
build

spouliot:
> 6 tests failed, 0 tests skipped, 228 tests passed
5x watchOS failures due to HE0038 https://github.com/xamarin/maccore/issues/581
1x failure due to expired certificates https://github.com/xamarin/maccore/issues/1597

rolfbjarne:
> this could be made conditional on some new property that the user can set to disable this behavior.

Ok, I've added a `DisableAutomaticSystemDrawingCommonReference` variable that can be set (if anyone has a better name I'm open to suggestions).

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
```diff
@@ -458,6 +458,13 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		<Delete SessionId="$(BuildSessionId)" Condition="'$(IsMacEnabled)' == 'true'" Files="@(_IpaPackageFile)" />
 	</Target>
 
+	<Target Name="_AddExtraReferences" BeforeTargets="ResolveAssemblyReferences" Condition="'$(DisableAutomaticSystemDrawingCommonReference)' == ''">
```
emaf: Same thing than in the Mac common targets

msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
```diff
@@ -186,6 +186,13 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		<RemoveDir Directories="$(IntermediateOutputPath)" />
 	</Target>
 
+	<Target Name="_AddExtraReferences" BeforeTargets="ResolveAssemblyReferences" Condition="'$(DisableAutomaticSystemDrawingCommonReference)' == ''">
```
emaf: I think `DisableAutomaticSystemDrawingCommonReference` should have a generic name to match the target name like `DisableExtraReferences` or `DisableAutomaticExtraReferences`.

Besides that, since it's a boolean property it'd be better give it a default value if it's empty (set it to false) or check if it's != 'true', because if someone sets it to false for any reason the target will be skipped when it shouldn't.

rolfbjarne: @emaf ok, I've changed it to use `DisableExtraReferences` and check for (not) `true` instead of empty string.

rolfbjarne:
@monojenkins backport d16-2

PR #6019, opened 2019-05-08
[VSTS] Ensure that the corrent tests are run on devices.

mandel-macaque:
The change allows to state the tests that have to be ran. ATM with these
changes, the vsts pipeline must add the following to the env vars:

* tvOS device pipelines: Must add 'run-tvos-tests' to the labels.
* iOS device pipelines: Must add 'run-ios-tests' to the labels.

This will ensure that only the tests for the devices are ran and if the
tests pass we get a green build with no unexpected skips.


VincentDondain:

PR #6020, opened 2019-05-08
[XHarness] List the simulators and devices in the html.

mandel-macaque:


rolfbjarne:
Maybe show the number of simulators instead of the entire list?

Screenshot 2019-05-09 07 24 53

spouliot:
a collapsing list would do too (but it's the device more than the sim that are interesting)

PR #6028, opened 2019-05-09
[mtouch/mmp] Make the mono native mode a Target-specific variable.

rolfbjarne:
The arm64_32 slice for watchOS apps will always use the 'unified' mode, while
the armv7k can be both 'unified' and 'compat' depending on the deployment
target, so we need to keep track of this per Target.

This PR does not change anything related to arm64_32, that will come in a
later PR.


rolfbjarne:
@monojenkins backport d16-2

PR #6030, opened 2019-05-09
[xharness] Flush the log after getting results from listing simulators/devices with mlaunch.

rolfbjarne:
Now constantly refreshing the device/simulator listing log will show the
output as mlaunch writes it, which makes the impatient me happier.


rolfbjarne:
Test failure is unrelated

* monotocuh-test/iOS 64-bits: https://github.com/xamarin/maccore/issues/1595

PR #6031, opened 2019-05-09
[xharness] Show links to previous test runs in html report in server mode.

rolfbjarne:
Show links to previous test runs in html report in server mode, which makes it
much easier to see what failed in previous test runs.

This required changing the URL scheme a little bit: the URL used to access the
local server now includes the immediate parent directory name (instead of
http://localhost/index.html it's something like
http://localhost/20190509_154249/index.html), so that the local server can
serve index.html files from other test runs.


rolfbjarne:
This is how it looks like (when hovering over "Previous test runs"):

Screenshot 2019-05-09 16 31 11

PR #6034, opened 2019-05-09
[Jenkins] Do not run device tests in the internal jenkins instance.

mandel-macaque:


rolfbjarne:
@monojenkins backport d16-2

PR #6060, opened 2019-05-16
[XHarness] Show when we have a watch HE0038 error.

mandel-macaque:
To simplify the life of the monitorer, if we get a crash with a HE0038
we will propagate the result to the html and will provide a link to the
issue so that it is easy to report it.


mandel-macaque:
Does not fix https://github.com/xamarin/maccore/issues/581 but simplifies the monitoring.

PR #6066, opened 2019-05-16
[xharness] 'Skipped' really meant 'DeviceNotFound', so rename things to avoid confusion.

rolfbjarne:


rolfbjarne:
The 'device not found' errors are expected (they're shown because I added the `run-device-tests` label so that the bot would try to run device tests, fail, and show those errors to ensure failures look as desired).

rolfbjarne:
@monojenkins backport d16-2

PR #6083, opened 2019-05-20
[XHarness] Remove the old style bcl tests.

mandel-macaque:
Remove all the old style bcl tests but the mscorlib ones since they are not yet running properly on Mac OS X.


mandel-macaque:
All crashes are due to the HE0038.

PR #6084, opened 2019-05-20
[apidiff] Fix a few issues with the apidiff when the mono hash or api diff url changes.

rolfbjarne:
* Change the name of the unzip stamp and download dir to contain the hash.
* Add rule to get mono-api-info.exe and mono-api-html.exe.
* Make make not delete temporary files.


rolfbjarne:
This isn't quite right, the api diff should be empty, not report breaking changes 😒

PR #6089, opened 2019-05-21
Fix ownership of CFSocketCreateRunLoopSource

Therzok:


Therzok:
I just noticed something. `CFSocketInvalidate` is not called, so the runloopsource is never removed. So the native leak of CFSocket will still be there. Should we bind CFSocket Invalidate to allow users to invalidate the socket before disposal?

Therzok:
There's a few other cans of worms I won't open in this PR:
a) The info pointer should actually contain a release call to the GCHandle, not Dispose
b) The info pointer is actually the wrong type.

src/CoreFoundation/CFSocket.cs
```diff
@@ -209,9 +212,62 @@ static byte[] CreateData (IPEndPoint endpoint)
...
+	}
+
 	public class CFSocket : CFType, INativeObject, IDisposable {
 		IntPtr handle;
 		GCHandle gch;
+		int retainCount = 1;
+
+		internal void Retain ()
+		{
+			++retainCount;
```
rolfbjarne: This should use `Interlocked.Increment`

src/CoreFoundation/CFSocket.cs
```diff
@@ -209,9 +212,62 @@ static byte[] CreateData (IPEndPoint endpoint)
...
+		}
+
+		internal void Release ()
+		{
+			if (--retainCount == 0 && gch.IsAllocated) {
```
rolfbjarne: And `Interlocked.Decrement`.

src/CoreFoundation/CFSocket.cs
```diff
@@ -300,90 +355,68 @@ public CFSocket (AddressFamily family, SocketType type, ProtocolType proto)
...
-				Marshal.FreeHGlobal (ptr);
-			}
-
-			if (handle == IntPtr.Zero)
-				throw new CFSocketException (CFSocketError.Error);
-			gch = GCHandle.Alloc (this);
-
-			var source = new CFRunLoopSource (CFSocketCreateRunLoopSource (IntPtr.Zero, handle, 0));
-			loop.AddSource (source, CFRunLoop.ModeDefault);
+			var ctx = new CFSocketContext (this);
```
rolfbjarne: You can move the creation of the `CFSocketContext` into the `Initialize` method as well, it's identical in all three cases.

tests/monotouch-test/CoreFoundation/SocketTest.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace MonoTouchFixtures.CoreFoundation {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class CFSocketTest {
+		static readonly byte[] dataToSend = { 0, 1, 2, 3 };
+
+		[Test]
+		public void RetainCount ()
```
rolfbjarne: I think it's also possible to create a test that checks if the CFSocket instance is actually collected by the GC:

1. Create several CFSockets on a background thread, create a weak gchandle to each, and then dispose them.
2. On the main thread, wait for the background thread to finish.
3. Run the GC a few times.
4. Assert that at least one of the weak gchandles doesn't have a target anymore.

This test should ensure that the refcounting logic for the context is correct.

src/CoreFoundation/CFSocket.cs
Therzok: How? We depend on the handle form the PInvoke. And the pinvoke depends on the struct.

tests/monotouch-test/CoreFoundation/SocketTest.cs
Therzok: Did manual tracing on the unit test, it was collecting them. But sure, can add that.

src/CoreFoundation/CFSocket.cs
rolfbjarne: Ah duh, nvm.

Therzok:
Thanks for taking over :heart:

rolfbjarne (approved):

dalexsoto (approved):
❤️

src/CoreFoundation/CFSocket.cs
```diff
@@ -272,28 +350,31 @@ static void OnCallback (IntPtr s, nuint type, IntPtr address, IntPtr data, IntPt
...
+				using (var cfdata = new CFData (data, false)) {
+					if (cfdata.Length > 0) {
+						var ep = CFSocketAddress.EndPointFromAddressPtr (address, false)!;
+						socket.OnData (new CFSocketDataEventArgs (ep, cfdata.GetBuffer ()));
+					}
+				}
 			} else if (cbType == CFSocketCallBackType.NoCallBack) {
 				// nothing to do
-			} else if (cbType == CFSocketCallBackType.ReadCallBack) {
+			} else if (cbType == CFSocketCallBackType.ReadCallBack && socket.ReadEvent != null) {
```
mandel-macaque, line 362:
```suggestion
			} else if (cbType == CFSocketCallBackType.ReadCallBack && socket.ReadEvent is not null) {
```

src/CoreFoundation/CFSocket.cs
```diff
@@ -272,28 +350,31 @@ static void OnCallback (IntPtr s, nuint type, IntPtr address, IntPtr data, IntPt
...
 				socket.OnRead (new CFSocketReadEventArgs ());
-			} else if (cbType == CFSocketCallBackType.WriteCallBack) {
+			} else if (cbType == CFSocketCallBackType.WriteCallBack && socket.WriteEvent != null) {
```
mandel-macaque, line 364:
```suggestion
			} else if (cbType == CFSocketCallBackType.WriteCallBack && socket.WriteEvent is not null) {
```

mandel-macaque (approved):

PR #6090, opened 2019-05-21
[CFUrl] Fix leaks of CFUrl instances.

Therzok:
Create methods give ownership of the object, so in this case, owns=true. Prevents leak of CFUrl


rolfbjarne:
Test failure is unrelated:

* monotouch-test/iOS 64-bits/Debug (LinkSdk): https://github.com/xamarin/maccore/issues/1595

PR #6114, opened 2019-05-23
[xharness] Don't use a dash in the bundle identifer for watchOS projects.

rolfbjarne:
It causes problems with the mscorlib test project, which can't be launched properly.

I'm not sure what's the underlying cause, but here are some of the symptoms:

* The watch app actually shows up fine on the device, but:
* mlaunch isn't notified about the new process, so it thinks the app didn't
  launch.
* The new process doesn't receive any environment variables we try to give it,
  which for instance means that it won't auto-start the tests upon launch.
* If we ask mlaunch to attach with lldb, mlaunch will ask watchOS to launch
  the process in a suspended state while lldb attaches. Yet the watch app
  shows up on the device as if not asked to be suspended upon launch.

It seems that the dash (I assume, because I haven't investigated this very
deeply, I just happened to find a solution that worked) makes watchOS launch
the app as if tapped, instead of launched from an IDE.

The strangest part is that this only happens with the mscorlib test project,
not any of the other test projects we run on the watch, and they all have
dashes in their bundle identifiers... yet replacing the dash with another
character (underscore, letter, removing it altogether) all made things work as
expected.

### DIFF
--- a/tests/xharness/WatchOSTarget.cs
+++ b/tests/xharness/WatchOSTarget.cs
@@ -154,7 +154,7 @@ namespace xharness
 			XmlDocument info_plist = new XmlDocument ();
 			var target_info_plist = Path.Combine (TargetDirectory, $"Info{Suffix}-extension.plist");
 			info_plist.LoadWithoutNetworkAccess (Path.Combine (TargetDirectory, "Info.plist"));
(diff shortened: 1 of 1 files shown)


rolfbjarne:
@monojenkins backport d16-2

PR #6115, opened 2019-05-23
[tests] Don't make the mscorlib tests exit upon suspend.

rolfbjarne:
It means that on watchOS they'll exit if the screen turns off.

I have no idea why we set it up like this in the first place, it's been like
this since we created the mscorlib tests ([1]).

[1]: https://github.com/xamarin/maccore/commit/76e15036e8d9f27281d1398d565feaf32bfebd8b


mandel-macaque:
@rolfbjarne we need to kill this tests if we do not want to build mono (use the downloaded sdk and trust the xunit corelib tests). I'll create a PR with a similar change for the BCL tests.

rolfbjarne:
Test failures are both due to https://github.com/xamarin/maccore/issues/581.

PR #6116, opened 2019-05-23
[xharness] Don't try to find simulators for ignored tests.

rolfbjarne:
Trying to find a simulator will mark the test as a failure if the simulator
couldn't be found, and we don't want that to happen to ignored tests.

This should fix an issue where xharness seems to try to run the 32-bit
simulator tests when asked to run only device tests.


rolfbjarne:
@monojenkins backport d16-2

rolfbjarne:
@monojenkins backport xcode11

mandel-macaque:
@monojenkins backport d16-3

PR #6121, opened 2019-05-23
[XHarness] Re-enable corlib tests on Mac Full and Modern

mandel-macaque:
Tests are readded from the test dll from mono (xunit).

Fixes: https://github.com/xamarin/maccore/issues/1203


tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -228,7 +228,6 @@ public enum WatchAppType {
 			(assembly: "xammac_net_4_5_Mono.Messaging_test.dll", platforms: new [] { Platform.MacOSModern}), // not present 
 			(assembly: "xammac_net_4_5_System.Data_test.dll", platforms: new [] { Platform.MacOSModern }), // tests use 'System.Configuration.IConfigurationSectionHandler' not present in modern 
 			(assembly: "xammac_net_4_5_System.Configuration_test.dll", platforms: new [] { Platform.MacOSModern }), // Not present in modern, ergo all tests will fail
```
akoeplinger: FWIW looks like we're still using the xammac_net_4_5 assemblies on Modern?

mandel-macaque: We don't, the importer does the right thing and ensures that the assembly does not have _net_t_45 when used in modern :)  https://github.com/xamarin/xamarin-macios/blob/master/tools/bcl-test-importer/BCLTestImporter/BCLTestAssemblyDefinition.cs#L50

mandel-macaque:
@spouliot will make sure they are!

mandel-macaque:
Both crashes on watch are due to https://github.com/xamarin/maccore/issues/581 merging.

PR #6124, opened 2019-05-24
Fix case sensitivity of MSBuildSDKsPath

Therzok:
This would result in locally for me. Seems like `SelectSingleNode` is case sensitive?

```
MSBUILD : Configuration error MSB4136: Error reading the toolset information from the configuration file "System.Configuration". Error deserializing configuration section msbuildToolsets: Multiple definitions were found for the property "MSBuildSDKsPath".
```


Therzok:
Interesting, the bots have the exact opposite of my local behaviour.

PR #6129, opened 2019-05-24
[xharness] A test fails if there are either errors or failures.

rolfbjarne:


rolfbjarne:
All test failures are known and will be fixed with PR #6128.

PR #6130, opened 2019-05-24
[Tests] Ensure that if trust exception is not raised we do not get a NRE

mandel-macaque:
This might be a race issue when dealing with the trust of the certs,
there is not much information in
https://github.com/xamarin/maccore/issues/1645 except the fact that we
are getting a null exception.

In this fix, we will first ensure that we did get the exception, later
confirm the exception type. To improve the debugging of the test, next
time it fails, we will get the content string to try and understand what
happened with the validation (did we return part of the stream or
headers when we shouldn't?)

### DIFF
--- a/tests/monotouch-test/System.Net.Http/MessageHandlers.cs
+++ b/tests/monotouch-test/System.Net.Http/MessageHandlers.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // MessageHandlers.cs
 //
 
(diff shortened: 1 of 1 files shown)


mandel-macaque:
Failures are related to https://github.com/xamarin/xamarin-macios/pull/6128

mandel-macaque:
@monojenkins backport to 16-2

PR #6132, opened 2019-05-27
[XHarness] Ignore a number of assemblies that do not have any tests.

mandel-macaque:
Mono did remove all the tests in a number of assemblies on
ios/tvos/watchos. With a recent change, we report test runs with no
tests as a failure (correctly since it will bring up issues with the
runners).

In this case, the tree assemblies have to be ignored because they truly
do not have tests and the runners are doing the right thing.

Fixes: https://github.com/xamarin/maccore/issues/1652


mandel-macaque:
Timeout is an already known issues: https://github.com/xamarin/maccore/issues/1651

PR #6137, opened 2019-05-27
[xharness] Timestamp all the things.

rolfbjarne:
I find myself wanting timestamps if I don't have them much more than wanting
them gone when I do.

Worst case scenario: timestamps can still be disabled for specific logs if
such a desire ever becomes overwhelming.


rolfbjarne:
build

PR #6141, opened 2019-05-27
[xharness] Remove mono tests built locally in favor of binary archive binaries

mandel-macaque:
Remove the old style test and replace it with the xunit equivalent which
has more tests and is provided by the mono package.

This is a draft since I have to test it on a 32b watchOS device to make sure we do not need to split the mscorlib tests into two.


rolfbjarne:
LGSF.

mandel-macaque:
@spouliot it is the old (we compile the tests) to the new (we got the test dlls) + using the xunit. But we could simply do the same changes and use the old NUnit ones, that is change line: https://github.com/xamarin/xamarin-macios/blob/master/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs#L103 instead and make it its own group. 

The rest of the changes are needed nevertheless to allow the name of the project to be mscorlib and have all the other logic work correctly.

mandel-macaque:
We have talked about this on the mono binaries sync, we will ignore the tests on watchOS 32 since the test assembly is too large.

mandel-macaque:
Xharness should show the following configs for mscorlib:

Screenshot 2019-05-28 at 17 17 32

tests/xharness/Harness.cs
```diff
@@ -307,12 +306,9 @@ void AutoConfigureIOS ()
 			var fsharp_test_suites = new string [] { "fsharp" };
 			var fsharp_library_projects = new string [] { "fsharplibrary" };
 			var bcl_suites = new string [] {
```
chamons: Should we just remove this code at some point?

mandel-macaque: Yes, next PR, I'm going to clean everything

mandel-macaque:
@akoeplinger no need, we removed the special case in the switch so we point to tests/bcl-test/BCLTests/mscorlib.csproj  which is generated and present

PR #6143, opened 2019-05-27
[Harness] Pass extra arguments to configure the linker for certain tests

mandel-macaque:
Some tests need a special linker configuration. Add the extra arguments
so that tests do pass on device.

Fixes: https://github.com/xamarin/maccore/issues/1608


tests/bcl-test/BCLTests/SystemXunitLinker.xml
```diff
@@ -0,0 +1,36 @@
+<linker>
+	<assembly fullname="System">
+		<type fullname="System.Text.RegularExpressions.CollectionDebuggerProxy`1" preserve="all" />
+    </assembly>
```
rolfbjarne: The indentation is off in this file (spaces vs tabs mixup?)

PR #6148, opened 2019-05-28
[generator] Check if an object/pointer is null before trying to call release on it.

rolfbjarne:
Note that the 'needsPtrZeroCheck' variable is not accurately named anymore,
but given that the code is somewhat impenetrable I wasn't able to come up with
a better name.

Also simplify two consecutive and identical null checks to a single check.


mandel-macaque:
Any way to add a test for this?

rolfbjarne:
@monojenkins backport d16-2

PR #6149, opened 2019-05-28
[xtro] Add new checker for '[return: Release]' attributes. Fixes #6014.

### LINKED ISSUE #6014: Add introspection test for [return: Release] attribute
### Steps to Reproduce

1. Read about [Objective-C method families](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#arc-method-families)
2. Observe that `[_]*new[A-Z]` selectors indicate that the returned value is already retained.
3. Look in our bindings, and see that we've bound many such selectors, but none of them have the `[return: Release]` attribute to tell the generator and our runtime that the return value is already retained. Example: https://github.com/xamarin/xamarin-macios/blob/62f0d294293ff53496bdaa6530d6356af97e9954/src/metal.cs#L795-L799

We should add an introspection test to verify that all the Objective-C method families are handled correctly (`new`, `alloc`, `copy`, `mutableCopy`,  and `init`).

We might want to look into making Objective-Sharpie spit out this attribute as well.

rolfbjarne:
Also add a 'u2ignore' tool that can copy the resulting unclassified (all 1612
of them) to the ignore files.

Fixes https://github.com/xamarin/xamarin-macios/issues/6014.


tests/xtro-sharpie/ReleaseAttributeCheck.cs
```diff
@@ -0,0 +1,127 @@
...
+					Log.On (framework).Add ($"!missing-release-attribute-on-return-value! {method.FullName}'s selector's ('{selector}') Objective-C method family ('{family}') indicates that the native method returns a retained object, and as such a '[return: Release]' attribute is required.");
+				}
+				break;
+			default:
+				break;
+			}
+
+		}
+
+		// We should also look at the native definition for family attributes: __attribute__((objc_method_family(...))
```
spouliot: filed an issue ?

rolfbjarne: no, I'm already fixing it in Objective-Sharpie, but it takes a while to get it in and released, so I thought I should just get the rest in first.

rolfbjarne:
Test failure is unrelated

rolfbjarne:
@monojenkins backport d16-2

PR #6152, opened 2019-05-28
[XHarness] Remove the mini tests since they are not longer needed.

mandel-macaque:
The tests in the mini project are the same ones found in
monotouch_Mono.Runtime.Tests_test.dll which is already ran via the
autogenerated bcl tests.

Mini tests therefore are redundant and require the mono sources which
will be removed.


tests/xharness/Jenkins.cs
```diff
@@ -676,10 +676,6 @@ void DisableKnownFailingDeviceTests ()
 		{
 			// https://github.com/xamarin/maccore/issues/1008
 			ForceExtensionBuildOnly = true;
-
```
akoeplinger: mini is still mentioned in this file on L318: https://github.com/xamarin/xamarin-macios/blob/5108f5659b4c680f825d94df0eb4f25e75d246e4/tests/xharness/Jenkins.cs#L318-L328

tests/mini/mini.csproj
```diff
@@ -1,176 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
akoeplinger: the csproj is still used in monotouch-tests and in tests.sln.

tests/xharness/Jenkins.cs
mandel-macaque: @akoeplinger nice catch, I'll remove that

PR #6158, opened 2019-05-29
[xtro] Implement support for auto-sanitizing the .ignore files by defining an environment variable.

rolfbjarne:
Going through each failure one-by-one after fixing something that fixed many
entries is annoying, so automate the process.

### DIFF
--- a/tests/xtro-sharpie/xtro-sanity/Sanitizer.cs
+++ b/tests/xtro-sharpie/xtro-sanity/Sanitizer.cs
@@ -189,6 +189,7 @@ namespace Extrospection {
 					else
 						raws [i] = new List<string> ();
 				}
+				var failures = new List<string> ();
 				foreach (var entry in common) {
 					if (!entry.StartsWith ("!", StringComparison.Ordinal))
 						continue;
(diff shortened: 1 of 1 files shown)


tests/xtro-sharpie/xtro-sanity/Sanitizer.cs
```diff
@@ -212,12 +221,21 @@ public static int Main (string [] args)
 					continue;
 				var rawfile = Path.ChangeExtension (file, ".raw");
 				var raws = new List<string> (File.ReadAllLines (rawfile));
-				foreach (var entry in File.ReadAllLines (file)) {
+				var failures = new List<string> ();
+				var lines = File.ReadAllLines (file);
```
mandel-macaque: We used to get issues when doing ReadAllLines in large files in xharness results. Can we use a reader instead?

spouliot: They are not the big - but an easier change is to use `foreach (var line in File.ReadLines(file))` which will read line-by-line (in memory)

spouliot:
@mandel-macaque the `File` class has easier API to do the same. It's also a change that would be needed in several places (in Xtro) and likely better in a different PR

mandel-macaque:
@spouliot fair enough, I just don't want to hit the memory issues we had in xharness, will approve. Should we file an issue?

rolfbjarne:
Yes, these files are quite small, and mostly hand-maintained, so they shouldn't grow so big they cause problems (if they do, we have even bigger problems).

spouliot:

Issue #6157, opened 2019-05-29
Xamarin.MTouch.BuildTestProject test failure in binary-artifacts branch


akoeplinger:
### Steps to Reproduce

1. do a git clean in tests/
2. comment out these lines so only mscorlib and BCL tests group 1 remain: https://github.com/xamarin/xamarin-macios/blob/b5958e89c4f620139b5145859c253e24eae996e8/tests/mtouch/MTouch.cs#L2552-L2555
2. run `make run-tests TEST_FIXTURE=-test=Xamarin.MTouch.BuildTestProject` in tests/mtouch
3. Observe test failure

### Expected Behavior

No failure.

### Actual Behavior

```
MTOUCH : error MT4116: Could not register the assembly 'com.xamarin.bcltests.BCL tests group 1':
error MT4118: Cannot register two managed types ('BCLTests.AppDelegate, com.xamarin.bcltests.BCL tests group 1' and 'BCLTests.AppDelegate, com.xamarin.bcltests.mscorlib') with the same native name ('AppDelegate').
```

I assume this is somehow related to mscorlib.csproj being built before BCL tests group 1 ?

rolfbjarne:
This fails the MTouch tests in PR as well.

mandel-macaque:
The BCLTests are sharing a dir, ergo some dlls are left after the ran, is as simple as adding "/t:Clean,Build" to make sure we do clean before. Got a branch running them to make sure and will propose the PR.

PR #6161, opened 2019-05-29
[XHarness] Do clean after mtouch build tests.

mandel-macaque:
The BCL projects do share the directory. This means that if we do not
clean before building we could have unknown side effects.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6157


rolfbjarne:
Test failure is unrelated:

* SystemXmlLinqTests/watchOS: https://github.com/xamarin/maccore/issues/581

Issue #6162, opened 2019-05-29
[Tests] Bcl test applications should be generated in their own test dir.


mandel-macaque:
The bcl tests generator is not creating a dir per testing application, this is an issue if we use tests (like the mtouch tests do) that use the project for testing. 

The generator has to create a dir per test application, that will allow to run the mtouch tests in parallel and not have issues like https://github.com/xamarin/xamarin-macios/issues/6157

Fixing the issue requires two different steps:

1. Remove the BCLTests directory. The directory is not longer needed since all the old style (compile from source) tests are not longer present.
2. The bcl tests generator use a directory per test app using the test project name as the name of the directory.

mandel-macaque:
Closed when it should not.

rolfbjarne:
We don't run BCL tests anymore.

PR #6165, opened 2019-05-29
[XHarness] Clean the bcl-tests directory.

mandel-macaque:
The change is a pre step to fix https://github.com/xamarin/xamarin-macios/issues/6162

We are removing the not longer child directory bcl-test/BCLTests since
there is not longer any reason for it to exist. The changes make sure
that the tests work as expected in the new directory.

Next step, move all the test apps to their own directory.


mandel-macaque:
Do not trust the files changed counter in github, most of them are moves, don't hate me.

mandel-macaque:
Errors are HE0038.

PR #6170, opened 2019-05-30
[tests] Run tests on watchOS in a background thread.

rolfbjarne:
Otherwise watchOS might terminate the process if we lock up the UI too long
while running tests.

It's also nice to see the progress as the tests are running.


rolfbjarne:
Test failure is unrelated

* mscorlib/iOS 32-bit: https://github.com/xamarin/maccore/issues/1659

PR #6178, opened 2019-05-30
[tests] Set environment variables so that the in-process generator finds the locally built platform assemblies.

rolfbjarne:
When running the in-process generator, we need to set the same environment
variables as we already do when running the out-of-process generator, so that
the generator looks for the platform assemblies in the local install directory
(and not the system one).


rolfbjarne:
Test failure is unrelated.

* Generator tests: https://github.com/xamarin/maccore/issues/1661

PR #6183, opened 2019-05-30
[tests] Ignore BaseOptimizeGeneratedCodeTest.SetupBlockPerfTest when doing CI. Fixes xamarin/maccore#649.

rolfbjarne:
The BaseOptimizeGeneratedCodeTest.SetupBlockPerfTest test is randomly failing
fairly often now, which means it turns CI builds red.

So disable it, but since I don't like disabling tests I've only disabled it
when doing CI. Hopefully we'll find out if there are any regressions when
running tests locally.

Fixes https://github.com/xamarin/maccore/issues/649.

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -103,6 +103,14 @@ partial class TestRuntime
 		return new Version (major, minor, build);
 	}
 
+	public static void IgnoreInCI (string message)
+	{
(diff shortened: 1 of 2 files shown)


rolfbjarne:
build

rolfbjarne:
The api diff should be fixed once the bcl test changes aren’t a difference between the PR and master.

PR #6191, opened 2019-05-31
[xtro] 'raw' files might not exist.

rolfbjarne:
If we just fixed all the entries in the corresponding '*.ignore' file, there
won't be a '*.raw' file.


spouliot:

PR #6193, opened 2019-05-31
[xharness] Improve marking and reporting of known failures in tests.

rolfbjarne:
* Add support for marking a test as a known failure and show the known failure
  text in the html report.

* Adjust HE0038 reporting to the new known failure logic.

* Don't run known failures when trying to run a collection of tests ('Run all'
  button).

* Mark the mscorlib interpreter tests as known failures.

* The interpreter has been fixed for monotouch-test, so we can unignore those
  tests. Fixes xamarin/maccore#1613.


rolfbjarne:
Api diff failure should go away once the bcl test changes isn’t a difference between PRs and the tip of master.

PR #6198, opened 2019-06-01
[XHarness] Bcl projects should use their own project directory

mandel-macaque:
BCL projects were sharing all the same directory, this means that there was a number of possible side effects when building more than one project without doing a /t:Clean,Build. The changes make sure that:

1. All bcl projects have their own dir.
2. Moved all the paths logic to its own class. Means it will later be easier to change the paths if needed.
3. Simplified the project generation methods.
4. Updated the mtouch tests to take into account the new location.

This is huge change, mtouch and bcl tests should be ran.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6162


mandel-macaque:
New tree structure of the tests is as follows: https://gist.github.com/mandel-macaque/7b9607ee9c65e81843f7d0f8fa7664f9

mandel-macaque:
@spouliot @rolfbjarne I would back port this to xcode11. Let me know if you have any issues with it.

tools/bcl-test-importer/BCLTestImporter/BCLPathManager.cs
```diff
@@ -0,0 +1,256 @@
...
+		/// Returns the path to the template to be used for the given watch app type.
+		/// </summary>
+		/// <returns>The path to the appropiate path.</returns>
+		/// <param name="appType">The type of watchOS application.</param>
+		public string GetPListTemplatePath (WatchAppType appType) => Path.Combine (PlistTemplateRootPath, watchOSPlistTemplateMatches [appType]);
+		
+		public string GetPListPath (WatchAppType appType)
+		{
+			switch (appType) {
+				case WatchAppType.App:
```
rolfbjarne: `case` statements should use the same indentation as the `switch`:

```csharp
switch (foo) {
case a:
    code ();
    break ();
default:
    break ();
}
```

mandel-macaque:
Closing for a working implementation here: https://github.com/xamarin/xamarin-macios/pull/6680

Issue #6195, opened 2019-05-31
CFProxy.ExecuteProxyAutoConfigurationScript API not public 


alexnesh:
I've been working with @mandel-macaque to address some issues with a client related to proxy configurations in their tenant. 

Specifically, we've been using the `CFNetwork.GetProxiesForUri` API in order to retrieve the system proxies and apply them to our web sockets. However, the sole proxy that we get back from this API is of type `CFProxyType.AutoConfigurationUrl`, for which the Hostname is, by design, null, so we cannot simply use this to construct our WebProxy. From the documentation I've read, it seems that we need to execute the PAC file located at this Url to retrieve the proxy results that we should apply to our web socket. 

Apple has the `CFNetworkExecuteProxyAutoConfigurationURL` API, and it looks like Xamarin has interop to expose this API in the form of `CFProxy.ExecuteProxyAutoConfigurationScript`. However, looking at `CFProxySupport.cs` in the xamarin-mac-5.10.0.147 tag, these API's are wrapped in an `#if notyet` block as they require bindings for `CFRunLoopSource` and `CFStreamClientContext`. Although it's possible I've missed something, it looks like these bindings exist in this version of Xamarin, so I'm inquiring as to whether or not this API is now safe to be made public.

PR #6201, opened 2019-06-03
[CFNetwork] Expose methods that were ignored until the dependencies were present.

mandel-macaque:
We can now expose the methods. Bindings had to be updated considering
the fact that we have a scrut and the type of callbacks.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6195


mandel-macaque:
Added it as a draft since I don't want to land it until I have written a test case, but I want the reviews to ensure that my train of though is correct.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,16 +580,23 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			}
+		}
 		
 		[DllImport (Constants.CFNetworkLibrary)]
-		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
+		extern unsafe static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
 			/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ void* cb,
```
rolfbjarne: Use `ref BlockLiteral` and you can remove the `unsafe` keyword.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -602,19 +609,27 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString prox
...
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+			unsafe {
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
+				block_handler.SetupBlockUnsafe (static_AutoConfigurationResult, resultCallback);
+		
+				try {
+					IntPtr clientPtr = Marshal.AllocHGlobal (Marshal.SizeOf (clientContext));
+					Marshal.StructureToPtr (clientContext, clientPtr, false);
+					IntPtr source = CFNetworkExecuteProxyAutoConfigurationScript (proxyAutoConfigurationScript.Handle, targetURL.Handle, (void*) block_ptr_handler, clientPtr);
```
rolfbjarne: Do: `ref block_handler` instead of `(void*) block_ptr_handler` and you can remove the `block_ptr_handler` variable and the `unsafe` block.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -602,19 +609,27 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString prox
...
-
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationScript (proxyAutoConfigurationScript.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+			unsafe {
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
+				block_handler.SetupBlockUnsafe (static_AutoConfigurationResult, resultCallback);
+		
+				try {
+					IntPtr clientPtr = Marshal.AllocHGlobal (Marshal.SizeOf (clientContext));
```
rolfbjarne: Where is this memory freed? Seems like you can pass the `clientContext` variable directly to the P/Invoke if you make it a ref parameter:

```csharp
extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
				/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
				/* CFURLRef __nonnull */ IntPtr targetURL,
				/* CFProxyAutoConfigurationResultCallback __nonnull */ ref BlockLiteral cb,
				/* CFStreamClientContext * __nonnull */ ref CFStreamClientContext clientContext);

// ...

IntPtr source = CFNetworkExecuteProxyAutoConfigurationScript (proxyAutoConfigurationScript.Handle, targetURL.Handle, ref block_handle, ref clientContext);
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -602,19 +609,27 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString prox
...
+			}
 		}
 		
 		[DllImport (Constants.CFNetworkLibrary)]
-		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
+		extern unsafe static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
 			/* CFURLRef __nonnull */ IntPtr proxyAutoConfigurationURL,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ void* cb,
```
rolfbjarne: Same: `ref BlockLiteral`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -627,14 +642,20 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationURL (NSUrl proxyAutoC
...
-				throw new ArgumentNullException ("clientContext");
-
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationURL (proxyAutoConfigurationURL.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+			unsafe {
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
+				block_handler.SetupBlockUnsafe (static_AutoConfigurationResult, resultCallback);
+				try {
+					IntPtr clientPtr = Marshal.AllocHGlobal (Marshal.SizeOf (clientContext));
```
rolfbjarne: Where is this memory freed?

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -627,14 +642,20 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationURL (NSUrl proxyAutoC
...
+					Marshal.StructureToPtr (clientContext, clientPtr, false);
+					IntPtr source = CFNetworkExecuteProxyAutoConfigurationURL (proxyAutoConfigurationURL.Handle, targetURL.Handle, (void*) block_ptr_handler, clientPtr);
```
rolfbjarne: Same: `ref block_handler`.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -602,19 +609,27 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString prox
...
 		[DllImport (Constants.CFNetworkLibrary)]
-		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
+		extern unsafe static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
 			/* CFURLRef __nonnull */ IntPtr proxyAutoConfigurationURL,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ void* cb,
 			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
 
 		public static CFRunLoopSource ExecuteProxyAutoConfigurationURL (NSUrl proxyAutoConfigurationURL, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
```
rolfbjarne: Casing: `ExecuteProxyAutoConfigurationUrl`

A signature like this would be much nicer:

```csharp
public delegate void AutoConfigurationResultCallback (NSArray proxyList, NSError error);
public static CFRunLoopSource ExecuteProxyAutoConfigurationUrl (NSUrl proxyAutoConfigurationUrl, NSUrl targetUrl, AutoConfigurationResultCallback callback)
{
    // ...
}
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,16 +580,23 @@ public static CFProxySettings GetSystemProxySettings ()
...
 		[DllImport (Constants.CFNetworkLibrary)]
-		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
+		extern unsafe static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
 			/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ void* cb,
 			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
 		
 		public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString proxyAutoConfigurationScript, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
```
rolfbjarne: A signature like this would be much nicer:

```csharp
delegate void AutoConfigurationResultCallback (NSArray proxyList, NSError error);
public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (string proxyAutoConfigurationScript, NSUrl targetUrl, AutoConfigurationResultCallback callback)
{
    // ...
}
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -602,19 +609,27 @@ public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString prox
 
 			if (resultCallback == null)
 				throw new ArgumentNullException ("resultCallback");
```
rolfbjarne: Nitpick: `nameof (resultCallback)` (same for all the other ones as well)

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,16 +580,23 @@ public static CFProxySettings GetSystemProxySettings ()
 			return new CFProxySettings (dict);
 		}
 		
-#if notyet
-		// FIXME: These require bindings for CFRunLoopSource and CFStreamClientContext
-		
 		public delegate void CFProxyAutoConfigurationResultCallback (IntPtr client, NSArray proxyList, NSError error);
```
alexnesh: @mandel-macaque and @rolfbjarne - is it possible for this callback to return a C# array of CFProxy rather than an NSArray? It would be quite a bit more convenient from the client side if we didn't have to parse the NSArray into the CFProxy array. Also, I've been testing this API with @mandel-macaque and am unsure how to parse the NSArray into the C# array of CFProxy - is this possible? Can one of you provide some guidance on this front?

rolfbjarne: We should be able to make it a managed array of something, but that something would probably be typed `NSObject`, since Apple doesn't explicitly state the type of the elements in the array. Then at runtime you can do a standard C# type check to check the actual type of the elements you get.

tests/monotouch-test/monotouch-test.csproj
```diff
@@ -365,7 +366,7 @@
     <Exec Command="xcrun -sdk $(_SmeltingSdk) metal -c @(CustomMetalSmeltingInput) -o $(IntermediateOutputPath)\fragmentShader.air $(_SmeltingMinOS)" />
     <Exec Command="xcrun -sdk $(_SmeltingSdk) metallib $(IntermediateOutputPath)\fragmentShader.air -o $(AppBundleDir)\fragmentShader.metallib" />
   </Target>
-  <Target Name="BeforeBuild" Inputs="@(GeneratedTestInput)" Outputs="@(GeneratedTestOutput)" DependsOnTargets="CustomMetalSmelting" >
+  <Target Name="BeforeBuild" Inputs="@(GeneratedTestInput)" Outputs="@(GeneratedTestOutput)" DependsOnTargets="CustomMetalSmelting">
```
rolfbjarne: Unnecessary whitespace change.

tests/monotouch-test/example.pac
```diff
@@ -0,0 +1,9 @@
+// Exmaple PAC file that returns a proxy for the urls that have the
+// xamarin domain name.
+function FindProxyForURL(url, host) {
+	if (dnsDomainIs(host, "xamarin.com"))
+		return "PROXY example.com:8080";
+	 
+	if (dnsDomainIs(host, "google.com"))
+		return "DIRECT";
+}
```
rolfbjarne: Add trailing newline.

tests/monotouch-test/CoreFoundation/ProxyTest.cs
```diff
@@ -42,5 +44,33 @@ public void Fields ()
...
+			var targetUri = new Uri ("http://docs.xamarin.com");
+			var proxies = CFNetwork.ExecuteProxyAutoConfigurationUrl (pacUri, targetUri);
+			Assert.AreEqual (1, proxies.Length, "Length");
+			// assert the data of the proxy, although we are really testing the js used
+			Assert.AreEqual (8080, proxies [0].Port, "Port");
+		}
+	}
+#endif
 	}
-}
```
rolfbjarne: Unnecessary whitespace change.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,61 +580,105 @@ public static CFProxySettings GetSystemProxySettings ()
...
-			if (clientContext == null)
-				throw new ArgumentNullException ("clientContext");
+				CFProxy[] proxies = null;
+				var runLoop = CFRunLoop.Current;
 
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationScript (proxyAutoConfigurationScript.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+				CFProxyAutoConfigurationResultCallback cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
+					if (proxyList != IntPtr.Zero) {
+						var array = new CFArray (proxyList, false);
```
rolfbjarne: `using (var array = new CFArray (proxyList, false)) {` avoids the manual Dispose later.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,61 +580,105 @@ public static CFProxySettings GetSystemProxySettings ()
...
-			if (clientContext == null)
-				throw new ArgumentNullException ("clientContext");
+				CFProxy[] proxies = null;
+				var runLoop = CFRunLoop.Current;
 
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationURL (proxyAutoConfigurationURL.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+				CFProxyAutoConfigurationResultCallback cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
+					if (proxyList != IntPtr.Zero) {
+						var array = new CFArray (proxyList, false);
```
rolfbjarne: `using (var array = new CFArray (proxyList, false)) {` avoids the manual Dispose later.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,61 +580,105 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					return null;
 
-			if (clientContext == null)
-				throw new ArgumentNullException ("clientContext");
+				CFProxy[] proxies = null;
+				var runLoop = CFRunLoop.Current;
 
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationScript (proxyAutoConfigurationScript.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+				CFProxyAutoConfigurationResultCallback cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
```
rolfbjarne: Any errors are ignored, those should be returned somehow.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -580,61 +580,105 @@ public static CFProxySettings GetSystemProxySettings ()
...
 		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
 			/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
-			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ CFProxyAutoConfigurationResultCallback cb,
+			/* CFStreamClientContext * __nonnull */ ref CFStreamClientContext  clientContext);
 		
-		public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString proxyAutoConfigurationScript, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
+		public static CFProxy[] ExecuteProxyAutoConfigurationScript (string proxyAutoConfigurationScript, Uri targetUrl)
```
rolfbjarne: While this overload is fine by itself, I think we should also provide a more advanced version that returns the CFRunLoopSource, so that users have more control over exactly which run loop to use, in particular because it's too easy to end up doing network requests on the main thread with this implementation, since it will execute synchronously on the current thread (which will most likely be the main thread).

Something like this maybe:

```csharp
public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (string script, Uri targetUrl, Action<CFProxy[], NSError> callback)
{
    // ...
}
```

another possible variation would be an async version:

```csharp
public static Task<CFProxy[]> ExecuteProxyAutoConfigurationScriptAsync (string script, Uri targetUrl)
{
    // ...
}
```

which schedules the runloop on a bacground/threadpool thread.

And then the same for `ExecuteProxyAutoConfigurationUrl` of course.

mandel-macaque: I'll provide both options, Task and CFRunLoopSource

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
-		
-		public delegate void CFProxyAutoConfigurationResultCallback (IntPtr client, NSArray proxyList, NSError error);
-		
+
+		delegate void CFProxyAutoConfigurationResultCallbackInternal (IntPtr client, IntPtr proxyList, IntPtr error);
+		public delegate void CFProxyAutoConfigurationResultCallback (NSObject client, CFProxy[] proxyList, NSError error);
+		// helper delegate to reuse code
+		delegate IntPtr CreatePACCFRunLoopSource (CFProxyAutoConfigurationResultCallbackInternal cb, ref CFStreamClientContext context);
+
+		static CFProxy[] ParseProxies (IntPtr proxyList) {
```
spouliot: minor/style: `{` on next line

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			CFProxy[] proxies = null;
+			if (proxyList != IntPtr.Zero) {
+				using (var array = new CFArray (proxyList, false)){
```
spouliot: minor/style: space before `{`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					proxies = new CFProxy [array.Count];
+					for (int i = 0; i < proxies.Length; i++) {
+						var dict = new NSDictionary (array.GetValue (i));
```
spouliot: you should use `CFDictionary` inside CoreFoundation'
even more if you think this code could move (like in mono) in the future

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		{
+			CFProxy[] proxies = null;
+			NSError outError = null;
+			await Task.Run (() => {
+				// we need the runloop of THIS thread, so it is important to get it in the correct context
+				var runLoop = CFRunLoop.Current;
+				CFProxyAutoConfigurationResultCallbackInternal cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
+					proxies = ParseProxies (proxyList);
+					if (error != IntPtr.Zero)
+						outError = new NSError (error);
```
spouliot: you parse the proxies even if there's an error - is that intentional ?

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					outError = new NSError (cbError);
+				return proxies;
+			}
+		}
+
+		static CFRunLoopSource ExecutePacCFRunLoopSourceBlocking (CreatePACCFRunLoopSource factory, CFProxyAutoConfigurationResultCallback clientCb, CFStreamClientContext clientContext)
+		{
+			CFProxyAutoConfigurationResultCallbackInternal cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
+				var proxies = ParseProxies (proxyList);
+				clientCb ((client != IntPtr.Zero)? null : new NSObject (client), proxies, (error == IntPtr.Zero) ? null : new NSError (error));
```
spouliot: `Runtime.GetNSObject` is likely better suited (for both `client` and `error`) since it won't create a 2nd managed instance over the same native one (if that's possible, not clear to me here)

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		}
+
 		[DllImport (Constants.CFNetworkLibrary)]
 		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationScript (
 			/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
-			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
-		
-		public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString proxyAutoConfigurationScript, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
```
spouliot: ^ `public` ? breaking change (or non public type?)

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		{
+			outError = null;
+			IntPtr cbError = IntPtr.Zero;
+			if (proxyAutoConfigurationScript == null)
+				throw new ArgumentNullException (nameof (proxyAutoConfigurationScript));
+
+			if (targetUrl == null)
+				throw new ArgumentNullException (nameof (targetUrl));
+
+			using (var pacScript = new NSString (proxyAutoConfigurationScript))
```
spouliot: minor: generated bindings do

```csharp
var p = NSString.CreateNative (proxyAutoConfigurationScript);
try {
...
}
finally {
   NSString.ReleaseNative (p);
}
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			if (proxyAutoConfigurationScript == null)
+				throw new ArgumentNullException (nameof (proxyAutoConfigurationScript));
+
+			if (targetUrl == null)
+				throw new ArgumentNullException (nameof (targetUrl));
+
+			using (var pacScript = new NSString (proxyAutoConfigurationScript)) 
+			using (var url = new NSUrl (targetUrl.AbsoluteUri)) {
+				if (url == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (url)}");
```
spouliot: `url` won't ever be `null` here since a `.ctor` can't return `null` - it will throw if invalid

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
```
spouliot: `url` won't ever be`null` here since a .ctor can't return null - it will throw if invalid

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			if (targetUrl == null)
+				throw new ArgumentNullException (nameof (targetUrl));
+			
+			if (clientCb == null)
+				throw new ArgumentNullException (nameof (clientCb));
+
+			using (var pacUrl = new NSUrl (proxyAutoConfigurationUrl.AbsoluteUri)) // toll free bridge to CFUrl
+			using (var url = new NSUrl (targetUrl.AbsoluteUri)) {
+				if (pacUrl == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (pacUrl)}");
```
spouliot: same, can't be `null`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
 
-			if (resultCallback == null)
-				throw new ArgumentNullException ("resultCallback");
+				if (url == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (url)}");
```
spouliot: same, can't be `null`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+
+			if (targetUrl == null)
+				throw new ArgumentNullException (nameof (targetUrl));
 
-			IntPtr source = CFNetworkExecuteProxyAutoConfigurationURL (proxyAutoConfigurationURL.Handle, targetURL.Handle, resultCallback, clientContext);
-			return (source == IntPtr.Zero) ? null : new CFRunLoopSource (source);
+			using (var pacUrl = new NSUrl (proxyAutoConfigurationUrl.AbsoluteUri)) // toll free bridge to CFUrl
+			using (var url = new NSUrl (targetUrl.AbsoluteUri)) {
+				if (pacUrl == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (pacUrl)}");
```
spouliot: same

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+
+				if (url == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (url)}");
```
spouliot: same

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			if (proxyAutoConfigurationUrl == null)
+				throw new ArgumentNullException (nameof (proxyAutoConfigurationUrl));
+
+			if (targetUrl == null)
+				throw new ArgumentNullException (nameof (targetUrl));
+			
+			using (var pacUrl = new NSUrl (proxyAutoConfigurationUrl.AbsoluteUri)) // toll free bridge to CFUrl
+			using (var url = new NSUrl (targetUrl.AbsoluteUri)) {
+				if (pacUrl == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (pacUrl)}");
```
spouliot: same

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,241 @@ public static CFProxySettings GetSystemProxySettings ()
...
+
+				if (url == null)
+					throw new ArgumentException ($"Could not get AbsoluteUri for {nameof (url)}");
```
spouliot: same

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: @spouliot it was ignored in the #if false so it was never exposed.

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: @spouliot perse, they will be set to null. But I understand it could be misleading, I'll rearrange so that it makes more sense.

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: @spouliot I'm using `using` since it is simpler and @rolfbjarne recommended to use it over manually releasing (diff instance, but was in the CFArray for the ParseProxies method).

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: There is no constructor for CFProxy that takes a CFDictionary :/

src/CoreFoundation/CFProxySupport.cs
spouliot: I have not read the associated documentation - so it's not clear if the proxy (if any) would be usable in case of errors... or if errors also include warnings...

src/CoreFoundation/CFProxySupport.cs
spouliot: it's an internal API so it could be changed - but since it's already existing (and not new to this PR) let's not introduce more changes than are required :)

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,217 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					outError = new NSError (cbError);
+				return proxies;
+			}
+		}
+
+		static CFRunLoopSource ExecutePacCFRunLoopSourceBlocking (CreatePACCFRunLoopSource factory, CFProxyAutoConfigurationResultCallback clientCb, CFStreamClientContext clientContext)
+		{
+			CFProxyAutoConfigurationResultCallbackInternal cb = delegate (IntPtr client, IntPtr proxyList, IntPtr error) {
+				var proxies = ParseProxies (proxyList);
+				clientCb ((client != IntPtr.Zero)? null : Runtime.GetNSObject<NSObject> (client), proxies, (error == IntPtr.Zero) ? null : Runtime.GetNSObject<NSError> (error));
```
spouliot: I meant that `Runtime.GetNSObject<NSObject> (client)` would replace the `client != IntPtr.Zero` check too, so it becomes:

```csharp
clientCb (Runtime.GetNSObject<NSObject> (client), proxies, Runtime.GetNSObject<NSError> (error));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,217 @@ public static CFProxySettings GetSystemProxySettings ()
...
 			/* CFStringRef __nonnull */ IntPtr proxyAutoConfigurationScript,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
-			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
-		
-		public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (NSString proxyAutoConfigurationScript, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ CFProxyAutoConfigurationResultCallbackInternal cb,
+			/* CFStreamClientContext * __nonnull */ ref CFStreamClientContext  clientContext);
+
+		public static CFRunLoopSource ExecuteProxyAutoConfigurationScript (string proxyAutoConfigurationScript, Uri targetUrl, CFProxyAutoConfigurationResultCallback clientCb, CFStreamClientContext clientContext)
```
spouliot: minor: no abbreviation in public parameters, i.e. `clientCb` -> `clientCallback`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,217 @@ public static CFProxySettings GetSystemProxySettings ()
...
 		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
 			/* CFURLRef __nonnull */ IntPtr proxyAutoConfigurationURL,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
-			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ CFProxyAutoConfigurationResultCallbackInternal cb,
+			/* CFStreamClientContext * __nonnull */ ref CFStreamClientContext clientContext);
 
-		public static CFRunLoopSource ExecuteProxyAutoConfigurationURL (NSUrl proxyAutoConfigurationURL, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
+		public static CFRunLoopSource ExecuteProxyAutoConfigurationUrl (Uri proxyAutoConfigurationUrl, Uri targetUrl, CFProxyAutoConfigurationResultCallback clientCb, CFStreamClientContext clientContext)
```
spouliot: minor: `clientCb` -> `clientCallback`

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,217 @@ public static CFProxySettings GetSystemProxySettings ()
...
+				}
+			}
+			return proxies;
+		}
+
+		static async Task<(CFProxy[] proxies, NSError error)> ExecutePacCFRunLoopSourceAsync (CreatePACCFRunLoopSource factory, CancellationToken cancellationToken)
+		{
+			CFProxy[] proxies = null;
+			NSError outError = null;
+			await Task.Run (() => {
```
whitneyschmidt: Can you provide more context on why we need to have async calls for this specific case?

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,217 @@ public static CFProxySettings GetSystemProxySettings ()
...
+				};
+				var clientContext = new CFStreamClientContext ();
+				using (var loopSource = new CFRunLoopSource (factory (cb, ref clientContext)))
+				using (var mode = new NSString ("Xamarin.iOS.Proxy")) {
+					runLoop.AddSource (loopSource, mode);
+					// blocks until stop is called, will be done in the cb set previously
+					runLoop.RunInMode (mode, double.MaxValue, false);
+					runLoop.RemoveSource (loopSource, mode);
+				}
+			}, cancellationToken).ConfigureAwait (false);
```
whitneyschmidt: `await` spawns a new thread that asynchronously runs some code...what does `.ConfigureAwait(false)` (or `true`) do?

tests/monotouch-test/CoreFoundation/ProxyTest.cs
```diff
@@ -42,5 +44,182 @@ public void Fields ()
...
+			var targetUri = new Uri ("http://docs.xamarin.com");
+			var runLoop = CFRunLoop.Current;
+
+			// get source, execute run loop in a diff thread
+
+			CFNetwork.CFProxyAutoConfigurationResultCallback cb = delegate (NSObject c, CFProxy [] l, NSError e) {
+				proxies = l;
+				error = e;
+				cbClient = c;
+				runLoop.Stop ();
```
whitneyschmidt: What does `runLoop.Stop()` do in this test code?

tests/monotouch-test/CoreFoundation/ProxyTest.cs
```diff
@@ -42,5 +44,182 @@ public void Fields ()
...
+			bool done = false;
+			string pacPath = Path.Combine (NSBundle.MainBundle.BundlePath, "example.pac");
+
+			var script = File.ReadAllText (pacPath);
+			var targetUri = new Uri ("http://docs.xamarin.com");
+			
+			Exception ex;
+			bool foundProxies;
+			// similar to the other tests, but we want to ensure that the async/await API works
+			TestRuntime.RunAsync (DateTime.Now.AddSeconds (30), async () => {
```
whitneyschmidt: How does this asynchronous (?) test code interact with the asynchronous source code/is there a pattern for writing async tests?

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: This implementation is similar to the sync once except that the execution of the **CFRunLoop** run method is executed in  **different thread**. You can think of the CFRunLoop  run method as a while loop that processes messages that are pumped to it via CFRunLoopSource objects. The *run* method will have a while loop that will process the events and will block until stopped.

Moving the run method to a **diff thread** ensures that we do not block the main thread (UI thread most probably) until the PAC script file is executed.

We are providing 3 APIs for this.

1. Sync one that blocks the thread until the PAC is parsed.
2. Sync one that returns the CFRunLoopSource for the user to add to a run loop to me processed.
3. Async one that does all the work for the user so that he does not have to look into the details of getting the correct CFRunLoop and call the run method.

This method is number 3. We ensure that we do not block the current thread and return the result via a task that the user can await.

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: Await spawns a thread yes does it always not block the thread, false.. You can imaging the await keyboard as to say 'execute this piece of code **somewhere** (we will get to the somewhere later) and get me back here to the same thread. How does .Net bring you back to the same thread? To do that, there is something called the SynchronizationContext. You can imaging everything as follows:

The following is the code you write:
```csharp
await GetMyDataAsync();
RestOfMethod();
```
This is the code *'generated'* by the runtime (not exactly, but more or less)
```csharp
var t = GetMyDataAsync();
var currentContext = SynchronizationContext.Current;
t.ContinueWith(delegate
{
    if (currentContext == null)
        RestOfMethod();
    else
        currentContext.Post(delegate { RestOfMethod(); }, null);
}, TaskScheduler.Current);
```
The above means, that when you are using the current synchronisation context, the task will be ran in it, which means that, if you are in the main thread (UI thread) the call **WILL BLOCK** the thread.

In our case, we do not want to be executed in the same sync context, we want it to be ran 'anywhere else' so we tell the system that we do not want to use the same sync context as the current one via **ConfigureAwait (false)**.

This is the most common use if you are writing libs or you do not want to block the current thread by using the same sync context (sometimes you do, since there is certain information we need in the current context etc.. (

tests/monotouch-test/CoreFoundation/ProxyTest.cs
mandel-macaque: The CFRunLoop is like a never ending while loop that processes all those events that have been added in it until it is stopped. In our case, we want to get the current loop, execute it until our callback is executed. At that point we have all the data but the loop is sitting for more events to be pumped it to be processed (via the add CFRunLoopSource method). Of course, we are done with the events, we know that as a fact, but the runloop does not know, so within the callback, once we have all what we needed, we tell the CFRunLoop to exit the *'while loop'*.

tests/monotouch-test/CoreFoundation/ProxyTest.cs
mandel-macaque: This is the pattern you can use to test async code within the MonoTouch tests without blocking the UI thread and getting into a deadlock. (There are not many tests like this in the tests).

mandel-macaque: To be more precise, the method will ensure that the cb is executed async without blocking the main thread.

src/CoreFoundation/CFProxySupport.cs
whitneyschmidt: So the pattern is: send off new thread to do something, return a `Task` object that will contain the newly spawned thread's results when it returns?

But if the current thread is awaiting the return from the `Task`, isn't it blocked? Or is there an event that notifies the main thread that the task is complete so that it can do something with the result?

src/CoreFoundation/CFProxySupport.cs
whitneyschmidt: Thank you for the detailed explanation - that makes a lot of sense! :)

tests/monotouch-test/CoreFoundation/ProxyTest.cs
whitneyschmidt: Ok. So the CFRunLoop sits there we do stuff in it until the callback calls stop on it - it is kind of a resource that the dev must manually manage since it is not smart enough to stop itself :)

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: So in this case. We crate a new thread that is squealed in the ThreadPool, we set the Task to ConfigureAwait (false) stating that we do not need the sync context for the calling thread and return it.

Because we are not requiring the sync context, the calling thread is not blocked, which is very important in threads that are dealing with events such as the UI thread, and whenever the thread completes, the code of the user will resume executing as if everything was sequential. 

For the point of view of our users, the should write the code as if it was sequential when in reality we are not blocking the UI thread with the CFRunLoop but a diff one.

tests/monotouch-test/CoreFoundation/ProxyTest.cs
mandel-macaque: Correct.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		public delegate void CFProxyAutoConfigurationResultCallback (NSObject client, CFProxy[] proxyList, NSError error);
+		// helper delegate to reuse code
+		delegate IntPtr CreatePACCFRunLoopSource (CFProxyAutoConfigurationResultCallbackInternal cb, ref CFStreamClientContext context);
+
+		static CFProxy[] ParseProxies (IntPtr proxyList)
+		{
+			CFProxy[] proxies = null;
+			if (proxyList != IntPtr.Zero) {
+				// it was retained in the cbs.
+				using (var array = new CFArray (proxyList, true)) {
```
rolfbjarne: You call `ParseProxies` from two places: one is a property that can call `ParseProxies` multiple times for the same pointer (without retaining it first). This will end up crashing.

Seems like it would be simpler to pass `owns: false`, and then in the other place that calls `ParseProxies`, stop retaining the pointer first.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					proxies = new CFProxy [array.Count];
+					for (int i = 0; i < proxies.Length; i++) {
+						var dict = new NSDictionary (array.GetValue (i));
```
rolfbjarne: I think you should use `Runtime.GetNSObject (array.GetValue (i))` instead of manually creating the `NSDictionary` using the pointer. That works better if the same native pointer shows up multiple times.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			}
+			return proxies;
+		}
+
+		// helper struct to contain all the data that was used in the callback
+		struct PACProxyCallbackData
+		{
+			public IntPtr ProxyListPtr; // Pointer to a CFArray to later be parsed
+			public IntPtr ErrorPtr; // Pointer to the Error
+			public IntPtr CFRunLoopPtr; // Pointer to the runloop, needed to be stoped
```
rolfbjarne:
```suggestion
			public IntPtr CFRunLoopPtr; // Pointer to the runloop, needed to be stopped
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			CFProxy[] proxies = null;
+			NSError outError = null;
+			await Task.Run (() => {
+				// we need the runloop of THIS thread, so it is important to get it in the correct context
+				var runLoop = CFRunLoop.Current;
+
+				// build a struct that will have all the needed info for the callback
+				var pacCbData = new PACProxyCallbackData ();
+				pacCbData.CFRunLoopPtr = runLoop.Handle;
+				var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf(pacCbData));
```
rolfbjarne:
```suggestion
				var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf (pacCbData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+					clientContext.Info = pacDataPtr;
+
+					using (var loopSource = new CFRunLoopSource (factory (ExecutePacCallback, ref clientContext)))
+					using (var mode = new NSString ("Xamarin.iOS.Proxy")) {
+						runLoop.AddSource (loopSource, mode);
+						// blocks until stop is called, will be done in the cb set previously
+						runLoop.RunInMode (mode, double.MaxValue, false);
+						runLoop.RemoveSource (loopSource, mode);
+					}
+					pacCbData = (PACProxyCallbackData) Marshal.PtrToStructure(pacDataPtr, typeof (PACProxyCallbackData));
```
rolfbjarne:
```suggestion
					pacCbData = (PACProxyCallbackData) Marshal.PtrToStructure (pacDataPtr, typeof (PACProxyCallbackData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		static CFProxy[] ExecutePacCFRunLoopSourceBlocking (CreatePACCFRunLoopSource factory, out NSError outError)
+		{
+			var runLoop = CFRunLoop.Current;
+			CFProxy [] proxies = null;
+			outError = null;
+
+			// build a struct that will have all the needed info for the callback
+			var pacCbData = new PACProxyCallbackData ();
+			pacCbData.CFRunLoopPtr = runLoop.Handle;
+			var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf(pacCbData));
```
rolfbjarne:
```suggestion
			var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf (pacCbData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+				var clientContext = new CFStreamClientContext ();
+				clientContext.Info = pacDataPtr;
+
+				using (var loopSource = new CFRunLoopSource (factory (ExecutePacCallback, ref clientContext)))
+				using (var mode = new NSString ("Xamarin.iOS.Proxy")) {
+					runLoop.AddSource (loopSource, mode);
+					runLoop.RunInMode (mode, double.MaxValue, false);
+					runLoop.RemoveSource (loopSource, mode);
+				}
+				pacCbData = (PACProxyCallbackData) Marshal.PtrToStructure(pacDataPtr, typeof (PACProxyCallbackData));
```
rolfbjarne:
```suggestion
				pacCbData = (PACProxyCallbackData) Marshal.PtrToStructure (pacDataPtr, typeof (PACProxyCallbackData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+				// get data from the struct
+				if (pacCbData.Error != null)
```
rolfbjarne: Can't you just get both values in both cases?

```csharp
outError = pacCbData.Error;
return pacCbData.ProxyList;
```

it's a few lines shorter :smile:

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		[MonoPInvokeCallback (typeof (CFProxyAutoConfigurationResultCallbackInternal))]
+		static void ExecutePacUserCallback (IntPtr client, IntPtr proxyListPtr, IntPtr errorPtr)
+		{
+			// grab the required structure and set the data, according apple docs:
+			// client
+			// The client reference originally passed in the clientContext parameter of the 
+			// CFNetworkExecuteProxyAutoConfigurationScript or CFNetworkExecuteProxyAutoConfigurationURL call
+			// that triggered this callback.
+			// Well, that is NOT TRUE, the client passed is the client.Info pointer not the client.
+			var pacCbData =  (PACProxyUserCallbackData) Marshal.PtrToStructure (client, typeof (PACProxyUserCallbackData));
```
rolfbjarne:
```suggestion
			var pacCbData = (PACProxyUserCallbackData) Marshal.PtrToStructure (client, typeof (PACProxyUserCallbackData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			NSError error = null;
+			if (errorPtr != IntPtr.Zero) {
+				error = Runtime.GetNSObject<NSError> (errorPtr);
+			} else {
+				// retain so that is nore released
+				CFObject.CFRetain (proxyListPtr);
+				proxies = ParseProxies (proxyListPtr);
+			}
+			// get the callback to be executed
+			pacCbData.UserCallback (Runtime.GetNSObject<NSObject> (pacCbData.UserClientDataPtr), proxies, error);
```
rolfbjarne: I think the above can be simplified to:

```csharp
var proxies = ParseProxies (proxyListPtr);
var error = (NSError) Runtime.GetNSObject (errorPtr);
pacDbData.UserCallback (Runtime.GetNSObject (pacCbData.UserClientDataPtr), proxies, error);
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			// we have to free the pointer here, else, if done in the caller, we will get garbage in this cb.
+			Marshal.FreeHGlobal(client);
```
rolfbjarne:
```suggestion
			Marshal.FreeHGlobal (client);
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+		{
+			// build a struct that will have all the needed info for the callback
+			var pacCbData = new PACProxyUserCallbackData ();
+
+			if (userContext != null)
+				pacCbData.UserClientDataPtr = userContext.Handle;
+			pacCbData.UserCallback = clientCb;
+
+			// this pointer will be cleaned in the wrapper cb
+			var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf(pacCbData));
```
rolfbjarne:
```suggestion
			var pacDataPtr = Marshal.AllocHGlobal (Marshal.SizeOf (pacCbData));
```

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			using (var pacScript = new NSString (proxyAutoConfigurationScript)) 
+			using (var url = new NSUrl (targetUrl.AbsoluteUri)) {
+				CreatePACCFRunLoopSource factory = delegate (CFProxyAutoConfigurationResultCallbackInternal cb, ref CFStreamClientContext context) {
+					return CFNetworkExecuteProxyAutoConfigurationScript (pacScript.Handle, url.Handle, cb, ref context);
+				};
+				return ExecutePacCFRunLoopSourceBlocking (factory, out outError);
+			}
 		}
 		
+		public static async Task<(CFProxy[] proxies, NSError error)> ExecuteProxyAutoConfigurationScriptAsync (string proxyAutoConfigurationScript, Uri targetUrl, CancellationToken cancellationToken)
```
rolfbjarne: The API takes an `CancellationToken`, but as far as I can see the code doesn't do anything in case of a cancellation request (and the native API isn't cancellable either).

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,328 @@ public static CFProxySettings GetSystemProxySettings ()
...
 		extern static /* CFRunLoopSourceRef __nonnull */ IntPtr CFNetworkExecuteProxyAutoConfigurationURL (
 			/* CFURLRef __nonnull */ IntPtr proxyAutoConfigurationURL,
 			/* CFURLRef __nonnull */ IntPtr targetURL,
-			/* CFProxyAutoConfigurationResultCallback __nonnull */ IntPtr cb,
-			/* CFStreamClientContext * __nonnull */ IntPtr clientContext);
+			/* CFProxyAutoConfigurationResultCallback __nonnull */ CFProxyAutoConfigurationResultCallbackInternal cb,
+			/* CFStreamClientContext * __nonnull */ ref CFStreamClientContext clientContext);
 
-		public static CFRunLoopSource ExecuteProxyAutoConfigurationURL (NSUrl proxyAutoConfigurationURL, NSUrl targetURL, CFProxyAutoConfigurationResultCallback resultCallback, CFStreamClientContext clientContext)
+		public static CFRunLoopSource ExecuteProxyAutoConfigurationUrl (Uri proxyAutoConfigurationUrl, Uri targetUrl, CFProxyAutoConfigurationResultCallback clientCallback, NSObject clientContext)
```
rolfbjarne: This API takes a `clientContext`, I suppose that's just anything the caller wants it to be (similar to the native API)? If so, it's redundant, since managed code can use anonymous delegates. It looks like just removing the `clientContext` parameter will make things somewhat simpler too.

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: I did it this way following @spouliot comment here: https://github.com/xamarin/xamarin-macios/pull/6201#discussion_r303969327

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: The cancellation token is passed to Task.Run, which should cancel the thread unless I am mistaken. It is passed to ExecutePacCFRunLoopSourceAsync and then passed to Task.Run here: https://github.com/xamarin/xamarin-macios/pull/6201/files#diff-edb0955a9801d2d134609eb5bd850840R686

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: Removing it will certainly make my life easier, but I'm not sure if clients will want it. In a sense, less Marshalling etc.. so better for us making the code simpler. But this is the 'advance' API (The async/await one is much nicer). I have no preference to be hones (well, yes, I prefer to have less code).

src/CoreFoundation/CFProxySupport.cs
rolfbjarne: No, cancellation is cooperative, nothing will cancel an executing thread (unless you do it yourself). It will however cancel the task if it hasn't been scheduled yet, but that's probably not what users want most of the time.

See also: https://stackoverflow.com/a/22637882/183422

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: I tested, if the user use an anonymous delegate, we get an exception when marshalling :/

src/CoreFoundation/CFProxySupport.cs
rolfbjarne: According to the documentation if one is not null the other will be (and vice versa).

My vote goes for not adding the IntPtr.Zero checks in this code, since that makes the code smaller. If there's confusion, it can be explained with comments.

src/CoreFoundation/CFProxySupport.cs
rolfbjarne: Then you have a bug somewhere :) There's no reason we can't make anonymous delegates work as long as the native API provides a context parameter of some sort.

rolfbjarne: Also it's not common in managed code to have a context parameter like this (people will wonder what the parameter is for), and we hide it everywhere else in our bindings too.

src/CoreFoundation/CFProxySupport.cs
mandel-macaque: I did the change and with just the Url method, the Retain deadlocks. The code paths are the same and jus the native API is changed.. I wonder what could happen. With the current implementation everything works.

src/CoreFoundation/CFProxySupport.cs
```diff
@@ -579,62 +581,252 @@ public static CFProxySettings GetSystemProxySettings ()
...
+			// CFNetworkExecuteProxyAutoConfigurationScript or CFNetworkExecuteProxyAutoConfigurationURL call
+			// that triggered this callback.
+			// Well, that is NOT TRUE, the client passed is the client.Info pointer not the client.
+			var pacCbData =  (PACProxyCallbackData) Marshal.PtrToStructure (client, typeof (PACProxyCallbackData));
+			// make sure is not released, will be released by the parsing method.
+			if (proxyList != IntPtr.Zero) {
+				CFObject.CFRetain (proxyList);
+				pacCbData.ProxyListPtr = proxyList;
+			}
+			pacCbData.ErrorPtr = error;
```
rolfbjarne: You must retain/release the error ptr as well.

mandel-macaque:

PR #6208, opened 2019-06-05
[XHarness] Point to the correct directory that has the linker config.

mandel-macaque:
The path is poiting to the old dir.

Fixes: https://github.com/xamarin/maccore/issues/1690


mandel-macaque:
All crashes are HE0038 related and the mac failure seems to be new but probably network related, created an issue here: https://github.com/xamarin/maccore/issues/1705

PR #6211, opened 2019-06-05
Fix wrong cast to IList

BrzVlad:
Fixes https://github.com/mono/mono/issues/14729

### DIFF
--- a/tests/bcl-test/templates/common/TestRunner.Core/TestRunner.cs
+++ b/tests/bcl-test/templates/common/TestRunner.Core/TestRunner.cs
@@ -30,7 +30,7 @@ namespace Xamarin.iOS.UnitTests
 			Logger = logger ?? throw new ArgumentNullException (nameof (logger));
 		}
 
-		public abstract void Run (IList <TestAssemblyInfo> testAssemblies);
+		public abstract void Run (IEnumerable<TestAssemblyInfo> testAssemblies);
 		public abstract string WriteResultsToFile ();
 		public abstract void WriteResultsToFile (TextWriter writer);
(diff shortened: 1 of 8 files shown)


rolfbjarne:
Why is that an invalid cast if the C# compiler accepts it?

BrzVlad:
I meant incorrect. An IEnumerable is not necessarily an IList so an exception can be thrown here

Therzok:
@BrzVlad Why not modify the signature of [TestRunner.Run](https://github.com/xamarin/xamarin-macios/blob/26a33946237ea0d9434687a03fb637ab90682445/tests/bcl-test/templates/common/TestRunner.Core/TestRunner.cs#L9) to take an IEnumerable? Don't see IList members being used.

rolfbjarne:
@BrzVlad 

> I meant incorrect. An IEnumerable is not necessarily an IList so an exception can be thrown here

That still doesn't explain why it only fails with the interpreter.

Therzok:
Can't repro the problem with Mac and Console apps, but can repro it not throwing on iOS debug builds.

Therzok:
I think [this](https://github.com/mono/mono/issues/14729#issuecomment-499306330) is why.

mandel-macaque:
Failures are due to issues with the simulators in the bots after the upgrade to Xcode 11. I have fixed the bots, will retrigger a build and everything should be back to normal.

mandel-macaque:
Mac failure is already known: https://github.com/xamarin/maccore/issues/1685 and the watchOS failures are due to https://github.com/xamarin/maccore/issues/581

Looks good to merge.

mandel-macaque:
API diff issue can also be ignored.

PR #6220, opened 2019-06-07
[XHanress] Move ignore file to the correct dir.

mandel-macaque:
There was a merge issue (two diff PRs landing at diff times) which made
the file to be added to the wrong dir.

Fixes: https://github.com/xamarin/maccore/issues/1689


mandel-macaque:
WatchOS error is https://github.com/xamarin/maccore/issues/581

PR #6228, opened 2019-06-10
[Bcl][Tests] Ignore the ssl related bcl tests that block.

mandel-macaque:
There are a number of tests that block in the BCL relatd to ssl, mono
issue https://github.com/mono/mono/issues/14938 was added to track them.


mandel-macaque:
Closing since we can push to the monthly branch directly.

PR #6237, opened 2019-06-10
[Tests] Try different urls before failing the test.

mandel-macaque:
We do have a test that fails when there are issues accessing the remote
resource. If an error is raised, rather than setting the tests as an
error, set them to inconclusive with the message of the error.

Related issue: https://github.com/xamarin/maccore/issues/1725

### DIFF
--- a/tests/monotouch-test/Foundation/NSDataTest.cs
+++ b/tests/monotouch-test/Foundation/NSDataTest.cs
@@ -155,15 +155,27 @@ namespace MonoTouchFixtures.Foundation {
 				Assert.Ignore ("NSData.FromUrl doesn't seem to work in watchOS");
 			}
 #endif
-			NSError error;
-			using (var url = new NSUrl ("https://www.microsoft.com/robots.txt"))
-			using (var x = NSData.FromUrl (url, NSDataReadingOptions.Uncached, out error)) {
-				if (error != null) {
-					Assert.Inconclusive ($"Could not access url error is '{error.Description}'");
-				} else {
-					Assert.That ((x != null) && (x.Length > 0));
+			// we have network issues, try several urls, if one works, be happy, else fail
+			var urls = new string [] {
+				"https://www.microsoft.com/robots.txt",
+				"https://www.xamarin.com/robots.txt",
+				"http://www.bing.com/robots.txt",
+				"https://www.xbox.com/robots.txt",
+				"https://www.msn.com/robots.txt",
+				"https://visualstudio.microsoft.com/robots.txt",
+			};
+			for (var i = 0; i < urls.Length; i++) {
+				NSError error;
+				using (var nsUrl = new NSUrl (urls [i]))
+				using (var x = NSData.FromUrl (nsUrl, NSDataReadingOptions.Uncached, out error)) {
+					if (error != null)
+						continue;
+					Assert.That (x != null);
+					Assert.That (x.Length > 0);
+					return;
 				}
 			}
+			Assert.Fail ("None of the urls could be fetch.");
 		}
 
 		[Test]
--- a/tests/monotouch-test/Foundation/NSDataTest.cs
+++ b/tests/monotouch-test/Foundation/NSDataTest.cs
@@ -155,9 +155,14 @@ namespace MonoTouchFixtures.Foundation {
 				Assert.Ignore ("NSData.FromUrl doesn't seem to work in watchOS");
 			}
 #endif
+			NSError error;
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/Foundation/NSDataTest.cs
```diff
@@ -155,9 +155,14 @@ public void Https ()
...
+			NSError error;
 			using (var url = new NSUrl ("https://www.microsoft.com/robots.txt"))
-			using (var x = NSData.FromUrl (url)) {
-				Assert.That ((x != null) && (x.Length > 0));
+			using (var x = NSData.FromUrl (url, NSDataReadingOptions.Uncached, out error)) {
+				if (error != null) {
+					Assert.Inconclusive ($"Could not access url error is '{error.Description}'");
+				} else {
+					Assert.That ((x != null) && (x.Length > 0));
+				}
```
spouliot: the test was to detect an error...
with that change I don't think you can trigger the 2nd assert (unless an empty file was present)
so basically the test can't fail (so it's not worth much)

mandel-macaque: @spouliot we can check the domain of the error and decide what to do?

mandel-macaque: Well.. thx Cocoa, we would get the not so useful "NSCocoaErrorDomain Code=256", which means an unknown error happened.

spouliot: yeah, a reachability (to host) error would be fine to ignore, but other errors would not be

spouliot:
for reference
```
	[INCONCLUSIVE] NSDataTest.Https : Could not access url error is 'Error Domain=NSCocoaErrorDomain Code=256 "The file “robots.txt” couldn’t be opened." UserInfo={NSURL=https://www.microsoft.com/robots.txt}'
		  at MonoTouchFixtures.Foundation.NSDataTest.Https () [0x00021] in /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/monotouch-test/Foundation/NSDataTest.cs:162 
		  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:391 
```
https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10214/Test_20Report/

tests/monotouch-test/Foundation/NSDataTest.cs
```diff
@@ -155,9 +155,14 @@ public void Https ()
...
 #endif
+			NSError error;
 			using (var url = new NSUrl ("https://www.microsoft.com/robots.txt"))
-			using (var x = NSData.FromUrl (url)) {
-				Assert.That ((x != null) && (x.Length > 0));
+			using (var x = NSData.FromUrl (url, NSDataReadingOptions.Uncached, out error)) {
+				if (error != null) {
+					Assert.Inconclusive ($"Could not access url error is '{error.Description}'");
+				} else {
+					Assert.That ((x != null) && (x.Length > 0));
```
rolfbjarne: Splitting the complex expression into multiple asserts would make failures easier to diagnose, since NUnit would provide better assertion messages:

```csharp
Assert.IsNotNull (x, "NotNull");
Assert.That (x, Is.Not.Empty, "NotEmpty");
```

mandel-macaque: I like the idea of the diff urls, I'll try that, make sense, this is a network failure due to the setup of the labs, which is very annoying :/

tests/monotouch-test/Foundation/NSDataTest.cs
mandel-macaque: I have used several company urls, lets hope trying several will work, else we can do an increasing timeout retry.

spouliot:
it's unclear why it fails so it's possible the complete list will fail (even more since it's all MS hosts) but if it works then it's much better than the current test :)

mandel-macaque:
Issues are related to the msbuild problems we have, all other tests, specially the NSDataTests passed.

mandel-macaque:

mandel-macaque:

VincentDondain:

PR #6277, opened 2019-06-12
Make NSUrlSessionHandler work with both HttpClient implementations.

baulig:
This is not a perfect solution, but the most robust and risk-free approach to fix https://github.com/mono/mono/issues/12996.

The `NSUrlSessionHandler` implementation depends on Mono-specific behavior, which makes
`SerializeToStreamAsync()` cancellable.  Unfortunately, the CoreFX implementation of
HttpClient does not support this.

By copying Mono's old implementation here, we ensure that we're compatible with both
HttpClient implementations, so when we eventually adopt the CoreFX version in all of
Mono's profiles, we don't regress here.

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -841,7 +841,7 @@ namespace Foundation {
 
 		// Needed since we strip during linking since we're inside a product assembly.
 		[Preserve (AllMembers = true)]
-		class NSUrlSessionDataTaskStreamContent : StreamContent
+		class NSUrlSessionDataTaskStreamContent : MonoStreamContent
 		{
 			Action disposed;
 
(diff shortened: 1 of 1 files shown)


baulig:
Another approach would be to make `NSUrlSessionHandler` not depend on `SerializeToStreamAsync()` being cancellable.  We can still do that at a later point.

At the moment, I would like to have something that will put us on the safe side and not regress once we port over the other Mono profiles to the CoreFX HttpClient.

spouliot:
The failure are related to known msbuild path issue mono/mono#14875 (and not the PR).
@baulig if complete then you can merge it (now is a good time before we branch for d16-3)
thanks!

directhex:
@spouliot I think @baulig is happy with it. Let's merge ASAP, as it's blocking VS 2019-06 integration

mandel-macaque:
@monojenkins backport d16-2

PR #6281, opened 2019-06-13
[tests] Fix monotouch-test for "iOS Unified 32-bits - simulator" | "Release (all optimizations)"

spouliot:
Photos.framework headers are broken in Xcode 11 beta 1 [1]. Optimizations
fails since the static registrar skip over the type [2]

```
	[FAIL] PHLivePhotoEditingContextTest.Linker : ObjCRuntime.RuntimeException : Can't register the class Photos.PHContentEditingInput when the dynamic registrar has been linked away.
		  at ObjCRuntime.Class.GetClassHandle (System.Type , System.Boolean , System.Boolean& ) [0x000de] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at ObjCRuntime.Class.GetClassHandle (System.Type ) [0x00000] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at ObjCRuntime.Class.GetHandle (System.Type ) [0x00000] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at Foundation.NSObject.AllocIfNeeded () [0x00019] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at Foundation.NSObject..ctor (Foundation.NSObjectFlag ) [0x00006] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at Photos.PHContentEditingInput..ctor () [0x00000] in <d93fa0efd60442128943a9eb0e82eb8d>:0
		  at MonoTouchFixtures.Photos.PHLivePhotoEditingContextTest.Linker () [0x00001] in <6e792af97a28410a80fa790ecfb26b9c>:0
		  at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
		  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object , System.Reflection.BindingFlags , System.Reflection.Binder , System.Object[] , System.Globalization.CultureInfo ) [0x0006a] in <3440dfbcd957471aabeed02fbc939d2a>:0
PHLivePhotoEditingContextTest : 2 ms
```

references
[1] https://github.com/xamarin/xamarin-macios/issues/6212
[2] https://github.com/xamarin/xamarin-macios/commit/60a5392d359ad76db334a3d9f764997f9b2fde02

Issue #6258, opened 2019-06-11
[xcode11] monotouch-test crashes in NSDecimalNumber.One


rolfbjarne:
https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10186/Test_20Report/

<details>
<summary>Crash details</summary>

```

=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Basic Fault Adddress Reporting
=================================================================
Memory around native instruction pointer (0x1166f9986):0x1166f9976  c0 48 39 d6 0f 94 c0 c3 c3 31 c9 48 85 f6 74 46  .H9......1.H..tF
0x1166f9986  8a 06 84 c0 74 40 48 83 c6 04 31 c9 0f b6 c0 31  ....t@H...1....1
0x1166f9996  c1 0f b6 46 fd 85 c0 74 2d c1 e0 08 31 c1 0f b6  ...F...t-...1...
0x1166f99a6  46 fe 85 c0 74 20 c1 e0 10 31 c8 0f b6 4e ff 85  F...t ...1...N..

=================================================================
	Native stacktrace:
=================================================================
	0x107084b1d - /Users/builder/Library/Developer/CoreSimulator/Devices/E10D999A-AA61-4129-838E-3F14B5627E4D/data/Containers/Bundle/Application/964376E2-1913-4F91-8501-AD4A3C7A537D/monotouchtest.app/monotouchtest : mono_dump_native_crash_info
	0x107078ac5 - /Users/builder/Library/Developer/CoreSimulator/Devices/E10D999A-AA61-4129-838E-3F14B5627E4D/data/Containers/Bundle/Application/964376E2-1913-4F91-8501-AD4A3C7A537D/monotouchtest.app/monotouchtest : mono_handle_native_crash
	0x10708bd98 - /Users/builder/Library/Developer/CoreSimulator/Devices/E10D999A-AA61-4129-838E-3F14B5627E4D/data/Containers/Bundle/Application/964376E2-1913-4F91-8501-AD4A3C7A537D/monotouchtest.app/monotouchtest : mono_sigsegv_signal_handler_debug
	0x11715fb5d - /usr/lib/system/libsystem_platform.dylib : _sigtramp
	0xf690400070001002 - Unknown
[...]

Issue #6255, opened 2019-06-11
[xcode11] link all/tvOS: crash


rolfbjarne:
https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10186/Test_20Report/

<details>
<summary>Crash details</summary>

```

=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Basic Fault Adddress Reporting
=================================================================
Memory around native instruction pointer (0x10ba6d986):0x10ba6d976  c0 48 39 d6 0f 94 c0 c3 c3 31 c9 48 85 f6 74 46  .H9......1.H..tF
0x10ba6d986  8a 06 84 c0 74 40 48 83 c6 04 31 c9 0f b6 c0 31  ....t@H...1....1
0x10ba6d996  c1 0f b6 46 fd 85 c0 74 2d c1 e0 08 31 c1 0f b6  ...F...t-...1...
0x10ba6d9a6  46 fe 85 c0 74 20 c1 e0 10 31 c8 0f b6 4e ff 85  F...t ...1...N..

=================================================================
	Native stacktrace:
=================================================================
	0x10782c615 - /Users/builder/Library/Developer/CoreSimulator/Devices/F5307DF2-F41A-499F-8C7E-B9B7CF85059C/data/Containers/Bundle/Application/A40F40C9-0BE3-4392-915C-FA72F6029EE6/link all.app/link all : mono_dump_native_crash_info
	0x107820df5 - /Users/builder/Library/Developer/CoreSimulator/Devices/F5307DF2-F41A-499F-8C7E-B9B7CF85059C/data/Containers/Bundle/Application/A40F40C9-0BE3-4392-915C-FA72F6029EE6/link all.app/link all : mono_handle_native_crash
	0x107832795 - /Users/builder/Library/Developer/CoreSimulator/Devices/F5307DF2-F41A-499F-8C7E-B9B7CF85059C/data/Containers/Bundle/Application/A40F40C9-0BE3-4392-915C-FA72F6029EE6/link all.app/link all : mono_sigsegv_signal_handler
	0x10c4cfb5d - /usr/lib/system/libsystem_platform.dylib : _sigtramp
	0x120f51058 - Unknown
[...]

Issue #6252, opened 2019-06-11
[xcode11] watchOS simulator projects fail to build


rolfbjarne:
Building for watchOS simulator (and probably device as well) fails with:

```
clang : error : linker command failed with exit code 1 (use -v to see invocation) [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/tmp-test-dir/dont link-watchos-extension/dont link-watchos-extension.csproj] 
MTOUCH : error MT5210: Native linking failed, undefined symbol: _main. Please verify that all the necessary frameworks have been referenced and native libraries are properly linked in. [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/tmp-test-dir/dont link-watchos-extension/dont link-watchos-extension.csproj] 
MTOUCH : error MT5202: Native linking failed. Please review the build log. [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/tmp-test-dir/dont link-watchos-extension/dont link-watchos-extension.csproj] 
```

https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10186/Test_20Report/.

spouliot:
Most failures seems fixed but I expected this one to be too
```
Xamarin.iOS.Tasks.WatchKit2("iPhoneSimulator").BasicTest: #RunTarget-ErrorCount
    redefinition of 'main'
    Failed to compile the file(s) '/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/msbuild/tests/MyWatchKit2IntentsExtension/obj/iPhoneSimulator/Debug/mtouch-cache/i386/main.m'. Please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new
Expected: 0
But was: 2
```
https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10296//Test_20Report

rolfbjarne:
@spouliot 

> Most failures seems fixed but I expected this one to be too
> 
> ```
> Xamarin.iOS.Tasks.WatchKit2("iPhoneSimulator").BasicTest: #RunTarget-ErrorCount
>     redefinition of 'main'
[...]

PR #6282, opened 2019-06-13
[replaykit] Update for xcode 11 beta 1

spouliot:
Added new enum value

Also removed availability attributes from other values - since they are
not helpful for error codes (e.g. you don't want warnings for them, nor
do you want to add version checks around them)


spouliot:
known failures due to

* msbuild path issue mono/mono#14875
* monotouch sim crash #6258
* and the test fixed by (just merged) https://github.com/xamarin/xamarin-macios/pull/6281

PR #6283, opened 2019-06-13
[watchconnectivity] Update for xcode 11 beta 1

spouliot:
Add new API for CompanionApp (watchOS only)


spouliot:
known failures due to

* msbuild path issue mono/mono#14875
* monotouch sim crash #6258

PR #6287, opened 2019-06-13
[XHarness] Allow parallel tests by default on the test runners.

mandel-macaque:
The runners are not running tests in parallel. This is not a problem in
the NUnit runner but results in tests blocking on the xUnit when those
tests do use async/await. The xUnit runner implementation used allows to
run tests in parallel this has two positive things:

1. Tests will run faster.
2. The tests that are async, will not block.

Fixes: https://github.com/mono/mono/issues/15058


mandel-macaque:
Closing in favour of https://github.com/xamarin/xamarin-macios/pull/6291 which has a nicer implementation.

PR #6291, opened 2019-06-14
[XHarness]: Convert xUnit / nUnit tests to run async.

baulig:
Make `TestRunner.Run()` async as we cannot sync-block on the UI thread.

### DIFF
--- a/tests/bcl-test/templates/common/TestRunner.NUnit/NUnitTestRunner.cs
+++ b/tests/bcl-test/templates/common/TestRunner.NUnit/NUnitTestRunner.cs
@@ -5,6 +5,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Threading.Tasks;
 
 using Foundation;
 
@@ -35,7 +36,9 @@ namespace Xamarin.iOS.UnitTests.NUnit
 			builderSettings = new Dictionary<string, object> (StringComparer.OrdinalIgnoreCase);
 		}
 
-		public override void Run (IEnumerable<TestAssemblyInfo> testAssemblies)
+#pragma warning disable 1998
+		public override async Task Run (IEnumerable<TestAssemblyInfo> testAssemblies)
+#pragma warning restore 1998
 		{
 			if (testAssemblies == null)
 				throw new ArgumentNullException (nameof (testAssemblies));
--- a/tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
+++ b/tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
@@ -692,7 +692,8 @@ namespace Xamarin.iOS.UnitTests.XUnit
 			}, tcs, timeout, true);
 
 			try {
-				await tcs.Task.ConfigureAwait (false);
+				if (!tcs.Task.IsCompleted)
+					await tcs.Task.ConfigureAwait (false);
 			} finally {
 				registration.Unregister (handle);
 			}
@@ -937,8 +938,10 @@ namespace Xamarin.iOS.UnitTests.XUnit
 					var tcs = new TaskCompletionSource<object> ();
 
 					try {
+						// set the wait for event cb first, then execute the tests
+						var resultTask = WaitForEvent (resultsSink.Finished, TimeSpan.FromDays (10)).ConfigureAwait (false);
 						frontController.RunTests (testCases, resultsSink, executionOptions);
-						await WaitForEvent (resultsSink.Finished, TimeSpan.FromDays (10)).ConfigureAwait (false);
+						await resultTask;
 					} finally {
 						resultsSink.Dispose ();
 					}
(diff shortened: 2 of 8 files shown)


mandel-macaque:
We can ignore the msbuild related failures and the API diff. But the fact that the xunit test apps block is worrying.

baulig:
It seems to block _after_ running the tests.  That's very weird.

mandel-macaque:
@baulig I'm going to take a look tonight to see what is going on

tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
```diff
@@ -916,8 +936,10 @@ XElement Run (Assembly assembly, string assemblyPath)
 					executionOptions.SetSynchronousMessageReporting (true);
 
 					try {
+						// set the wait for event cb first, then execute the tests
+						var resultTask = WaitForEvent (resultsSink.Finished, TimeSpan.FromDays (10)).ConfigureAwait (false);
 						frontController.RunTests (testCases, resultsSink, executionOptions);
-						resultsSink.Finished.WaitOne ();
+						await resultTask;
```
baulig: Do we need to continue on the main thread here?

```suggestion
						await resultTask.ConfigureAwait (false);
```

tests/bcl-test/templates/tvOSApp/ViewController.cs
```diff
@@ -85,7 +85,7 @@ public async override void ViewDidLoad ()
 				runner.SkipTests (skippedTests);
 			}
 
-			runner.Run (testAssemblies);
+			await runner.Run (testAssemblies);
```
baulig: You moved the `TerminateWithSuccess ()` into `BeginInvokeOnMainThread ()`, so I suppose we don't need to continue on the main thread here?

```suggestion
			await runner.Run (testAssemblies).ConfigureAwait (false);
```

baulig:
Had some minor cosmetic comments regarding `.ConfigureAwait (false)`, but looks good to me :-)

Thank you so much for looking into this!

tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
mandel-macaque: You are 100% right, good catch (I was working late).

tests/bcl-test/templates/tvOSApp/ViewController.cs
mandel-macaque: You are 100% right, good catch (I was working late).

mandel-macaque:
@baulig we can ignore those failures, there are due to HE0038, which is well known and a gc issue on mono, already fix that will mean we have to bump mono :)

spouliot:
@mandel-macaque please add links to known issues - it creates an entry on them (which tells us how many time a failure happens).

tests/bcl-test/templates/iOSApp/ViewController.cs
```diff
@@ -99,7 +99,7 @@ public async override void ViewDidLoad ()
 			
 			logger.Info ($"Tests run: {runner.TotalTests} Passed: {runner.PassedTests} Inconclusive: {runner.InconclusiveTests} Failed: {runner.FailedTests} Ignored: {runner.FilteredTests}");
 			if (options.TerminateAfterExecution)
-				TerminateWithSuccess ();
+				BeginInvokeOnMainThread (TerminateWithSuccess);
```
whitneyschmidt: This is what tells main() that the async call has completed?

tests/bcl-test/templates/common/TestRunner.NUnit/NUnitTestRunner.cs
```diff
@@ -35,7 +36,9 @@ public NUnitTestRunner (LogWriter logger) : base (logger)
 			builderSettings = new Dictionary<string, object> (StringComparer.OrdinalIgnoreCase);
 		}
 
-		public override void Run (IEnumerable<TestAssemblyInfo> testAssemblies)
+#pragma warning disable 1998
+		public override async Task Run (IEnumerable<TestAssemblyInfo> testAssemblies)
```
whitneyschmidt: async + param + function name = normal declaration for asynchronous functions in C#?

tests/bcl-test/templates/iOSApp/ViewController.cs
mandel-macaque: Correct, if you noticed, we are using ConfigureAwait (false) that means that we leave the main thread, but the method needs to be executed in that one, so we need to call it from the main thread to work.

tests/bcl-test/templates/common/TestRunner.NUnit/NUnitTestRunner.cs
mandel-macaque: In this case, @baulig is defining a async method, that is in reality, sync. He does that so that both runners have the same interface. That will make the c# compiler throw a warning, and we want it to be muted since we know what we are doing.

mandel-macaque:
@spouliot sure, known issues are:

* HE0038: https://github.com/xamarin/maccore/issues/581
* GC issue: https://github.com/mono/mono/issues/14975
* Assertion issue: https://github.com/xamarin/maccore/issues/1685
* mmptests: https://github.com/mono/mono/issues/11564

Issue #6298, opened 2019-06-14
[tests][tvOS] SecStatusCodeTest.ErrorDescriptionTest fails on a device with a non en language.


mandel-macaque:
The tests fails on a tvOS with the lang set to spanish.

mandel-macaque:
Of course, this is due to the fact that the error code is translated and my device is in spanish 🤦‍♂

PR #6299, opened 2019-06-14
[Test][Security] Ensure that test does not fail depending on the device lang.

mandel-macaque:
We have a test that depends on the lang of the device. If the device is
not using 'en' set it as inconclusive.

Fixes https://github.com/xamarin/xamarin-macios/issues/6298

### DIFF
--- a/tests/monotouch-test/Security/SecStatusCodeTest.cs
+++ b/tests/monotouch-test/Security/SecStatusCodeTest.cs
@@ -9,6 +9,7 @@
 
 using System;
 using NUnit.Framework;
+using System.Globalization;
 
 #if XAMCORE_2_0
 using Foundation;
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/Security/SecStatusCodeTest.cs
```diff
@@ -32,9 +33,14 @@ public void ErrorDescriptionTest ()
 			var desc = SecStatusCode.Success.GetStatusDescription ();
 			Assert.NotNull (desc, $"{nameof (desc)} not null");
 
-			// This value generated from objc enum documentation and very unlikely to change.
-			var noErr = "No error.";
-			Assert.That (desc, Is.EqualTo (noErr), $"{nameof (desc)} == {noErr}");
+			var lang = CultureInfo.CurrentUICulture.TwoLetterISOLanguageName;
+			if (lang.Equals ("en", StringComparison.InvariantCultureIgnoreCase)) {
```
spouliot: you compare native (iOS) and .net languages here
we try our best but they can mismatch (e.g. iOS allows "DE-fr" while .net does not)
would be safer to use native API to retrieve the current language

mandel-macaque:
Failing issues is known: https://github.com/mono/mono/issues/14876

PR #6307, opened 2019-06-14
[watch4] set right min version for arm64_32 based watch devices

lewurm:
Fixes the confusion around `libmono-native*` (see for example https://github.com/xamarin/xamarin-macios/commit/ce5ba1e41d526eddd9b61f68fae22a747831e9f2#commitcomment-33834491 ) when building with `MONO_BUILD_FROM_SOURCE=1`.

Not exactly happy with the naming, but at least it's consistently bad with the Mono SDK.


lewurm:
My vote is for `MIN_WATCHOS64_32_SDK_VERSION`.

I'll change the name in Mono SDK as well once we settled for a naming here (want to wait for @rolfbjarne ).

mandel-macaque:
+1 on what said, also, 5 reads like an S if the font is not good.

lewurm:
Thanks for voting! 🙂 will merge it into `mono-2019-06` and follow up there when the Mono SDK fix has landed.

lewurm:
https://github.com/mono/mono/pull/15202

PR #6308, opened 2019-06-14
[apidiff] Track changes based on d16-2

spouliot:
since the `xcode11` branch is based on `d16-2` (and not the current
`d16-1` stable).


spouliot:
known failures due to

* monotouch sim crash #6258

PR #6279, opened 2019-06-12
Bump VSMac to 8.1.0.2742 to fix msbuild issues

radical:

PR #6313, opened 2019-06-15
[nuget] Use xibuild to run nuget

radical:
Fix errors seen during `nuget restore` for tests:

```
Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xammac_tests/xammac_tests.csproj(213,3): error MSB4024: The imported project file "/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Mac/Xamarin.Mac.CSharp.targets" could not be loaded. Could not find file "/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Mac/Xamarin.Mac.CSharp.targets"
```

`nuget.exe` loads the msbuild assemblies, which ends up with no toolset
for mono (which has the fallback paths where XM/XI are installed). So,
use `xibuild` to use this correctly.

### DIFF
--- a/tests/generator/Makefile
+++ b/tests/generator/Makefile
@@ -5,7 +5,7 @@ include $(TOP)/Make.config
 all-local:: run-unit-tests
 
 build-unit-tests:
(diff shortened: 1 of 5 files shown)


radical:
This depends on the vsmac 8.1 bump, so will have to be part of https://github.com/xamarin/xamarin-macios/pull/6279 .

Issue #6322, opened 2019-06-17
[xcode11] mono-native-*/macOS/Modern tests fail to run due to DeviceCheck framework


rolfbjarne:
See https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10301/Test_20Report/

Both mono-native-compat/Mac/Modern and mono-native-unified/Mac/Modern fail to launch due to:

```
Dyld Error Message:
  Library not loaded: /System/Library/Frameworks/DeviceCheck.framework/Versions/A/DeviceCheck
  Referenced from: /Users/USER/*/mononativetests-compat-modern.app/Contents/MacOS/mononativetests-compat-modern
  Reason: image not found
```

rolfbjarne:
Fixed in https://github.com/xamarin/xamarin-macios/commit/69bd0d822397237eb2d480291f132428033e6563.

PR #6354, opened 2019-06-18
[wkwebkit] Update for Xcode 11 beta 2

spouliot:


src/wkwebkit.cs
```diff
@@ -493,16 +493,19 @@ interface WKUIDelegate {
 		void DidClose (WKWebView webView);
 
 		[iOS (10,0)][NoMac]
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'TBD' instead.")]
```
rolfbjarne: I wondered a while where `TBD` was... so maybe phrase is without an abbreviation? "Deprecated API, the replacement API will be implemented later" or something along those lines?

spouliot: It's not meant to be released (final) like this :) but it follows the final format
Added to https://github.com/xamarin/xamarin-macios/issues/6212

spouliot:
known issues
* mono-native*/Mac -> #6322
* msbuild -> xamarin/maccore#1266

PR #6356, opened 2019-06-18
[CoreGraphics] Fix lifetime of delegates passed to CGPatternCreate.

rolfbjarne:
We may get callbacks after the managed CGPattern instance has been
disposed/garbage collected, so make sure our delegates survives that long.

Since the delegates don't need any instance state, just make them static.


rolfbjarne:
Test failures are unrelated

* Xtro: https://github.com/xamarin/maccore/issues/1742
* MSBuild tests/iOS: https://github.com/xamarin/maccore/issues/1266

spouliot:
Should be back ported to `xcode11` or `d16-3`

rolfbjarne:
@monojenkins backport d16-2

PR #6389, opened 2019-06-20
[Network] Improve bindings for NWProtocolMetadata.

rolfbjarne:
It turns out the NWProtocolMetadata can contain metadata for different
protocols (Ip/Tls/Tcp). This is important; if someone tries to get a value for
one protocol and the metadata is for another protocol, then they invoke the
wrath of superior beings who will smite that poor someone with uninitialized
memory.

At that point there's not much left but to pray.

I don't like to depend on divine intervention, so I've modified the API here
to check if the metadata's protocol is the required type for the native API
we're calling, and if the check fails, we throw a nice and dependable managed
exception.

This is a functional breaking change; but if there are any lost souls who
likes to pray, they can always re-implement the P/Invokes themselves and skip
the sanity checks.

In addition I've renamed a few properties whose name didn't clearly specify
which protocol type they operate on.

Ref: Apple feedback FB6155967.
Ref: https://trello.com/c/1TW0BSKJ/145-fb6155967-nwipcreatemetadata-returns-uninitialized-metadata-in-ios-13


src/Network/NWProtocolMetadata.cs
```diff
@@ -87,7 +87,35 @@ public static NWProtocolMetadata CreateUdpMetadata ()
...
+		void CheckIsIP ()
+		{
+			if (!IsIP)
+				throw new InvalidOperationException ("This metadata is not IP metadata.");
+		}
+
+		void CheckIsTcp ()
+		{
+			if (!IsTcp)
+				throw new InvalidOperationException ("This metadata is not Tcp metadata.");
```
spouliot: `TCP` in the text itself

rolfbjarne:
Test failures are unrelated

PR #6397, opened 2019-06-21
[coreanimation] Update for Xcode 11 beta 1 and 2

spouliot:
There's no change in watchOS SDK headers... but they are marked as
available (to watchOS) inside iOS, tvOS and macOS. Not a new situation
and IIRC we reported it before (will check and report again if needed)


src/coreanimation.cs
```diff
@@ -584,6 +586,35 @@ interface CALayer : CAMediaTiming, NSSecureCoding {
...
+		[TV (13,0)][Mac (10, 15, onlyOn64: true)][iOS (13, 0)]
+		[Export ("cornerCurve")]
+		NSString CornerCurve { get; set; }
+
+		[NoWatch] // headers not updated
+		[TV (13,0)][Mac (10, 15, onlyOn64: true)][iOS (13, 0)]
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		[Static]
+		[Export ("cornerCurveExpansionFactor:")]
+		nfloat GetCornerCurveExpansionFactor (NSString curve);
```
dalexsoto: Why not `nfloat GetCornerCurveExpansionFactor ([BindAs (typeof (CACornerCurve))] NSString curve);` instead of the manual wrap? Since you are already doing it for the property.

spouliot: I forgot about it - and updated the others afterward (and missed that one) :D

PR #6401, opened 2019-06-21
[xharness] Fix Makefile-mac.inc generation after recent disruptive changes.

rolfbjarne:
This boils down to the makefile-generation code having the information it
needs (and that information being correct).

This fixes running of tests on other macOS bots (older/newer), because now we
can build the test package again.


rolfbjarne:
build

PR #6406, opened 2019-06-21
[system-dependencies] Fix xcode-provisioning extraction

VincentDondain:
This should fix:

```
        Extracting /tmp/x-provisioning/Xcode_11_Beta_2.xip...
        Installing Xcode 11.0 to /Applications/Xcode11-beta2.app...
mv: rename *.app to /Applications/Xcode11-beta2.app: No such file or directory
```

This happened because we're copying the .xip to /tmp/x-provisioning, then try to extract it but for some reason that extraction always end up in ~/Downloads.
Therefore the move fails because the extracted .app is not in /tmp/x-provisioning.
Note: Apple's Xcode .xip are always extracted with a name of Xcode[-beta].app so hardcode that.
I don't know if it's because the original .xip was downloaded to ~/Download or if macOS's archive utility is just ignoring its preferences (save expended files into same directory as archive).
I couldn't find the option to force the archiver to extract to a specific location (feedback welcome).

### DIFF
--- a/system-dependencies.sh
+++ b/system-dependencies.sh
@@ -336,10 +336,17 @@ function install_specific_xcode () {
 	local XCODE_NAME=`basename $XCODE_URL`
 	local XCODE_DMG=$PROVISION_DOWNLOAD_DIR/$XCODE_NAME
 
+	# Apple always uses the same name when extracting Xcode, only adding -beta if needed
+	if [[ $XCODE_NAME == *"Beta"* ]]; then
+		local XCODE_DOWNLOAD_PATH=$HOME/Downloads/Xcode-beta.app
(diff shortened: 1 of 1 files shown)


VincentDondain:

VincentDondain:
FYI this is what caused https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2786559 to fail at the provisioning step.

rolfbjarne:
This is a preference for Archive Utility 😠 

Screenshot 2019-06-24 10 20 27

so this fix isn't always correct.

rolfbjarne:
Superseded by #6413.

PR #6408, opened 2019-06-21
[authenticationservices] Update for Xcode 11 beta 1 and 2

spouliot:
Also enable the framework for macOS, watchOS and tvOS


rolfbjarne:
build

src/authenticationservices.cs
```diff
@@ -192,5 +227,481 @@ interface ASWebAuthenticationSession {
...
+	[Native]
+	enum ASAuthorizationAppleIdProviderCredentialState : long {
+		Revoked,
+		Authorized,
+		NotFound,
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+	[BaseType (typeof (NSObject), Name = "ASAuthorizationAppleIDProvider")]
+	[DisableDefaultCtor] // NSInvalidArgumentException Reason: -[ASAuthorizationAppleIDRequest init]: unrecognized selector sent to instance 0x600000950000
```
rolfbjarne: How are you supposed to create/get an instance of this type then? From the code sample here, it seems you're supposed to use the default constructor: https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidprovider?language=objc

src/authenticationservices.cs
```diff
@@ -192,5 +227,481 @@ interface ASWebAuthenticationSession {
...
+	enum ASAuthorizationAppleIdButtonStyle : long {
+		White = 0,
+		WhiteOutline = 1,
+		Black = 2,
+	}
+
+	[TV (13,0), NoWatch, Mac (10,15, onlyOn64: true), iOS (13,0)]
+	[BaseType (typeof (UIControl))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationAppleIDButton {
```
rolfbjarne: You forgot one: `ASAuthorizationAppleIDButton` => `ASAuthorizationAppleIdButton` :smile:

dalexsoto:
I see some `[Unavailable (PlatformName.UIKitForMac)]`, I think we also need the following on those until we either teach the generator or we do the right thing with that information from the IDE.

```
```

src/authenticationservices.cs
```diff
@@ -192,5 +227,481 @@ interface ASWebAuthenticationSession {
...
+		IntPtr Constructor (ASAuthorizationRequest[] authorizationRequests);
+	}
+
+	interface IASAuthorizationProvider { }
+
+	[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+	[Protocol]
+	interface ASAuthorizationProvider { }
+
+	// [Unavailable (PlatformName.UIKitForMac)]
```
VincentDondain: Why are those commented out?

src/authenticationservices.cs
```diff
@@ -192,5 +227,481 @@ interface ASWebAuthenticationSession {
...
+	}
+
+	[NoWatch, NoTV, Mac (10,15, onlyOn64: true), iOS (13,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationSingleSignOnProvider : ASAuthorizationProvider {
+
+		[Static]
+		[Export ("authorizationProviderWithIdentityProviderURL:")]
+		ASAuthorizationSingleSignOnProvider CreateProvider (NSUrl url);
```
VincentDondain: The parameter could be `identityProviderUrl` to give more context to people using this API.

src/authenticationservices.cs
spouliot: I added the value later :)

spouliot:
```
14:57:05 /Applications/Xcode11-beta2.app/Contents/Developer/usr/bin/make test-local
14:57:05 /Applications/Xcode11-beta2.app/Contents/Developer/usr/bin/make test-local
14:57:05 CSC     [xammac_net_4_5] xammac_net_4_5_corlib_test.resources.dll
14:57:06 Unable to load assembly: ../../../class/lib/xammac_net_4_5/nunitlite.dll
14:57:06 
make[10]: *** [../../../class/lib/xammac_net_4_5/nunitlite.dll] Error 1
14:57:06 make[9]: *** [do-all] Error 2
14:57:06 make[8]: *** [all-recursive] Error 1
14:57:06 make[7]: *** [../../build/deps/nunit-xammac_net_4_5.stamp] Error 2
14:57:06 make[7]: *** Waiting for unfinished jobs....
```
...
```
14:58:02 make[1]: *** [.stamp-build-tools64] Error 2
```

spouliot:
build

src/authenticationservices.cs
spouliot:
```Objective-C exception thrown.  Name: NSInvalidArgumentException Reason: -[ASAuthorizationAppleIDRequest init]: unrecognized selector sent to instance 0x600001421770
Native stack trace:
	0   CoreFoundation                      0x0000000102e3422e __exceptionPreprocess + 350
	1   libobjc.A.dylib                     0x00000001101d4f57 objc_exception_throw + 48
	2   CoreFoundation                      0x0000000102e547f4 -[NSObject(NSObject) doesNotRecognizeSelector:] + 132
	3   AuthenticationServices              0x000000010ffbe7b5 -[ASAuthorizationRequest init] + 25
	4   introspection                       0x00000001026a3f59 xamarin_dyn_objc_msgSend + 217
	5   ???                                 0x000000013e4b8902 0x0 + 5340104962
```
So it's not `init` itself but something it calls... we might be missing some configuration steps

spouliot: oops, was simply copy/pasted on the wrong line (comment is fine)

FinHorsley:
Sorry if this is the wrong place, but i've come from https://github.com/xamarin/xamarin-macios/wiki/AuthenticationServices-iOS-xcode11-beta1. I'm familiar with Xamarin, but this is my first time actually digging into the underlying code, so apologies for any rookie questions 
Is `ASPresentationAnchor` intentionally missing, or is this something that is coming in a later release? 
Also, why is `IASWebAuthenticationPresentationContextProviding` just an interface? Some of the examples given in Swift are inheriting from a class, and look like they don't require overrides for `Handle` and `Dispose`?
```
extension MyClass: ASWebAuthenticationPresentationContextProviding {
    @available(iOS 13.0, *)
    func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
        return UIApplication.shared.keyWindow!
    }
}
```

spouliot:
@FinHorsley no one is actively monitoring closed PR to the repo. The broader developer community would be the best and quickest place for additional troubleshooting help on this issue.
Posting a question on Stack Overflow [0] or the Xamarin Forums [1][2] would be the best next step.

[0] https://stackoverflow.com/questions/tagged/xamarin
[1] https://forums.xamarin.com/categories/ios
[2] https://forums.xamarin.com/categories/mac

> Is ASPresentationAnchor intentionally missing ?

Yes. `ASPresentationAnchor` is simply a cross-platform synonym of `UIWindow` or `NSWindow`, depending if you're on iOS or macOS. See below:

```diff
+#if __has_include(<UIKit/UIWindow.h>)
+typedef UIWindow * ASPresentationAnchor;
+#endif
+#elif __has_include(<AppKit/AppKit.h>)
+#import <AppKit/AppKit.h>
+typedef NSWindow * ASPresentationAnchor;
+#endif
```

> why is IASWebAuthenticationPresentationContextProviding just an interface? 

Because it's a protocol and (ObjC/Swift) classes *conforms* to a protocol - they don't subclass it (unless they are themselves protocol definitions). The closest analogy in C# is an interface - which is why it's bound that way.

```diff
+@protocol ASWebAuthenticationPresentationContextProviding;
```
https://developer.apple.com/documentation/authenticationservices/aswebauthenticationpresentationcontextproviding

> look like they don't require overrides for Handle and Dispose?

Those are C#/.net construct that are required for bindings. You normally don't have to call those yourself.

FinHorsley:
@spouliot Thanks for such a detailed response 🙂 

For anyone else having the same problem, here is the solution we came up with. **Note** `KeyWindow` is deprecated in iOS 13 (important for multi scene apps)

```
public sealed class ASWebAuthenticationDelegate : NSObject, IASWebAuthenticationPresentationContextProviding
{
       UIWindow IASWebAuthenticationPresentationContextProviding.GetPresentationAnchor(ASWebAuthenticationSession session)
            => UIApplication.SharedApplication.KeyWindow;
}
```

```
//This is an iOS 13 only feature, so running this on <= iOS 12 will cause your app to crash 
if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
{
         asWebAuthenticationSession.PresentationContextProvider = new ASWebAuthenticationDelegate();
}

PR #6415, opened 2019-06-24
[PencilKit] Add Xcode 11 Beta 1 & Beta 2 bindings

tj-devel709:


src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		void EndUsingTool (PKCanvasView canvasView);
+	}
+
+	interface IPKCanvasViewDelegate {}
+
+	[iOS (13, 0), NoMac]
+	[BaseType (typeof (UIScrollView))]
+	interface PKCanvasView : PKToolPickerObserver {
+
+		// This exists in the base class
```
whitneyschmidt: Should this comment be left in?

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+using UIImage = AppKit.NSImage;
+
+using UIScrollViewDelegate = Foundation.NSObjectProtocol;
+using UIScrollView = Foundation.NSObject;
+using UIGestureRecognizer = Foundation.NSObject;
+using UIResponder = Foundation.NSObject;
+using UIView = Foundation.NSObject;
+using UIWindow = Foundation.NSObject;
+#else
+using UIKit;
```
whitneyschmidt: I'm confused re this #if/#else macro - is UIKit only available on non-mac devices?

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		Marker,
+	}
+
+	[iOS (13, 0), NoMac]
+	[BaseType (typeof (NSObject))]
+	[Model (AutoGeneratedName = true)] [Protocol]
+	interface PKCanvasViewDelegate : UIScrollViewDelegate {
+
+		[Export ("canvasViewDrawingDidChange:")]
+		void DrawingDidChange (PKCanvasView canvasView);
```
whitneyschmidt: Nit: should this be DrawingChanged? Or do we try and keep the C# name as close to the ObjC name as possible?

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		[Export ("width")]
+		nfloat Width { get; }
+	}
+
+	[iOS (13, 0), NoMac]
+	[BaseType (typeof (PKTool))]
+	[DisableDefaultCtor]
+	interface PKLassoTool {
+
+		[DesignatedInitializer]
```
whitneyschmidt: Does (instancetype)init NS_DESIGNATED_INITIALIZER; translate into [DesignatedInitializer]? Why is it 
"designated" in the first place?

src/pencilkit.cs
rolfbjarne: @whitneyschmidt correct, UIKit is not available on macOS, only iOS, tvOS and watchOS (on macOS there's AppKit instead, which is much older than UIKit).

src/pencilkit.cs
rolfbjarne: How close to ObjC we want to be is a subjective decision: in general we try to keep close to ObjC unless it makes sense not to (usually because it makes it easier to map between ObjC and C# for customers: like reading documentation, samples, stackoverflow, etc. If we name things differently, it can become quite hard to write the C# code if you know the particular ObjC API).

However, we do change things. For instance: we try to not add redundant information (thus `DrawingDidChange` and not `CanvasViewDrawingDidChange`, since the type is a `*CanvasView*` of some sort). Another typical case is that we move the first parameter name to the actual parameter (for instance `drawWithFoo:` would become `Draw (object foo)`).

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+	[iOS (13, 0), Mac (10, 15, onlyOn64: true)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface PKDrawing : NSCopying, NSSecureCoding {
+
+		[Field ("PKAppleDrawingTypeIdentifier")]
+		NSString AppleDrawingTypeIdentifier { get; }
+
+		[DesignatedInitializer]
+		[Export("init")]
```
rolfbjarne: Nitpick:

```suggestion
		[Export ("init")]
```

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		IntPtr Constructor();
```
rolfbjarne:
```suggestion
		IntPtr Constructor ();
```

src/pencilkit.cs
rolfbjarne: Yes, that's what `[DesignatedInitializer]` means. [This](https://stackoverflow.com/a/26186421/183422) explains the Objective-C part fairly well.

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		nfloat Width { get; }
+	}
+
+	[iOS (13, 0), NoMac]
+	[BaseType (typeof (PKTool))]
+	[DisableDefaultCtor]
+	interface PKLassoTool {
+
+		[DesignatedInitializer]
+		[Export("init")]
```
rolfbjarne:
```suggestion
		[Export ("init")]
```

src/pencilkit.cs
```diff
@@ -0,0 +1,243 @@
...
+		IntPtr Constructor();
```
rolfbjarne:
```suggestion
		IntPtr Constructor ();
```

tests/xtro-sharpie/iOS-PencilKit.todo
```diff
@@ -1,53 +1,3 @@
...
-!missing-field! PKInkTypePen not bound
-!missing-field! PKInkTypePencil not bound
-!missing-protocol! PKCanvasViewDelegate not bound
-!missing-protocol! PKToolPickerObserver not bound
-!missing-selector! +PKInkingTool::defaultWidthForInkType: not bound
-!missing-selector! +PKInkingTool::maximumWidthForInkType: not bound
-!missing-selector! +PKInkingTool::minimumWidthForInkType: not bound
-!missing-selector! +PKToolPicker::sharedToolPickerForWindow: not bound
-!missing-selector! PKCanvasView::allowsFingerDrawing not bound
+# adding these since weak delegate exists in base class
```
rolfbjarne: Entires that won't be fixed should be moved to the `iOS-PencilKit.ignore` file.

src/pencilkit.cs
spouliot: not strictly needed - but it explains why it's missing (which someone could comment on)

src/pencilkit.cs
spouliot: This can be simplified with `[DesignatedDefaultCtor]` on the type
i.e. remove `[DisableDefaultCtor]` and `[DesignatedInitializer][Export("init")]IntPtr Constructor();`

src/pencilkit.cs
spouliot: same, use the simpler `[DesignatedDefaultCtor]`

src/pencilkit.cs
whitneyschmidt: thanks @rolfbjarne!

src/pencilkit.cs
tj-devel709: So just to make sure, you are suggesting that I remove the method completely and where DisableDefaultCtor is right now, I change that to DesignatedDefaultCtor?

tj-devel709: Also I have those DisableDefaultCtor 's there since some of them inherit from PKTool which has a disable default constructor, so is that still okay?

tj-devel709: @spouliot Sorry, I'm not sure if I was supposed to tag you.

dalexsoto:
build

PR #6417, opened 2019-06-25
[GameController] GCGamepad is deprecated in the headers, so mark it as such. Fixes xamarin/maccore#1742.

rolfbjarne:
Fixes https://github.com/xamarin/maccore/issues/1742.

### DIFF
--- a/src/gamecontroller.cs
+++ b/src/gamecontroller.cs
@@ -123,6 +123,7 @@ namespace GameController {
 	[Mac (10,9, onlyOn64: true)]
 	[Deprecated (PlatformName.MacOSX, 10, 12)]
 	[Deprecated (PlatformName.iOS, 10, 0)]
+	[Deprecated (PlatformName.TvOS, 10, 0)]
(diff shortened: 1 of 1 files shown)


src/gamecontroller.cs
```diff
@@ -121,6 +121,8 @@ partial interface GCControllerDirectionPad {
 
 	[iOS (7,0)]
 	[Mac (10,9, onlyOn64: true)]
+	[Deprecated (PlatformName.MacOSX, 10, 12)]
+	[Deprecated (PlatformName.iOS, 10, 0)]
```
spouliot: You'll need a tvOS entry too - it's implied as always available in tvOS (by `[iOS (7,0)]`) but that does not apply to deprecation

spouliot: Still strange that this happens only in some, not all bots :|

rolfbjarne:
@monojenkins backport d16-2
@monojenkins backport d16-3

PR #6432, opened 2019-06-26
[pushkit] Update for Xcode 11 beta 1

spouliot:
No change in beta 2

This enable PushKit on watchOS and macOS.

Availability macros suggest this is available on tvOS 13 but the headers
are not in the tvOS SDK (at least in beta 2)


src/pushkit.cs
```diff
@@ -53,19 +62,32 @@ interface PKPushRegistry {
 	[Static]
 	interface PKPushType {
 		
+		[Unavailable (PlatformName.UIKitForMac)]
```
dalexsoto: `[Advice ("This API is not available when using UIKit on macOS.")]`

src/pushkit.cs
```diff
@@ -53,19 +62,32 @@ interface PKPushRegistry {
...
+		[NoWatch]
+		[NoMac]
 		[Field ("PKPushTypeVoIP")]
 		NSString Voip { get; }
 
 		[iOS (9,0)]
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use directly from watchOS instead.")]
+		[Unavailable (PlatformName.UIKitForMac)]
```
dalexsoto: `[Advice ("This API is not available when using UIKit on macOS.")]`

PR #6442, opened 2019-06-26
[carplay] Update for Xcode 11 beta 1

whitneyschmidt:
#if false because Apple is trying to use an API that isn't yet exposed. Xtro-ignore additions due to this too.


src/carplay.cs
```diff
@@ -785,8 +805,15 @@ interface CPWindow {
 
 		[Export ("mapButtonSafeAreaLayoutGuide")]
 		UILayoutGuide MapButtonSafeAreaLayoutGuide { get; }
+
+#if false
+// Apple is trying to use CPTemplateApplicationScene but it is not exposed. FB6231565
```
whitneyschmidt: This needs to be updated w/ the radar #, right now it's the Apple bug report #.

src/carplay.cs
```diff
@@ -378,8 +382,14 @@ interface CPListTemplateDelegate {
 	[BaseType (typeof (NSObject))]
 	interface CPManeuver : NSCopying, NSSecureCoding {
 
+		[Introduced (PlatformName.iOS, 12,0)]
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'CPManeuver.symbolImage' instead.")]
```
dalexsoto: Just a small casing issue: 

```suggestion
		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'CPManeuver.SymbolImage' instead.")]
```

src/carplay.cs
```diff
@@ -785,8 +805,15 @@ interface CPWindow {
...
 		UILayoutGuide MapButtonSafeAreaLayoutGuide { get; }
+
+#if false
+// Apple is trying to use CPTemplateApplicationScene but it is not exposed. FB6231565
+		[NullAllowed, Export ("templateApplicationScene", ArgumentSemantic.Weak)]
+		CPTemplateApplicationScene TemplateApplicationScene { get; set; }
+#endif
 	}
 
+
```
VincentDondain: Please remove that whitespace :P

mattleibow:
```suggestion
```

src/carplay.cs
spouliot: yes, we now track them in https://github.com/xamarin/maccore/issues?q=is%3Aissue+is%3Aopen+label%3Aapple-feedback

tests/xtro-sharpie/iOS-CarPlay.ignore
```diff
@@ -1,2 +1,4 @@
 !incorrect-protocol-member! CPBarButtonProviding::backButton is REQUIRED and should be abstract
 !incorrect-protocol-member! CPBarButtonProviding::setBackButton: is REQUIRED and should be abstract
+!missing-selector! CPWindow::setTemplateApplicationScene: not bound
+!missing-selector! CPWindow::templateApplicationScene not bound
```
spouliot: you can add comments, any line that starts with `#`, to the internal bug you filed
not critical since `git blame` can find the commit/PR and that will link with it...

src/carplay.cs
tj-devel709: So this deprecated attribute, when we use it and add a message, the user will receive the message when they try to use SymbolSet in iOS13?

tj-devel709:
TJ Lambert approved these changes now - not sure how to approve changes!

src/carplay.cs
dalexsoto: @tj-devel709 yes, Visual Studio will show this warning depending on the minimum version registered in the app's info.plist

src/carplay.cs
whitneyschmidt: Added maccore issue #, let me know if this is the correct format!

tests/xtro-sharpie/iOS-CarPlay.ignore
whitneyschmidt: Added the issue # as a comment here too...let me know if the formatting is correct!

src/carplay.cs
```diff
@@ -807,13 +807,12 @@ interface CPWindow {
 		UILayoutGuide MapButtonSafeAreaLayoutGuide { get; }
 
 #if false
-// Apple is trying to use CPTemplateApplicationScene but it is not exposed. FB6231565
+// Apple is trying to use CPTemplateApplicationScene but it is not exposed. xamarin/maccore#1822
```
VincentDondain: Could you just use the full link? https://github.com/xamarin/maccore/issues/1822

Simpler

tests/xtro-sharpie/iOS-CarPlay.ignore
VincentDondain: Same in your comment here, please use the full link: https://github.com/xamarin/maccore/issues/1822

PR #6457, opened 2019-06-28
[mediaaccessibility] Update for Xcode 11 beta 1

spouliot:
No change in beta 2

Unit tests added - but new API behaviour is not quite what I expected.
Maybe future docs will make this clearer...

### DIFF
--- a/src/MediaAccessibility/MAImageCaptioning.cs
+++ b/src/MediaAccessibility/MAImageCaptioning.cs
@@ -1,0 +1,57 @@
+using System;
+using System.Runtime.InteropServices;
+
+using ObjCRuntime;
+using CoreFoundation;
+using CoreGraphics;
+using CoreText;
+using Foundation;
+
+namespace MediaAccessibility {
+
+	[TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+	public static partial class MAImageCaptioning {
+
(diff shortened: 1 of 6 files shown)


src/MediaAccessibility/MAImageCaptioning.cs
```diff
@@ -0,0 +1,57 @@
...
+using CoreGraphics;
+using CoreText;
+using Foundation;
+
+namespace MediaAccessibility {
+
+	[TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+	public static partial class MAImageCaptioning {
+
+		[DllImport (Constants.MediaAccessibilityLibrary)]
```
whitneyschmidt: Based on the .md it looks like this is coming from C/C++. I see the using directives at the top of the file...what is [DlImport...] for? Is it another way of including files?

spouliot: yes, it's a C based framework :)

so this uses the _normal_ .net mechanisms for interop, known as **p/invoke**

`[DlImport...]` tells where the native (C) symbol reside so it can be looked up at runtime (JIT) or at build time (AOT)

you can learn more about p/invoke it on microsoft.com
but there's a lot of mono specific/important pieces in https://www.mono-project.com/docs/advanced/pinvoke/ too

PR #6467, opened 2019-07-01
[xharness] Propagate the BUILD_REVISION variable to test apps to specify if we're in CI or not. Fixes xamarin/maccore#649.

rolfbjarne:
We have tests whose behavior changes when executed on CI, and those tests use
the BUILD_REVISION variable to detect where they're being executed. For this
to work we need to propagate the BUILD_REVISION variable to the test
executable.

Hopefully fixes https://github.com/xamarin/maccore/issues/649 now.


rolfbjarne:
@monojenkins backport xcode11

PR #6475, opened 2019-07-02
[CoreBluetooth] Update bindings for Xcode 11 beta 1 & beta 2.

mandel-macaque:
Updated the bindings to match Xcode 11 and took the time to add a StrongDict (new keys were added) to improve the API a little.


src/corebluetooth.cs
```diff
@@ -7,6 +7,8 @@
 //
 // Copyright 2011-2013 Xamarin Inc
 //
+using System.ComponentModel;
+
```
whitneyschmidt: extra line intended?

src/corebluetooth.cs
```diff
@@ -7,6 +7,8 @@
 //
 // Copyright 2011-2013 Xamarin Inc
```
whitneyschmidt: nit: update

src/corebluetooth.cs
```diff
@@ -43,6 +45,59 @@ interface CBCentralInitOptions {
 	interface CBManager {
 		[Export ("state", ArgumentSemantic.Assign)]
 		CBManagerState State { get; }
+
+		[iOS (13,0), TV (13,0), Watch (6,0), Mac (10, 15)]
```
spouliot: I think we still need the `only64: true` on the bindings...

src/corebluetooth.cs
```diff
@@ -43,6 +45,59 @@ interface CBCentralInitOptions {
...
+		[Export ("authorization", ArgumentSemantic.Assign)]
+		CBManagerAuthorization Authorization { get; }
+	}
+
+	[iOS (13,0), TV (13,0), Watch (6,0)]
```
spouliot: missing `[NoMac]` ? unless it's present (then missing `Mac (10,15, onlyOn64: true)]` or you're inside `!MONOMAC` (then better add `[NoMac]` anyway to ease review and help the IDE)

src/corebluetooth.cs
```diff
@@ -43,6 +45,59 @@ interface CBCentralInitOptions {
...
+	[StrongDictionary ("CBConnectionEventMatchingOptionsKeys")]
+	interface CBConnectionEventMatchingOptions {
+		NSUuid [] PeripheralUuids { get; set; }
+		CBUUID [] ServiceUuids { get; set;}
+	}
+
+	[iOS (13,0), TV (13,0), Watch (6,0)]
```
spouliot: same

src/corebluetooth.cs
```diff
@@ -43,6 +45,59 @@ interface CBCentralInitOptions {
...
+		NSString PeripheralUuidsKey { get; }
+		[Field ("CBConnectionEventMatchingOptionServiceUUIDs")]
+		NSString ServiceUuidsKey { get; }
+	}
+
+	[StrongDictionary ("CBConnectPeripheralOptionsKeys")]
+	interface CBConnectPeripheralOptions {
+		[Mac (10,13)] [iOS (6,0)]
+		bool NotifyOnConnection { get; set; }
+		bool NotifyOnDisconnection { get; set; }
```
spouliot: that one has no availability ? looks weird (add comment if that's the case)
if possible move some (all) availability attributes on the type itself

src/corebluetooth.cs
```diff
@@ -43,6 +45,59 @@ interface CBCentralInitOptions {
...
+	}
+
+	[Static]
+	[Internal]
+	interface CBConnectPeripheralOptionsKeys {
+		[Mac (10,13)] [iOS (6,0)]
+		[Field ("CBConnectPeripheralOptionNotifyOnConnectionKey")]
+		NSString NotifyOnConnectionKey { get; }
+		[Field ("CBConnectPeripheralOptionNotifyOnDisconnectionKey")]
+		NSString NotifyOnDisconnectionKey { get; }
```
spouliot: same

src/corebluetooth.cs
```diff
@@ -858,6 +946,8 @@ interface CBPeripheralManager {
 		NSString RestoredStateAdvertisementDataKey { get; }
 
 #if !MONOMAC || !XAMCORE_4_0
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'CBManager.Authorization' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use 'CBManager.Authorization' instead.")]
```
spouliot: looks like tvOS should also have the deprecation

spouliot:
should be update tp beta 3 before merging

src/corebluetooth.cs
rolfbjarne: Actually if the build doesn't fail and tests don't fail, it shouldn't matter anymore (and any lingering need will be resolved before the final release anyway; it's on our todo list).

mandel-macaque: I suppose since that we are not longer supporting 32b, it makes sense, nevertheless, I'll add it. It won't hurt

mandel-macaque:
@spouliot I checked the headers, looks like it beta3 ready since they just changed the availability macro. :) :success:

spouliot:
@mandel-macaque nice, please update the wiki and the description (when you merge)

src/CoreBluetooth/Enums.cs
```diff
@@ -172,4 +172,26 @@ public enum CBPeripheralManagerConnectionLatency : long {
...
 	}
+
+	[iOS (13,0), TV (13,0), Watch (6,0), Mac (10,15)]
+	[Native]
+	public enum CBConnectionEvent : long {
+		Disconnected = 0,
+		Connected = 1,
+	}
+
+	[Flags, iOS (13,0), TV (13,0), Watch (6,0)]
```
spouliot: Mac (No or version) is unspecified

src/corebluetooth.cs
mandel-macaque: yes, I like to separate the bcl using from ours

src/corebluetooth.cs
mandel-macaque: That is the only one that had no availability attributes from the old constant used: https://github.com/xamarin/xamarin-macios/blob/master/src/corebluetooth.cs#L105 

The other attributes I used from the already present (set to obsolete) constants. So I don't think I can move [Mac (10,13)] [iOS (6,0)] to the top of the class :/

mandel-macaque:
@spouliot my fault, I though that [Introduced (PlatformName.UIKitForMac, 13,0)] would be the correct thing to do.

mandel-macaque:
Midi test failures are un related, I have seen this in other PRs. Looks like a device doing something funny.

spouliot:
It's `[NoMac]` not `[NotMac]`
Please build and test locally before committing :)

spouliot:
@mandel-macaque it might be `[Introduced (PlatformName.UIKitForMac, 13,0)]` even if not available on macOS itself (a bit weird for CoreBluetooth)

PR #6479, opened 2019-07-02
[quicklook] Update for Xcode11 beta 1-3

whitneyschmidt:
Changes for b1-3, no updates from b3 according to wiki page.

### DIFF
--- a/src/quicklook.cs
+++ b/src/quicklook.cs
@@ -131,15 +131,15 @@ namespace QuickLook {
 
 		[iOS (13,0)]
 		[Export ("previewController:editingModeForPreviewItem:"), DelegateName("QLEditingMode"), DefaultValue ("QLPreviewItemEditingMode.Disabled")]
-		QLPreviewItemEditingMode EditingModeForPreviewItem (QLPreviewController controller, IQLPreviewItem item);
+		QLPreviewItemEditingMode GetEditingMode (QLPreviewController controller, IQLPreviewItem previewItem);
 
 		[iOS (13,0)]
 		[Export ("previewController:didUpdateContentsOfPreviewItem:"), EventArgs ("QLPreviewControllerDelegateDidUpdate")]
-		void DidUpdateContentsOfPreviewItem (QLPreviewController controller, IQLPreviewItem item);
+		void DidUpdateContents (QLPreviewController controller, IQLPreviewItem previewItem);
 
 		[iOS (13,0)]
 		[Export ("previewController:didSaveEditedCopyOfPreviewItem:atURL:"), EventArgs ("QLPreviewControllerDelegateDidSave")]
-		void DidSaveEditedCopyOfPreviewItem (QLPreviewController controller, IQLPreviewItem item, NSUrl modifiedContentsUrl);
+		void DidSaveEditedCopy (QLPreviewController controller, IQLPreviewItem previewItem, NSUrl modifiedContentsUrl);
 
 #endif
 	}
--- a/src/quicklook.cs
+++ b/src/quicklook.cs
@@ -7,6 +7,7 @@
 //
 // Copyright 2009, Novell, Inc.
 // Copyright 2012 Xamarin Inc
+// Copyright 2019 Microsoft Corporation
 //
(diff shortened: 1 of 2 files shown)


src/quicklook.cs
```diff
@@ -119,6 +128,19 @@ interface QLPreviewControllerDelegate {
 		[Export ("previewController:transitionViewForPreviewItem:"), DelegateName ("QLTransitionView"), DefaultValue (null)]
 		[return: NullAllowed]
 		UIView TransitionViewForPreviewItem (QLPreviewController controller, IQLPreviewItem item);
+
+		[iOS (13,0)]
+		[Export ("previewController:editingModeForPreviewItem:"), DelegateName("QLEditingMode"), DefaultValue ("QLPreviewItemEditingMode.Disabled")]
+		QLPreviewItemEditingMode EditingModeForPreviewItem (QLPreviewController controller, IQLPreviewItem item);
```
spouliot: prefix: use a verb for methods, e.g. `Get...`
suffix: avoid repeating parameter names, e.g. drop `ForPreviewItem`

so `GetEditingMode`

note: ObjC selectors are `{method_name}{param1}:{param2}` - i.e. there's no `:` between the method name and the **first** parameter name. In doubt check the name used by Apple for the Swift API.

src/quicklook.cs
```diff
@@ -119,6 +128,19 @@ interface QLPreviewControllerDelegate {
...
+
+		[iOS (13,0)]
+		[Export ("previewController:didUpdateContentsOfPreviewItem:"), EventArgs ("QLPreviewControllerDelegateDidUpdate")]
+		void DidUpdateContentsOfPreviewItem (QLPreviewController controller, IQLPreviewItem item);
```
spouliot: same -> `DidUpdateContents`

src/quicklook.cs
```diff
@@ -119,6 +128,19 @@ interface QLPreviewControllerDelegate {
...
+
+		[iOS (13,0)]
+		[Export ("previewController:didSaveEditedCopyOfPreviewItem:atURL:"), EventArgs ("QLPreviewControllerDelegateDidSave")]
+		void DidSaveEditedCopyOfPreviewItem (QLPreviewController controller, IQLPreviewItem item, NSUrl modifiedContentsUrl);
```
spouliot: same -> `DidSaveEditedCopy`

src/quicklook.cs
rolfbjarne: This also means the parameter can be better named as `previewItem`:

```csharp
QLPreviewItemEditingMode GetEditingMode (..., IQLPreviewItem previewItem);
```

src/quicklook.cs
rolfbjarne: same -> `IQLPreviewItem previewItem`

src/quicklook.cs
rolfbjarne: same -> `IQLPreviewItem previewItem`

src/quicklook.cs
whitneyschmidt: Updated this and others! Thanks @spouliot and @rolfbjarne :)

src/quicklook.cs
```diff
@@ -119,6 +128,19 @@ interface QLPreviewControllerDelegate {
 		[Export ("previewController:transitionViewForPreviewItem:"), DelegateName ("QLTransitionView"), DefaultValue (null)]
 		[return: NullAllowed]
 		UIView TransitionViewForPreviewItem (QLPreviewController controller, IQLPreviewItem item);
+
+		[iOS (13,0)]
+		[Export ("previewController:editingModeForPreviewItem:"), DelegateName("QLEditingMode"), DefaultValue ("QLPreviewItemEditingMode.Disabled")]
```
tj-devel709: How did you know to use the DelegateName and DefaultValue attributes and their values?

whitneyschmidt: Looking at documentation and other similar code - ie, transitionViewForPreviewItem has DelegateName + DefaultValue. Then asking Alex :)

PR #6485, opened 2019-07-03
[AppKit] Implement [NSView sortSubviewsUsingFunction:context:]

rolfbjarne:


src/AppKit/NSView.cs
```diff
@@ -0,0 +1,55 @@
...
+				var a = (NSView) Runtime.GetNSObject (view1);
+				var b = (NSView) Runtime.GetNSObject (view2);
+				return (nint) (long) data.Comparer (a, b);
+			} catch (Exception e) {
+				data.Exception = e;
+				return (nint) (long) NSComparisonResult.Same;
+			}
+		}
+
+		public void SortSubviews (Func<NSView, NSView, NSComparisonResult> comparer)
```
spouliot: we generally define a custom delegate when there's more than one generic argument - but I think this one is self documenting.

abock: It might be nice to have a `Comparison<T>` override though.

src/AppKit/NSView.cs
```diff
@@ -0,0 +1,55 @@
...
+using ObjCRuntime;
+
+namespace AppKit {
+
+	public partial class NSView {
+
+		delegate nint view_compare_func (IntPtr view1, IntPtr view2, IntPtr context);
+		static view_compare_func view_comparer = view_compare;
+
+		class SortData
```
abock: `sealed`

src/AppKit/NSView.cs
rolfbjarne: We can't propagate `T` to the P/Invoke callback (`view_compare`), so a `Comparison<T>` overload wouldn't be trivial to implement (it would probably need to use reflection to call the managed comparison function).

spouliot:
I think this should be back ported to `d16-3` or `xcode11` (16.4 is a long wait)

rolfbjarne:
> I think this should be back ported to `d16-3` or `xcode11` (16.4 is a long wait)

I think we're going to re-branch d16-3 from master once mono 2019-06 has been merged.

Therzok:
Was this included in d16-3?

Therzok:
I have answered my own question, it's a yes.

PR #6497, opened 2019-07-04
[mmptest] Remove the link-gdiplus-1/2 tests, they can't be ported to Unified. Partial fix for #4975.

rolfbjarne:
These tests verify that a linked project that references Bitmap[Data] from
System.Drawing ends up with a libgdiplus.dylib (or not) in the .app.

The problem when porting to Unified is that the only way to depend on
libgdiplus.dylib is using the system mono, and in that mode mmp doesn't allow
linking.

So just remove these tests.

Fixes part of https://github.com/xamarin/xamarin-macios/issues/4975.


rolfbjarne:
Test failures are unrelated

* monotouch-test: https://github.com/xamarin/maccore/issues/1834
* introspection: https://github.com/xamarin/maccore/issues/1835

PR #6503, opened 2019-07-05
[xcode11] Remove the WatchKit framework from iOS while keeping API stability. Fixes #6492.

### LINKED ISSUE #6492: [xcode11] The WatchKit framework has been removed from the iOS SDK
The WatchKit framework has been removed from the iOS SDK.

In [Xcode 11 beta 3's release notes][1], Apple explains:

> The WatchKit framework is no longer included in the iOS SDK. If you’re using WatchKit APIs from iOS, you need to remove this use. The WatchKit framework remains available on watchOS. If you’re using WatchKit APIs from iOS to infer availability of features on the paired Apple Watch, include information about your use case when you submit feedback to [Feedback Assistant][2].

Tasks:

* [x] Remove support for linking with the WatchKit framework in mtouch.
* [x] Replace the WatchKit bindings with stubs that throw InvalidPlatformExceptions.

[1]: https://developer.apple.com/documentation/xcode_release_notes/xcode_11_beta_3_release_notes?language=objc
[2]: https://feedbackassistant.apple.com/

rolfbjarne:
* Copy all generated sources and modify them to throw PlatformNotSupported exceptions.
* Adjust some existing source code to also throw PlatformNotSupported exceptions.
* Sprinkle Obsolete attributes generously.
* Stop generating code for the WatchKit framework for iOS.
* Fix mtouch to not link with WatchKit, and show a warning if we detect code
  that want to use WatchKit.
* A few test fixes were also needed.

Fixes https://github.com/xamarin/xamarin-macios/issues/6492.


src/WatchKit/WKDefs.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace WatchKit {
 	[iOS (8,2)]
+	[Unavailable (PlatformName.iOS)]
```
spouliot: not sure how the IDE will interpret (and show) this mix of `[Introduced]` and `[Available]` ?!?

tools/common/Target.cs
```diff
@@ -205,9 +205,9 @@ public void GatherFrameworks ()
 							}
 							break;
 						case "WatchKit":
-							// Xcode 11 beta 1 doesn't ship WatchKit for iOS
+							// Xcode 11 doesn't ship WatchKit for iOS
 							if (Driver.XcodeVersion.Major == 11 && App.Platform == ApplePlatform.iOS) {
-								ErrorHelper.Warning (99, "Not linking with WatchKit because Xcode 11 beta 1 does not ship with the WatchKit framework.");
+								ErrorHelper.Warning (5219, "Not linking with WatchKit because it has been removed from iOS.");
```
spouliot: does the condition still makes sense ? as we'll never link with WatchKit anymore (whatever Xcode version is being used)

rolfbjarne: It shouldn't ever be hit, but _shouldn't_ and _won't_ are quite far apart in the dictionary.

docs/website/mtouch-errors.md
```diff
@@ -2545,6 +2545,14 @@ If this is not the case, please file a [bug report](https://github.com/xamarin/x
 
 The name of an Objective-C protocol can't contain certain characters which means that the `Adopts` attribute on the corresponding managed class can't have the `ProtocolType` parameter containing them. Please refer to the provided error message and fix accordingly.
 
+### MT4178: The class '{class}' will not be registered because the WatchKit framework has been removed from iOS.
+
+Apple removed the WatchKit framework on iOS, so any code that uses it will not work anymore.
```
spouliot: That seems inaccurate as introspection did not complain. Only the headers (not the framework) was removed (at least so far)

src/WatchKit/WKDefs.cs
rolfbjarne: The [Unavailable] makes the generator not generate anything (for iOS, still generates for watchOS), so there's no mix in the final source code.

The source code used for iOS is the manual code elsewhere in this PR.

rolfbjarne:
This is how it looks like in the IDE:

Screen Shot 2019-07-15 at 19 37 00

rolfbjarne:
```
08:11:13 sh: /opt/X11/bin/cp: No such file or directory
08:11:13 make[1]: *** [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/builds/install/bcl/monotouch/Facades/System.Reflection.DispatchProxy.dll] Error 127
08:11:13 make[1]: *** Waiting for unfinished jobs....
```

rolfbjarne:
build

spouliot:
https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10714/API_20diff_20_28PR_20only_29/ shows some API are removed (not deprecated)

rolfbjarne:
@spouliot those are all API that were added with Xcode 11, so we don't need them to maintain backwards compatibility.

PR #6511, opened 2019-07-05
[identitylookup] Update for Xcode 11 beta 1

spouliot:
No change in beta 2 and beta 3

### DIFF
--- a/src/identitylookup.cs
+++ b/src/identitylookup.cs
@@ -5,6 +5,7 @@
 //	Alex Soto  <alexsoto@microsoft.com>
 //
 // Copyright 2017 Xamarin Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


src/identitylookup.cs
```diff
@@ -15,6 +16,8 @@
 namespace IdentityLookup {
 
 	[iOS (11,0)]
+	[Introduced (PlatformName.UIKitForMac, 13,0)]
```
whitneyschmidt: Do we use [Introduced] for external APIs and [platform (version,version)] for internal (?) APIs?

spouliot:
known issue: watchOS 3.2 introspection timeout -> xamarin/maccore#1835

PR #6512, opened 2019-07-05
[carplay] Update for Xcode11 beta 3

whitneyschmidt:
https://github.com/xamarin/maccore/issues/1822 - beta3 resolves this issue.

Should I add git commit -m "Fixes xamarin/xamarin-macios#1234" to close the issue when this code is merged?


src/carplay.cs
```diff
@@ -72,6 +71,14 @@ enum CPLimitableUserInterface : ulong {
 		Lists = 1uL << 1,
 	}
 
+	[NoWatch, NoTV, NoMac, iOS (13,0)]
+	[Flags]
+	[Native]
+	enum CPAmbientLightStatus : ulong {
+		Light = 1uL << 0,
+		Dark = 1uL << 1
```
mandel-macaque: Small thing, add ',' after the last enum value, will minimise the next diff.

mandel-macaque:
@whitneyschmidt if you add fixes, it will close the issue, which looks like it is resolved. That way github will close the issue once the PR is merged.

src/carplay.cs
```diff
@@ -686,9 +699,12 @@ interface ICPSessionConfigurationDelegate { }
 	[BaseType (typeof (NSObject))]
 	interface CPSessionConfigurationDelegate {
 
-		[Abstract]
```
dalexsoto: Add `[Abstract]` again, removing it causes a breaking change

src/carplay.cs
```diff
@@ -878,6 +891,50 @@ interface CPAlertTemplate {
...
 	}
-}
 
+	interface ICPTemplateApplicationSceneDelegate { }
+
+	[NoWatch, NoTV, NoMac, iOS (13,0)]
+	[Protocol, Model (AutoGeneratedName = true)]
+	[BaseType (typeof (NSObject))]
+	interface CPTemplateApplicationSceneDelegate : UISceneDelegate
+	{
```
dalexsoto: It should be moved to declaration line above

src/carplay.cs
```diff
@@ -878,6 +891,50 @@ interface CPAlertTemplate {
...
+
+		[Export ("templateApplicationScene:didSelectNavigationAlert:")]
+		void DidSelect (CPTemplateApplicationScene templateApplicationScene, CPNavigationAlert navigationAlert);
+
+		[Export ("templateApplicationScene:didSelectManeuver:")]
+		void DidSelect (CPTemplateApplicationScene templateApplicationScene, CPManeuver maneuver);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (13,0)]
+	[BaseType (typeof(UIScene))]
```
dalexsoto: `typeof(UIScene)` needs a space before opening parentheses

src/carplay.cs
```diff
@@ -878,6 +891,50 @@ interface CPAlertTemplate {
...
+
+		[Export ("interfaceController", ArgumentSemantic.Strong)]
+		CPInterfaceController InterfaceController { get; }
+
+		[Export ("carWindow", ArgumentSemantic.Strong)]
+		CPWindow CarWindow { get; }
+
+		[iOS (13, 0)]
+		[Field ("CPTemplateApplicationSceneSessionRoleApplication")]
+		NSString CPTemplateApplicationSceneSessionRoleApplication { get; }
```
dalexsoto: You need a better name here like `SessionRoleApplication`

src/carplay.cs
```diff
@@ -28,7 +28,6 @@ enum CPAlertActionStyle : ulong {
 		Cancel,
 		Destructive,
 	}
-
```
dalexsoto: Remove this change, it causes diff noise 🙂

src/carplay.cs
spouliot: If it's not `@required` you can add a comment like `// this is not @required since Xcode 11 but kept for API compatibility`

src/carplay.cs
```diff
@@ -806,11 +822,8 @@ interface CPWindow {
 		[Export ("mapButtonSafeAreaLayoutGuide")]
 		UILayoutGuide MapButtonSafeAreaLayoutGuide { get; }
 
-#if false
-// Apple is trying to use CPTemplateApplicationScene but it is not exposed. https://github.com/xamarin/maccore/issues/1822
```
spouliot: don't forget to update/close that issue (it can be done at the same time as the PR, no need to wait for it to merge since it won't change Apple's headers)

src/carplay.cs
spouliot: and, if that's the case, you'll likely need to update `iOS-CarPlay.ignore` too :)

whitneyschmidt: updated .cs and .ignore! :)

whitneyschmidt:
Totally messed several things up with the last 2 commits. Going through and fixing everything again right now...

src/carplay.cs
```diff
@@ -796,6 +815,51 @@ interface CPImageSet : NSSecureCoding {
...
+		[NullAllowed, Export ("delegate", ArgumentSemantic.Strong)]
+		NSObject WeakDelegate { get; set; }
+
+		[Export ("interfaceController", ArgumentSemantic.Strong)]
+		CPInterfaceController InterfaceController { get; }
+
+		[Export ("carWindow", ArgumentSemantic.Strong)]
+		CPWindow CarWindow { get; }
+
+		[iOS (13, 0)]
```
spouliot: minor (don't update the PR only for that, unless other changes are needed - it can be changed in future betas)

There's no need for an `[iOS (13,0)]` here since it's already on the type-level (so it's inherited, by defaults, on all members - unless they specify something else)

PR #6516, opened 2019-07-08
[apitest] Remove comment about fixed crasher.

rolfbjarne:
Apple fixed the crasher on their side, so the comment is obsolete now.

I'm leaving the workaround, since it's still sensible code: the test class is
using a NSTextView instance across all test methods, and the workaround leaves
the NSTextView instance like it was before this particular test.


rolfbjarne:
Test failures are unrelated

* introspection/watchOS 3.2: https://github.com/xamarin/maccore/issues/1835
* monotouch-test: https://github.com/xamarin/maccore/issues/1605

Issue #6387, opened 2019-06-20
System.ArgumentOutOfRangeException NsUrlSessionHandler


czuvich:
### Environment

Xamarin.iOS Application
iOS 12.3.1 and 11.4

### Problem

I am getting the following exception in our iOS mobile Xamarin app.

```
Application Specific Information:
*** Terminating app due to uncaught exception 'System.ArgumentOutOfRangeException', reason: 'System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.'

Xamarin Exception Stack:
Parameter name: index
  at System.Collections.Generic.List`1[T].get_Item (System.Int32 index) <0x100a1cf00 + 0x0004b> in <4db5e915761a454b810fa3913e31be34#17161b22946fba777b3431f3b3c47e37>:0 
  at Foundation.NSNotificationCenter.RemoveObserver (Foundation.NSObject observer) <0x1014ace20 + 0x00053> in <497181d9d26447828f8574819afeb3d5#17161b22946fba777b3431f3b3c47e37>:0 
  at System.Net.Http.NSUrlSessionHandler.RemoveInflightData (Foundation.NSUrlSessionTask task, System.Boolean cancel) <0x1015127dc + 0x0011f> in <0da8ced3e38d42039920e99e08ca9321#17161b22946fba777b3431f3b3c47e37>:0 
  at System.Net.Http.NSUrlSessionHandler+NSUrlSessionHandlerDelegate.DidCompleteWithError (Foundation.NSUrlSession session, Foundation.NSUrlSessionTask task, Foundation.NSError error) <0x101512ec4 + 0x000bf> in <0da8ced3e38d42039920e99e08ca9321#17161b22946fba777b3431f3b3c47e37>:0
```

This issue has been recently introduced to our latest release which upgraded Xamarin.iOS.  We see the issue when we are trying to make an HttpClient call within a UIApplication.BackgroundTask ```UIApplication.SharedApplication.BeginBackgroundTask```.  The exceptions seems to occur intermittently.

### Steps to Reproduce

I can't quite reproduce the issue yet; however, I am getting crash logs via HockeyApp.  I've also been able to identify the exact line of code in the Xamarin code that is throwing the exception.  Line 191 in https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSUrlSessionHandler.cs.

### Workaround

I'm probably just going to catch all exceptions during this web call.

Visual Studio Info
```
[...]

PR #6537, opened 2019-07-09
[Foundation] Remove a possible race condition when interacting with the notification centre.

mandel-macaque:
Although it looked that it was not possible, due to the fact that the
RemoveObserversFromList does lock on the list of observers, we have
threads that are stepping on each other when the list is cleaned up.
Adding a lock around the AddObserver and RemoveObserver will ensure that
the handler does not call the removal more than once with an object that
was already removed.

Fixes https://github.com/xamarin/xamarin-macios/issues/6387


mandel-macaque:
We should back port this to d16-3. @chamons I don't think we do have the opportunity to do it for d16-2, do we?

mandel-macaque:
@rolfbjarne the initial implementation considered that the NotificationCenter was thread safe, which should be. But here I'm worried about getting a NRE when doing the RemoveObserver with more than one thread.

mandel-macaque:
This is when we summon @spouliot for a tiebreaker ;)

rolfbjarne:
If you look at the code that calls `AddNotification`: https://github.com/xamarin/xamarin-macios/blob/0ceced29a9b314b8ff6d287b58f167ba2bc45272/src/Foundation/NSUrlSessionHandler.cs#L397-L401

You see that it's already called with a lock held (`inflightRequestsLock`).

Now if you look at the code that calls `RemoveNotification`: 

https://github.com/xamarin/xamarin-macios/blob/0ceced29a9b314b8ff6d287b58f167ba2bc45272/src/Foundation/NSUrlSessionHandler.cs#L210-L221

and https://github.com/xamarin/xamarin-macios/blob/0ceced29a9b314b8ff6d287b58f167ba2bc45272/src/Foundation/NSUrlSessionHandler.cs#L230-L236

you can see that the same lock is held in the first call to `RemoveNotification`, but not the second. One potential improvement would be to fix the second call to only occur when lock is held, even though I believe this is unnecessary, because if you're calling Dispose on one thread while doing things on another thread, you're already in deep trouble.

There's a second point to make: these locks do not protect against calling `NSNotificationCenter` from different `NSUrlSessionHandler`  instances simultaneously, only if it's the same instance. The fix in `NSNotificationCenter` fixes that scenario too.

spouliot:
I'm with @rolfbjarne on this. We should fix the existing code before adding another layer of locks.

Also without a test case we're not 100% sure we fully understand the issue, nor the fix. It should **not** be backported to `d16-2` unless we get a test case or a confirmation it happens (on d16-2, as it might not) and that the fix (in master) to avoid making the situation any worse.

mandel-macaque:
@spouliot @rolfbjarne the less locks the better. I think that the issue as rolf says is in the NotificationCenter and already fixed. The reported case was on 8.1 and looks like and old release. Will comment and will close the bug and PR.

mandel-macaque:
Reopening due to  https://github.com/xamarin/xamarin-macios/issues/6691

mandel-macaque:
All failures are network related BUT not related with the handler:  LinkAllRegressionTest.TrustUsingOldPolicy : System.Net.WebException : Error: ConnectFailure (Connection timed out)

rolfbjarne:
I agree with Sebastien, we should produce a package with just the fix in the second commit, give that to the #6691 reporter and see if it fixes the issue.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -181,16 +182,20 @@ public NSUrlSessionHandler (NSUrlSessionConfiguration configuration)
...
 
 		void RemoveNotification ()
 		{
-			if (notificationToken != null) {
-				NSNotificationCenter.DefaultCenter.RemoveObserver (notificationToken);
-				notificationToken = null;
-			}
+			lock (notificationTokenLock) {
+				if (notificationToken != null) {
+					NSNotificationCenter.DefaultCenter.RemoveObserver (notificationToken);
```
rolfbjarne: I think something like this would be better, to avoid calling RemoveObserver from inside the lock:

```csharp
NSObject localNotificationToken;
lock (notificationTokenLock) {
    localNotificationToken = notificationToken;
    notificationToken = null;
}
if (localNotificationToken != null)
    NSNotificationCenter.DefaultCenter.RemoveObserver (localNotificationToken);
```

mandel-macaque:
Failures are unrelated to the PR and have been issued here: https://github.com/xamarin/maccore/issues/1939

mandel-macaque:

PR #6538, opened 2019-07-09
[foundation] Partial update for Xcode 11 beta 1 to 3

spouliot:
Some generic API require more work (manual bindings or generator
enhancements) before being included. xtro files are up to date.

### DIFF
--- a/src/Foundation/Enum.cs
+++ b/src/Foundation/Enum.cs
@@ -376,6 +376,15 @@ namespace Foundation  {
 		CloudSharingErrorMinimum = 5120,
 		[Mac (10,12)][iOS (10,0)][NoTV][NoWatch]
 		CloudSharingErrorMaximum = 5375,
+
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		CompressionFailedError = 5376,
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		DecompressionFailedError = 5377,
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		CompressionErrorMinimum = 5376,
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		CompressionErrorMaximum = 5503,
 	}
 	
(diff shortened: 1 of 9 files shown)


src/foundation.cs
```diff
@@ -15646,4 +16027,346 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+
+		[Export ("locale", ArgumentSemantic.Copy)]
+		NSLocale Locale { get; set; }
+
+		[NullAllowed, Export ("itemFormatter", ArgumentSemantic.Copy)]
+		NSFormatter ItemFormatter { get; set; }
+
+		[Static]
+		[Export ("localizedStringByJoiningStrings:")]
+		// using `NSString[]` since they might be one (of many) `NSString` subclass that hanldle localization
```
rolfbjarne: nitpick:

```suggestion
		// using `NSString[]` since they might be one (of many) `NSString` subclasses that handle localization
```

src/foundation.cs
```diff
@@ -15646,4 +16027,346 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		[Static]
+		[Export ("bits", ArgumentSemantic.Copy)]
+		NSUnitInformationStorage Bits { get; }
+
+		[Static]
+		[Export ("nibbles", ArgumentSemantic.Copy)]
+		NSUnitInformationStorage Nibbles { get; }
+
+		[Static]
+		[Export ("yottabytes", ArgumentSemantic.Copy)]
```
rolfbjarne: Can't say Apple doesn't aim high...

spouliot: I can't wait for that new iPhone :)

src/Foundation/Enum.cs
```diff
@@ -376,6 +376,15 @@ public enum NSCocoaError : int {
 		CloudSharingErrorMinimum = 5120,
 		[Mac (10,12)][iOS (10,0)][NoTV][NoWatch]
 		CloudSharingErrorMaximum = 5375,
+
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
```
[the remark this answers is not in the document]
spouliot: you're gith, I followed the previous members but it's not helpful (and extra metadata baggage)
I'll remove (all of) them in a future foundation PR :)

PR #6540, opened 2019-07-10
Martin tests amazing things

baulig:


baulig:
This is based on the current 2019-06 integration, with https://github.com/mono/mono/pull/15642 pushed on top of it.

PR #6546, opened 2019-07-10
[mmptest] Port link-uithread-1/2 to an NUnit LinkAll test. Partial fix for #4975.

rolfbjarne:
Fixes part of https://github.com/xamarin/xamarin-macios/issues/4975.


tests/linker/mac/link all/LinkAllTest.cs
```diff
@@ -25,5 +29,41 @@ public void Calendars ()
...
+			// works on main/ui thread
+			NSApplication.EnsureUIThread ();
+
+			ThreadPool.QueueUserWorkItem ((v) => Tester.Test ());
+			Assert.IsTrue (Tester.mre.WaitOne (TimeSpan.FromSeconds (10)), "Successful wait");
+#if DEBUG
+			var expected_ex_thrown = true;
+#else
+			var expected_ex_thrown = false;
+#endif
```
mandel-macaque: Approving, but @rolfbjarne could we have a comment of why is this added? I'm sure we will forget

rolfbjarne: It's because the UI thread check exception only occur in debug builds; in release builds the linker removes the check.

PR #6553, opened 2019-07-11
[UIKit] Partial update to Xcode 11 Beta 1, 2 and 3 - Part 2 of ?

dalexsoto:
Missing bindings for

* NSDiffableDataSourceSnapshot
* UICollectionViewDiffableDataSource
* UITableViewDiffableDataSource

due to a registrar issue with generics.

intro tested in the following versions locally (green)

- iOS 9.0 and iOS 13.0
- tvOS 9.0 and tvOS 13.0

There are some availability attributes older than 13 here. Also pardon some spaces v.s. tabs changes, those are intentional when I introduced something either on my last PR or I modified something in this PR.

### Type Changed: UIKit.UIAccessibilityElement

Modified base type:

	Foundation.NSObject UIKit.UIResponder

> New BaseType `NSObject` -> `UIResponder`

### Type Changed: UIKit.UIDragPreviewParameters

Modified base type:

	Foundation.NSObject UIKit.UIPreviewParameters

> New BaseType `NSObject` -> `UIPreviewParameters`

Removed properties:

	public virtual UIColor BackgroundColor { get; set; }
	public virtual UIBezierPath VisiblePath { get; set; }

> These now come from the new parent class `UIPreviewParameters`

### Type Changed: UIKit.UIDragPreviewTarget

Modified base type:

	Foundation.NSObject UIKit.UIPreviewTarget

> New BaseType `NSObject` -> `UIPreviewTarget`

Removed properties:

	public virtual CoreGraphics.CGPoint Center { get; }
	public virtual UIView Container { get; }
	public virtual CoreGraphics.CGAffineTransform Transform { get; }

> These now come from the new parent class `UIPreviewTarget`

### Type Changed: UIKit.UIKeyCommand
Modified base type:

	Foundation.NSObject UIKit.UICommand

> New BaseType `NSObject` -> `UICommand`

Removed method:

	public virtual void EncodeTo (Foundation.NSCoder encoder);

> No longer adopts `NSSecureCoding` protocol

### Type Changed: UIKit.UIMenu
Removed method:

	public virtual void EncodeTo (Foundation.NSCoder encoder);

> No longer adopts `NSSecureCoding` protocol

### Type Changed: UIKit.UITargetedDragPreview
Modified base type:

	Foundation.NSObject UIKit.UITargetedPreview

> New BaseType `NSObject` -> `UITargetedPreview`


src/uikit.cs
```diff
@@ -13970,6 +14520,34 @@ interface UIView : UIAppearance, UIAppearanceContainer, UIAccessibility, UIDynam
 		[TV (13,0), NoWatch, iOS (13,0)]
 		[Export ("overrideUserInterfaceStyle", ArgumentSemantic.Assign)]
 		UIUserInterfaceStyle OverrideUserInterfaceStyle { get; set; }
+
+		[NoWatch, TV (12,0), iOS (12,0)] // Yes, 12 and intro is happy on 12 device.
+		[Export ("transform3D", ArgumentSemantic.Assign)]
+		CATransform3D Transform3D { get; set; }
+
+#if !TVOS // avoid a CS0109 also these are not in TVOS anyways...
```
rolfbjarne: This might be more self-explanatory:

```csharp
#if TVOS
#pragma warning disable 0109 // The member 'member' does not hide an inherited member. The new keyword is not required
#endif
...
#if TVOS
#pragma warning restore 0109
#endif

File: src/uikit.cs [hunk 2]
```diff
@@ -14301,10 +14884,16 @@ interface UIViewController : NSCoding, UIAppearanceContainer, UIContentContainer
...
 
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'CanPerformUnwind2' instead.")]
+		[Deprecated (PlatformName.TvOS, 13, 0, message: "Use 'CanPerformUnwind2' instead.")]
 		[iOS (6,0)]
 		[Export ("canPerformUnwindSegueAction:fromViewController:withSender:")]
 		bool CanPerformUnwind (Selector segueAction, UIViewController fromViewController, NSObject sender);
 
+		[TV (13,0), iOS (13,0)]
+		[Export ("canPerformUnwindSegueAction:fromViewController:sender:")]
+		bool CanPerformUnwind2 (Selector segueAction, UIViewController fromViewController, [NullAllowed] NSObject sender);
```
rolfbjarne: So they added a new function that ends up with the same signature for us 😒 

The `2` suffix makes it sound like we did something wrong... but no other good option comes to mind either. Maybe `CanPerformUnwindAction`? `CanPerformSegueUnwind`? `CanPerformUnwindSegueAction`?

Also maybe add some `#if XAMCORE_4_0` spaghetti to rename the old function `CanPerformUnwindDeprecated` and the new one `CanPerformUnwind`.

src/uikit.cs
```diff
@@ -18848,13 +19572,1736 @@ interface UIPencilInteractionDelegate {
...
+	[DisableDefaultCtor]
+	interface NSCollectionLayoutItem : NSCopying {
+
+		[Static]
+		[Export ("itemWithLayoutSize:")]
+		NSCollectionLayoutItem Create (NSCollectionLayoutSize layoutSize);
+
+		[Static]
+		[Export ("itemWithLayoutSize:supplementaryItems:")]
+		NSCollectionLayoutItem Create (NSCollectionLayoutSize layoutSize, NSCollectionLayoutSupplementaryItem [] supplementaryItems);
```
rolfbjarne:
```suggestion
		NSCollectionLayoutItem Create (NSCollectionLayoutSize layoutSize, params NSCollectionLayoutSupplementaryItem [] supplementaryItems);
```

src/uikit.cs
```diff
@@ -18848,13 +19572,1736 @@ interface UIPencilInteractionDelegate {
...
+	[DisableDefaultCtor]
+	interface NSCollectionLayoutGroup : NSCopying {
+
+		[Static]
+		[Export ("horizontalGroupWithLayoutSize:subitem:count:")]
+		NSCollectionLayoutGroup CreateHorizontal (NSCollectionLayoutSize layoutSize, NSCollectionLayoutItem subitem, nint count);
+
+		[Static]
+		[Export ("horizontalGroupWithLayoutSize:subitems:")]
+		NSCollectionLayoutGroup CreateHorizontal (NSCollectionLayoutSize layoutSize, NSCollectionLayoutItem [] subitems);
```
rolfbjarne:
```suggestion
		NSCollectionLayoutGroup CreateHorizontal (NSCollectionLayoutSize layoutSize, params NSCollectionLayoutItem [] subitems);
```

src/uikit.cs
```diff
@@ -18848,13 +19572,1736 @@ interface UIPencilInteractionDelegate {
...
+
+		[Static]
+		[Export ("verticalGroupWithLayoutSize:subitem:count:")]
+		NSCollectionLayoutGroup CreateVertical (NSCollectionLayoutSize layoutSize, NSCollectionLayoutItem subitem, nint count);
+
+		[Static]
+		[Export ("verticalGroupWithLayoutSize:subitems:")]
+		NSCollectionLayoutGroup CreateVertical (NSCollectionLayoutSize layoutSize, NSCollectionLayoutItem [] subitems);
```
rolfbjarne:
```suggestion
		NSCollectionLayoutGroup CreateVertical (NSCollectionLayoutSize layoutSize, params NSCollectionLayoutItem [] subitems);
```

src/uikit.cs
```diff
@@ -281,29 +332,6 @@ interface NSAttributedString_NSAttributedStringKitAdditions {
 		bool ContainsAttachments (NSRange range);
 	}
 
-	[NoWatch]
-	[iOS (9,0)]
-	[BaseType (typeof(NSObject))]
-	[DisableDefaultCtor] // NSInvalidArgumentException Reason: -[NSDataAsset init]: unrecognized selector sent to instance 0x7f6c8cc0
-	interface NSDataAsset : NSCopying
```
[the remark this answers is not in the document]
dalexsoto: Yes, to move it away from a `#if !WATCH`

src/uikit.cs
```diff
@@ -5379,6 +5832,12 @@ interface UIFont : NSCopying {
 		[Internal] // bug 25511
 		IntPtr _FromName (string name, nfloat size);
 
+		[iOS (13,0), TV (13,0)]
+		[Static]
+		[Internal] // bug 25511
```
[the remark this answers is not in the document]
dalexsoto: Just c/p the one above but I will look for it. IIRC it was the one where the life of the UIFont mattered because apple was reusing it under the hood, something around those lines.

src/uikit.cs
dalexsoto: I think I like `CanPerformUnwindSegueAction` better

dalexsoto:
Thank you @rolfbjarne I addressed your comments :)

src/UIKit/UIEnums.cs
```diff
@@ -1125,9 +1137,14 @@ public enum UITextDirection : long {
...
 	
 	// NSInteger -> UITextInput.h
+	// NSWritingDirection == UITextWritingDirection
 	[Native]
 	[NoWatch]
+#if XAMCORE_4_0
+	public enum NSWritingDirection : long {
+#else
 	public enum UITextWritingDirection : long {
+#endif
```
spouliot: This is becoming confusing
```
src/AppKit/Enums.cs:    public enum NSWritingDirection : long {
src/Foundation/Enum.cs: [Obsolete ("Use NSWritingDirection in AppKit instead.")]
src/Foundation/Enum.cs: public enum NSWritingDirection : long {
```
and, for us, it's really different types (if some new, shared macOS+iOS API, start using them)

src/UIKit/UIEnums.cs
```diff
@@ -2318,16 +2368,14 @@ public enum UIImageSymbolWeight : long
 
 	[Mac (10,15, onlyOn64: true), iOS (13,0), TV (13,0)]
 	[Native]
-	public enum NSTextScalingType : long
-	{
+	public enum NSTextScalingType : long {
```
spouliot: try not to fix style in the same commit that adds code
ease reviewing (now) and blaming (later)

src/uikit.cs
```diff
@@ -328,25 +356,47 @@ partial interface NSFileProviderExtension {
...
+		[Export ("itemChangedAtURL:")]
+		void ItemChangedAtUrl (NSUrl url);
 	
-	    [Export ("stopProvidingItemAtURL:")]
-	    void StopProvidingItemAtUrl (NSUrl url);
+		[Deprecated (PlatformName.iOS, 13, 0)] // Undocumented replacement
+		[Export ("stopProvidingItemAtURL:")]
+		void StopProvidingItemAtUrl (NSUrl url);
+
+		// TODO: Enable with new iOS 13 FileProvider bindings.
```
spouliot: better file an issue (or add it to the xcode11 checklist) or we might forget about it...

src/uikit.cs
```diff
@@ -6265,14 +6759,15 @@ interface UIKeyCommand : NSCopying, NSSecureCoding {
 		NSString Escape { get; }
 
 		[iOS (9,0)]
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'UIKeyCommand' instead.")]
```
spouliot: this is `UIKeyCommand` ?!?

src/uikit.cs
```diff
@@ -6265,14 +6759,15 @@ interface UIKeyCommand : NSCopying, NSSecureCoding {
...
 		[Static]
 		[Export ("keyCommandWithInput:modifierFlags:action:discoverabilityTitle:")]
 		UIKeyCommand Create (NSString keyCommandInput, UIKeyModifierFlags modifierFlags, Selector action, NSString discoverabilityTitle);
 
 		[iOS (9,0)]
 		[NullAllowed]
 		[Export ("discoverabilityTitle")]
-		NSString DiscoverabilityTitle { get; set; }
+		NSString DiscoverabilityTitle { get; [Deprecated (PlatformName.iOS, 13, 0, message: "Use 'UIKeyCommand.DiscoverabilityTitle' instead.")] set; }
```
spouliot: This is `UIKeyCommand.DiscoverabilityTitle` ?!?
also nothing mention it's setter-only in the comment

src/uikit.cs
```diff
@@ -7752,6 +8017,19 @@ interface UITargetedPreview : NSCopying {
 	[BaseType (typeof (NSObject))]
 	[DesignatedDefaultCtor]
 	partial interface UIImageAsset : NSSecureCoding {
+
+		[iOS (13,0), TV (13,0)]
+		[Export ("imageWithConfiguration:")]
+		UIImage FromConfiguration (UIImageConfiguration configuration);
```
spouliot: did we not chose `Create*` last time ?

src/uikit.cs
```diff
@@ -10429,8 +10816,21 @@ interface UIScrollView : UIFocusItemScrollableContainer {
...
+		[Export ("horizontalScrollIndicatorInsets", ArgumentSemantic.Assign)]
+		UIEdgeInsets HorizontalScrollIndicatorInsets { get; set; }
+
 		[Export ("scrollIndicatorInsets")]
-		UIEdgeInsets ScrollIndicatorInsets { get; set; } 
+		UIEdgeInsets ScrollIndicatorInsets {
+			[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'VerticalScrollIndicatorInsets' or 'HorizontalScrollIndicatorInsets' instead.")]
+			[Deprecated (PlatformName.TvOS, 13, 0, message: "Use 'VerticalScrollIndicatorInsets' or 'HorizontalScrollIndicatorInsets' instead.")]
+			get;
+			set;
```
spouliot: weird, only the `getter` is deprecated ?

src/uikit.cs
```diff
@@ -11376,6 +11809,15 @@ interface UIStringAttributeKey {
 		[iOS (7,0)]
 		[Internal, Field ("NSDefaultTabIntervalDocumentAttribute")]
 		NSString NSDefaultTabIntervalDocumentAttribute { get; }
+
+		// we do not seem to expose other options like NSDefaultAttributesDocumentOption so keeping these as is for now
+		[iOS (13,0), TV (13,0), Watch (6,0)]
+		[Internal, Field ("NSTargetTextScalingDocumentOption")]
+		NSString NSTargetTextScalingDocumentOption { get; }
```
spouliot: drop `NS` prefix

src/uikit.cs
```diff
@@ -11376,6 +11809,15 @@ interface UIStringAttributeKey {
...
+
+		[iOS (13,0), TV (13,0), Watch (6,0)]
+		[Internal, Field ("NSSourceTextScalingDocumentOption")]
+		NSString NSSourceTextScalingDocumentOption { get; }
```
spouliot: drop `NS` prefix (minor since internal)

src/uikit.cs
```diff
@@ -18848,13 +19595,1736 @@ interface UIPencilInteractionDelegate {
...
+		UITabBarItemStateAppearance Selected { get; }
+
+		[Export ("disabled", ArgumentSemantic.Strong)]
+		UITabBarItemStateAppearance Disabled { get; }
+
+		[Export ("focused", ArgumentSemantic.Strong)]
+		UITabBarItemStateAppearance Focused { get; }
+	}
+
+	// TODO: The following break the registrar with:
```
spouliot: file (or link to) an issue - and include the stack trace there (not here)

spouliot:
api-diff complains about

* Type Changed: UIKit.UIMenu / Removed method: `public virtual void EncodeTo (Foundation.NSCoder encoder);`
* Type Changed: UIKit.UIKeyCommand / Removed method: `public virtual void EncodeTo (Foundation.NSCoder encoder);`

Those might be fine depending on the type hierarchy - in doubt add a test ;-)

* Removed Type UIKit.UIActionOptions
* Removed Type UIKit.UICommandState

Those are breaking changes

dalexsoto:
> Removed Type UIKit.UIActionOptions
> Removed Type UIKit.UICommandState
> Those are breaking changes

Yes but those were introduced in iOS 13 in my last PR.

src/UIKit/UIEnums.cs
dalexsoto: Yeah I agree but (🙃) I decided to fix this on the stuff that I included in my last PR it would not ease review but will still blame back to me

src/uikit.cs
dalexsoto: Xtro never forgets! This is not removed from Xtro todo files, do you still want an issue for this? I just created the bindings because I needed the name in one of the Deprecated attributes

src/uikit.cs
dalexsoto: c/p will indeed destroy the world some day 🙈

src/UIKit/UIEnums.cs
dalexsoto: Oh woow nice catch @spouliot will expose AppKit's into xkit.cs and use that instead 👍

src/uikit.cs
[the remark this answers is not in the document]
rolfbjarne: Surrounding this block with an inverted `#if` would have been easier to review I think:

```diff
+#endif // !WATCH
	interface NSDataAsset : NSCopying {
        // ...
        }
+#if !WATCH
```

It results in more conditional code, but I think that's fine in this case (since the amount of lines inside conditional directives decreases).

src/uikit.cs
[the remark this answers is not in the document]
rolfbjarne: OK, that made me look here: https://bugzilla.xamarin.com/show_bug.cgi?id=25511, which is the right one. I still think the full url should be in the comment :smile:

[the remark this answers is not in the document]
dalexsoto: I did change it, it just did not reflect here

```
			[iOS (13,0), TV (13,0), Watch (6,0)]
			[Static]
			[Internal] // bug https://xamarin.github.io/bugzilla-archives/25/25511/bug.html
			[Export ("monospacedSystemFontOfSize:weight:")]
			IntPtr _MonospacedSystemFontOfSize (nfloat fontSize, double weight);
```

src/UIKit/NSLayoutManager.cs
```diff
@@ -98,6 +107,24 @@ partial class NSLayoutManager {
 				}
 			}
 		}
+#endif // !XAMCORE_4_0
+
+		[Watch (6,0), TV (13,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
```
tj-devel709: Do you need to create this because it is not supported in C#? When do you know if you need to create an additional file?

src/UIKit/UIAccessibility.cs
```diff
@@ -339,6 +346,20 @@ static unsafe void TrampolineRequestGuidedAccessSession (IntPtr block, bool enab
 			}
 		}
 
+		[iOS (13,0), TV (13,0)]
+		[DllImport (Constants.UIKitLibrary)]
```
tj-devel709: Is Constants.UIKitLibrary a third party dll?

src/UIKit/UITypes.cs
```diff
@@ -201,13 +201,11 @@ public UIFloatRange (nfloat minimum, nfloat maximum)
...
-		static extern bool UIFloatRangeIsEqualToRange (UIFloatRange range, UIFloatRange otherRange);
+		// Got replaced by a macro in iOS 13 (Xcode 11)...
+		// [DllImport (Constants.UIKitLibrary)]
+		// static extern bool UIFloatRangeIsEqualToRange (UIFloatRange range, UIFloatRange otherRange);
 
-		public bool Equals (UIFloatRange other)
-		{
-			return UIFloatRangeIsEqualToRange (this, other);
-		}
+		public bool Equals (UIFloatRange other) => this.Minimum == other.Minimum && this.Maximum == other.Maximum;
```
stephen-hawley: This is absolutely bizarre because the documentation says that it uses an epsilon for the comparison. The swift equivalent does exactly this. Comparing floating point numbers for equality is generally a Very Bad Idea.

dalexsoto: So bad that it breaks our tests...

src/UIKit/NSLayoutManager.cs
dalexsoto: I need to create this because the **generator** does not know how to handle certain types and these need to be manually bound, meaning C# does support this it is just the generator does not know how to output the right C# code for these.

Generally the generator will yell at you when it does not know how to handle a type but basically C arrays and types that do not inherit from NSObject.

src/UIKit/UIAccessibility.cs
dalexsoto: Nope, `UIKitLibrary` points to the system's path where this library is located, the value changes between simulator and devices.

src/UIKit/UITypes.cs
spouliot: -> feedback

src/UIKit/UIEnums.cs
dalexsoto: Fixed in https://github.com/xamarin/xamarin-macios/pull/6553/commits/692b68bae277a1f1474d28e5d4535b39c6122704

src/UIKit/UITypes.cs
dalexsoto: Feedback sent, will add to our tracker. https://feedbackassistant.apple.com/feedback/6629947

src/uikit.cs
dalexsoto: Unfortunately there are other `From` members and IIRC we agreed that create would be for new types to be consistent 

```
[Export ("imageWithTraitCollection:")]
UIImage FromTraitCollection (UITraitCollection traitCollection);
```

src/uikit.cs
dalexsoto: Yup

```
@property(nonatomic)         UIEdgeInsets                 scrollIndicatorInsets; // use the setter only, as a convenience for setting both verticalScrollIndicatorInsets and horizontalScrollIndicatorInsets to the same value. if those properties have been set to different values, the return value of this getter (deprecated) is undefined.
- (UIEdgeInsets)scrollIndicatorInsets API_DEPRECATED("The scrollIndicatorInsets getter is deprecated, use the verticalScrollIndicatorInsets and horizontalScrollIndicatorInsets getters instead.", ios(2.0, 13.0), tvos(9.0, 13.0));
```

src/uikit.cs
dalexsoto: Linked: https://github.com/xamarin/xamarin-macios/issues/6567

dalexsoto:
@spouliot I think I have answered and/or fixed all your feedback, thanks!! 👍

dalexsoto:
build

dalexsoto:
@spouliot corrected the enum situation as discussed and created an issue for it https://github.com/xamarin/xamarin-macios/issues/6573, thanks for the feedback!

src/uikit.cs
```diff
@@ -18848,13 +19595,1736 @@ interface UIPencilInteractionDelegate {
...
+
+		[BindAs (typeof (UIWindowSceneSessionRole))]
+		[Export ("role")]
+		NSString Role { get; }
+
+		[Advice ("You can use 'SceneType' with a 'Type' instead.")]
+		[NullAllowed, Export ("sceneClass", ArgumentSemantic.Assign)]
+		Class SceneClass { get; set; }
+
+		Type SceneType {
```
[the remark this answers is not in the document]
dalexsoto: @tj-devel709 `SceneClass` is type of `Class` which is an Objective-C type that is not as .NET friendly as its C# counterpart `Type`. I created a nicer wrapper around it called `SceneType` you can set/get a C# `Type` and this property will convert it into an Objective-C `Class` and set/get it to  `SceneClass`. This allows you to do something like `SceneType = typeof (NSObject)` which is a little cleaner code that looking for a `Class` manually.

```csharp
[Advice ("You can use 'SceneType' with a 'Type' instead.")]
[NullAllowed, Export ("sceneClass", ArgumentSemantic.Assign)]
Class SceneClass { get; set; }

Type SceneType {
	[Wrap ("Class.Lookup (SceneClass)")] get;
	[Wrap ("SceneClass = value == null ? null : new Class (value)")] set;
}
```

PR #6559, opened 2019-07-11
[xcode11] Fix monotouch-test to not crash on device.

rolfbjarne:
* Add permission description for CoreBluetooth. Fixes xamarin/maccore#1880.
* Don't check for the macOS system version when running on other platforms.
* Don't try to set image captions on read-only files.

### DIFF
--- a/tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
+++ b/tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
@@ -63,17 +63,20 @@ namespace MonoTouchFixtures.MediaAccessibility {
 
 			var temp = String.Empty;
 			using (NSUrl url = new NSUrl (NSBundle.MainBundle.ResourceUrl.AbsoluteString + "basn3p08.png")) {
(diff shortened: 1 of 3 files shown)


tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
```diff
@@ -63,17 +63,20 @@ public void SetCaption ()
 
 			var temp = String.Empty;
 			using (NSUrl url = new NSUrl (NSBundle.MainBundle.ResourceUrl.AbsoluteString + "basn3p08.png")) {
-				Assert.True (MAImageCaptioning.SetCaption (url, "xamarin", out var e), "Set");
-				Assert.Null (e, "ro / set / no error"); // weird, it can't be saved back to the file metadata
+				// In Xcode 11 beta 3, the native P/Invoke will very helpfully crash if called with a read-only file path.
```
rolfbjarne: I've added this to https://github.com/xamarin/xamarin-macios/issues/6212.

Issue #6561, opened 2019-07-11
[xcode11] b1 Contacts.cs non-generic type errors for enumeratorForContactFetchRequest, enumeratorForChangeHistoryFetchRequest 


whitneyschmidt:
### Steps to Reproduce

1. Open src/Contacts.cs

2. Inside of `interface CNContactStore` uncomment or add:

		[Unavailable (PlatformName.Swift)]
		[Export ("enumeratorForContactFetchRequest:error:")]
		[return: NullAllowed]
		CNFetchResult<NSEnumerator<CNContact>> GetEnumeratorForContact (CNContactFetchRequest request, [NullAllowed] out NSError error);

		[Unavailable (PlatformName.Swift)]
		[Export ("enumeratorForChangeHistoryFetchRequest:error:")]
		[return: NullAllowed]
		CNFetchResult<NSEnumerator<CNChangeHistoryEvent>> GetEnumeratorForChangeHistory (CNChangeHistoryFetchRequest request, [NullAllowed] out NSError error);

3. make all -j8 yields the following errors:

```
GEN      [ios] generated_sources
GEN      [mac/mobile] generated-sources
GEN      [mac/full] generated-sources
GEN      [watch] generated_sources
CSC      [tvos] Xamarin.TVOS.dll
contacts.cs(1477,3): error CS0308: The non-generic type 'CNFetchResult' cannot be used with type arguments
contacts.cs(1477,17): error CS0308: The non-generic type 'NSEnumerator' cannot be used with type arguments
contacts.cs(1482,3): error CS0308: The non-generic type 'CNFetchResult' cannot be used with type arguments
contacts.cs(1482,17): error CS0308: The non-generic type 'NSEnumerator' cannot be used with type arguments
contacts.cs(1474,30): error CS0117: 'PlatformName' does not contain a definition for 'Swift'
contacts.cs(1479,30): error CS0117: 'PlatformName' does not contain a definition for 'Swift'
bgen: API binding contains errors.
make: *** [build/watch/watch/generated_sources] Error 1
make: *** Waiting for unfinished jobs....
contacts.cs(1477,3): error CS0308: The non-generic type 'CNFetchResult' cannot be used with type arguments
contacts.cs(1477,17): error CS0308: The non-generic type 'NSEnumerator' cannot be used with type arguments
contacts.cs(1482,3): error CS0308: The non-generic type 'CNFetchResult' cannot be used with type arguments
[...]

PR #6563, opened 2019-07-11
[contacts] xcode11 beta1 - beta4 updated bindings

whitneyschmidt:
Apple issue filed for typo: https://feedbackassistant.apple.com/feedback/6623415
Generics issues filed here: https://github.com/xamarin/xamarin-macios/issues/6561#issue-467045824

Generics code:

```
 [Unavailable (PlatformName.Swift)]
 [Export ("enumeratorForContactFetchRequest:error:")]
 [return: NullAllowed]
 CNFetchResult<NSEnumerator<CNContact>> GetEnumeratorForContact (CNContactFetchRequest request, [NullAllowed] out NSError error);

 [Unavailable (PlatformName.Swift)]
 [Export ("enumeratorForChangeHistoryFetchRequest:error:")]
 [return: NullAllowed]
 CNFetchResult<NSEnumerator<CNChangeHistoryEvent>> GetEnumeratorForChangeHistory (CNChangeHistoryFetchRequest request, [NullAllowed] out NSError error);
```


src/contacts.cs
```diff
@@ -463,16 +1457,30 @@ interface CNContactStore {
 
 		[Export ("unifiedContactsMatchingPredicate:keysToFetch:error:")]
 		[Protected] // we cannot use ICNKeyDescriptor as Apple (and others) can adopt it from categories
-		CNContact[] GetUnifiedContacts (NSPredicate predicate, NSArray keys, out NSError error);
+		[return: NullAllowed]
+		CNContact[] GetUnifiedContacts (NSPredicate predicate, NSArray keys, [NullAllowed] out NSError error);
```
spouliot: Sharpie generate them like this `[NullAllowed] out ...` but, in _C#_ reality, we can't use `null` on an `out` parameter.
Not bad (and not worth changing)

`[return: NullAllowed]` is also not used (by the generator) but it could become useful soon (with C# 8) so it's nice too add them when missing :)

src/contacts.cs
```diff
@@ -449,6 +1237,212 @@ interface CNLabelContactRelationKey {
...
+		[Export ("subgroup", ArgumentSemantic.Strong)]
+		CNGroup Subgroup { get; }
+
+		[Export ("group", ArgumentSemantic.Strong)]
+		CNGroup Group { get; }
+	}
+
+	[Watch (6,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+	[BaseType (typeof (NSObject))]
+	interface CNFetchRequest {}
```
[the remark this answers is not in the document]
whitneyschmidt: Added `[DisableDefaultCtor]` :)

src/contacts.cs
whitneyschmidt: Thanks for the info! :)

src/contacts.cs
```diff
@@ -441,6 +441,794 @@ interface CNLabelContactRelationKey {
...
+		[Field ("CNLabelContactRelationCousinOrSiblingsChild")]
+		NSString CousinOrSiblingsChild { get; }
+
+		[iOS (13, 0)]
+		[Field ("CNLabelContactRelationNieceOrCousin")]
+		NSString NieceOrCousin { get; }
+
+		[iOS (13, 0)]
+		[Field ("CNLabelContactRelationNephewOrCousin")]
+		NSString CNephewOrCousin { get; }
```
dalexsoto: `CNephewOrCousin` => `NephewOrCousin`

src/contacts.cs
```diff
@@ -449,6 +1237,213 @@ interface CNLabelContactRelationKey {
...
+	[BaseType (typeof (CNFetchRequest))]
+	interface CNChangeHistoryFetchRequest : NSSecureCoding
+	{
+		[NullAllowed, Export ("startingToken", ArgumentSemantic.Copy)]
+		NSData StartingToken { get; set; }
+
+		[NullAllowed, Export ("additionalContactKeyDescriptors", ArgumentSemantic.Copy)]
+		// we cannot use ICNKeyDescriptor as Apple (and others) can adopt it from categories
+		// cannot be exposed as NSString since they could be internalized types, like CNAggregateKeyDescriptor
+		NSArray AdditionalContactKeyDescriptors { get; set; }
```
dalexsoto: What about `NSObject []`? at least more C# friendly

src/contacts.cs
```diff
@@ -463,16 +1458,30 @@ interface CNContactStore {
 
 		[Export ("unifiedContactsMatchingPredicate:keysToFetch:error:")]
 		[Protected] // we cannot use ICNKeyDescriptor as Apple (and others) can adopt it from categories
-		CNContact[] GetUnifiedContacts (NSPredicate predicate, NSArray keys, out NSError error);
+		[return: NullAllowed]
+		CNContact[] GetUnifiedContacts (NSPredicate predicate, NSArray keys, [NullAllowed] out NSError error);
```
dalexsoto: What about `NSObject [] keys`? Nicer C# also for the ones below, @spouliot what do you think?

src/contacts.cs
rolfbjarne: I don't quite understand why you can't use `ICNKeyDescriptor[]` - that's how it's defined in the header, so it should work.

src/contacts.cs
rolfbjarne: Same: why not `ICNKeyDescriptor[]`?

src/contacts.cs
spouliot: it's marked as resolved - but it's still has the `C` prefix
please don't resolve discussions before the PR is fixed (it's too easy to forget those items)

whitneyschmidt: `C` removed in PR 🙂

src/contacts.cs
whitneyschmidt: Updated to `ICNKeyDescriptor[]` here.

src/contacts.cs
whitneyschmidt: @dalexsoto and I looked at this one...it appears that a lot of the bindings were written with a custom wrapper `CNContact` for `ICNKeyDescriptor[]`. I am leaving the old bindings with the wrapper as updating them would be a breaking change, but for new bindings using `ICNKeyDescriptor[]`.

src/contacts.cs
```diff
@@ -441,6 +452,795 @@ interface CNLabelContactRelationKey {
...
+		[Field ("CNLabelContactRelationUncleFathersSistersHusband")]
+		NSString UncleFathersSistersHusband { get; }
+
+		[Watch (6,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		[Field ("CNLabelContactRelationUncleFathersYoungerSistersHusband")]
+		NSString UncleFathersYoungerSistersHusband { get; }
+
+		[Watch (6,0), Mac (10,15, onlyOn64: true), iOS (13,0)]
+		[Field ("CNLabelContactRelationUncleFathersElderSistersHusband")]
+		NSString UncleFathersElderSistersHusband { get; }
```
tj-devel709: These relations are insane! Everything looks good to me though! Perhaps remove the newline before the curly braces when starting interfaces though for style, but your call!

whitneyschmidt: Updated to remove extra lines :)

spouliot:
The 5 failing tests seems related to

* `CNLabelSchool` - likely returning `null` (Apple bug that needs feedback)
* `[FAIL] ContactFetchRequestTest.Ctor_ICNKeyDescriptorArray : System.PlatformNotSupportedException :` which fails on 32bits iOS (so the test are likely not checking if the latest Xcode SDK is used)

src/contacts.cs
```diff
@@ -289,8 +289,8 @@ interface CNContactKey {
 		NSString InstantMessageAddresses { get; }
 	}
 
-	[iOS (9,0), Mac (10,11)]
-	[BaseType (typeof (NSObject))]
+ 	[iOS (9,0), Mac (10,15), Watch (2,0)]
```
spouliot: The original `[iOS (9,0), Mac (10,11)]` does not have to change here, because
- the type was available in macOS 10.11
- `[Watch (2,0)]` is implied since the framework/bindings are present (and it's the minimum we support)

src/contacts.cs
```diff
@@ -399,36 +399,47 @@ interface CNContactRelation : NSCopying, NSSecureCoding, INSCopying, INSSecureCo
 	[EditorBrowsable (EditorBrowsableState.Advanced)]
 	interface CNLabelContactRelationKey {
 
+		[Watch (6,0)]
```
spouliot: those were already available in watchOS (2.0+)

```diff
-CONTACTS_EXTERN NSString * const CNLabelContactRelationFather             API_AVAILABLE(macos(10.11), ios(9.0));
+CONTACTS_EXTERN NSString * const CNLabelContactRelationFather                                                       NS_AVAILABLE(10_11, 9_0);
```

That's identical - just different macros (noise) and there's no mention of watchOS - so that can be confusing. It's often (nothing is ever 100% with Apple) because it's implied - i.e. the header/framework is in watchOS ? so all API that are not decorated are _presumed_ to have been available since the beginning (2.0 for watchOS).

Other ways to double check / reduce confusion is to check if they are already in `Xamarin.WatchOS.dll` (or read the generated code). If they are then the introspection tests will verify that the are present (at runtime).

src/contacts.cs
```diff
@@ -449,6 +1249,199 @@ interface CNLabelContactRelationKey {
...
+	interface CNChangeHistoryRemoveSubgroupFromGroupEvent {
+		[Export ("subgroup", ArgumentSemantic.Strong)]
+		CNGroup Subgroup { get; }
+
+		[Export ("group", ArgumentSemantic.Strong)]
+		CNGroup Group { get; }
+	}
+
+	// this type is new in Xcode11 but is decorated with earlier versions since it's used as a
+	// base type for older types (and that confuse the generator for 32bits availability)
```
spouliot: 👍

src/contacts.cs
whitneyschmidt: Got it. Removed `[Watch (6,0)]`. 👍

whitneyschmidt:
build

dalexsoto:
@whitneyschmidt the www.xamarin.com test failures are not your fault, but you still have some xtro failures

src/contacts.cs
```diff
@@ -441,6 +441,811 @@ interface CNLabelContactRelationKey {
 		[Watch (4,0)]
 		[Field ("CNLabelContactRelationDaughter")]
 		NSString Daughter { get; }
+
+		[Watch (6,0), Mac (10,15), iOS (13,0)]
+		[Field ("CNLabelContactRelationColleauge")] // Apple issue: https://feedbackassistant.apple.com/feedback/6623415
+		NSString Colleauge { get; }
```
dalexsoto: Any reason we can't make this right for us? `Colleauge` => `Colleague`?

whitneyschmidt:
build

whitneyschmidt:
build

PR #6564, opened 2019-07-11
Ensure Local TimeZone Id is Set For TimeZoneInfo.Local

steveisok:
Mirrors the runtime change found in https://github.com/mono/mono/pull/15667.

The fix here is to return the name of the local time zone so that TimeZoneInfo.Local
can have the correct Id.

### DIFF
--- a/runtime/xamarin-support.m
+++ b/runtime/xamarin-support.m
@@ -103,8 +103,6 @@ xamarin_timezone_get_names (int *count)
 
 //
 // Returns the geopolitical region ID of the local timezone.
-// Previously we just provided the data to TimeZoneInfo.MonoTouch.cs and that
-// defaulted the ID to "Local", which is incorrect. 
 
 const char *
 xamarin_timezone_get_local_name ()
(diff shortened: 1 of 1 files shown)


spouliot:
It looks like this code should be (have been) removed with the SDK work.

akoeplinger:
@spouliot we didn't share/move these runtime glue pieces as part of the SDKs work so this is still necessary.

spouliot:
duplicated files will cause issues sooner than later :(

anyway when was `xamarin_timezone_get_local_name` introduced (used) inside mono ?
does it make sense in `2019-02` (used in master) as-is ? or with a bump ?
or should it be re-based on the `2019-06` branch ? or even later

akoeplinger:
It was merged today in Mono and backported to 2019-06 so we should be good having it just in master here.

runtime/xamarin-support.m
```diff
@@ -96,6 +101,20 @@
...
+// Previously we just provided the data to TimeZoneInfo.MonoTouch.cs and that
+// defaulted the ID to "Local", which is incorrect. 
+
+const char *
+xamarin_timezone_get_local_name ()
+{
+	NSTimeZone *tz = nil;
+	tz = [NSTimeZone localTimeZone];
+	NSString *name = [tz name];
+	return strdup ([name UTF8String]);
```
rolfbjarne: This will crash if `name` is NULL (which should probably not happen, but I've seen strange data come out of these Apple API before, and they're a pain to debug if they only occur under certain circumstances).

rolfbjarne:
build

PR #6569, opened 2019-07-12
[mmptest] Various cleanups.

rolfbjarne:
* Simplify Makefile
	* Remove logic to build using vstool since it's not needed anymore.
	* Remove logic to build 32-bit test apps.
* tests.sln: add existing projects, remove deleted projects.
* Delete MainMenu.xib, it's not needed since none of these tests show a UI.
* Modify the csprojs to be much more similar to eachother, which makes it
  easier to find the significant difference between them. Also add
  properties/nodes that VSfM insists on adding, so that future diffs should
  not contain gratuitous changes by VSfM.


rolfbjarne:
Test failure is unrelated

* link all/Mac/Release: https://github.com/xamarin/maccore/issues/649

PR #6588, opened 2019-07-15
[quicklookUI] xcode11 beta1 - beta3 updated bindings

whitneyschmidt:


src/quicklookUI.cs
```diff
@@ -168,5 +168,9 @@ interface QLPreviewingController {
 #endif
 		[Export ("preparePreviewOfSearchableItemWithIdentifier:queryString:completionHandler:")]
 		void PreparePreviewOfSearchableItem (string identifier, string queryString, Action<NSError> ItemLoadingHandler);
+
+		[Mac (10,15, onlyOn64: true)]
+		[Export ("preparePreviewOfFileAtURL:completionHandler:")]
+		void PreparePreviewOfFile (NSUrl url, Action<NSError> ItemLoadingHandler);
```
whitneyschmidt: Not sure about naming for `Action<NSError> ItemLoadingHandler` - I went on the param naming of the binding directly above.

The documentation for `preparePreviewOfFileAtURL` says:

```
+ @param handler The completion handler should be called whenever the view is ready to be displayed. 
A loading spinner will be shown until the handler is called.
+ It can be called asynchronously after the method has returned.
```

spouliot: I would remove the `OfFile` suffix but Swift kept it... not a huge thing :-)

But the second parameter should be `completionHandler`, not `ItemLoadingHandler` (it looks like a _copy/paste_ error in previous API declaration) as it should match the ObjC parameter name and **not* start with a capital letter.

whitneyschmidt: `completionHandler` makes a lot more sense - thank you! New API is updated.

Leaving previous API as-is to avoid diff noise unless otherwise advised :)

spouliot: yeah, copy/paste is not always safe from past errors ;-)

src/quicklookUI.cs
```diff
@@ -168,5 +168,9 @@ interface QLPreviewingController {
 #endif
 		[Export ("preparePreviewOfSearchableItemWithIdentifier:queryString:completionHandler:")]
 		void PreparePreviewOfSearchableItem (string identifier, string queryString, Action<NSError> ItemLoadingHandler);
+
+		[Mac (10,15, onlyOn64: true)]
```
spouliot: In preparation of https://github.com/xamarin/xamarin-macios/pull/6595 please remove `, onlyOn64: true`

PR #6589, opened 2019-07-15
[GameController] Add Xcode 11 Beta 1-5 bindings

tj-devel709:


tj-devel709:
Here is a link to the issue: https://github.com/xamarin/xamarin-macios/issues/6590
When adding attributes to the GCAcceleration, GCRotationRate, and GCQuaternion structs, it will not compile, but compiles fine without the attributes

src/GameController/Enums.cs
```diff
@@ -12,6 +12,9 @@
 using Foundation;
 
 namespace GameController {
+	[Deprecated (PlatformName.MacOSX, 10, 15, message: "GCMicroGamepadSnapshot has been deprecated, use [GCController controllerWithMicroGamepad] instead")]
```
rolfbjarne: In deprecation messages that point elsewhere, we point to the managed name instead of the Objective-C name, since that's much more useful for our developers.

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -49,13 +49,22 @@ public struct GCExtendedGamepadSnapShotDataV100 {
 		// Buttons in the range [0.0, 1.0]
 		public float /* float_t = float */ LeftTrigger;
 		public float /* float_t = float */ RightTrigger;
+		[iOS (12,2), Mac (10,14,4, onlyOn64: true), TV (12,2)]
+		public bool supportsClickableThumbsticks;
```
rolfbjarne: Public API shouldn't start with a lower-cased letter: `SupportsClickableThumbsticks`.

Same for the to other fields just below.

src/GameController/GCMotion.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using ObjCRuntime;
+using Foundation;
+
+namespace GameController {
+
+	// Removed the Headers for these structs because I could not compile with them.
+	// [iOS (13,0), Mac (10,15, onlyOn64: true), TV (13,0)]
+	public struct GCAcceleration {
+
```
rolfbjarne: Not a big deal, but I think this looks better without a newline between each field.

src/gamecontroller.cs
```diff
@@ -115,10 +123,16 @@ partial interface GCControllerDirectionPad {
 
 		[Export ("right")]
 		GCControllerButtonInput Right { get; }
+
+		[Export ("setValueForXAxis:yAxis:")]
+		void SetValueForXAxis (float xAxis, float yAxis);
```
rolfbjarne: Just `SetValue` is enough here.

src/gamecontroller.cs
```diff
@@ -186,6 +203,9 @@ partial interface GCExtendedGamepad {
 		[Export ("valueChangedHandler", ArgumentSemantic.Copy)]
 		GCExtendedGamepadValueChangedHandler ValueChangedHandler { get; set; }
 
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "GCExtendedGamepadSnapshot has been deprecated, use [GCController capture] instead")]
+		[Deprecated (PlatformName.iOS, 13, 0, message: "GCExtendedGamepadSnapshot has been deprecated, use [GCController capture] instead")]
+		[Deprecated (PlatformName.TvOS, 13, 0, message: "GCExtendedGamepadSnapshot has been deprecated, use [GCController capture] instead")]
```
joconte: Hello, is this part of the message ```GCExtendedGamepadSnapshot has been deprecated``` worth having ? I thought we just needed ```use xxx instead```

spouliot: yes, we can't use the text from header files
As much as possible (for direct replacements) we try to stick to this format: `"Use 'XXX' instead."`, where `XXX` is the *managed* name of the API providing the replacement

src/gamecontroller.cs
spouliot: yes, an ObjC API with two parameters looks like `methodParam1:Param2:`
IOW there's **no** `:` between the method name and the first parameter, so
* method -> `SetValueFor`
* parameter 1 -> `xAxis`
* parameter 2 -> `yAxis`

and we remove suffix like `For` (and `With`...) so it becomes 
```csharp
void SetValue (float xAxis, float yAxis);
```

src/gamecontroller.cs
```diff
@@ -307,16 +353,38 @@ partial interface GCController {
...
+		[Export ("productCategory")]
+		string ProductCategory { get; }
+
+		[TV (13, 0), Mac (10, 15, onlyOn64: true), iOS (13, 0)]
+		[Export ("snapshot")]
+		bool Snapshot { [Bind ("isSnapshot")] get; }
+
+		[TV (13, 0), Mac (10, 15, onlyOn64: true), iOS (13, 0)]
+		[Export ("capture")]
+		GCController Capture { get; }
```
spouliot: That sounds like a method - and not a property
Please check headers and docs (both ObjC and Swift)

src/gamecontroller.cs
```diff
@@ -307,16 +353,38 @@ partial interface GCController {
...
+
+		[TV (13, 0), Mac (10, 15, onlyOn64: true), iOS (13, 0)]
+		[Static]
+		[Export ("controllerWithMicroGamepad")]
+		GCController ControllerWithMicroGamepad { get; }
```
spouliot: No verb in method
I think `GetMicroGamepadController` is better

src/gamecontroller.cs
```diff
@@ -307,16 +353,38 @@ partial interface GCController {
...
+
+		[TV (13, 0), Mac (10, 15, onlyOn64: true), iOS (13, 0)]
+		[Static]
+		[Export ("controllerWithExtendedGamepad")]
+		GCController ControllerWithExtendedGamepad { get; }
```
spouliot: No verb in method
I think `GetExtendedGamepadController ()` is better

src/gamecontroller.cs
tj-devel709: This one, ControllerWithMicroGamepad, and ControllerWithExtendedGamepad were all methods but sharpie suggested changing them to properties since they just return a GCController, should I leave them as methods?

src/gamecontroller.cs
tj-devel709: Right now it is set as a property though and properties don't need a verb correct? I can revert to a method though.

src/gamecontroller.cs
tj-devel709: Same as above, Right now it is set as a property though and properties don't need a verb correct? I can revert to a method though.

src/GameController/GCMotion.cs
```diff
@@ -0,0 +1,48 @@
...
+// Copyright 2019 Microsoft Corporation.
+
+using System;
+
+using ObjCRuntime;
+using Foundation;
+
+namespace GameController {
+
+	// Removed the Headers for these structs because I could not compile with them.
```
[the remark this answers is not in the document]
tj-devel709: Oops sorry, I need to get my vocab straight haha

src/gamecontroller.cs
tj-devel709: Good catch! I will fix these, thanks!

src/gamecontroller.cs
tj-devel709: Thanks, will change!

src/GameController/GCMotion.cs
tj-devel709: Changed!

src/GameController/Enums.cs
tj-devel709: Ah man looked right past that

src/GameController/GCExtendedGamepadSnapshot.cs
tj-devel709: Missed this one too, thanks!

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -49,13 +49,22 @@ public struct GCExtendedGamepadSnapShotDataV100 {
...
+		[iOS (12,2), Mac (10,14,4, onlyOn64: true), TV (12,2)]
+		public bool leftThumbstickButton;
+		[iOS (12,2), Mac (10,14,4, onlyOn64: true), TV (12,2)]
+		public bool rightThumbstickButton;
 
 		// radar: https://trello.com/c/7FoGTORD (GCExtendedGamepadSnapShotDataV100 struct size / alignment not backward compatible)
 		// [TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
 		// public bool LeftThumbstickButton;
 		// [TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
 		// public bool RightThumbstickButton;
```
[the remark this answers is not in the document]
tj-devel709: So far checking this, the link provided is dead. They also added the "SupportsClickableThumbsticks" which may have helped the functionality of those removed buttons. Where would you recommend checking if these should be here or not?

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -49,13 +49,22 @@ public struct GCExtendedGamepadSnapShotDataV100 {
 		// Buttons in the range [0.0, 1.0]
 		public float /* float_t = float */ LeftTrigger;
 		public float /* float_t = float */ RightTrigger;
+		[iOS (12,2), Mac (10,14,4, onlyOn64: true), TV (12,2)]
```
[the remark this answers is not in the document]
tj-devel709: Will do!

src/GameController/GCExtendedGamepadSnapshot.cs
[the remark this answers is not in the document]
spouliot: it's not dead - but you need to login (and have access) to trello to see it
Nothing much to see since Apple closed it as a duplicate. The original issue was that the structure was not backward compatible (since fields were added in the middle, not at end).

src/gamecontroller.cs
spouliot: yes, they are not state (but actions) so they should be methods

side note: in _old_ ObjC there was no syntax for property so sharpie does its best... but it's only a suggestion (and newer headers generally use `@property` so there's no/less need to guess properties).

src/gamecontroller.cs
spouliot: It should be a method, as this is not a really a state of the current type.

src/gamecontroller.cs
spouliot: yeah, I was not clear - it should be a method (and have a verb)

src/GameController/Enums.cs
```diff
@@ -12,9 +12,9 @@
 using Foundation;
 
 namespace GameController {
-	[Deprecated (PlatformName.MacOSX, 10, 15, message: "GCMicroGamepadSnapshot has been deprecated, use [GCController controllerWithMicroGamepad] instead")]
-	[Deprecated (PlatformName.iOS, 13, 0, message: "GCMicroGamepadSnapshot has been deprecated, use [GCController controllerWithMicroGamepad] instead")]
-	[Deprecated (PlatformName.TvOS, 13, 0, message: "GCMicroGamepadSnapshot has been deprecated, use [GCController controllerWithMicroGamepad] instead")]
+	[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use [GCController ControllerWithMicroGamepad] instead")]
+	[Deprecated (PlatformName.iOS, 13, 0, message: "Use [GCController ControllerWithMicroGamepad] instead")]
+	[Deprecated (PlatformName.TvOS, 13, 0, message: "Use [GCController ControllerWithMicroGamepad] instead")]
```
spouliot: That `[...]` is the ObjC syntax - please use the managed name, like
`Use 'GCController.GetMicroGamePadController()' instead.`

note the `.` at the end :)

src/GameController/GCGamepadSnapshot.cs
```diff
@@ -40,9 +40,9 @@ public struct GCGamepadSnapShotDataV100 {
 		public float /* float_t = float */ LeftShoulder;
 		public float /* float_t = float */ RightShoulder;
 
-		[Deprecated (PlatformName.MacOSX, 10, 15, message: "GCGamepad has been deprecated, use GCExtendedGamepad instead")]
-		[Deprecated (PlatformName.iOS, 13, 0, message: "GCGamepad has been deprecated, use GCExtendedGamepad instead")]
-		[Deprecated (PlatformName.TvOS, 13, 0, message: "GCGamepad has been deprecated, use GCExtendedGamepad instead")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use GCExtendedGamepad instead")]
```
spouliot: need `'` around the type name
and `.` at the end of the string

src/GameController/GCExtendedGamepadSnapshot.cs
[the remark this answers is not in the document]
tj-devel709: Ah okay, so should I remove the LeftThumbsitckButton and RightThumbsitckButton but leave in the SuportsClickableThumbsticks?

src/gamecontroller.cs
tj-devel709: Okay thanks!

src/gamecontroller.cs
tj-devel709: Changing!

src/gamecontroller.cs
tj-devel709: Changing!

src/GameController/Enums.cs
tj-devel709: Oh thank you

src/gamecontroller.cs
```diff
@@ -115,10 +125,17 @@ partial interface GCControllerDirectionPad {
...
+		[Mac (10,15), iOS (13,0)]
+		[Export ("setValueForXAxis:yAxis:")]
+		void SetValue (float xAxis, float yAxis);
 	}
 
 	delegate void GCGamepadValueChangedHandler (GCGamepad gamepad, GCControllerElement element);
 
+	[Deprecated (PlatformName.MacOSX, 10, 12, message: "Use GCExtendedGamepad instead.")]
+	[Deprecated (PlatformName.iOS, 10, 0, message: "Use GCExtendedGamepad instead.")]
+	[Deprecated (PlatformName.TvOS, 10, 0, message: "Use GCExtendedGamepad instead.")]
```
spouliot: missing single `'` around the type name

src/gamecontroller.cs
```diff
@@ -307,16 +356,38 @@ partial interface GCController {
...
+		bool Snapshot { [Bind ("isSnapshot")] get; }
+
+		[TV (13, 0), Mac (10, 15), iOS (13, 0)]
+		[Export ("capture")]
+		GCController Capture ();
+
+		[TV (13, 0), Mac (10, 15), iOS (13, 0)]
+		[Static]
+		[Export ("controllerWithMicroGamepad")]
+		GCController GetControllerWithMicroGamepad ();
```
spouliot: -> `GetMicroGamepadController` sounds better

src/gamecontroller.cs
```diff
@@ -307,16 +356,38 @@ partial interface GCController {
...
+
+		[TV (13, 0), Mac (10, 15), iOS (13, 0)]
+		[Static]
+		[Export ("controllerWithExtendedGamepad")]
+		GCController GetControllerWithExtendedGamepad ();
```
spouliot: -> `GetExtendedGamepadController` sounds better

src/gamecontroller.cs
```diff
@@ -405,6 +514,9 @@ interface GCMicroGamepadSnapshot {
 		[Export ("initWithController:snapshotData:")]
 		IntPtr Constructor (GCController controller, NSData data);
 
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use 'GCController.ControllerWithMicroGamepad()' instead.")]
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use 'GCController.ControllerWithMicroGamepad()' instead.")]
+		[Deprecated (PlatformName.TvOS, 13, 0, message: "Use 'GCControler.ControllerWithMicroGamepad()' instead.")]
```
spouliot: That's not the managed name you have given (missing `Get` prefix)
Please review all attributes (since it was copy/pasted in many places)

src/GameController/GCExtendedGamepadSnapshot.cs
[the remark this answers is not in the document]
spouliot: no, if the new fields are present (in the headers) at the end of the structure then we must match the memory representation (so we must have the elements in the same order)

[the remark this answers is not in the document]
tj-devel709: Sorry I am a little confused. In the xcode11 headers, it seems that it should be in the order of SupportsClickableThumbsticks
LeftThumbstickButton
RightThumbstickButton. 
So you are saying I should include these 3 and put them in this order?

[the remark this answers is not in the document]
spouliot: I have not looked at the headers... but, since it's a structure, they *must* match (same order, same size...) so that the code from both ObjC and C# can write to the (same) memory location (and make sense of what the other wrote)

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -49,13 +51,22 @@ public struct GCExtendedGamepadSnapShotDataV100 {
 		// Buttons in the range [0.0, 1.0]
 		public float /* float_t = float */ LeftTrigger;
 		public float /* float_t = float */ RightTrigger;
+
```
chamons: Unneeded white space

tj-devel709:
Okay this should be good now! @chamons @mandel-macaque @whitneyschmidt @spouliot @dalexsoto

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -50,12 +52,18 @@ public struct GCExtendedGamepadSnapShotDataV100 {
...
+		// Apple inserted these in the middle of the struct previously so it broke the struct for backwards compatibility,
+		// but now they are at the end of the struct so they are allowed and do not prevent backwards compatibility
+		[iOS (12,2), Mac (10,14,4), TV (12,2)]
+		public bool SupportsClickableThumbsticks;
+		[TV (12, 2), Mac (10, 14, 4), iOS (12, 2)]
+		public bool LeftThumbstickButton;
+		[TV (12, 2), Mac (10, 14, 4), iOS (12, 2)]
+		public bool RightThumbstickButton;
+
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use 'GCController.ControllerWithExtendedGamepad()' instead.")]
```
rolfbjarne: The managed name is `GCController. GetExtendedGamepadController ()`

src/GameController/GCExtendedGamepadSnapshot.cs
```diff
@@ -50,12 +52,18 @@ public struct GCExtendedGamepadSnapShotDataV100 {
...
-		// radar: https://trello.com/c/7FoGTORD (GCExtendedGamepadSnapShotDataV100 struct size / alignment not backward compatible)
-		// [TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
-		// public bool LeftThumbstickButton;
-		// [TV (12, 1), Mac (10, 14, 1, onlyOn64: true), iOS (12, 1)]
-		// public bool RightThumbstickButton;
-
+		// Apple inserted these in the middle of the struct previously so it broke the struct for backwards compatibility,
+		// but now they are at the end of the struct so they are allowed and do not prevent backwards compatibility
+		[iOS (12,2), Mac (10,14,4), TV (12,2)]
+		public bool SupportsClickableThumbsticks;
```
rolfbjarne: I looked at the header, and these fields are not in the `GCExtendedGamepadSnapShotDataV100` struct, they're only in the `GCExtendedGamepadSnapshotData` struct.

tj-devel709: Oh man big rookie mistake, thank you for catching this!

src/GameController/GCExtendedGamepadSnapshot.cs
tj-devel709: Thanks for catching this also!

mandel-macaque:
There is a failing tests that might affect this PR that is unrelated: https://github.com/xamarin/maccore/issues/1903

PR #6595, opened 2019-07-16
[src] Remove/obsolete 'onlyOn64: true' in attributes.

rolfbjarne:
It's not needed anymore, since we don't support anything but 64-bit for macOS.


dalexsoto:
With the help of Oleg and Marius looked at it and we think it should be safe to land.

PR #6596, opened 2019-07-16
[MediaPlayer] Add Xcode 11 Beta 1-3 bindings

tj-devel709:


tj-devel709:
Dependent on AVFoundation. Has AVMediaSelectionGroup and AVMediaSelectionOption set to NSObject for the watch currently. https://github.com/xamarin/xamarin-macios/issues/6597

src/mediaplayer.cs
```diff
@@ -1280,12 +1280,17 @@ interface MPVolumeView {
 		[Export ("initWithFrame:")]
 		IntPtr Constructor (CGRect frame);
 
+		[Deprecated (PlatformName.iOS, 13, 0, message: "Use AVRoutePickerView instead")]
```
joconte: Maybe add a ```.```at the end of the message for consistency.

tj-devel709: Good catch! Thanks :)

src/mediaplayer.cs
```diff
@@ -20,7 +20,7 @@
 using UIKit;
 #endif
 using System;
-#if WATCH
+#if WATCH // TODO Will be removed after CoreMedia and AVFoundation are updated
 using AVMediaSelectionGroup = Foundation.NSObject;
 using AVMediaSelectionOption = Foundation.NSObject;
 #endif
```
spouliot: ^ that creates unusable API and it's too easy to forget
Better not update `watchOS-MediaPlayer.todo` so it stays visible (and under our radar).

IOW just do the iOS parts until other parts are ready

tj-devel709: Okay I will do that!

PR #6600, opened 2019-07-17
[corelocation] Update for Xcode 11 beta 1 to 4

spouliot:


src/corelocation.cs
```diff
@@ -293,18 +285,24 @@ partial interface CLLocationManager {
...
+		[Deprecated (PlatformName.iOS, 13,0, message: "Not used anymore. Call will not have any effect.")]
+		[Unavailable (PlatformName.UIKitForMac)][Advice ("This API is not available when using UIKit on macOS.")]
 		[Export ("disallowDeferredLocationUpdates")]
 		void DisallowDeferredLocationUpdates ();
 #endif
 
 		[NoWatch][NoTV]
 		[iOS (6,0)]
 		[Mac (10,9)]
+		[Deprecated (PlatformName.iOS, 13,0, message: "Not used anymore. It will always return .false'.")]
```
tj-devel709: Just wondering if ".false" is a typo?

spouliot: yes it should have been `'false'`
good catch !

mandel-macaque:
We probably want to double check with the new Xcode 11 beta 4.

src/corelocation.cs
```diff
@@ -428,13 +446,25 @@ partial interface CLLocationManagerDelegate
...
 #if !MONOMAC
 		[NoWatch][NoTV]
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'DidRangeBeaconsSatisfyingConstraint' instead.")]
 		[iOS (7,0), Export ("locationManager:didRangeBeacons:inRegion:"), EventArgs ("CLRegionBeaconsRanged")]
 		void DidRangeBeacons (CLLocationManager manager, CLBeacon [] beacons, CLBeaconRegion region);
 
+		[NoWatch, NoTV, NoMac, iOS (13,0)]
+		[Export ("locationManager:didRangeBeacons:satisfyingConstraint:")]
+		[EventArgs ("CLRegionBeaconsConstraintRanged")]
+		void DidRangeBeaconsSatisfyingConstraint (CLLocationManager manager, CLBeacon[] beacons, CLBeaconIdentityConstraint beaconConstraint);
```
whitneyschmidt: Can `DidRangeBeaconsSatisfyingConstraint` be shortened to `DidRangeBeacons` - since the params are different it won't cause a namespace issue, right?

src/corelocation.cs
```diff
@@ -736,4 +806,30 @@ interface CLVisit : NSSecureCoding, NSCopying {
...
+		IntPtr Constructor (NSUuid uuid, ushort major);
+
+		[Export ("initWithUUID:major:minor:")]
+		IntPtr Constructor (NSUuid uuid, ushort major, ushort minor);
+
+		[Export ("UUID", ArgumentSemantic.Copy)]
+		NSUuid Uuid { get; }
+
+		[NullAllowed, Export ("major", ArgumentSemantic.Copy)]
+		[BindAs (typeof (short?))]
```
whitneyschmidt: When do we use `BindAs`?

src/corelocation.cs
spouliot: Good catch! It was my first attempt too - but sadly it can't :( 

The problem are that:
* the name is already used in the type (even if overloads are fine most of the time); 
* a `*Delegate` type;
* it uses  `[EventArgs]`;

so this cause a conflict in the generated code.

I don't have this PR built right now - but I'll update the comment with the code (easier to see what's happening this way).

src/corelocation.cs
spouliot: There's a few cases, the most common one is `NSNumber`, like above. Another one are to support "smart enums" (using a C# enum that maps to several `NSString` constants).

So for some API Apple prefers to use a reference type, like `NSNumber`,  over a more specific value type. Why ? mainly because they want a nullable properties

But
```csharp
NSNumber Major { get; }
```
does not tell the API consumer if it should except a `short` (like it's meant to), a `float` or even a `bool`. That does not create a very discoverable API and we can do a lot better in C# with nullable `?` types :)

Why not
```csharp
short? Major { get; }
```

Because our runtime must know this an `NSNumber` natively - e.g. it's needed to allow us to override the property in a subclass and have some ObjC code call us. So the declaration remains an `NSNumber` but we give the generator a hint to add extra (conversion) code to expose a nicer API to our developers.

spouliot:
@mandel-macaque there's no change in corelocation for beta 4 (but I'll fix the typo found by @tj-devel709)

src/corelocation.cs
dalexsoto: Also we have some docs about it :) https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/objective-c-libraries?tabs=macos#binding-nsvalue-nsnumber-and-nsstring-to-a-better-type

whitneyschmidt: Thank you both for the detailed explanation! 👍

PR #6602, opened 2019-07-17
[xcode11] [ItunesLibrary] macOS b1 bindings

joconte:
diff : https://github.com/xamarin/xamarin-macios/wiki/iTunesLibrary-macOS-xcode11-beta1

It probably needs more work but we can discuss about it here 😄

### DIFF
--- a/src/iTunesLibrary/ITCompat.cs
+++ b/src/iTunesLibrary/ITCompat.cs
@@ -1,0 +1,14 @@
+using System;
+
+namespace iTunesLibrary
+{
+#if !XAMCORE_2_0
+    public partial class ITLibrary {
+
+        [Obsolete ("Does not return a valid instance on macOS 14.")]
+        public ITLibrary ()
+        {
+        }
+    }
+#endif
+}
--- a/src/iTunesLibrary/ITCompat.cs
+++ b/src/iTunesLibrary/ITCompat.cs
@@ -2,10 +2,10 @@ using System;
 
 namespace iTunesLibrary
 {
-#if !XAMCORE_2_0
+#if XAMCORE_4_0
     public partial class ITLibrary {
 
-        [Obsolete ("Does not return a valid instance on macOS 14.")]
+        [Obsolete ("This constructor does not create a valid instance of the type.")]
         public ITLibrary ()
         {
         }
--- a/Make.config
+++ b/Make.config
@@ -70,9 +70,9 @@ MIN_CMAKE_URL=https://cmake.org/files/v3.6/cmake-3.6.2-Darwin-x86_64.dmg
 MIN_CMAKE_VERSION=2.8.8
 
 # ObjectiveSharpie min/max versions
-MIN_SHARPIE_VERSION=3.4.35
+MIN_SHARPIE_VERSION=3.4.43
 MAX_SHARPIE_VERSION=3.4.99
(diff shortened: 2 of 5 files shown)


joconte:
@mandel-macaque Oh It's today ! Yeah I will double check.

joconte:
Well, according to the wiki, nothing changed in beta 4.

dalexsoto:
build

joconte:
Ouch build failed

spouliot:
@joconte there's an [example](https://github.com/xamarin/xamarin-macios/blob/master/src/CoreLocation/CLCompat.cs) and the new file needs to be added to `src/frameworks.sources`

spouliot:
The build error is a known random issue -> https://github.com/xamarin/maccore/issues/1888

joconte:
@spouliot Wow thanks for the detailed answer. I will fix it.

joconte:
@spouliot What about this ?

src/iTunesLibrary/ITCompat.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace iTunesLibrary
+{
+#if !XAMCORE_2_0
```
spouliot: close :) I should have mentioned it but `XAMCORE_2_0` is a few years back (when we did 32/64 unified)

What you need if `#if XAMCORE_4_0` so, the next time we read API compatibility (if ever), that code can be removed.

and, just to paint a complete picture, `XAMCORE_3_0` was used to exclude stuff when we added tvOS and watchOS support (so we did not bring `[Obsolete]`d API in the new platforms.

src/iTunesLibrary/ITCompat.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace iTunesLibrary
+{
+#if !XAMCORE_2_0
+    public partial class ITLibrary {
+
+        [Obsolete ("Does not return a valid instance on macOS 14.")]
```
spouliot: There's no need to mention the version - it's not clear if this a change in macOS 10.15 (or if it was already the case) and the type is new in macOS 10.14.

Also it's not `macOS 14` ;-) even if I wish that least year `X` removal (from Mac OSX) also removed the `10`

-> `[Obsolete ("This constructor does not create a valid instance of the type.")]`

^ it's an existing string (so it won't increase the metadata size)

src/iTunesLibrary/ITCompat.cs
joconte: Thanks a lot for the detailed explanations, I'm learning so much about the project everyday !

spouliot:
build

spouliot:
Fixed :) If the bots are happy (and they should be) then it can be merged. Thanks!

joconte:
Ahhh I forgot to delete the todo file

joconte:
I will to do it later today

joconte:
Todo file deleted

rolfbjarne:
build

dalexsoto:
build

dalexsoto:
It just needed one last bit @joconte, we use framework.sources to track files in our build system, since this PR added `ITCompat.cs` we just need our build system to be aware of it.

dalexsoto:
Thanks a lot for your contribution ❤️

PR #6613, opened 2019-07-18
[generator] Allow the generation of smart enums `*Extensions` code even if no fields are present

spouliot:
This is needed in a case like:

```csharp
[Unavailable (PlatformName.UIKitForMac)][Advice ("This API is not available when using UIKit on macOS.")]
[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
enum ASAuthorizationProviderAuthorizationOperation {
	// no value yet - but we must handle `nil` as a default value
	[DefaultEnumValue]
	[Field (null)]
	None,
}
```

IOW we want to expose the smart enum for it's only `null` value
but the general conversion code is not available (for a `Weak`
version of the property)

The upcoming update for `authenticationservices.cs` depends on this fix.

### DIFF
--- a/src/generator-enums.cs
+++ b/src/generator-enums.cs
@@ -97,7 +97,7 @@ public partial class Generator {
 
 		var library_name = type.Namespace;
 		var error = AttributeManager.GetCustomAttribute<ErrorDomainAttribute> (type);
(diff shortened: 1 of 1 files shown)


spouliot:
✅ Generator Diff (no change)

^ 😸

PR #6619, opened 2019-07-19
[AuthenticationServices] Fix ASAuthorization and ASAuthorizationRequest

dalexsoto:
Apple returns an internal wrapper object depending on the context the API
is used for the following properties

* `ASAuthorization.Provider`
* `ASAuthorization.Credential`
* `ASAuthorizationRequest.Provider`

The provider objects have a common interface `IASAuthorizationProvider`
and the Credential objects have` IASAuthorizationCredential` since the
objects that implement these protocols inherit from `NSObject` and Apple
returns the internal wrapper object we cannot have the runtime do the
right cast for us so we expose a generic API constrained to

* `NSObject, IASAuthorizationProvider`
* `NSObject, IASAuthorizationCredential`

Respectively so now we are able to do something like

```csharp
GetCredential<ASAuthorizationAppleIdCredential> ();
```

No unit test added because this requires special entitlements but
this is working as expected in an internal sample that it is being
worked on.

sample

### DIFF
--- a/src/AuthenticationServices/ASAuthorization.cs
+++ b/src/AuthenticationServices/ASAuthorization.cs
@@ -1,0 +1,20 @@
+//
+// ASAuthorization.cs
+//
+// Authors:
+//	Alex Soto <alexsoto@microsoft.com>
+//
+// Copyright 2019 Microsoft Corporation
+//
+
+using Foundation;
+using ObjCRuntime;
+
+namespace AuthenticationServices {
+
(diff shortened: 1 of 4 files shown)


dalexsoto:
Breaking changes are expected, this is all new API introduced by Xcode 11

dalexsoto:
Oh I forgot to mention here is the internal wrapper object

image

PR #6620, opened 2019-07-19
[Generator] Do not used harcoded 'error' var name, use the one in the declaration.

mandel-macaque:


dalexsoto:
```
tmp-test-dir/Xamarin.Tests.BGenTool.EnsureTempDir68/ObjCRuntime/Trampolines.g.cs(69,34): error CS1061: 'AudioBuffers' does not contain a definition for 'GetHandle' and no accessible extension method 'GetHandle' accepting a first argument of type 'AudioBuffers' could be found (are you missing a using directive or an assembly reference?)
```

Ouch you are ruining into this because `GetHandle` is internal, you might have to change the generator a little bit to remove `GetHandle` and do the `IntPtr.Zero` check yourself so this works with 3rd party bindings (which our tests are).

See? test all the things! 🙈

mandel-macaque:

PR #6624, opened 2019-07-19
[PhotosUI] Add Xcode 11 Beta 1 Binding 

tj-devel709:
No changes in beta 2-4

### DIFF
--- a/src/photosui.cs
+++ b/src/photosui.cs
@@ -471,7 +471,7 @@ namespace PhotosUI {
 	[iOS (8,0)]
 	[NoMac][NoTV]
 	[DisableDefaultCtor]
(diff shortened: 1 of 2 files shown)


src/photosui.cs
```diff
@@ -24,6 +24,8 @@ namespace PhotosUI {
 	[iOS (8, 0)]
 	[Mac (10, 13)]
 	[Protocol]
+	[Unavailable (PlatformName.UIKitForMac)]
```
whitneyschmidt: @tj-devel709 Can you share more about `UIKitForMac` and `[Unavailable]`? :)

tj-devel709: Hi Whitney, do you mean in these github comments or in the PR?

PR #6632, opened 2019-07-23
[system-dependencies] Improve error message when siminstaller can't find simulators to explain what to do.

rolfbjarne:


rolfbjarne:
It looks like this now (in case it fails):

Screenshot 2019-07-23 11 43 09

PR #6634, opened 2019-07-23
[Provisioning] Use correct public VSMac url.

mandel-macaque:


mandel-macaque:

mandel-macaque:
Failure is unrelated since we wanted to fix the provisioning issues.

PR #6638, opened 2019-07-24
[xtro] Sharpie has been fixed, so no need to avoid looking at some previously troublesome frameworks.

rolfbjarne:


rolfbjarne:
Looks like I didn't quite test this correctly, it's still quite broken 😒

Issue #4969, opened 2018-10-11
[RFC] Change what [Abstract] means when binding classes


rolfbjarne:
Current behavior: adding `[Abstract]` to a class in an Api definition makes the corresponding generated class abstract.

This is problematic because native code might create a custom subclass of the abstract native class. If we don't know about this custom subclass, it won't be bound, and when the time comes to create a managed wrapper for an instance of such a class, we can't, because the corresponding managed class is abstract.

References:

* [Cannot create an instance of UserNotifications.UNNotificationTrigger](https://github.com/xamarin/xamarin-macios/issues/3935)
* [BGTaskScheduler crash](https://github.com/xamarin/xamarin-macios/issues/7456)
* https://stackoverflow.com/questions/52710769/system-memberaccessexception-cannot-create-an-instance-of-usernotifications-unn

### Potential solution A

Change the meaning of `[Abstract]` for classes (not members) in an API definition to:

 * All constructors are automatically protected (if not already internal).
* The class is not abstract.

This will allow our runtime to create instances of such classes, when we get a native instance of a custom subclass. It also prevents users from trying to create instances of these specific classes, without subclassing them.

### Potential solution B

Generate an internal non-abstract subclass that can be used when we would otherwise have tried to create an instance of the abstract class.

We already do something similar for protocols (we have an internal managed class we can use to wrap a handle typed as a protocol).

Con: lots of boilerplate code (which the linker might not be able to remove).

migueldeicaza:
I do not like the approach of the second solution due to the additional code that we could not remove, while there are no downsides to the former

spouliot:
It would not be a lot of code (there's not a lot of `abstract` types) and the linker would remove any that are not needed (where the `abstract` type is not included).

It's a bit sad to lose the information that `abstract` convey to the developer. However we can output an `[Advice ("This type is meant to be subclassed.")]` that should give an hint to developers trying to use it.

alaskanrogue:
Given this issue, then the "Enhanced User Notification" approach to push notifications is not usable? Is the older deprecated approach still usable?

rolfbjarne:
With PR #6659 and [this diff](https://gist.github.com/rolfbjarne/47f5f8e2316b9bf8828c6192fed4079c), there are 143 API that reports to use abstract types (derived from NSObject): https://gist.github.com/rolfbjarne/9281b92c74af819a61423dabb9582790

PR #6659, opened 2019-07-26
[registrar] Report a warning when the registrar export an abstract INativeObject type to Objective-C.

rolfbjarne:
Exporting abstract types to Objective-C can lead to problems when at runtime
we're asked to create an instance of such a type (which we can't), so warn
when this happens.

This would have caught #6655, and the problems explained in #4969 as well.

Since this may trigger for code that's currently working fine, I'm making it a
warning instead of an error (which means adding some extra code to be able to
easily report warnings from the generator code).


rolfbjarne:
The test failure is unrelated and will be fixed once the xcode11 branch is merged into master.

PR #6671, opened 2019-07-29
[Tests] Ignore some tests on the CI due to the settings of the bots.

mandel-macaque:
In some cases some bots will return a proxy present, not because of the
PAC being parsed, but due to the bot settings. Ignore the tests that
expect no proxies in the CI fixes the issue.

Fixes https://github.com/xamarin/maccore/issues/1901


mandel-macaque:

mandel-macaque:
API diff issues can be ignored. They are not related.

PR #6680, opened 2019-07-29
[XHarness] Move the bcl tests to have their own dir.

mandel-macaque:


mandel-macaque:
Example of the generated tree: https://gist.github.com/mandel-macaque/28b18dd31ba8f5bbf760935e24dd1509 

How to know this works, well, this branch will have tests passing and no API diff annoying message.

PR #6683, opened 2019-07-29
[Introspection] ABRecord is abstract and cannot be instantiated.

mandel-macaque:
Following https://github.com/xamarin/xamarin-macios/pull/6655 the class was made non abstract but it cannot be instantiated. Ignore it.

Fixes: https://github.com/xamarin/maccore/issues/1903


mandel-macaque:
API diff can be ignored.

mandel-macaque:

PR #6707, opened 2019-08-02
[Tests] Ignore tests that only fail on tvOS devices.

mandel-macaque:
A mono issue has been reported here: https://github.com/mono/mono/issues/15983


mandel-macaque:

mandel-macaque:
No need to backport to Xcode 11 because we use a diff mono.

PR #6710, opened 2019-08-02
[xcode11] [HomeKit] Bindings for HomeKit beta1 - beta5 - all platforms

joconte:
diff : 
https://github.com/xamarin/xamarin-macios/wiki/HomeKit-iOS-xcode11-beta1
https://github.com/xamarin/xamarin-macios/wiki/HomeKit-iOS-xcode11-beta2
https://github.com/xamarin/xamarin-macios/wiki/HomeKit-iOS-xcode11-beta3
https://github.com/xamarin/xamarin-macios/wiki/HomeKit-iOS-xcode11-beta5

It's not complete yet, I will annotate where I need to clarify things.


src/HomeKit/HMEnums.cs
```diff
@@ -107,6 +107,14 @@ public enum HMError : long {
 		NoHomeHub = 91,
 		IncompatibleHomeHub = 92, // HMErrorCodeNoCompatibleHomeHub introduced and deprecated on iOS 11. HMErrorCodeIncompatibleHomeHub = HMErrorCodeNoCompatibleHomeHub.
 		IncompatibleAccessory = 93,
+		[iOS (12,0)]
+		ObjectWithSimilarNameExists = 95,
```
joconte: This one was not in the diff, maybe we forgot it before ?

src/HomeKit/HMEnums.cs
```diff
@@ -1136,6 +1144,19 @@ public enum HMSignificantEvent {
 		Sunset,
 	}
 
+	[iOS (13,0), TV (13,0), Watch (6,0), NoMac]
+	[Native]
```
joconte: I don't know if I need ```[Native]```or not. Maybe I need to dig further in the documentation.

src/homekit.cs
```diff
@@ -56,6 +56,10 @@ partial interface HMHomeManager {
 		[Async]
 		[Export ("removeHome:completionHandler:")]
 		void RemoveHome (HMHome home, Action<NSError> completion);
+		
+		[iOS (13,0), Watch (6,0), TV (13,0), NoMac]
+		[Export ("authorizationStatus")]
+		HMHomeManagerAuthorizationStatus AuthorizationStatus { get; [NotImplemented] set; }
```
joconte: ```[NotImplemented] set;```or just no ```set;```?

src/homekit.cs
```diff
@@ -1671,12 +1683,64 @@ interface HMNumberRange {
...
+		[NullAllowed, Export ("payloadWithURL:ownershipToken:")]
+		HMAccessorySetupPayload GetPayloadWithUrlAndOwnershipToken (NSUrl setupPayloadUrl, HMAccessoryOwnershipToken ownershipToken);
+	}
+
+	[iOS (13,0), Watch (6,0), TV (13,0), NoMac]
+	[DisableDefaultCtor]
+	interface HMNetworkConfigurationProfile : HMAccessoryProfile {
+		[Export ("delegate"), ArgumentSemantic.Weak)]
+		[NullAllowed]
+		HMNetworkConfigurationProfileDelegate WeakDelegate { get; set; }
```
joconte: I find this in the todo file :
```!missing-selector! HMNetworkConfigurationProfile::setDelegate: not bound```

But I don't find it in the diff files. 
Should I create separate getter and setter ?

File: src/homekit.cs [hunk 5]
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+		
+		[Export ("isNetworkAccessRestricted")]
+		bool NetworkAccessRestricted { get; [NotImplemented] set; }
+	}
+
+	[iOS (13,0), Watch (6,0), TV (13,0), NoMac]
+	interface HMNetworkConfigurationProfileDelegate {
```
joconte: It's probably incomplete, I didn't find 
```(void)profileDidUpdateNetworkAccessMode:(HMNetworkConfigurationProfile *)profile;```
in the todo file but it's here https://github.com/xamarin/xamarin-macios/wiki/HomeKit-iOS-xcode11-beta3

File: src/homekit.cs [hunk 6]
```diff
@@ -75,6 +79,14 @@ partial interface HMHomeManagerDelegate {
 
 		[Export ("homeManager:didRemoveHome:"), EventArgs ("HMHomeManager")]
 		void DidRemoveHome (HMHomeManager manager, HMHome home);
+
+		[iOS (13,0), NoWatch, NoTV, NoMac]
+		[Export("homeManager:didReceiveAddAccessoryRequest:")]
+		void DidReceiveAddAccessoryRequest (HMHomeManager manager, HMAddAccessoryRequest request);
```
whitneyschmidt: Nit: update to something more like `DidReceiveAddAccessory` or `DidReceiveAdd` since `AddAccessoryRequest` is a suffix that describes one of the parameter values. I'm not sure how familiar with ObjC you are (I'm new to it!) but generally method names tend to contain the names/types of one of their parameters. When translating to C# we usually remove the suffix :)

@dalexsoto can you confirm?

src/homekit.cs
```diff
@@ -75,6 +79,14 @@ partial interface HMHomeManagerDelegate {
...
+		
+		[iOS (13,0), NoWatch, NoTV, NoMac]
+		[Export("homeManager:didUpdateAuthorizationStatus:")]
+		void DidUpdateAuthorizationStatus (HMHomeManager manager, HMHomeManagerAuthorizationStatus status);
```
whitneyschmidt: Same as ^, `DidUpdateAuthorization`

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
 		NSNumber Max { get; }
 	}
 
+	[iOS (13,0), NoWatch, NoMac, NoTV]
+	[DisableDefaultCtor]
+	interface HMAccessoryOwnershipToken
+	{
+		[Export("initWithData:")]
+		IntPtr Constructor(NSData data);
```
dalexsoto: Missing space

```suggestion
		IntPtr Constructor (NSData data);
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+	}
+
+	[iOS (13,0), NoWatch, NoMac, NoTV]
+	[DisableDefaultCtor]
+	interface HMAddAccessoryRequest
+	{
+		[Export ("home")] 
+		HMHome Home { get; [NotImplemented] set; }
```
dalexsoto:
```suggestion
		HMHome Home { get; }
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+		
+		[Export ("accessoryCategory")]
+		HMAccessoryCategory AccessoryCategory { get; [NotImplemented] set; }
```
dalexsoto:
```suggestion
		HMAccessoryCategory AccessoryCategory { get; }
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+		
+		[Export ("accessoryName")]
+		NSString AccessoryName { get; [NotImplemented] set; }
```
dalexsoto:
```suggestion
		NSString AccessoryName { get; }
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+		
+		[Export ("requiresSetupPayloadURL")]
+		bool RequiresSetupPayloadURL { get; [NotImplemented] set; }
```
dalexsoto: URL -> Url naming conventions

```suggestion
		bool RequiresSetupPayloadUrl { get; }
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+		
+		[Export ("requiresOwnershipToken")]
+		bool RequiresOwnershipToken { get; [NotImplemented] set; }
```
dalexsoto:
```suggestion
		bool RequiresOwnershipToken { get; }
```

src/homekit.cs
dalexsoto: No, this should be bound as

```csharp
[Wrap ("WeakDelegate")]
[NullAllowed]
IHMNetworkConfigurationProfileDelegate Delegate { get; set; }

```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+
+	[iOS (13,0), Watch (6,0), TV (13,0), NoMac]
+	[DisableDefaultCtor]
+	interface HMNetworkConfigurationProfile : HMAccessoryProfile {
+		[Export ("delegate"), ArgumentSemantic.Weak)]
+		[NullAllowed]
+		HMNetworkConfigurationProfileDelegate WeakDelegate { get; set; }
+		
+		[Export ("isNetworkAccessRestricted")]
+		bool NetworkAccessRestricted { get; [NotImplemented] set; }
```
dalexsoto: The right selector is `isNetworkAccessRestricted`

```suggestion
		bool NetworkAccessRestricted { [Bind ("isNetworkAccessRestricted")] get; }
```

src/homekit.cs
```diff
@@ -1671,12 +1683,65 @@ interface HMNumberRange {
...
+	}
+
+	[iOS (13,0), Watch (6,0), TV (13,0), NoMac]
+	[DisableDefaultCtor]
+	interface HMNetworkConfigurationProfile : HMAccessoryProfile {
+		[Export ("delegate"), ArgumentSemantic.Weak)]
+		[NullAllowed]
+		HMNetworkConfigurationProfileDelegate WeakDelegate { get; set; }
+		
+		[Export ("isNetworkAccessRestricted")]
```
dalexsoto: While what you did is correct, we like to preserve the information like in the headers so this is what we would do.

```suggestion
		[Export ("networkAccessRestricted")]
```

src/homekit.cs
dalexsoto: Yup this looks like you want 

```csharp

interface IHMNetworkConfigurationProfileDelegate { }

[Watch (6,0), TV (13,0), NoMac, iOS (13,0)]
[Protocol, Model (AutoGeneratedName = true)]
interface HMNetworkConfigurationProfileDelegate {

   [Export ("profileDidUpdateNetworkAccessMode:")]
   void DidUpdateNetworkAccessMode (HMNetworkConfigurationProfile profile);
}
```

src/homekit.cs
```diff
@@ -1671,12 +1683,63 @@ interface HMNumberRange {
...
+		NSString AccessoryName { get; }
+		
+		[Export ("requiresSetupPayloadURL")]
+		bool RequiresSetupPayloadUrl { get; }
+		
+		[Export ("requiresOwnershipToken")]
+		bool RequiresOwnershipToken { get; }
+
+		[NullAllowed, Export ("payloadWithOwnershipToken:")]
+		HMAccessorySetupPayload GetPayloadWithOwnershipToken (HMAccessoryOwnershipToken ownershipToken);
```
VincentDondain:
```suggestion
		HMAccessorySetupPayload GetPayload (HMAccessoryOwnershipToken ownershipToken);
```

src/homekit.cs
```diff
@@ -1671,12 +1683,63 @@ interface HMNumberRange {
...
+		
+		[NullAllowed, Export ("payloadWithURL:ownershipToken:")]
+		HMAccessorySetupPayload GetPayloadWithUrlAndOwnershipToken (NSUrl setupPayloadUrl, HMAccessoryOwnershipToken ownershipToken);
```
VincentDondain:
```suggestion
		HMAccessorySetupPayload GetPayload (NSUrl setupPayloadUrl, HMAccessoryOwnershipToken ownershipToken);
```

This is the Objective-C naming, in C# it's inferred from the parameter names (:

src/homekit.cs
```diff
@@ -75,6 +79,14 @@ partial interface HMHomeManagerDelegate {
 
 		[Export ("homeManager:didRemoveHome:"), EventArgs ("HMHomeManager")]
 		void DidRemoveHome (HMHomeManager manager, HMHome home);
+
+		[iOS (13,0), NoWatch, NoTV, NoMac]
+		[Export("homeManager:didReceiveAddAccessoryRequest:")]
```
VincentDondain:
```suggestion
		[Export("homeManager:didReceiveAddAccessoryRequest:"), EventArgs ("HMHomeManager")]
```

src/homekit.cs
```diff
@@ -75,6 +79,14 @@ partial interface HMHomeManagerDelegate {
...
+		void DidReceiveAddAccessoryRequest (HMHomeManager manager, HMAddAccessoryRequest request);
+		
+		[iOS (13,0), NoWatch, NoTV, NoMac]
+		[Export("homeManager:didUpdateAuthorizationStatus:")]
```
VincentDondain:
```suggestion
		[Export("homeManager:didUpdateAuthorizationStatus:"), EventArgs ("HMHomeManager")]
```

src/homekit.cs
```diff
@@ -1671,12 +1683,71 @@ interface HMNumberRange {
...
 	[NoWatch, NoTV, iOS (11,3)]
 	[BaseType (typeof(NSObject))]
 	[DisableDefaultCtor]
 	interface HMAccessorySetupPayload {
-		[Export ("initWithURL:")]
-		IntPtr Constructor (NSUrl setupPayloadUrl);
+		[NullAllowed, Export ("initWithURL:")]
+		IntPtr Constructor ([NullAllowed] NSUrl setupPayloadUrl);
+		
+		[NullAllowed, Export ("initWithURL:ownershipToken:")]
```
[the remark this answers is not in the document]
joconte: Nice catch :)

src/homekit.cs
joconte: Can I have some explanation about this one ? Or maybe a link to the documentation ? I had no idea about this one 👀

VincentDondain: @joconte https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#generating-events 😉

src/HomeKit/HMEnums.cs
```diff
@@ -1136,6 +1140,18 @@ public enum HMSignificantEvent {
 		Sunset,
 	}
 
+	[iOS (13,0), TV (13,0), Watch (6,0), NoMac]
+	public enum HMHomeManagerAuthorizationStatus : long {
```
spouliot: there's no need to use `: long` here - the default 32bits enum has ample space

src/homekit.cs
```diff
@@ -1671,12 +1683,72 @@ interface HMNumberRange {
...
+	[DisableDefaultCtor]
+	interface HMAddAccessoryRequest {
+		[Export ("home")] 
+		HMHome Home { get; }
+		
+		[Export ("accessoryCategory")]
+		HMAccessoryCategory AccessoryCategory { get; }
+		
+		[Export ("accessoryName")]
+		NSString AccessoryName { get; }
```
spouliot: why not `string` ?
we normally use `NSString` for constants that are defined by Apple (and which, sometimes, compare pointers and not the string content)

src/HomeKit/HMEnums.cs
joconte: I just kinda followed what was already in the file but yes if It's better to use the default 32bits let's go for it.
e.g. : 
```C#
[iOS (9,0)]
[TV (10,0)]
[Native]
public enum HMCharacteristicValueAirQuality : long {
	Unknown = 0,
	Excellent,
	Good,
	Fair,
	Inferior,
	Poor
}
```

spouliot: I see, but that's not the same case :) 

A `[Native]` enum means it exists, as an `enum` in Objective-C and we **must** use the same size (which is `NSInteger`, so 64 bits open 64 bits architectures). Also special care is needed when used in 32bits (e.g. for p/invoke).

But in this case you are creating a new enum (there's nothing in ObjC to match it) to better represent string constants (more type safe, helping code completion...). Since there's no contract (w/ObjC) you can use the default (smaller) enum size (which is `int` for C#).

If this was a `[Flag]` enum and you could not be sure if more than 32 bits would be needed (in the future) then `[u]long` could make sense too.

joconte: Wow thanks for the detailed answer 👍

src/HomeKit/HMEnums.cs
```diff
@@ -1141,7 +1141,7 @@ public enum HMSignificantEvent {
 	}
 
 	[iOS (13,0), TV (13,0), Watch (6,0), NoMac]
-	public enum HMHomeManagerAuthorizationStatus : long {
+	public enum HMHomeManagerAuthorizationStatus : int {
```
spouliot: minor (no need to change it if it's complete) but
```csharp
public enum HMHomeManagerAuthorizationStatus {
```
is identical (since `int` is the default)

joconte: Thanks

VincentDondain:
build

VincentDondain:
This should be good to go now. I double checked everything with Sharpie and fixed a couple of things (including the intro tests).
This was pretty minor though so thanks again @joconte for the amazing work ❤️

_Note: everyone please re-review (:_

joconte:
@VincentDondain TBH the first commit was really far from what we have now but I learned a **LOT** about the project on this one. I will need to review this PR when everything is good to fully understand and capitalize skills. This way I will be able to provide better contributions in the future.

BTW thanks to all the team for the support !

VincentDondain:
build

VincentDondain:
Known issue: https://github.com/xamarin/maccore/issues/649

PR #6717, opened 2019-08-06
[UIKit] Update bindings to Xcode 11 Beta 5

dalexsoto:


dalexsoto:
Breaking changes expected on new type

src/uikit.cs
```diff
@@ -21315,4 +21340,103 @@ interface UITableViewDiffableDataSource<SectionIdentifierType,ItemIdentifierType
...
+		[Abstract]
+		[Export ("itemProvidersForActivityItemsConfiguration", ArgumentSemantic.Copy)]
+		NSItemProvider [] ItemProvidersForActivityItemsConfiguration { get; }
+
+		[Export ("activityItemsConfigurationSupportsInteraction:")]
+		bool GetActivityItemsConfigurationSupportsInteraction (NSString activityItemsConfigurationInteraction);
+
+		[Export ("activityItemsConfigurationMetadataForKey:")]
+		[return: NullAllowed]
+		NSObject GetActivityItemsConfigurationMetadata (NSString activityItemsConfigurationMetadataKey);
```
mandel-macaque: Do we have an enum for this keys? Use a smart enum, expose the string one as an advance one.

src/uikit.cs
```diff
@@ -21315,4 +21340,103 @@ interface UITableViewDiffableDataSource<SectionIdentifierType,ItemIdentifierType
...
+	[NoWatch, NoTV, iOS (13,0)]
+	[Protocol]
+	interface UIActivityItemsConfigurationReading {
+
+		[Abstract]
+		[Export ("itemProvidersForActivityItemsConfiguration", ArgumentSemantic.Copy)]
+		NSItemProvider [] ItemProvidersForActivityItemsConfiguration { get; }
+
+		[Export ("activityItemsConfigurationSupportsInteraction:")]
+		bool GetActivityItemsConfigurationSupportsInteraction (NSString activityItemsConfigurationInteraction);
```
mandel-macaque: Do we have an enum for this and then expose the string one for advance users?

src/uikit.cs
```diff
@@ -21315,4 +21340,103 @@ interface UITableViewDiffableDataSource<SectionIdentifierType,ItemIdentifierType
...
+		NSObject GetActivityItemsConfigurationMetadata (NSString activityItemsConfigurationMetadataKey);
+
+		[Export ("activityItemsConfigurationMetadataForItemAtIndex:key:")]
+		[return: NullAllowed]
+		NSObject GetActivityItemsConfigurationMetadata (nint index, NSString activityItemsConfigurationMetadataKey);
+
+		[Export ("activityItemsConfigurationPreviewForItemAtIndex:intent:suggestedSize:")]
+		[return: NullAllowed]
+		NSItemProvider GetActivityItemsConfigurationPreview (nint index, NSString activityItemsConfigurationPreviewIntent, CGSize suggestedSize);
+
```
mandel-macaque: In all the methods, or most of them we have `GetActivityItemsConfiguration` then either preview or metadata. Can we just use GetMetadata and GetPreview?

src/uikit.cs
dalexsoto: You can't because this is a protocol :)

src/uikit.cs
dalexsoto: You can't because this is a protocol :)

src/uikit.cs
dalexsoto: You can't because this is a protocol :) Also I think the name is important because this is a protocol and can be implemented by several, also I think this avoids a round trip to documentation

dalexsoto: and also goes by hand with the property `ItemProvidersForActivityItemsConfiguration`

mandel-macaque: Meh.. I'm going to accept it because Swift uses the same.. but what a horrible name.

dalexsoto: And swift uses the same

```
optional func activityItemsConfigurationPreviewForItem(at index: Int, 
                                                intent: UIActivityItemsConfigurationPreviewIntent, 
                                         suggestedSize: CGSize) -> NSItemProvider?
```

PR #6720, opened 2019-08-06
[publishing] Added conditional behavior to CreateIpa and its dependen…

mauroa:
…cies

The targets execution needs to change based on the 'IsAppDistribution' property value, which would be set from the IDE.
According to the value of this property, some dependencies should be run or should be skipped.
The strategy adopted is to use the BeforeTargets value because doing this way we will force the conditionals over '$(IsAppDistribution)' to be evaluated each time a target is attempted to be run.
The most intuitive way of doing it is by directly add DependsOnTargets and conditions on the PropertyGroup items, but doing this will not work because the IDE usually runs the targets when the MSBuild project is already loaded, which means that the properties have already been evaluated so they will not be evaluated again, which causes conditional values to be out of date.
The only way of making it work in these cases is by forcing the conditional evaluations of each target execution attempt, and that is accomplished by adding the conditions directly on the Target 'Condition' property.


emaf:
Closing in favor of https://github.com/xamarin/xamarin-macios/pull/6753

PR #6731, opened 2019-08-07
[coretext] Update for Xcode 11 beta 5

spouliot:
Beta 4 added some new API and constants
Beta 5 removed `CTFontManagerCopyRegisteredFontDescriptors` from all OS but iOS

### DIFF
--- a/src/CoreText/CTFontManager.cs
+++ b/src/CoreText/CTFontManager.cs
@@ -401,17 +401,20 @@ namespace CoreText {
 			}
 		}
 
-		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+#if __IOS__
+		[iOS (13,0)]
 		[DllImport (Constants.CoreTextLibrary)]
 		static extern /* CFArrayRef */ IntPtr CTFontManagerCopyRegisteredFontDescriptors (CTFontManagerScope scope, bool enabled);
 
-		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[iOS (13,0)]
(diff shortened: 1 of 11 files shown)


src/CoreText/CTFontManager.cs
```diff
@@ -401,17 +401,20 @@ public static void UnregisterFontDescriptors (CTFontDescriptor[] fontDescriptors
...
 
-		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+#if __IOS__
+		[iOS (13,0)]
 		[DllImport (Constants.CoreTextLibrary)]
 		static extern /* CFArrayRef */ IntPtr CTFontManagerCopyRegisteredFontDescriptors (CTFontManagerScope scope, bool enabled);
 
-		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[iOS (13,0)]
+		[NoWatch][NoTV][NoMac]
```
dalexsoto: Do we really want this if it is inside a `#if __IOS__`?

spouliot:
build

src/CoreText/CTFontManager.cs
spouliot: Not sure but no harm :) 

FWIW we're not fully consistent with `[No*]` attributes (e.g. we have quite a few `[NoMac]` in frameworks not available on macOS).

Also it might have (future) value if we could get the IDE to show, somehow, all platform availability. E.g. you might be writing code for iOS but you might want to share that code at some point and pick the API that are more cross-platform.

PR #6733, opened 2019-08-07
[mediaaccessibility] Update for Xcode 11 beta 5

spouliot:


src/MediaAccessibility/MediaAccessibility.cs
```diff
@@ -262,6 +262,34 @@ public static MACaptionAppearanceTextEdgeStyle GetTextEdgeStyle (MACaptionAppear
...
+			}
+		}
+
+		[TV (13,0), Mac (10,15), iOS (13,0)]
+		public static void DidDisplayCaptions (NSAttributedString[] strings)
+		{
+			// CFAttributedString is “toll-free bridged” with its Foundation counterpart, NSAttributedString.
+			// https://developer.apple.com/documentation/corefoundation/cfattributedstring?language=objc
+			if ((strings == null) || (strings.Length == 0))
+				MACaptionAppearanceDidDisplayCaptions (IntPtr.Zero);
```
whitneyschmidt: Why do we use `IntPtr.Zero` in place of `null`?

spouliot: It's basically all the same, just like `0` would be - but
* C# is type safe (unlike C/C++); and
* `IntPtr` is a `struct` not a reference type so it can't be set to `null`; and

We could use pointers in C# but that would expose a lot of `unsafe` code everywhere... which does not make great API.

We could have a 2nd `[DllImport]` with a different signature - but that would bloat the code.

whitneyschmidt: Thank you for the detailed explanation, makes a lot of sense! ❤️

PR #6738, opened 2019-08-08
[master] Bump mono to head of 2019-06 (a791989)

dalexsoto:
Diff: https://github.com/mono/mono/compare/17ac7dcfc74e7371b99a1b7bf73d17e78df90c55...a7919897e8bc1268547d60ca2283e619a3b3cc45


mandel-macaque:
@dalexsoto there are a number of HE0038 issues, which we should not worry about, but I'm curious about the tvOS symlink tests and the ProcessTest crashing on the Mac. The embedded resources can be ignored on the tvOS because is only under the interpreter.

PR #6745, opened 2019-08-09
[photos][macos] Fix breaking changes for XM

spouliot:


spouliot:
Good catch, I did not see it - only the missing final `.` caught my eye. Not worth a few more bot hours but we can fix it in the upcoming betas.

PR #6746, opened 2019-08-09
[externalaccessory] Update for Xcode 11 beta 5

spouliot:


spouliot:
I rather have the default / empty text when no alternative is suggested. That goes back to the original meaning, i.e. "{Apple} does not approve", without bloating the assemblies with strings that might not be (or remain) accurate.

PR #6748, opened 2019-08-10
[xcode11] [AVKit] bindings for all platforms

joconte:
diff : 
https://github.com/xamarin/xamarin-macios/wiki/AVKit-iOS-xcode11-beta1
https://github.com/xamarin/xamarin-macios/wiki/AVKit-iOS-xcode11-beta2
https://github.com/xamarin/xamarin-macios/wiki/AVKit-macOS-xcode11-beta1
https://github.com/xamarin/xamarin-macios/wiki/AVKit-tvOS-xcode11-beta1

I had to refactor some of the actual code because of the fact that some interfaces are now available to macOS and were inside a ```if !MONOMAC```block. 

I also don't know which names I should choose for the events.


src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
 		[NoiOS, TV (11,0), NoWatch, NoMac]
 		[Export ("playerViewController:willTransitionToVisibilityOfTransportBar:withAnimationCoordinator:")]
 		void WillTransitionToVisibilityOfTransportBar ([NullAllowed] AVPlayerViewController playerViewController, bool visible, [NullAllowed] IAVPlayerViewControllerAnimationCoordinator coordinator);
+		
+		[iOS (12,0), NoTV, NoWatch, NoMac]
+		[Export ("playerViewController:willBeginFullScreenPresentationWithAnimationCoordinator:")]
+		void PlayerViewControllerWillBeginFullScreenPresentation (AVPlayerViewController playerViewController, UIViewControllerTransitionCoordinator coordinator);
```
joconte: Here is the error about UIViewControllerTransitionCoordinator not being in UIKit (despite the fact it is).

src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
...
+		
+		[iOS (12,0), NoTV, NoWatch, NoMac]
+		[Export ("playerViewController:willEndFullScreenPresentationWithAnimationCoordinator:")]
+		void PlayerViewControllerWillEndFullScreenPresentation (AVPlayerViewController playerViewController, UIViewControllerTransitionCoordinator coordinator);
```
joconte: Here is the error about UIViewControllerTransitionCoordinator not being in UIKit (despite the fact it is).

src/avkit.cs
```diff
@@ -383,8 +359,57 @@ interface AVPlayerView {
...
+	interface AVPlayerViewPictureInPictureDelegate
+	{
+		[Export ("playerViewWillStartPictureInPicture:")]
+		void PlayerViewWillStartPictureInPicture (AVPlayerView playerView);
+
+		[Export ("playerViewDidStartPictureInPicture:")]
+		void PlayerViewDidStartPictureInPicture (AVPlayerView playerView);
+
+		[Export ("playerView:failedToStartPictureInPictureWithError:")]
+		void PlayerView (AVPlayerView playerView, NSError error);
```
joconte: What name for the ```EventArgs```?

src/avkit.cs
```diff
@@ -383,8 +359,57 @@ interface AVPlayerView {
...
+
+		[Export ("playerViewWillStopPictureInPicture:")]
+		void PlayerViewWillStopPictureInPicture (AVPlayerView playerView);
+
+		[Export ("playerViewDidStopPictureInPicture:")]
+		void PlayerViewDidStopPictureInPicture (AVPlayerView playerView);
+
+		[Export ("playerView:restoreUserInterfaceForPictureInPictureStopWithCompletionHandler:")]
+		void PlayerView (AVPlayerView playerView, Action<bool> completionHandler);
```
joconte: What name for the ```EventArgs```?

src/avkit.cs
```diff
@@ -598,4 +594,164 @@ interface UIWindow_AVAdditions {
...
+	interface AVPictureInPictureControllerDelegate
+	{
+		[Export ("pictureInPictureControllerWillStartPictureInPicture:")]
+		void WillStartPictureInPicture (AVPictureInPictureController pictureInPictureController);
+	
+		[Export ("pictureInPictureControllerDidStartPictureInPicture:")]
+		void DidStartPictureInPicture (AVPictureInPictureController pictureInPictureController);
+	
+		[Export ("pictureInPictureController:failedToStartPictureInPictureWithError:")]
+		void FailedToStartPictureInPicture (AVPictureInPictureController pictureInPictureController, NSError error);
```
joconte: What name for the ```EventArgs```?

src/avkit.cs
```diff
@@ -598,4 +594,164 @@ interface UIWindow_AVAdditions {
...
+	
+		[Export ("pictureInPictureControllerWillStopPictureInPicture:")]
+		void WillStopPictureInPicture (AVPictureInPictureController pictureInPictureController);
+	
+		[Export ("pictureInPictureControllerDidStopPictureInPicture:")]
+		void DidStopPictureInPicture (AVPictureInPictureController pictureInPictureController);
+	
+		[Export ("pictureInPictureController:restoreUserInterfaceForPictureInPictureStopWithCompletionHandler:")]
+		void RestoreUserInterfaceForPictureInPicture (AVPictureInPictureController pictureInPictureController, Action<bool> completionHandler);
```
joconte: What name for the ```EventArgs```?

src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
 		[NoiOS, TV (11,0), NoWatch, NoMac]
 		[Export ("playerViewController:willTransitionToVisibilityOfTransportBar:withAnimationCoordinator:")]
 		void WillTransitionToVisibilityOfTransportBar ([NullAllowed] AVPlayerViewController playerViewController, bool visible, [NullAllowed] IAVPlayerViewControllerAnimationCoordinator coordinator);
+		
+		[iOS (12,0), NoTV, NoWatch, NoMac]
+		[Export ("playerViewController:willBeginFullScreenPresentationWithAnimationCoordinator:")]
+		void WillBeginFullScreenPresentation (AVPlayerViewController playerViewController, UIViewControllerTransitionCoordinator coordinator);
```
joconte: Here is the error about UIViewControllerTransitionCoordinator not being in UIKit (despite the fact it is).

src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
...
+		
+		[iOS (12,0), NoTV, NoWatch, NoMac]
+		[Export ("playerViewController:willEndFullScreenPresentationWithAnimationCoordinator:")]
+		void WillEndFullScreenPresentation (AVPlayerViewController playerViewController, UIViewControllerTransitionCoordinator coordinator);
```
joconte: Here is the error about UIViewControllerTransitionCoordinator not being in UIKit (despite the fact it is).

src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
...
+		
+		[TV (13,0)]
+		[Export ("nextChannelInterstitialViewControllerForPlayerViewController:")]
+		UIViewController NextChannelInterstitialViewController (AVPlayerViewController playerViewController);
+		
+		[TV (13,0)]
+		[Export ("playerViewController:skipToNextChannel:")]
+		void SkipToNextChannel (AVPlayerViewController playerViewController, Action<bool> completion);
```
joconte: What name for the ```EventArgs```?

src/avkit.cs
```diff
@@ -309,6 +243,48 @@ interface AVPlayerViewControllerDelegate
...
+		
+		[TV (13,0)]
+		[Export ("playerViewController:skipToPreviousChannel:")]
+		void SkipToPreviousChannel (AVPlayerViewController playerViewController, Action<bool> completion);
```
joconte: What name for the ```EventArgs```?

spouliot:
build

src/AVKit/Enums.cs
```diff
@@ -31,16 +31,32 @@ public enum AVPlayerViewTrimResult : long {
...
 		None = 0,
 		Unknown = -1000,
-		PictureInPictureStartFailed = -1001
+		PictureInPictureStartFailed = -1001,
+		[TV (13,0)]
+		ContentRatingUnknown = -1100,
+		[TV (13,0)]
+		ContentDisallowedByPasscode = -1101,
+		[TV (13,0)]
+		ContentDisallowedByProfile = -1102,
```
spouliot: all those new members will also be new in iOS 13 but...
we do not add availability attributes on error codes (not useful, they are output only and you can't ignore them if your code will be run on newer OS).

src/avfoundation.cs
```diff
@@ -10930,6 +10931,18 @@ interface AVPlayerItem_AVPlayerItemProtectedContent {
 		AVContentAuthorizationStatus GetContentAuthorizationRequestStatus ();
 	}
 
+	[TV (13,0), NoWatch, NoMac, NoiOS]
+	[Category]
+	[BaseType (typeof(AVPlayerItem))]
```
spouliot: minor: space before `(`

src/avkit.cs
```diff
@@ -16,86 +16,12 @@
 #else
 using AppKit;
 #endif
+#if MONOMAC
+using UIImage=AppKit.NSImage;
+#endif
 
 namespace AVKit {
 #if !MONOMAC
-	[NoTV]
```
spouliot: [re]move the `#if !MONOMAC` directive
this will reduce the size of the diff to review and makes it much easier to see any other changes

src/avkit.cs
```diff
@@ -383,8 +359,57 @@ interface AVPlayerView {
...
 		[Export ("showsSharingServiceButton")]
 		bool ShowsSharingServiceButton { get; set; }
+		
+		[Mac (10, 15)]
+		[Export ("allowsPictureInPicturePlayback")]
+		bool AllowsPictureInPicturePlayback { get; set; }
+		
+		[Wrap ("WeakPictureInPictureDelegate")]
+		[NullAllowed]
+		IAVPlayerViewPictureInPictureDelegate PictureInPictureDelegate { get; set; }
```
spouliot: that should also have it's `[Mac (10,15)]` attribute

src/avkit.cs
```diff
@@ -518,23 +543,7 @@ interface AVKitMetadataIdentifier {
...
-		IAVRoutePickerViewDelegate Delegate { get; set; }
-
-		[Export ("activeTintColor", ArgumentSemantic.Assign), NullAllowed]
-		UIColor ActiveTintColor { get; set; }
-
-		[NoiOS]
-		[Export ("routePickerButtonStyle", ArgumentSemantic.Assign)]
-		AVRoutePickerViewButtonStyle RoutePickerButtonStyle { get; set; }
-	}
+
```
spouliot: same, put an `#endif` and a `#if MONOMAC` after it, i.e. don't move the sources

src/avkit.cs
```diff
@@ -544,19 +553,6 @@ public enum AVRoutePickerViewButtonStyle : long {
...
-	[Protocol, Model]
-	[BaseType (typeof (NSObject))]
-	interface AVRoutePickerViewDelegate {
-
-		[Export ("routePickerViewWillBeginPresentingRoutes:")]
-		void WillBeginPresentingRoutes (AVRoutePickerView routePickerView);
-
-		[Export ("routePickerViewDidEndPresentingRoutes:")]
-		void DidEndPresentingRoutes (AVRoutePickerView routePickerView);
-	}
```
spouliot: same

spouliot:
build

src/AVKit/Enums.cs
```diff
@@ -31,16 +31,32 @@ public enum AVPlayerViewTrimResult : long {
 		OKButton,
 		CancelButton
 	}
+	
+	[Mac (10,15)]
+	[Native]
+	public enum AVRoutePickerViewButtonState : long
+	{
```
VincentDondain: The brace should be on the same line as the type. https://www.mono-project.com/community/contributing/coding-guidelines. See the little summary table on that page. If you're using Visual Studio for Mac, you can setup the coding guidelines to `mono` and it will do that automatically for you.

src/avfoundation.cs
```diff
@@ -10869,8 +10869,9 @@ interface AVPlayerItem : NSCopying {
 		[Export ("navigationMarkerGroups", ArgumentSemantic.Copy)]
 		AVNavigationMarkersGroup[] NavigationMarkerGroups { get; set; }
 
-		[NoiOS][NoMac]
+		[NoMac]
 		[TV (9,0)]
+		[iOS (12,0)]
```
VincentDondain: This is new in iOS 12? Not 13? Ideally you'd need to run the introspection tests on an iOS 12 simulator to confirm (you can do so by opening the `test.sln`, select introspection as a target (top left) and deploy on an iOS 12 sim (you might need to go to Xcode's preferences and download an older simulator under `Components`).

src/avkit.cs
```diff
@@ -208,6 +134,14 @@ interface AVPlayerViewController {
 		[NoiOS, TV (11,2), NoMac, NoWatch]
 		[Export ("appliesPreferredDisplayCriteriaAutomatically")]
 		bool AppliesPreferredDisplayCriteriaAutomatically { get; set; }
+		
+		[iOS (9, 0), TV (13,0), NoWatch]
```
VincentDondain: It was probably private before iOS 13, a quick simulator test to ensure it's actually available on older versions would help.

src/avkit.cs
VincentDondain: Everywhere there's an `Action` you're gonna need the `[Async]` attribute. See https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#asyncattribute

src/avkit.cs
```diff
@@ -323,7 +299,7 @@ interface AVPlayerViewControllerAnimationCoordinator {
 	}
 
 #else
-
+
```
VincentDondain: Minor but careful with whitespace changes (please do not commit them).

A nice trick to avoid that is when you need to stage your changes before committing, try to use `git add -i`. Then you do (5 to patch), select all your files (1,2,3...) and that will have you review every single line of code you need to stage (you can use `s` to split if needed).
So you go line by line and press (y/n) and can easily ignore the whitespace changes this way.
You'll end up with staged and unstaged changes (all unstaged would be the things you don't want to commit). Then you can `git commit` and `git reset --hard` all the unstaged / unwanted code.

VincentDondain:
@joconte for the `EventArgs` it's on a case by case basis, see https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#generating-events for inspiration, try to use a name that makes sense. Could be a combination of the method's name and parameters or just method name. Sometimes that can be shares if it's on multiple methods with the same parameters so you need something more generic.
However ultimately this doesn't matter too much as people will not really use that name you provide.
See link I provided but they would just do `scrollView.ShouldScrollToTop = (sv) => MakeDecision ();`

src/avfoundation.cs
```diff
@@ -10930,6 +10931,18 @@ interface AVPlayerItem_AVPlayerItemProtectedContent {
 		AVContentAuthorizationStatus GetContentAuthorizationRequestStatus ();
 	}
 
+	[TV (13,0), NoWatch, NoMac, NoiOS]
+	[Category]
+	[BaseType (typeof (AVPlayerItem))]
+	interface AVPlayerItem_AVPlaybackRestrictions
+	{
+		[Export ("requestPlaybackRestrictionsAuthorization:")]
+		void RequestPlaybackRestrictionsAuthorization (Action<bool, NSError> completion);
```
VincentDondain: `[Async]`

src/avfoundation.cs
```diff
@@ -10930,6 +10931,18 @@ interface AVPlayerItem_AVPlayerItemProtectedContent {
 		AVContentAuthorizationStatus GetContentAuthorizationRequestStatus ();
 	}
 
+	[TV (13,0), NoWatch, NoMac, NoiOS]
+	[Category]
+	[BaseType (typeof (AVPlayerItem))]
+	interface AVPlayerItem_AVPlaybackRestrictions
+	{
```
VincentDondain: Same line (:

src/avkit.cs
spouliot: No, that's fine here since this is a protocol

src/avkit.cs
joconte: That's a nice trick right there 😄 thanks

src/avfoundation.cs
joconte: Yes I found it strange also but : https://developer.apple.com/documentation/avfoundation/avplayeritem/1627623-externalmetadata

src/avkit.cs
```diff
@@ -544,9 +688,10 @@ public enum AVRoutePickerViewButtonStyle : long {
 		Custom,
 	}
 
+#endif
 	interface IAVRoutePickerViewDelegate { }
 
-	[TV (11,0), iOS (11,0)]
+	[TV (11,0), iOS (11,0), Mac (10,15)]
 	[Protocol, Model]
```
joconte: I think I should add ```Model (AutoGeneratedName = true)```here to use ```IAVRoutePickerViewDelegate```right ?

src/avkit.cs
```diff
@@ -518,23 +631,54 @@ interface AVKitMetadataIdentifier {
...
+	interface AVRoutePickerView
+	{
+#if !MONOMAC
 		[Export ("initWithFrame:")]
 		IntPtr Constructor (CGRect frame);
-
-		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
+#endif
+		[Wrap ("WeakDelegate")]
+		[NullAllowed]
```
joconte: Actually there is a breaking change because of this modification : https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/10962/API_20diff_20_28PR_20only_29/

What do you think about it?

src/avkit.cs
```diff
@@ -62,18 +64,31 @@ interface AVPictureInPictureController
...
 		UIImage CreateStopButton ([NullAllowed] UITraitCollection traitCollection);
+#endif
 	}
+	
+	interface IAVPictureInPictureControllerDelegate {}
 
-	[NoTV]
-	[Protocol, Model]
+	[iOS (9,0), Mac (10,15), NoTV]
+	[Protocol, Model (AutoGeneratedName = true)]
```
rolfbjarne: Setting `AutoGeneratedName` to true when it previously wasn't is a breaking change.

src/avkit.cs
```diff
@@ -309,6 +333,49 @@ interface AVPlayerViewControllerDelegate
...
+		UIViewController PreviousChannelInterstitialViewController (AVPlayerViewController playerViewController);
+	}
+	
+	[NoWatch, NoTV, NoMac, iOS (13,0)]
+	[Native]
+	public enum AVAudioSessionRouteSelection : long
+	{
+		None = 0,
+		Local = 1,
+		External = 2
```
rolfbjarne:
```suggestion
		External = 2,
```

src/avkit.cs
rolfbjarne: One solution here is to not change the `Delegate` property, and add `[Sealed]` to the new `WeakDelegate` property (it's only a problem to have two members with the same Objective-C selector they're both virtual):

```csharp
[NullAllowed, Export ("delegate", ArgumentSemenatic.Weak)]
[Sealed]
```

src/avkit.cs
rolfbjarne: No, don't add `AutoGeneratedName = true` to existing protocols, it's a breaking change.

spouliot:
build

src/avkit.cs
joconte: I just checked, it's fine : 
BCE57F5F-FB71-4C52-BB87-43B4C612D964

src/avfoundation.cs
joconte: I ran the introspection test and an error occurred, I change it to ``` iOS (13,0)```:

rolfbjarne:
build

src/avkit.cs
```diff
@@ -647,11 +647,6 @@ interface AVRoutePickerView
 #endif
 		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
 		IAVRoutePickerViewDelegate Delegate { get; set; }
-
-		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
-		[Sealed]
-		NSObject WeakDelegate { get; set; }
-
```
spouliot: Not sure that's correct. It's the `Delegate` that should be using `[Wrap]` over the `WeakDelegate`

joconte: Yes, before the changes I had this (not for Mac) : 
```C#
IAVRoutePickerViewDelegate Delegate { get; set; }
```
But I had to add it for Mac, sharpie gave me this : 
```C#
[Wrap ("WeakDelegate")]
[NullAllowed]
IAVRoutePickerViewDelegate Delegate { get; set; }

```

But this way I change it for iOS too. Maybe I should just put the weak delegate only for Mac ? Like this : 

```C#
#if !MONOMAC
IAVRoutePickerViewDelegate Delegate { get; set; }
#else
[Wrap ("WeakDelegate")]
[NullAllowed]
IAVRoutePickerViewDelegate Delegate { get; set; }

#endif
```

As I understood the use of ```Wrap``` allow us to wrap a weakly-typed object with a strongly-typed object. So It's "better" but I don't have to change the way It was for iOS.

spouliot: For a _short_ while we did not provide `WeakDelegate` properties but that decision was reversed :) We have not gone thru everything, but we're _fixing_ (adding) them as we update.

If done correctly it's not a breaking change (just verify the API diff to be sure).

spouliot:
build

joconte:
@VincentDondain Thanks a lot

PR #6751, opened 2019-08-11
[xcode11] [EventKit] bindings for all platforms

joconte:
Finally a lot of noise but not that much to change as I understood.

### DIFF
--- a/src/EventKit/EKEnums.cs
+++ b/src/EventKit/EKEnums.cs
@@ -75,15 +75,10 @@ namespace EventKit {
 		ProcedureAlarmsNotMutable,
 		EventStoreNotAuthorized,
 		OSNotSupported,
-		[iOS (13,0), Mac (10,15), Watch (6,0)]
 		InvalidInviteReplyCalendar,
-		[iOS (13,0), Mac (10,15), Watch (6,0)]
 		NotificationsCollectionFlagNotSet,
-		[iOS (13,0), Mac (10,15), Watch (6,0)]
 		SourceMismatch,
-		[iOS (13,0), Mac (10,15), Watch (6,0)]
 		NotificationCollectionMismatch,
-		[iOS (13,0), Mac (10,15), Watch (6,0)]
 		NotificationSavedWithoutCollection,
 	}
 
--- a/src/EventKit/EKEnums.cs
+++ b/src/EventKit/EKEnums.cs
@@ -75,6 +75,16 @@ namespace EventKit {
 		ProcedureAlarmsNotMutable,
 		EventStoreNotAuthorized,
 		OSNotSupported,
+		[iOS (13,0), Mac (10,15)]
+		InvalidInviteReplyCalendar,
+		[iOS (13,0), Mac (10,15)]
+		NotificationsCollectionFlagNotSet,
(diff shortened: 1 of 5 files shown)


src/EventKit/EKEnums.cs
```diff
@@ -75,6 +75,16 @@ public enum EKErrorCode : long {
...
+		[iOS (13,0), Mac (10,15)]
+		InvalidInviteReplyCalendar,
+		[iOS (13,0), Mac (10,15)]
+		NotificationsCollectionFlagNotSet,
+		[iOS (13,0), Mac (10,15)]
+		SourceMismatch,
+		[iOS (13,0), Mac (10,15)]
+		NotificationCollectionMismatch,
+		[iOS (13,0), Mac (10,15)]
+		NotificationSavedWithoutCollection,
```
joconte: I don't know which version of ```Watch```I should put.

spouliot:
build

src/EventKit/EKEnums.cs
spouliot: since this is added in Xcode 11 then it's the watchOS SDK version that was provided, i.e. 6.0

rolfbjarne:
build

src/eventkit.cs
```diff
@@ -508,16 +508,14 @@ interface EKEventStore {
...
-
-#if !MONOMAC
-		[NoWatch]
+		
+		[NoWatch, Mac (10,14)]
 		[Export ("saveEvent:span:error:")]
 		bool SaveEvent (EKEvent theEvent, EKSpan span, out NSError error);
 
-		[NoWatch]
+		[NoWatch, Mac (10,14)]
```
spouliot: Tests, run on 10.14, failed - that shows that this did not exists back then
Please update to `[Mac (10,15)]`

src/eventkit.cs
```diff
@@ -508,16 +508,14 @@ interface EKEventStore {
 
 		[Export ("defaultCalendarForNewEvents"), NullAllowed]
 		EKCalendar DefaultCalendarForNewEvents { get;  }
-
-#if !MONOMAC
-		[NoWatch]
+		
+		[NoWatch, Mac (10,14)]
```
spouliot: Tests, run on 10.14, failed - that shows that this did not exists back then
Please update to `[Mac (10,15)]`

src/eventkit.cs
joconte: Come on Apple.. ahah thanks

src/EventKit/EKEnums.cs
```diff
@@ -75,6 +75,16 @@ public enum EKErrorCode : long {
 		ProcedureAlarmsNotMutable,
 		EventStoreNotAuthorized,
 		OSNotSupported,
+		[iOS (13,0), Mac (10,15), Watch (6,0)]
```
dalexsoto: `EKErrorCode` is an error enum so it is unlikely that you will have to use its values as input so you can remove all the availability information (`[iOS (13,0), Mac (10,15), Watch (6,0)]`) from all the new enum members.

dalexsoto:
Build

spouliot:
Thank again for the contribution!

PR #6753, opened 2019-08-12
[xcode11][publishing] Added conditional behavior to CreateIpa and its dependencies

emaf:
Backport of https://github.com/xamarin/xamarin-macios/pull/6720, since publishing is part of the Xcode11 effort.

The targets execution needs to change based on the 'IsAppDistribution' property value, which will be set from the IDE.
According to the value of this property, some dependencies should be run or should be skipped.
The strategy adopted is to use the BeforeTargets value because doing this way we will force the conditionals over '$(IsAppDistribution)' to be evaluated each time a target is attempted to be run.
The most intuitive way of doing it is by directly add DependsOnTargets and conditions on the PropertyGroup items, but doing this will not work because the IDE usually runs the targets when the MSBuild project is already loaded, which means that the properties have already been evaluated so they will not be evaluated again, which causes conditional values to be out of date.
The only way of making it work in these cases is by forcing the conditional evaluations of each target execution attempt, and that is accomplished by adding the conditions directly on the Target 'Condition' property.

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
@@ -356,7 +356,7 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		</CreateAppBundleDependsOn>
 	</PropertyGroup>
 
-	<Target Name="_CreateAppBundle" Condition="'$(_CanOutputAppBundle)' == 'true'" DependsOnTargets="$(CreateAppBundleDependsOn)" />
+	<Target Name="_CreateAppBundle" Condition="'$(_CanOutputAppBundle)' == 'true' And '$(IsAppDistribution)' != 'true'" DependsOnTargets="$(CreateAppBundleDependsOn)" />
 
 	<PropertyGroup>
 		<_CodesignAppBundleDependsOn>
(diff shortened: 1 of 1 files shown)


msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
```diff
@@ -1772,8 +1802,16 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 		</CollectFrameworks>
 	</Target>
 
+	<PropertyGroup>
+		<_BeforeCodesignNativeLibrariesForBuildDependsOn>
+			_CompileToNative
+		</_BeforeCodesignNativeLibrariesForBuildDependsOn>
+	</PropertyGroup>
+
+	<Target Name="_BeforeCodesignNativeLibrariesForBuild" Condition="'$(IsAppDistribution)' != 'True'" BeforeTargets="_CodesignNativeLibraries" DependsOnTargets="$(_BeforeCodesignNativeLibrariesForBuildDependsOn)" />
```
rolfbjarne: It doesn't look like anything depends on `_BeforeCodesignNativeLibrariesForBuild`, is that on purpose?

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
```diff
@@ -376,34 +376,55 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
...
-		</CreateIpaDependsOn>
+		</_BeforeCreateIpaDependsOn>
+		
+		<_BeforeCreateIpaForBuildDependsOn>
+			$(_BeforeCreateIpaDependsOn)
+		</_BeforeCreateIpaForBuildDependsOn>
 
+		<_BeforeCreateIpaForDistributionDependsOn>
+			$(_BeforeCreateIpaDependsOn)
+		</_BeforeCreateIpaForDistributionDependsOn>
```
rolfbjarne: It seems both `_BeforeCreateIpaForBuildDependsOn` and `_BeforeCreateIpaForDistributionDependsOn` are identical, so doesn't this distinction just introduce unnecessary complexity? Or are there plans to make these different in the future?

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
emaf: No, nothing depends on it but it will always run before `_CodesignNativeLibraries` because of the `BeforeTargets` property.

msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
emaf: That's a really good point. The main purpose for those is to have specific extensibility points for both actions (create an IPA for distribution or as part of a full build). For instance, from a VS perspective, both are different situations, when creating an IPA for distribution (from an archive) we need to copy the xcarchive file to the Mac to then create the IPA from there, but when doing the same thing as part of a full build there's no archive involved since we'll use the app bundle generated by the build process.

I can add some comments on those property groups to clarify this if needed.

rolfbjarne: Yes, a comment or two would be nice, there are other changes needed as well because quite a few tests are failing.

emaf:
hey @rolfbjarne I've fixed all the issues and simplified Mauro's approach, could you take a look at the new changes?

PR #6755, opened 2019-08-12
[master] Bump mono to head of 2019-06 (4365dce)

dalexsoto:
Diff: https://github.com/mono/mono/compare/17ac7dcfc74e7371b99a1b7bf73d17e78df90c55...4365dce8b7dfdb42837bfd11bb61284b3477a45e


dalexsoto:
- A bunch of Known failure: HE0038
- All other did not fail on previous builds so it seems that connection was lost

PR #6759, opened 2019-08-13
[fileprovider[ui]] Update for Xcode 11 beta 1 to 5

spouliot:
Moved some code from uikit.cs since the type moved a while ago. That
ease code sharing with macOS (XM) but it stays into the UIKit namespace
(for XI) until `XAMCORE_4_0` to ensure binary compatibility.


src/foundation.cs
```diff
@@ -4132,19 +4132,27 @@ interface NSError : NSSecureCoding, NSCopying {
...
+		[Mac (10,15)]
 		[Static]
 		[Export ("fileProviderErrorForNonExistentItemWithIdentifier:")]
 		NSError GetFileProviderError (string nonExistentItemIdentifier);
+
+		[iOS (13,0)]
+		[Mac (10,15)]
+		[Static]
+		[Export ("fileProviderErrorForOutOfDateItem:")]
+		NSError fileProviderErrorForOutOfDateItem (INSFileProviderItem updatedVersion);
```
rolfbjarne:
```suggestion
		NSError GetFileProviderErrorForOutOfDateItem (INSFileProviderItem updatedVersion);
```

Sadly there's already a `GetFileProviderError` with the same signature, so that won't work.

spouliot: yeah, copy-paste mistake from me - after that same signature error
it should have been `GetFileProviderErrorForOutOfDateItem` like you suggested

spouliot:
unrelated build failure
`make[1]: *** [.stamp-build-tools64] Error 2`

spouliot:
build

PR #6771, opened 2019-08-14
[contactsui] Update for Xcode 11 beta 1 to 5

spouliot:
Mark the `iconWithContact:` API as not available with Catalyst

Note that the category `UIApplicationShortcutIcon (ContactsUI)` was
inlined in UIKit's `UIApplicationShortcutIcon` so only that file needs
to be updated.


spouliot:
Unrelated xtro failure fixed with https://github.com/xamarin/xamarin-macios/pull/6776

PR #6773, opened 2019-08-14
[gamekit] Update for Xcode 11 beta 1 to 5

spouliot:
`gamePlayerID` and `teamPlayerID` are decorated as `[iOS (12,4)]...` since
the headers mention so in both Xcode 11 betas and the recent 10.3 (stable)
https://github.com/xamarin/xamarin-macios/wiki/GameKit-iOS-xcode103-final

The `enum GKError` has been unified (at some point) so it was simplified.


spouliot:
Unrelated xtro failure fixed with #6776

PR #6781, opened 2019-08-15
[xtro] check new enum members

joconte:
It's not working actually but it's a start. 

The problem I'm facing is sometimes the enum members doesn't have exactly the same name in Xamarin and new Apple lib, so I have false positive.

I tried to use this version by going into xtro folder and ```make``` but the ```!missing-enum-member!```appears in ```*.unclassified```files (until I ```make unclassified2todo```)

Here the ```name.Replace ("Code","")``` is because the enum members doesn't have the same name, I already see there is the same problem with ```Mask```. 

I also saw sometimes the enums members starts with the framework "short name" , e.g. : EventKit -> EKSomething, EKSomethingElse, etc.

### DIFF
--- a/tests/xtro-sharpie/EnumCheck.cs
+++ b/tests/xtro-sharpie/EnumCheck.cs
@@ -58,9 +58,9 @@ namespace Extrospection {
 				Log.On (framework).Add ($"!missing-enum! {name} not bound");
 				return;
 			} else {
-            	var enumMemberInPreviousLib = type.Fields.Select (e => name.Replace ("Code","") + e.Name);
-            	var enumMemberInNewAppleLib = decl.Values.Select (e => e.Name);
-            	var missingEnums = enumMemberInNewAppleLib.Except (enumMemberInPreviousLib);
+            	var enumMembersInXamarin = type.Fields.Select (e => name.Replace ("Code","") + e.Name);
+            	var enumMembersInNewAppleLib = decl.Values.Select (e => e.Name);
+            	var missingEnums = enumMembersInNewAppleLib.Except (enumMembersInXamarin);
             	if (missingEnums.Any ()) {
             		foreach (var missingEnum in missingEnums) {
             			Log.On (framework).Add ($"!missing-enum-member! {name} : {missingEnum.Replace (name.Replace ("Code",""),"")} not bound");
--- a/tests/xtro-sharpie/EnumCheck.cs
+++ b/tests/xtro-sharpie/EnumCheck.cs
@@ -58,15 +58,17 @@ namespace Extrospection {
 				Log.On (framework).Add ($"!missing-enum! {name} not bound");
 				return;
 			} else {
-				var missingEnums = type.Fields.Select (e => e.Name).Except (decl.Values.Select (e => e.Name));
-				if (missingEnums.Any ()) {
-					foreach (var missingEnum in missingEnums) {
-						Log.On (framework).Add ($"!missing-enum-member! {name} : {missingEnum} not bound");
-					}
-					return;
-				}
-				enums.Remove (mname);
-			}
+            	var enumMemberInPreviousLib = type.Fields.Select (e => name.Replace ("Code","") + e.Name);
+            	var enumMemberInNewAppleLib = decl.Values.Select (e => e.Name);
+            	var missingEnums = enumMemberInNewAppleLib.Except (enumMemberInPreviousLib);
+            	if (missingEnums.Any ()) {
+            		foreach (var missingEnum in missingEnums) {
+            			Log.On (framework).Add ($"!missing-enum-member! {name} : {missingEnum.Replace (name.Replace ("Code",""),"")} not bound");
+            		}
+            		return;
+            	}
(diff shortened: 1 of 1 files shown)


dalexsoto:
@joconte IIRC all you have to do is to add said members to this file 

https://github.com/xamarin/xamarin-macios/blob/d24b652274c410ebd36ea3425afbe5c2ce912851/tests/xtro-sharpie/Helpers.cs#L23-L52

so Xtro is aware of those differences

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -57,8 +57,18 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
...
+            	var enumMembersInNewAppleLib = decl.Values.Select (e => e.Name);
+            	var missingEnums = enumMembersInNewAppleLib.Except (enumMembersInXamarin);
+            	if (missingEnums.Any ()) {
+            		foreach (var missingEnum in missingEnums) {
+            			Log.On (framework).Add ($"!missing-enum-member! {name} : {missingEnum.Replace (name.Replace ("Code",""),"")} not bound");
+            		}
+            		return;
+            	}
+            	enums.Remove (mname);
+            }
```
spouliot: Not critical but your indentation is incorrect :)

Now to be really useful the logic needs to be a bit more complex. In contrast to API you can always use a missing value, e.g. `Call ((MyEnum)42))`, so a missing member is not critical (as much as other things) since it's not blocking (just annoying). Also adding them later is not a breaking change...

The most important thing for enum members is their values, e.g. if we have `MissingMember = 43` while the right value is `42`. That's very hard to diagnose such errors (and easy to get wrong since, most of the time, the values are assigned automatically by the C# compiler). It's also a breaking change to fix them.

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -57,8 +57,18 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
 			if (!enums.TryGetValue (mname, out var type)) {
 				Log.On (framework).Add ($"!missing-enum! {name} not bound");
 				return;
-			} else
-				enums.Remove (mname);
+			} else {
+            	var enumMembersInXamarin = type.Fields.Select (e => name.Replace ("Code","") + e.Name);
```
spouliot: `name.Replace ("Code","")` should not be needed

The earlier logic has already figured out the managed and native name (and `Code` is only one of many difference that can happen).

spouliot:
build

tests/xtro-sharpie/EnumCheck.cs
joconte: Thanks for the feedback, I will try to add the value logic.

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -57,8 +57,18 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
...
+            	var enumMembersInNewAppleLib = decl.Values.Select (e => e.Name);
+            	var missingEnums = enumMembersInNewAppleLib.Except (enumMembersInXamarin);
+            	if (missingEnums.Any ()) {
+            		foreach (var missingEnum in missingEnums) {
+            			Log.On (framework).Add ($"!missing-enum-member! {name} : {missingEnum.Replace (name.Replace ("Code",""),"")} not bound");
+            		}
+            		return;
```
spouliot: I think the _new_ xtro errors are because of this `return` here.
In that case the `enums.Remove (mname);` is not executed and this throws off other results.

joconte: Yes but overall I think I shouldn't add new code in this file/class because I kinda break the Single Responsibility Principle. It was a fast edit for Proof of Concept. The ```EnumCheck```class is responsible for checking if enums are bound or not (and It's perfectly working). AFAIK the enum members (+ correct value) check should belong in another file/class, maybe in ```EnumMemberCheck```class ?

tests/xtro-sharpie/EnumCheck.cs
joconte: I'm having a hard time finding a way to get the enum member **value** of Apple lib. 
The enum is of type ```Clang.Ast.EnumDecl```and the enum members type is ```Clang.Ast.EnumConstantDecl```. I didn't find anything online plus I can't find it by debugging. If someone have an idea I would be happy to know about it :)

tests/xtro-sharpie/EnumCheck.cs
spouliot: Agreed :) but it means the PoC has not shown anything on the bots
Also you can do use GitHub to create _draft PR_ when things are not meant to be committed (not a big deal but it's make the goal clearer).

joconte:
@spouliot I'm kinda stuck because of this : https://github.com/xamarin/xamarin-macios/pull/6781#discussion_r314727049
I think yes you can close it and do what you just said : issue (enhancement).

Thanks for taking the time to ask.

PR #6784, opened 2019-08-15
[CoreML] Update bindings to Xcode 11 Beta 1 - 5

dalexsoto:


src/coreml.cs
```diff
@@ -82,6 +90,25 @@ public enum MLComputeUnits : long {
...
+		Failed = 5
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[Flags]
+	[Native]
+	public enum MLUpdateProgressEvent : long {
+		TrainingBegin = 1L << 0,
+		EpochEnd = 1L << 1,
+		MiniBatchEnd = 1L << 2
```
spouliot: minor `,`

src/coreml.cs
```diff
@@ -82,6 +90,25 @@ public enum MLComputeUnits : long {
...
+		Running = 2,
+		Cancelling = 3,
+		Completed = 4,
+		Failed = 5
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[Flags]
+	[Native]
+	public enum MLUpdateProgressEvent : long {
```
spouliot: `[Flags]` are generally unsigned - every bit counts :)

src/coreml.cs
```diff
@@ -209,6 +236,123 @@ interface MLFeatureValue : NSCopying {
...
 		[Export ("isEqualToFeatureValue:")]
 		bool IsEqual (MLFeatureValue value);
+
+		// From MLFeatureValue (MLImageConversion)
+
+		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[Static]
+		[Export ("featureValueWithImageAtURL:pixelsWide:pixelsHigh:pixelFormatType:options:error:")]
+		[return: NullAllowed]
+		MLFeatureValue Create (NSUrl url, nint pixelsWide, nint pixelsHigh, uint pixelFormatType, [NullAllowed] NSDictionary options, [NullAllowed] out NSError error);
```
spouliot: there's nothing else (than `uint` / `OSType`) to represent `pixelFormatType` ?

spouliot: looks like it's often `CVPixelFormatType`
```
src/CoreVideo/CVPixelBuffer.cs:                 /* OSType */ CVPixelFormatType pixelFormatType,
```

src/coreml.cs
```diff
@@ -584,6 +754,233 @@ interface MLModelConfiguration : NSCopying, NSSecureCoding {
...
+		[Export ("maxNumber")]
+		NSNumber MaxNumber { get; } // no better type found on docs nor swift
+
+		[NullAllowed, Export ("enumeratedNumbers")]
+		NSSet<NSNumber> EnumeratedNumbers { get; }
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (NSObject))]
+	interface MLParameterDescription : NSSecureCoding {
```
spouliot: sounds like a candidate for `[DisableDefaultCtor]` since there are no setters, only getters - what's the point of creating an instance ?

src/coreml.cs
```diff
@@ -584,6 +754,233 @@ interface MLModelConfiguration : NSCopying, NSSecureCoding {
...
+		[Export ("epochIndex")]
+		MLMetricKey EpochIndex { get; }
+
+		[Static]
+		[Export ("miniBatchIndex")]
+		MLMetricKey MiniBatchIndex { get; }
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (NSObject))]
```
spouliot: sounds like a candidate for `[DisableDefaultCtor]` since there are no setters, only getters - what's the point of creating an instance ?

src/coreml.cs
```diff
@@ -584,6 +754,233 @@ interface MLModelConfiguration : NSCopying, NSSecureCoding {
...
+		NSDictionary<MLParameterKey, NSObject> Parameters { get; }
+	}
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface MLUpdateProgressHandlers {
+
+		[Export ("initForEvents:progressHandler:completionHandler:")]
+		IntPtr Constructor (MLUpdateProgressEvent interestedEvents, [NullAllowed] Action<MLUpdateContext> progressHandler, Action<MLUpdateContext> completionHandler);
```
spouliot: does `[Async]` make sense ?

src/coreml.cs
```diff
@@ -584,6 +754,233 @@ interface MLModelConfiguration : NSCopying, NSSecureCoding {
...
+
+	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (MLTask))]
+	[DisableDefaultCtor]
+	interface MLUpdateTask {
+
+		[Static]
+		[Export ("updateTaskForModelAtURL:trainingData:configuration:completionHandler:error:")]
+		[return: NullAllowed]
+		MLUpdateTask Create (NSUrl modelUrl, IMLBatchProvider trainingData, [NullAllowed] MLModelConfiguration configuration, Action<MLUpdateContext> completionHandler, [NullAllowed] out NSError error);
```
spouliot: `[Async]` ?
I think our missing-async test will not spot such case since it's not the last parameter !?!
In fact I'm not sure if the generator would be happy...

src/coreml.cs
dalexsoto: Unfortunately this is a `ctor`

src/coreml.cs
dalexsoto: I do not think `[Async]` is needed here since this creates a Task object but its completion handler is not called when this task object is finished being created but when the task is completed

PR #6789, opened 2019-08-16
[quicklookthumbnailing] Update for Xcode11 beta 1 to 5

spouliot:
New framework - but it includes some of iOS API that were previously in
QuickLook.framework. Types were moved but remains in the old namespace
until `XAMCORE_4_0` is defined.


src/quicklookthumbnailing.cs
```diff
@@ -0,0 +1,169 @@
...
+
+	[ThreadSafe] // Members get called inside 'QLThumbnailProvider.ProvideThumbnail' which runs on a background thread.
+	[iOS (11,0)]
+	[Mac (10,15)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface QLThumbnailReply {
+		[Static]
+		[Export ("replyWithContextSize:drawingBlock:")]
+		QLThumbnailReply CreateReply (CGSize contextSize, Func<CGContext, bool> drawingBlock);
```
dalexsoto: Nitpick: 

```suggestion
		QLThumbnailReply CreateReply (CGSize contextSize, Func<CGContext, bool> drawingHandler);
```

src/quicklookthumbnailing.cs
```diff
@@ -0,0 +1,169 @@
...
+
+		[Static]
+		[Export ("replyWithContextSize:currentContextDrawingBlock:")]
+		QLThumbnailReply CreateReply (CGSize contextSize, Func<bool> drawingBlock);
```
dalexsoto: Nitpick:

```suggestion
		QLThumbnailReply CreateReply (CGSize contextSize, Func<bool> drawingHandler);
```

src/quicklookthumbnailing.cs
rolfbjarne: This is technically a source breaking change, and since this is really minor, I'd leave it as is.

src/quicklookthumbnailing.cs
rolfbjarne: Same.

spouliot:
While the suggested name is better it's existing code being moved, so a minor breaking change.

PR #6791, opened 2019-08-19
Update Versions-ios.plist.in with `altool` info

olegoid:
Needed for the IDEs to recognize if the SDK has support for App Store Connect upload/validation.


olegoid:
@rolfbjarne  good point 👍 Both XI and XM have the support of `altool` targets

PR #6802, opened 2019-08-20
[Tests] Allow to ignore tests that fail on the interpreter.

mandel-macaque:
There are some failures that only happen in the interpreter. We add a
TestRuntime method that allow to check if the interpreter is used.

The resources tests are shared between different solutions and the
TestRuntime is not present, so we have to do check in the test.

### DIFF
--- a/tests/EmbeddedResources/ResourcesTest.cs
+++ b/tests/EmbeddedResources/ResourcesTest.cs
@@ -32,17 +32,11 @@ namespace EmbeddedResources {
 		[Test]
 		public void Embedded ()
 		{
+
 #if __TVOS__
-			bool in_interpreter = false;
-			try {
-				AssemblyName aName = new AssemblyName ("DynamicAssemblyExample");
-				AssemblyBuilder ab = AppDomain.CurrentDomain.DefineDynamicAssembly (aName, AssemblyBuilderAccess.RunAndSave);
-				in_interpreter = true;
-			} catch (PlatformNotSupportedException) {
-				// we do not have the interpreter, lets continue
-			}
-			if (in_interpreter)
+			if (TestRuntime.CheckExecutingWithInterpreter ())
 				Assert.Ignore ("This test is disabled on TVOS."); // Randomly crashed on tvOS -> https://github.com/xamarin/maccore/issues/1909
+
 #endif
 
 #if MONOMAC
--- a/tests/EmbeddedResources/ResourcesTest.cs
+++ b/tests/EmbeddedResources/ResourcesTest.cs
@@ -11,6 +11,8 @@ using System;
 using System.IO;
 using System.Resources;
 using System.Globalization;
+using System.Reflection;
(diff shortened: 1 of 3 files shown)


tests/EmbeddedResources/ResourcesTest.cs
```diff
@@ -31,7 +33,16 @@ public class ResourcesTest {
 		public void Embedded ()
 		{
 #if __TVOS__
-			Assert.Ignore ("This test is disabled on TVOS."); // Randomly crashed on tvOS -> https://github.com/xamarin/maccore/issues/1909
+			bool in_interpreter = false;
+			try {
+				AssemblyName aName = new AssemblyName ("DynamicAssemblyExample");
+				AssemblyBuilder ab = AppDomain.CurrentDomain.DefineDynamicAssembly (aName, AssemblyBuilderAccess.RunAndSave);
```
spouliot: not a big deal but you can just do
```csharp
AppDomain.CurrentDomain.DefineDynamicAssembly (null, , AssemblyBuilderAccess.RunAndSave);
```
since it's default implementation to throw when REPL is not used (which the interpreter reuses)

tests/common/TestRuntime.cs
```diff
@@ -111,6 +113,24 @@ public static void IgnoreInCI (string message)
...
+			AssemblyName aName = new AssemblyName ("DynamicAssemblyExample");
+			AssemblyBuilder ab = AppDomain.CurrentDomain.DefineDynamicAssembly (aName, AssemblyBuilderAccess.RunAndSave);
+			in_interpreter = true;
+		} catch (PlatformNotSupportedException) {
+			// we do not have the interpreter, lets continue
+		}
+		if (!in_interpreter)
+			return;
+		NUnit.Framework.Assert.Ignore (message);
+	}
```
spouliot: that seems to differ from other checks...
we normally have a `bool Check...` to have conditional tests
and a `void Assert...` to ignore tests

it would also be nice if the `bool CheckInterpreter (...)` would be re-used in `tests/EmbeddedResources/ResourcesTest.cs`

mandel-macaque: @spouliot I followed the implementation used in the IgnoreInCI. I have no preference, maybe IgnoreInCI is outdated? If that is the case I can change both if needed.

mandel-macaque: TestRuntime is not available in the EmebededResources project, that is why I did the copy of the logic.

rolfbjarne: Did you try adding the TestRuntime.cs file to the EmbeddedResources project? That would be a better solution that duplicating the code.

spouliot: it's minor but can't we add/link the file to the project ?

I like the `Check`/`Assert` pattern better but that's also minor

mandel-macaque:

PR #6806, opened 2019-08-20
[tests] Adjust expected result of looking for 'mono_native_initialize'. Fixes xamarin/maccore#1809.

rolfbjarne:
Fixes https://github.com/xamarin/maccore/issues/1809.

### DIFF
--- a/tests/mono-native/Introspection.cs
+++ b/tests/mono-native/Introspection.cs
@@ -107,7 +107,9 @@ namespace Xamarin.Tests
 
 			try {
 #if MONOTOUCH_TV || MONOTOUCH_WATCH // on tvOS/watchOS we emit a native reference for P/Invokes in all assemblies, so we'll strip away the 'mono_native_initialize' symbol when we're linking statically (since we don't need the symbol).
-				var has_symbol = MonoNativeConfig.LinkMode != MonoNativeLinkMode.Static || Runtime.Arch == Arch.SIMULATOR;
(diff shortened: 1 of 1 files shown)


tests/mono-native/Introspection.cs
```diff
@@ -107,7 +107,9 @@ public void CheckSymbols ()
 
 			try {
 #if MONOTOUCH_TV || MONOTOUCH_WATCH // on tvOS/watchOS we emit a native reference for P/Invokes in all assemblies, so we'll strip away the 'mono_native_initialize' symbol when we're linking statically (since we don't need the symbol).
-				var has_symbol = MonoNativeConfig.LinkMode != MonoNativeLinkMode.Static || Runtime.Arch == Arch.SIMULATOR;
+				// The 'strip' tool that comes with Xcode 11 has some black magic where it removes the symbol from the executable, yet tvOS is still able to lookup the address from the symbol.
```
mandel-macaque: The comment says tvOS but we are inside a if MONOTOUCH_TV || MONOTOUCH_WATCH. Either the comment is wrong or this was added to WATCH too when it shouldn't.

rolfbjarne: I'll check this and create a new PR if something needs fixing.

rolfbjarne: These tests work fine on the watch, so no need to fix anything.

PR #6807, opened 2019-08-20
[UIKit] Update bindings to Xcode 11 Beta 6

dalexsoto:
Breaking changes are expected, Apple deprecated some methods that we better remove now.


src/uikit.cs
```diff
@@ -3284,14 +3284,14 @@ interface UIContextMenuInteractionDelegate {
...
 
-		[Export ("contextMenuInteractionWillPresent:")]
-		void WillPresent (UIContextMenuInteraction interaction);
+		[Export ("contextMenuInteraction:willDisplayMenuForConfiguration:animator:")]
+		void WillDisplayMenu (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, [NullAllowed] IUIContextMenuInteractionAnimating animator);
 
-		[Export ("contextMenuInteractionDidEnd:")]
-		void DidEnd (UIContextMenuInteraction interaction);
+		[Export ("contextMenuInteraction:willEndForConfiguration:animator:")]
+		void WillEndForConfiguration (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, [NullAllowed] IUIContextMenuInteractionAnimating animator);
```
spouliot: `ForConfiguration` seems redundant with the parameter. Swift uses overloads for all `contextMenuInteraction` so it's not really a good substitute for the name.

src/uikit.cs
```diff
@@ -3284,14 +3284,14 @@ interface UIContextMenuInteractionDelegate {
 		[return: NullAllowed]
 		UITargetedPreview GetPreviewForDismissingMenu (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration);
 
-		[Export ("contextMenuInteraction:willCommitWithAnimator:")]
-		void WillCommit (UIContextMenuInteraction interaction, IUIContextMenuInteractionCommitAnimating animator);
+		[Export ("contextMenuInteraction:willPerformPreviewActionForMenuWithConfiguration:animator:")]
+		void WillPerformPreviewActionForMenu (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```
VincentDondain: Why are we keeping all the `ForXXX` in this type? Isn't that referring to the first argument which is a `UIContextMenuInteraction`.

src/uikit.cs
VincentDondain: And I'm not quite sure about the other `ForXXX`

dalexsoto: Yeah I think plain `WillEnd` is ok

```suggestion
		void WillEnd (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, [NullAllowed] IUIContextMenuInteractionAnimating animator);
```

src/uikit.cs
```diff
@@ -4267,8 +4255,8 @@ interface UICollectionViewDelegate : UIScrollViewDelegate {
 		UITargetedPreview GetPreviewForDismissingContextMenu (UICollectionView collectionView, UIContextMenuConfiguration configuration);
 
 		[NoWatch, NoTV, iOS (13,0)]
-		[Export ("collectionView:willCommitMenuWithAnimator:")]
-		void WillCommitMenu (UICollectionView collectionView, IUIContextMenuInteractionCommitAnimating animator);
+		[Export ("collectionView:willPerformPreviewActionForMenuWithConfiguration:animator:")]
+		void WillPerformPreviewActionForMenu (UICollectionView collectionView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```
VincentDondain: Not sure the `ForMenu` adds much value here.

spouliot: agreed, it's not a direct `UIMenu` argument - but the signature makes it obvious

src/uikit.cs
dalexsoto: Agreed, it adds very little value

```suggestion
		void WillPerformPreviewAction (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```

src/uikit.cs
dalexsoto: Fixing!

```suggestion
		void WillPerformPreviewAction (UICollectionView collectionView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```

dalexsoto:
Thanks @VincentDondain and @spouliot I addressed your comments

src/uikit.cs
```diff
@@ -12620,8 +12608,8 @@ interface UITableViewSource {
 		UITargetedPreview GetPreviewForDismissingContextMenu (UITableView tableView, UIContextMenuConfiguration configuration);
 
 		[NoWatch, NoTV, iOS (13,0)]
-		[Export ("tableView:willCommitMenuWithAnimator:")]
-		void WillCommitMenu (UITableView tableView, IUIContextMenuInteractionCommitAnimating animator);
+		[Export ("tableView:willPerformPreviewActionForMenuWithConfiguration:animator:")]
+		void WillPerformPreviewActionForMenu (UITableView tableView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```
dalexsoto:
```suggestion
		void WillPerformPreviewAction (UITableView tableView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```

VincentDondain: @dalexsoto I think you can remove this ForMenu and the one below too

src/uikit.cs
```diff
@@ -13002,8 +12990,8 @@ interface UITableViewDelegate {
...
```
dalexsoto:
```suggestion
		void WillPerformPreviewAction (UITableView tableView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```

src/uikit.cs
```diff
@@ -3284,14 +3284,14 @@ interface UIContextMenuInteractionDelegate {
 		[return: NullAllowed]
 		UITargetedPreview GetPreviewForDismissingMenu (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration);
 
-		[Export ("contextMenuInteraction:willCommitWithAnimator:")]
-		void WillCommit (UIContextMenuInteraction interaction, IUIContextMenuInteractionCommitAnimating animator);
+		[Export ("contextMenuInteraction:willPerformPreviewActionForMenuWithConfiguration:animator:")]
+		void WillPerformPreviewAction (UIContextMenuInteraction interaction, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
```
VincentDondain: @dalexsoto and this animator is not `[NullAllowed]` somehow?

src/uikit.cs
```diff
@@ -12620,8 +12608,8 @@ interface UITableViewSource {
 		UITargetedPreview GetPreviewForDismissingContextMenu (UITableView tableView, UIContextMenuConfiguration configuration);
 
 		[NoWatch, NoTV, iOS (13,0)]
-		[Export ("tableView:willCommitMenuWithAnimator:")]
-		void WillCommitMenu (UITableView tableView, IUIContextMenuInteractionCommitAnimating animator);
+		[Export ("tableView:willPerformPreviewActionForMenuWithConfiguration:animator:")]
+		void WillPerformPreviewAction (UITableView tableView, UIContextMenuConfiguration configuration, IUIContextMenuInteractionCommitAnimating animator);
 
 		// WARNING: If you add more methods here, add them to UITableViewController as well.
```
VincentDondain: Is that warning needed? I know you didn't touch it but it's a scary warning

src/uikit.cs
dalexsoto: Yes, headers do not have `nullable` for that one

```objectivec
- (void)contextMenuInteraction:(UIContextMenuInteraction *)interaction willPerformPreviewActionForMenuWithConfiguration:(UIContextMenuConfiguration *)configuration animator:(id<UIContextMenuInteractionCommitAnimating>)animator;
```

src/uikit.cs
dalexsoto: Yup, that is needed so we do not forget to do that

src/uikit.cs
VincentDondain: Could just be missing in the headers but I guess it's minor let's not bother

src/uikit.cs
VincentDondain: Buuuuuut you didn't do that then

dalexsoto: I did but you are right, the WARNING is wrong since what you want is `UITableViewControllerDelegate`

dalexsoto:
```suggestion
		// WARNING: If you add more methods here, add them to UITableViewControllerDelegate as well.
```

PR #6824, opened 2019-08-21
[msbuild] Fixes publishing targets

emaf:
- Sets the missing SessionId property on the altool targets to enable running those from Visual Studio.
- Ensure Entitlements are compiled before creating an IPA. When creating an IPA from an archive the `CreateAppBundle` target will be skipped because we already have a bundle and the same will happen with the Entitlements compilation, but we need to re-compile those with the new certificate and profile. Adding `_CompileEntitlements` as a `CreateIpa` dependency will ensure it's run when creating an IPA file from an archive. Nothing changes when creating IPAs from a regular build since at the point the of creating the IPA the `_CompileEntitlements` target was already run, so it will be skipped.


emaf:
@rolfbjarne @spouliot is there any other approval needed to merge this PR?

spouliot:
@emaf no, if you have a green PR with approval from team member(s) then you can merge :)

Issue #6297, opened 2019-06-14
[Mono 2019-06][Interpreter][tvOS]ResourcesTest.Embedded : System.IO.FileNotFoundException : Invalid Image 


mandel-macaque:
Test fails when executed on a tvOS device when running with the interpreter and mono 2019-06.

marek-safar:
@mandel-macaque could you include some log info?

mandel-macaque:
The only info I have is the static trace of the failure:

> [FAIL] ResourcesTest.Embedded : System.IO.FileNotFoundException : Invalid Image
> 		  at System.Resources.ResourceManager+ResourceManagerMediator.GetResourceFileName (System.Globalization.CultureInfo culture) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/resources/resourcemanager.cs:1654 
> 		  at System.Resources.ManifestBasedResourceGroveler.GrovelForResourceSet (System.Globalization.CultureInfo culture, System.Collections.Generic.Dictionary`2[TKey,TValue] localResourceSets, System.Boolean tryParents, System.Boolean createIfNotExists, System.Threading.StackCrawlMark& stackMark) [0x00065] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/resources/manifestbasedresourcegroveler.cs:104 
> 		  at System.Resources.ResourceManager.InternalGetResourceSet (System.Globalization.CultureInfo requestedCulture, System.Boolean createIfNotExists, System.Boolean tryParents, System.Threading.StackCrawlMark& stackMark) [0x00099] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/resources/resourcemanager.cs:815 
> 		  at System.Resources.ResourceManager.InternalGetResourceSet (System.Globalization.CultureInfo culture, System.Boolean createIfNotExists, System.Boolean tryParents) [0x00002] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/resources/resourcemanager.cs:759 
[...]

PR #6832, opened 2019-08-22
[Tests] Pass a parameter to avoid a NRE.

mandel-macaque:
An example of the failure can be found here: http://xamarin-storage/jenkins/xamarin-macios/master/55a95cff879db23b19a84b4bf94a52b8b1135659/2964024/device-tests/jenkins-results/tests/index.html

```
[FAIL] ResourcesTest.Embedded : System.ArgumentNullException : Value cannot be null. 
```

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Reflection;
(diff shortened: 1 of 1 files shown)


mandel-macaque:
Ignoring until we get https://github.com/xamarin/xamarin-macios/issues/6297 fixed.

mandel-macaque:
@spouliot but the NRE catching.. could happen when the method is there too. Which we don't want. I'll make it static.

PR #6837, opened 2019-08-23
[Vision] Update bindings to Xcode 11 Beta 1 - 6

dalexsoto:


src/Vision/VNUtils.cs
```diff
@@ -61,6 +61,13 @@ public static CGPoint GetImagePoint (Vector2 faceLandmarkPoint, CGRect faceBound
 
 			return result;
 		}
+
+		[TV (13,0), Mac (10,15), iOS (13,0)]
+		[DllImport (Constants.VisionLibrary, EntryPoint = "VNElementTypeSize")]
+		static extern nuint _GetElementTypeSize (nuint elementType);
```
spouliot: why use `EntryPoint` ? it's an internal symbol so `VNElementTypeSize` should be fine right ?

src/vision.cs
```diff
@@ -251,6 +255,85 @@ enum VNBarcodeObservationRequestRevision : ulong {
...
+		Unspecified = 0,
+		One = 1,
+	}
+
+	[TV (13,0), Mac (10,15), iOS (13,0)]
+	[Native]
+	enum VNElementType : ulong {
+		Unknown = 0,
+		Float = 1,
+		Double = 2
```
spouliot: minor: missing `,`

src/vision.cs
```diff
@@ -1153,6 +1258,11 @@ interface VNFaceObservation {
 		[Export ("faceObservationWithRequestRevision:boundingBox:roll:yaw:")]
 		VNFaceObservation FromBoundingBox (VNFaceObservationRequestRevision requestRevision, CGRect boundingBox, [NullAllowed] [BindAs (typeof (nfloat?))] NSNumber roll, [NullAllowed] [BindAs (typeof (nfloat?))] NSNumber yaw);
 
+		[TV (13,0), Mac (10,15), iOS (13,0)]
+		[BindAs (typeof (nfloat?))]
```
spouliot: doc suggest it's `float?` not `nfloat?`
Not a big deal since it goes thru `NSNumber` but other API are also using `float`

src/Vision/VNUtils.cs
dalexsoto: Yeah, I have no answer but I just like the other name heh, anyways I'll change it :)

src/Vision/VNFeaturePrintObservation.cs
```diff
@@ -0,0 +1,26 @@
+//
+// VNFeaturePrintObservation.cs
+//
+// Authors:
+//	Alex Soto  <alexsoto@microsoft.com>
+//
+// Copyright 2019 Xamarin Inc. All rights reserved.
```
VincentDondain: Microsoft copyright (;

src/vision.cs
```diff
@@ -1,4 +1,4 @@
-﻿//
+//
```
VincentDondain: Dunno what happened here :P

src/vision.cs
```diff
@@ -144,6 +145,7 @@ enum VNDetectFaceLandmarksRequestRevision : ulong {
 		Unspecified = 0,
 		One = 1,
 		Two = 2,
+		Three = 3,
```
VincentDondain: Why isn't this getting an availability attribute?

src/vision.cs
```diff
@@ -59,7 +60,7 @@ enum VNRequestTrackingLevel : ulong {
 	enum VNImageCropAndScaleOption : ulong {
 		CenterCrop = 0,
 		ScaleFit = 1,
-		ScaleFill,
+		ScaleFill = 2,
```
VincentDondain: Why isn't this getting an availability attribute?

src/vision.cs
```diff
@@ -1775,5 +1916,393 @@ interface VNRequestRevisionProviding {
...
+
+	[TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (VNImageBasedRequest))]
+	[DisableDefaultCtor]
+	interface VNRecognizeTextRequest : VNRequestProgressProviding {
+
+		[Static]
+		[Export ("supportedRecognitionLanguagesForTextRecognitionLevel:revision:error:")]
+		[return: NullAllowed]
+		string [] GetSupportedRecognitionLanguages (VNRequestTextRecognitionLevel recognitionLevel, VNRecognizeTextRequestRevision requestRevision, [NullAllowed] out NSError error);
```
VincentDondain: `recognitionLevel` should be `textRecognitionLevel` here

src/vision.cs
```diff
@@ -1775,5 +1916,393 @@ interface VNRequestRevisionProviding {
...
+	[TV (13,0), Mac (10,15), iOS (13,0)]
+	[BaseType (typeof (VNImageBasedRequest))]
+	[DisableDefaultCtor]
+	interface VNRecognizeAnimalsRequest {
+
+		[Static]
+		[Export ("knownAnimalIdentifiersForRevision:error:")]
+		[return: NullAllowed]
+		[return: BindAs (typeof (VNAnimalIdentifier []))]
+		NSString [] GetKnownAnimalIdentifiers (VNRecognizeAnimalsRequestRevision requestRevision, [NullAllowed] out NSError error);
```
VincentDondain: I think `requestRevision` should just be `revision`. It's the **revision** for `VNRecognizeAnimalsRequest`, a `VNRecognizeAnimalsRequestRevision` object.

src/vision.cs
VincentDondain: Also just `revision`

src/vision.cs
```diff
@@ -1775,5 +1916,393 @@ interface VNRequestRevisionProviding {
...
+	interface VNClassifyImageRequest {
+
+		[Export ("initWithCompletionHandler:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([NullAllowed] VNRequestCompletionHandler completionHandler);
+
+		[Static]
+		[Export ("knownClassificationsForRevision:error:")]
+		[return: NullAllowed]
+		VNClassificationObservation [] GetKnownClassifications (VNClassifyImageRequestRevision requestRevision, [NullAllowed] out NSError error);
```
VincentDondain: Just `revision` instead of `requestRevision`.

src/vision.cs
```diff
@@ -376,9 +467,18 @@ interface VNDetectFaceLandmarksRequest : VNFaceObservationAccepting {
 		[DesignatedInitializer]
 		IntPtr Constructor ([NullAllowed] VNRequestCompletionHandler completionHandler);
 
+		[TV (13,0), Mac (10,15), iOS (13,0)]
+		[Static]
+		[Export ("revision:supportsConstellation:")]
+		bool SupportsConstellation (VNDetectFaceLandmarksRequestRevision requestRevision, VNRequestFaceLandmarksConstellation constellation);
```
VincentDondain: `Revision` not `requestRevision`

src/vision.cs
dalexsoto: Oh I changed from UTF8 BOM to just plain UTF8

VincentDondain: Scandalo!

src/vision.cs
dalexsoto: Because it is not needed just matching headers value which is the same implicit value

src/vision.cs
dalexsoto: Nice catch

src/vision.cs
VincentDondain: Oh yea thought it was new like the one I commented above 👍

dalexsoto:
> Also you have "We must inline the following 5 properties" in a bunch of places. I believe we should have a test that enforces that propagation on all the children (we definitely have that for other cases like this).

I agree, just not part of this PR I'll file an enhancement bug for this, specially if we have a customer waiting for these bindings

dalexsoto:
@spouliot / @VincentDondain comments addressed 👍 Thanks for the feedback!

PR #6844, opened 2019-08-25
[VisionKit] Add Xcode 11 Beta 1 - 6 bindings

dalexsoto:


src/visionkit.cs
```diff
@@ -0,0 +1,66 @@
+//
+// Vision C# bindings
```
rolfbjarne:
```suggestion
// VisionKit C# bindings
```

PR #6848, opened 2019-08-26
Add a dummy x86_64 slice to all our native libraries that don't have one.

rolfbjarne:
Apple's notarization tool has a bug where they incorrectly flag Mach-O
binaries without an x86_64 slice, so make sure all our libraries have one.


rolfbjarne:
build

PR #6851, opened 2019-08-26
[xcode11] Update with Xcode 10.3 changes and fix intro on iOS 12.4

spouliot:
Xcode 10.3 was released over the summer with a very small subset
of the (already out) Xcode 11 betas API.

This PR fix some availability attributes and also ensure we can
run introspection tests successfully on an iOS 12.4 device.

### DIFF
--- a/src/videotoolbox.cs
+++ b/src/videotoolbox.cs
@@ -552,11 +552,11 @@ namespace VideoToolbox {
 		[Field ("kVTVideoEncoderSpecification_EncoderID")]
 		NSString EncoderID { get; }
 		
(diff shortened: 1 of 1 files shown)


src/networkextension.cs
```diff
@@ -290,6 +290,10 @@ interface NEFilterControlProvider
 		[NullAllowed, Export ("URLAppendStringMap", ArgumentSemantic.Copy)]
 		NSDictionary<NSString,NSString> UrlAppendStringMap { get; set; }
 	
+		[iOS (11,0)] // also in base type - but only on iOS13+ (so we rre-define it here)
```
rolfbjarne: Rrrrr!

PR #6852, opened 2019-08-26
[AVFoundation] Add support up to Xcode 11 beta 7.

mandel-macaque:
This adds most of the framework to watchOS and a few number of classes and methods from iOS to macOS.


src/AVFoundation/AVUrlAssetOptions.cs
```diff
@@ -67,6 +66,4 @@ public AVUrlAssetOptions (NSDictionary dictionary)
 		}
 #endif
 	}
-}
-
-#endif
+}
```
rolfbjarne: Missing newline at eof.

src/AVFoundation/Enums.cs
```diff
@@ -1009,4 +1071,46 @@ public enum AVAudioSessionPromptStyle : ulong
 		Short = 0x73687274, // 1936224884 - 'shrt'
 		Normal = 0x6e726d6c, //1852992876 - 'nrml'
 	}
+
+	[Watch (6,0), iOS (13,0), TV (13,0), Mac (10,15)]
+	[Native]
+	public enum AVSpeechSynthesisVoiceGender : long {
+		Unspecified,
+		Male,
+		Female
```
rolfbjarne: Missing trailing comma.

src/avfoundation.cs
```diff
@@ -6930,7 +6970,7 @@ interface AVMediaDataStorage
 		NSUrl URL { get; }
 	}
 
-	[Mac (10,10)]
+	[Mac (10,11), Watch (6,0), iOS (13,0), NoTV]
```
rolfbjarne: macOS 10.10 -> 10.11 doesn't look quite right.

src/avfoundation.cs
```diff
@@ -7142,13 +7183,13 @@ interface AVMutableMovieTrack_AVMutableMovieTrackTrackAssociations
...
 
-	[Mac (10,10)]
+	[Mac (10,10), NoTV, Watch (6,0), iOS (13,0)]
 	[BaseType (typeof(AVMovieTrack))]
 	[DisableDefaultCtor]
 	interface AVFragmentedMovieTrack
 	{
-#if !XAMCORE_4_0
-		[Mac (10, 10)]
+#if !XAMCORE_4_
```
rolfbjarne:
```suggestion
#if !XAMCORE_4_0
```

src/avfoundation.cs
```diff
@@ -7665,10 +7740,17 @@ interface AVAssetExportSession {
 		[Export ("directoryForTemporaryFiles", ArgumentSemantic.Copy), NullAllowed]
 		[Advice ("This property cannot be set after the export has started.")]
 		NSUrl DirectoryForTemporaryFiles { get; set; }
+
+		[TV (13,0), NoWatch, Mac (10,15), iOS (13,0)]
+		[Export ("estimateMaximumDurationWithCompletionHandler:")]
+		void EstimateMaximumDuration (Action<CMTime, NSError> handler);
```
rolfbjarne: `[Async]`?

src/avfoundation.cs
```diff
@@ -7665,10 +7740,17 @@ interface AVAssetExportSession {
...
+
+		[TV (13,0), NoWatch, Mac (10,15), iOS (13,0)]
+		[Export ("estimateOutputFileLengthWithCompletionHandler:")]
+		void EstimateOutputFileLength (Action<long, NSError> handler);
```
rolfbjarne: `[Async]`?

src/avfoundation.cs
```diff
@@ -8507,6 +8621,14 @@ interface AVCaptureDeviceInput {
 		[NoWatch, NoTV, NoMac, iOS (12, 0)]
 		[Export ("unifiedAutoExposureDefaultsEnabled")]
 		bool UnifiedAutoExposureDefaultsEnabled { get; set; }
+
+		[NoMac, iOS (13,0)]
+		[Export ("portsWithMediaType:sourceDeviceType:sourceDevicePosition:")]
+		AVCaptureInputPort[] GetPortsWithMediaType ([BindAs (typeof (AVMediaTypes))] [NullAllowed] NSString mediaType, [BindAs (typeof (AVCaptureDeviceType))][NullAllowed] NSString sourceDeviceType, AVCaptureDevicePosition sourceDevicePosition);
```
rolfbjarne:
```suggestion
		AVCaptureInputPort[] GetPorts ([BindAs (typeof (AVMediaTypes))] [NullAllowed] NSString mediaType, [BindAs (typeof (AVCaptureDeviceType))][NullAllowed] NSString sourceDeviceType, AVCaptureDevicePosition sourceDevicePosition);
```

src/avfoundation.cs
```diff
@@ -9290,14 +9431,21 @@ interface AVCaptureResolvedPhotoSettings
 		[NoWatch, NoTV, NoMac, iOS (12, 0)]
 		[Export ("redEyeReductionEnabled")]
 		bool RedEyeReductionEnabled { [Bind ("isRedEyeReductionEnabled")] get; }
+
+		[NoWatch, NoTV, NoMac, iOS (13,0)]
+		[Export ("dimensionsForSemanticSegmentationMatteOfType:")]
+		CMVideoDimensions GetDimensionsForSemanticSegmentationMatte ([BindAs (typeof (MatteType))]NSString semanticSegmentationMatteType);
```
rolfbjarne:
```suggestion
		CMVideoDimensions GetDimensions ([BindAs (typeof (MatteType))]NSString semanticSegmentationMatteType);
```

src/avfoundation.cs
```diff
@@ -13224,19 +13224,21 @@ interface AVPersistableContentKeyRequest {
 	[DisableDefaultCtor]
 	[BaseType (typeof (NSObject))]
 	interface AVContentKeyResponse {
+		[Internal]
```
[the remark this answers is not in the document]
VincentDondain: @mandel-macaque no it doesn't work, the old code is internal customers cannot use it.

If you don't trust me it's reported as a breaking change:

```
Namespace AVFoundation

Type Changed: AVFoundation.AVContentKeyResponse

Removed method:

	public static AVContentKeyResponse Create (Foundation.NSData fairPlayStreamingKeyResponseData);
```

https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/11124/API_20diff_20_28PR_20only_29/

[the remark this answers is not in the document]
mandel-macaque: Meh.. I'm not very happy, but that means that the enum cannot be optional, which is a PITA.

mandel-macaque:
Failing tests are due to https://github.com/xamarin/maccore/issues/1939

spouliot:
@mandel-macaque some errors are related to https://github.com/xamarin/maccore/issues/1939 but the time outs are not

they happened 22hours ago
https://github.com/xamarin/xamarin-macios/pull/6852#issuecomment-526352519
another 2 days ago
https://github.com/xamarin/xamarin-macios/pull/6852#issuecomment-525980765

src/avfoundation.cs
```diff
@@ -11229,7 +11229,8 @@ interface AVPlayerItem_AVPlaybackRestrictions {
 		void CancelPlaybackRestrictionsAuthorizationRequest ();
 	}
 	
-	[iOS (13,0)][Watch (6,0)]
+	// trick the generator to believe it is present in an earlier version of the OS.
```
spouliot: It seems that the type was already present
https://developer.apple.com/documentation/avfoundation/avplayeritemoutput?language=objc
and marking it as `[iOS (13,0)]` is what cause the crash on 32 bits
IOW the comment is wrong

but if it has been correct (new base types do happen) then all members would need to have updated availability attributes,

src/AVFoundation/AVTypes.cs
```diff
@@ -322,4 +322,13 @@ public struct AVSampleCursorChunkInfo {
...
 	}
+
+	[NoTV, NoWatch, NoiOS, Mac (10,15)]
+	[StructLayout (LayoutKind.Sequential)]
+	public struct AVSampleCursorAudioDependencyInfo {
+		[MarshalAs (UnmanagedType.I1)]
+		public bool IsIndependentlyDecodable;
+
+		public nint PacketRefreshCount;
+	}
```
spouliot: That's incorrect. This file is not processed by the generator so the `[No*]` attributes are not affecting the output.

IOW `AVSampleCursorAudioDependencyInfo` is inside `Xamarin.iOS.dll` (and `*.TVOS.dll` and `*.WatchOS.dll`) even if not available on those platforms (check API diff).

src/avfoundation.cs
```diff
@@ -7519,6 +7601,9 @@ interface AVAssetExportSession {
 		[Export ("progress")]
 		float Progress { get;  } // defined as 'float'
 
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'EstimateMaximumDuration' enum instead.")]
+		[Deprecated (PlatformName.TvOS, 13,0, message: "Use 'EstimateMaximumDuration' enum instead.")]
+		[Deprecated (PlatformName.MacOSX, 10,15, message: "Use 'EstimateMaximumDuration' enum instead.")]
 		[Export ("maxDuration")]
 		CMTime MaxDuration { get;  }
```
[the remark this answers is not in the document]
spouliot: @mandel-macaque ^^^ no action/reply here

src/avfoundation.cs
```diff
@@ -7552,10 +7642,29 @@ interface AVAssetExportSession {
...
+		[Field ("AVAssetExportPresetHEVCHighestQualityWithAlpha")]
+		NSString PresetHevchHighestQualityWithAlpha { get; }
+
 		// 5.0 APIs
 		[Export ("asset", ArgumentSemantic.Retain)]
 		AVAsset Asset { get; }
 
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'EstimateOutputFileLength' for more precise results.")]
+		[Deprecated (PlatformName.TvOS, 13,0, message: "Use 'EstimateOutputFileLength' for more precise results.")]
+		[Deprecated (PlatformName.MacOSX, 10,15, message: "Use 'EstimateOutputFileLength' for more precise results.")]
```
spouliot: same as earlier
add [NoWatch] here
there's no point in exposing the deprecated API in a new platform

src/avfoundation.cs
```diff
@@ -9183,6 +9373,7 @@ interface AVCaptureResolvedPhotoSettings
 		[Export ("flashEnabled")]
 		bool IsFlashEnabled { [Bind ("isFlashEnabled")] get; }
 
+		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'AVCaptureResolvedPhotoSettings.PhotoProcessingTimeRange' instead.")]
```
spouliot: add `[NoMac]` (no point in adding deprecated API) unless the new API is not available on macOS

src/avfoundation.cs
```diff
@@ -9642,274 +9877,319 @@ enum AVAuthorizationMediaType {
 		Audio,
 	}
 
-	[NoWatch]
```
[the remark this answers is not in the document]
spouliot: I'm not sure it's correct - the registrar will still see (and generate code for) this type and Apple could reject that. c.c. @rolfbjarne 

I think a better way is to have the type `[Static]` on watchOS (since the field is `static` binary code will remain compatible). That way there will not be trace of the type in native code.

_note: remove `#if` is nice - but it does not achieve the same result here_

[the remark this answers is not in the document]
spouliot: This is what **API diff** shows on watchOS

```
New Type AVFoundation.AVCaptureDevice

public class AVCaptureDevice : Foundation.NSObject, Foundation.INSObjectProtocol, ObjCRuntime.INativeObject, System.IDisposable, System.IEquatable<Foundation.NSObject> {
	// constructors
	protected AVCaptureDevice (Foundation.NSObjectFlag t);
	protected AVCaptureDevice (IntPtr handle);
	// properties
	public override IntPtr ClassHandle { get; }
	public static float LensPositionCurrent { get; }
}
```

src/avfoundation.cs
```diff
@@ -12645,49 +13068,64 @@ interface AVContentKeySession_AVContentKeyRecipients
 		IAVContentKeyRecipient[] GetContentKeyRecipients ();
 	}
 
-	[TV (10,2), Mac (10,12,4), iOS (10,3), NoWatch]
+	[TV (10,2), Mac (10,12,4), iOS (10,3), Watch (6,0)]
```
spouliot: that looks like the same situation - that type should not be exposed in watchOS, see
https://developer.apple.com/documentation/avfoundation/avcontentkeyrequest

tests/xtro-sharpie/watchOS-AVFoundation.ignore
```diff
@@ -17,3 +17,10 @@
 !unknown-field! AVAudioUnitTypeOfflineEffect bound
 !unknown-field! AVAudioUnitTypeOutput bound
 !unknown-field! AVAudioUnitTypePanner bound
+
+# added the class to expose a filed and match the other APIs.
+!unknown-type! AVCaptureDevice bound
+!unknown-type! AVContentKeyRequest bound
```
spouliot: ^^ happy to see xtro caught them (and that there's no more than the two I spotted)

Issue #6704, opened 2019-08-01
Notification Callback crashes because collection is not locked


SOFSPEEL:
Crash report from AppCenter is:
```
Application Specific Information:
*** Terminating app due to uncaught exception 'SIGABRT', reason: 'Collection was modified; enumeration operation may not execute.'

Xamarin Exception Stack:
System.InvalidOperationException: **Collection was modified**; enumeration operation may not execute.
  at System.Collections.Generic.Dictionary`2+ValueCollection+Enumerator[TKey,TValue].MoveNext () <0x101176360 + 0x0010c> in <4fffe1dab0494c34a2a213216932b318#a1561c6d5fc7357f1676f8d092d865e6>:0
  at System.Net.Http.NSUrlSessionHandler.BackgroundNotificationCb (Foundation.NSNotification obj) <0x101e1f270 + 0x00087> in <0da8ced3e38d42039920e99e08ca9321#a1561c6d5fc7357f1676f8d092d865e6>:0
  at Foundation.InternalNSNotificationHandler.Post (Foundation.NSNotification s) <0x1013bdec0 + 0x0002f> in <b28d782d99614fe0b872da2da418b7b8#a1561c6d5fc7357f1676f8d092d865e6>:0
  at (wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
  at UIKit.UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate) <0x10137d0c0 + 0x0002b> in <b28d782d99614fe0b872da2da418b7b8#a1561c6d5fc7357f1676f8d092d865e6>:0
  at UIKit.UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName) <0x10137d030 + 0x00043> in <b28d782d99614fe0b872da2da418b7b8#a1561c6d5fc7357f1676f8d092d865e6>:0
  at ACS.FundRaising.Application.Main (System.String[] args) <0x100e47f20 + 0x0004b> in <1463607bddc044fcbe415fbcafd2d18a#a1561c6d5fc7357f1676f8d092d865e6>:0
```

Note that I suspect the crash occurs because the following code in NsUrlSessionHandler is not locked:

https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSUrlSessionHandler.cs

So this:

```csharp
void BackgroundNotificationCb (NSNotification obj)
{
[...]

PR #6877, opened 2019-08-29
[Foundation] Ensure that the collection is not modified during the loop. #6704

mandel-macaque:
Collections should not be modified during the loop, this is bad
practice and was a side effect of the TrySetCanceled. Is better to
create a temp collection with all the sources and cancel each of them.

Fixes https://github.com/xamarin/xamarin-macios/issues/6704


mandel-macaque:

mandel-macaque:

PR #6881, opened 2019-08-30
[generator] Fix CS1522 warning building generated bindings

spouliot:
It's now possible to have an empty smart `enum` but this was generating
an empty switch statement that `csc` would warn us about, e.g.

```
build/{profile}/ASAuthorizationProviderAuthorizationOperation.g.cs(64,24): warning CS1522: Empty switch block
```

This fix the generation to skip the `switch` generation when no fields
are present in the enum.

### DIFF
--- a/src/generator-enums.cs
+++ b/src/generator-enums.cs
@@ -165,20 +165,23 @@ public partial class Generator {
 			print ("{");
 			indent++;
 			print ("IntPtr ptr = IntPtr.Zero;");
-			print ("switch (({0}) self) {{", underlying_type);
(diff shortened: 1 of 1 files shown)


spouliot:
Both failures are: `[FAIL] UrlSessionTest.CreateDataTaskAsync : CreateDataTask a ` and seems related to a network issue in the lab (another PR failed in the same test). It's definitively not related to the change.

spouliot:
NSUrlSession failures now tracked in https://github.com/xamarin/maccore/issues/1939

PR #6890, opened 2019-08-30
[tests] Change URL for UrlSessionTest.CreateDataTaskAsync

spouliot:
The current https://www.xamarin.com is being redirected to https://dotnet.microsoft.com/apps/xamarin but it takes a long time to completely load (not sure why) and that could be why we fail since we have a 30 seconds timeout - and it took more than that to (stop) load into safari.
    
Switching to https://www.microsoft.com/ is a lot faster and seems to fix the issue.
    
Reference:
https://github.com/xamarin/maccore/issues/1939


spouliot:
Not sure an old title/description was used in the PR ?!?

To be backported to open branches (xcode11) once fix is confirmed.

spouliot:

PR #6893, opened 2019-08-31
Remove lists.xamarin.com from Readme

akoeplinger:
The mailing list no longer exists.


mandel-macaque:
Ignore the red check mark.

PR #6897, opened 2019-09-03
[imageio] Update for Xcode11 beta 7

spouliot:


spouliot:
CGImageAnimation.h not included in ImageIO umbrella header
https://github.com/xamarin/maccore/issues/1941
https://feedbackassistant.apple.com/form-response/15197631 (issue submitting)

src/imageio.cs
```diff
@@ -306,6 +310,18 @@ interface CGImageProperties {
 		[Field ("kCGImagePropertyGIFUnclampedDelayTime")]
 		NSString GIFUnclampedDelayTime { get; }
 
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Field ("kCGImagePropertyGIFCanvasPixelWidth")]
+		NSString GifCanvasPixelWidth { get; }
```
rolfbjarne: All the other fields here use `GIF` and not `Gif`.

spouliot: yes, I'm a bit ambivalent between old stuff and current rules :)

spouliot: Code completion does not care about casing (it's still alphabetical order) so I tend toward keeping the new stuff following the newer rules (i.e. it's not inconsistent in the usage, only the look).

PR #6901, opened 2019-09-03
[Master] Bump mono for the latest mono fixes related to culture.

mandel-macaque:
Diff (Single commit): https://github.com/mono/mono/compare/6434153d8b7d20b96e510e43031c5968d4042dda..2c3aeaf3780de7392a0d3cbe4dcf86846eb4dffa

PR #6902, opened 2019-09-03
[d16-3] Bump mono for the latest mono fixes related to culture.

mandel-macaque:
Diff (single commit) https://github.com/mono/mono/compare/6434153d8b7d20b96e510e43031c5968d4042dda..2c3aeaf3780de7392a0d3cbe4dcf86846eb4dffa


dalexsoto:
@rolfbjarne I agree but is the hash that android needs, that said there are some other commits that we may or may not care about like `[utils] Don't use MAP_32BIT on Apple platforms, fixes crash with XCode 11 beta6` anyways the decision to merge this has not been made yet so until we get green lights on this we will not merge it.

dalexsoto:
build

dalexsoto:
I do not see all of those watch failures in its counterpart PR https://github.com/xamarin/xamarin-macios/pull/6901 so I think we got a bad test run, triggering this again to be really sure about this bump

mandel-macaque:
hmmm @rolfbjarne good catch

mandel-macaque:
Mac failure is known: https://github.com/xamarin/maccore/issues/1932

PR #6908, opened 2019-09-04
[CoreMidi] Update framework to the latests Xcode11 beta 7.

mandel-macaque:


src/CoreMidi/MidiServices.cs
```diff
@@ -1542,6 +1544,15 @@ public int Add (string name, bool embedded, nuint numSourceEndpoints, nuint numD
 			}
 		}
 
+		[TV (13,0), Mac (10,15), iOS (13,0)]
```
spouliot: why `[TV]` ? it was not reported (or updated) in the xtro files

mandel-macaque: But it i present in the headers, from the diff:

extern const CFStringRef    kMIDIPropertyNameConfigurationDictionary            API_AVAILABLE(macos(10.15), ios(13.0), tvos(13.0));

spouliot: Headers can lie ;-)
The framework is not available on tvOS, so the API can't be present.

PR #6911, opened 2019-09-04
[AVFoundation] Add constructor for AVCaptureSynchronizedDepthData to fix breaking change.

rolfbjarne:
AVCaptureSynchronizedDepthData's init method is not available, which means we
should make the corresponding binding obsolete. We've already removed the
bound constructor, because it made introspection crash, this is just the
second part.


rolfbjarne:
build

Issue #6762, opened 2019-08-13
HttpClient fails to reach any URL via mobile data


ArrowCase:
### Steps to Reproduce

1. Create a Xamarin.Forms app that makes an HttpClient SendAsync call to any URL and deploy it to a physical iOS device.
2. Disconnect from Wi-Fi and connect to mobile data (4G LTE in my case).
3. Make the HttpClient call.

### Expected Behavior

The call succeeds.

### Actual Behavior

```
System.Net.WebException: The Internet connection appears to be offline.
--->
Foundation.NSError-Exception: Error Domain=NSURLError-Domain Code=-1009
"The Internet connection appears to be offline."

UserInfo={_kCFStreamErrorCodeKey=50, NSUnderlyingError=0x281f164c0 {Error Domain=kCFErrorDomainCFNetwork Code=-1009 "(null)" UserInfo={_kCFStreamErrorCodeKey=50, _kCFStreamErrorDomainKey=1}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <29082FC1-D10D-41CE-A44E-D5D41503D0BB>.<1>, _NSURLErrorRelatedURLSessionTaskErrorKey=( "LocalDataTask <29082FC1-D10D-41CE-A44E-D5D41503D0BB>.<1>" ), NSLocalizedDescription=The Internet connection appears to be offline., NSErrorFailingURLStringKey=https://some.url/path, NSErrorFailingURLKey=https://some.url/path, _kCFStreamErrorDomainKey=1}
```

The request is not even being attempted, and instead this error is immediately returned, exactly the same behavior seen when the device is legitimately not connected to the Internet.

Obviously, I've verified that I do in fact have an Internet connection when testing this. Xamarin Essentials correctly reports that there is connectivity, and I can open Safari and browse the same websites that HttpClient is failing to reach.

### Environment

```
iPhone 7 Plus
iOS 12.3.1 (also seen reports from 12.2 and 12.4)
VS2019 16.3.1 P1 & P2
Xamarin.Forms 4.1.0.673156 (also seen with 4.0.0 and 4.2.0-pre)
```

This works fine on Android and does not work with any one of the three iOS HttpClient implementations.

Related: https://forums.xamarin.com/discussion/162993/http-request

rolfbjarne:
Thank you for your feedback!

[...]

PR #6916, opened 2019-09-05
[Foundation] Ensure that we allow celullar data by default until the user says otherwise. #6762

mandel-macaque:
The default value in the NSUrlSession is to allow cellular data. This
small change closes the issue, since users will not have an unexpected
result.

Later we need to provide a proper fix to allow the property to be
exposed AND use the value of the session.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6762

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -158,7 +158,7 @@ namespace Foundation {
 			// therefore, we do not have to listen to the notifications.
 			isBackgroundSession = !string.IsNullOrEmpty (configuration.Identifier);
 #endif
-
+			allowsCellularAccess = configuration.AllowsCellularAccess;
 			AllowAutoRedirect = true;
 
 			// we cannot do a bitmask but we can set the minimum based on ServicePointManager.SecurityProtocol minimum
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -280,7 +280,7 @@ namespace Foundation {
 			}
 		}
 
-		bool allowsCellularAccess;
+		bool allowsCellularAccess = true;
 
 		public bool AllowsCellularAccess {
 			get {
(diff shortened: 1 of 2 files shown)


src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -280,7 +280,7 @@ protected override void Dispose (bool disposing)
 			}
 		}
 
-		bool allowsCellularAccess;
+		bool allowsCellularAccess = true;
```
rolfbjarne: I think that honoring the NSUrlSessionConfiguration passed to the instance would be better.

In other words, add the following line here https://github.com/xamarin/xamarin-macios/blob/ef4f329fa813b09efabdd029213db571b1c02788/src/Foundation/NSUrlSessionHandler.cs#L151-L163:

```csharp
allowsCellularAccess = configuration.AllowsCellularAccess;
```

spouliot: agreed - the default `.ctor` calls this one (and that 2nd .ctor needs to consider it too)

spouliot: minor (might be postponed if nothing else needs change) but the `= true;` is not needed anymore (and gets over-written in all, current, cases).

mandel-macaque: I'll merge and will do in a new one.

ideveloperaz:
hello @mandel-macaque  I have the same issue when wifi off but only 3/4G on ... where I could get packages to fix that? Thanks

ideveloperaz:
Hi @mandel-macaque , @rolfbjarne  I wrote to you regarding that crash:  https://github.com/xamarin/xamarin-macios/issues/6704

and you gave me these packages: xamarin.mac-5.15.0.317.pkg  and xamarin.ios-12.15.0.317.pkg to fix that issue ... but after updating VS that included fixes in these  #6762 #6916 case ... I;m still having issue with #6704 ...

how to get all that fixes (6704, 6916, 6762) in one package? As I could see #6704 not included

please, merge them to a package, we are having issue with customers :-(

mandel-macaque:
@ideveloperaz sorry I missed the message because the PR was close, let mi find that for you.

ideveloperaz:
@mandel-macaque  thank you a lot. Will wait packages from you to include all that critical fixes for our app :  #6762 #6916 and #6704

PR #6919, opened 2019-09-05
[coreaudio] Update for Xcode 11 (up to beta 7)

spouliot:
CoreAudioTypes is _new_ but it's just some stuff that moved around,
however it now shows up separately in our API diff (and was quite
large because of the removal/addition caused by moving headers)

### DIFF
--- a/src/AVFoundation/Enums.cs
+++ b/src/AVFoundation/Enums.cs
@@ -387,22 +387,27 @@ namespace AVFoundation {
 
 	[Native]
 	// NSInteger - AVAudioSession.h
+	// typedef CF_ENUM(NSInteger, AVAudioSessionErrorCode) -> CoreAudioTypes.framework/Headers/AudioSessionTypes.h
 	public enum AVAudioSessionErrorCode : long {
 		None = 0,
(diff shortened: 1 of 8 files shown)


src/AVFoundation/Enums.cs
```diff
@@ -387,22 +387,27 @@ public enum AVAudioSessionInterruptionType : ulong  {
...
-		Unspecified = 0x77686174,
-		InsufficientPriority = 0x21707269,
-		[iOS (9,0)]
-		CodeResourceNotAvailable = 0x21726573
+		InsufficientPriority = 0x21707269, // '!pri'
+#if !XAMCORE_4_0
+		[Obsolete ("Use 'ResourceNotAvailable' instead.")]
+		CodeResourceNotAvailable = 0x21726573,
+#endif
+		ResourceNotAvailable = 0x21726573, // '!res'
```
mandel-macaque: Changing my mind about the previous review. It had a iOS (9,0) attr and it does no longer have. Is that correct>

spouliot: yes, we don't put availability attributes on error codes (they are output only)

PR #6922, opened 2019-09-05
Implement notarization.

rolfbjarne:


dalexsoto:
This was partially shut down but do not close yet, since who knows what we will run into tomorrow heh

rolfbjarne:
d16-3 is stable now, so I think we can safely say we don't need this anymore.

PR #6927, opened 2019-09-06
[d16-3] Implement notarization.

rolfbjarne:


rolfbjarne:
Test failures are both due to https://github.com/xamarin/maccore/issues/581.

PR #6928, opened 2019-09-06
Implement notarization.

rolfbjarne:

### DIFF
--- a/builds/.gitignore
+++ b/builds/.gitignore
@@ -23,3 +23,5 @@ targettv
 targetwatch
 watchbcl
 mono-ios-sdk-destdir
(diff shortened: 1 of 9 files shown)


rolfbjarne:
> Looks good - but won't we get this into master w/xcode11 merge ?

Yes, but we've been asked to do it anyway.

Also it does not merge cleanly, so doing only these changes is not a bad idea anyway.

rolfbjarne:
Test failures are unrelated

* Mono Mac OS X BCL tests group 2/Mac/Modern: https://github.com/xamarin/maccore/issues/1932

PR #6930, opened 2019-09-06
[ModelIO] Implement new API introduced with Xcode 11 b1-7.

rolfbjarne:
Diff: https://github.com/xamarin/xamarin-macios/wiki/ModelIO-iOS-xcode11-beta1


src/modelio.cs
```diff
@@ -2946,6 +2947,16 @@ interface MDLTransformMatrixOp : MDLTransformOp {
 		MDLAnimatedMatrix4x4 AnimatedValue { get; }
 	}
 
+	[iOS (12,0), Mac (10,14), TV (12,0)]
+	[BaseType (typeof (NSObject))]
+	interface MDLTransformOrientOp : IMDLTransformOp {
```
spouliot: the type is marked as 12.0
https://developer.apple.com/documentation/modelio/mdltransformorientop?language=objc
but all members are 13.0
https://developer.apple.com/documentation/modelio/mdltransformorientop/2890059-animatedvalue?language=objc
https://developer.apple.com/documentation/modelio/mdltransformorientop/2889963-name?language=objc

my guess is that the former was an error - but, in any case, the member are now incorrectly attributed

src/modelio.cs
```diff
@@ -2971,6 +2982,10 @@ interface MDLTransformStack : NSCopying, MDLTransformComponent {
 		[Export ("addMatrixOp:inverse:")]
 		MDLTransformMatrixOp AddMatrixOp (string animatedValueName, bool inverse);
 
+		[iOS (13,0), Mac (10,15), TV (13,0)]
+		[Export ("addOrientOp:inverse:")]
+		MDLTransformMatrixOp AddOrientOp (string animatedValueName, bool inverse);
+
```
spouliot: should not this return a `MDLTransformOrientOp` ?https://developer.apple.com/documentation/modelio/mdltransformstack/2890002-addorientop?language=objc

src/modelio.cs
```diff
@@ -3026,5 +3041,39 @@ interface MDLMatrix4x4Array : NSCopying {
 		[Export ("getDouble4x4Array:maxCount:")]
 		nuint _GetDouble4x4Array (IntPtr valuesArray, nuint maxCount);
 	}
+
+	[iOS (12,0), Mac (10,14), TV (12,0)]
+	[BaseType (typeof (MDLAnimatedValue))]
+	interface MDLAnimatedQuaternion {
```
spouliot: same situation - the type _documented_ as part of iOS 12.0
but all members are in 13.0
https://developer.apple.com/documentation/modelio/mdlanimatedquaternion/3237269-doublequaternionattime?language=objc

tests/monotouch-test/Asserts.cs
```diff
@@ -178,6 +178,14 @@ public static void AreEqual (Quaternion expected, Quaternion actual, string mess
 		Assert.AreEqual (expected.W, actual.W, message + " (W)");
 	}
 
+	public static void AreEqual (Quaternion [] expected, Quaternion [] actual, string message)
+	{
+		Assert.AreEqual (expected.Length, actual.Length, message + " array lengths");
```
spouliot: minor: will NRE is one (or both) is/are `null`

tests/monotouch-test/Asserts.cs
```diff
@@ -186,6 +194,22 @@ public static void AreEqual (Quaterniond expected, Quaterniond actual, string me
...
+	{
+		Assert.AreEqual (expected.X, actual.X, delta, message + " (X)");
+		Assert.AreEqual (expected.Y, actual.Y, delta, message + " (Y)");
+		Assert.AreEqual (expected.Z, actual.Z, delta, message + " (Z)");
+		Assert.AreEqual (expected.W, actual.W, delta, message + " (W)");
+	}
+
+	public static void AreEqual (Quaterniond [] expected, Quaterniond [] actual, string message)
+	{
+		Assert.AreEqual (expected.Length, actual.Length, message + " array lengths");
```
spouliot: minor: will NRE is one (or both) is/are `null`

PR #6932, opened 2019-09-06
[MetalPerformanceShaders] Update for Xcode 11 beta 1 to 7

praeclarum:
Only 1 API has been added, and it was actually added in iOS 12.2.

### DIFF
--- a/src/MetalPerformanceShaders/MPSDefs.cs
+++ b/src/MetalPerformanceShaders/MPSDefs.cs
@@ -526,7 +526,7 @@ namespace MetalPerformanceShaders {
 	{
 		Default = 0x0,
 		LowPower = 0x1,
-		SkipRemovable = 0x2
+		SkipRemovable = 0x2,
 	}
 }
 #endif
--- a/src/MetalPerformanceShaders/MPSDefs.cs
+++ b/src/MetalPerformanceShaders/MPSDefs.cs
@@ -517,5 +517,16 @@ namespace MetalPerformanceShaders {
 		GruOutputGateBiasTerms,
 		//Count, // must always be last, and because of this it will cause breaking changes.
 	}
+
+	[Flags]
+	[Native]
+	[Introduced (PlatformName.MacCatalyst, 13, 0)]
+	[TV (12,2), Mac (10,14,4), iOS (12,2)]
+	public enum MPSDeviceOptions : ulong
+	{
+		Default = 0x0,
+		LowPower = 0x1,
+		SkipRemovable = 0x2
+	}
 }
 #endif
--- a/src/MetalPerformanceShaders/MPSKernel.cs
+++ b/src/MetalPerformanceShaders/MPSKernel.cs
@@ -29,7 +29,7 @@ namespace MetalPerformanceShaders {
 			var h = MPSGetPreferredDevice (options);
 			if (h == IntPtr.Zero)
 				return null;
-			return new MTLDeviceWrapper (h, false);
+			return Runtime.GetINativeObject<IMTLDevice> (h, false);
 		}
 
 		internal unsafe static float [] GetTransform (IntPtr transform)
--- a/src/MetalPerformanceShaders/MPSKernel.cs
+++ b/src/MetalPerformanceShaders/MPSKernel.cs
@@ -20,6 +20,18 @@ namespace MetalPerformanceShaders {
 			return MPSSupportsMTLDevice (device == null ? IntPtr.Zero : device.Handle);
 		}
 
+		[DllImport (Constants.MetalPerformanceShadersLibrary)]
+		[Introduced (PlatformName.MacCatalyst, 13, 0)]
+		[TV (12,2), Mac (10,14,4), iOS (12,2)]
+		static extern /* id<MTLDevice> _Nullable */ IntPtr MPSGetPreferredDevice (MPSDeviceOptions options);
+		public static IMTLDevice GetPreferredDevice (MPSDeviceOptions options)
+		{
+			var h = MPSGetPreferredDevice (options);
+			if (h == IntPtr.Zero)
+				return null;
+			return new MTLDeviceWrapper (h, false);
+		}
+
 		internal unsafe static float [] GetTransform (IntPtr transform)
 		{
 			var t = (float*) transform;
--- a/src/MetalPerformanceShaders/MPSKernel.cs
+++ b/src/MetalPerformanceShaders/MPSKernel.cs
@@ -23,10 +23,10 @@ namespace MetalPerformanceShaders {
 		[DllImport (Constants.MetalPerformanceShadersLibrary)]
 		[Introduced (PlatformName.MacCatalyst, 13, 0)]
 		[TV (12,2), Mac (10,14,4), iOS (12,2)]
-		static extern /* id<MTLDevice> _Nullable */ IntPtr MPSGetPreferredDevice (MPSDeviceOptions options);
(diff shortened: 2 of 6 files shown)


src/MetalPerformanceShaders/MPSKernel.cs
```diff
@@ -20,6 +20,18 @@ public static bool Supports (IMTLDevice device)
 			return MPSSupportsMTLDevice (device == null ? IntPtr.Zero : device.Handle);
 		}
 
+		[DllImport (Constants.MetalPerformanceShadersLibrary)]
+		[Introduced (PlatformName.MacCatalyst, 13, 0)]
+		[TV (12,2), Mac (10,14,4), iOS (12,2)]
+		static extern /* id<MTLDevice> _Nullable */ IntPtr MPSGetPreferredDevice (MPSDeviceOptions options);
```
whitneyschmidt: Is it standard practice to leave the `/*...*/` in code?

src/MetalPerformanceShaders/MPSDefs.cs
```diff
@@ -517,5 +517,16 @@ public enum MPSRnnMatrixId : ulong {
...
+
+	[Flags]
+	[Native]
+	[Introduced (PlatformName.MacCatalyst, 13, 0)]
+	[TV (12,2), Mac (10,14,4), iOS (12,2)]
+	public enum MPSDeviceOptions : ulong
+	{
+		Default = 0x0,
+		LowPower = 0x1,
+		SkipRemovable = 0x2
```
mandel-macaque: Lets add a `,` to minimize the next diff.
```suggestion
		SkipRemovable = 0x2,
```

whitneyschmidt: style nit: add `,` after `SkipRemovable` to minimize diff noise if the enum is updated in the future :)

src/MetalPerformanceShaders/MPSKernel.cs
mandel-macaque: Yes. we do that as a reference for the next person.

praeclarum: The function just above it did, so I was copying that. I can go either way.

I hope we can get null annotations some day :-)

src/MetalPerformanceShaders/MPSDefs.cs
praeclarum: Will fix!

praeclarum:
Not sure why "[FAIL] MetalPerformanceShaders.MPSKernel.MPSGetPreferredDevice has a [Native] enum parameter in its signature: MetalPerformanceShaders.MPSDeviceOptions options" is a failure.

I can remove the `[Native]` from `MPSDeviceOptions` but would like someone to chime in whether that's the right way to go.

src/MetalPerformanceShaders/MPSKernel.cs
```diff
@@ -20,6 +20,18 @@ public static bool Supports (IMTLDevice device)
 			return MPSSupportsMTLDevice (device == null ? IntPtr.Zero : device.Handle);
 		}
 
+		[DllImport (Constants.MetalPerformanceShadersLibrary)]
+		[Introduced (PlatformName.MacCatalyst, 13, 0)]
+		[TV (12,2), Mac (10,14,4), iOS (12,2)]
+		static extern /* id<MTLDevice> _Nullable */ IntPtr MPSGetPreferredDevice (nuint options);
+		public static IMTLDevice GetPreferredDevice (MPSDeviceOptions options)
```
spouliot: You need to have the attribute on the public API too.

They are needed twice

1. the public API is what the IDE can use to tell people if an API is available in the OS version they target;

2. the internal/native API (like the p/invoke above) is needed by our tools (and tests) so we can ensure they API is _really_ available (because headers are full of _lies_)

src/MetalPerformanceShaders/MPSKernel.cs
```diff
@@ -20,6 +20,18 @@ public static bool Supports (IMTLDevice device)
 			return MPSSupportsMTLDevice (device == null ? IntPtr.Zero : device.Handle);
 		}
 
+		[DllImport (Constants.MetalPerformanceShadersLibrary)]
+		[Introduced (PlatformName.MacCatalyst, 13, 0)]
+		[TV (12,2), Mac (10,14,4), iOS (12,2)]
```
spouliot: It did not show in previous API diff, e.g. for Xcode 10.3 -> https://github.com/xamarin/xamarin-macios/wiki/xcode103-Bindings-Status

So it _might_ be available in earlier OS (it happens) but even more commonly it's a mistake header (soft _lie_) from Apple. Running the tests on older devices or simulator (if available for those versions) is the only way to be sure if the correct availability. IOW don't trust headers if they do not match the current SDK version :)

spouliot:
@praeclarum `[Native]` means it's an enum based on `NS[U]Integer`.

So natively it's 32bits on 32bits platforms/architectures and 64bits on 64bits.

That's a problem in managed code because we can only define one size for enums - so we make them `[u]long`.

Fixing that in managed code creates a problem to interop with native code, in particular with p/invokes. You need to ensure the native signature is correct (for both 32/64 variants) and that the managed caller(s) are casting the value to match.

src/MetalPerformanceShaders/MPSKernel.cs
praeclarum: 👍 Yep, I figured as much.

src/MetalPerformanceShaders/MPSKernel.cs
praeclarum: Yeah, it's odd. There are a lot of APIs, especially in MPSNeuralNetwork, that get this iOS 12.2 designation.

So what should I do for the attributes? Manually test? Or is there build step configuration for that version?

Also, Apple definitely didn't talk about these APIs until iOS 13, so I don't see any harm in fibbing a little and marking these iOS 13.

src/MetalPerformanceShaders/MPSKernel.cs
praeclarum: 👍

praeclarum:
@Rolf Can you give me an example of an API that uses a [Native] enum so that I can see how they handle the 32-bit/64-bit difference?

It seems like a lot of enums just strip it off.

src/MetalPerformanceShaders/MPSKernel.cs
spouliot: The easy way is to run the introspection tests on iOS 12.2 and see if it can find the selectors.

In doubt (e.g. does not work for protocols) when use iOS 13.0 (the current SDK version) and we can refine this later (it's not a breaking change) and, like you said, it won't cause much of a problem for most developers :)

spouliot:
@praeclarum the change that @dalexsoto pushed in your PR does what's needed already (and yes it's generally just casting - what we want are a correct p/invoke signature since we deal with different ABI across 32/64 and Intel/ARM processors).

praeclarum:
@rolfbjarne @dalexsoto Can you recommend how to proceed with this PR given that this branch has been released? Should I rebase and force push? Or start a new PR?

rolfbjarne:
@praeclarum I think it's best to rebase on the master branch and creating a new PR is the best way to go (alternatively rebase on the xcode11.3 branch, which will get the PR in a stable release faster, but the down side is that the code will get less tested before it's released).

praeclarum:
OK, will open a new PR against xcode11.3

PR #6935, opened 2019-09-07
[MPSCore] Update for Xcode 11 beta 1 to 7

praeclarum:
This updates all the APIs to beta7.

Notes:

* `MPSCommandBuffer` wraps an `IMTLCommandBuffer` - I think I got it right but please check (it's important).
* `MPSNDArray` is the big new data type that is similar to `MPSMatrix` and `MPSImage`.
* `MPSNDArray` has some methods that optionally return strides. I offer 3 versions of those functions: parameterless, IntPtr, and nint[].
* `MPSNDArray.WriteBytes` buffer is an IntPtr which isn't ideal. But NDArrays can be any
data type so the alternative is to write primitive type overloads. I'm putting the burden on the app developer to `fixed` their data.
* `MPSNDArrayAllocators` aren't really user definable but I wasn't able to get `NSCopying` and friends working without the `[Model]`
* This also removes `MetalPerformanceShaders` from `Dlfcn` as it was conflicting with the one defined in Libraries.g.cs


src/MetalPerformanceShaders/MPSKernel.cs
```diff
@@ -58,6 +58,15 @@ public static bool Supports (IMTLDevice device)
 	}
 
 #if !COREBUILD
+	public partial class MPSImage {
+
+		[DllImport (Constants.MetalPerformanceShadersLibrary)]
+		static extern MPSImageType MPSGetImageType (IntPtr image);
+
+		[iOS (13,0), TV (12,0), Mac (10,15)][Introduced (PlatformName.MacCatalyst, 13, 0)]
+		public MPSImageType ImageType => MPSGetImageType (Handle);
```
dalexsoto: mmm this reminds me a little of https://github.com/xamarin/xamarin-macios/issues/6530 (Which I still need to fix heh).
Will the property it be safe for the debugger?
what happens if Handle is invalid (`IntPtr.Zero`)?

src/metalperformanceshaders.cs
```diff
@@ -1775,6 +1784,84 @@ interface MPSTemporaryImage {
...
+		[Export ("initWithDevice:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (IMTLDevice device);
+	}
+
+	interface IMPSHeapProvider {}
+
+	[Protocol]
+	[BaseType (typeof(NSObject))]
+	[Model]
```
dalexsoto:
```suggestion
	[Model (AutoGeneratedName = true)]
```

For any new `Model` please add the `AutoGeneratedName = true`

src/metalperformanceshaders.cs
```diff
@@ -1775,6 +1784,84 @@ interface MPSTemporaryImage {
...
+	interface MPSHeapProvider
+	{
+		[Abstract]
+		[Export ("newHeapWithDescriptor:")]
+		[return: NullAllowed]
+		IMTLHeap NewHeapWithDescriptor (MTLHeapDescriptor descriptor);
```
dalexsoto: What about 

```suggestion
		IMTLHeap GetNewHeap (MTLHeapDescriptor descriptor);
```

src/metalperformanceshaders.cs
```diff
@@ -1775,6 +1784,84 @@ interface MPSTemporaryImage {
...
+
+		[Export ("initWithCommandBuffer:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (IMTLCommandBuffer commandBuffer);
+
+		[Export ("commitAndContinue")]
+		void CommitAndContinue ();
+
+		[Export ("prefetchHeapForWorkloadSize:")]
+		void PrefetchHeapForWorkloadSize (nuint size);
```
dalexsoto: What about

```suggestion
		void PrefetchHeap (nuint workloadSize);
```

src/metalperformanceshaders.cs
```diff
@@ -5877,6 +6007,154 @@ interface MPSKeyedUnarchiver : MPSDeviceProvider {
...
+
+		[Export ("reshapeWithShape:")]
+		void Reshape (NSNumber[] shape);
+	}
+
+	interface IMPSNDArrayAllocator { }
+
+	[Protocol]
+	[BaseType (typeof(NSObject))]
+	[Model]
```
dalexsoto:
```suggestion
	[Model (AutoGeneratedName = true)]
```

src/metalperformanceshaders.cs
```diff
@@ -5877,6 +6007,154 @@ interface MPSKeyedUnarchiver : MPSDeviceProvider {
...
+		// This is implemented using params and descriptorWithDataType:dimensionCount:dimensionSizes: in MPSNDArray.cs
+		// [Static, Internal]
+		// [Export ("descriptorWithDataType:dimensionSizes:", IsVariadic = true)]
+		// MPSNDArrayDescriptor Create (MPSDataType dataType, nuint dimension0, IntPtr varArgs);
+
+		[Export ("reshapeWithDimensionCount:dimensionSizes:")]
+		void Reshape (nuint numberOfDimensions, IntPtr dimensionSizes);
+
+		[Export ("reshapeWithShape:")]
+		void Reshape (NSNumber[] shape);
```
dalexsoto: could this be a more concrete type? do you happen to know? if you do then you can do something like this

```csharp
		void Reshape ([BindAs (typeof (nint []))] NSNumber[] shape);
```

BindAs docs https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/objective-c-libraries?tabs=macos#binding-nsvalue-nsnumber-and-nsstring-to-a-better-type

src/MetalPerformanceShaders/MPSKernel.cs
praeclarum: Fair :-) 

Yes it's safe because, ImageType is just an integer.

The argument is marked nonnull so I assume bad things if null is passed. FWIW, I don't know if any NSObject properties are protected from Zero Handles.

I think this is fine, but I also don't think this function will be used a lot in real code, so I won't push the property.

src/metalperformanceshaders.cs
praeclarum: Got it!

src/metalperformanceshaders.cs
praeclarum: Ugh, missed that one :-( Thanks!

src/metalperformanceshaders.cs
praeclarum: Thanks again!

src/metalperformanceshaders.cs
praeclarum: Ah, no I didn't know that, thanks!

Thing is, I *really* want `params` for these puppies and that's why I provided the `params` version in manual code. Is there an attribute to get `params`?

src/MetalPerformanceShaders/MPSNDArray.cs
```diff
@@ -0,0 +1,90 @@
+#if (XAMCORE_2_0 && !MONOMAC) && !XAMCORE_4_0
```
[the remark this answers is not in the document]
praeclarum: I'll be honest I have no idea what any of these defines they mean. I stole them from *MPSMatrixDescriptor.cs*.

MPS files use a combination of:

```
#if (XAMCORE_2_0 && !MONOMAC) && !XAMCORE_4_0
```

```
#if XAMCORE_2_0 || !MONOMAC
```

I would like to include every platform. So I'm guessing `XAMCORE_2_0 || !MONOMAC` is the one I want? Are those just guarding against the old APIs?

src/MetalPerformanceShaders/MPSNDArray.cs
```diff
@@ -0,0 +1,90 @@
...
+using Metal;
+using Foundation;
+
+namespace MetalPerformanceShaders {
+	public partial class MPSNDArrayDescriptor {
+
+		public unsafe static MPSNDArrayDescriptor Create (MPSDataType dataType, params nuint[] dimensionSizes)
+		{
+			fixed (nuint* p = dimensionSizes) {
+				return Create (dataType, (nuint)dimensionSizes.Length, (IntPtr)p);
```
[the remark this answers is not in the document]
praeclarum: Yep!

src/MetalPerformanceShaders/MPSNDArray.cs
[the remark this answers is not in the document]
rolfbjarne: Just remove the entire `#if ...` condition, as far as I can tell it should be fine.

praeclarum:
@rolfbjarne @dalexsoto Can you recommend how to proceed with this PR given that this branch has been released? Should I rebase and force push? Or start a new PR?

rolfbjarne:
@praeclarum I think it's best to rebase on the master branch and creating a new PR is the best way to go (alternatively rebase on the xcode11.3 branch, which will get the PR in a stable release faster, but the down side is that the code will get less tested before it's released).

praeclarum:
OK, will open a new PR against xcode11.3

PR #6943, opened 2019-09-09
[CoreHaptics] Add support for the new framework up to beta 7.

mandel-macaque:


src/CoreHaptics/Enums.cs
```diff
@@ -0,0 +1,144 @@
...
+		AudioPan,
+
+		[Field ("CHHapticEventParameterIDAudioBrightness")]
+		AudioBrightness,
+	}
+
+	[Mac (10, 15), iOS (13, 0)]
+	public enum CHHapticDynamicParameterId {
+		[Field ("CHHapticDynamicParameterIDHapticIntensityControl")]
+		HapticDynamicParameterIDHapticIntensityControl,
```
spouliot: typo -> `HapticIntensityControl`

src/CoreHaptics/Enums.cs
```diff
@@ -0,0 +1,144 @@
...
+		InvalidTime = -4840,
+		UnknownError = -4898,
+		MemoryError = -4899,
+	}
+
+	[Mac (10,15), iOS (13,0)]
+	[Native]
+	public enum CHHapticEngineFinishedAction : long {
+		StopEngine = 1,
+		LeaveEngineRunning = 2
```
spouliot: minor: end with `,` so future diffs are smaller :)

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+
+		[Export ("relativeTime")]
+		double RelativeTime { get; set; }
+
+		[Export ("controlPoints")]
+		CHHapticParameterCurveControlPoint[] ControlPoints { get; }
+
+		[Export ("initWithParameterID:controlPoints:relativeTime:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([BindAs (typeof (CHHapticDynamicParameterId))]NSString parameterID, CHHapticParameterCurveControlPoint[] controlPoints, double relativeTime);
```
spouliot: `parameterID` -> `parameterId`

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		[Export ("initWithAudioResourceID:parameters:relativeTime:")]
+		IntPtr Constructor (nuint resourceId, CHHapticEventParameter[] eventParams, double time);
+
+		[Export ("initWithAudioResourceID:parameters:relativeTime:duration:")]
+		IntPtr Constructor (nuint resourceId, CHHapticEventParameter[] eventParams, double time, double duration);
+	}
+
+	[Mac (10,15), iOS (13,0)]
+	[Protocol]
+	[BaseType (typeof(NSObject))]
```
spouliot: no `[BaseType]` if it's not a `[Model]` - that generates incorrect code (and I thought we had a test for that)

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		float MaxValue { get; }
+
+		[Abstract]
+		[Export ("defaultValue")]
+		float DefaultValue { get; }
+	}
+
+	interface ICHHapticDeviceCapability { } 
+
+	[iOS (13,0)]
```
spouliot: either `[NoMac]` or `[Mac (10,15)]` ?

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+
+		[Abstract]
+		[Export ("supportsAudio")]
+		bool SupportsAudio { get; }
+
+		// Protocols do not like BindAs yet we know is the enum CHHapticEventType
+		[Abstract]
+		[Export ("attributesForEventParameter:eventType:error:")]
+		[return: NullAllowed]
+		CHHapticParameterAttributes GetAttributes (string eventParameter, string type, [NullAllowed] out NSError outError);
```
spouliot: using `NSString` since the constant comparison will need to allocate otherwise

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+
+		// Protocols do not like BindAs yet we know is the enum CHHapticEventType
+		[Abstract]
+		[Export ("attributesForDynamicParameter:error:")]
+		[return: NullAllowed]
+		CHHapticParameterAttributes GetAttributes (string eventParameter, [NullAllowed] out NSError outError);
```
spouliot: same -> `NSString`

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	[NoMac, iOS (13,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface CHHapticEngine
+	{
+		[Static]
+		[Export ("capabilitiesForHardware")]
+		ICHHapticDeviceCapability CapabilitiesForHardware { get; }
```
spouliot: minor: `HardwareCapabilities` sounds more like .net api

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		NSDictionary Event { get; set; }
+		CHHapticEventType EventType { get; set; }
+		double Time { get; set; }
+		double EventDuration { get; set; }
+		NSString EventWaveformPath { get; set; }
+		NSArray EventParameters { get; set; }
+		NSDictionary Parameter { get; set; }
+		// we need to do a NSString because it can be a CHHapticEventParameterID or a CHHapticDynamicParameterID 
+		// which are two different enums. User will have to do the conversion
+		NSString ParameterId { get; set; }
```
spouliot: add an `[Advice]` attribute since this is quite useful to the developer

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		double ParameterValue { get; set; }
+		// we do not have docs or header information about the exact type
+		NSObject ParameterCurve { get; set; }
```
spouliot: `CHHapticParameterCurve` ?

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		NSObject ParameterCurveControlPoints { get; set; }
```
spouliot: `CHHapticParameterCurveControlPoint` ?

src/corehaptics.cs
spouliot: now since there's some doubt name it `Weak*` until we know for sure
that way we can add better bindings later

src/corehaptics.cs
spouliot: now since there's some doubt name it `Weak*` until we know for sure
that way we can add better bindings later

src/frameworks.sources
```diff
@@ -487,7 +487,12 @@ COREGRAPHICS_SOURCES = \
 	CoreGraphics/CGPDFStream.cs \
 	CoreGraphics/CGPDFString.cs \
 	CoreGraphics/CGShading.cs \
-	
+
+# CoreHaptics
+
+COREHAPTICS_API_SOURCES = \
+		CoreHaptics/Enums.cs \
```
spouliot: alignment

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+		[Export ("autoShutdownEnabled")]
+		bool AutoShutdownEnabled { [Bind ("isAutoShutdownEnabled")] get; set; }
+
+		[Export ("initAndReturnError:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([NullAllowed] out NSError error);
+
+		[Export ("initWithAudioSession:error:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([NullAllowed] AVAudioSession audioSession, [NullAllowed] out NSError error);
```
spouliot: should this be the only API marked with `[NoMac]` instead of the whole type ?

src/corehaptics.cs
```diff
@@ -0,0 +1,394 @@
...
+	
+using AVFoundation;
+using CoreFoundation;
+using Foundation;
+using ObjCRuntime;
+
+namespace CoreHaptics {
+
+#if MONOMAC
+	interface AVAudioSession {}
```
spouliot: add link to the open feedback/shadow issue

mandel-macaque: yep, I was going to add it in a comment. Will add it in the source.

src/corehaptics.cs
mandel-macaque: yes, it is plural nevertheless. But i don't want to risk it with an NSArray. No docs, better not to write a broken API.

src/corehaptics.cs
prmallic: What does the `[BindAs]` attribute do? Does it map the NSString of parameterID to CHHapticDynamicParameterId somehow? When do we usually use it?

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+		NSString ParameterKey { get; }
+
+		[Field ("CHHapticPatternKeyParameterID")]
+		NSString ParameterIdKey { get; }
+
+		[Field ("CHHapticPatternKeyParameterValue")]
+		NSString ParameterValueKey { get; }
+
+		[Field ("CHHapticPatternKeyParameterCurve")]
+		NSString WeakCHHapticParameterCurveKey { get; }
```
rolfbjarne: I think this could be just `WeakParameterCurveKey`.

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+
+		[Field ("CHHapticPatternKeyParameterCurveControlPoints")]
+		NSString WeakCHHapticParameterCurveControlPointsKey { get; }
```
rolfbjarne: Same: `WeakParameterCurveControlPointsKey`

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+#endif
+
+	[Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface CHHapticEventParameter {
+		[BindAs (typeof (CHHapticEventParameterId))]
+		[Export ("parameterID")]
+		NSString ParameterId { get; }
+
```
prmallic: What does the `[BindAs]` attribute do? Does it like map `CHHapticEventParameterId` to `NSString` here? When do we use this?

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+		[Export ("initWithDictionary:error:")]
+		IntPtr Constructor (NSDictionary patternDict, [NullAllowed] out NSError outError);
+
+		[Wrap ("this (patternDefinition?.Dictionary, out outError)")]
+		IntPtr Constructor (CHHapticPatternDefinition patternDefinition, [NullAllowed] out NSError outError);
+
+		[Export ("exportDictionaryAndReturnError:")]
+		[return: NullAllowed]
+		NSDictionary<NSString, NSObject> _ExportDictionary ([NullAllowed] out NSError outError);
```
rolfbjarne: Exposed methods should not start with an underscore. Missing an `[Internal]`?

tests/introspection/iOS/iOSApiWeakPropertyTest.cs
```diff
@@ -53,8 +53,10 @@ protected override bool Skip (PropertyInfo property)
 			case "UIViewControllerPreviewingWrapper":
 				return property.Name == "WeakDelegate";
 #endif
-			}
-			return base.Skip (property);
+			case "CHHapticPatternDefinition": // special case for a strong dictionary
+				return property.Name == "WeakCHHapticParameterCurve" || property.Name == "WeakCHHapticParameterCurveControlPoints";
+            }
+            return base.Skip (property);
```
rolfbjarne: Space vs tab issue

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+
+		[Wrap ("new CHHapticPatternDefinition (_ExportDictionary (out outError))")]
+		CHHapticPatternDefinition ExportDictionary ([NullAllowed] out NSError outError);
```
spouliot: but here you're not exporting a dictionary :) -> `Export`

tests/introspection/iOS/iOSApiWeakPropertyTest.cs
spouliot: indentation is incorrect

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	[Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface CHHapticEngine
+	{
+		[Static]
+		[Export ("capabilitiesForHardware")]
+		ICHHapticDeviceCapability GetHardwareCapabilities ();
```
spouliot: it's indeed a method
```objc
/*! @method capabilitiesForHardware
 	@abstract
 		Get the protocol that describes haptic and audio capabilities on this device.
 	@discussion
 		Detailed description on the capability protocol is in CHHapticDeviceCapability.h.
 */
+ (id<CHHapticDeviceCapability>)capabilitiesForHardware;
```
https://developer.apple.com/documentation/corehaptics/chhapticengine/3081788-capabilitiesforhardware?language=objc

src/corehaptics.cs
spouliot: `WeakCHHapticParameterCurveKey` -> `WeakParameterCurveKey`

src/corehaptics.cs
spouliot: `WeakCHHapticParameterCurveControlPointsKey` -> `WeakParameterCurveControlPointsKey`

src/corehaptics.cs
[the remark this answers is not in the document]
mandel-macaque: Argh, I did dont want to expose it, I prefer it to be internal, and use the wrapper.

src/corehaptics.cs
mandel-macaque: Fair enough, ExportAsPatternDefinition?

src/corehaptics.cs
mandel-macaque: This is an internal class used as the holder of the key for the strong dictionary. This means that I'd have to add an extra [Export ("TheNewKeyName")] in  CHHapticPatternDefinition. For that matter, we might as well remove the Weak, call it ParameterCurveKey and leave it cleaner.

src/corehaptics.cs
mandel-macaque: Same as https://github.com/xamarin/xamarin-macios/pull/6943/files#r322459040

mandel-macaque: Same as https://github.com/xamarin/xamarin-macios/pull/6943/files#r322459040

tests/introspection/iOS/iOSApiWeakPropertyTest.cs
mandel-macaque: I have VSMac.

src/corehaptics.cs
spouliot: `[Export]` is enough, the IDE gives you all other information

src/corehaptics.cs
spouliot: not sure I understand

the problem is having the prefix/part of the type name after the `Weak`

src/corehaptics.cs
```diff
@@ -0,0 +1,398 @@
...
+		NSArray EventParameters { get; set; }
+		NSDictionary Parameter { get; set; }
+		// we need to do a NSString because it can be a CHHapticEventParameterID or a CHHapticDynamicParameterID 
+		// which are two different enums. User will have to do the conversion
+		[Advice ("Value can be either a CHHapticEventParameterId or a CHHapticDynamicParameterId. A check is needed against both enums.")]
+		NSString ParameterId { get; set; }
+		double ParameterValue { get; set; }
+		// we do not have docs or header information about the exact type
+		NSObject WeakCHHapticParameterCurve { get; set; }
+		NSObject WeakCHHapticParameterCurveControlPoints { get; set; }
```
spouliot: same as above: the `CHHaptic` text should **not** be part of the public name

if we add _strong_ versions later then the names should be `WeakFoo` and `Foo`, not `WeakCHHapticFoo` and `Foo`

src/corehaptics.cs
spouliot: Whenever `NSString` constants are used with enum-like macros or a subclass of `NSString` is used inside ObjC headers.

An API that accept `NSString` is not very useful (e.g. code completion) to _discover_ the right values. For such API we use a _smart_ enum that maps the members with constants. It's _smart_ because the mapping code is generated automatically (so you can't get it wrong by copy/pasting code or forgetting some entries...)

src/corehaptics.cs
mandel-macaque: That there is a better, way, let me push the diff, better with code

tests/introspection/iOS/iOSApiWeakPropertyTest.cs
spouliot: it's not the IDE it's the settings ;-)

tools/common/Frameworks.cs
```diff
@@ -188,6 +188,7 @@ public Framework Find (string framework)
 					{ "PencilKit", "PencilKit", 10,15 },
 					{ "Speech", "Speech", 10,15 },
 					{ "LinkPresentation", "LinkPresentation", 10,15 },
+					{ "CoreHaptics", "CoreHaptics", 10,15 },
```
rolfbjarne: You have a merge conflict here; when fixing it make sure to keep this list alphabetically ordered.

PR #6953, opened 2019-09-09
[LinkPresentation] Add the framework to simlauncher.

mandel-macaque:


spouliot:
we'll do a single PR for all missing frameworks

PR #6961, opened 2019-09-10
[generator] Handle inheriting the same member from multiple protocols. Fixes #6366.

### LINKED ISSUE #6366: Generated binding compile time error when conforming to 2 protocols with the same @required member
Since two protocols have the same required member we generate it twice - so the C# compiler is not happy.

AFAIK the two protocols can be used separately, so the definition should be in both, or together (and the declaration should be merged)

### Steps to Reproduce

1. Uncomment `bool Available { [Bind ("isAvailable")] get; }` in `NFCNdefTag` (corennfc.cs)
2. Build

### Expected Behavior

No error.

### Actual Behavior

```
error CS0102: The type 'NFCFeliCaTagWrapper' already contains a definition for 'Available'
```

### Environment

`xcode11`

### Example Project (If Possible)

See (upcoming) PR for `corenfc` support

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/issues/6366.

### DIFF
--- a/docs/website/generator-errors.md
+++ b/docs/website/generator-errors.md
@@ -196,6 +196,14 @@ This usually indicates a bug in Xamarin.iOS/Xamarin.Mac; please [file a bug repo
 
 ### <a name='BI1066'/>BI1066: Unsupported return type '{type}' in {method}.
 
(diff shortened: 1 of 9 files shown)


mandel-macaque:
Love it, I have seen an issue like that already!

PR #6962, opened 2019-09-10
[xharness] Bring more joy to the developer when all tests pass :tada:

VincentDondain:
- When everything just works, skip the "test results" header and output that all tests pass with a :tada: (:

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -1757,8 +1757,11 @@ namespace xharness
 			var buildingQueuedTests = allTasks.Where ((v) => v.Building && v.Waiting);
 
 			if (markdown_summary != null) {
-				markdown_summary.WriteLine ("# Test results");
-				markdown_summary.WriteLine ();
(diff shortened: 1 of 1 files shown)


VincentDondain:
Preview of what it will look like:

Screenshot 2019-09-10 10 51 08

VincentDondain:
Here's the *sad* version for reference (: 

Screenshot 2019-09-10 10 57 45

VincentDondain:
@mandel-macaque sounds like you're talking about an edge case where we'd print: " 🎉 All 0 tests passed 🎉 ". That would add a hint of sarcasm so why not (:

mandel-macaque:
@VincentDondain but that only happens in some of the tests, so you will never have the 🎉 All 0 tests passed 🎉, you will get the same number but you will not be able to expand that you got 0 in on of the tests.

@rolfbjarne am I missing something?

rolfbjarne:
@mandel-macaque if there were no passed test, `No tests selected.` should show up in the report, not `All 0 tests passed`.

VincentDondain:
@monojenkins backport to xcode11.1

PR #6974, opened 2019-09-11
[fileprovider][macos] Update (remove some) API for Xcode 11 GM

spouliot:
For some reason (likely to be added back later ?) Xcode 11 GM removed
most of new macOS 10.15 API for FileProvider.

So instead of deleting stuff this uses a lot of `[NoMac]` even if some
API are actually not part of any platform anymore, e.g. you can see the
following line in the GM headers

```
API_UNAVAILABLE(watchos, tvos) API_UNAVAILABLE(ios, macos)
```


spouliot:
msbuild failures are known issues filed as https://github.com/xamarin/xamarin-macios/issues/6970

PR #6981, opened 2019-09-11
[CoreBluetooth] Update framework for Xcode 11 GM.

mandel-macaque:


src/corebluetooth.cs
```diff
@@ -46,9 +46,12 @@ interface CBManager {
 		[Export ("state", ArgumentSemantic.Assign)]
 		CBManagerState State { get; }
 
+/* headers defer from iOS/tvOS/watchOS and MacOS. Feedback: https://github.com/xamarin/maccore/issues/1962 
+		[Static]
 		[iOS (13,0), TV (13,0), Watch (6,0), Mac (10, 15)]
 		[Export ("authorization", ArgumentSemantic.Assign)]
 		CBManagerAuthorization Authorization { get; }
+*/
```
spouliot: tricky - because it might be needed for apps

it looks iOS headers were not updated on purpose (at least it seems justifiable) as iOS 13.0 gets out before other OS (at least the GM was) and 13.1 might be out around the time others get released.

```
+@property(nonatomic, assign, readonly) CBManagerAuthorization authorization API_DEPRECATED_WITH_REPLACEMENT("CBManager.authorization", ios(13.0,13.1));
```

```
+@property(class, nonatomic, assign, readonly) CBManagerAuthorization authorization API_AVAILABLE(macos(10.15), ios(13.1), watchos(6.0), tvos(13.0));
```

spouliot: my guess is that iOS 13.0 won't ever have the static version of the API

but we can probably hack this to expose it (on iOS) as static for the managed API with a version check

dalexsoto: We could also file an issue in our repository and link it in release notes, that way we can add a workaround code so anyone really needing this can just bring this temporary code until we get the right picture from Apple.

mandel-macaque:
Closing since @spouliot took a better approach.

PR #6984, opened 2019-09-11
[AVFoundation] Add support up to Xcode 11 GM.

mandel-macaque:


src/avfoundation.cs
```diff
@@ -13469,4 +13583,34 @@ interface AVCaptureMultiCamSession {
 		[Export ("systemPressureCost")]
 		float SystemPressureCost { get; }
 	}
+
+	[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(AVMetadataObject))]
```
spouliot: add `[DisableDefaultCtor]` here since
* it's not used anywhere;
* the base type is abstract without an `init`
* we can remove it (it's not a breaking change) if we're wrong

src/avfoundation.cs
```diff
@@ -13469,4 +13583,34 @@ interface AVCaptureMultiCamSession {
...
+	interface AVMetadataBodyObject : NSCopying {
+		[Export ("bodyID")]
+		nint BodyId { get; }
+	}
+
+	[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(AVMetadataBodyObject))]
```
spouliot: same

src/avfoundation.cs
```diff
@@ -13469,4 +13583,34 @@ interface AVCaptureMultiCamSession {
...
+	interface AVMetadataCatBodyObject : NSCopying {
+	}
+
+	[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(AVMetadataBodyObject))]
```
spouliot: same

src/avfoundation.cs
```diff
@@ -13469,4 +13583,34 @@ interface AVCaptureMultiCamSession {
...
+	interface AVMetadataDogBodyObject : NSCopying {
+	}
+
+	[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(AVMetadataBodyObject))]
```
spouliot: same

src/avfoundation.cs
```diff
@@ -13469,4 +13583,34 @@ interface AVCaptureMultiCamSession {
...
+	interface AVMetadataHumanBodyObject : NSCopying {
+	}
+
+	[NoWatch, NoTV, Mac (10,15), iOS (13,0)]
+	[BaseType (typeof(AVMetadataObject))]
```
spouliot: same

PR #6986, opened 2019-09-11
[CoreHaptics] Expose protocol method that was ignored due to inheritance.

mandel-macaque:
After PR https://github.com/xamarin/xamarin-macios/pull/6961 we can now
do the right thing.

### DIFF
--- a/src/corehaptics.cs
+++ b/src/corehaptics.cs
@@ -198,7 +198,7 @@ namespace CoreHaptics {
 
 	[Mac (10,15), iOS (13,0)]
 	[Protocol]
-	interface CHHapticAdvancedPatternPlayer {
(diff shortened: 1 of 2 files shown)


src/corehaptics.cs
```diff
@@ -198,7 +198,7 @@ interface ICHHapticAdvancedPatternPlayer {}
 
 	[Mac (10,15), iOS (13,0)]
 	[Protocol]
-	interface CHHapticAdvancedPatternPlayer : CHHapticPatternPlayer {
```
dalexsoto: This is wrong :)

`protocol CHHapticAdvancedPatternPlayer<CHHapticPatternPlayer>`

spouliot:
build

PR #6987, opened 2019-09-11
[CoreHaptics] Remove the framework from Mac OS X until we know better.

mandel-macaque:


mandel-macaque:
@spouliot no need, it was never there, I forgot to comment this files. I knew about the other.

PR #6988, opened 2019-09-11
[corebluetooth] Update for Xcode 11 GM

spouliot:
This should let us provide a nicer API for the GM change about
`CBManager authorization` moving from an instance to a static
property (in all but iOS 13)

Alternative to https://github.com/xamarin/xamarin-macios/pull/6981

### DIFF
--- a/src/corebluetooth.cs
+++ b/src/corebluetooth.cs
@@ -46,9 +46,23 @@ namespace CoreBluetooth {
 		[Export ("state", ArgumentSemantic.Assign)]
 		CBManagerState State { get; }
 
-		[iOS (13,0), TV (13,0), Watch (6,0), Mac (10, 15)]
+#if IOS
+		[Internal]
+		[iOS (13,0)]
+		[Export ("authorization", ArgumentSemantic.Assign)]
+		CBManagerAuthorization _IAuthorization { get; }
+
+		[Internal]
+		[iOS (13,1)]
+		[Static]
+		[Export ("authorization", ArgumentSemantic.Assign)]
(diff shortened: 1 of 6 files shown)


mandel-macaque:
Much better, closing the other one.

src/corebluetooth.cs
```diff
@@ -46,9 +46,23 @@ interface CBManager {
...
+		[iOS (13,0)]
+		[Export ("authorization", ArgumentSemantic.Assign)]
+		CBManagerAuthorization _IAuthorization { get; }
+
+		[Internal]
+		[iOS (13,1)]
+		[Static]
+		[Export ("authorization", ArgumentSemantic.Assign)]
+		CBManagerAuthorization _SAuthorization { get; }
+#else
```
prmallic: Noob question 😅Does `[iOS (13,0)]` not mean anything 13+? Then are there two Exports for the same selector? How does it get resolved?

dalexsoto:
build

dalexsoto:
it is hung again...

dalexsoto:
build

src/corebluetooth.cs
spouliot: @prmallic good questions :)

Yes, `[iOS]` is a shortcut for `[Introduced (PlatformName.iOS, 13, 0)]` which means it's the first version where the API is available.

The _exported_ strings are identical but the selector is not. Static selector are prefixed with `+` so they can't conflict with the instance selectors of the same name. Our `[Export]` attribute hides that bit (it's implied with the `[Static]` attribute).

PR #7001, opened 2019-09-15
Add bindings for NSXpcConnection and related types

wjk:
Does what the title says, mostly. There was a previous PR that was also working on this task, but I cannot find it in the list now for some weird reason. This PR also adds MSBuild support for writing XPC services in C#. Note that since I cannot create new templates (since those are in the private md-addins repo), I have piggybacked off of the existing support for creating App Extensions.

This PR also contains changes to mmp that are required to convince Cocoa to accept a protocol written in C# as a valid XPC interface. (Apparently, dynamically registered protocols cannot be used as XPC interfaces; they must be compiled into the binary by clang to work.) I have tested this PR with [my test application](https://github.com/wjk/XamarinXpcSample), and it correctly passes parameters into and out of the bundled XPC service. This PR is now ready for review. Thanks so much!


dalexsoto:
Awesome ❤️ !! 

A small comment, would you be able to remove all the whitespace noise? Makes foundation.cs harder to review.

src/foundation.cs
```diff
@@ -15567,6 +15567,170 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		[Export ("interfaceWithProtocol:")]
+		[Static]
+		IntPtr CreateForProtocol(Protocol protocol);
+
+		[Export ("protool", ArgumentSemantic.Assign)]
+		Protocol Protocol { get; get; }
+
+		// FIXME: How do I hide the Selector type here?
+		[Export ("setClasses:forSelector:argumentIndex:ofReply:")]
+		void SetAllowedClassesForMethod (NSSet<Class> allowedClasses, Selector methodSelector, nuint argumentIndex, bool forReplyBlock);
```
Therzok: Disclaimer: not a corecontributor, so I may be completely wrong.

This could probably be done by including a custom method, similarly to how NSString2.cs and NSObject.mac.cs do with partial classes.

The custom bound method would hide the MethodInfo (or whatever we want exposed) being mapped to Selector

src/foundation.cs
```diff
@@ -15567,6 +15567,170 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+	[BaseType (typeof (NSObject), Name = "NSXPCInterface")]
+	[Mac (10, 8)][iOS (6, 0)][Watch (2, 0)][TV (9, 0)]
+	[DisableDefaultCtor]
+	interface NSXpcInterface
+	{
+		[Export ("interfaceWithProtocol:")]
+		[Static]
+		IntPtr CreateForProtocol(Protocol protocol);
+
+		[Export ("protool", ArgumentSemantic.Assign)]
```
Therzok: Typo: protocol

wjk:
@Therzok I added wrapper methods that take Type and MethodInfo instances, and extract the Objective-C metadata through the registration attributes. Look good?

rolfbjarne:
If you've bound the protocol, you should be able to do this:

```csharp
var obj = Runtime.GetINativeObject<IProtocol> (ptr, false);
```

and `obj` will be an instance you of an internal class that implements the protocol (so you can call all the API defined on that protocol with the `obj` object).

src/foundation.cs
```diff
@@ -16086,7 +16086,168 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		[NoiOS][NoWatch][NoTV]
+		IntPtr Constructor (string xpcServiceName);
+
+		[Export ("serviceName")]
+		string ServiceName { get; }
+
+		[Export ("initWithMachServiceName:")]
+		[NoiOS][NoWatch][NoTV]
+		[Internal] // See FIXME in Foundation/NSXpcConnection.cs
+		IntPtr Constructor (string machServiceName, nint options);
```
mandel-macaque: So, my question regarding this and the comment in NSXpcConnection.cs. Does it mean that you are trying to pass the enum to the constructor? You probably can do what you want using the BindAssAttribute on the parameter to go from nint to the enum. Else, you can always use the WrapAttribute to to the conversion.

wjk:
As of aedda49, `make` in the root directory succeeds with no issues. 🎉 I'll be starting on the MSBuild integration next. Thanks!

wjk:
@rolfbjarne I'm a bit confused about what you said about `Runtime.GetINativeObject<T>`. To get my wrapper methods to compile, I needed to add `where TProtocol : class, INativeObject`. As I understand it, this will bar use of the wrapper methods when passed the protocol type directly, since the interface type is not a "class". Will we still need the ObjC binding project, or am I missing something here? Thanks!

wjk:
As of c79a0f3, the build process successfully creates an XPC service and copies it into the app bundle, and all XPC-related types in the class library are ready to use. However, a test app (which I will be posting in a repo shortly) hangs within `NSXpcInterface.CreateForProtocol()`; I don't know why yet, but I suspect it is a binding bug. Any assistance in debugging this would be appreciated.

Also, @rolfbjarne, please disregard my previous comment. A binding project is in fact required, and passing the interface type into `Runtime.GetINativeObject()` works just fine. Thanks!

wjk:
The sample that I am using to test this PR is here: https://github.com/wjk/XamarinXpcSample

wjk:
And I've ran into a bit of bad news. Here's the exception string that I'm seeing thrown by `NSXpcInterface.CreateForProtocol()`:

```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason:
'NSXPCInterface: Unable to get extended method signature from Protocol data
(XamarinXpcProtocol / getHelloString:returnBlock:). Use of clang is required for NSXPCInterface.'
```

If we take the exception message at face value, there is no way to implement XPC in C#. 😢 However, I believe that Cocoa is simply querying for ObjC runtime data that we don't provide, and if we find a way to generate it, we can fool NSXPCInterface into thinking we were compiled with Clang.

msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
```diff
@@ -206,9 +207,11 @@ protected override string GenerateCommandLineCommands ()
 				foreach (var nr in NativeReferences)
 					args.AddQuotedLine ("/native-reference:" + Path.GetFullPath (nr.ItemSpec));
 			}
-
```
rolfbjarne: Whitespace noise.

msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.AppExtension.Common.targets
```diff
@@ -21,12 +21,12 @@ Copyright (C) 2013-2014 Xamarin. All rights reserved.
 
 	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Mac.Common.targets" />
 
-	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Mac.AppExtension.Common.props"
```
rolfbjarne: Whitespace noise here and a few other places in this file too.

msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
```diff
@@ -727,15 +729,25 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 
 	<Target Name="_CopyAppExtensionsToBundle"  Condition="'$(IsAppExtension)' != 'true'" DependsOnTargets="_ResolveAppExtensionReferences">
 		<MakeDir Directories="$(_AppBundlePath)Contents\PlugIns" Condition="'@(_ResolvedAppExtensionReferences)' != ''" />
+		<MakeDir Directories="$(_AppBundlePath)Contents\XPCServices" Condition="'@(_ResolvedAppExtensionReferences)' != ''" />
```
rolfbjarne: Would this mean every Xamarin.Mac app now ends up with an XPCServices directory?

src/Foundation/Enum.cs
```diff
@@ -1332,4 +1332,9 @@ public enum NSUrlSessionDelayedRequestDisposition : long {
 		UseNewRequest = 1,
 		Cancel = 2,
 	}
+
+	[Native]
+	public enum NSXpcConnectionOptions : ulong {
+		Privileged = (1 << 12)
```
rolfbjarne: We try to add a comma at the end of enum values (so that the diff only shows added lines if more values are added later):

```suggestion
		Privileged = (1 << 12),
```

src/Foundation/NSXpcConnection.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using ObjCRuntime;
+
+namespace Foundation {
+	public partial class NSXpcConnection : NSObject {
+		public TProtocol CreateRemoteObjectProxy<TProtocol> ()
+			where TProtocol : class, INativeObject
+		{
+			IntPtr nativeProxyPtr = _CreateRemoteObjectProxy ();
+			return (TProtocol)Runtime.GetINativeObject<TProtocol> (nativeProxyPtr, true);
```
rolfbjarne: The `(TProtocol)` cast should be unnecessary.

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
+//
+// Copyright 2009-2010, Novell, Inc.
+// Copyright 2011, 2012 Xamarin Inc
```
rolfbjarne: The copyrights aren't quite right :smile:

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using System;
+using System.Reflection;
+using ObjCRuntime;
+
+namespace Foundation {
+	public partial class NSXpcInterface : NSObject {
+		public static NSXpcInterface CreateForType (Type interfaceType)
+		{
+			if (interfaceType == null) throw new ArgumentNullException (nameof (interfaceType));
```
rolfbjarne: Style: put the `throw new ...` statement on the next line (indented).

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+			return CreateForProtocol (new Protocol (interfaceType));
+		}
+
+		public NSSet<Class> GetAllowedClassesForMethod (MethodInfo method, nuint argumentIndex, bool forReplyBlock)
+		{
+			if (method == null) throw new ArgumentNullException (nameof (method));
```
rolfbjarne: Style: put the `throw new ...` statement on the next line (indented).

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+
+			ExportAttribute attribute = method.GetCustomAttribute<ExportAttribute> ();
+			if (attribute == null)
+			{
```
rolfbjarne: We don't use braces if the if block is a single statement:

```csharp
if (attribute == null)
    throw new ...
```

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+				throw new ArgumentException ($"Method {method.Name} is not exposed to Objective-C", nameof (method));
+			}
+
+			// The runtime ensures that the Selector property is non-null and a valid selector.
+			Selector sel = new Selector(attribute.Selector);
```
rolfbjarne:
```suggestion
			Selector sel = new Selector (attribute.Selector);
```

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+			return GetAllowedClassesForMethod (sel, argumentIndex, forReplyBlock);
+		}
+
+		public void SetAllowedClassesForMethod (MethodInfo method, NSSet<Class> allowedClasses, nuint argumentIndex, bool forReplyBlock)
+		{
+			if (method == null) throw new ArgumentNullException (nameof (method));
```
rolfbjarne: Style: put the `throw new ...` statement on the next line (indented).

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+
+			ExportAttribute attribute = method.GetCustomAttribute<ExportAttribute> ();
+			if (attribute == null)
+			{
```
rolfbjarne: Style: remove braces.

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+				throw new ArgumentException ($"Method {method.Name} is not exposed to Objective-C", nameof (method));
+			}
+
+			// The runtime ensures that the Selector property is non-null and a valid selector.
+			Selector sel = new Selector(attribute.Selector);
```
rolfbjarne:
```suggestion
			Selector sel = new Selector (attribute.Selector);
```

src/Foundation/NSXpcInterface.cs
```diff
@@ -0,0 +1,67 @@
...
+			SetAllowedClassesForMethod(allowedClasses, sel, argumentIndex, forReplyBlock);
```
rolfbjarne:
```suggestion
			SetAllowedClassesForMethod (allowedClasses, sel, argumentIndex, forReplyBlock);
```

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		[Export ("resume")]
+		void Resume ();
+
+		[Export ("suspend")]
+		void Suspend ();
+
+		[Export ("invalidate")]
+		void Invalidate ();
+
+		// FIXME: How to bind auditSessionIdentifier?
```
rolfbjarne: `auditSessionIdentifier` is defined as a property whose type is `au_asid_t`: https://developer.apple.com/documentation/foundation/nsxpcconnection/1410393-auditsessionidentifier?language=objc
* `au_asid_t` is a `pid_t`: https://developer.apple.com/documentation/kernel/au_asid_t?language=objc
* `pid_t` is a `__darwin_pid_t`: https://developer.apple.com/documentation/kernel/pid_t?language=objc
* `__darwin_pid_t` is an `__int32_t`: https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/sys/_types.h#L72

which means that the property can be bound as an int:

```csharp
[Export ("auditSessionIdentifier")]
int AuditSessionIdentifier { get; }
```

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		int PeerEffectiveGroupId { get; }
+
+		[Export ("currentConnection")]
+		[Static]
+		[Mac (10, 8)][iOS (6, 0)][Watch (2, 0)][TV (9, 0)]
+		NSXpcConnection CurrentConnection { [return: NullAllowed] get; }
+
+		[Export ("scheduleSendBarrierBlock:")]
+		[Mac (10, 15)][iOS (13, 0)][Watch (6, 0)][TV (13, 0)]
+		void ScheduleSendBarrier(Action block);
```
rolfbjarne:
```suggestion
		void ScheduleSendBarrier (Action block);
```

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+	}
+
+	[BaseType (typeof (NSObject), Name = "NSXPCInterface")]
+	[Mac (10, 8)][iOS (6, 0)][Watch (2, 0)][TV (9, 0)]
+	[DisableDefaultCtor]
+	interface NSXpcInterface
+	{
+		[Export ("interfaceWithProtocol:")]
+		[Static]
+		NSXpcInterface CreateForProtocol (Protocol protocol);
```
rolfbjarne: I that just `Create` works fine:

```suggestion
		NSXpcInterface Create (Protocol protocol);
```

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+
+		[Export ("protocol", ArgumentSemantic.Assign)]
+		Protocol Protocol { get; set; }
+
+		[Export ("setClasses:forSelector:argumentIndex:ofReply:")]
+		[Internal]
+		void SetAllowedClassesForMethod (NSSet<Class> allowedClasses, Selector methodSelector, nuint argumentIndex, bool forReplyBlock);
```
rolfbjarne: I thinks is nicer:

```suggestion
		void SetAllowedClasses (NSSet<Class> allowedClasses, Selector methodSelector, nuint argumentIndex, bool forReplyBlock);
```

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+
+		[Export ("classesForSelector:argumentIndex:ofReply:")]
+		[Internal]
+		NSSet<Class> GetAllowedClassesForMethod (Selector methodSelector, nuint argumentIndex, bool forReplyBlock);
```
rolfbjarne: And the same here:

```suggestion
		NSSet<Class> GetAllowedClasses (Selector methodSelector, nuint argumentIndex, bool forReplyBlock);
```

src/frameworks.sources
```diff
@@ -1,6 +1,6 @@
 #
 # The source files used to build the platform assemblies:
-#
```
rolfbjarne: Whitespace noise (there are several others in this file too).

tools/mmp/driver.cs
```diff
@@ -1724,7 +1725,12 @@ static void CreateSymLink (string directory, string real, string link)
 				contents_dir = Path.Combine (App.AppDirectory, "Versions", "A");
 				macos_dir = contents_dir;
 			} else {
-				App.AppDirectory = Path.Combine (output_dir, string.Format ("{0}.{1}", app_name, is_extension ? "appex" : "app"));
+				string bundle_ext;
+				if (is_extension && is_xpc_service) bundle_ext = "xpc";
```
rolfbjarne: We format our `if`s like this:

```csharp
if (is_extension && is_xpc_service)
    bundle_ext = "xpc";
else if (is_extension)
    bundle_ext = "appex";
else
    bundle_ext = "app";
```

tools/mmp/driver.cs
```diff
@@ -327,6 +327,7 @@ static void Main2 (string [] args)
 				{ "tls-provider=", "Specify the default TLS provider", v => { tls_provider = v; }},
 				{ "http-message-handler=", "Specify the default HTTP Message Handler", v => { http_message_provider = v; }},
 				{ "extension", "Specifies an app extension", v => is_extension = true },
+				{ "xpc", "Specifies an XPC service. Use with --extension", v => is_xpc_service = true },
```
rolfbjarne: Maybe show an error if `--xpc` is passed, but not `--extension`?

src/generator.cs
```diff
@@ -5981,6 +5983,15 @@ void WriteMarkDirtyIfDerived (StreamWriter sw, Type type)
 		}
 	}
 
+	public void PrintXpcInterfaceAttribute (ICustomAttributeProvider mi)
+	{
+		var p = AttributeManager.GetCustomAttribute<XpcInterfaceAttribute> (mi);
```
rolfbjarne: If possible, use `AttributeManager.HasAttribute<XpcInterfaceAttribute> (mi)` instead, since it can be faster.

tools/common/StaticRegistrar.cs
```diff
@@ -2767,7 +2781,7 @@ void Specialize (AutoIndentStringBuilder sb)
 					if (token_ref == uint.MaxValue)
 						token_ref = CreateTokenReference (@class.Type, TokenType.TypeDef);
 					protocol_wrapper_map.Add (token_ref, new Tuple<ObjCType, uint> (@class, CreateTokenReference (@class.ProtocolWrapperType, TokenType.TypeDef)));
-					if (needs_protocol_map) {
```
rolfbjarne: Seems like it would be simpler to set `needs_protocol_map` to true like this somewhere further up:

```csharp
if (TryGetAttribute (@class.Type, "Foundation", "XpcInterfaceAttribute", out var xpcAttr))
    needs_protocol_map = true;
```

That would make the next change below redundant as well.

tools/common/StaticRegistrar.cs
```diff
@@ -496,8 +496,22 @@ public string ToObjCType (TypeDefinition type)
 			if (IsNativeObject (type))
 				return "id";
 
-			if (IsDelegate (type))
-				return "id";
+			if (IsDelegate (type)) {
+				if (!delegateToBlockType)
```
rolfbjarne: Did you try to always return the block signature instead of just `id` (i.e. skip the `delegateToBlockType` check and go straight to the code below? If it didn't work, do you remember why?

lemonmojo:
I'm very excited about this as I've been waiting for XPC support in Xamarin.Mac for a very long time.

One question: Will this allow building a standalone XPC service in C# which can then be embedded in a native ObjC/Swift app, including all necessary dependencies (Mono runtime, referenced DLLs, etc)?

tools/common/StaticRegistrar.cs
wjk: I tried always returning the block signature, but then I got build errors. The reason is this:

```objc
- (void)doSomething:(void (^)(BOOL))paramName; // This is valid syntax.
void doSomethingElse(void (^)(BOOL) paramName); // This is a syntax error.
void doSomethingElse(void (^paramName)(BOOL)); // This is valid syntax.
```

Without the delegateToBlockType check, we get the second output in the above example, and a broken build as a result. I will try threading the parameter name into the method so we get the third output, but I don't know if it will work. Thanks!

msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
wjk: As it currently exists, yes. However, AFAIK an empty XPCServices directory does no harm. I'll try adding `Condition="'%(_ResolvedAppExtensionReferences.Extension)' == '.xpc'"`.

wjk:
@lemonmojo As a matter of fact, yes! The way I've written it, a C# XPC service is packaged just like a standalone app: the Xamarin runtime and all required DLLs are embedded within the bundle.

lemonmojo:
@wjk Wow, awesome!
Is it functional right now so that I could already give it a spin? Or is there anything missing before it can be tested?

wjk:
It is ready for ad-hoc testing now, but there are a few quirks you need to know about:

* The static registrar *must* be enabled for all build configurations for all projects that reference `NSXpcInterface`, or else it will throw an ObjC exception and crash the process. NSXpcInterface refuses to work with protocols that were added at runtime using the dynamic registrar.
* You must decorate all protocols that are used with `NSXpcInterface` with `XpcInterfaceAttribute`, or else the Xamarin runtime will throw a C# exception. This attribute forces `mmp` to emit the ObjC protocol data into the binary for Cocoa to find later.
* Use only ObjC types in the XPC service interface, not C# ones. In particular, use `NSString` instead of `System.String`.
* Specify block callback types in your XPC interface as delegate types in the API definition, not as inline `Action<T>` delegates. If you don't do this, then Xamarin won't generate the necessary reverse block-to-delegate marshalling code required for the XPC service to make its callback.
* AFAIK, there is no way to attach the C# debugger to the XPC service process due to how it is launched. Furthermore, I have found that Xamarin ASL calls show up in the Console log garbled, probably due to an encoding mismatch. (Console logs in an XPC service are discarded by the system.) Even if this were practical, it would still require work from Microsoft as the Xamarin debugger isn't open-source.

Check out the test application linked from the OP, as it illustrates all of the above points. Hope this helps!

src/Foundation/NSXpcInterface.cs
wjk: What header do you want me to replace it with? I copy/pasted these as-is from another C# file. The copyright headers are all over the place in this repo.

tools/common/StaticRegistrar.cs
wjk: Unfortunately, this won't really work. If I were to move the parameter name into the type, I would then need to ensure that it is no emitted after the type as well [here](https://github.com/xamarin/xamarin-macios/blob/cd5917650e40cab7ed70cd75db3efe8220c60ba4/tools/common/StaticRegistrar.cs#L3988-L3989). I could do this, but then I'd have to implement an out-parameter from the ToObjCParameterType method to tell the caller that the parameter was already emitted, which IMHO isn't worth the increased complexity. Marshalling block types as `id` everywhere other than in Objective-C method declarations works just fine.

lemonmojo:
@wjk Tried compiling xamarin-macios from source with your changes but ultimately failed and gave up for now. Guess I'll be waiting desperately for the day this ends up in a Xamarin.Mac release.
Anyway, I wanted to thank you for getting this rolling! There are users like me out there waiting for this!

src/Foundation/NSXpcInterface.cs
rolfbjarne: Let me get back to you on that.

src/foundation.cs
```diff
@@ -16086,7 +16086,155 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+		[NoiOS][NoTV][NoWatch]
+		IntPtr Constructor (string machServiceName);
+
+		[Export ("delegate", ArgumentSemantic.Assign)]
+		[NullAllowed]
+		NSObject WeakDelegate { get; set; }
+
+		[Wrap ("WeakDelegate")]
+		[Protocolize]
+		NSXpcListenerDelegate Delegate { get; set; }
```
rolfbjarne: Instead of using `[Protocolize]`, create an `INSXpcListenerDelegate` interface and use that instead:

```csharp
interface INSXpcListenerDelegate {}

interface NSXpcListenerDelegate
{
    [Wrap ("WeakDelegate")]
    INSXpcListenerDelegate Delegate { get; set; }
}
```

src/foundation.cs
```diff
@@ -16086,7 +16086,155 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+
+		[Export ("suspend")]
+		void Suspend ();
+
+		[Export ("invalidate")]
+		void Invalidate ();
+	}
+
+	[BaseType (typeof (NSObject), Name = "NSXPCListenerDelegate")]
+	[Model, Protocol]
```
rolfbjarne: This is just a technically we have to do:

```suggestion
	[Model (AutoGeneratedName = true), Protocol]
```

rolfbjarne:
build

src/foundation.cs
wjk: I'm afraid I don't quite follow what you're asking me to do here. Could you please clarify your suggested change? Thanks!

rolfbjarne: 1. Remove the `[Protocolize]` line.

2. Change the property type to use I... instead:

```suggestion
		INSXpcListenerDelegate Delegate { get; set; }
```

3.  Declare a new interface (just like this, no members, etc):

```csharp
interface INSXpcListenerDelegate {}
```

VincentDondain:
build

VincentDondain:
@wjk Thanks for the contribution ❤️ I just wanted to add some info about the template part.

The Xamarin templates are indeed private but you might be able to play with https://github.com/dotnet/templating if you are interested in getting something going. This is the engine we use.

Otherwise, I can certainly create a template for you when this PR lands but having a small sample of what the template should look like would definitely help (: 

Probably a stripped down version of https://github.com/wjk/XamarinXpcSample

wjk:
@VincentDondain Check out [the template branch on my repo](https://github.com/wjk/XamarinXpcSample/tree/template). It contains a fully pared-down, ready-to-use `dotnet new` template for creating a Xamarin.Mac XPC service. I expect you'll be able to import this into md-addins as-is. If you have any problems, pleasae let me know. Thanks!

VincentDondain:
That is awesome thanks!

wjk:
I think I fixed the API availability test failures and the sharpie failures. The test failures about MM4105 not being emitted, I have no clue about. Some more eyes would be appreciated there. Thanks!

rolfbjarne:
The MM4105 test failure is probably related to the changes in the static registrar, because the test verifies that the 4105 error is shown with delegates in method signatures for methods exported to Objective-C: https://github.com/xamarin/xamarin-macios/blob/ea829b5a3fffe6ade0a6ac8720b431bff849f5c0/tests/common/BundlerTest.cs#L210-L219

The way to debug these is:

* Open tests/tests-mac.sln, build mmptest, open the Unit Test pad, and then you can drill down to the MX4105 test and run only that test.
* In the Test Results pad, you'll see the executed mmp command. It shows up like this:

    > /path/to/xamarin-macios/tools/xibuild/xibuild -t -- /path/to/xamarin-macios/_mac-build/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/mmp ...
* Copy this command (everything after '/path/to/.../mmp').
* Open Xamarin.Mac.sln, open the project options for the 'mmp' project, go to Run/Configurations/Default and paste into Arguments.
* Run the mmp project, and you'll run as the failing test did. You should be able to put breakpoints and debug using the debugger as normal.

wjk:
@rolfbjarne MX4105 test is working again. CI can be rerun at your convenience. Thanks!

rolfbjarne:
build

msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
```diff
@@ -209,6 +210,8 @@ protected override string GenerateCommandLineCommands ()
 				
 			if (IsAppExtension)
 				args.AddQuotedLine ("/extension");
+			if (IsXPCService)
```
mandel-macaque: No supper important, but we are mixing methods with the PR that use XPC with others in the bindings that use Xpc. I could choose one and be consistent.

src/Foundation/Enum.cs
```diff
@@ -1332,4 +1332,9 @@ public enum NSUrlSessionDelayedRequestDisposition : long {
 		UseNewRequest = 1,
 		Cancel = 2,
 	}
+
+	[Native]
+	public enum NSXpcConnectionOptions : ulong {
+		Privileged = (1 << 12),
+	}
```
mandel-macaque: Is this enum present in all platforms in the earliest version that we support?

src/Foundation/NSXpcConnection.cs
```diff
@@ -0,0 +1,51 @@
+//
+// Copyright 2009-2010, Novell, Inc.
+// Copyright 2011, 2012 Xamarin Inc
+//
```
mandel-macaque: Xamarin is no more.. and I wonder, is this a new file? Then Novell and Xamarin do not make much sense, right?

src/Foundation/NSXpcConnection.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using ObjCRuntime;
+
+namespace Foundation {
+	public partial class NSXpcConnection : NSObject {
+		public TProtocol CreateRemoteObjectProxy<TProtocol> ()
+			where TProtocol : class, INativeObject
+		{
+			IntPtr nativeProxyPtr = _CreateRemoteObjectProxy ();
```
mandel-macaque: Can this IntPtr ever be IntPtr.Zero?

src/Foundation/Enum.cs
wjk: The enum is marked as `API_AVAILABLE(macos(10.8), ios(6.0), watchos(2.0), tvos(9.0))` in the SDK header. I see no reason why it wouldn't be.

src/Foundation/NSXpcConnection.cs
wjk: It shouldn't be, because the return value of the `-[NSXPCConnection remoteObjectProxy]` method is marked as nonnull via the `NS_ASSUME_NONNULL_BEGIN` macro located above it in the header. If this method ever did return null, it would cause the Swift runtime serious problems, where nullable types and non-nullable types are completely different.

src/Foundation/Enum.cs
mandel-macaque: I needed the double check :)

rolfbjarne:
build

src/foundation.cs
```diff
@@ -16086,7 +16086,156 @@ interface NSMeasurementFormatter : NSSecureCoding {
...
+
+		[Export ("effectiveUserIdentifier")]
+		int PeerEffectiveUserId { get; }
+
+		[Export ("effectiveGroupIdentifier")]
+		int PeerEffectiveGroupId { get; }
+
+		[Export ("currentConnection")]
+		[Static]
+		[Mac (10, 8)][iOS (6, 0)][Watch (2, 0)][TV (9, 0)]
```
rolfbjarne: These availability attributes are redundant (the minimum OS versions Xamarin.iOS/Mac require are at least these versions, which means that this property will always be available) and should be removed.

It's also causing a test failure (in the introspection tests).

tests/xtro-sharpie/common-Foundation.ignore
```diff
@@ -834,41 +831,9 @@
...
-!missing-selector! NSXPCConnection::remoteObjectProxyWithErrorHandler: not bound
-!missing-selector! NSXPCConnection::serviceName not bound
-!missing-selector! NSXPCConnection::setExportedInterface: not bound
-!missing-selector! NSXPCConnection::setExportedObject: not bound
-!missing-selector! NSXPCConnection::setInterruptionHandler: not bound
-!missing-selector! NSXPCConnection::setInvalidationHandler: not bound
-!missing-selector! NSXPCConnection::setRemoteObjectInterface: not bound
-!missing-selector! NSXPCConnection::synchronousRemoteObjectProxyWithErrorHandler: not bound
-!missing-selector! NSXPCInterface::classesForSelector:argumentIndex:ofReply: not bound
-!missing-selector! NSXPCInterface::interfaceForSelector:argumentIndex:ofReply: not bound
```
rolfbjarne: The removal of this line is causing the xtro tests to fail.

tests/xtro-sharpie/common-Foundation.ignore
```diff
@@ -834,41 +831,9 @@
...
-!missing-selector! NSXPCInterface::protocol not bound
-!missing-selector! NSXPCInterface::setClasses:forSelector:argumentIndex:ofReply: not bound
-!missing-selector! NSXPCInterface::setInterface:forSelector:argumentIndex:ofReply: not bound
```
rolfbjarne: Same: this line should be put back.

tests/xtro-sharpie/macOS-Foundation.ignore
```diff
@@ -4,8 +4,6 @@
 ## XPC not supported
 !missing-selector! NSXPCCoder::decodeXPCObjectOfType:forKey: not bound
 !missing-selector! NSXPCCoder::encodeXPCObject:forKey: not bound
-!missing-selector! NSXPCInterface::setXPCType:forSelector:argumentIndex:ofReply: not bound
-!missing-selector! NSXPCInterface::XPCTypeForSelector:argumentIndex:ofReply: not bound
```
rolfbjarne: Same: these two lines should be put back.

rolfbjarne:
build

rolfbjarne:
So close!

The xtro tests still fail. This is for macOS:

    !extra-designated-initializer! NSXPCConnection::initWithServiceName: is incorrectly decorated with an [DesignatedInitializer] attribute
    !missing-designated-initializer! NSXPCListener::initWithMachServiceName: is missing an [DesignatedInitializer] attribute

so it looks like you put the `[DesignatedInitializer]` attribute on the wrong constructor.

and this is for iOS, tvOS and watchOS:

    !missing-protocol-conformance! NSXPCConnection should conform to NSXPCProxyCreating
    !missing-selector! NSXPCInterface::interfaceForSelector:argumentIndex:ofReply: not bound
    !missing-selector! NSXPCInterface::setInterface:forSelector:argumentIndex:ofReply: not bound

which means that you need to (re?-)add those lines to `[iOS|tvOS|watchOS]-Foundation.ignore`.

VincentDondain:
build

rolfbjarne:
@emaf can you review the msbuild changes?

wjk:
I've just rebased this PR to get rid of the merge conflicts. You may want to rerun CI just to be certain that nothing was broken by mistake. Thanks!

rolfbjarne:
build

mandel-macaque:
lets try a rebuild in case it was a CI issue.

mandel-macaque:
build

rolfbjarne:
Test failures are unrelated:

* mmptest/macOS: https://github.com/xamarin/maccore/issues/2035

Now we're just waiting for a review from @emaf.

rolfbjarne:
@wjk thanks a lot for your contribution!

Issue #6863, opened 2019-08-27
[Generator] A type mismatch on a BindAs attribute results in the generator crashing


mandel-macaque:
The situation is quite simple to explain, I forgot to add [] to a BindAs type in a property that returns an array. It is easier to understand with an example:

```csharp

[BindAs (typeof (MySuperEnum)]
NSString[] MySuperProperty { get; set; }
```

The above code, will result in the following exception/crash:

```
error BI0000: Unexpected error - Please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new
System.NullReferenceException: Object reference not set to an instance of an object
  at Generator.GetReturnsWrappers (IKVM.Reflection.MethodInfo mi, MemberInformation minfo, IKVM.Reflection.Type declaringType, System.String& cast_a, System.String& cast_b, EnumMode enum_mode, System.Text.StringBuilder postproc) [0x00647] in /Users/mandel/Xamarin/xcode11/xamarin-macios/src/generator.cs:3642 
  at Generator.GenerateInvoke (System.Boolean stret, System.Boolean supercall, IKVM.Reflection.MethodInfo mi, MemberInformation minfo, System.String selector, System.String args, System.Boolean assign_to_temp, IKVM.Reflection.Type category_type, System.Boolean aligned, EnumMode enum_mode) [0x002a3] in /Users/mandel/Xamarin/xcode11/xamarin-macios/src/generator.cs:3721 
  at Generator.GenerateNewStyleInvoke (System.Boolean supercall, IKVM.Reflection.MethodInfo mi, MemberInformation minfo, System.String selector, System.String[] args, System.Boolean assign_to_temp, IKVM.Reflection.Type category_type) [0x002bd] in /Users/mandel/Xamarin/xcode11/xamarin-macios/src/generator.cs:3806 
  at Generator+<>c__DisplayClass175_0.<GenerateMethodBody>b__2 () [0x00001] in /Users/mandel/Xamarin/xcode11/xamarin-macios/src/generator.cs:4345 
  at Generator.WriteIsDirectBindingCondition (System.IO.StreamWriter sw, System.Int32& tabs, System.Nullable`1[T] is_direct_binding, System.String is_direct_binding_value, System.Func`1[TResult] trueCode, System.Func`1[TResult] falseCode) [0x0006a] in /Users/mandel/Xamarin/xcode11/xamarin-macios/src/generator.cs:3111 
[...]

PR #7016, opened 2019-09-17
[Generator] Provide clear error message on type mismatch with BindAs error

wachaudh:
Fixes https://github.com/xamarin/xamarin-macios/issues/6863.


src/generator.cs
```diff
@@ -3699,6 +3699,9 @@ void GetReturnsWrappers (MethodInfo mi, MemberInformation minfo, Type declaringT
 				print ("IntPtr retvalarrtmp;");
 				cast_a = "((retvalarrtmp = ";
 				cast_b = ") == IntPtr.Zero ? null : (";
+				if (bindAsT == null) {
+					throw new BindingException (1067, true, "BindAs attribute of selector \"{0}\" is missing '[]' in the definition", minfo.selector);
```
mandel-macaque: I think we can do a lot better in the message that we throw and get a more decent information out of it. This exception is going to be misleading when we have a BindAs attribute that has been defined and that returns as null due to a type mismatch.

I would look at the GetBindAsAttribute method. There we are doing:

```csharp
BindAsAttribute rv;
		if (cu != null && (rv = AttributeManager.GetCustomAttribute<BindAsAttribute> (cu)) != null)
			return rv;

		var minfo = cu as MethodInfo;
		if (minfo?.ReturnParameter != null && (rv = AttributeManager.GetCustomAttribute<BindAsAttribute> (minfo.ReturnParameter)) != null)
			return rv;

		return null;
```
I first, would move any logic that does validation of the BindAs type there, meaning that if we have other cases in which we have a mismatch, this would work. Secondly, in the exception, we know that we tried to get the type of the BindAs attribute when the return type of a method being an Array, so at the very least, rather than saying:

> BindAs attribute of selector \"{0}\" is missing '[]' in the definition

We should be telling the developer something of the kind:

> {selector} return type does not match the type defined in the BindAs attribute, are you missing a []?

That will be a lot more helpful to developers since its is stating the problem and is possible since we have all the required information.

wachaudh: As we discussed offline, there was a misinterpretation of what the error message was saying so I've left it the same. Instead of doing a null check with the result of `GetBindAsAttribute`, I've updated it so it explicitly checks for `isArray` and avoids misleading exceptions. Adding validation to the `GetBindAsAttribute` would've required significantly changing how the generator is structured.

docs/website/generator-errors.md
```diff
@@ -198,6 +198,10 @@ This usually indicates a bug in Xamarin.iOS/Xamarin.Mac; please [file a bug repo
 
 ### <a name='BI1066'/>BI1066: Unsupported return type '{type}' in {method}.
 
+### <a name='BI1067'/>BI1067: BindAs attribute of the selector \"{selector}\" is missing '[]' in the definition.
```
rolfbjarne: I'd say something like: `The BindAs type for the member "{type}.{member}" must be an array when the member's type is an array.`

This could also benefit from some text that explains the problem (it doesn't have to be more than a bad example and then a good/corrected example).

docs/website/generator-errors.md
```diff
@@ -198,6 +198,10 @@ This usually indicates a bug in Xamarin.iOS/Xamarin.Mac; please [file a bug repo
...
+
+### <a name='BI1068'/>BI1068: BindAs attribute of the parameter \"{parameter}\" is missing '[]' in the definition.
```
rolfbjarne: `The BindAs type for the parameter "{parameterName}" in the method "{type}.{method}" must be an array when the parameter's type is an array.`

Note that I also include type and the method name in the message, because just the parameter name can be quite annoying if there are many methods that use the same parameter name.

src/generator.cs
```diff
@@ -1382,6 +1382,10 @@ string GetToBindAsWrapper (MemberInformation minfo = null, ParameterInfo pi = nu
 			temp = isNullable ? $"{parameterName} == null ? null : " : string.Empty;
 			temp += $"{FormatType (retType.DeclaringType, retType)}Extensions.GetConstant ({parameterName}{denullify});";
 		} else if (originalType.IsArray && originalType.GetArrayRank () == 1) {
+			if (!retType.IsArray) {
+				throw new BindingException (1072, true, "The BindAs type for the parameter \"{0}\" in the method \"{1}.{2}\" must be an array when the parameter's type is an array.",
+					parameterName,mi.ReturnType.Name, mi.Name);
```
rolfbjarne: The type name is wrong :smile: I think this is right:

```suggestion
					parameterName, mi.DeclaringType.FullName, mi.Name);
```

src/generator.cs
```diff
@@ -3650,7 +3654,11 @@ void GetReturnsWrappers (MethodInfo mi, MemberInformation minfo, Type declaringT
 		} else if (mai.Type.IsArray){
 			Type etype = mai.Type.GetElementType ();
 			if (minfo != null && minfo.is_bindAs) {
-				var bindAsT = GetBindAsAttribute (minfo.mi).Type.GetElementType ();
+				var bindAttrType = GetBindAsAttribute (minfo.mi).Type;
+				if (!bindAttrType.IsArray) {
+					throw new BindingException (1071, true, "The BindAs type for the member \"{0}.{1}\" must be an array when the member's type is an array.", mai.Type.Name, minfo.mi.Name);
```
rolfbjarne: Same, the type name is the member's type, I think this is right:

```suggestion
					throw new BindingException (1071, true, "The BindAs type for the member \"{0}.{1}\" must be an array when the member's type is an array.", minfo.mi.DeclaringType.FullName, minfo.mi.Name);
```

docs/website/generator-errors.md
```diff
@@ -212,6 +212,10 @@ This usually indicates a bug in Xamarin.iOS/Xamarin.Mac; please [file a bug repo
 
 ### <a name='BI1070'/>BI1070: The type '{type}' is trying to inline the property '{property}' from the protocols '{protocol1}' and '{protocol2}', but the inlined properties are of different types ('{property1}' is int, while '{property2}' is int).
 
+### <a name='BI1071'/>BI1071: The BindAs type for the member "{type}.{member}" must be an array when the member's type is an array. See src/tests/generator/ghissue6863_method.cs for example.
```
rolfbjarne: This documentation ends up on our website, which means that the reference to `src/tests/generator/ghissue6864_method.cs` will not be very meaningful.

It would be much better to copy the example code into this document.

src/generator.cs
wachaudh: Oops, misunderstood your previous comment. I think it's correct now.

docs/website/generator-errors.md
wachaudh: Where can I access the website? I've added an example to the doc

tests/generator/ErrorTests.cs
```diff
@@ -149,6 +149,28 @@ public void BI1049 ()
...
+		[Test]
+		public void GH6863_method ()
+		{
+			var bgen = new BGenTool ();
+			bgen.Profile = Profile.iOS;
+			bgen.CreateTemporaryBinding (File.ReadAllText (Path.Combine (Configuration.SourceRoot, "tests", "generator", "ghissue6863_method.cs")));
+			bgen.AssertExecuteError ("build");
+			bgen.AssertError (1072, "The BindAs type for the parameter \"id_test\" in the method \"GH6863_method.MyFooClass.StringMethod\" must be an array when the parameter's type is an array.");
+		}
+
```
prmallic: Is it possible to have both the property and method tests in a single test function? Also there is probably something obvious here, but where does the "GH6863" come from?

docs/website/generator-errors.md
```diff
@@ -212,6 +212,22 @@ This usually indicates a bug in Xamarin.iOS/Xamarin.Mac; please [file a bug repo
 
 ### <a name='BI1070'/>BI1070: The type '{type}' is trying to inline the property '{property}' from the protocols '{protocol1}' and '{protocol2}', but the inlined properties are of different types ('{property1}' is int, while '{property2}' is int).
 
+### <a name='BI1071'/>BI1071: The BindAs type for the member "{type}.{member}" must be an array when the member's type is an array. See example binding below.
+
+Correct:
+
+[BindAs (typeof (string []))]
```
rolfbjarne: Much better!

Just one tiny thing left: make the code fenced blocks, so that they're rendered like source code:

\`\`\`csharp
[BindAs (typeof (string []))]
[Export ("stringProp")]
NSString[] StringProp { get; }
\`\`\`

docs/website/generator-errors.md
rolfbjarne: Actually the documentation for the generator isn't available on the web (yet!), but we're publishing documentation for other tools (such as mtouch: https://docs.microsoft.com/en-us/xamarin/ios/troubleshooting/mtouch-errors - the generator errors would one day show up similiarly)

rolfbjarne:
I'm fine with this PR as-is.

PR #7019, opened 2019-09-18
Add SECURITY.MD to repo (as suggested by OSPO)

spouliot:


SECURITY.MD
```diff
@@ -0,0 +1,41 @@
+<!-- BEGIN MICROSOFT SECURITY.MD V0.0.2 BLOCK -->
```
mandel-macaque: +1, don't know if we need this

SECURITY.MD
```diff
@@ -0,0 +1,41 @@
...
+
+## Preferred Languages
+
+We prefer all communications to be in English.
+
+## Policy
+
+Microsoft follows the principle of [Coordinated Vulnerability Disclosure](https://www.microsoft.com/en-us/msrc/cvd).
+
+<!-- END MICROSOFT SECURITY.MD BLOCK -->
```
mandel-macaque: Ditto, looks good, no idea if we want the comment.

SECURITY.MD
spouliot: I almost removed it, but it tells us the version, and (afaik) it won't show up

PR #7023, opened 2019-09-18
[Versions] Use odd versions (.3) because Xcode 11.1 isn't stable yet

VincentDondain:


spouliot:
My bad, the numbers were meant as the _final_ (post d16-4 branch) ones, not the intermediate steps to get there. I expected the `d16-4` branch to happen rather quickly (if not immediately after the `xcode11` merge).

Ideally `xcode11.x` branches _should_ be *odd* before the associated Xcode becomes GM.
* xcode11.1 -> x.3.0 (previews) -> x.4.0 (stable)
* xcode11.2 -> x.5.0 (preview) -> x.6.0 (stable) _assuming it comes out before d16-4 and based on d16-3_

Whatever we have in `master` (that is not back ported) won't be available (in stable) before `d16-4` (x.8.0) so a `master` *pre* `d16-4` branch should be -> 0.7.x, like:

* master (pre 16-4 branch) -> x.7.0
* d16-4 -> x.8.0
* master (post 16-4 branch) -> x.9.0

VincentDondain:
Device tests failures are known (and fixed) xamarin/maccore#1983

PR #7036, opened 2019-09-19
[coop runtime] add a GC transition and relax some GC related assertions

lewurm:
Contributes to https://github.com/mono/mono/issues/16819


mandel-macaque:
Agreed, and it will make it easier for the average user.

rolfbjarne:
build

PR #7040, opened 2019-09-20
Revert "[Generator] Provide clear error message on type mismatch with BindAs error"

wachaudh:
Reverts xamarin/xamarin-macios#7016

Accidentally merged into master without approvals.


rolfbjarne:
I'm OK with the original PR, so from my side it's not necessary to revert.

PR #7072, opened 2019-09-23
Use Xcode 11 final

VincentDondain:
It's the same hash as Xcode 11 GM 2 but it's cleaner to actually use `Xcode11.app`.

In addition: the device bots are looking for "GM" and "beta" in the Xcode name to run on either the stable or beta pool.


VincentDondain:

VincentDondain:
@monojenkins backport to xcode11.1

Issue #7045, opened 2019-09-20
[xcode11.1] Update Versions file for 13.1


VincentDondain:
Need to update this https://github.com/xamarin/xamarin-macios/blob/xcode11.1/Versions-ios.plist.in#L36 for 13.1.

spouliot:
merged, closing

PR #7081, opened 2019-09-23
[Versions] Add 13.1 to our version files

VincentDondain:
Add dedicated MAX_XXX_VERSION variables because the OS version doesn't always match the SDK version (happened before and will happen in the future).

- Fixes https://github.com/xamarin/xamarin-macios/issues/7045

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -96,7 +96,7 @@ TVOS_SDK_VERSION=13.0
 # If the max OS version does not match the SDK version (for instance the iOS
 # 12.2 SDK supported both iOS 12.3 and iOS 12.4), then these variables can be
 # set to something other than the corresponding SDK versions.
-MAX_IOS_VERSION=$(IOS_SDK_VERSION)
+MAX_IOS_VERSION=13.1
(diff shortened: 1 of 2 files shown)


Make.config
```diff
@@ -93,6 +93,13 @@ OSX_SDK_VERSION=10.15
 WATCH_SDK_VERSION=6.0
 TVOS_SDK_VERSION=13.0
 
+# Maximum OS versions for running XI/XM apps.
```
rolfbjarne: We already have this logic in master, but slightly different: https://github.com/xamarin/xamarin-macios/blob/2adecee3d98ad93599141db53376971d4eb73633/Make.config#L92-L97

Could you make it identical (or as close as possible)? That will hopefully make merging this branch into master later easier.

Also you're assigning strings to the MACOS/WATCHOS/TVOS variables, not version represented by those strings.

VincentDondain: 🤦‍♂ I forgot I added that to master

VincentDondain:
Superseded by https://github.com/xamarin/xamarin-macios/pull/7090

Issue #7084, opened 2019-09-24
ERROR while running iOS Project **System.Reflection.TargetInvocationException:** 'Exception has been thrown by the target of an invocation.'


manonaga2188:
### Steps to Reproduce

1. Run the iOS application in real device.
2. 
3. 

### Actual Behavior
Return the following error.
**System.Reflection.TargetInvocationException:** 'Exception has been thrown by the target of an invocation.'

### Environment
Microsoft Visual Studio Community 2019
Version 16.3.0
VisualStudio.16.Release/16.3.0+29318.209
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Community

Application Insights Tools for Visual Studio Package   9.1.00913.1
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2019   16.3.282.33872
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   16.3.282.33872
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.3.282.33872
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   16.3.282.33872
Azure Functions and Web Jobs Tools

C# Tools   3.3.1-beta3-19461-02+2fd12c210e22f7d6245805c60340f6a34af6875b
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.0 (d16-2@8b56e20)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Microsoft Azure Tools   2.9
Microsoft Azure Tools for Microsoft Visual Studio 0x10 - v2.9.20816.1

Microsoft Continuous Delivery Tools for Visual Studio   0.4
Simplifying the configuration of Azure DevOps pipelines from within the Visual Studio IDE.

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft Library Manager   2.0.83+gbc8a4b23ec
Install client-side libraries easily to any web project

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Microsoft Visual Studio Tools for Containers   1.1
[...]

Issue #7083, opened 2019-09-24
Shell Flyout crashes on iPad 2 and iOS 9.3


xperseguers:
### Steps to Reproduce

1. Download the official sample application xaminals from https://docs.microsoft.com/en-gb/samples/xamarin/xamarin-forms-samples/userinterface-xaminals/
2. Upgrade Xamarin.Forms package to latest (4.2.0.815419)
3. Run a simulation on iPhone 6s and iOS 9.3 => works
4. Run a simulation on iPad 2 and iOS 9.3 => crashes

### Expected Behavior

The application should not crash.

### Actual Behavior

The application is crashing. Could not get a stacktrace with this project but with a real project of mine, and although I have other 3rd-party libraries initialization lines crashing earlier, it seems to crash as well with:

```
[...]

PR #7085, opened 2019-09-24
[generator] Do not generate PlatformNotSupportedException in chaining .ctor

spouliot:
Types that are new in 64bits only OS are generated differently on 32bits
bindings. They mainly throw a `PlatformNotSupportedException` so it's
easier to diagnose (than a crash) what's happening at runtime.

This works well in all cases except one. When a new type, let's say
`UIMenuElement` is added **and** serves as a new base type for existing
types.

`UIKeyCommand` (iOS 7) -> `UICommand` (iOS 13)-> `UIMenuElement` (iOS 13)

This is _correct_ as new base types can be added (in ObjC and C#).
However the generated code for the constructors of `UICommand` and
`UIMenuElement` would be throwing a `PlatformNotSupportedException`
which breaks the `UIKeyCommand` on 32 bits devices.

We fixed this in a few places by tweaking the availability attribute
but that requires spotting the new base type while doing bindings and
that is error prone [1][2].

This PR simply does let the `protected` constructor, using when chaining,
be generated normally. It's simpler and will cover all the cases (without
requiring hacks in the availability of those types)

[1] https://github.com/xamarin/xamarin-macios/issues/7083
[2] https://github.com/xamarin/xamarin-macios/issues/7084

### DIFF
--- a/src/generator.cs
+++ b/src/generator.cs
@@ -6479,16 +6479,9 @@ $" using methods with different signatures ('{distinctMethodsBySignature [0].Met
 						sw.WriteLine ("\t\t[EditorBrowsable (EditorBrowsableState.Advanced)]");
 						sw.WriteLine ("\t\tprotected {0} (NSObjectFlag t) : base (t)", TypeName);
 						sw.WriteLine ("\t\t{");
-						if (is32BitNotSupported) {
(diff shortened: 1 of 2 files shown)


spouliot:
@monojenkins backport to master

PR #7090, opened 2019-09-24
Bump for Xcode 11.1 GM

VincentDondain:


Make.config
```diff
@@ -46,9 +46,9 @@ IOS_PACKAGE_VERSION_BUILD=$(IOS_COMMIT_DISTANCE)
 IOS_PACKAGE_UPDATE_ID=$(shell printf "2%02d%02d%02d%03d" $(IOS_PACKAGE_VERSION_MAJOR) $(IOS_PACKAGE_VERSION_MINOR) $(IOS_PACKAGE_VERSION_REV) $(IOS_PACKAGE_VERSION_BUILD))
 
 # Xcode version should have both a major and a minor version (even if the minor version is 0)
-XCODE_VERSION=11.0
-XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_11.xip
-XCODE_DEVELOPER_ROOT=/Applications/Xcode11.app/Contents/Developer
+XCODE_VERSION=11.1
+XCODE_URL=http://xamarin-storage/bot-provisioning/xcodes/Xcode_11.1_GM_Seed.xip
+XCODE_DEVELOPER_ROOT=/Applications/Xcode11.1-GM.app/Contents/Developer
```
dalexsoto: I think this should be `Xcode111-GM.app` at least that is what we did for Xcode 10 minors

```suggestion
XCODE_DEVELOPER_ROOT=/Applications/Xcode111-GM.app/Contents/Developer
```

spouliot: is that the location on the bots ?
it does not match https://xamarinhq.slack.com/archives/C99ER4A1M/p1569347681015300

dalexsoto: Yeah I think we went with this https://xamarinhq.slack.com/archives/C03CEMRFL/p1569348227198300

`Xcode111-GM.app`

mandel-macaque:
@VincentDondain  CoreBluetooth is making the tests on Mac OS X fail, we did have some changes in there, we need to update the tests.

VincentDondain:
So I'm not quite sure how we end up in this situation on macOS:

```
ObjCRuntime.RuntimeException: Failed to marshal the Objective-C object 0x7ff62ff227f0 (type: MonoTouchFixtures_CoreBluetooth_CBCentralManagerTest_ManagerDelegate). Could not find an existing managed instance for this object, nor was it possible to create a new managed instance (because the type 'MonoTouchFixtures.CoreBluetooth.CBCentralManagerTest+ManagerDelegate' does not have a constructor that takes one IntPtr argument).
```

I was also thinking there could be a problem with the test `Setup` where the `ManagerDelegate` would not be initialized somehow. I cannot repro locally so it doesn't help with the investigation of the issue and I am not up to date on the CoreBluetooth-related things that happened this summer.

I'll take thoughts on this (:

tests/monotouch-test/CoreBluetooth/CentralManagerTest.cs
```diff
@@ -36,6 +36,10 @@ public class CBCentralManagerTest {
 		class ManagerDelegate : CBCentralManagerDelegate {
 			public AutoResetEvent Event { get; private set; }
 
+			public ManagerDelegate (IntPtr handle) : base (handle)
+			{
+			}
+
```
spouliot: That won't set the managed state (`Event`) back to a valid (non-null) value. It will likely work, just like rebuilding would have - i.e. your code change being irrelevant to the test success/failure.

My _guess_ is that the 5sec timeout occurred on the bot and, while `TearDown` happened the callback occurred, throwing before the timeout assertion could occur.

You can try to duplicate this locally with a much shorter timeout.

VincentDondain: Still can't repro...

I tried playing with `mgrDelegate.Event.WaitOne` and the test's timeout `[Timeout]` and couldn't get the `TearDown` to happen before the Asserts...

I'm gonna ignore the test, file a bug and merge that PR so we can get going.

VincentDondain: Mmh actually it might be `ScanForPeripherals` failing... not `RetrievePeripherals`

dalexsoto:
build

PR #7091, opened 2019-09-24
[Networking] Bring the framework up to Xcode 11.1

mandel-macaque:
Draft PR, tests needs to be added, there are two methods from NWFramer that I need to understand and a small API that is Mac OS X only.


src/Network/NWDataTransferReport.cs
```diff
@@ -0,0 +1,174 @@
+//
+// NWDataTransferReport.cs: Bindings the Network nw_data_transfer_report_t API.
+//
+// Authors:
+//   Manuel de la Pena (mandel@microsoft.com)
+//
+// Copyrigh 2019 Microsoft Inc
```
whitneyschmidt: `Copyright`

src/Network/NWEstablishmentReport.cs
```diff
@@ -0,0 +1,135 @@
+
+//
+// NWDataTransferReport.cs: Bindings the Network nw_data_transfer_report_t API.
+//
+// Authors:
+//   Manuel de la Pena (mandel@microsoft.com)
+//
+// Copyrigh 2019 Microsoft Inc
```
whitneyschmidt: typo

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
+//
+// NWFramer.cs: Bindings the Network nw_framer_t API.
+//
+// Authors:
+//   Manuel de la Pena (mandel@microsoft.com)
+//
+// Copyrigh 2019 Microsoft Inc
```
whitneyschmidt: typo

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace Network {
+
+	[Flags, TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public enum NWBrowseResultChange : ulong {
+		Invalid = 0x00,
+		Identical = 0x01,
+		ResultAdded = 0x02,
+		ResultRemoved = 0x04,
+		TxtRecordChanged = 0x20,
+		InterfaceAdded = 0x08,
```
stephen-hawley: For maintenance, it's probably best to put these into numerical order rather than how they are documented.
For flags like this, I prefer to use shift constants like `Identical = 1 << 0, ResultAdded = 1 << 1, ResultRemoved = 1 << 2` etc.

src/Network/NWTxtRecord.cs
```diff
@@ -0,0 +1,240 @@
...
+
+namespace Network {
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public enum NWTxtRecordFindKey {
+		Invalid = 0,
+		NotPresent = 1,
+		NoValue = 2,
+		EmptyValue = 3,
+		NonEmptyValue = 4
```
whitneyschmidt: super nit: `,` for last enum

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_endpoint nw_browse_result_copy_endpoint (OS_nw_browse_result result);
+
+		NWEndpoint EndPoint => new NWEndpoint (nw_browse_result_copy_endpoint (GetCheckedHandle ()), owns: true);
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern nuint nw_browse_result_get_interfaces_count (OS_nw_browse_result result);
+
+		nuint InterfacesCount => nw_browse_result_get_interfaces_count (GetCheckedHandle ());
```
stephen-hawley: This is private and is never used. Why?

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+			var del = BlockLiteral.GetTarget<Action<NWInterface>> (block);
+			if (del != null) {
+				var nwInterface = new NWInterface (inter, owns: false);
+				del (nwInterface);
+			}
+		}
+
+		public void EnumerateInterfaces (Action<NWInterface> handler)
+		{
+			unsafe {
```
stephen-hawley: I wonder about thread safety on this. If a consumer of this API uses the same object from two different threads, will there be an issue?
Is this something that wants to be async?
Or is it something that wants to have an event?

src/Network/NWBrowser.cs
```diff
@@ -0,0 +1,152 @@
...
+				if (value == null)
+					throw new ArgumentNullException (nameof (value));
+				nw_browser_set_queue (GetCheckedHandle (), value.Handle);
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_browser_start (OS_nw_browser browser);
+
+		public void Start () => nw_browser_start (GetCheckedHandle ());
```
stephen-hawley: I know that this is not reflected in the Apple API, but if I have a `Start` and a `Cancel`, I'd like to have `bool IsStarted { get; private set; }` The Apple documentation doesn't say if you can call `Start` multiple times.

src/Network/NWBrowser.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using ObjCRuntime;
+using Foundation;
+using CoreFoundation;
+
+using OS_nw_browser=System.IntPtr;
+using OS_nw_browse_descriptor=System.IntPtr;
+using OS_nw_parameters=System.IntPtr;
+using dispatch_queue_t =System.IntPtr;
```
prmallic: Is this like using `#define` in c++? also, maybe an extra space here?

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+		}
+
+		public bool ParseOutput (nuint minimumIncompleteLength, nuint maximumLength, byte[] tempBuffer, Action<byte[], bool> handler)
+		{
+			unsafe {
+				if (handler == null) {
+					if (tempBuffer == null)
+						return nw_framer_parse_output (GetCheckedHandle (), minimumIncompleteLength, maximumLength, IntPtr.Zero, null);
+					else
+						fixed (void* tmpBufferPtr = tempBuffer)
```
wachaudh: Seems like something is missing here. You're missing a semicolon on `fixed` or the indentation of the return statement is wrong.

src/Network/NWDataTransferReport.cs
```diff
@@ -0,0 +1,174 @@
...
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern ulong nw_data_transfer_report_get_duration_milliseconds (OS_nw_data_transfer_report report);
+
+		public ulong Duration => nw_data_transfer_report_get_duration_milliseconds (GetCheckedHandle ());
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_interface nw_data_transfer_report_copy_path_interface (OS_nw_data_transfer_report report, uint path_index);
+
+		public NWInterface GetInterface (uint pathIndex)
```
stephen-hawley: All of these properties with `Get` should either be made methods or have the `Get` prefix removed. As a property, the `Get` is implicit in using the getter.

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+using OS_dispatch_data=System.IntPtr;
+using OS_nw_protocol_definition=System.IntPtr;
+using OS_nw_protocol_options=System.IntPtr;
+using OS_nw_endpoint=System.IntPtr;
+using OS_nw_parameters=System.IntPtr;
+
+namespace Network {
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public class NWFramer : NWProtocolMetadata {
```
mandel-macaque: @migueldeicaza I'm curious about this inheritance, looking at the headers, and the properties expose by the NWProtocolMetadata object my understanding is that NWProtocolMetadata is the parent class.

src/Network/NWWebSocketOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+	// this maps to `nw_ws_version_t` in Network.framework/Headers/ws_options.h (and not the enum from NetworkExtension)
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public enum NWWebSocketVersion {
+		Invalid = 0,
+		Version13 = 1,
+	}
+
+	[TV (12,0), Mac (10,14), iOS (12,0)]
+	[Watch (6,0)]
+	public class NWWebSocketOptions : NWProtocolOptions {
```
mandel-macaque: @migueldeicaza same here, I think I did choose the correct base class but I'd like some extra pair of eyes.

src/Network/NWTxtRecord.cs
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern byte nw_txt_record_set_key(IntPtr handle, string key, IntPtr value, nuint valueLen);
+
+		public bool SetKey (string key, byte [] value)
+		{
+			if (value == null)
+				return nw_txt_record_set_key (Handle, key, IntPtr.Zero, 0) != 0;
+			else {
```
wachaudh: Redundant else statement that causes unnecessary indentation. You don't need an else because the if statement returns.

src/Network/NWEstablishmentReport.cs
```diff
@@ -0,0 +1,135 @@
...
+		{
+			var del = BlockLiteral.GetTarget<Action<NWReportResolutionSource, nuint, int, NWEndpoint, NWEndpoint>> (block);
+			if (del != null) {
+				var nwSuccesfulEndpoint = new NWEndpoint (successful_endpoint, owns: false);
+				var nwPreferredEndpoint = new NWEndpoint (preferred_endpoint, owns: false);
+				del (source, milliseconds, endpoint_count, nwSuccesfulEndpoint, nwPreferredEndpoint);
+			}
+		}
+
+		public void ResolutionEnumeration (Action<NWReportResolutionSource, nuint, int, NWEndpoint, NWEndpoint> handler)
```
stephen-hawley: Better to be named `EnumerateResolutions` I would also be happier if this was an `async` or if there were an `async` version.

src/Network/NWDataTransferReport.cs
```diff
@@ -0,0 +1,174 @@
...
+		static void TrampolineCollectHandler (IntPtr block, IntPtr report)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWDataTransferReport>> (block);
+			if (del != null) {
+				var nwReport = new NWDataTransferReport (report, owns: false);
+				del (nwReport);
+			}
+		}
+
+		public void Collect (DispatchQueue queue, Action<NWDataTransferReport> handler)
```
stephen-hawley: `async`?

src/Network/NWFramer.cs
prmallic: I think its alright. Check out - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/fixed-statement

src/Network/NWBrowser.cs
whitneyschmidt: Yes, `using` is like `#define`/macros in c++.

But why are we mapping all of these strings to `System.IntPtr`?

src/Network/NWBrowseResult.cs
mandel-macaque: Because I missed the public :facepalm:

src/Network/NWBrowser.cs
mandel-macaque: Yes, I'm defining the dispatch_queue_t, OS_nw_parameters etc to be IntPtrs, so that when you read the PInvokes it makes more sense, is nicer to have: 

```csharp
static extern void nw_browser_set_queue (OS_nw_browser browser, dispatch_queue_t queue);
```
than
```csharp
static extern void nw_browser_set_queue (IntPtr browser, IntPtr queue);
```

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_browse_result_enumerate_interfaces (OS_nw_browse_result result, void *enumerator);
+
+		delegate void nw_browse_result_enumerate_interfaces_t (IntPtr block, IntPtr nwInterface);
+		static nw_browse_result_enumerate_interfaces_t static_EnumerateInterfacesHandler = TrampolineEnumerateInterfacesHandler;
+
+		[MonoPInvokeCallback (typeof (nw_browse_result_enumerate_interfaces_t))]
+		static void TrampolineEnumerateInterfacesHandler (IntPtr block, IntPtr inter)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWInterface>> (block);
```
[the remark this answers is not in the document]
mandel-macaque: The BlockLiteral is wrapping the Action as a Objc block to be called by the Objc runtime. So, when I get the IntPtr of the blockLiteral, I'm using the method to get the managed code to be executed.

As a ref: https://docs.microsoft.com/en-us/dotnet/api/objcruntime.blockliteral?view=xamarin-ios-sdk-12

src/Network/NWBrowseResult.cs
mandel-macaque: It is 0 thread safe, if two threads have the same object, change and change the callback, bad things will happen. The problem is that the C API relies on C callbacks, so unless I do something very smart, people should be doing the thread management in the code. I agree that the pattern sucks.

src/Network/NWDataTransferReport.cs
mandel-macaque: ah! but they take a parameter! we are getting the interface in an index, all other methods that do not take a parameter (unless I missed one) are exposed as properties.

src/Network/NWFramer.cs
mandel-macaque: +1 on @prmallic I am pinning the object to ensure is not GC and passing it to the C function.

src/Network/NWDataTransferReport.cs
mandel-macaque: @stephen-hawley I think I am able to write the full API with the async/await support and using completion sources, might be a good thing as a second pass for the API. @chamons can I have the time for that between Xcodes?

src/Network/NWAdvertiseDescriptor.cs
```diff
@@ -61,5 +62,16 @@ public void SetTxtRecord (string txt)
...
 			get => nw_advertise_descriptor_get_no_auto_rename (GetCheckedHandle ());
 		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_txt_record nw_advertise_descriptor_copy_txt_record_object (OS_nw_advertise_descriptor advertise_descriptor);
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_advertise_descriptor_set_txt_record_object (OS_nw_advertise_descriptor advertise_descriptor, OS_nw_txt_record txt_record);
+
+		public NWTxtRecord TxtRecord {
```
mandel-macaque: @mandel-macaque Add availability attributes.

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public class NWBrowseResult : NativeObject {
+
+		public NWBrowseResult (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_endpoint nw_browse_result_copy_endpoint (OS_nw_browse_result result);
+
+		NWEndpoint EndPoint => new NWEndpoint (nw_browse_result_copy_endpoint (GetCheckedHandle ()), owns: true);
```
mandel-macaque: public missing.

src/Network/NWBrowseResult.cs
```diff
@@ -0,0 +1,93 @@
...
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern nuint nw_browse_result_get_interfaces_count (OS_nw_browse_result result);
+
+		nuint InterfacesCount => nw_browse_result_get_interfaces_count (GetCheckedHandle ());
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_txt_record nw_browse_result_copy_txt_record_object (OS_nw_browse_result result);
+
+		NWTxtRecord TxtRecord => new NWTxtRecord (nw_browse_result_copy_txt_record_object (GetCheckedHandle ()), owns: true);
```
mandel-macaque: same.

src/Network/NWBrowseResult.cs
mandel-macaque: same

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+using OS_nw_browse_descriptor=System.IntPtr;
+
+namespace Network {
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public class NWBrowserDescriptor: NativeObject {
+
+		public NWBrowserDescriptor (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
```
mandel-macaque: Really need all the properties?

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+			return new NWBrowserDescriptor (nw_browse_descriptor_create_bonjour_service (type, domain), owns: true);
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern bool nw_browse_descriptor_get_include_txt_record (OS_nw_browse_descriptor descriptor);
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_browse_descriptor_set_include_txt_record (OS_nw_browse_descriptor descriptor, bool include_txt_record);
+
+		bool IncludeTxtRecord {
```
mandel-macaque: public missing.

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+			get => nw_browse_descriptor_get_include_txt_record (GetCheckedHandle ()); 
+			set => nw_browse_descriptor_set_include_txt_record (GetCheckedHandle (), value); 
+		}
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_browse_descriptor_get_bonjour_service_type (OS_nw_browse_descriptor descriptor);
+
+		string BonjourType => nw_browse_descriptor_get_bonjour_service_type (GetCheckedHandle ());
```
mandel-macaque: public missing.

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_browse_descriptor_get_bonjour_service_type (OS_nw_browse_descriptor descriptor);
+
+		string BonjourType => nw_browse_descriptor_get_bonjour_service_type (GetCheckedHandle ()); 
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_browse_descriptor_get_bonjour_service_domain (OS_nw_browse_descriptor descriptor);
+
+		string BonjourDomain => nw_browse_descriptor_get_bonjour_service_domain (GetCheckedHandle ());
```
mandel-macaque: public missing.

src/Network/NWConnection.cs
```diff
@@ -503,5 +504,43 @@ public void Batch (Action method)
...
+		static void TrampolineGetEstablishmentReportHandler (IntPtr block, nw_establishment_report_t report)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWEstablishmentReport>> (block);
+			if (del != null) {
+				var nwReport = new NWEstablishmentReport (report, owns: true);
+				del (nwReport);
+			}
+		}
+
+		public void GetEstablishmentReport (DispatchQueue queue, Action<NWEstablishmentReport> handler)
```
mandel-macaque: Add availability attributes.

src/Network/NWProtocolDefinition.cs
```diff
@@ -61,5 +61,9 @@ public bool Equals (object other)
 
 		public static NWProtocolDefinition TlsDefinition => new NWProtocolDefinition (nw_protocol_copy_tls_definition (), owns: true);
 
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_protocol_definition nw_protocol_copy_ws_definition ();
+
+		public static NWProtocolDefinition WebServiceDefinition => new NWProtocolDefinition (nw_protocol_copy_ws_definition (), owns: true);
```
mandel-macaque: Add missing availability attributes.

src/Network/NWProtocolOptions.cs
```diff
@@ -169,21 +185,23 @@ public void IPSetCalculateReceiveTime (bool calculateReceiveTime)
 		extern static void nw_tcp_options_set_disable_ecn (IntPtr handle, [MarshalAs (UnmanagedType.U1)] bool disableEcn);
 		public void TcpSetDisableEcn (bool disableEcn) => nw_tcp_options_set_disable_ecn (GetCheckedHandle (), disableEcn);
 
-//
-// UDP Options
-//
+		//
+		// UDP Options
+		//
+
```
mandel-macaque: Remove diff noise.

src/Network/NWProtocolOptions.cs
```diff
@@ -169,21 +185,23 @@ public void IPSetCalculateReceiveTime (bool calculateReceiveTime)
...
 		extern static void nw_udp_options_set_prefer_no_checksum (IntPtr handle, [MarshalAs (UnmanagedType.U1)] bool preferNoChecksums);
 
 		public void UdpSetPreferNoChecksum (bool preferNoChecksums) => nw_udp_options_set_prefer_no_checksum (GetCheckedHandle (), preferNoChecksums);
 
-//
-// TLS options
-//
+		//
+		// TLS options
+		//
```
mandel-macaque: Remove diff noise.

src/Network/NWProtocolOptions.cs
```diff
@@ -169,21 +185,23 @@ public void IPSetCalculateReceiveTime (bool calculateReceiveTime)
...
 
 		[DllImport (Constants.NetworkLibrary)]
 		extern static IntPtr nw_tls_copy_sec_protocol_options (IntPtr options);
 
 		public SecProtocolOptions TlsProtocolOptions => new SecProtocolOptions (nw_tls_copy_sec_protocol_options (GetCheckedHandle ()), owns: true);
+
```
mandel-macaque: Remove diff noise.

src/Network/NWWebSocketMetadata.cs
```diff
@@ -0,0 +1,118 @@
...
+using CoreFoundation;
+
+using OS_nw_protocol_metadata=System.IntPtr;
+using OS_nw_ws_response=System.IntPtr;
+using dispatch_queue_t =System.IntPtr;
+
+namespace Network {
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public enum NWWebSocketOpCode : long {
```
rolfbjarne: `nw_ws_opcode_t` doesn't have a specific type in the headers, which means it's an `int`, not `long`.

src/Network/NWWebSocketMetadata.cs
```diff
@@ -0,0 +1,118 @@
...
+		Text = 0x1,
+		Binary = 0x2,
+		Close = 0x8,
+		Ping = 0x9,
+		Pong = 0xA,
+		Invalid = -1,
+	}
+
+	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+	public enum NWWebSocketCloseCode : long {
```
rolfbjarne: Same: `int`, not `long`.

src/Network/NWWebSocketRequest.cs
```diff
@@ -0,0 +1,90 @@
...
+			var del = BlockLiteral.GetTarget<Action<string, string>> (block);
+			if (del != null) {
+				del (header, value);
+			}
+		}
+
+		public void EnumerateAdditionalHeaders (Action<string, string> handler)
+		{
+			unsafe {
+				if (handler == null) {
```
rolfbjarne: A null handler doesn't make much sense for this method.

src/Network/NWWebSocketRequest.cs
```diff
@@ -0,0 +1,90 @@
...
+			if (del != null) {
+				del (subprotocol);
+			}
+		}
+
+		public void EnumerateSubprotocols (Action<string> handler)
+		{
+			unsafe {
+				if (handler == null) {
+					nw_ws_request_enumerate_subprotocols (GetCheckedHandle (), null);
```
rolfbjarne: Same: null handler doesn't make much sense.

src/Network/NWWebSocketResponse.cs
```diff
@@ -0,0 +1,86 @@
...
+		public NWWebSocketResponse (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern unsafe OS_nw_ws_response nw_ws_response_create (NWWebSocketResponseStatus status, string selected_subprotocol);
+
+		public NWWebSocketResponse (NWWebSocketResponseStatus status, string subprotocol)
+			=> InitializeHandle (nw_ws_response_create (status, subprotocol));
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern unsafe string nw_ws_response_get_selected_subprotocol (OS_nw_ws_response response);
```
rolfbjarne:
```suggestion
		static extern string nw_ws_response_get_selected_subprotocol (OS_nw_ws_response response);
```

rolfbjarne: Also I think you need to marshal the return value manually, because by default the managed runtime will try to free the native char* pointer.

src/Network/NWWebSocketResponse.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		public string SelectedSubprotocol => nw_ws_response_get_selected_subprotocol (GetCheckedHandle ()); 
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern NWWebSocketResponseStatus nw_ws_response_get_status (OS_nw_ws_response response);
+
+		public NWWebSocketResponseStatus Status => nw_ws_response_get_status (GetCheckedHandle ()); 
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern bool nw_ws_response_enumerate_additional_headers (OS_nw_ws_response response, void *enumerator);
```
rolfbjarne:
```suggestion
	static extern bool nw_ws_response_enumerate_additional_headers (OS_nw_ws_response response, void *enumerator);
```

src/Network/NWWebSocketResponse.cs
```diff
@@ -0,0 +1,86 @@
...
+			var del = BlockLiteral.GetTarget<Action<string, string>> (block);
+			if (del != null) {
+				del (header, value);
+			}
+		}
+
+		public bool EnumerateAditionalHeaders (Action<string, string> handler)
+		{
+			unsafe {
+				if (handler == null) {
```
rolfbjarne: Does a null handler make sense?

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+		public bool WriteOutput (nuint outputLength) => nw_framer_write_output_no_copy (GetCheckedHandle (), outputLength);
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_framer_write_output_data (OS_nw_framer framer, OS_dispatch_data output_data);
+
+		public void WriteOutput (DispatchData data)
+		{
+			if (data == null)
+				throw new ArgumentNullException (nameof (data));
+			nw_framer_write_output_data (GetCheckedHandle(), data.Handle);
```
rolfbjarne:
```suggestion
			nw_framer_write_output_data (GetCheckedHandle (), data.Handle);
```

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern unsafe void nw_framer_write_output (OS_nw_framer framer, IntPtr output_buffer, nuint output_length);
```
rolfbjarne:
```suggestion
		static extern void nw_framer_write_output (OS_nw_framer framer, byte[] output_buffer, nuint output_length);
```

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+
+		public void WriteOutput (byte[] data)
+		{
+			if (data == null)
+				throw new ArgumentNullException (nameof (data));
+			unsafe {
+				fixed (void* dataHandle = data)
+					nw_framer_write_output (GetCheckedHandle (), (IntPtr) dataHandle, (nuint)data.Length);
```
rolfbjarne:
```suggestion
					nw_framer_write_output (GetCheckedHandle (), data, (nuint) data.Length);
```

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+
+		public bool ParseOutput (nuint minimumIncompleteLength, nuint maximumLength, byte[] tempBuffer, Action<byte[], bool> handler)
+		{
+			unsafe {
+				if (handler == null) {
+					if (tempBuffer == null)
+						return nw_framer_parse_output (GetCheckedHandle (), minimumIncompleteLength, maximumLength, IntPtr.Zero, null);
+					else
+						fixed (void* tmpBufferPtr = tempBuffer)
+							return nw_framer_parse_output (GetCheckedHandle (), minimumIncompleteLength, maximumLength, (IntPtr)tmpBufferPtr, null);
```
rolfbjarne: If you change the P/Invoke to take a byte[] instead of IntPtr, you can simplify a lot of the code here.

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+
+		public bool ParseInput (nuint minimumIncompleteLength, nuint maximumLength, byte[] tempBuffer, Action<byte[], bool> handler)
+		{
+			unsafe {
+				if (handler == null) {
+					if (tempBuffer == null)
+						return nw_framer_parse_input (GetCheckedHandle (), minimumIncompleteLength, maximumLength, IntPtr.Zero, null);
+					else
+						fixed (void* tmpBufferPtr = tempBuffer)
+							return nw_framer_parse_input (GetCheckedHandle (), minimumIncompleteLength, maximumLength, (IntPtr)tmpBufferPtr, null);
```
rolfbjarne: Same: make the P/Invoke take a byte[] instead and this code can be simplified.

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+				fixed (void* valuePtr = value)
+					nw_framer_message_set_value (GetCheckedHandle (), key, (IntPtr) valuePtr, null);
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_framer_message_set_object_value (OS_nw_protocol_metadata message, string key, IntPtr value);
+
+		public void SetKey (string key, NSObject value)
+			=> nw_framer_message_set_object_value (GetCheckedHandle (), key, value?.Handle ?? IntPtr.Zero);
```
rolfbjarne:
```suggestion
			=> nw_framer_message_set_object_value (GetCheckedHandle (), key, value.GetHandle ()); 
```

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+			unsafe {
+				fixed (void* valuePtr = value)
+					nw_framer_message_set_value (GetCheckedHandle (), key, (IntPtr) valuePtr, null);
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_framer_message_set_object_value (OS_nw_protocol_metadata message, string key, IntPtr value);
+
+		public void SetKey (string key, NSObject value)
```
rolfbjarne:
```suggestion
		public void SetObject (string key, NSObject value)
```

src/Network/NWFramer.cs
```diff
@@ -0,0 +1,474 @@
...
+			=> nw_framer_message_set_object_value (GetCheckedHandle (), key, value?.Handle ?? IntPtr.Zero); 
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern IntPtr nw_framer_message_copy_object_value (OS_nw_protocol_metadata message, string key);
+
+		public NSObject GetValue (string key)
+			=> new NSObject (nw_framer_message_copy_object_value (GetCheckedHandle (), key));
```
rolfbjarne:
```suggestion
			=> new Runtime.GetNSObject (nw_framer_message_copy_object_value (GetCheckedHandle (), key));
```

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_browse_descriptor_set_include_txt_record (OS_nw_browse_descriptor descriptor, bool include_txt_record);
+
+		bool IncludeTxtRecord {
+			get => nw_browse_descriptor_get_include_txt_record (GetCheckedHandle ()); 
+			set => nw_browse_descriptor_set_include_txt_record (GetCheckedHandle (), value); 
+		}
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_browse_descriptor_get_bonjour_service_type (OS_nw_browse_descriptor descriptor);
```
viniciusjarina: AFAIK any p/Invoke should only return string if the CLR/Mono needs to free the string, otherwise you will have a native crash. (as rolf stated here https://github.com/xamarin/xamarin-macios/pull/7091/files/a5fdf0e6b004dfaea4643c3d095e9fb5b2456646#r327966707) 

You will probably need to manually marshal the string manually for almost all method which return a `char *`
```
            IntPtr ptr = nw_browse_descriptor_get_bonjour_service_type(descriptor)
            return Marshal.PtrToStringAnsi(ptr);
```

src/Network/NWBrowserDescriptor.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		string BonjourType => nw_browse_descriptor_get_bonjour_service_type (GetCheckedHandle ()); 
+
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_browse_descriptor_get_bonjour_service_domain (OS_nw_browse_descriptor descriptor);
```
viniciusjarina: Same here https://github.com/xamarin/xamarin-macios/pull/7091/files/a5fdf0e6b004dfaea4643c3d095e9fb5b2456646#r328214811

src/Network/NWEndpoint.cs
```diff
@@ -112,5 +112,27 @@ public static NWEndpoint CreateBonjourService (string name, string serviceType,
...
+				throw new ArgumentNullException (nameof (url));
+			var handle = nw_endpoint_create_url (url);
+			if (handle == IntPtr.Zero)
+				return null;
+			return new NWEndpoint (handle, owns: true);
+		}
+
+		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[DllImport (Constants.NetworkLibrary, CharSet = CharSet.Ansi, CallingConvention = CallingConvention.Cdecl)]
+		static extern string nw_endpoint_get_url (OS_nw_endpoint endpoint);
```
viniciusjarina: Same here https://github.com/xamarin/xamarin-macios/pull/7091/files/a5fdf0e6b004dfaea4643c3d095e9fb5b2456646#r328214811

src/Network/NWDataTransferReport.cs
mandel-macaque: @stephen-hawley I'm going to add and indexer, which is nice to have :)

tests/monotouch-test/Network/NWTxtRecordTest.cs
```diff
@@ -0,0 +1,168 @@
...
+	[Preserve(AllMembers = true)]
+	public class NWTxtRecordTest
+	{
+		NWTxtRecord record;
+		string randomKey = "MyData";
+		[SetUp]
+		public void SetUp ()
+		{
+			record = NWTxtRecord.CreateDictionary ();
+            record.SetValue (randomKey, new byte[3] { 0, 0, 0 });
```
spouliot: minor: alignment (tab/spaces?)

tests/monotouch-test/Network/NWTxtRecordTest.cs
```diff
@@ -0,0 +1,168 @@
...
+		public void TestFromBytes ()
+		{
+			// get the raw data from the dictionary create txt record, and recreate a new one
+			ReadOnlyMemory<byte> rawData = null;
+			record.GetRawBytes (
+				(d) => {
+					rawData = d;
+				}
+			);
+            Assert.AreNotEqual (0, rawData.Length, "Raw data length.");
```
spouliot: minor: alignment (tab/spaces?)

tests/monotouch-test/Network/NWTxtRecordTest.cs
```diff
@@ -0,0 +1,168 @@
...
+			// do the required conversion
+			using (var otherRecord = NWTxtRecord.FromBytes (rawData))
+	            Assert.IsTrue (record.Equals (otherRecord), "Equals");
```
spouliot: minor: alignment (tab/spaces?)

tests/monotouch-test/Network/NWTxtRecordTest.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		[TearDown]
+		public void TearDown ()
+		{
+			record?.Dispose ();
```
spouliot: `record` can't be `null` or `SetUp` would throw an NRE

src/Network/NWBrowserDescriptor.cs
rolfbjarne: We could create automated tests for this (https://github.com/xamarin/xamarin-macios/issues/7124).

src/Network/NWConnection.cs
```diff
@@ -503,5 +504,44 @@ public void Batch (Action method)
...
+		}
+
+		[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
+		public void GetEstablishmentReport (DispatchQueue queue, Action<NWEstablishmentReport> handler)
+		{
+			if (queue == null)
+				throw new ArgumentNullException (nameof (queue));
+
+			unsafe {
+				if (handler == null) {
```
rolfbjarne: Can the handler really be null? That would make the API quite useless.

tests/monotouch-test/Network/NWTxtRecordTest.cs
mandel-macaque: I can't but I don't trust people that change the code to be honest.

src/Network/NWWebSocketMetadata.cs
```diff
@@ -0,0 +1,122 @@
...
+			set => nw_ws_metadata_set_close_code (GetCheckedHandle (), value);
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern NWWebSocketOpCode nw_ws_metadata_get_opcode (OS_nw_protocol_metadata metadata);
+
+		public NWWebSocketOpCode OpCode => nw_ws_metadata_get_opcode (GetCheckedHandle ());
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_ws_metadata_set_pong_handler (OS_nw_protocol_metadata metadata, dispatch_queue_t client_queue, void *pong_handler);
```
rolfbjarne:
```suggestion
	static extern void nw_ws_metadata_set_pong_handler (OS_nw_protocol_metadata metadata, dispatch_queue_t client_queue, ref BlockLiteral pong_handler);
```

This would also require related changes elsewhere.

tests/monotouch-test/Network/NWTxtRecordTest.cs
spouliot: I'd rather have the test fail if someone made such a change without running them - clearly worth investigating

tests/monotouch-test/Network/NWWebSocketMetadataTest.cs
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		[Test]
+		public void TestConstructor ()
+		{
+			foreach (var opCode in new [] { NWWebSocketOpCode.Binary, NWWebSocketOpCode.Close, NWWebSocketOpCode.Cont, NWWebSocketOpCode.Invalid, NWWebSocketOpCode.Ping, NWWebSocketOpCode.Pong, NWWebSocketOpCode.Text }) {
+				Assert.DoesNotThrow (() => {
+					var newMetadata = new NWWebSocketMetadata (opCode);
+					newMetadata.Dispose ();
+				});
```
spouliot: add label so we can know which `opCode` failed

tests/monotouch-test/Network/NWWebSocketMetadataTest.cs
```diff
@@ -0,0 +1,74 @@
...
+			}
+		}
+
+		[Test]
+		public void TestCloseCodeProperty ()
+		{
+
+		}
```
spouliot: I know we don't publish *close code* but it seems missing ;-)

tests/monotouch-test/Network/NWWebSocketOptionsTest.cs
```diff
@@ -0,0 +1,104 @@
...
+		[SetUp]
+		public void SetUp ()
+		{
+			options = new NWWebSocketOptions (NWWebSocketVersion.Version13);
+		}
+
+		[TearDown]
+		public void TearDown ()
+		{
+			options?.Dispose ();
```
spouliot: same - remove `?`, if some test mess with it we need to know (since the order of execution is not guaranteed)

src/Network/NWBrowser.cs
```diff
@@ -0,0 +1,182 @@
...
+
+		public void Cancel ()
+		{
+			lock (startLock) {
+				nw_browser_cancel (GetCheckedHandle ());
+				started = false;
+			}
+		}
+
+		public bool Started  {
```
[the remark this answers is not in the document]
mandel-macaque: Active is much better.

src/Network/NWBrowser.cs
```diff
@@ -45,22 +45,20 @@ public NWBrowser (NWBrowserDescriptor descriptor, NWParameters parameters)
 			if (descriptor == null)
 				throw new ArgumentNullException (nameof (descriptor));
 
-			InitializeHandle (nw_browser_create (descriptor.Handle, parameters?.Handle ?? IntPtr.Zero));
+			InitializeHandle (nw_browser_create (descriptor.Handle, parameters?.GetHandle () ?? IntPtr.Zero));
```
spouliot: -> simply `parameters.GetHandle ()`
that's an extension method that does the same, i.e. `return self == null ? IntPtrZero : self.Handle;`

src/Network/NWBrowser.cs
```diff
@@ -45,22 +45,20 @@ public NWBrowser (NWBrowserDescriptor descriptor, NWParameters parameters)
...
 
-		public DispatchQueue DispatchQueue {
-			set {
-				if (value == null)
-					throw new ArgumentNullException (nameof (value));
-				lock (startLock) {
-					nw_browser_set_queue (GetCheckedHandle (), value.Handle);
-					queueSet = true;
-				}
+		public void SetDispatchQueue (DispatchQueue queue){
```
spouliot: minor: `{` on next line

src/Network/NWBrowser.cs
```diff
@@ -119,28 +120,26 @@ static void TrampolineChangesHandler (IntPtr block, IntPtr oldResult, IntPtr new
...
-					}
-					BlockLiteral block_handler = new BlockLiteral ();
-					BlockLiteral *block_ptr_handler = &block_handler;
-					block_handler.SetupBlockUnsafe (static_ChangesHandler, value);
-					try {
-						nw_browser_set_browse_results_changed_handler (GetCheckedHandle (), (void*) block_ptr_handler);
-					} finally {
-						block_handler.CleanupBlock ();
-					}
+		public void SetChangesHandler (Action<NWBrowseResult, NWBrowseResult> handler) {
```
spouliot: minor: `{` on next line

src/Network/NWFramer.cs
```diff
@@ -24,8 +24,8 @@
 namespace Network {
 
 	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
-	public class NWFramer : NWProtocolMetadata {
-		public NWFramer (IntPtr handle, bool owns) : base (handle, owns) {}
+	public class NWFramer : NativeObject /* : NWProtocolMetadata */ {
```
spouliot: why comment ?

tests/monotouch-test/Network/NWBrowserTest.cs
```diff
@@ -0,0 +1,91 @@
...
+
+		[Test]
+		public void TestDispatchQueuPropertyNull () => Assert.Throws<ArgumentNullException> (() => { browser.DispatchQueue = null; });
+
+		[Test]
+		public void TestStart ()
+		{
+			browser.Start ();
+			Assert.IsTrue (browser.Started);
+			browser.Cancel ();
```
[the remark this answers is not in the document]
spouliot: ^^^ check for `IsActive` to be `false` after `Cancel` is called

src/Network/NWFramer.cs
mandel-macaque: @spouliot for me to remember, but I can remove it

spouliot: it's not about removal - it's about _why_ ? :)
if it's a subclass of `NWProtocolMetadata` then why not make it so today ?
if not then why keep it as a comment ?

PR #7096, opened 2019-09-25
Use the mono archive again.

rolfbjarne:


rolfbjarne:
build

tools/install-source/Program.cs
```diff
@@ -207,10 +207,7 @@ public static int Main (string[] arguments)
 			}
 			if (verbose)
 				Console.WriteLine ($"Original source is {src}");
-			string fixedSource = src;
-
-			if (!File.Exists (fixedSource))
-				fixedSource = mangler.GetSourcePath (src);
+			var fixedSource = mangler.GetSourcePath (src);
```
spouliot: out of curiosity, why is this needed now ?

rolfbjarne: It's a revert of this fix, which turned out to be wrong: cf126061a98b81300ded3974b60817af0efcbff4.

rolfbjarne:
I skipped the api comparison because when checking out the previous commit to build it, the previous hash was the mono archive disabled, and the api comparison tries to find mono-api-info.exe from the mono submodule, which ends up failing because the mono submodule isn't there.

PR #7097, opened 2019-09-25
[mtouch] Tweak watchOS architecture test to build for bitcode when building with llvm. Fixes xamarin/maccore#1994.

rolfbjarne:
Tweak the watchOS architecture test to build for bitcode when building with
llvm, since that's what's usually done. This makes mtouch stop complaining about having to enable device-specific builds.

Also tweak the MT0145 error message a bit.

Fixes https://github.com/xamarin/maccore/issues/1994.


lewurm:
Added a mental note add the `run-mtouch-tests` label next time. Thank you!

PR #7125, opened 2019-09-27
Bump mono 2019-08

VincentDondain:
This is to match Xamarin.Android

New commits in mono/mono:

* mono/mono@d8441deead6 [crashing] Remove Mono signal handlers when starting to handle a crash (#17078)
* mono/mono@7ae0f38fd7e [runtime] Respect runtime_version in mono_init_internal (#17086)
* mono/mono@61ca7fe3a46 [sdks] Bump min iOS version to 7.0. (#17070)

Diff: https://github.com/mono/mono/compare/c99adb97edc6a57d934105d5c13506518134b585..d8441deead62b3af591c6daefb5013b09ef1ab8b

- Fixes https://github.com/xamarin/maccore/issues/1996

PR #7127, opened 2019-09-27
[d16-4] Bump mono 2019-08@d8441dee

VincentDondain:
New commits in mono/mono:

* mono/mono@d8441deead6 [crashing] Remove Mono signal handlers when starting to handle a crash (#17078)
* mono/mono@7ae0f38fd7e [runtime] Respect runtime_version in mono_init_internal (#17086)
* mono/mono@61ca7fe3a46 [sdks] Bump min iOS version to 7.0. (#17070)

Diff: https://github.com/mono/mono/compare/c99adb97edc6a57d934105d5c13506518134b585..d8441deead62b3af591c6daefb5013b09ef1ab8b

- Fixes https://github.com/xamarin/maccore/issues/1996


dalexsoto:
build

mandel-macaque:
We should at least have linked to https://github.com/xamarin/maccore/issues/1996 @rolfbjarne does it mean we can close the issue since the condition is removed?

dalexsoto:
Looks like we ran out of memory 

> Error: Garbage collector could not allocate 1875002 bytes of memory for cardtable-mod-union.

dalexsoto:
This looks random since its counterpart PR went green https://github.com/xamarin/xamarin-macios/pull/7125

VincentDondain:
@mandel-macaque have you seen that during monitoring? Maybe we need to file a new issue against mono

mandel-macaque:
@VincentDondain  nope, this is not an issue that happens in the current branches, the issue is that there is a new tests.It does look like it is specially a test that does fail on 32 because is too much memory. Need to indemnify the issue, ignore it JUST on 32b and at the same time file an issue with mono. This will have to be done in all the branches that for the same bump.

VincentDondain:
@dalexsoto @mandel-macaque file mono issue here: https://github.com/mono/mono/issues/17129. As Alex said it seems random but I will still ignore the test. Though for the sake of expediting the insertions let's merge this now and ignore the test in a following PR.

PR #7140, opened 2019-09-30
[SceneKit] Update for Xcode 11

VincentDondain:
And `Xcode 11.1` by extension but nothing changed.

Note: it's the first time `Vector3` are used in SceneKit APIs (usually it's `SCNVector3`).

It's debatable if we *actually* want to expose those APIs but it doesn't look like there are alternatives to those APIs.
In the case of other simd specific APIs using `Vector3` or `Vector4`, we did not port them because they had an `SCNVector` equivalent (the simd being only more performant for Apple, not for us).


src/scenekit.cs
```diff
@@ -1340,6 +1378,83 @@ interface SCNLight : SCNAnimatable, SCNTechniqueSupport, NSCopying, NSSecureCodi
...
+
+		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[Export ("probeExtents", ArgumentSemantic.Assign)]
+		Vector3 ProbeExtents {
+			[MarshalDirective (NativePrefix = "xamarin_simd__", Library = "__Internal")]
+			get;
+			[MarshalDirective (NativePrefix = "xamarin_simd__", Library = "__Internal")]
+			set;
+		}
+
```
prmallic: Noob question - What does the `MarshalDirective` do here? Does `NativePrefix = "xamarin_simd__"` mean that this binding has something to do AOT compilation?

src/scenekit.cs
```diff
@@ -3894,6 +4080,13 @@ interface SCNTechnique : SCNAnimatable, NSCopying, NSSecureCoding {
 		[iOS (9,0)][Mac (10,11)]
 		[Internal, Export ("setObject:forKeyedSubscript:")]
 		void _SetObject ([NullAllowed] NSObject obj, INSCopying key);
+
+#if XAMCORE_2_0 // Metal is 64 bit only
```
rolfbjarne: Everything is `XAMCORE_2_0` now, so this `#if` is not needed.

src/scenekit.cs
rolfbjarne: `MarshalDirective` is a workaround we have to work around a calling convention limitation with P/Invokes.

The problem is that certain native types (such as `Vector3`) are not put in the normal registers when used as parameters to methods, instead they're passed in special SIMD registers. Unfortunately managed code (i.e. P/Invokes) doesn't know about this, so we can't call these native methods using P/Invokes because the arguments would be in the wrong registers.

Our workaround is a bit involved: we generate a native wrapper method where the parameters are passed in the expected registers, and that calls the actual native function, so that the native compiler puts the arguments in the right registers. The `MarshalDirective` attribute tells the generator that it should generate a call to the wrapper method instead of the normal P/Invoke (and that the name of the wrapper method is the name of the native function prefixed by `xamarin_simd__`, and the library where the wrapper method is located - `__Internal` means the executable itself).

prmallic: 🤯 (_takes it in slowly_)
Thanks! 😃

src/scenekit.cs
VincentDondain: True it's redundant but the PR is mergeable right now so I'm not gonna rebuild it (:
Will avoid next time.

VincentDondain:
Known watchOS issue: https://github.com/xamarin/maccore/issues/1605

```
[FAIL] NSMutableDictionary2Test.AddEntries :   ICE 1
```

PR #7141, opened 2019-10-01
os_log(3) support

wjk:
Does what the title says, really. The ASL API used in the Darwin.SystemLog class has been deprecated for several major macOS versions now (IIRC ASL was deprecated starting in 10.11). Binding this API was quite simple, and I have verified using a test app that messages do show up in Console.app as expected (and not encoding-mangled!).

This API is available on all four Apple platforms (macOS, iOS, tvOS, and watchOS), but the C# wrapper is currently only compiled for macOS and iOS. Let me know if you need it added to the other two DLLs. Thanks!


runtime/xamarin-support.m
```diff
@@ -2,12 +2,27 @@
...
+{
+	// Logging a dynamic string as-is using %{public}s is, strictly speaking,
+	// an antipattern. However, there is no way to call os_log directly from
+	// C#. It can only be called from Objective-C, and even then it requires
+	// that the format string be compiled into the binary. This really is
+	// our only option to pass a string from C# into os_log.
+	if (logger == NULL)
+		logger = OS_LOG_DEFAULT;
+
+	os_log_with_type (logger, type, "%{public}s", message);
```
Therzok: Will this contain symbolicated information for JIT-ed code? (AOT is going to be fine, I guess)

src/Darwin/OSLog.cs
```diff
@@ -0,0 +1,121 @@
...
+				throw new ArgumentNullException (nameof (subsystem));
+			if (category == null)
+				throw new ArgumentNullException (nameof (category));
+
+			handle = os_log_create (subsystem, category);
+		}
+
+		public void Log (string message)
+		{
+			if (message == null)
```
Therzok: Nit: just call `Log(OSLogLevel.Default, message);`

It'll most likely be inlined anyway

src/Darwin/OSLog.cs
```diff
@@ -0,0 +1,121 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using ObjCRuntime;
+using System.Runtime.InteropServices;
+using System.Collections.Generic;
+
+namespace Darwin {
+	public class OSLog : IDisposable, INativeObject {
```
Therzok: Nit: sealed

runtime/monotouch-support.h
```diff
@@ -11,6 +13,7 @@ void *         xamarin_timezone_get_data (const char *name, int *size);
 char **        xamarin_timezone_get_names (int *count);
 char *         xamarin_timezone_get_local_name ();
 void           xamarin_start_wwan (const char *uri);
+void           xamarin_os_log (os_log_t logger, os_log_type_t type, const char *mesage);
```
rolfbjarne:
```suggestion
void           xamarin_os_log (os_log_t logger, os_log_type_t type, const char *message);
```

src/Darwin/OSLog.cs
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	public enum OSLogLevel
+	{
+		// These values must match the os_log_type_t enum in <os/log.h>.
+		Default = 0x00,
+		Info    = 0x01,
+		Debug   = 0x02,
+		Error   = 0x10,
+		Fault   = 0x11
```
rolfbjarne:
```suggestion
		Fault   = 0x11,
```

src/Makefile
```diff
@@ -321,6 +322,7 @@ MAC_SOURCES += \
 	Obsoletes.cs \
 	Darwin/KernelNotification.cs \
 	Darwin/SystemLog.cs \
+	Darwin/OSLog.cs \
```
rolfbjarne: Nit: order alphabetically.

src/Darwin/OSLog.cs
rolfbjarne: Inherit from `NativeObject`, and you'll get some of the Handle/Dispose/dtor logic for free, you'll only have to provide an implementation for the ` Release` (call `os_release`) and `Retain` (call `os_retain`) methods.

runtime/xamarin-support.m
rolfbjarne: No, it won't contain symbolicated information for JIT-ed code.

src/Darwin/OSLog.cs
```diff
@@ -0,0 +1,121 @@
...
+		public void Log (OSLogLevel level, string message)
+		{
+			if (message == null)
+				throw new ArgumentNullException (nameof (message));
+
+			xamarin_os_log (handle, level, message);
+		}
+	}
+
+	public enum OSLogLevel
```
rolfbjarne: `os_log_type_t` is defined to be a `uint8_t`:

```suggestion
	public enum OSLogLevel : byte
```

src/Darwin/OSLog.cs
```diff
@@ -0,0 +1,121 @@
...
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using ObjCRuntime;
+using System.Runtime.InteropServices;
+using System.Collections.Generic;
+
+namespace Darwin {
```
rolfbjarne: I'm wondering if `ObjCRuntime` wouldn't be a better namespace for this. @spouliot what do you think?

rolfbjarne:
It would also be nice to have a test in monotouch-test and/or xammac_test  (just something very simple like `OSLog.Default.Log ("Hello world from OSLog");`).

src/Darwin/OSLog.cs
spouliot: You need to add availability attributes on the type so it will apply to all public members and the p/invokes (even private ones). Our introspection tests requires them to ensure we're not exposing unavailable symbols.

It's also helpful, from the IDE, to know the availability of an API. Figuring it out at runtime is much less fun :)

src/Darwin/OSLog.cs
spouliot: hmm... we have the reflex of using `ObjCRuntime` as both `ObjC` support and `Runtime` service types :) but

* The Mac-only `SystemLog` type is already in `Darwin`;
* the `Darwin` namespace is not part of Xamarin.iOS/tvOS/watchOS

so I don't have really hard feelings on either location. I would have used `ObjCRuntime` myself but, like mentioned, more a reflex that a thoughtful decision :)

spouliot:
build

src/CoreFoundation/OSLog.cs
```diff
@@ -0,0 +1,100 @@
+//
+// OSLog.cs: Bindings to the os_log(3) API from Apple
+//
+// Authors:
+//  William Kent
+//
+// Copyright 2019, Xamarin Inc.
```
mandel-macaque: Xamarin does not longer exists :(

rolfbjarne:
build

rolfbjarne:
build

mandel-macaque:
build

tests/xtro-sharpie/common-ObjCRuntime.ignore
```diff
@@ -38,3 +38,8 @@
 ## only on iOS and macOS
 !unknown-pinvoke! object_getInstanceVariable bound
 !unknown-pinvoke! object_setInstanceVariable bound
+
+## present in <os/log.h>
+!unknown-pinvoke! os_log_create bound
```
Therzok: It's actually in Kernel.framework.

wjk: @Therzok I think you meant System.framework. Kernel.framework is usable in KEXTs only.

Therzok: I grepped for os_log_create in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform`. But yeah, seems like System.framework! Can we make xtro know about this and not have to do ignore it @rolfbjarne ?

Therzok: Gotcha!

mandel-macaque:
Failure is known: https://github.com/xamarin/maccore/issues/581

src/CoreFoundation/OSLog.cs
```diff
@@ -0,0 +1,95 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using ObjCRuntime;
+using System.Runtime.InteropServices;
+using System.Collections.Generic;
+
+namespace CoreFoundation {
+	[Mac (10,12), iOS (10,0)]
```
spouliot: the API are included in tvOS and watchOS (see API diff)
unless it was present in tvOS 9.0 (which I doubt) then the attribute `[TV (10,0)]` is missing
same for watchOS (was it in 2.0?)

src/CoreFoundation/OSLog.cs
```diff
@@ -0,0 +1,95 @@
...
+		public void Log (OSLogLevel level, string message)
+		{
+			if (message == null)
+				throw new ArgumentNullException (nameof (message));
+
+			xamarin_os_log (Handle, level, message);
+		}
+	}
+
+	public enum OSLogLevel : byte
```
spouliot: not a big deal but we normally put availability attributes on enums

runtime/xamarin-support.m
```diff
@@ -8,6 +8,20 @@
...
```
spouliot: https://developer.apple.com/documentation/os/os_log_with_type
```
iOS 10.0+
macOS 10.12+
Mac Catalyst 13.0+
tvOS 10.0+
watchOS 3.0+
```

mandel-macaque:
build

mandel-macaque:
@wjk building again just because we want to make sure that the CI passes before merging. Once done I'll land it.

Thanks A LOT for the work!

mandel-macaque:
Same known failure xamarin/maccore#581 and a timeout with msbuild tests that is not related. Merging.

PR #7154, opened 2019-10-02
[monotouch-test] Fix UrlProtocolTest after a server we don't control changed its output. Fixes xamarin/maccore#2006.

rolfbjarne:
microsoft.com is doing user agent sniffing, and broke our our test since their
output is now different. Switch to example.com instead.

Fixes https://github.com/xamarin/maccore/issues/2006.


dalexsoto:
@monojenkins backport d16-4

spouliot:

rolfbjarne:
@monojenkins backport d16-3

PR #7158, opened 2019-10-02
[Tests] Ignore some tests when using the interpreter

mandel-macaque:
Some tests do fail with the interpreter on tvOS. Ignoring them only
there.

Fixes https://github.com/xamarin/maccore/issues/1909


mandel-macaque:
Failures are kwon and have been fixed already: https://github.com/xamarin/xamarin-macios/pull/7154

PR #7161, opened 2019-10-03
Fix version numbers in FrameworkList.xml files

akoeplinger:
The existing test only checked that an assembly was mentioned in the file, but not that its version etc matches.
Updated the test and fixed the differences.


akoeplinger:
@monojenkins backport d16-4

PR #7165, opened 2019-10-03
[prototype][linker] Do not mark NSObject subclasses from Xamarin.Forms.* assemblies

spouliot:


spouliot:
@VincentDondain waiting on feedback, tests, numbers... from XF

PureWeen:
@spouliot So for this one I'm wondering if we can just start the work required to at least opt in linking on "Xamarin.Forms.Platform.iOS"

I realize we're wanting to see a few gains from this to validate the effort
- IPA size changes
- startup performance
- Getting rid of the UIWebView warning message

But the work to make Xamarin.Forms properly link out unused types to affect the IPA is going to be a medium to high effort as our current strategy for this doesn't work

But with this package I've already tested and the necessary WebViewRenderer gets linked out
https://github.com/xamarin/Xamarin.Forms/blob/c57b2064eacb8edb2ba24982e068e14d30bae6e4/Xamarin.Forms.Platform.iOS/Renderers/WebViewRenderer.cs#L13

I feel like that's enough motivation to make linking work on at least "Xamarin.Forms.Platform.iOS" work

Especially given the considerable interest on the issue here
https://github.com/xamarin/Xamarin.Forms/issues/7323

If that seems valid can we modify this to just opt in "Xamarin.Forms.Platform.iOS" and then we can make sure that all still works from our side?

jfversluis:
In regard to the WebView problem I can confirm that his works with the pkgs from this issue and the NuGet produced by the [PR on our side](https://github.com/xamarin/Xamarin.Forms/pull/8001) that removes the last of the references to the UIWebView(Renderer).

This does two out of the three bullets mentioned by Shane in the comment above:
- IPA size changes
- Getting rid of the UIWebView warning message

The test app I have used for this was just showing a web view and primarily focused on the warning going away, but the IPA size changed as well. Before this change it was 8.8 MB now it dropped to 7.8 MB. That is a 12-13% drop in size.

I didn't really investigate the startup time and I'm note sure if that is something that can be done by this very small testing app, but I can only imagine it _must_ have impact there as well.

PureWeen:
@chamons @spouliot @samhouts @jfversluis 

Using a very simple app that consists of only a Xamarin.Forms ContentPage and a Button.
I removed all of our RenderWith attributes so the final result was a platform DLL that linked out all of our renderers except for PageRenderer and ButtonRenderer

### App Size (Full Linking)
- Without this PKG  25.46 MB App Store Size 7.8 MB IPA
- With this PKG  18.88 MB App Store Size 6.3 MB IPA
- For comparison a native Xamarin.iOS app based on the MDP Template has a 5.03 MB app size and 1.9 MB IPA

I want to emphasize that I hacked our library so it would link out everything not needed to run the app in order to get an idea of potential benefits. It would take additional effort to actually see these gains even if this PR is merged.

### Startup times
Running on an iPhone 4S with iOS 9.3.5
- I noticed no obvious difference in startup times. I just did wall clock measurements and wasn't able to capture anything that seemed like an improvement or change

### Build Times using our large Control Gallery app on debug configuration with Full Linker enabled
- I didn't notice any changes in build time. The build server that has this package installed takes the same amount of time as the one that doesn't

chamons:
@PureWeen - Couldn't you just use monop or reflection to see what's left in the assemblies in question? You could also write a tool to process the output from monodis to give you a rough idea.

PureWeen:
@chamons 
>  Couldn't you just use monop or reflection to see what's left in the assemblies in question?

Maybe?  The file in question has no extension
image

I assume it's a binary file but when I drop it into dotpeek (it says invalid file) or into VSMAC to inspect it I just get a hexdump of the file. I tried giving it a ".dll" extension to maybe trigger VSMAC to read it that way but it didn't work.

So I probably have to go a bit lower level and manual to parse it out via the methods you've articulated. Though I'm thinking that level of parsing isn't necessary to validate the usefulness of this PR?

chamons:
@rolfbjarne - Any clue what that file is?

rolfbjarne:
That's the native executable.

If you want to look at which functions it contains, you can use `nm` from the command line:

```shell
nm path/to/Xamarin.Forms.Sandbox.iOS
```

But this is just a reflection of the AOT-compiled managed code. The less managed code, the less AOT-native code. If you want to look at what's left of the managed code, you can look into the temporary obj directory and find the assemblies in question before we remove the IL from them, or you can build for debug mode and look at the assemblies in the final app (we don't remove IL from assemblies in debug builds).

spouliot:
@PureWeen on Windows the easiest solution is to use ILSpy. Beside showing you what's included it can also show you how things are referenced (main reason for being kept, beside preserve attributes and xml files)

spouliot:
I'm closing this PR in favour of another one, specific to Xamarin.Forms.Platform.iOS.dll

Still we can continue to discuss here to have all informations at the same place.

> Build Times using our large Control Gallery app on debug configuration with Full Linker enabled

@PureWeen the size difference should be noticeable in build times for the above (Archive) sample you got the size difference. At least for device builds as the linker removes stuff faster than the AOT compiler can generate them.

Can you add `--timestamp --timestamp --timestamp` to the **Additional touch arguments** ?

PR #7187, opened 2019-10-07
[AppKit] Bindings - part 1

prmallic:
Adding bindings bit by bit. Part 1


src/appkit.cs
```diff
@@ -26692,4 +26692,47 @@ interface NSWindowTabGroup
...
+	[DisableDefaultCtor]
+	interface NSButtonTouchBarItem
+	{
+		[Export ("initWithIdentifier:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string identifier);
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:title:target:action:")]
+		NSButtonTouchBarItem ButtonTouchBarItemWithIdentifier (string identifier, string title, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: `ButtonTouchBarItemWithIdentifier` sounds too much like ObjC

For .net we want 
* method names that starts with a verb (action);
* method names that do not duplicate the type name  (like `ButtonTouchBarItem`);
* method names that do not duplicate the parameter names;

The last one can be a bit confusing, e.g. `buttonTouchBarItemWithIdentifier:title:target:action:` means `buttonTouchBarItem` along with arguments
* `WithIdentifier:` (a good hint is `WIth` or `For` - they, and what's after, are generally not required)
* `title:`
* `target:
* `action:`

IOW ` {method}[{arg1}:][{arg2:}]...`, not the lack of `:` after the method name.

So a good .net method name would simply be `Create`

In other cases, like if the API was not creating instance of `NSButtonTouchBarItem` then you would need to add a description, e.g. `CreateFoo`.

src/appkit.cs
```diff
@@ -26692,4 +26692,47 @@ interface NSWindowTabGroup
...
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:image:target:action:")]
+		NSButtonTouchBarItem ButtonTouchBarItemWithIdentifier (string identifier, NSImage image, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: -> `Create`

src/appkit.cs
```diff
@@ -26692,4 +26692,47 @@ interface NSWindowTabGroup
...
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:title:image:target:action:")]
+		NSButtonTouchBarItem ButtonTouchBarItemWithIdentifier (string identifier, string title, NSImage image, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: -> `Create`

src/appkit.cs
```diff
@@ -26693,6 +26693,14 @@ interface NSWindowTabGroup
 		void Remove (NSWindow window);
 	}
 
+	[Mac (10,15)]
+	[Native]
+	public enum NSTextScalingType : long
+	{
+		Standard = 0,
+		iOS
```
spouliot: minor: add a `,` after the last member

while it's not required (for the compiler) it will make smaller diff in the future (so it ease reviews)

src/appkit.cs
```diff
@@ -26704,15 +26712,15 @@ interface NSButtonTouchBarItem
 
 		[Static]
 		[Export ("buttonTouchBarItemWithIdentifier:title:target:action:")]
-		NSButtonTouchBarItem ButtonTouchBarItemWithIdentifier (string identifier, string title, [NullAllowed] NSObject target, [NullAllowed] Selector action);
+		NSButtonTouchBarItem GetButtonTouchBarItem (string identifier, string title, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: `Get` is a verb - but it sounds like you're not _getting_ something that exists but that you're _creating_ a new item

prmallic: Aah. Right! thanks!

prmallic:
I'm not completely sure of this `NSTargetTextScalingDocumentOption` binding here. The header says 
```
APPKIT_EXTERN NSAttributedStringDocumentAttributeKey const NSTextScalingDocumentAttribute API_AVAILABLE(macos(10.15), ios(13.0));  // @"TextScaling", one of the text scaling types declared above. Only affects RTF documents.  For reading methods, this indicates the type of text scaling used in the returned attributed string.  For write methods, this attribute can be used in two different ways: (1) Passing only this attribute will overwrite the text scaling metadata in the document, but will not perform any conversion on the font sizes in the document, or (2) Passing both this attribute and NSSourceTextScalingDocumentAttribute will convert the font sizes in the document from the source text scaling type to the text scaling type specified by this attribute.
```
 which is the earlier added `NSTextScalingType` enum. 
Is this the right way to go about it?

src/Foundation/NSAttributedString.mac.cs
```diff
@@ -296,6 +296,14 @@ public partial class NSAttributedStringDocumentAttributes : DictionaryContainer
...
 			set { SetNumberValue (NSStringAttributeKey.NSTimeoutDocumentOption, (float?)value); }
 		}
+
+		public NSTextScalingType SourceTextScaling {
+			get { return (NSTextScalingType)GetLongValue (NSStringAttributeKey.NSSourceTextScalingDocumentOption); }
+		}
+
+		public NSTextScalingType TargetTextScaling {
+			get { return (NSTextScalingType)GetLongValue (NSStringAttributeKey.NSTargetTextScalingDocumentOption); }
+		}
```
prmallic: The header says - 
```
APPKIT_EXTERN NSAttributedStringDocumentReadingOptionKey const NSTargetTextScalingDocumentOption API_AVAILABLE(macos(10.15), ios(13.0));  // @"TargetTextScaling", one of the text scaling types declared above.  Only affects RTF documents.  For reading methods, you can pass this option to request that the returned attributed string uses the specified scaling.  The font point sizes in the document may be converted if necessary.  If this option is not provided, the system will deduce the target text scaling type based on application framework and platform.
```
which i believe refers to the `NSTextScalingType` enum defined here.

src/appkit.cs
```diff
@@ -2109,10 +2109,12 @@ interface NSButtonCell {
 		[Export ("keyEquivalentModifierMask")]
 		NSEventModifierMask KeyEquivalentModifierMask { get; set; }
 	
-		[Export ("keyEquivalentFont", ArgumentSemantic.Retain)]
+		[NullAllowed, Export ("keyEquivalentFont", ArgumentSemantic.Strong)]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "The keyEquivalentFont property is no longer used. It always returns the NSButtonCell's font, and setting it has no effect.")]
```
spouliot: Our API is named `KeyEquivalentFont` (casing) and it's also not formatted correctly. Also you cannot copy/paste messages from the header files.

src/appkit.cs
```diff
@@ -2109,10 +2109,12 @@ interface NSButtonCell {
...
 		NSFont KeyEquivalentFont { get; set; }
 	
 		[Export ("setKeyEquivalentFont:size:")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "The keyEquivalentFont property is no longer used. Setting it has no effect.")]
```
spouliot: same

src/appkit.cs
```diff
@@ -2717,6 +2719,7 @@ interface NSClipView {
 		void ViewBoundsChanged (NSNotification  notification);
 	
 		[Export ("copiesOnScroll")]
+		[Deprecated (PlatformName.MacOSX, 100000, 0, message: "NSClipView will always minimize the area of the document view that is invalidated. To force invalidation of the document view, use SetNeedsDisplayInRect.")]
```
spouliot: incorrect version (Apple did not update it), if that was added in Xcode 11[.x] then it was deprecated in `10,15`

src/appkit.cs
```diff
@@ -3062,6 +3065,7 @@ partial interface NSCollectionViewDelegate {
 		bool CanDragItems (NSCollectionView collectionView, NSIndexSet indexes, NSEvent evt);
 
 		[Export ("collectionView:writeItemsAtIndexes:toPasteboard:")]
+		[Deprecated (PlatformName.MacOSX, 100000, 0, message: "Use GetPasteboardWriter instead")]
```
spouliot: same

src/appkit.cs
```diff
@@ -3087,6 +3091,7 @@ partial interface NSCollectionViewDelegate {
 
 		[Mac (10,11)]
 		[Export ("collectionView:writeItemsAtIndexPaths:toPasteboard:")]
+		[Deprecated (PlatformName.MacOSX, 100000, 0, message: "Use GetPasteboardWriter instead")]
```
spouliot: same

src/appkit.cs
```diff
@@ -9039,6 +9044,7 @@ partial interface NSOutlineViewDataSource {
 		void SortDescriptorsChanged (NSOutlineView outlineView, NSSortDescriptor [] oldDescriptors);
 	
 		[Export ("outlineView:writeItems:toPasteboard:")]
+		[Deprecated (PlatformName.MacOSX, 100000, 0)]
```
spouliot: same

src/appkit.cs
```diff
@@ -17182,6 +17198,7 @@ interface NSTableViewDataSource {
 		void SortDescriptorsChanged (NSTableView tableView, NSSortDescriptor [] oldDescriptors);
 	
 		[Export ("tableView:writeRowsWithIndexes:toPasteboard:")]
+		[Deprecated (PlatformName.MacOSX, 100000, 0, message: "Use GetPasteboardWriterForRow instead")]
```
spouliot: same

src/appkit.cs
```diff
@@ -26692,4 +26724,138 @@ interface NSWindowTabGroup
...
+	[DisableDefaultCtor]
+	interface NSButtonTouchBarItem
+	{
+		[Export ("initWithIdentifier:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string identifier);
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:title:target:action:")]
+		NSButtonTouchBarItem Create (string identifier, string title, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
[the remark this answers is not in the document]
prmallic: NSTouchBarItemIdentifier is a typedef of NSString - https://developer.apple.com/documentation/appkit/nstouchbaritemidentifier?language=objc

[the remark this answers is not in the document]
wachaudh: There's a `NSTouchBarItemIdentifier` enum already defined in appkit.cs (line:19831). I think setting the identifier type to `NSTouchBarItemIdentifier` makes a lot more sense than using a string.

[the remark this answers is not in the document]
prmallic: nice spotting! thanks!

src/appkit.cs
```diff
@@ -14538,6 +14544,7 @@ interface NSSplitViewController : NSSplitViewDelegate, NSUserInterfaceValidation
 		// 'new' since it's inlined from NSSplitViewDelegate as this instance needs [RequiresSuper]
 		[RequiresSuper]
 		[Export ("splitView:shouldCollapseSubview:forDoubleClickOnDividerAtIndex:")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "This delegate method is never called, and NSSplitViewController's implementation always returns NO.")]
```
spouliot: `NO` -> `false` (`NO` is ObjC)

src/appkit.cs
```diff
@@ -21275,12 +21292,15 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
 		NSNotificationCenter NotificationCenter { get; }
 		
 		[Export ("openFile:"), ThreadSafe]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use the 'OpenURL' method instead.")]
```
spouliot: it's named `OpenUrl` (casing)

src/Foundation/NSAttributedString.mac.cs
spouliot: are you sure they are ready-only ? that does not seem very useful

prmallic: Its a `const` actually

src/appkit.cs
```diff
@@ -26692,4 +26724,138 @@ interface NSWindowTabGroup
...
+	[DisableDefaultCtor]
+	interface NSButtonTouchBarItem
+	{
+		[Export ("initWithIdentifier:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string identifier);
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:title:target:action:")]
+		NSButtonTouchBarItem Create (NSTouchBarItemIdentifier identifier, string title, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: Using `NSTouchBarItemIdentifier` is much better but it's not the right syntax.

The API accept `NSString` not an `enum`. To use our _smart _ (field-backed) `enum` you need to use `[BindAs]`, e.g.

```
NSButtonTouchBarItem Create ([BindAs (typeof (NSTouchBarItemIdentifier))] NSString identifier, ...
```

This will instruct the generator to add the conversion logic (from `enum` to `NSString`) in the generated code.

You can look it up by finding the file `NSButtonTouchBarItem.g.cs` (under `/src/build/`) and see how it differs from what you have right now.

src/appkit.cs
```diff
@@ -26692,4 +26724,138 @@ interface NSWindowTabGroup
...
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:image:target:action:")]
+		NSButtonTouchBarItem Create (NSTouchBarItemIdentifier identifier, NSImage image, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: same

src/appkit.cs
```diff
@@ -26692,4 +26724,138 @@ interface NSWindowTabGroup
...
+
+		[Static]
+		[Export ("buttonTouchBarItemWithIdentifier:title:image:target:action:")]
+		NSButtonTouchBarItem Create (NSTouchBarItemIdentifier identifier, string title, NSImage image, [NullAllowed] NSObject target, [NullAllowed] Selector action);
```
spouliot: same

src/appkit.cs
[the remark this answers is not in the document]
spouliot: That's great advice 👍 but using the smart enum needs to be bound differently (see my comment)

PR #7202, opened 2019-10-09
[WatchKit] Add WatchKit bindings for Xcode 11.2

wachaudh:
Add WatchKit bindings for Xcode 11.2

### DIFF
--- a/src/watchkit.cs
+++ b/src/watchkit.cs
@@ -515,6 +515,14 @@ namespace WatchKit {
 		[Watch (6,0)][NoiOS]
 		[Export ("supportsAudioStreaming")]
 		bool SupportsAudioStreaming { get; }
+        
+		[Watch (6,1)][NoiOS]
+		[Export ("enableWaterLock")]
+		void EnableWaterLock ();
+
+		[Watch (6,1)][NoiOS]
+		[Export ("isWaterLockEnabled")]
+		bool IsWaterLockEnabled { get; }
 	}
 
 	[iOS (8,2)]
@@ -704,6 +712,18 @@ namespace WatchKit {
 
 		[Export ("removeAllAnnotations")]
 		void RemoveAllAnnotations ();
+
+		[Watch (6,1)]
+		[Export ("setShowsUserHeading:")]
+		void SetShowsUserHeading (bool showsUserHeading);
+
+		[Watch (6,1)]
+		[Export ("setShowsUserLocation:")]
+		void SetShowsUserLocation (bool showsUserLocation);
+
+		[Watch (6,1)]
+		[Export ("setUserTrackingMode:animated:")]
+		void SetShowsUserLocation (WKInterfaceMapUserTrackingMode mode, bool animated);
 	}
 
 	[iOS (8,2)]
@@ -983,6 +1003,7 @@ namespace WatchKit {
 
 		[Watch (4,0)]
 		[Export ("enableWaterLock")]
+		[Deprecated (PlatformName.WatchOS, 6,1, message: "Use WKInterfaceDevice.enableWaterLock")]
 		void EnableWaterLock ();
 
 		[Watch (6,0)]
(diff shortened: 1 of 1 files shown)


src/watchkit.cs
```diff
@@ -1576,6 +1597,20 @@ enum WKInterfaceVolumeControlOrigin : long {
...
 	}
 
+	[Watch (6,1), NoiOS]
+	[Native]
+	enum WKInterfaceAuthorizationAppleIDButtonStyle: long {
+		Default,
+		White,
+	}
+
+	[Watch (6,1), NoiOS]
```
prmallic: Is the `NoiOS` necessary?

src/watchkit.cs
```diff
@@ -1637,9 +1672,13 @@ interface WKExtendedRuntimeSession {
 	[BaseType (typeof (WKInterfaceObject), Name = "WKInterfaceAuthorizationAppleIDButton")]
 	[DisableDefaultCtor] // Handle is `nil`
 	interface WKInterfaceAuthorizationAppleIdButton {
-
 		[Export ("initWithTarget:action:")]
+		[Deprecated (PlatformName.WatchOS, 6,1, message: "Use initWithStyle:target:action:")]
```
prmallic: I think we have to use the C# method names here instead of the ObjC selectors

src/watchkit.cs
wachaudh: Yes, the parent interface has properties/methods that also support iOS.

src/watchkit.cs
wachaudh: Yup, not sure what the convention is for referring to a c# constructor that's replacing the existing constructor. cc: @spouliot

src/watchkit.cs
```diff
@@ -704,6 +712,18 @@ interface WKInterfaceMap {
...
+		[Export ("setShowsUserHeading:")]
+		void SetShowsUserHeading (bool showsUserHeading);
+
+		[Watch (6,1)]
+		[Export ("setShowsUserLocation:")]
+		void SetShowsUserLocation (bool showsUserLocation);
+
+		[Watch (6,1)]
+		[Export ("setUserTrackingMode:animated:")]
+		void SetShowsUserLocation (WKInterfaceMapUserTrackingMode mode, bool animated);
```
spouliot: wrong name -> should be `SetUserTrackingMode`

src/watchkit.cs
```diff
@@ -983,6 +1003,7 @@ interface WKExtension {
 
 		[Watch (4,0)]
 		[Export ("enableWaterLock")]
+		[Deprecated (PlatformName.WatchOS, 6,1, message: "Use WKInterfaceDevice.enableWaterLock")]
```
spouliot: Use C# (not ObjC) syntax + our format, e.g. `Use 'X' instead.` so

`"Use 'WKInterfaceDevice.EnableWaterLock ()' instead.`

src/watchkit.cs
```diff
@@ -1576,6 +1597,20 @@ enum WKInterfaceVolumeControlOrigin : long {
 		Companion,
 	}
 
+	[Watch (6,1), NoiOS]
+	[Native]
+	enum WKInterfaceAuthorizationAppleIDButtonStyle: long {
```
spouliot: `Id`, no `ID` since it's not an acronym

src/watchkit.cs
spouliot: I don't think it's needed (but there's no harm in it).
WatchKit used to be available on iOS - but it's not anymore (and there are stubs to maintain binary compatibility).

src/watchkit.cs
spouliot: I think we used `.ctor(...)` in some places but
`Use "new WKInterfaceAuthorizationAppleIdButton (WKInterfaceVolumeControlOrigin,NSObject,Selector)' instead."`
might be better

in doubt use `grep` on the `/src/` directory ;-)

spouliot: > Yup, that's a a placeholder,

Put a `do-not-merge` label if there are pending questions on your side. That way someone else won't merge it (e.g. if you're not online). You can remove the label once you have answers.

PR #7214, opened 2019-10-10
[Networking] Improve the NWBrowser API.

mandel-macaque:
Kept the API as it was but added a new handler that will be called when
all the changes from the browser have been done. The old Set* method
could be deleted in a future release and just expose the property.

### DIFF
--- a/src/Network/NWBrowser.cs
+++ b/src/Network/NWBrowser.cs
@@ -40,7 +40,8 @@ namespace Network {
 		bool queueSet = false;
 		object startLock = new Object ();
 
-		internal NWBrowser (IntPtr handle, bool owns) : base (handle, owns) {
+		internal NWBrowser (IntPtr handle, bool owns) : base (handle, owns)
+		{
 			SetChangesHandler (InternalChangesHandler);
 		}
 
@@ -130,8 +131,8 @@ namespace Network {
 			var del = BlockLiteral.GetTarget<NWBrowserChangesDelegate> (block);
 			if (del != null) {
 				// we do the cleanup of the objs in the internal handlers
-				NWBrowseResult nwOldResult = (oldResult == IntPtr.Zero)? null : new NWBrowseResult (oldResult, owns: false);
-				NWBrowseResult nwNewResult = (newResult == IntPtr.Zero)? null : new NWBrowseResult (newResult, owns: false);
+				NWBrowseResult nwOldResult = (oldResult == IntPtr.Zero) ? null : new NWBrowseResult (oldResult, owns: false);
+				NWBrowseResult nwNewResult = (newResult == IntPtr.Zero) ? null : new NWBrowseResult (newResult, owns: false);
 				del (nwOldResult, nwNewResult, completed);
 			}
 		}
@@ -149,32 +150,33 @@ namespace Network {
 			// call the cb, completed, we need to get a collection and call the cb when completed
 			var individualCb = IndividualChangesDelegate;
 			individualCb?.Invoke (oldResult, newResult);
-			lock (changesLock) {
-				var completeCb = CompleteChangesDelegate;
-				if (completeCb == null) {
-					// we do not want to keep a list of the new results if the user does not care, dipose and move on
-					// results can be null, since we could have a not old one
-					oldResult?.Dispose ();
-					newResult?.Dispose ();
-					return; 
-				}
-				// get the change, add it to the list
-				var change = NWBrowseResult.GetChanges (oldResult, newResult);
-				var result = (result: newResult, change: change);
-				// at this point, we do not longer need the old result
-				// results can be null
+			var completeCb = CompleteChangesDelegate;
+			if (completeCb == null) {
+				// we do not want to keep a list of the new results if the user does not care, dispose and move on
+				// results can be null, since we could have a not old one
 				oldResult?.Dispose ();
+				newResult?.Dispose ();
+				return; 
+			}
+			// get the change, add it to the list
+			var change = NWBrowseResult.GetChanges (oldResult, newResult);
+			var result = (result: newResult, change: change);
+			// at this point, we do not longer need the old result
+			// results can be null
+			oldResult?.Dispose ();
+			List<(NWBrowseResult result, NWBrowseResultChange change)> tmp_changes = null;
+			lock (changesLock) {
 				changes.Add (result);
 				// only call when we know we are done
 				if (completed)  {
-					completeCb.Invoke (changes);
-					// clean resources, we never cleaned the new results, therefore we need to dispose them at this stage
-					foreach (var c in changes) {
-						c.result?.Dispose ();
-					}
-					// be ready for the next collection
-					changes.Clear ();
+					tmp_changes = changes;
+					changes = new List<(NWBrowseResult result, NWBrowseResultChange change)> ();
 				}
+			}
+			if (completed) {
+				completeCb.Invoke (tmp_changes);
+				foreach (var c in tmp_changes)
+					c.result?.Dispose ();
 			}
 		}
 
@@ -199,6 +201,7 @@ namespace Network {
 
 		// let to not change the API, but would be nice to remove it in the following releases.
 #if !XAMCORE_4_0
+		[Obsolete ("Uset the 'IndividualChangesDelegate' instead.")]
 		public void SetChangesHandler (Action<NWBrowseResult, NWBrowseResult> handler) => IndividualChangesDelegate = handler;
 #endif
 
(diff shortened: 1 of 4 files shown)


src/Network/NWBrowser.cs
```diff
@@ -113,22 +117,65 @@ public NWParameters Parameters
...
-				using (var nwNewResult = new NWBrowseResult (newResult, owns: false))
-					del (nwOldResult, nwNewResult);
+				// we do the cleanup of the objs in the internal handlers
+				var nwOldResult = new NWBrowseResult (oldResult, owns: false);
+				var nwNewResult = new NWBrowseResult (newResult, owns: false);
+				del (nwOldResult, nwNewResult, completed);
+			}
+		}
+
+		public delegate void NWBrowserChangesDelegate (NWBrowseResult oldResult, NWBrowseResult newResult, bool completed);
```
spouliot: it's harder to find as an inner type

mandel-macaque: You mean move it outside the class, fair enough

mandel-macaque:
build

tests/monotouch-test/Network/NWBrowserTest.cs
```diff
@@ -78,14 +80,70 @@ public void TestStart ()
...
+					}
+					changesEvent.Set ();
+					eventsDone = true;
+
+				});
+				browser.Start ();
+				browserReady.WaitOne (30000);
+				// we are going to start a browser, which will not pick any results, because
+				// we have not listeners, in another thread, the listener will be started and
+				// changes will get the changes for a new service
```
whitneyschmidt: Maybe it's my lack of NW knowledge, but I'm not sure what's going on in this comment or the following code 🤔

src/Network/NWListener.cs
```diff
@@ -26,6 +26,8 @@ public enum NWListenerState {
 	[TV (12,0), Mac (10,14), iOS (12,0)]
 	[Watch (6,0)]
 	public class NWListener : NativeObject {
+		bool connectionHandlerWasSet = false;
+		object connectionHandlerLock = new object ();
```
whitneyschmidt: C# doesn't have a specific `Lock` type? We just create a generic `object` and use it with `lock` statements?

src/Network/NWListener.cs
```diff
@@ -163,20 +172,23 @@ static void TrampolineNewConnection (IntPtr block, IntPtr connection)
...
-				BlockLiteral *block_ptr_handler = &block_handler;
-				block_handler.SetupBlockUnsafe (static_NewConnection, callback);
-
-				try {
-					nw_listener_set_new_connection_handler (GetCheckedHandle (), (void*) block_ptr_handler);
-				} finally {
-					block_handler.CleanupBlock ();
+			lock (connectionHandlerLock) {
+				unsafe {
+					if (callback == null){
```
whitneyschmidt: super nit: add space between `)` and `{`

src/Network/NWListener.cs
rolfbjarne: Correct, it's possible to lock on any object.

tests/monotouch-test/Network/NWBrowserTest.cs
mandel-macaque: The NW API has different classes to interact with available services. With NWListener, they have chosen a name I would personally not have. We could translate NWLister to NWServer. A NWListener is an object/class that is listening to connection which we can interact with using clients, and that is the reason for the ugly name.

The NWBrowser on the hand is a class that allows to browse/find services (listeners) that are exposing a service over Bonjour. So a browser could look in our network for services of a type (titactoe.tcp.) and on a domain ("mycooldomain").

In the test we are doing the following:

1. Start a browser. At this point, we have no listeners (unless someone is exposing it in the lab) and therefore the browser cannot find any services/listeners.
2. Start a listener that is using the same type/domain pair that the browser expects. 
3. Browser picks up the new listener, and sends an event (service found).
4. Listener stops, and the service disappears.
5. The browser is not yet canceled, so it picks up that the service/listener is not longer then and returns it.

My comment is a little cryptic because it expects knowledge on all that, sorry. I'll update it with the above explanation.

src/Network/NWBrowser.cs
```diff
@@ -113,22 +121,65 @@ public NWParameters Parameters
...
+		object changesLock = new object ();
+		List<(NWBrowseResult result, NWBrowseResultChange change)> changes = new List<(NWBrowseResult result, NWBrowseResultChange change)> ();
+
+		void InternalChangesHandler (NWBrowseResult oldResult, NWBrowseResult newResult, bool completed)
+		{
+			// we allow the user to listen to both, individual changes AND complete ones, individual is simple, just
+			// call the cb, completed, we need to get a collection and call the cb when completed
+			var individualCb = IndividualChangesDelegate;
+			individualCb?.Invoke (oldResult, newResult);
+			lock (changesLock) {
```
rolfbjarne: The lock is surrounding a rather big block of code, if I understand correctly you only want to lock around accesses to the `changes` variable.

My suggestion would be to only lock when accessing the `changes` variable, and also capture the variable and hoist it outside of the lock statement when invoking the complete handler and disposing the changes (this would start where you're currently calling `changes.Add (result)`):

```csharp

List<(NWBrowseResult result, NWBrowseResultChange change)> tmp_changes;
lock (changesLock) {
    changes.Add (result);
    if (complete) {
        tmp_changes = changes;
        changes = new List<(NWBrowseResult result, NWBrowseResultChange change)> ();
    }
}
if (completed) {
    completeCb.Invoke (tmp_changes);
    foreach (var c in tmp_changes)
        c.result?.Dispose ();
}
```

In general the code inside locks should be as short as possible, and we should really try to avoid executing code we can't control (such as calling user delegates).

src/Network/NWBrowser.cs
```diff
@@ -28,14 +29,20 @@ public enum NWBrowserState {
...
+
 	[TV (13,0), Mac (10,15), iOS (13,0), Watch (6,0)]
 	public class NWBrowser : NativeObject {
 
 		bool started = false;
 		bool queueSet = false;
 		object startLock = new Object ();
 
-		internal NWBrowser (IntPtr handle, bool owns) : base (handle, owns) {}
+		internal NWBrowser (IntPtr handle, bool owns) : base (handle, owns) {
```
spouliot: minor/style: `{` on new line for methods

src/Network/NWBrowser.cs
```diff
@@ -113,22 +121,65 @@ public NWParameters Parameters
...
 		{
-			var del = BlockLiteral.GetTarget<Action<NWBrowseResult, NWBrowseResult>> (block);
+			var del = BlockLiteral.GetTarget<NWBrowserChangesDelegate> (block);
 			if (del != null) {
-				using (var nwOldResult = new NWBrowseResult (oldResult, owns: false))
-				using (var nwNewResult = new NWBrowseResult (newResult, owns: false))
-					del (nwOldResult, nwNewResult);
+				// we do the cleanup of the objs in the internal handlers
+				NWBrowseResult nwOldResult = (oldResult == IntPtr.Zero)? null : new NWBrowseResult (oldResult, owns: false);
+				NWBrowseResult nwNewResult = (newResult == IntPtr.Zero)? null : new NWBrowseResult (newResult, owns: false);
```
spouliot: minor/style: space before `?`

src/Network/NWBrowser.cs
```diff
@@ -113,22 +121,65 @@ public NWParameters Parameters
...
+		void InternalChangesHandler (NWBrowseResult oldResult, NWBrowseResult newResult, bool completed)
+		{
+			// we allow the user to listen to both, individual changes AND complete ones, individual is simple, just
+			// call the cb, completed, we need to get a collection and call the cb when completed
+			var individualCb = IndividualChangesDelegate;
+			individualCb?.Invoke (oldResult, newResult);
+			lock (changesLock) {
+				var completeCb = CompleteChangesDelegate;
+				if (completeCb == null) {
+					// we do not want to keep a list of the new results if the user does not care, dipose and move on
```
spouliot: typo: `diSpose`

src/Network/NWBrowser.cs
```diff
@@ -146,6 +197,11 @@ public void SetChangesHandler (Action<NWBrowseResult, NWBrowseResult> handler)
 			}
 		}	
 
+		// let to not change the API, but would be nice to remove it in the following releases.
+#if !XAMCORE_4_0
+		public void SetChangesHandler (Action<NWBrowseResult, NWBrowseResult> handler) => IndividualChangesDelegate = handler;
```
spouliot: add an `[Obsolete]` with the alternative
Goal is that people building without (obsolete) warnings will have an easy migration to newer profiles

mandel-macaque:

PR #7120, opened 2019-09-26
[coreimage] Update for Xcode 11[.1]

spouliot:
```
Incomplete (draft) - it's missing a bunch of `[Obsolete]` but those are
not breaking changes so the _core_ can be reviewed.

Pay attention to naming (many conflicts) and comment if you have better
names suggestions.

It's also a big PR with a lot of copy/paste so watch out for typos!
```

Apple decided to expose most (but not all) `CIFilter` using protocols
(instead of weakly named dictionaries). Most of this maps well with
our strong bindings but there are cases where we:

* missing some properties (easy, there were added); or
* used a different types [1] and new members were required

[1] Often ours are better (using `float` for a `bool` is not optimal)
but we do not have `[BindAs]` for protocols :( to _fix_ them

Issue #6785, opened 2019-08-15
Simpler bindings for [StrongDictionary]


spouliot:
We can simplify them by re-use the smart enums pattern, inlining the `[Field]` into the dictionary. E.g.

Today
```csharp
	[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
		[Internal]
		[Static]
		interface MLFeatureValueImageOptionKeys {
	
			[Field ("MLFeatureValueImageOptionCropRect")]
			NSString CropRectKey { get; }
	
			[Field ("MLFeatureValueImageOptionCropAndScale")]
			NSString CropAndScaleKey { get; }
		}
	
		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
		[StrongDictionary ("MLFeatureValueImageOptionKeys")]
		interface MLFeatureValueImageOption {
			CGRect CropRect { get; set; }
			VNImageCropAndScaleOption CropAndScale { get; set; }
		}
```

Future
```csharp
		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
		[StrongDictionary]
		interface MLFeatureValueImageOption {
			[Field ("MLFeatureValueImageOptionCropRect")]
			CGRect CropRect { get; set; }
			[Field ("MLFeatureValueImageOptionCropAndScale")]
			VNImageCropAndScaleOption CropAndScale { get; set; }
		}
```

spouliot:
We could even expose an auto-generated `NSString GetConstant (string propertyName);` method for cases where the fields can be used differently (but without exposing all of them individually, just like we do on smart enums).

dalexsoto:
That would enable something like

```csharp
var fooconst = StrongDict.GetConstant (nameof (FooProperty));
```

I like it!

PR #7216, opened 2019-10-10
[coreimage] Update for Xcode 11[.2]

spouliot:
Apple decided to expose most (but not all) `CIFilter` using protocols
(instead of weakly named dictionaries). Most of this maps well with
our strong bindings but there are cases where we:

* missing some properties (easy, there were added); or
* used a different types [1] and that requires new members / obsoletion

[1] Often ours are better (using `float` for a `bool` value is not
optimal) but we do not have `[BindAs]` for protocols :( to _fix_ them

Note: this replace draft PR https://github.com/xamarin/xamarin-macios/pull/7120 but it's has quite a bit of changes in filter generation (inlining protocols) and that affected bindings too.


src/coreimage.cs
```diff
@@ -5673,85 +5498,3365 @@ interface CIKeystoneCorrection {
...
 
 	[CoreImageFilter]
 	[iOS (13,0)]
 	[TV (13,0)]
 	[Mac (10,15)]
 	[BaseType (typeof (CIKeystoneCorrection))]
-	interface CIKeystoneCorrectionHorizontal {
+	interface CIKeystoneCorrectionHorizontal : ICIKeystoneCorrectionHorizontalProtocol {
+
+#if false // no documentation about the type
```
VincentDondain: I thought we tried not to push dead code 😉. Couldn't it just be in the xtro ignore file with a comment?

src/coreimage.cs
```diff
@@ -5673,85 +5498,3365 @@ interface CIKeystoneCorrection {
...
+	interface ICISmoothLinearGradientProtocol {}
+
+	[iOS (13,0)]
+	[TV (13,0)]
+	[Mac (10,15)]
+	[Protocol (Name = "CISmoothLinearGradient")]
+	// `CILinearGradientProtocol` is a bit of a lie - but it would not compile (registrar) otherwise
+	interface CISmoothLinearGradientProtocol : CILinearGradientProtocol {
+
+		/* we get those from ICILinearGradientProtocol
```
VincentDondain: So if we're keeping this code in case it's ever added again, wouldn't a `#if false` be better?
Similar to my comment above I don't know if we want dead code (:

src/coreimage.cs
spouliot: there's (almost) no headers for CIFilters, so no xtro reminders

_almost_ because some stuff is now defined in protocols - and those are in headers, but the rest is not (it's all key/value based)

src/CoreImage/CIFilter.cs
```diff
@@ -199,6 +221,16 @@ internal int GetInt (string key)
 			}
 		}
 
+		internal nint GetNInt (string key)
+		{
+			using (var nskey = new NSString (key)){
+				var v = ValueForKey (nskey);
+				if (v is NSNumber)
+					return (v as NSNumber).NIntValue;
```
mandel-macaque: Can it be done with just using the as? Like:

```csharp
var n = v as NSNumber;
return n?.NIntValue ?? 0;
```

src/CoreImage/CIFilter.cs
```diff
@@ -237,31 +269,16 @@ internal IntPtr GetHandle (string key)
...
-		internal CIColor GetColor (string key)
-		{
-			return ValueForKey (key) as CIColor;
-		}
-		
-		internal CIImage GetInputImage ()
+		internal CGPoint GetPoint (string key)
 		{
-			return ValueForKey (CIFilterInputKey.Image) as CIImage;
+			var v = ValueForKey (key) as CIVector;
```
mandel-macaque: as can return null, right? So unless we are 100% sure.. I'd check that.

src/CoreImage/CIFilter.cs
```diff
@@ -237,31 +269,16 @@ internal IntPtr GetHandle (string key)
...
 		{
-			SetValueForKey (value, CIFilterInputKey.Image);
-		}
-
-		internal CIImage GetImage (string key)
-		{
-			using (var nsstr = new NSString (key))
-				return ValueForKey (nsstr) as CIImage;
+			var v = ValueForKey (key) as CIVector;
+			return new CGRect (v.X, v.Y, v.Z, v.W);
```
mandel-macaque: Same.

src/coreimage.cs
```diff
@@ -1328,6 +1351,18 @@ interface CIImageInitializationOptionsKeys {
...
+		[Field ("kCIImageAuxiliarySemanticSegmentationSkinMatte")]
+		NSString AuxiliarySemanticSegmentationSkinMatteKey { get; }
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Field ("kCIImageAuxiliarySemanticSegmentationHairMatte")]
+		NSString AuxiliarySemanticSegmentationHairMatteKey { get; }
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Field ("kCIImageAuxiliarySemanticSegmentationTeethMatte")]
+		NSString AuxiliarySemanticSegmentationTeethMatteKey { get; }
```
mandel-macaque: This look like they could be a smart enum.

src/coreimage.cs
```diff
@@ -2645,29 +2788,23 @@ interface CIImageProcessorKernel {
 	[iOS (8,0)]
 	[Mac (10,10)]
 	[BaseType (typeof (CIFilter))]
-	interface CIAccordionFoldTransition {
+	interface CIAccordionFoldTransition : ICIAccordionFoldTransitionProtocol {
```
dalexsoto: @spouliot I see you are using the `I` version of `CIAccordionFoldTransitionProtocol` is that intentional? IIRC you have to use the `I` less version when doing this else the generator won't do the right thing.

Also I see this being mixed in the whole PR, for example the following does what I expect :)

`interface CIAccordionFoldTransitionProtocol : CITransitionFilterProtocol {`

src/coreimage.cs
```diff
@@ -1346,6 +1381,17 @@ interface CIImage : NSSecureCoding, NSCopying {
...
 
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		[Static]
+		[Export ("imageWithCGImageSource:index:options:")]
+		CIImage FromCGImageSource (CGImageSource source, nuint index, [NullAllowed] NSDictionary options);
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Static]
+		[Wrap ("FromCGImageSource (source, index, options == null ? null : options.Dictionary)")]
```
mandel-macaque: No need to change, but I think options?.Dictionary would be the same.

src/coreimage.cs
```diff
@@ -1483,6 +1529,15 @@ interface CIImage : NSSecureCoding, NSCopying {
 		[Wrap ("this (image, options == null ? null : options.Dictionary)")]
 		IntPtr Constructor (CGImage image, [NullAllowed] CIImageInitializationOptionsWithMetadata options);
 
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		[Export ("initWithCGImageSource:index:options:")]
+		IntPtr Constructor (CGImageSource source, nuint index, [NullAllowed] NSDictionary options);
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Wrap ("this (source, index, options == null ? null : options.Dictionary)")]
```
mandel-macaque: Same options?.Dictionary should be the same.

src/coreimage.cs
```diff
@@ -1927,6 +1986,32 @@ interface CIImage : NSSecureCoding, NSCopying {
...
+
+		[TV (13,0), iOS (13,0), Mac (10,15)]
+		[Export ("initWithSemanticSegmentationMatte:")]
+		IntPtr Constructor (AVSemanticSegmentationMatte matte);
+
+		[TV (13,0), iOS (13,0), Mac (10,15)]
+		[Static]
+		[Export ("imageWithSemanticSegmentationMatte:options:")]
+		[return: NullAllowed]
+		CIImage FromSemanticSegmentationMatte (AVSemanticSegmentationMatte matte, [NullAllowed] NSDictionary options);
```
mandel-macaque: Do we know about the options? Could it be a strong dictionary?

src/coreimage.cs
```diff
@@ -1927,6 +1986,32 @@ interface CIImage : NSSecureCoding, NSCopying {
...
 
+		// CIImage_AVSemanticSegmentationMatte
+
+		[TV (13,0), iOS (13,0), Mac (10,15)]
+		[NullAllowed, Export ("semanticSegmentationMatte")]
+		AVSemanticSegmentationMatte SemanticSegmentationMatte { get; }
+
+		[TV (13,0), iOS (13,0), Mac (10,15)]
+		[Export ("initWithSemanticSegmentationMatte:options:")]
+		IntPtr Constructor (AVSemanticSegmentationMatte matte, [NullAllowed] NSDictionary options);
```
mandel-macaque: Do we know about the options? Could it be a string dictionary?

src/coreimage.cs
```diff
@@ -5452,6 +5252,22 @@ partial interface CIImageRepresentationKeys {
...
+		[Field ("kCIImageRepresentationSemanticSegmentationSkinMatteImage")]
+		NSString SemanticSegmentationSkinMatteImageKey { get; }
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Field ("kCIImageRepresentationSemanticSegmentationHairMatteImage")]
+		NSString SemanticSegmentationHairMatteImageKey { get; }
+
+		[iOS (13,0)][TV (13,0)][Watch (6,0)][Mac (10,15)]
+		[Field ("kCIImageRepresentationSemanticSegmentationTeethMatteImage")]
+		NSString SemanticSegmentationTeethMatteImageKey { get; }
```
mandel-macaque: These look like they could be a smart enum.

src/coreimage.cs
```diff
@@ -5673,85 +5498,3365 @@ interface CIKeystoneCorrection {
...
+	// CIFilter already exists so we're using the Swift name
+	interface CIFilterProtocol {
+
+		[Abstract]
+		[NullAllowed, Export ("outputImage")]
+		CIImage OutputImage { get; }
+
+		[Static]
+		[NullAllowed, Export ("customAttributes")]
+		NSDictionary<NSString, NSObject> CustomAttributes { get; }
```
mandel-macaque: Do we know anything about the attributes? Strong dict will be nicer if possible.

src/coreimage.cs
```diff
@@ -5673,85 +5498,3365 @@ interface CIKeystoneCorrection {
...
+
+	[iOS (13,0)]
+	[TV (13,0)]
+	[Mac (10,15)]
+	[Protocol (Name = "CILinearGradient")]
+	interface CILinearGradientProtocol : CIFilterProtocol {
+
+		[Abstract]
+		[Export ("point0", ArgumentSemantic.Assign)]
+		CGPoint InputPoint0 { get; set; }
```
mandel-macaque: Why do we add 'Input'?

src/coreimage.cs
spouliot: good catch, I'll review those

`CIFilter` subclasses are generated differently that _normal_ `NSObject`so the end-result might be (nearly) identical. IOW I think my generator changes are not correct, with incorrect data giving the correct output.

src/CoreImage/CIFilter.cs
rolfbjarne:
```csharp
if (v is NSNumber n)
    return n.NIntValue;
```

spouliot: it was a copy/paste of the existing `GetInt` method, but I can fix all of them
who's counting the lines anyway :)

src/coreimage.cs
spouliot: Good catch, left over from the original attempt. I think that's fixed (properly) now. I'll uncomment it

src/generator-filters.cs
```diff
@@ -126,28 +176,57 @@ public void GenerateFilter (Type type)
...
 
+			// an export will be present (only) if it's defined in a protocol
+			var export = AttributeManager.GetCustomAttribute<ExportAttribute> (p);
+
 			var name = AttributeManager.GetCustomAttribute<CoreImageFilterPropertyAttribute> (p)?.Name;
-			if (p.GetGetMethod () != null)
+			// we can skip the name when it's identical to a protocol selector
+			if (name == null) {
+				if (export == null)
+					throw new BindingException (9999, true, $"Missing [CoreImageFilterProperty] attribute on {type.Name} property {ptype}");
```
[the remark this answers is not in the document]
spouliot: hmm... how would translating `{0}` be different from `{ptype}` ?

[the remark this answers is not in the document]
rolfbjarne: The problem isn't translating it, but using the translated string in C#. String interpolation requires the string to be a literal string, and translated strings aren't literal (https://stackoverflow.com/a/32360534/183422)

src/coreimage.cs
spouliot: uncommented (that scenario works fine now)

src/CoreImage/CIFilter.cs
spouliot: fixed in newer commits

src/coreimage.cs
spouliot: it's actually a inside a type used tp create a strong dictionary `[StrongDictionary ("CIImageInitializationOptionsKeys")]` but I did not add the keys to it :| I'll fix this

Makes me wish for https://github.com/xamarin/xamarin-macios/issues/6785 even more

src/coreimage.cs
spouliot: string ? no, strong ? maybe :)

It was not documented before but now that 
`typedef NSString *CIImageOption;`
was added we can be pretty sure it's the same as the one I mentioned above

However there's a lot of older members that also use `NSDictionary` so, to keep this PR from calling unicorns, I'll fix that in a separate PR.

src/coreimage.cs
spouliot: no, but it's part of the type used for a strong dictionary: `CIImageRepresentationOptions`

I'll look if the types are documented, we are already missing a few of those

src/coreimage.cs
spouliot: no :(
https://developer.apple.com/documentation/coreimage/cifilter/3228047-customattributes?language=objc

spouliot:
build

src/coreimage.cs
spouliot: yes, headers (not web doc) do mention the types of those (and the one we were missing from last year)

spouliot:
@mandel-macaque @dalexsoto your feedback should all be addressed now

PR #7217, opened 2019-10-10
[build] lipo x86-64-slice into libs according to their kind

lewurm:
The result of `lipo`ing `.a`s and `.dylib`s together is not well defined.
In this specific case, using `lipo` on `libmonosgen-2.0.a` to inject `x86-64-slice.dylib` resulted into `dsymutil` (run by `mtouch` when building the app) not finding the proper debug information anymore.

Also putting the dummy slice into a file and adding `-ggdb3` to avoid those warnings:

```console
$ dsymutil ./_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/SDKs/MonoTouch.iphoneos.sdk/usr/lib/libmonosgen-2.0.dylib
warning: (x86_64) /var/folders/p3/5279mmgn1p575bz28j0ngfqw0000gn/T/--19c1b5.o unable to open object file: No such file or directory
warning: no debug symbols in executable (-arch x86_64)
```

### DIFF
--- a/builds/.gitignore
+++ b/builds/.gitignore
@@ -25,3 +25,4 @@ watchbcl
 mono-ios-sdk-destdir
 *.dylib
 *.o
+x86-64-slice.*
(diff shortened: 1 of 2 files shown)


jaykrell:
Wierd.
How does anyone else build their software?
What does Xcode do?

PR #7229, opened 2019-10-14
[xcode11.1] Bump mono 2019-06@476d72b

VincentDondain:
New commits in mono/mono:

* mono/mono@476d72b [IO] Remove read-only logic in mono_w32_get_disk_free_space
* mono/mono@bdc5576 [2019-06][merp] Don't install SIGTERM handler in EnableMicrosoftTelemetry (#17296)
* mono/mono@1e2d850 [tests] Bump corefx to get Azure testhost change (#17275)
* mono/mono@703e9a2 [2019-06] Add drawing type converters to mobile profiles (#17241)
* mono/mono@4ec605f If there is a perform_wait_callback in the stack there will be another catch generated by the method called in the owner thread, so we don't need to throw as unhandled exception, we can continue and find the next catch.
* mono/mono@061c225 [merp] exit_status is 0 if we ran the uploader successfully (#17186)
* mono/mono@07c23f2 [crashing] Remove Mono signal handlers when starting to handle a crash

Diff: https://github.com/mono/mono/compare/5608fe0abb3e501c461b2979a913c0bed3fe2a6d..476d72b9e32ea58a7a8bda53af8cc8f5b7ffe6bb


VincentDondain:
Adding `do-not-merge` as we're not sure yet when this should land.

VincentDondain:
Equivalent Xamarin.Android PR: https://github.com/xamarin/xamarin-android/pull/3774

mandel-macaque:
@VincentDondain  @dalexsoto  I'm confident that we do not need to worry about those two old NUnit tests.

VincentDondain:
We do not need this PR anymore, the shipping vehicle for those mono commits was xcode11.2.

PR #7230, opened 2019-10-14
[SampleTester] Bump prebuilt-apps for SampleTester

wachaudh:
Bump prebuilt-apps for SampleTester to latest on master. Fixes have been merged to prebuilt-apps which should resolve CI failures.


rolfbjarne:
I've pushed this to the origin repo (https://github.com/xamarin/xamarin-macios/commit/bump-prebuilsamples), so that our internal jerkins instance will build it, and then trigger a sample test run.

mandel-macaque:
WatchOS failure can be ignored is the HE0038

PR #7238, opened 2019-10-16
Bumping mono might require re-creating Make.config.inc.

rolfbjarne:
If the new mono commit has (or doesn't have) mono archives.


rolfbjarne:
Test failure is unrelated

PR #7252, opened 2019-10-17
[Network] Add more convinient callbacks.

mandel-macaque:
Providing callbacks with DispatchData and Int + length is not nice for
the non advance users. Added set of methods that will use callbacks that
will provide the users with ReadOnlySpan, which is nicer and makes it
work with the API easier.

Also, the change allows users to avoid unsafe code.

### DIFF
--- a/src/Network/NWConnection.cs
+++ b/src/Network/NWConnection.cs
@@ -372,7 +372,7 @@ namespace Network {
 		}
 
 		[BindingImpl (BindingImplOptions.Optimizable)]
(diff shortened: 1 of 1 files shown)


mandel-macaque:
I agree on the need of a test, but lets wait until we have NWFramer which will simplify the test of all the callbacks in the NWConnection class.

mandel-macaque:
build

mandel-macaque:
Failures are unrelated to the change.

PR #7256, opened 2019-10-17
[Network] Add NWFramerMessage implementation.

mandel-macaque:
Adds the needed object for the NWFramer AND deprecates some factory methods that were properties.


src/Network/NWProtocolDefinition.cs
```diff
@@ -44,28 +44,75 @@ public bool Equals (object other)
 		[DllImport (Constants.NetworkLibrary)]
 		static extern OS_nw_protocol_definition nw_protocol_copy_ip_definition ();
 
+		[Obsolete ("Use 'CreateIPDefinition' method instead.")]
```
rolfbjarne: Use `XAMCORE_4_0` to remove obsolete API in the future.

src/Network/NWFramerMessage.cs
```diff
@@ -0,0 +1,106 @@
...
+
+		public void SetData (string key, ReadOnlySpan<byte> value)
+		{
+			// the method takes a callback to cleanup the data, but we do not need that since we are managed
+			if (key == null)
+				throw new ArgumentNullException (nameof (key));
+
+			unsafe {
+				fixed (byte* mh = value)
+					nw_framer_message_set_value (GetCheckedHandle (), key,  mh, null);
```
rolfbjarne: This is not safe: you're storing a pointer to memory referenced by the ReadOnlySpan, but the ReadOnlySpan will be collected soon after this function returns, and that might include the memory pointed to by the ReadOnlySpan.

You must keep the pointer passed to the framer message alive for as long needed (until either the value is overwritten or the framer message freed).

As such I don't think it's a good idea to use a ReadOnlySpan as the parameter type, because it can only live on the stack and there's no way for this code to make sure it's not collected once the call stack unwinds. I think an IntPtr version could work, but that would also require the API to have a dispose callback to notify the caller when the IntPtr can be freed:

```csharp
public void SetData (string key, IntPtr value, Action<IntPtr> dispose_value);
```

another alternative is a byte[] version:

```csharp
public void SetData (string key, byte[] data);
```

and where you store a strong GCHandle to the byte[] in the framer message (and free the GCHandle in the dispose value callback).

mandel-macaque: byte[] using the cleanup in the c call is doable and probably the way to go.

src/Network/NWFramer.cs
```diff
@@ -238,23 +250,25 @@ static void TrampolineCleanupHandler (IntPtr block, OS_nw_framer framer)
 		static extern void nw_framer_schedule_wakeup (OS_nw_framer framer, ulong milliseconds);
 
 		public void ScheduleWakeup (ulong milliseconds) => nw_framer_schedule_wakeup (GetCheckedHandle (), milliseconds);
+		*/
 
 		[DllImport (Constants.NetworkLibrary)]
 		static extern OS_nw_protocol_metadata nw_framer_protocol_create_message (OS_nw_protocol_definition definition);
 
-		public static NWFramer CreateMessage (NWProtocolDefinition protocolDefinition)
+		public static NWFramerMessage CreateMessage (NWProtocolDefinition protocolDefinition)
```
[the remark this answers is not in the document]
mandel-macaque: Well, I was doubting. Could move the constructor/factory to the NSFramerMessage

src/Network/NWProtocolDefinition.cs
```diff
@@ -44,28 +44,75 @@ public bool Equals (object other)
...
+		delegate NWFramerStartResult nw_framer_create_definition_t (IntPtr block, IntPtr framer);
+		static nw_framer_create_definition_t static_CreateFramerHandler = TrampolineCreateFramerHandler;
+
+		[MonoPInvokeCallback (typeof (nw_framer_create_definition_t))]
+		static NWFramerStartResult TrampolineCreateFramerHandler (IntPtr block, IntPtr framer)
+		{
+			// get and call, this is internal and we are trying to do all the magic in the call
+			var del = BlockLiteral.GetTarget<Func<NWFramer,NWFramerStartResult>> (block);
+			if (del != null) {
+				var nwFramer = new NWFramer (framer, owns: true);
```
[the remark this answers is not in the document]
mandel-macaque: That has to be managed by the user. The callback will provide the framer to the user, the framer is later used to read and write message that are sent to the NWConnection, so you want to create a framer per protocol definition, keep it in memory and use it during the life of the application or the usefulness of the object. Then be disposed.

If I do a dispose here, the entire API is useless, since you won't the able to write the required data to the 'wire'.

src/Network/NWFramer.cs
```diff
@@ -238,23 +250,25 @@ static void TrampolineCleanupHandler (IntPtr block, OS_nw_framer framer)
...
-			return new NWFramer (nw_framer_protocol_create_message (protocolDefinition.Handle), owns: true);
+			return new NWFramerMessage (nw_framer_protocol_create_message (protocolDefinition.Handle), owns: true);
 		}
 
 		[DllImport (Constants.NetworkLibrary)]
 		static extern OS_nw_protocol_metadata nw_framer_message_create (OS_nw_framer framer);
 
-		public NWProtocolMetadata CreateMessage ()
-			=> new NWFramer (nw_framer_message_create (GetCheckedHandle ()), owns: true);
+		public NWFramerMessage CreateMessage ()
```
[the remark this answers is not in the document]
mandel-macaque: This is not a static method, but a instance one. Needs the pointer to the framer. The one above is kept in the same class since it makes sense to group them. Would look funny to create a message in two different classes.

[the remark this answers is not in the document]
spouliot: ok, I think it's better not have `CreateMessage` as both static and instance methods (so this one is fine)

src/Network/NWProtocolDefinition.cs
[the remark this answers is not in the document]
spouliot: ok, sounds like it's not called often

src/Network/NWFramer.cs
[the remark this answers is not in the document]
spouliot: I think it's better located in the type itself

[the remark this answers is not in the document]
mandel-macaque: Ok, will move it to FramerMessage then.

mandel-macaque:
build

src/Network/NWFramer.cs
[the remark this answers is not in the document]
spouliot: Sorry, I think I changed my mind. You're right it's not `static` but that's how you chose to expose it, not how Apple did it.

The function is named `nw_framer_message_create` so the prefix makes me think it should be inside `NWFramerMessage` not here in `NWFramer`

IOW it's `nw_framer_message_create` -> `NWFramerMessage.Create`
and not `NWFramer.MessageCreate`

src/Network/NWFramerMessage.cs
```diff
@@ -0,0 +1,137 @@
...
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_protocol_metadata nw_framer_protocol_create_message (OS_nw_protocol_definition definition);
+
+		public static NWFramerMessage CreateMessage (NWProtocolDefinition protocolDefinition)
+		{
+			if (protocolDefinition == null)
+				throw new ArgumentNullException (nameof (protocolDefinition));
+			return new NWFramerMessage (nw_framer_protocol_create_message (protocolDefinition.Handle), owns: true);
+		}
```
spouliot: wrong place, sorry again :)

I think `nw_framer_protocol_create_message` should be `NWFramerProtocol.CreateMessage`

mandel-macaque: So this, is a complicated and API because it is not very consistent with the naming. For what I have been able to see, we have:

* nw_framer_protocol_create_message which is the only C call that has the nw_framer_protocol_* pattern. And that takes a OS_nw_protocol_definition which is static.
* nw_framer_ which is the pattern used by the C library when working with OS_nw_framer which is what I'm mapping to NWFramer.
* nw_framer_message_* pattern. Which maps got a NWFramerMessage which is a NWProtocolMetadata.

Looking at swift, the code is not very clear because it does a lot of magic, and they expose a NWFramerProtocol convenience class, that contains a NWFramer and hooks to all the different callbacks.  My reasoning between moving the CreateMessage **static** method to the NWFramer is because in order to work with the NWFramer, you need to be able to write NWFramerMessage (which is just a specific type of metadata) which is later passed to the NWConneciton, something like the following:

```csharp
var message = protocol.CreateMessage (GameMessageType.Move);
var context = new NWContentContext ("Move");
context.SetMetadata (message);

connection.Send (Encoding.UTF8.GetBytes (move), context, true, (e) => {
	if (e != null) {
		Console.WriteLine ($"Error sending move: error number {e.ErrorCode} domain {e.ErrorDomain}");
	}
});
```

In that case we are creating a NWFramerMessage that has a header which contains the GamerMessageType (it is added as a string in one of the headers using a key).

I would consider nw_framer_protocol_ to be similar to nw_framer_ since the naming of the API is inconsistent. Now, where to place the CreateMessage is mostly on what we want to expose as an API. In the  swift code there is no public implementation of the NWFramer, but if you look at it using Xcode you get the [following](https://gist.github.com/mandel-macaque/abda55c3b12ec50a9e5d2766a7274b0b).

There are using two types, NWFramer, and a wrapper NWProtocolFramer that contains and instance of the NWFramer and can be init with a NWProtocolDefinition. I don't want to write magic, at least not at this stage and not with the same pattern they have used (they have inner classes, so rather than a NWTlsProtocolOptions, they have an Options class inside the NWTlsProtocol). I have no preference per se, either one is good as long as we do not create NWFramerProtocol, but use NWFramer. I think that NWFramerMessage.Create (definition) is nice, even when you can create the message in two diff classes.

src/Network/NWFramerMessage.cs
```diff
@@ -0,0 +1,137 @@
...
+				return del (data);
+			}
+			return false;
+		}
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public bool GetData (string key, int dataLength, out ReadOnlySpan<byte> outData) {
+			IntPtr outPointer = IntPtr.Zero;
+			// create a function that will get the data, and the data length passed and will set the out param returning the value
+			Func<IntPtr,bool> callback = (inData) => {
```
prmallic: Minor style - `Func<IntPtr, bool>`?

src/Network/NWFramerMessage.cs
```diff
@@ -0,0 +1,137 @@
...
+			}; 
+
+			BlockLiteral block_handler = new BlockLiteral ();
+			block_handler.SetupBlockUnsafe (static_AccessValueHandler, callback);
+			try {
+				// the callback is inlined!!!
+				var found = nw_framer_message_access_value (GetCheckedHandle (), key, ref block_handler);
+				if (found) {
+					unsafe {
+						outData = new ReadOnlySpan<byte>((void*)outPointer, dataLength);
```
prmallic: Minor style - `(void*) outPointer`

src/Network/NWProtocolDefinition.cs
```diff
@@ -44,28 +44,85 @@ public bool Equals (object other)
...
+				return del (nwFramer);
+			}
+			return NWFramerStartResult.Unknown;
+		}
+
+		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public static NWProtocolDefinition CreateFramerDefinition (string identifier, NWFramerCreateFlags flags, Func<NWFramer,NWFramerStartResult> startCallback)
```
prmallic: `Func<NWFramer, NWFramerStartResult>`

src/Network/NWFramerMessage.cs
spouliot: ok :) just so people reading the source does not wonder why (and try to reach the same conclusion as you did) please add a short code comment, which can be a link back to your detailed answer here

PR #7268, opened 2019-10-21
[HomeKit] Xcode11.2 bindings

prmallic:
Bindings for HomeKit for Xcode11.2


spouliot:
Removed `no-notes-worthy` since we normally list all updated (API) frameworks in release notes (and the flag would mean the tooling will skip over the PR)

PR #7270, opened 2019-10-21
[Network] Cleanup the NWProtocolMetadata. 

mandel-macaque:
Add missing types so that users do not mix the diff protocol metadata. Also makes the API cleaner since we will have a WebSocektsMetadata and a NWFramerMessage which are of the metadata type.

### DIFF
--- a/src/Network/NWIPMetadata.cs
+++ b/src/Network/NWIPMetadata.cs
@@ -30,7 +30,7 @@ namespace Network {
 		public TimeSpan ReceiveTime {
 			get {
 				var time = nw_ip_metadata_get_receive_time (GetCheckedHandle ());
-				return (time == 0) ? TimeSpan.Zero : TimeSpan.FromTicks ((long) time / 100);
+				return TimeSpan.FromTicks ((long) time / 100);
 			}
 		}
 
--- a/src/Network/NWIPMetadata.cs
+++ b/src/Network/NWIPMetadata.cs
@@ -1,0 +1,42 @@
+//
+// NWIPMetadata.cs: Bindings the Netowrk nw_protocol_metadata_t API that is an IP.
+//
+// Authors:
+//   Manuel de la Pena <mandel@microsoft.com>
+//
+// Copyrigh 2019 Microsoft
+//
+using System;
+using ObjCRuntime;
+using Foundation;
+using CoreFoundation;
+
+namespace Network {
+
+	[TV (12,0), Mac (10,14), iOS (12,0), Watch (6,0)]
+	public class NWIPMetadata : NWProtocolMetadata {
+
+		internal NWIPMetadata (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		public NWIPMetadata () : this (nw_ip_create_metadata (), owns: true) {}
+
+		public NWIPEcnFlag EcnFlag {
+			get => nw_ip_metadata_get_ecn_flag (GetCheckedHandle ());
+			set => nw_ip_metadata_set_ecn_flag (GetCheckedHandle (), value);
+		}
+
+		// A single tick represents one hundred nanoseconds, API returns: the time at which a packet was received, in nanoseconds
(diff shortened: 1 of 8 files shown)


src/Network/NWIPMetadata.cs
```diff
@@ -0,0 +1,42 @@
...
+
+		public NWIPMetadata () : this (nw_ip_create_metadata (), owns: true) {}
+
+		public NWIPEcnFlag EcnFlag {
+			get => nw_ip_metadata_get_ecn_flag (GetCheckedHandle ());
+			set => nw_ip_metadata_set_ecn_flag (GetCheckedHandle (), value);
+		}
+
+		// A single tick represents one hundred nanoseconds, API returns: the time at which a packet was received, in nanoseconds
+		// so we get nanoseconds, divicde by 100 and use from ticks
```
rolfbjarne:
```suggestion
		// so we get nanoseconds, divide by 100 and use from ticks
```

src/Network/NWIPMetadata.cs
```diff
@@ -0,0 +1,42 @@
...
+		public TimeSpan ReceiveTime {
+			get {
+				var time = nw_ip_metadata_get_receive_time (GetCheckedHandle ());
+				return (time == 0) ? TimeSpan.Zero : TimeSpan.FromTicks ((long) time / 100);
```
stephen-hawley: I don't think this is needed. There's no indication that `FromTicks (0)` returns an illegal value. I looked at the MS reference source for `FromTicks` and it returns `new TimeSpan (ticks)`. Since `TimeSpan` is a struct, this is cheap.

src/Network/NWTcpMetadata.cs
```diff
@@ -0,0 +1,25 @@
...
+using CoreFoundation;
+
+namespace Network {
+
+	[TV (12,0), Mac (10,14), iOS (12,0), Watch (6,0)]
+	public class NWTcpMetadata : NWProtocolMetadata {
+
+		internal NWTcpMetadata (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		public uint AvailableReceiveBuffer () => nw_tcp_get_available_receive_buffer (GetCheckedHandle ());
```
spouliot: why not a property ?

The `=>` syntax is easy to mis-read between methods and properties. However I don't see a good reason for not using a property.

src/Network/NWTcpMetadata.cs
```diff
@@ -0,0 +1,25 @@
...
+
+		public uint AvailableSendBuffer () => nw_tcp_get_available_send_buffer (GetCheckedHandle ());
```
spouliot: same

src/Network/NWIPMetadata.cs
mandel-macaque: More of not using a new struct every time, and use the static one. Can be removed.

stephen-hawley: The internal representation is a long. Returning a long-sized struct is essentially free.

tests/monotouch-test/Network/NWIPProtocolMetadataTest.cs
```diff
@@ -0,0 +1,61 @@
...
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class NWIPProtocolMetadataTest {
+		NWIPMetadata metadata;
+
+		[TestFixtureSetUp]
+		public void Init () => TestRuntime.AssertXcodeVersion (10, 0);
+
+		[SetUp]
+		public void SetUp () {
```
prmallic: minor style - `{` in next line?

src/Network/NWIPMetadata.cs
mandel-macaque: Removed it, so not a big issue.

mandel-macaque:
Failures are due to known issues: https://github.com/xamarin/maccore/issues/940

PR #7192, opened 2019-10-08
Bump to mono:2019-10

directhex:

PR #7282, opened 2019-10-22
[tests] Add make target to manually launch a device test run for the current commit.

rolfbjarne:


mandel-macaque:
not needed for this commit, but we might want to remove wrench... (I know, the feelings..)

rolfbjarne:
> not needed for this commit, but we might want to remove wrench... (I know, the feelings..)

That would need maccore fixes too... maybe next time that code has to be changed.

rolfbjarne:
> Code looks good but I think I'm missing the context / point. When/why will we need that?

@VincentDondain: this is when it's useful: https://github.com/xamarin/xamarin-macios/pull/7192#issuecomment-549537581

PR #7292, opened 2019-10-23
[TVMLKit] Add TVMLKit binding for xcode11.2b2

wachaudh:
Add TVMLKit binding for xcode11.2b2

### DIFF
--- a/src/tvmlkit.cs
+++ b/src/tvmlkit.cs
@@ -120,6 +120,7 @@ namespace TVMLKit {
 	[ErrorDomain ("TVMLKitErrorDomain")]
 	public enum TVMLKitError : long {
 		Unknown = 1,
+        InvalidArguments,
 		InternetUnavailable,
 		FailedToLaunch,
 		Last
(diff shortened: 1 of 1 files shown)


src/tvmlkit.cs
```diff
@@ -120,6 +120,7 @@ public enum TVImageType : long {
 	[ErrorDomain ("TVMLKitErrorDomain")]
 	public enum TVMLKitError : long {
 		Unknown = 1,
+		InvalidArguments,
```
mandel-macaque: The order of the enums matter. This changes makes InternetUnavailable to be the old one +1. Which is bad it the users do depend on the casting of enums to long/

wachaudh:
maccore issue: https://github.com/xamarin/maccore/issues/2039

src/tvmlkit.cs
Therzok: It's basically an ABI break.

Therzok: Not just enum to long. It breaks interop too.

PR #7293, opened 2019-10-23
[Tests] Fix the Network tests.

mandel-macaque:
The pattern used to Cancel then Dispose the connection is wrong and
ended up crashing the tests sometimes. It is a race condition and is
very rare, but makes the tests flaky. The changes ensures that we
Dispose the connection, which internally closes it.

Also cleaned some on the CWL that were not needed.

PR #7295, opened 2019-10-23
[Appkit] Bindings part 4

prmallic:
AppKit bindings for xcode11. I've added some bindings (enums) that were introduced before so that the newer bindings are compatible.


src/appkit.cs
```diff
@@ -6523,6 +6538,12 @@ partial interface NSFont : NSSecureCoding, NSCopying {
 		[Internal]
 		[Export ("getAdvancements:forCGGlyphs:count:")]
 		void _GetAdvancements (IntPtr advancements, IntPtr glyphs, nuint glyphCount);
+
+		[Mac (10,15)]
+		[Static]
+		[Export ("monospacedSystemFontOfSize:weight:")]
+		[Internal]
+		IntPtr _MonospacedSystemFont (nfloat fontSize, double weight);
```
VincentDondain:
```suggestion
		IntPtr _MonospacedSystemFont (nfloat fontSize, nfloat weight);
```

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+
+	interface INSTextCheckingClient { }
+
+	[Protocol]
+	interface NSTextCheckingClient : NSTextInputTraits, NSTextInputClient
+	{
+		[Abstract]
+		[Export ("annotatedSubstringForProposedRange:actualRange:")]
+		[return: NullAllowed]
+		NSAttributedString AnnotatedSubstring (NSRange range, [NullAllowed] NSRange actualRange);
```
prmallic: In ObjC this is a required method of a protocol that returns a pointer to `NSAttributedString`. Does it make sense to make this `[Internal]` and do a manual implementation mapping a `IntPtr` returned to `NSAttributedString`?

src/appkit.cs
```diff
@@ -6745,6 +6766,11 @@ interface NSFontDescriptor : NSSecureCoding, NSCopying {
 		[Mac (10, 13)]
 		[Export ("requiresFontAssetRequest")]
 		bool RequiresFontAssetRequest { get; }
+
+		[Mac (10,15)]
+		[Export ("fontDescriptorWithDesign:")]
+		[return: NullAllowed]
+		NSFontDescriptor FontDescriptorWithDesign (NSFontDescriptorSystemDesign design);
```
mandel-macaque: No need for WithDesign, is the parameter.

src/appkit.cs
```diff
@@ -7750,6 +7776,27 @@ interface NSEvent : NSCoding, NSCopying {
...
+	{
+		CapsLock = 1uL << 16,
+		Shift = 1uL << 17,
+		Control = 1uL << 18,
+		Option = 1uL << 19,
+		Command = 1uL << 20,
+		NumericPad = 1uL << 21,
+		Help = 1uL << 22,
+		Function = 1uL << 23,
+		DeviceIndependentFlagsMask = 0xffff0000L
```
mandel-macaque: minor, missing ','
```suggestion
		DeviceIndependentFlagsMask = 0xffff0000L,
```

src/appkit.cs
```diff
@@ -15099,6 +15172,15 @@ interface NSStoryboard {
...
 		NSStoryboard MainStoryboard { get; }
+
+		[Mac (10,15)]
+		[Export ("instantiateInitialControllerWithCreator:")]
+		[return: NullAllowed]
+		NSObject InstantiateInitialController ([NullAllowed] NSStoryboardControllerCreator block);
+
+		[Mac (10,15)]
+		[Export ("instantiateControllerWithIdentifier:creator:")]
+		NSObject InstantiateController (string identifier, [NullAllowed] NSStoryboardControllerCreator block);
```
mandel-macaque: we prefer handler over block @dalexsoto dixit.
```suggestion
		NSObject InstantiateController (string identifier, [NullAllowed] NSStoryboardControllerCreator handler);
```

src/appkit.cs
VincentDondain: Ugh, other APIs are horrible unfortunately, so in the name of constancy we could keep it as is but I think this is better:

```suggestion
		NSFontDescriptor GetFontDescriptor (NSFontDescriptorSystemDesign design);
```

Need other opinions on this.

src/appkit.cs
VincentDondain: `,` at the end of the enum to avoid diff nose when adding values

src/appkit.cs
```diff
@@ -21600,6 +21728,18 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
 		[Export ("requestAuthorizationOfType:completionHandler:")]
 		void RequestAuthorization (NSWorkspaceAuthorizationType type, Action<NSWorkspaceAuthorization, NSError> completionHandler);
+
+		[Mac (10,15)]
+		[Export ("openApplicationAtURL:configuration:completionHandler:")]
+		void OpenApplication (NSUrl applicationURL, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
+
+		[Mac (10,15)]
+		[Export ("openURL:configuration:completionHandler:")]
+		void OpenURL (NSUrl url, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```
mandel-macaque: In C# we prefer Url.
```suggestion
		void OpenUrl (NSUrl url, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```

src/appkit.cs
```diff
@@ -21600,6 +21728,18 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Mac (10,15)]
+		[Export ("openURLs:withApplicationAtURL:configuration:completionHandler:")]
+		void OpenURLs (NSUrl[] urls, NSUrl applicationURL, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```
mandel-macaque: Same, 'Url'
```suggestion
		void OpenUrls (NSUrl[] urls, NSUrl applicationURL, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		[return: NullAllowed]
+		NSMenu MenuAtIndex (nuint location, bool clickedOnSelection, NSRange effectiveRange);
+
+		[Export ("spellCheckerDocumentTag")]
+		nint SpellCheckerDocumentTag { get; set; }
+	}
+
+	delegate NSCollectionViewItem NSCollectionViewDiffableDataSourceItemProvider (NSCollectionView collectionView, NSIndexPath indexPath, NSObject itemIdentifierType);
+
+	delegate NSView NSCollectionViewDiffableDataSourceSupplementaryViewProvider (NSCollectionView collectionView, string str, NSIndexPath indexPath);
```
mandel-macaque: Don't use acronyms for the delegates, is there a nicer parameter name so that it makes more sense for the users, 'str' is not very precise.

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		where ItemIdentifierType : NSObject {
+
+		[Export ("initWithCollectionView:itemProvider:")]
+		IntPtr Constructor (NSCollectionView collectionView, NSCollectionViewDiffableDataSourceItemProvider itemProvider);
+
+		[Export ("snapshot")]
+		NSDiffableDataSourceSnapshot<SectionIdentifierType, ItemIdentifierType> Snapshot { get; }
+
+		[Export ("applySnapshot:animatingDifferences:")]
+		void ApplySnapshot (NSDiffableDataSourceSnapshot<SectionIdentifierType, ItemIdentifierType> snapshot, bool animatingDifferences);
```
mandel-macaque: Apply sounds nicer, since we know is a Snapshot as the first parameter.

src/appkit.cs
VincentDondain:
```suggestion
		void OpenUrl (NSUrl url, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```

3 letter, PascalCase

src/appkit.cs
VincentDondain:
```suggestion
		void OpenUrls (NSUrl[] urls, NSUrl applicationUrl, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```

src/xkit.cs
```diff
@@ -1190,4 +1190,87 @@ interface NSLayoutManagerDelegate {
...
+		SectionIdentifierType GetSectionIdentifierForSection (ItemIdentifierType itemIdentifier);
+
+		[Export ("indexOfItemIdentifier:")]
+		nint GetIndex (ItemIdentifierType itemIdentifier);
+
+		[Export ("indexOfSectionIdentifier:")]
+		nint GetIndex (SectionIdentifierType sectionIdentifier);
+
+		[Export ("appendItemsWithIdentifiers:")]
+		void AppendItems (ItemIdentifierType [] identifiers);
```
mandel-macaque: Append?

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
 		[Export ("customizationLabel")]
 		string CustomizationLabel { get; set; }
 	}
+
+	[Protocol]
+	interface NSTextInputTraits
```
VincentDondain: Missing availability attribute

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		[Export ("linkDetectionType", ArgumentSemantic.Assign)]
+		NSTextInputTraitType LinkDetectionType { get; set; }
+
+		[Export ("textCompletionType", ArgumentSemantic.Assign)]
+		NSTextInputTraitType TextCompletionType { get; set; }
+	}
+
+	interface INSTextCheckingClient { }
+
+	[Protocol]
```
VincentDondain: Missing availability attribute

src/xkit.cs
```diff
@@ -1190,4 +1190,87 @@ interface NSLayoutManagerDelegate {
...
+
+		[Export ("appendItemsWithIdentifiers:intoSectionWithIdentifier:")]
+		void AppendItems (ItemIdentifierType [] identifiers, SectionIdentifierType sectionIdentifier);
```
mandel-macaque: Same, Append?

src/xkit.cs
```diff
@@ -1190,4 +1190,87 @@ interface NSLayoutManagerDelegate {
...
+
+		[Export ("insertItemsWithIdentifiers:beforeItemWithIdentifier:")]
+		void InsertItemsBefore (ItemIdentifierType [] identifiers, ItemIdentifierType itemIdentifier);
+
+		[Export ("insertItemsWithIdentifiers:afterItemWithIdentifier:")]
+		void InsertItemsAfter (ItemIdentifierType [] identifiers, ItemIdentifierType itemIdentifier);
+
+		[Export ("deleteItemsWithIdentifiers:")]
+		void DeleteItems (ItemIdentifierType [] identifiers);
```
mandel-macaque: Delete?

src/xkit.cs
```diff
@@ -1190,4 +1190,87 @@ interface NSLayoutManagerDelegate {
...
+
+		[Export ("deleteAllItems")]
+		void DeleteAllItems ();
```
mandel-macaque: Maybe DeleteAll?

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+
+	public interface INSSharingServicePickerToolbarItemDelegate { }
+
+	[Protocol, Model (AutoGeneratedName = true)]
+	[BaseType (typeof (NSSharingServicePickerDelegate))]
+	interface NSSharingServicePickerToolbarItemDelegate
+	{
+		[Abstract]
+		[Export ("itemsForSharingServicePickerToolbarItem:")]
+		NSObject[] ItemsForSharingServicePickerToolbarItem (NSSharingServicePickerToolbarItem pickerToolbarItem);
```
VincentDondain:
```suggestion
		NSObject[] GetItems (NSSharingServicePickerToolbarItem pickerToolbarItem);
```

src/xkit.cs
VincentDondain: Mmmmh not too sure about that. I mean you'd loose a bit of the context.

spouliot: too late anyway, that has already shipped (this is code being moved from files)

src/xkit.cs
spouliot: too late anyway, that has already shipped (this is code being moved from files)

src/xkit.cs
spouliot: too late anyway, that has already shipped (this is code being moved from files)

src/xkit.cs
spouliot: too late anyway, that has already shipped (this is code being moved from files)

src/appkit.cs
```diff
@@ -4098,6 +4098,16 @@ partial interface NSColor : NSCoding, NSCopying, NSSecureCoding, NSPasteboardRea
 		[Export ("systemGrayColor", ArgumentSemantic.Strong)]
 		NSColor SystemGrayColor { get; }
 
+		[Mac (10, 15)]
+		[Static]
+		[Export ("systemIndigoColor", ArgumentSemantic.Strong)]
+		NSColor SystemIndigoColor { get; }
+
+		[Mac (10, 12)]
```
spouliot: hmm... I'm always wary of new Xcode adding old stuff
make sure you follow up, after the merge, if the old-macos bots are happy

src/appkit.cs
spouliot: -> `typedef CGFloat NSFontWeight;` and `CGFloat` is `nfloat`

src/appkit.cs
spouliot: It's a method so it needs a verb. Since it's inside `NSFontDescriptor` and returns a `NSFontDescriptor` then it's best to use `Create` for the verb. Not sure we need anything more (the parameter self-explains the rest)

Sadly most older API did not follow that rule :(

src/appkit.cs
```diff
@@ -7750,6 +7776,27 @@ interface NSEvent : NSCoding, NSCopying {
 		[Mac (10,12,2)]
 		[Export ("coalescedTouchesForTouch:")]
 		NSTouch[] GetCoalescedTouches (NSTouch touch);
+
+		[Mac (10,15)]
+		[Export ("charactersByApplyingModifiers:")]
+		[return: NullAllowed]
+		string GetCharacters (NSEventModifierFlags modifiers);
```
spouliot: a bit weird to have `GetCharacters` returns a `string`... but it matches ObjC and I don't have a better suggestion

src/appkit.cs
```diff
@@ -15099,6 +15172,15 @@ interface NSStoryboard {
 		[Static]
 		[NullAllowed, Export ("mainStoryboard", ArgumentSemantic.Strong)]
 		NSStoryboard MainStoryboard { get; }
+
+		[Mac (10,15)]
+		[Export ("instantiateInitialControllerWithCreator:")]
+		[return: NullAllowed]
+		NSObject InstantiateInitialController ([NullAllowed] NSStoryboardControllerCreator block);
```
spouliot: docs says it returns `id`
but I suspect it returns a (subclass of) `NSViewController`
double check how iOS `UIStoryboard` does its binding

src/appkit.cs
spouliot: same, returning an `NSObject` is best avoided, when possible

src/appkit.cs
```diff
@@ -17906,7 +17996,7 @@ interface NSTextAttachmentCell {
 	}
 
 	[BaseType (typeof (NSObject))]
-	interface NSTextAttachment : NSCoding {
+	interface NSTextAttachment : NSCoding, NSSecureCoding {
```
spouliot: minor: `NSSecureCoding` is a subclass of `NSCoding` so you don't really need to keep the later

src/appkit.cs
spouliot: Not sure if I get your concern correctly but it's a protocol, so an interface once generated
Making a member internal would not be helpful, even less if required since no one would be able to conform

spouliot: Method name lacks a verb, e.g. `Get...`

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		void ReplaceCharacters (NSRange range, NSAttributedString annotatedString);
+
+		[Abstract]
+		[Export ("selectAndShowRange:")]
+		void SelectAndShow (NSRange range);
+
+		[Abstract]
+		[Export ("viewForRange:firstRect:actualRange:")]
+		[return: NullAllowed]
+		NSView GetView (NSRange range, [NullAllowed] CGRect firstRect, [NullAllowed] NSRange actualRange);
```
spouliot: `[NullAllowed] CGRect firstRect` that sounded weird, since `CGRect` is a struct

https://developer.apple.com/documentation/appkit/nstextcheckingclient/3242726-viewforrange?language=objc
```
- (NSView *)viewForRange:(NSRange)range 
               firstRect:(NSRectPointer)firstRect 
             actualRange:(NSRangePointer)actualRange;
```

^ those are not the same types, you're hiding the pointer, which is not correct and that won't work at runtime.

If you mapped this in other API please update them ASAP.

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		NSCandidateListTouchBarItem CandidateListTouchBarItem { get; }
+	}
+
+	[Mac (10,15)]
+	[BaseType (typeof (NSObject))]
+	interface NSWorkspaceOpenConfiguration : NSCopying
+	{
+		[Static]
+		[Export ("configuration")]
+		NSWorkspaceOpenConfiguration Configuration ();
```
spouliot: Looks like that's either a `Create` method *or* it's a property

If it's a method it's also likely that `[DisableDefaultCtor]` should be used.

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		string[] Arguments { get; set; }
+
+		[Export ("environment", ArgumentSemantic.Copy)]
+		NSDictionary<NSString, NSString> Environment { get; set; }
+
+		[NullAllowed, Export ("appleEvent", ArgumentSemantic.Strong)]
+		NSAppleEventDescriptor AppleEvent { get; set; }
+
+		[Export ("architecture")]
+		int Architecture { get; set; }
```
spouliot: `cpu_type_t` -> we might have an enum that map to this

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+
+		[Export ("updateCandidates")]
+		void UpdateCandidates ();
+
+		[Export ("validAnnotations")]
+		string[] ValidAnnotations { get; }
+
+		[Export ("menuAtIndex:clickedOnSelection:effectiveRange:")]
+		[return: NullAllowed]
+		NSMenu MenuAtIndex (nuint location, bool clickedOnSelection, NSRange effectiveRange);
```
spouliot: -> `GetMenu`, i.e. verb and suffix looks like a dupe of the first argument

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		[Field ("NSFontDescriptorSystemDesignSerif")]
+		Serif,
+
+		[Mac (10, 15)]
+		[Field ("NSFontDescriptorSystemDesignMonospaced")]
+		Monospaced,
+
+		[Mac (10, 15)]
+		[Field ("NSFontDescriptorSystemDesignRounded")]
+		Rounded,
```
spouliot: move `[Mac(10,15)]` on the type (enum) instead of having each member decorated

src/appkit.cs
```diff
@@ -21600,6 +21728,18 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
 		[Mac (10,14)]
 		[Export ("requestAuthorizationOfType:completionHandler:")]
 		void RequestAuthorization (NSWorkspaceAuthorizationType type, Action<NSWorkspaceAuthorization, NSError> completionHandler);
+
+		[Mac (10,15)]
+		[Export ("openApplicationAtURL:configuration:completionHandler:")]
+		void OpenApplication (NSUrl applicationURL, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```
rolfbjarne: Minor:

```suggestion
		void OpenApplication (NSUrl applicationUrl, NSWorkspaceOpenConfiguration configuration, [NullAllowed] Action<NSRunningApplication, NSError> completionHandler);
```

rolfbjarne: Does `[Async]` make sense for this method?

src/appkit.cs
rolfbjarne: Same: does `[Async]` make sense for this method?

src/appkit.cs
rolfbjarne: Same: does `[Async]` make sense for this method?

src/appkit.cs
prmallic: I think it does. The header reads - `Launch the application at the specified URL with the given configuration. The completion handler is called on a concurrent queue with either the NSRunningApplication on success, or an NSError on failure.` Thanks!

src/appkit.cs
VincentDondain: Or play it safe, make it 10.15 cause that's when it appeared for the first time and if it's really needed on lower platforms we can re-adjust later.

src/appkit.cs
VincentDondain: I my mind `Create` was reserved for static APIs but I guess that works yes.

src/appkit.cs
VincentDondain: @spouliot here's the code Pramit is trying to bind:

```
unsafe NSAttributedString AttributedSubstringForProposedRange (NSRange range, [NullAllowed] NSRangePointer* actualRange);
```

Similar to https://github.com/xamarin/xamarin-macios/pull/7295/files#r338313032 he's hiding a pointer here.

A similar API:
```
unsafe CGRect FirstRectForCharacterRange (NSRange range, [NullAllowed] NSRangePointer* actualRange);
```
was bound like this: https://github.com/xamarin/xamarin-macios/blob/master/src/appkit.cs#L19304

So you can use ref/out in those cases because you're dealing with a pointer.

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		[NullAllowed]
+		INSSharingServicePickerToolbarItemDelegate Delegate { get; set; }
+
+		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
+		NSObject WeakDelegate { get; set; }
+	}
+
+	public interface INSSharingServicePickerToolbarItemDelegate { }
+
+	[Protocol, Model (AutoGeneratedName = true)]
```
VincentDondain: Missing availability attribute

src/appkit.cs
VincentDondain: Swift version is `applySnapshot` -> https://developer.apple.com/documentation/uikit/uitableviewdiffabledatasourcereference/3356326-applysnapshot

Arguably

src/appkit.cs
VincentDondain: Also hiding a pointer: `unsafe NSMenu MenuAtIndex (nuint location, bool clickedOnSelection, NSRangePointer* effectiveRange);`

src/appkit.cs
```diff
@@ -27427,4 +27575,265 @@ interface NSStepperTouchBarItem
...
+		void InsertedText (NSRange range);
+
+		[Export ("didChangeSelectedRange")]
+		void DidChangeSelectedRange ();
+
+		[Export ("considerTextCheckingForRange:")]
+		void ConsiderTextChecking (NSRange range);
+
+		[Export ("checkTextInRange:types:options:")]
+		void CheckText (NSRange range, NSTextCheckingTypes checkingTypes, NSDictionary<NSString, NSObject> options);
```
VincentDondain: Looking at the [doc](https://developer.apple.com/documentation/appkit/nstextcheckingoptionkey?language=objc), I believe you can use the strong dictionary that already exists for that: `NSTextCheckingOptions`.

_Note: that strong dictionary might be missing some keys 🕵. I count 10 in the doc and 8 in our strong dictionary (:_

src/appkit.cs
prmallic: So there is the `CFBundle.Architecture` enum that maps the `cpu_type_t`s. I'll move it to CoreFoundation_Core_Sources to make it available here. However there is a generator issue https://github.com/xamarin/maccore/issues/2041 that is causing an error right now. @mandel-macaque and I found a fix for it, so we'll get that working

src/appkit.cs
```diff
@@ -27427,4 +27579,278 @@ interface NSStepperTouchBarItem
...
+		[Internal]
+		[Export ("architecture")]
+		int _LaunchArchitecture { get; set; }
+
+		CFBundle.Architecture LaunchArchitecture {
+			[Wrap ("(CFBundle.Architecture) this._LaunchArchitecture")]
+			get;
+			[Wrap ("this._LaunchArchitecture = (int) value")]
+			set;
+		}
```
prmallic: Working on this. Pending issue - https://github.com/xamarin/maccore/issues/2041

src/appkit.cs
prmallic: Thanks for pointing me to `[EditorBrowsable (EditorBrowsableState.Advanced)]` and `StrongDictionary`!

src/appkit.cs
```diff
@@ -27427,4 +27579,279 @@ interface NSStepperTouchBarItem
...
+		// [Internal]
+		// [Export ("architecture")]
+		// int _LaunchArchitecture { get; set; }
+
+		// CFBundle.Architecture LaunchArchitecture {
+		// 	[Wrap ("(CFBundle.Architecture) this._LaunchArchitecture")]
+		// 	get;
+		// 	[Wrap ("this._LaunchArchitecture = (int) value")]
+		// 	set;
+		// }
```
mandel-macaque: Once https://github.com/xamarin/xamarin-macios/pull/7309 lands you can uncomment this. Will ping you when done.

src/appkit.cs
```diff
@@ -13153,6 +13201,18 @@ partial interface NSScreen {
...
+		[Export ("maximumPotentialExtendedDynamicRangeColorComponentValue")]
+		nfloat MaximumPotentialExtendedDynamicRangeColorComponentValue { get; }
+
+		[Mac (10, 15)]
+		[Export ("maximumReferenceExtendedDynamicRangeColorComponentValue")]
+		nfloat MaximumReferenceExtendedDynamicRangeColorComponentValue { get; }
+
+		[Mac (10, 15)]
+		[Export ("localizedName", ArgumentSemantic.Copy)]
+		string LocalizedName { get; }
```
wachaudh: `localizedName` is an optional in swift according to the [doc](https://developer.apple.com/documentation/appkit/nsrunningapplication/1526751-localizedname). Shouldn't yhis have a `NullAllowed` attribute?

mandel-macaque:
Crashes should have been fixed after the merge of https://github.com/xamarin/xamarin-macios/pull/7293

PR #7298, opened 2019-10-24
[runtime] Ensure binding-generator full initialize NVector3[d] members

spouliot:


spouliot:
@VincentDondain the member does not make sense (it does not exists in native code) so it should not be used (in managed code) or your code has other issues. Still better 0 than something random.

PR #7306, opened 2019-10-25
[runtime] Simplify `xamarin_switch_gchandle` by removed dead code

spouliot:
If `old_gchandle` is not set (0) then the function returns early,
so can't be 0 after the `MONO_THREAD_ATTACH` meaning:

* the `else` branch is never executed;

* the `exception_gchandle` is never set to something other than 0

* ` if (exception_gchandle == 0) {` is always `true` and can be removed (diff looks bigger than it is because of indentation change)

* the call to `xamarin_process_managed_exception_gchandle` is unneeded


mandel-macaque:
Ugh the indentation before the changes..

spouliot:
unrelated failure, HE0038 -> https://github.com/xamarin/maccore/issues/581

PR #7307, opened 2019-10-25
[runtime] Simplify Vision dlsym'ed functions

spouliot:
More code sharing - DRYer :)

It also remove complaints (from static analysis tools) that a `dlclose`
should be present. That's not really an issue here since it's reference
counted and won't be unloaded if you use the code (and such a call is
not added in the PR). It silence the warning since it's not a local
variable anymore.

### DIFF
--- a/runtime/bindings.m
+++ b/runtime/bindings.m
@@ -103,10 +103,12 @@ static void* vision_handle;
 static vision_func
 get_vision_func (const char *func_name, const char **error_msg)
 {
-	vision_handle = dlopen ("/System/Library/Frameworks/Vision.framework/Vision", RTLD_LAZY);
 	if (vision_handle == NULL) {
-		*error_msg = "Could not open Vision.framework";
-		return NULL;
+		vision_handle = dlopen ("/System/Library/Frameworks/Vision.framework/Vision", RTLD_LAZY);
+		if (vision_handle == NULL) {
+			*error_msg = "Could not open Vision.framework";
+			return NULL;
+		}
 	}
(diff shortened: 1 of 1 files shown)


runtime/bindings.m
```diff
@@ -98,25 +98,31 @@
 
 typedef CGPoint (*vision_func) (vector_float2 faceLandmarkPoint, CGRect faceBoundingBox, size_t imageWidth, size_t imageHeight);
 
+static void* vision_handle;
+
+static vision_func
+get_vision_func (const char *func_name, const char **error_msg)
+{
+	vision_handle = dlopen ("/System/Library/Frameworks/Vision.framework/Vision", RTLD_LAZY);
```
rolfbjarne: I think you can check if `vision_handle` is NULL before calling dlopen:

```c
if (vision_handle == NULL) {
    vision_handle = dlopen (...);
    if (vision_handle == NULL) {
        *error_msg = "...";
        return NULL;
    }
}
return (vision_func) dlsym (vision_handle, func_name);
```

runtime/bindings.m
```diff
@@ -98,25 +98,33 @@
 
 typedef CGPoint (*vision_func) (vector_float2 faceLandmarkPoint, CGRect faceBoundingBox, size_t imageWidth, size_t imageHeight);
 
+static void* vision_handle;
```
Therzok: Make this a static field? No use exporting it outside the function if nothing uses it.

spouliot:
unrelated failure due to HE0038 on watchOS => https://github.com/xamarin/maccore/issues/581

PR #7308, opened 2019-10-25
[runtime] Throw a OutOfMemoryException when processing parameters

spouliot:
`calloc` can return `null` and we're writing to the memory location
which would crash the process.

An `OutOfMemoryException` is the correct way to handle this (even if
will likely crash the process anyway).


spouliot:
unrelated HE0038 (watchOS) error https://github.com/xamarin/maccore/issues/581

PR #7318, opened 2019-10-28
[Tests] Ignore BCL tests on 32b that fail due to threads failing to be created.

mandel-macaque:
Related mono issue: https://github.com/mono/mono/issues/17588


mandel-macaque:
@monojenkins backport d16-5

PR #7322, opened 2019-10-28
[Xcode 11.2] Callkit updates

whitneyschmidt:


src/callkit.cs
```diff
@@ -30,7 +30,13 @@ public enum CXCallDirectoryEnabledStatus : long {
...
 	[Native]
 	public enum CXErrorCode : long {
-		Unknown = 0
+		Unknown = 0,
+		[iOS (13, 2)]
+		[Introduced (PlatformName.UIKitForMac, 13,2)]
+		Unentitled = 1,
+		[iOS (13, 2)]
+		[Introduced (PlatformName.UIKitForMac, 13,2)]
+		InvalidArgument = 2,
```
spouliot: Every rule has its exception - and error enums don't get availability attributes.

Why ? they are output (not input) to API so it's not a problem to have them in your code without a version check.

whitneyschmidt: Thank you @spouliot :)

PR #7324, opened 2019-10-29
[ARKit] Xcode11.2 Binding

prmallic:
Xcode11.2 Update - Diff - https://github.com/xamarin/xamarin-macios/wiki/ARKit-iOS-xcode11.2-beta1

### DIFF
--- a/src/arkit.cs
+++ b/src/arkit.cs
@@ -67,7 +67,9 @@ namespace ARKit {
 		InvalidReferenceObject = 301,
 		InvalidWorldMap = 302,
 		InvalidConfiguration = 303,
+		[Deprecated (PlatformName.iOS, 13, 2, message : "Please use the 'InvalidCollaborationData' value instead.")]
 		CollaborationDataUnavailable = 304,
+		InvalidCollaborationData = 304,
 		InsufficientFeatures = 400,
 		ObjectMergeFailed = 401,
 		FileIOFailed = 500,


src/arkit.cs
```diff
@@ -67,7 +67,9 @@ public enum ARErrorCode : long {
 		InvalidReferenceObject = 301,
 		InvalidWorldMap = 302,
 		InvalidConfiguration = 303,
+		[Deprecated (PlatformName.iOS, 13, 2, message : "Please use the 'InvalidCollaborationData' value instead.")]
 		CollaborationDataUnavailable = 304,
```
prmallic: I realize that error codes don't get their own availability attributes (https://github.com/xamarin/xamarin-macios/pull/7322/commits/36c637056af3b932341f20aad8c9b19c63aecc46#r339817547). How would we handle deprecation of the same though? Or give the suggested enum value?

mandel-macaque: Deprecation AFAIK is correct, but I would have done the following:
```csharp
CollaborationDataUnavailable = InvalidCollaborationData,
InvalidCollaborationData = 304,
```

So that we are always sure they are the same value.

src/arkit.cs
```diff
@@ -67,7 +67,9 @@ public enum ARErrorCode : long {
...
+		InvalidCollaborationData = 304,
```
wachaudh: missing availability attribute.

prmallic: But its an error enum. https://github.com/xamarin/xamarin-macios/pull/7322#discussion_r339817547 Does it make sense to add it then?

mandel-macaque:
Merging, intro tests should be fixed by https://github.com/xamarin/xamarin-macios/pull/7328 xls issue is under investigation: https://github.com/xamarin/maccore/issues/2048 and last is a HE0038 error.

mandel-macaque:
Will merge after @spouliot +1.

PR #7330, opened 2019-10-29
[MPSCore] Xcode11.1 GM Binding

prmallic:
Xcode11.1 GM Diff - https://github.com/xamarin/xamarin-macios/wiki/MPSCore-iOS-xcode11.1-GM
Added missing 'offset' selector

### DIFF
--- a/src/metalperformanceshaders.cs
+++ b/src/metalperformanceshaders.cs
@@ -1900,7 +1900,7 @@ namespace MetalPerformanceShaders {
 		[Export ("matrixBytes")]
 		nuint MatrixBytes { get; }
 
-		[TV (13, 0), Mac (10, 15), iOS (13, 0)]
+		[TV (13, 0), Mac (10, 15), iOS (13, 1)]
 		[Export ("offset")]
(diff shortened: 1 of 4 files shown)


src/metalperformanceshaders.cs
```diff
@@ -2089,6 +2093,10 @@ interface MPSVector {
 		[TV (11,3), Mac (10,13,4), iOS (11,3)]
```
whitneyschmidt: nit: spacing for version #

VincentDondain:
@prmallic I can help you with that (:

VincentDondain:
Mmh indeed, looks like Xcode 11.1(.2) is not listing older 13.X simulators... https://forums.developer.apple.com/thread/123347

VincentDondain:
_@prmallic note that if it comes from the 13.1 API diffs, it's fair to use that version for the availability attribute instead of being too smart and go for the older version (which might be incorrect)._

prmallic:
Device test on iOS13.0 failed. So bumping the version to 13.1

PR #7331, opened 2019-10-29
[MapKit] Xcode 11.2 update

whitneyschmidt:


src/mapkit.cs
```diff
@@ -327,6 +331,17 @@ interface MKMapItem : NSSecureCoding
 		[Export ("openMapsWithItems:launchOptions:"), Internal]
 		bool _OpenMaps ([NullAllowed] MKMapItem [] mapItems, [NullAllowed] NSDictionary launchOptions);
 
+		[iOS (13, 2), NoMac, NoTV, NoWatch]
+		[Introduced (PlatformName.UIKitForMac, 13, 2)]
```
whitneyschmidt: Wasn't sure about Catalyst - according to Apple docs, this will be supported in Catalyst: https://developer.apple.com/documentation/mapkit/mkmapitem/3395859-openinmapswithlaunchoptions?language=objc

However, it was not explicitly listed as supported in the diffs.

src/mapkit.cs
```diff
@@ -327,6 +331,17 @@ interface MKMapItem : NSSecureCoding
...
+		[Export ("openInMapsWithLaunchOptions:fromScene:completionHandler:"), Internal]
```
rolfbjarne: This API is internal, and you're not exposing it through any manually written API, so it's not usable for customers. Did you forget to add anything to this file: https://github.com/xamarin/xamarin-macios/blob/e8409f28347d5c688dfe8d45406be3512cefdf9e/src/MapKit/MKMapItem.cs#L112

Or maybe a method here with `[Wrap (...)]?`

mandel-macaque: Or remove Internal and the _ I think this was left here by accident.

src/mapkit.cs
```diff
@@ -327,6 +331,17 @@ interface MKMapItem : NSSecureCoding
...
+		void _OpenInMaps ([NullAllowed] NSDictionary  launchOptions, [NullAllowed] UIScene fromScene, Action<NSError> completionHandler);
+
+		[iOS (13, 2), NoMac, NoTV, NoWatch]
+		[Introduced (PlatformName.UIKitForMac, 13, 2)]
+		[Static]
+		[Export ("openMapsWithItems:launchOptions:fromScene:completionHandler:"), Internal]
+		void _OpenMaps ([NullAllowed] MKMapItem [] mapItems, [NullAllowed] NSDictionary launchOptions, [NullAllowed] UIScene fromScene, Action<NSError> completionHandler);
```
mandel-macaque: Same as with the other method and either what @rolfbjarne said or me. Internal and the _ naming.

src/mapkit.cs
```diff
@@ -327,6 +331,17 @@ interface MKMapItem : NSSecureCoding
 		[Export ("openMapsWithItems:launchOptions:"), Internal]
 		bool _OpenMaps ([NullAllowed] MKMapItem [] mapItems, [NullAllowed] NSDictionary launchOptions);
 
+		[iOS (13, 2), NoMac, NoTV, NoWatch]
+		[Introduced (PlatformName.UIKitForMac, 13, 2)]
+		[Export ("openInMapsWithLaunchOptions:fromScene:completionHandler:")]
+		void OpenInMaps ([NullAllowed] NSDictionary  launchOptions, [NullAllowed] UIScene fromScene, Action<NSError> completionHandler);
```
prmallic: Can we add `[Async]` to these?

src/mapkit.cs
```diff
@@ -46,6 +46,10 @@
 
 namespace MapKit {
 
+#if !IOS
+	interface UIScene {}
```
VincentDondain: Minor: doing it this way is fair but it looks this file is using the `using` pattern (above) to achieve the same result. Might have been best to add to that (;

src/mapkit.cs
VincentDondain: Minor: style, you have 2 spaces before `launchOptions`

src/mapkit.cs
```diff
@@ -32,6 +32,7 @@
 using UIView=AppKit.NSView;
 using UIEdgeInsets=AppKit.NSEdgeInsets;
 using UIColor=AppKit.NSColor;
+using UIScene=AppKit.NSColor;
```
whitneyschmidt: AppKit doesn't have an equivalent to `UIScene`, maybe it would be better to use `NSObject` here?

PR #7334, opened 2019-10-30
[simlauncher] Move simlauncher source to .mm file extension (ObjC++)

spouliot:
This is already compiled as ObjC++ but this gives an additional (and
required) hint to some tools that it is Objective C++ code.

This avoid a lot of build errors (under static analysis tools),
most of them not really helpful (a trait shared with many C++ compiler
warnings and errors) except for this one:

```
error: invalid argument ‘-std=c++14’ not allowed with ‘Objective-C’
```


spouliot:
unrelated watchOS HE0038 error -> https://github.com/xamarin/maccore/issues/581

PR #7337, opened 2019-10-30
[Tests] Fix some network tests that were leaving the connection in a bad state

mandel-macaque:
Some of the tests were creating connections when not needed, this
'apparetly' was leaving the internal state of the Networking API in a
bad state in which it would throw a SIGSEGV and will make the connection
of other tests fail. In this case, the Tls tests that use the Network
API.

Cleaning the tests and removing those badly managed connections ensured
that the SIGSEGV would not be thrown and got all the other tests to
work.

This is a blackbox, test now passes without problems.

Fixes: https://github.com/xamarin/maccore/issues/2048

### DIFF
--- a/tests/monotouch-test/Network/NWParametersTest.cs
+++ b/tests/monotouch-test/Network/NWParametersTest.cs
@@ -1,4 +1,4 @@
-﻿#if !__WATCHOS__
+#if !__WATCHOS__
 using System;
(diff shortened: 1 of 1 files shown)


mandel-macaque:
Crash is known https://github.com/xamarin/maccore/issues/581

PR #7338, opened 2019-10-30
[Networking] Add the NWFramer implementation.

mandel-macaque:
Uncommented the sources and update some mistakes after following the
sample provided by Apple. Initially tests were going to be added but
they resulted to be too flacky and would make the CI red too often to be
adding value.

Porting the sample will ensure that it works are the bindings are not
broken.


src/Network/NWFramer.cs
```diff
@@ -399,23 +399,24 @@ public bool ParseOutput (nuint minimumIncompleteLength, nuint maximumLength, Mem
...
 				var mValue = new Memory<byte>(bBuffer);
-				del (mValue, is_complete);
+				return del (mValue, is_complete);
 			}
+			return 0;
 		}
 
 		[BindingImpl (BindingImplOptions.Optimizable)]
-		public bool ParseInput (nuint minimumIncompleteLength, nuint maximumLength, Memory<byte> tempBuffer, Action<Memory<byte>, bool> handler)
+		public bool ParseInput (nuint minimumIncompleteLength, nuint maximumLength, Memory<byte> tempBuffer, Func<Memory<byte>, bool, int> handler)
```
spouliot: `Func<Memory<byte>, bool, int>` -> use a custom delegate because I can't guess that the arguments should be

src/Network/NWFramer.cs
```diff
@@ -180,21 +180,22 @@ static void TrampolineOutputHandler (IntPtr block, OS_nw_framer framer, OS_nw_pr
...
 				var nwFramer = new NWFramer (framer, owns: true);
-				del (nwFramer);
+				return del (nwFramer);
 			}
+			return 0;
 		}
 
 		[BindingImpl (BindingImplOptions.Optimizable)]
-		public Action<NWFramer> InputHandler {
+		public Func<NWFramer, int> InputHandler {
```
spouliot: `Func<NWFramer, int>` hard to guess that the `int` is without reading the docs
and the best way not to write docs is to have a custom delegate :)

src/Network/NWFramer.cs
```diff
@@ -180,21 +183,22 @@ static void TrampolineOutputHandler (IntPtr block, OS_nw_framer framer, OS_nw_pr
 		[DllImport (Constants.NetworkLibrary)]
 		unsafe static extern void nw_framer_set_input_handler (OS_nw_framer framer, void *input_handler);
 
-		delegate void nw_framer_set_input_handler_t (IntPtr block, OS_nw_framer framer);
+		delegate int nw_framer_set_input_handler_t (IntPtr block, OS_nw_framer framer);
```
Therzok: this is size_t.

Therzok: `nint`

src/Network/NWFramer.cs
```diff
@@ -399,23 +402,24 @@ public bool ParseOutput (nuint minimumIncompleteLength, nuint maximumLength, Mem
 		[DllImport (Constants.NetworkLibrary)]
 		static extern unsafe bool nw_framer_parse_input (OS_nw_framer framer, nuint minimum_incomplete_length, nuint maximum_length, byte *temp_buffer, ref BlockLiteral parse);
 
-		delegate void nw_framer_parse_input_t (IntPtr block, IntPtr buffer, nuint buffer_length, bool is_complete);
+		delegate int nw_framer_parse_input_t (IntPtr block, IntPtr buffer, nuint buffer_length, bool is_complete);
 		static nw_framer_parse_input_t static_ParseInputHandler = TrampolineParseInputHandler;
 
 		[MonoPInvokeCallback (typeof (nw_framer_parse_input_t))]
-		static void TrampolineParseInputHandler (IntPtr block, IntPtr buffer, nuint buffer_length, bool is_complete)
+		static int TrampolineParseInputHandler (IntPtr block, IntPtr buffer, nuint buffer_length, bool is_complete)
```
Therzok: nuint

mandel-macaque:
@spouliot fixe, and retargeted the PR.

spouliot:
@mandel-macaque 
> Happens to the *hest* of us
It sounds like a middle ground between "best" and "rest" ;-)

PR #7339, opened 2019-10-30
[Runtime] Ensure that all comparisons between signed and unsigned values are fixed.

mandel-macaque:
Enable the -Wsign-compare which will raise issues when a comparison
between signed and unsigned values could produce an incorrect result
and fix all the raised warnings.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -1598,7 +1598,7 @@ objc_skip_type (const char *type)
 	}
 }
 
-int
+unsigned long
 xamarin_objc_type_size (const char *type)
 {
 	const char *original_type = type;
@@ -1642,7 +1642,7 @@ xamarin_objc_type_size (const char *type)
 			xamarin_assertion_message ("Unhandled type encoding: %s", type);
 			break;
 		case _C_ARY_B: {
-			int size = 0;
+			unsigned long size = 0;
 			int len = atoi (type+1);
 			do {
 				type++;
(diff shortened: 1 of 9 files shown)


runtime/runtime.m
```diff
@@ -1598,7 +1598,7 @@ -(void) xamarinSetGCHandle: (int) gc_handle;
 	}
 }
 
-int
+unsigned long
```
rolfbjarne: Why `long` and not continue using `int`? Not that it matters much, I'm just curious.

mandel-macaque: is the defined value for size_t I think I could change it to use size_t which is probably nicer, what do you think?

rolfbjarne: I don't think it matters much, the current code is fine with me.

runtime/xamarin-classic/runtime.h
```diff
@@ -169,7 +169,7 @@ void			xamarin_set_bundle_path (const char *path); /* Public API */
 MonoObject *	xamarin_get_managed_object_for_ptr (id self, guint32 *exception_gchandle);
 MonoObject *	xamarin_get_managed_object_for_ptr_fast (id self, guint32 *exception_gchandle);
 void			xamarin_check_for_gced_object (MonoObject *obj, SEL sel, id self, MonoMethod *method, guint32 *exception_gchandle);
-int				xamarin_objc_type_size (const char *type);
+unsigned long	xamarin_objc_type_size (const char *type);
```
wachaudh: nitpicking but the indentation is inconsistent versus the declarations above/below.

PR #7341, opened 2019-10-30
[runtime] Remove some classic (pre-unified) code that is not used anymore.

spouliot:


spouliot:
unrelated watchOS HE0038 failure -> https://github.com/xamarin/maccore/issues/581

PR #6103, opened 2019-05-22
[foundation] Add custom trust/certificate validation to NSUrlSessionHandler. Fix #4170

spouliot:
Basic application (size) for doing an `HttpClient.GetAsync`, release/llvm, 64bits only

- NSUrlSessionHandler (master): 6.4 MB
- NSUrlSessionHandler (PR #5936): 7.7 MB
- NSUrlSessionHandler (this PR): 6.4 MB

The size increase occurs because of the reference to .net `X509*` types.
This brings a lot of additional code, including managed cryptographic
code, inside the application - even when the feature is **not** used.

The solution is to expose an API that only use native (OS) types, which
are mostly already part of the application. This has a very low impact
on existing applications.

It's still possible to hook back to .NET validation if needed (it should
not in most cases) but, in this case, the extra price will only be
_paid_ if used (and can be lower if the code is needed by something else
from the application).

In comparison using other `HttpClient` handler produce app sizes of

- HttpClientHandler (managed): 10.4 MB
- CFNetworkHandler: 6.8 MB

Based on/supersede https://github.com/xamarin/xamarin-macios/pull/5733
Fix https://github.com/xamarin/xamarin-macios/issues/4170

PR #7346, opened 2019-10-31
Bring HttpClient from CoreFX.

baulig:
This bumps Mono to use https://github.com/mono/mono/pull/17645 (which is the 2019-10 backport of https://github.com/mono/mono/pull/17628).

The big user-visible change is in regards to certificate validation, everything below are just some minor adjustments to tests.

## SocketsHttpHandler

CoreFX uses a completely new `HttpClientHandler` implementation called `SocketsHttpHandler`, which you can find [here](https://github.com/dotnet/corefx/tree/release/3.0/src/System.Net.Http/src/System/Net/Http/SocketsHttpHandler).

Since this is not based on the web stack anymore, it does not use any of the related APIs such as `ServicePointManager` or `WebException`.

## Certificate Validation Changes

There is a new API called `HttpClientHandler.ServerCertificateCustomValidationCallback`.  You can find the documentation [here](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=netframework-4.8) and the code [here](https://github.com/dotnet/corefx/blob/c1778515a3bee34cc09c757b5563d0af0c8b1e99/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Unix.cs#L154)

The `ServicePointManager.ServerCertificateValidationCallback` is no longer invoked and on certificate validation failure, `AuthenticationException` (from `System.Security.Authentication`) is thrown instead of `WebException`.

At the moment, the `NSUrlSessionHandler` still uses it's own validation callback and also still throws `WebException` on failure; we should probably look into making this consistent with the other handlers.

## Minor adjustments related to internal Mono APIs

* `HttpContent.SerializeToStreamAsync()` is now `protected` (changed from `protected internal`).
  - src/Foundation/NSUrlSessionHandler.cs: changed overload accordingly.
  - src/System.Net.Http/CFContentStream.cs: likewise.

* `HttpHeaders.GetKnownHeaderKind()` is an internal Mono API.
   There is a new internal API called `System.Net.Http.PlatformHelper.IsContentHeader(key)` which exists in both the old as well as the new implementation.
   The correct way of doing it with the CoreFX handler is
   `HeaderDescriptor.TryGet (key, out var descriptor) && descriptor.HeaderType == HttpHeaderType.Content`

## Minor adjustments to tests.

* `HttpClientHandler.MaxRequestContentBufferSize` is now longer supported, you can set it to any non-negative value, the getter will always return 0.  See the [source code](https://github.com/dotnet/corefx/blob/c1778515a3bee34cc09c757b5563d0af0c8b1e99/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Core.cs#L18) for details.
  - tests/linker/ios/link sdk/HttpClientHandlerTest.cs: removed assertion from test.

* `HttpMessageInvoker.handler` is a `protected private` field - in the CoreFX handler, it is called `_handler` and `private`.  This is accessed via reflection by some of the tests, which are now using the new name.
  - tests/mmptest/src/MMPTest.cs: here
  - tests/mtouch/MTouch.cs: here

* tests/monotouch-test/System.Net.Http/MessageHandlers.cs:
  Adjust `RejectSslCertificatesServicePointManager` to reflect the certificate validation changes described above.
  - FIXME: There was an `Assert.Ignore()` related to `NSUrlSessionHandler` and macOS 10.10;
    I removed that to reenable the test because the description linked to an old issue in the private repo that was referenced by several "Merged" PR's, so it looked to me that this might have already been fixed - and I also didn't see why it would fail there.


spouliot:
Changes LGTM

WRT `public Func<System.Net.Http.HttpRequestMessage,System.Security.Cryptography.X509Certificates.X509Certificate2,System.Security.Cryptography.X509Certificates.X509Chain,System.Net.Security.SslPolicyErrors,bool> ServerCertificateCustomValidationCallback { get; set; }`

we can't expose this in `NSUrlSessionHandler` without a huge impact on app size (about 1.3MB extra code, see https://github.com/xamarin/xamarin-macios/pull/6103) as it brings most of the crypto stack in the app (which might not be needed in many cases). 

However we can document how to map it with the existing certificate handler - so the _price to pay_ becomes optional for people wanting to share/re-use the same code.

filipnavara:
> However we can document how to map it with the existing certificate handler - so the price to pay becomes optional for people wanting to share/re-use the same code.

Would it make sense to expose that as extension method? Presumably it would get linked out if not used.

spouliot:
@filipnavara no, in this case that would not help. The reason is that it's a delegate used as a callback.

IOW the handler code (not the user code) has to create an `X509Certificate2` and a `X509Chain` to be able to call the delegate (and your code). Even if the property is `null` it will be reachable (part of the code will check it for null).

Those types will recursively mark a lot of other (mostly security/crypto) types that most application are likely not using (if using the `NSUrlSessionHandler` for example).

This was the problem with the original approach in https://github.com/xamarin/xamarin-macios/pull/5733/files#diff-7f619be41b43eaaed811ff8b330f769cR511 that resulted in a 1.3MB app size increase.

The (existing) native handler version of the callback workaround this by using fewer types (mostly only `SecTrust`) that have nearly no other dependencies (p/invokes to native iOS functions).

The trick is (or will be) to set that (existing) delegate to call the one defined in the new handler, i.e. get the `X509Certificate2` (and optionally a `X509Chain`, less useful for pinning) out of `SecTrust` and call the delegate. That way only people using the feature will pay the extra price (in app size).

filipnavara:
Perhaps I should have been more clear but that approach still sounds like what I had in mind:
- `NSUrlSessionHandler` would expose only callback with native macOS objects, as it does now
- There would be `IDisposable NSUrlSessionHandlerExtensions.RegisterServerValidationCallback(this NSUrlSessionHandler handler, Func<System.Net.Http.HttpRequestMessage,System.Security.Cryptography.X509Certificates.X509Certificate2,System.Security.Cryptography.X509Certificates.X509Chain,System.Net.Security.SslPolicyErrors,bool> validationCallback)` (or similar prototype) that would register a delegate that translates from native Cocoa objects into the X509* ones and call `validationCallback`

mandel-macaque:
@spouliot we can adde the extension method, I can create a PR with it.

baulig:
build

xamarin-release-manager:
❌ [Build was (probably) aborted](https://jenkins.internalx.com/job/macios/job/PR-7346/2/display/redirect)

🔥 [Jenkins job](https://jenkins.internalx.com/job/macios/job/PR-7346/2/display/redirect) (on internal Jenkins) failed in stage(s) 'Build' 🔥 : hudson.AbortException: script returned exit code 2

🔥 [Build failed](https://jenkins.internalx.com/job/macios/job/PR-7346/2//console) 🔥

xamarin-release-manager:
❌ [Build was (probably) aborted](https://jenkins.internalx.com/job/macios/job/PR-7346/3/display/redirect)

🔥 [Jenkins job](https://jenkins.internalx.com/job/macios/job/PR-7346/3/display/redirect) (on internal Jenkins) failed in stage(s) 'Packaging' 🔥 : hudson.AbortException: script returned exit code 2

✅ [Build succeeded](https://jenkins.internalx.com/job/macios/job/PR-7346/3//console)
    ⚠️ Mono built from source

baulig:
build

baulig:
build

rolfbjarne:
build

xamarin-release-manager:
❌ [Build was (probably) aborted](https://jenkins.internalx.com/job/macios/job/PR-7346/4/display/redirect)

🔥 [Jenkins job](https://jenkins.internalx.com/job/macios/job/PR-7346/4/display/redirect) (on internal Jenkins) failed in stage(s) 'Packaging' 🔥 : hudson.AbortException: script returned exit code 2

✅ [Build succeeded](https://jenkins.internalx.com/job/macios/job/PR-7346/4//console)
    ⚠️ Mono built from source

xamarin-release-manager:
❌ [Build was (probably) aborted](https://jenkins.internalx.com/job/macios/job/PR-7346/5/display/redirect)

🔥 [Jenkins job](https://jenkins.internalx.com/job/macios/job/PR-7346/5/display/redirect) (on internal Jenkins) failed in stage(s) 'Packaging' 🔥 : hudson.AbortException: script returned exit code 2

✅ [Build succeeded](https://jenkins.internalx.com/job/macios/job/PR-7346/5//console)
    ⚠️ Mono built from source

rolfbjarne:
Packaging fails due to https://github.com/xamarin/maccore/issues/2053.

@mandel-macaque could you have a look?

mandel-macaque:
Merging, and I'll fix the pkg issue in a diff PR.

xamarin-release-manager:
❌ [Build was (probably) aborted](https://jenkins.internalx.com/job/macios/job/PR-7346/6/display/redirect)

🔥 [Jenkins job](https://jenkins.internalx.com/job/macios/job/PR-7346/6/display/redirect) (on internal Jenkins) failed in stage(s) 'Build' 🔥 : hudson.AbortException: script returned exit code 2

✅ [Provisioning succeeded](https://jenkins.internalx.com/job/macios/job/PR-7346/6//console)
🔥 [Build failed](https://jenkins.internalx.com/job/macios/job/PR-7346/6//console) 🔥

PR #7350, opened 2019-11-01
Added Static to UITraitCollection.FromPreferredContentSizeCategory

### LINKED ISSUE #7267: [Missing API] UITraitCollection.WithPreferredContentSizeCategory
### Actual Behavior
Method is not exposed as static like the other `UITraitCollection.From....` methods. Only available as a member method. 

I believe this is in error. 

```C#
var tc = new UITraitCollection()
                   .FromPreferredContentSizeCategory( /* ... */ );
```

### Expected Behavior
```C#
var tc = UITraitCollection.FromPreferredContentSizeCategory( /* ... */ );
```

Apple documentation link:

https://developer.apple.com/documentation/uikit/uitraitcollection/1771748-traitcollectionwithpreferredcont?language=objc

prmallic:
Fixes #7267

### DIFF
--- a/src/UIKit/UITraitCollection.cs
+++ b/src/UIKit/UITraitCollection.cs
@@ -14,6 +14,14 @@ using Foundation;
 namespace UIKit {
 #if !WATCH
 	public partial class UITraitCollection {
+#if !XAMCORE_4_0
+		[Obsolete ("Please use the static method instead.")]
+		public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
+		{
+			return FromPreferredContentSizeCategory (category.GetConstant ());
+		}
+#endif
+
 		public static UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
 		{
 			return FromPreferredContentSizeCategory (category.GetConstant ());
--- a/src/UIKit/UITraitCollection.cs
+++ b/src/UIKit/UITraitCollection.cs
@@ -14,18 +14,15 @@ using Foundation;
 namespace UIKit {
 #if !WATCH
 	public partial class UITraitCollection {
+
 #if !XAMCORE_4_0
-		[Obsolete ("Please use the static method instead.")]
+		[Obsolete ("Please use the static 'Create' method instead.")]
 		public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
-		{
-			return FromPreferredContentSizeCategory (category.GetConstant ());
-		}
+			=> Create (category);
 #endif
 
-		public static UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
-		{
-			return FromPreferredContentSizeCategory (category.GetConstant ());
-		}
+		public static UITraitCollection Create (UIContentSizeCategory category)
+			=> FromPreferredContentSizeCategory (category.GetConstant ());
 	}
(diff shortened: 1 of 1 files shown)


src/UIKit/UITraitCollection.cs
```diff
@@ -14,7 +14,7 @@
 namespace UIKit {
 #if !WATCH
 	public partial class UITraitCollection {
-		public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
+		public static UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
```
mandel-macaque: As @spouliot mentioned, it is changing the API. Since we are moving to static.. I wonder if we can be smart, leave the instance  one and call the static one from that. Expose both and remove the instance one on XAMCORE_4.

prmallic: Something like - 
```
#if XAMCORE_4_0
		[Obsolete ("Please use the static method instead.")]
#endif
		public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
		{
			return FromPreferredContentSizeCategory (category.GetConstant ());
		}

#if XAMCORE_4_0
		public static UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
		{
			return FromPreferredContentSizeCategory (category.GetConstant ());
		}
#endif
```
should work?

prmallic: Also, the instance one is already calling an internal static method

mandel-macaque: Close, we could name the static method create since it is a factory method, and do:
```csharp
#if !XAMCORE_4_0
[Obsolete ("Please use the static method 'StaticFromPreferredContentSizeCategory' instead.")]
public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category) => Create (category);
#endif

public static UITraitCollection Create (UIContentSizeCategory category)
{
  return FromPreferredContentSizeCategory (category.GetConstant ());
}
```

@spouliot  thoughts?

spouliot: close but not quite :)

We want to avoid breaking changes _before_ `XAMCORE_4_0` so we add an `[Obsolete]` until it's defined. Once it's define then we can remove the old code (and attribute) from the build. E.g.

```csharp
#if !XAMCORE_4_0
[Obsolete ("Please use the static method 'StaticFromPreferredContentSizeCategory' instead.")]
public UITraitCollection FromPreferredContentSizeCategory (UIContentSizeCategory category)
{
  return FromPreferredContentSizeCategory (category.GetConstant ());
}
#endif

public static UITraitCollection StaticFromPreferredContentSizeCategory (UIContentSizeCategory category)
{
  return FromPreferredContentSizeCategory (category.GetConstant ());
}
```

prmallic: Actually, won't this cause an error considering we have a static and an instance method of the same signature?

mandel-macaque:
Expected compilation failure:

 UIKit/UITraitCollection.cs(25,35): error CS0111: Type 'UITraitCollection' already defines a member called 'FromPreferredContentSizeCategory' with the same parameter types

That is why I mentioned to use Create. I'll try to find a name that works or a nice workaround.

mandel-macaque:

PR #7351, opened 2019-11-01
[runtime] Merge two identical code paths

spouliot:


runtime/trampolines.m
```diff
@@ -80,8 +80,6 @@
 	/* Any changes in this method probably need to be reflected in the static registrar as well */
 	switch (type [0]) {
 		case _C_CLASS:
```
wachaudh: shouldn't the case be removed all together?

wachaudh: shouldn't something be returned in this

runtime/trampolines.m
```diff
@@ -80,8 +80,6 @@
...
-			return xamarin_get_handle_for_inativeobject (retval, exception_gchandle);
```
wachaudh: Isn't the function expected to return some sort of void pointer? If the function hits this empty case it will never return.

spouliot: it's not an empty case, i.e. there's no `break` or `continue`, so it just share the same code path as `_C_SEL`

mandel-macaque: @wachaudh that is what they mean when they say a switch case will fall-through.

mandel-macaque:

PR #7352, opened 2019-11-01
[runtime] Free `env-argv` in `xamarin-main`

spouliot:
`get_mono_env_options` allocate memory so it must [1] be freed.

[1] to be polite since it will be freed when the app exit, which
will free everything anyway

### DIFF
--- a/runtime/launcher.m
+++ b/runtime/launcher.m
@@ -669,6 +669,7 @@ int xamarin_main (int argc, char **argv, enum XamarinLaunchMode launch_mode)
 		}
 
 		free (new_argv);
+		free (env_argv);
 	}
 
 	return rv;


mandel-macaque:

whitneyschmidt:
I looked through the documentation for ObjC/@releasepool and am a little confused.

It seems like the `@autorelease` automatically frees memory of objects allocated within the `@autorelease` block when it ends, unless you specify `retain`. How does this interact with our calls to `free` and why do we call `free` when objects are `released` by default at the end of the block?

Text blob from some probably outdated Apple docs:

In a garbage-collected environment, there is no need for autorelease pools. You may, however, write a framework that is designed to work in both a garbage-collected and reference-counted environment. In this case, you can use autorelease pools to hint to the collector that collection may be appropriate. In a garbage-collected environment, sending a drain message to a pool triggers garbage collection if necessary; release, however, is a no-op. In a reference-counted environment, drain has the same effect as release. Typically, therefore, you should use drain instead of release.

spouliot:
@whitneyschmidt good observations :)

* `@autorelease` only works for reference-counted objects (e.g. NSObject) so it won't help in the case where `malloc` was used. That was the case here so the call to `free` is needed.

* _In a garbage-collected environment,_ yes, that's from old, outdated documentation from the time Apple had it's own GC. It also would not mean it applied to us (being largely in a GC-world), nor would it apply when we use manual memory allocation.

PR #7357, opened 2019-11-02
Remove redundant class_ptr initialization

Therzok:
This is already done by the generated code, as seen below.

```csharp
static NSApplication()
{
...
selWindowWithWindowNumber_Handle = Selector.GetHandle ("windowWithWindowNumber:");
selWindowsHandle = Selector.GetHandle ("windows");
selWindowsMenuHandle = Selector.GetHandle ("windowsMenu");
class_ptr = Class.GetHandle ("NSApplication");
class_ptr = Class.GetHandle ("NSApplication");
}
```

With this change, we've also removed the static constructor, therefore all the fields are going to be initialized when they are used, as opposed to being forcefully loaded on the first access. This may improve startup times

### DIFF
--- a/src/AppKit/NSApplication.cs
+++ b/src/AppKit/NSApplication.cs
@@ -37,10 +37,6 @@ namespace AppKit {
 
 		private static Thread mainThread;
 
-		static NSApplication () {
-			class_ptr = Class.GetHandle ("NSApplication");
-		}
-
 		[DllImport (Constants.AppKitLibrary)]
 		extern static int /* int */ NSApplicationMain (int /* int */ argc, string [] argv);
 


Therzok:
HE0038 failed, which is known.

Therzok:
So, the generated code (ilspy on Xamarin.Mac) contains `class_ptr = Class.GetHandle ("NSApplication");` *twice* at the end of the _static constructor_. What I meant to say is that the manual value init is _redundant_.

All that happens eagerly in the static constructor, because of `static NSApplication()` constructor being present, thus all fields need to be initialized because of the static constructor.

With this change, all fields are initialized only on use.

PR #7391, opened 2019-11-08
[CoreFoundation] Ensure that we are thread safe.

mandel-macaque:
CFBundleGetAllBundles as per Apple docs is not thread safe, this means
that when we loop over the bundles, and use a property, the value is
updated by a different thread. This means that in some weird cases we
get an IndexOutOfBounds exception. We now store the count and use it in
the loop.

Fixes: https://github.com/xamarin/maccore/issues/940

### DIFF
--- a/src/CoreFoundation/CFBundle.cs
+++ b/src/CoreFoundation/CFBundle.cs
@@ -103,9 +103,18 @@ namespace CoreFoundation {
 		
 		public static CFBundle[] GetAll ()
 		{
+			// as per apple documentation: 
+			// CFBundleGetAllBundles
+			//
+			// 	'This function is potentially expensive and not thread-safe'
+			//
+			// This means, that we should not trust the size of the array, since is a get and
+			// might be modified by a diff thread. We just use the size returned by the first
+			// count to avoid a System.IndexOutOfRangeException when looping
 			using (var cfBundles = new CFArray (CFBundleGetAllBundles ())) {
-				var managedBundles = new CFBundle [cfBundles.Count];
-				for (int index = 0; index < cfBundles.Count; index++) {
+				var bundleCount = cfBundles.Count;
+				var managedBundles = new CFBundle [bundleCount];
+				for (int index = 0; index < bundleCount; index++) {
 					// follow the get rule, we do not own the object
 					managedBundles [index] = new CFBundle (cfBundles.GetValue (index), false);
 				}


src/CoreFoundation/CFBundle.cs
```diff
@@ -103,9 +103,18 @@ public static CFBundle[] GetBundlesFromDirectory (NSUrl directoryUrl, string bun
...
+			//
+			// This means, that we should not trust the size of the array, since is a get and
+			// might be modified by a diff thread. We just use the size returned by the first
+			// count to avoid a System.IndexOutOfRangeException when looping
 			using (var cfBundles = new CFArray (CFBundleGetAllBundles ())) {
-				var managedBundles = new CFBundle [cfBundles.Count];
-				for (int index = 0; index < cfBundles.Count; index++) {
+				var bundleCount = cfBundles.Count;
+				var managedBundles = new CFBundle [bundleCount];
+				for (int index = 0; index < bundleCount; index++) {
```
wachaudh: Why would `cfBundles` be getting updated if `CFBundleGetAllBundles()` is called in another thread? Type `CFArray` is immutable so `cfBundles` can't be changed, right?

mandel-macaque: Looking at the C code we have the following:

```c
FArrayRef CFBundleGetAllBundles(void) {
    // To answer this properly, we have to have created the static bundles!
#if !defined(AVOID_WEAK_COLLECTIONS)
    static CFMutableArrayRef externalAllBundles = NULL;
#endif /* AVOID_WEAK_COLLECTIONS */
    CFArrayRef bundles;
    __CFSpinLock(&CFBundleGlobalDataLock);
    _CFBundleEnsureAllBundlesUpToDateAlreadyLocked();
#if defined(AVOID_WEAK_COLLECTIONS)
    bundles = _allBundles;
#else /* AVOID_WEAK_COLLECTIONS */
    if (!externalAllBundles) {
        CFArrayCallBacks nonRetainingArrayCallbacks = kCFTypeArrayCallBacks;
        nonRetainingArrayCallbacks.retain = NULL;
        nonRetainingArrayCallbacks.release = NULL;
        externalAllBundles = CFArrayCreateMutable(kCFAllocatorSystemDefault, 0, &nonRetainingArrayCallbacks);
    }
    CFArrayRemoveAllValues(externalAllBundles);
    for (id value in _allBundles) CFArrayAppendValue(externalAllBundles, value);
    bundles = externalAllBundles;
#endif /* AVOID_WEAK_COLLECTIONS */
    __CFSpinUnlock(&CFBundleGlobalDataLock);
    return bundles;
}
```

The important thing there that I noticed:

```c
externalAllBundles = CFArrayCreateMutable
```
Now, if we are not thread safe as apple states. We can be getting bundles to be a CFMutableArray that is getting a new bundle added. When we look, the Count changes (because our implementation of the property calls a C call on the pointer) and we get the exception.

wachaudh: 🍎 🤦‍♂ 😡

mandel-macaque:

mandel-macaque:

mandel-macaque:
@monojenkins backport d16-5

mandel-macaque:
@monojenkins backport d16-4

src/CoreFoundation/CFBundle.cs
spouliot: @mandel-macaque the addition are inside a lock so you should not get different number of items (at least not for the reason you mention).

mandel-macaque: Well, the pointer is alway the same one, so, I don't think the Lock is very useful since it does not lock the GetCount call. So that lock is just useful if several calls are happening, but getting the count is not valid. 

The way I see it:

1. Thread 1 gets bundles (count 5 for example).
2. We start the loop.
3. Thread 2 gets bundles (no other thread can call the method) and new bundles are added.
4. Thread 1 calls count once again, but we got a larger count!!! Got exception.

We could use a list, append and then return as an array. But I have 0 confidence in the order of the bundles. So we would have to, get count, check from the previous result. If it is the same, continue, else, re-call. Not a big fan of it :/

rolfbjarne: Apple's code is just plain broken, there's just no way to use the return value from `CFBundleGetAllBundles` in any safe way, it can be modified by any other thread at any time :/

I think the best thing we can do is to document that this function should not be used (or as Apple says: only while debugging).

spouliot: it's indeed broken but if we clone it then the items will be constant (even if another call would return something different), which is what developers would expect of a `CFArray`. 

This, at least, removes the `IndexOutOfBoundException` which makes it unusable.

Still it would be a good idea to add
```csharp
[Advice ("Subsequent calls to this method will return a different array instance, possibly with different items.")]
```

We can also use the same lock (as the code above) if the clone operation is not atomic.

PR #7401, opened 2019-11-11
[Tests] If we have server errors. Mark test as inconclusive.

mandel-macaque:
If we are getting errors (500,401..) do not mark a link all test as a
failure, but as inconclusive.

Fixes: https://github.com/xamarin/maccore/issues/2056

### DIFF
--- a/tests/linker/mac/LinkAnyTest.cs
+++ b/tests/linker/mac/LinkAnyTest.cs
@@ -31,7 +31,7 @@ namespace LinkAnyTest {
 		{
 			// do not use GetStringAsync, we are going to miss useful data, such as the resul code
 			using (var client = new HttpClient ()) {
-				HttpResponseMessage response = await client.GetAsync ("http://msdn.microsoft.com"); 
+				HttpResponseMessage response = await client.GetAsync ("http://example.com"); 
 				if(!response.IsSuccessStatusCode) {
 					requestError = true;
 					statusCode = response.StatusCode;
--- a/tests/linker/mac/LinkAnyTest.cs
+++ b/tests/linker/mac/LinkAnyTest.cs
@@ -21,15 +21,27 @@ namespace LinkAnyTest {
 		}
 
 		static bool waited;
+		static bool requestError;
+		static HttpStatusCode statusCode;
+
+
 		// http://blogs.msdn.com/b/csharpfaq/archive/2012/06/26/understanding-a-simple-async-program.aspx
 		// ref: https://bugzilla.xamarin.com/show_bug.cgi?id=7114
 		static async Task GetWebPageAsync ()
 		{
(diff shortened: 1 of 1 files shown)


mandel-macaque:
All failures are https://github.com/xamarin/maccore/issues/581

tests/linker/mac/LinkAnyTest.cs
```diff
@@ -21,15 +21,27 @@ public void AES ()
...
+			// do not use GetStringAsync, we are going to miss useful data, such as the resul code
+			using (var client = new HttpClient ()) {
+				HttpResponseMessage response = await client.GetAsync ("http://msdn.microsoft.com"); 
+				if(!response.IsSuccessStatusCode) {
+					requestError = true;
+					statusCode = response.StatusCode;
+				} else {
+					string content = await response.Content.ReadAsStringAsync ();
+					waited = true;
+					bool success = !String.IsNullOrEmpty (content);
```
whitneyschmidt: tiny nit: it's a little confusing to have a whole `bool success` variable that we expect to be true always. 

In the context of http request code, I would expect a variable named `success` to indicate something like `response.IsSuccessStatusCode`, but it seems like by the time we hit the `else` branch we're already in the success case :)

Otherwise, looks good to me.

mandel-macaque: good point. I think we are just testing that async/await works. If that is the case, even the errno would be good.

@rolfbjarne do you know why do we have this tests exactly? I looked at the bugzilla issue and might just be for the async support.

rolfbjarne: The test was added here: https://github.com/xamarin/maccore/commit/ca7dd9294d173893fbb2b635175141797261ead8, and yes, it looks like it was just to make sure async/await works.

But did you try just changing the url to `http://example.com`? That's a url we're already using in other places, and which will fail if the url goes down. It's possible http://msdn.microsoft.com will always return us a 500 code, and it doesn't feel right to ignore any such failures (eventually we'll end up ignoring legitimate failures).

spouliot: make sense to reduce the number of endpoint our tests depends on... maybe we should have them as constants (in a different PR)

mandel-macaque:
@spouliot @rolfbjarne I'll update the url, seems like msd is giving a lot of problems. Ignoring anyway seems good, since inconclusive in this case is what we want. BUT! Maybe just the HttpResponse with the errno is good enough to test the async pattern, and in this particular case we do not care about the content of the response.

mandel-macaque:

mandel-macaque:

mandel-macaque:
@monojenkins backport d16-5

mandel-macaque:
@monojenkins backport d16-4

PR #7403, opened 2019-11-11
[CoreFoundation] Add Clone method for CFArray.

mandel-macaque:
When working on https://github.com/xamarin/maccore/issues/940 we noticed
that clone the array would be a better approach.

The CFArrayCreateCopy add some nice things:

* The pointer values from theArray are copied into the new array.
* The values are also retained by the new array.
* The count of the new array is the same as theArray
* The new array uses the same callbacks as theArray. [IMPORTANT]

With this in, we can have a better fix for https://github.com/xamarin/maccore/issues/940

### DIFF
--- a/src/CoreFoundation/CFArray.cs
+++ b/src/CoreFoundation/CFArray.cs
@@ -35,6 +35,8 @@ using Foundation;
 using ObjCRuntime;
 
 using CFIndex = System.nint;
+using CFArrayRef = System.IntPtr;
+using CFAllocatorRef = System.IntPtr;
 
 namespace CoreFoundation {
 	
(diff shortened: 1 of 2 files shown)


src/CoreFoundation/CFArray.cs
```diff
@@ -35,6 +35,8 @@
 using ObjCRuntime;
 
 using CFIndex = System.nint;
+using CFArrayRef = System.IntPtr;
+using CFAllocatorRef = System.IntPtr;
```
wachaudh: Why do we name these like this and not just use `IntPtr`directly for the parmeter types in `Clone()`? There's other bindings that just use IntPtr directly.

mandel-macaque: It is easier to read. I would have changed all, but would have increase the diff. It does not change the output result of the binding.

dalexsoto: Haha nice question, but yeah just for readability.

giphy

mandel-macaque:
@spouliot CFArray is not public. Do you want me to change the internals visibility to be able to write the test?

spouliot:
@mandel-macaque forgot about that, it will be tested w/CFBundle anyway :)

mandel-macaque:

PR #7409, opened 2019-11-12
[runtime] Clarify redirect_io behaviour

spouliot:
Change `redirect_io` to be easier to re-use correctly, i.e. returning the
file descriptor (instead of `0`).

Comment why the file descriptor are never released. It makes sense but
it's not obvious (even less in `redirect_io`) without comments.

Using static fields also makes it easier for static analysis (locals are
easier to process).

Finally reuse, mostly, the same string for the error message, i.e.

```
$ strings runtime//.libs/mac/libxammac.x86_64.dylib | grep "Could not redirect"
Xamarin.Mac: Could not redirect %s to `%s': %s
```
versus
```
$ strings runtime//.libs/mac/libxammac.x86_64.dylib | grep "Could not redirect"
Xamarin.Mac: Could not redirect stdout to `%s': %s
Xamarin.Mac: Could not redirect stderr to `%s': %s
```

### DIFF
--- a/runtime/launcher.m
+++ b/runtime/launcher.m
@@ -42,12 +42,16 @@ redirect_io (int from_fd, const char *to_path)
 		return -1;
 	}
 
-	return 0;
+	return fd;
 }
 
 static void
 init_logdir (void)
 {
+	// If redirected we will not be closing the returned file descriptors anywhere.
+	// That's "by design" so they will keep logging as long as the app is alive.
+	static int redirected_stdout = -1;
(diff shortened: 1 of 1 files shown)


spouliot:
build

spouliot:
many, many HE0038 w/watchOS -> https://github.com/xamarin/maccore/issues/581
a few HTTP 500 errors -> fixed with inflight https://github.com/xamarin/xamarin-macios/pull/7401

PR #7411, opened 2019-11-12
[runtime] Zeroize dummy member of structure in binding-generator

spouliot:
This is not exposed so it should never be an issue - but just in
case something else goes wrong then it will only be 0 (and not
random values).


spouliot:
many, many HE0038 w/watchOS -> https://github.com/xamarin/maccore/issues/581
a few HTTP 500 errors -> fixed with inflight https://github.com/xamarin/xamarin-macios/pull/7401

PR #7420, opened 2019-11-13
[runtime] Filter isblank input to 8 bits values

spouliot:
Help static analyzer by clarifying that only `char` will be given to
`isblank` so it does not look like it could be used to index an array
and gets out of bounds (important since this is user supplied values).

Digging down into `isblank` will get you to
> Using tainted variable _c as an index to array _DefaultRuneLocale.__runetype.


spouliot:
Multiple HE0038 on watchOS -> https://github.com/xamarin/maccore/issues/581

PR #7421, opened 2019-11-13
[runtime] Catch and log exceptions in ObjC++ destructor

spouliot:
`xamarin_notify_dealloc` can throw an exception (depending on the mode
used).

Throwing exception in a destructor is problematic in [Obj]C++ and can
abort the application. That might be fine in some cases but there's
not much point in doing so when we're about to forget everything about
that specific object.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -2771,7 +2771,11 @@ xamarin_get_managed_method_for_token (guint32 token_ref, guint32 *exception_gcha
 XamarinObject::~XamarinObject ()
 {
 	// COOP: no managed memory access: any mode.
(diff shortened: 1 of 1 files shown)


spouliot:
> xammac tests/Mac Modern/Release: Failed (Test run crashed (exit code: 134).)

```
09:32:19.0184500 2019-11-13 09:32:19.018 xammac_tests[2226:54867970] error: mono_coop_mutex_lock Cannot transition thread 0x110def5c0 from STATE_BLOCKING with DO_BLOCKING
```

filed as https://github.com/xamarin/maccore/issues/2063

It seems unrelated but I'll re-run it again (to be sure) before merging

spouliot:
build

PR #7425, opened 2019-11-13
[CoreFoundation] CFBundle.GetAll better thread safe.

mandel-macaque:
The initial solution fixed the rance condition in which the index
changed, but that did not guarantee that we would get the correct
bundles. We now clone the CFArray (which also clones the callbacks set
to the array) and iterate over it to make sure Apple does not do evil
things while we are iteraing.

Better Fixes: https://github.com/xamarin/maccore/issues/940


mandel-macaque:
HE0038 issues: https://github.com/xamarin/maccore/issues/581

mandel-macaque:

mandel-macaque:
@monojenkins backport d16-5

mandel-macaque:

mandel-macaque:
@monojenkins backport d16-4

PR #7426, opened 2019-11-13
[Networking] Add missing Mac OS X APIs.

mandel-macaque:
Add missing APIs and complete the bindings for Xcode 11.

### DIFF
--- a/src/Network/NWEthernetChannel.cs
+++ b/src/Network/NWEthernetChannel.cs
@@ -1,0 +1,194 @@
+//
+// NWEthernetChannel.cs: Bindings the Network nw_ethernet_channel_t API.
+//
+// Authors:
+//   Manuel de la Pena (mandel@microsoft.com)
+//
+// Copyright 2019 Microsoft
+//
+#if MONOMAC
+using System;
+using System.Text;
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using ObjCRuntime;
+using Foundation;
+using CoreFoundation;
+
+using OS_nw_ethernet_channel=System.IntPtr;
+using OS_nw_interface=System.IntPtr;
+using OS_dispatch_data=System.IntPtr;
+
+namespace Network {
+
+	[NoWatch, NoTV, NoiOS, Mac (10,15)]
+	public delegate void NWEthernetChannelReceiveDelegate (DispatchData content, ushort vlanTag, string localAddress, string remoteAddress);
+
+	[NoWatch, NoTV, NoiOS, Mac (10,15)]
+	public enum NWEthernetChannelState 
+	{
+		Invalid = 0,
+		Waiting = 1,
+		Preparing = 2,
+		Ready = 3,
+		Failed = 4,
+		Cancelled = 5,
+	}
+
+	[NoWatch, NoTV, NoiOS, Mac (10,15)]
+	public class NWEthernetChannel : NativeObject {
+
+		internal NWEthernetChannel (IntPtr handle, bool owns) : base (handle, owns) {}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_ethernet_channel nw_ethernet_channel_create (ushort ether_type, OS_nw_interface networkInterface);
+
+		// we cannot pass an enum! As per docs in the headers:
+		// The custom EtherType to be used for all Ethernet frames in this channel. The
+		// EtherType is the two-octet field in an Ethernet frame, indicating the protocol
+		// encapsulated in the payload of the frame.  This parameter is in little-endian
+		// byte order.  Only custom EtherType values are supported. This parameter cannot
+		// be an EtherType already handled by the system, such as IPv4, IPv6, ARP, VLAN Tag,
+		// or 802.1x.
+		// 
+		// Calling processes must hold the "com.apple.developer.networking.custom-protocol"
+		// entitlement.
+		public NWEthernetChannel (ushort ethernetType, NWInterface networkInterface)
+		{
+			if (networkInterface == null)
+				throw new ArgumentNullException (nameof (networkInterface));
+
+			InitializeHandle (nw_ethernet_channel_create (ethernetType, networkInterface.Handle));
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_ethernet_channel_start (OS_nw_ethernet_channel ethernet_channel);
+
+		public void Start () => nw_ethernet_channel_start (GetCheckedHandle ()); 
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_ethernet_channel_cancel (OS_nw_ethernet_channel ethernet_channel);
+
+		public void Cancel () => nw_ethernet_channel_cancel (GetCheckedHandle ());
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_ethernet_channel_set_queue (OS_nw_ethernet_channel ethernet_channel, IntPtr queue);
+
+		public void SetQueue (DispatchQueue queue)
+		{
+			if (queue == null)
+				throw new ArgumentNullException (nameof (queue));
+			nw_ethernet_channel_set_queue (GetCheckedHandle (), queue.Handle);
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		static extern void nw_ethernet_channel_send (OS_nw_ethernet_channel ethernet_channel, OS_dispatch_data content, ushort vlan_tag, string remote_address, ref BlockLiteral completion);
+
+		delegate void nw_ethernet_channel_send_completion_t (IntPtr block, IntPtr error);
+		static nw_ethernet_channel_send_completion_t static_SendCompletion = TrampolineSendCompletion;
+
+		[MonoPInvokeCallback (typeof (nw_ethernet_channel_send_completion_t))]
+		static void TrampolineSendCompletion (IntPtr block, IntPtr error)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWError>> (block);
+			if (del != null) {
+				var err = error == IntPtr.Zero ? null : new NWError (error, owns: false);
+				del (err);
+				err?.Dispose ();
+			}
+		}
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void Send (ReadOnlySpan<byte> content, ushort vlanTag, string remoteAddress, Action<NWError> callback)
+		{
+			if (callback == null)
+				throw new ArgumentNullException (nameof (callback));
+
+			using (var dispatchData = DispatchData.FromByteBuffer (content.ToArray (), 0, content.Length)) {
+				BlockLiteral block_handler = new BlockLiteral ();
+				block_handler.SetupBlockUnsafe (static_SendCompletion, callback);
+
+				try {
+					nw_ethernet_channel_send (GetCheckedHandle (), dispatchData.GetHandle (), vlanTag, remoteAddress, ref block_handler);
+				} finally {
+					block_handler.CleanupBlock ();
+				}
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_ethernet_channel_set_receive_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ BlockLiteral *handler);
+
+		delegate void nw_ethernet_channel_receive_handler_t (IntPtr block, OS_dispatch_data content, ushort vlan_tag, byte[] local_address, byte[] remote_address);
+		static nw_ethernet_channel_receive_handler_t static_ReceiveHandler = TrampolineReceiveHandler;
+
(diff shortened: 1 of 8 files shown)


src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
+//
+// NWEthernetChannel.cs: Bindings the Network nw_ethernet_channel_t API.
+//
+// Authors:
+//   Manuel de la Pena (mandel@microsoft.com)
+//
+// Copyrigh 2018 Microsoft
```
prmallic:
```suggestion
// Copyright 2019 Microsoft
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+				IntPtr bufferAddress = IntPtr.Zero;
+				nuint bufferSize = 0;
+
+				if (content != IntPtr.Zero) {
+					dispatchData = new DispatchData (content, owns: false);
+					dataCopy = dispatchData.CreateMap (out bufferAddress, out bufferSize);
+				}
+
+				unsafe {
+					var spanData = new ReadOnlySpan<byte> ((void*)bufferAddress, (int)bufferSize);
```
prmallic:
```suggestion
					var spanData = new ReadOnlySpan<byte> ((void*) bufferAddress, (int) bufferSize);
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+					dataCopy.Dispose ();
+					dispatchData.Dispose ();
+				}
+			}
+		}
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void SetReceiveHandler (NWEthernetChannelReceiveDelegate handler)
+		{
+			unsafe
```
prmallic:
```suggestion
			unsafe {
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+
+		delegate void nw_ethernet_channel_state_changed_handler_t (IntPtr block, NWEthernetChannelState state, IntPtr error);
+		static nw_ethernet_channel_state_changed_handler_t static_StateChangesHandler = TrampolineStateChangesHandler;
+
+		[MonoPInvokeCallback (typeof (nw_ethernet_channel_state_changed_handler_t))]
+		static void TrampolineStateChangesHandler (IntPtr block, NWEthernetChannelState state, IntPtr error)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWEthernetChannelState, NWError>> (block);
+			if (del != null) {
+				var nwError = (error == IntPtr.Zero)? null : new NWError (error, owns: false);
```
prmallic:
```suggestion
				var nwError = (error == IntPtr.Zero) ? null : new NWError (error, owns: false);
```

src/Network/NWParameters.cs
```diff
@@ -177,6 +177,29 @@ public unsafe static NWParameters CreateUdp (Action<NWProtocolOptions> configure
 			return new NWParameters (ptr, owns: true);
 		}
 
+		[NoWatch, NoTV, NoiOS, Mac (10,15)]
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern IntPtr nw_parameters_create_custom_ip (byte custom_ip_protocol_number, void *configure_ip);
+
+		[NoWatch, NoTV, NoiOS, Mac (10,15)]
+		public unsafe static NWParameters CreateCustomIP (byte protocolNumber, Action<NWProtocolOptions> configureCustomIP = null)
```
prmallic: are there performance differences between declaring the function as `unsafe` vs encapsulating the code (i think most of it here) in an `unsafe` block?

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+			}
+		}
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void Send (ReadOnlySpan<byte> content, ushort vlanTag, string remoteAddress, Action<NWError> callback)
+		{
+			if (callback == null)
+				throw new ArgumentNullException (nameof (callback));
+
+			using (var dispatchData = DispatchData.FromByteBuffer (content.ToArray (), 0, content.Length)) {
```
Therzok: DispatchData should have a ROSpan overload

src/Network/NWParameters.cs
Therzok: Marking the whole method as unsafe does prevent some optimizations (no idea on a specific list) being applied to the whole method as just to the respective block of code - as unsafe code can be used in the whole method now.

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+				} finally {
+					block_handler.CleanupBlock ();
+				}
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_ethernet_channel_set_receive_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ void *handler);
+
+		delegate void nw_browser_set_state_changed_handler_t (IntPtr block, OS_dispatch_data content, ushort vlan_tag, byte[] local_address, byte[] remote_address);
```
rolfbjarne: Shouldn't this be named `nw_ethernet_channel_receive_handler_t` instead?

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+		[MonoPInvokeCallback (typeof (nw_browser_set_state_changed_handler_t))]
+		static void TrampolineReceiveHandler (IntPtr block, OS_dispatch_data content, ushort vlanTag, byte[] localAddress, byte[] remoteAddress)
+		{
+			var del = BlockLiteral.GetTarget<NWEthernetChannelReceiveDelegate> (block);
+			if (del != null) {
+				DispatchData dispatchData = null, dataCopy = null;
+				IntPtr bufferAddress = IntPtr.Zero;
+				nuint bufferSize = 0;
+
+				if (content != IntPtr.Zero) {
```
rolfbjarne: I think it would be better to use DispatchData in the delegate callback instead if ReadOnly<byte>, because it can be much faster, depending on how the callback is consumed. We could create a DispatchData.CreateMap overload that outputs a ReadOnly<byte> instance if we want to make it easier to consume.

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+					dispatchData = new DispatchData (content, owns: false);
+					dataCopy = dispatchData.CreateMap (out bufferAddress, out bufferSize);
+				}
+
+				unsafe {
+					var spanData = new ReadOnlySpan<byte> ((void*)bufferAddress, (int)bufferSize);
+					del (spanData, vlanTag, Encoding.UTF8.GetString (localAddress), Encoding.UTF8.GetString (remoteAddress));
```
rolfbjarne: What if `localAddress` or  `remoteAddress` are null?

src/Network/NWEthernetChannel.cs
rolfbjarne:
```suggestion
				var nwError = (error == IntPtr.Zero) ? null : new NWError (error, owns: false);
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+				try {
+					nw_ethernet_channel_set_receive_handler (GetCheckedHandle (), (void*) block_ptr_handler);
+				} finally {
+					block_handler.CleanupBlock ();
+				}
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_ethernet_channel_set_state_changed_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ void *handler);
```
rolfbjarne: More exact type is always welcome:

```suggestion
		unsafe static extern void nw_ethernet_channel_set_state_changed_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ BlockLiteral *handler);
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,208 @@
...
+			unsafe {
+				if (handler == null) {
+					nw_ethernet_channel_set_state_changed_handler (GetCheckedHandle (), null);
+					return;
+				}
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
+				block_handler.SetupBlockUnsafe (static_StateChangesHandler, handler);
+				try {
+					nw_ethernet_channel_set_state_changed_handler (GetCheckedHandle (), (void*) block_ptr_handler);
```
rolfbjarne: I believe this would work if the P/Invoke declaration is changed as suggested above:

```suggestion
					nw_ethernet_channel_set_state_changed_handler (GetCheckedHandle (), &block_handler);
```

src/Network/NWParameters.cs
```diff
@@ -177,6 +177,29 @@ public unsafe static NWParameters CreateUdp (Action<NWProtocolOptions> configure
 			return new NWParameters (ptr, owns: true);
 		}
 
+		[NoWatch, NoTV, NoiOS, Mac (10,15)]
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern IntPtr nw_parameters_create_custom_ip (byte custom_ip_protocol_number, void *configure_ip);
```
rolfbjarne:
```suggestion
		unsafe static extern IntPtr nw_parameters_create_custom_ip (byte custom_ip_protocol_number, BlockLiteral *configure_ip);
```

src/Network/NWEthernetChannel.cs
mandel-macaque: I have been thinking about updating DispatchData to be a more useful class. As it is atm is not very useful :/

src/Network/NWEthernetChannel.cs
spouliot: worth filing an issue (not the right PR to do this) so we don't forget

mandel-macaque: @spouliot @Therzok take a look https://github.com/xamarin/xamarin-macios/issues/7446 any other possible methods to add?

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,195 @@
...
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void SetStateChangesHandler (Action<NWBrowserState, NWError> handler)
+		{
+			unsafe {
+				if (handler == null) {
+					nw_ethernet_channel_set_state_changed_handler (GetCheckedHandle (), null);
+					return;
+				}
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
```
rolfbjarne: This line is redundant.

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,195 @@
...
+				try {
+					nw_ethernet_channel_send (GetCheckedHandle (), dispatchData.GetHandle (), vlanTag, remoteAddress, ref block_handler);
+				} finally {
+					block_handler.CleanupBlock ();
+				}
+			}
+		}
+
+		[DllImport (Constants.NetworkLibrary)]
+		unsafe static extern void nw_ethernet_channel_set_receive_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ void *handler);
```
rolfbjarne:
```suggestion
		static extern void nw_ethernet_channel_set_receive_handler (OS_nw_ethernet_channel ethernet_channel, /* [NullAllowed] */ BlockLiteral *handler);
```

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,195 @@
...
+		public void SetReceiveHandler (NWEthernetChannelReceiveDelegate handler)
+		{
+			unsafe {
+				if (handler == null) {
+					nw_ethernet_channel_set_receive_handler (GetCheckedHandle (), null);
+					return;
+				}
+
+				BlockLiteral block_handler = new BlockLiteral ();
+				BlockLiteral *block_ptr_handler = &block_handler;
```
rolfbjarne: This line becomes redundant with the other changes.

src/Network/NWEthernetChannel.cs
```diff
@@ -0,0 +1,195 @@
...
+				block_handler.SetupBlockUnsafe (static_ReceiveHandler, handler);
+				try {
+					nw_ethernet_channel_set_receive_handler (GetCheckedHandle (), (void*) block_ptr_handler);
```
rolfbjarne:
```suggestion
					nw_ethernet_channel_set_receive_handler (GetCheckedHandle (), &block_ptr_handler);
```

src/Network/NWEthernetChannel.cs
Therzok: Unsafe is needed with pointers.

rolfbjarne: Ah, that's right, I was thinking about something else.

PR #7428, opened 2019-11-13
[Runtime] Enable the -Wsign-conversion on trampolines-varargs files.

mandel-macaque:
Enable the flag that was disabled via pragmas.

Continuation of PR: https://github.com/xamarin/xamarin-macios/pull/7405


runtime/trampolines-varargs.h
```diff
@@ -16,7 +16,7 @@ struct XamarinCallState {
 	union {  // floating point return value
 		double double_ret;
 		float float_ret;
-		long long longlong_ret;
+		uint64_t ulonglong_ret;
```
rolfbjarne: What were the sign conversions warnings? Seems that fixing those without changing the field type would be a  easier and shorter fix.

runtime/trampolines-varargs.m
```diff
@@ -132,7 +128,7 @@
 	case _C_ULNG:
 	case _C_LNG_LNG:
 	case _C_ULNG_LNG:
-		state->longlong_ret = *(uint64_t *) mono_object_unbox (value);
```
mandel-macaque: @rolfbjarne the answer to your question is this case here. We could change it to long long, I did not want to loose precision and supposed you did the explicit cast for a reason.

runtime/trampolines-varargs.h
mandel-macaque: See https://github.com/xamarin/xamarin-macios/pull/7428#pullrequestreview-316590686

runtime/trampolines-varargs.m
rolfbjarne: The explicit cast is there because `mono_object_unbox` returns a void*.

This can just as easily be:

```suggestion
		state->longlong_ret = *(int64_t *) mono_object_unbox (value);
```

mandel-macaque: Reverted and made the commit with the suggestion.

mandel-macaque:

PR #7432, opened 2019-11-13
[Runtime] Add extra flags from preventing us to add code with warnings.

mandel-macaque:


mandel-macaque:
There are no new warnings, this is for future changes.

mandel-macaque:
Know issues: https://github.com/xamarin/maccore/issues/581

PR #7433, opened 2019-11-13
[Xcode11.3] Make CFBundle.GetAll thread safe.

mandel-macaque:


mandel-macaque:
Needed manual backport since monojenkins can't deal with it.

PR #7436, opened 2019-11-13
[d16-4][CoreFoundation] Make CFBundle.GetAll thread safe.

mandel-macaque:


dalexsoto:
I have to agree with @rolfbjarne on this one, I do not think this deserves a QB insertion at this point

mandel-macaque:
Closing then.

PR #7454, opened 2019-11-15
Keep track of SDK versions and OS versions separately. Fixes xamarin/maccore#2066.

rolfbjarne:
The latest SDK version and the latest OS version does not necessarily have to
match (for instance the iOS 13.2 SDK can support both iOS 13.2 and iOS 13.3),
so keep track of them separately.

Also use the latest OS version to determine which simulator to run, instead of
the latest SDK version (Xcode 11.3 ships with the iOS 13.2 SDK but only has an
iOS 13.3 simulator, not an iOS 13.2 simulator).

Fixes https://github.com/xamarin/maccore/issues/2066.


rolfbjarne:
Test failure is unrelated

* introspection/iOS 64-bits: https://github.com/xamarin/maccore/issues/2067

PR #7493, opened 2019-11-25
[xharness] Fix 32 bit device tests selection

VincentDondain:
With the `run-ios-32-tests` label, all iOS device tests were ignored.
This is because the ignore condition was incorrect, `IncludeiOS` is not set for 32 bit device tests.
Now we only enable/disable the 32 bit device tests based on `IncludeiOS32` which is set by the `run-ios-32-tests` label.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -22,7 +22,7 @@ namespace xharness
 		public bool IncludeAll;
 		public bool IncludeBcl;
 		public bool IncludeMac = true;
-		public bool IncludeiOS = true;
+		public bool IncludeiOS64 = true;
 		public bool IncludeiOS32 = true;
 		public bool IncludeiOSExtensions;
 		public bool ForceExtensionBuildOnly;
@@ -481,9 +481,9 @@ namespace xharness
 
 				var ps = new List<Tuple<TestProject, TestPlatform, bool>> ();
 				if (!project.SkipiOSVariation)
-					ps.Add (new Tuple<TestProject, TestPlatform, bool> (project, TestPlatform.iOS_Unified, ignored || !IncludeiOS));
+					ps.Add (new Tuple<TestProject, TestPlatform, bool> (project, TestPlatform.iOS_Unified, ignored || !IncludeiOS64));
 				if (project.MonoNativeInfo != null)
-					ps.Add (new Tuple<TestProject, TestPlatform, bool> (project, TestPlatform.iOS_TodayExtension64, ignored || !IncludeiOS));
+					ps.Add (new Tuple<TestProject, TestPlatform, bool> (project, TestPlatform.iOS_TodayExtension64, ignored || !IncludeiOS64));
 				if (!project.SkiptvOSVariation)
 					ps.Add (new Tuple<TestProject, TestPlatform, bool> (project.AsTvOSProject (), TestPlatform.tvOS, ignored || !IncludetvOS));
 				if (!project.SkipwatchOSVariation)
@@ -555,7 +555,7 @@ namespace xharness
 						TestName = project.Name,
 					};
 					build64.CloneTestProject (project);
-					projectTasks.Add (new RunDeviceTask (build64, Devices.Connected64BitIOS.Where (d => d.IsSupported (project))) { Ignored = !IncludeiOS });
+					projectTasks.Add (new RunDeviceTask (build64, Devices.Connected64BitIOS.Where (d => d.IsSupported (project))) { Ignored = !IncludeiOS64 });
 
 					var build32 = new XBuildTask {
 						Jenkins = this,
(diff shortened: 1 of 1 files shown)


VincentDondain:
FYI: tried locally and 124 tests are enabled (all 32 bit)

tests/xharness/Jenkins.cs
```diff
@@ -565,7 +565,7 @@ async Task<IEnumerable<TestTask>> CreateRunDeviceTasksAsync ()
 						TestName = project.Name,
 					};
 					build32.CloneTestProject (project);
-					projectTasks.Add (new RunDeviceTask (build32, Devices.Connected32BitIOS.Where (d => d.IsSupported (project))) { Ignored = !IncludeiOS || !IncludeiOS32 });
+					projectTasks.Add (new RunDeviceTask (build32, Devices.Connected32BitIOS.Where (d => d.IsSupported (project))) { Ignored = !IncludeiOS32 });
```
mandel-macaque: I don't think the fix is correct. As far as I can remember, the addition of the  !IncludeiOS is to ensure, that if ALL the iOS tests are ignored, the iOS32 label is not needed. So the old check is correct. That is:

1. Ignore 32b when !IncludeiOS (skip-ios label)
2. Ignore 32b when !IncludeiOS32 (ios label + skip-ios-32 labels)
3. !IncludeiOS && IncludeiOS32 Add 32b when all the other iOS tests are ignored (that is, we do not want to run the 64 ones, but we want to run the 32). (skip-ios + iOS-32 labels)

With the above change, the logic changes too:

1. **RUN** 32b when !IncludeiOS  (skip-ios label)
2. Ignore 32b when !IncludeiOS32 (ios label + skip-ios-32 labels)
3. !IncludeiOS && IncludeiOS32 Add 32b when all the other iOS tests are ignored (that is, we do not want to run the 64 ones, but we want to run the 32).  (skip-ios + iOS-32 labels)

So yes, with the change, the 32b devices will run, AFAIK even with the skip-ios label set.

mandel-macaque: @rolfbjarne the bool check changed a little: https://github.com/xamarin/xamarin-macios/commit/9c29746825e99b2b18d0683a008d441463eb0072 maybe the removal of the ignored value change things and that is why @VincentDondain is not seeing the expected results :/

rolfbjarne: I think the behavior we want is that if either `skip-ios-tests` or `skip-ios-32-tests` is set, then the iOS 32-bit tests are not executed:

| skip-ios-tests | skip-ios-32-tests | Execute 32-bit tests? |
| - | - | - | 
| true | true | **no** |
| true | false | **no** |
| false | true | **no** |
| false | false | **yes** |

However, with this setup it's not possible to run only the 32-bit tests:

* Run all the iOS tests: set no label.
* Run only the 64-bit iOS tests: `skip-ios-32-bit`
* Run only the 32-bit iOS tests:  _not possible_

One solution would be to add a `skip-ios-64-bit` label and a corresponding `IncludeiOS64` variable in the code:

* Run all the iOS tests: set no label.
* Run only the 64-bit iOS tests: `skip-ios-32-bit`
* Run only the 32-bit iOS tests:  `skip-ios-64-bit`

Note: both the iOS and iOS 32-bit tests are enabled by default:

https://github.com/xamarin/xamarin-macios/blob/532059daf6a9a066cf0677b61fc1602d48eebc2f/tests/xharness/Jenkins.cs#L25-L26

VincentDondain: @rolfbjarne I believe your proposed solution to add `IncludeiOS64` is close to what I'm doing (or expecting). The `||` that I updated would remain as is, tracking only `!IncludeiOS32` and we'd make the other tests explicitly `IncludeiOS64` which automatically allows to `skip-ios-64-tests`.

Do we agree that in this scenario you have to do `skip-ios-64-tests` **and** `skip-ios-32-tests` to **skip all iOS tests** right?

Let me know if I misunderstood.

VincentDondain: Ok how about https://github.com/xamarin/xamarin-macios/pull/7493/commits/ffab20c465e7238233fbe44c55a2a09fe41646bc?
Things behave as I expect locally. We'd probably have to update the pipeline for `ios-64` and `ios-32` but that's okay.

`skip-all-tests,run-device-tests` -> everything ignored, except all device tests (didn't work before)
`skip-all-tests,run-ios-32-tests` -> everything ignored (even iOS 64 bit), except 32 bit device tests

rolfbjarne: > Do we agree that in this scenario you have to do `skip-ios-64-tests` **and** `skip-ios-32-tests` to **skip all iOS tests** right?

I was thinking that we continue having `skip-ios-tests` that would skip both 64-bit and 32-bit tests.

VincentDondain:
Superseded by https://github.com/xamarin/xamarin-macios/pull/7540

Issue #7317, opened 2019-10-28
Smart enums vs flags


whitneyschmidt:
There are some cases where enum values we would like to convert to smart enums are used as flags. It would be great to redo these bindings as smart enums.

For example:

In https://github.com/xamarin/xamarin-macios/pull/7316

We add new `enum` values to `AVFoundation.cs` and `AVFoundation\Enums.cs`. This seems like a great use case for smart enums but we can't use a smart enum because it's used as a flag in `AVMetadataObject.cs`:

```
		internal static AVMetadataObjectType ArrayToEnum (NSString[] arr)
		{
			AVMetadataObjectType rv = AVMetadataObjectType.None;

			if (arr == null || arr.Length == 0)
				return rv;

			foreach (var str in arr) {
				rv |= ObjectToEnum (str);
			}

			return rv;
		}
```

spouliot:
So what was done before, e.g. in `src/Vision/VNBarcodeSymbologyExtensions.cs`, is not identical to this case - it was multiple values, but without being a `[Flag]`

spouliot:
Since the **to/from-flag logic** resides in 2 methods, separate from the NSString/enum conversions, I think this (and similar cases) can be moved to smart enums.

IOW we just need to change the **to/from-flag** methods to use the smart enum helpers - and since that's independent of the items then we have the benefits (no manual updates required)

We can also generalize the ** to/from-flag** code to avoid using the enum type - so it can be reused elsewhere (if needed). Still that does not require generator changes IMO.

spouliot:
yes, it's orthogonal (does not stop using smart enums)
PR coming (works but missing tests), but not a generator enhancement - closing

PR #7497, opened 2019-11-26
[avfoundation] Use a smart enum for AVMetadataObjectType

spouliot:
and drop the manual, error prone, conversion code.

The fact that this is also used as a `[Flag]` enum does not mean smart
enum could not be used [1]. The flag code only needed to be updated to
use the smart enum generated code.

[1] https://github.com/xamarin/xamarin-macios/issues/7317

Added unit tests.

### DIFF
--- a/src/AVFoundation/AVMetadataObject.cs
+++ b/src/AVFoundation/AVMetadataObject.cs
@@ -36,7 +36,7 @@ namespace AVFoundation {
 	public partial class AVMetadataObject {
 		public AVMetadataObjectType Type {
 			get {
-				return ObjectToEnum (WeakType);
+				return AVMetadataObjectTypeExtensions.GetValue (WeakType);
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


spouliot:
c.c. @whitneyschmidt @prmallic @mandel-macaque

src/AVFoundation/AVMetadataObject.cs
```diff
@@ -36,7 +36,7 @@ namespace AVFoundation {
 	public partial class AVMetadataObject {
 		public AVMetadataObjectType Type {
 			get {
-				return ObjectToEnum (WeakType);
+				return AVMetadataObjectTypeExtensions.GetValue (WeakType);
```
prmallic: Noob question - does the `AVMetadataObjectTypeExtension` get generated by the generator? I can't seem to find it in my local repo. Also, does it get generated for every `enum` with `[Field]` attributes defined?

spouliot: @prmallic yes, that's what makes it a _smart_ enum - when `[Field]` are added to enum members then the generator will create the conversion methods (NSString->enum and enum->NSString) automagically.

Issue #7481, opened 2019-11-22
FSEventStream bindings are incomplete


Therzok:
### Steps to Reproduce

EDIT: SEE COMMENTS BELOW!

~1. Try to use FSEventStream~
~2. Nothing to use~

### Environment

Xamarin.Mac
Version: 6.14.0.7 (Visual Studio Enterprise)
Hash: 28a560c8
Branch: d16-5
Build date: 2019-11-12 09:11:24-0500

prmallic:
@Therzok . I think the bindings are there in `src/CoreServices/FSEvents.cs`. Is there some specific API that you want to use and is missing?

Therzok:
Hah, could not find it initially.

It seems though that the manual binding could cripple GC performance under VSMac (due to marshalling arrays on every event).

Might be solvable if we let macOS reduce event duplication, but need to check.

From the looks of it, the binding is incomplete. FSEventStreamSet DispatchQueue/ExclusionPaths are missing

Therzok:
Also, the API is not high performance suited - array allocations on every native callback instead of having a callback per file.

PR #7507, opened 2019-11-27
[CoreServices] Completing FSEventStream bindings

prmallic:
Should address https://github.com/xamarin/xamarin-macios/issues/7481


prmallic:
Based the bindings on the other `FSEventStream` ones in the file

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,30 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
...
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern IntPtr FSEventStreamCreateRelativeToDevice (IntPtr allocator, FSEventStreamCallback callback, 
+		IntPtr context, ulong deviceToWatch, IntPtr pathsToWatchRelativeToDevice, 
+		ulong sinceWhen, double latency, FSEventStreamCreateFlags flags);
+
+		public FSEventStream (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice, 
+			ulong sinceWhen, TimeSpan latency, FSEventStreamCreateFlags flags)
+		{
+			if (pathsToWatchRelativeToDevice == null) {
+				throw new ArgumentNullException ("pathsToWatchRelativeToDevice");
```
spouliot: Instead of a string use `name(pathsToWatchRelativeToDevice)` to avoid typos

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,30 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
...
+			}
+
+			eventsCallback = new FSEventStreamCallback (EventsCallback);
+
+			handle =  FSEventStreamCreateRelativeToDevice (
+				allocator ==  null ? IntPtr.Zero : allocator.Handle,
```
spouliot: you can use the extension method `GetHandle ()` here, e.g.
```csharp
allocator.GetHandle (),
```
This will do the null check for you (and it's easier to read)

src/CoreServices/FSEvents.cs
```diff
@@ -307,6 +331,31 @@ public void ScheduleWithRunLoop (NSRunLoop runLoop)
...
 		}
 
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern void FSEventStreamUnscheduleFromRunLoop (IntPtr handle,
+			IntPtr runLoop, IntPtr runLoopMode);
+
+		public void UnscheduleWithRunLoop (CFRunLoop runLoop, NSString runLoopMode)
+		{
+			CheckDisposed ();
+			FSEventStreamUnscheduleFromRunLoop (handle, runLoop.Handle, runLoopMode.Handle);
```
spouliot: you'll get `NRE` if other arguments are `null`

If null is allowed then use the `GetHandle` extension method
if not, then add a null check and throw an `ANE`

src/CoreServices/FSEvents.cs
```diff
@@ -307,6 +331,31 @@ public void ScheduleWithRunLoop (NSRunLoop runLoop)
...
+		}
+
+		public void UnscheduleWithRunLoop (CFRunLoop runLoop)
+		{
+			UnscheduleWithRunLoop (runLoop, CFRunLoop.ModeDefault);
+		}
+
+		public void UnscheduleWithRunLoop (NSRunLoop runLoop, NSString runLoopMode)
+		{
+			UnscheduleWithRunLoop (runLoop.GetCFRunLoop (), runLoopMode);
```
spouliot: same -> NRE

src/CoreServices/FSEvents.cs
```diff
@@ -307,6 +331,31 @@ public void ScheduleWithRunLoop (NSRunLoop runLoop)
...
+		}
+
+		public void UnscheduleWithRunLoop (NSRunLoop runLoop)
+		{
+			UnscheduleWithRunLoop (runLoop.GetCFRunLoop (), CFRunLoop.ModeDefault);
```
spouliot: same -> NRE

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +409,39 @@ public void Invalidate ()
...
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException ("pathsToExclude");
+			}
+			CheckDisposed ();
+			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.Handle);
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetDispatchQueue (IntPtr handle, DispatchQueue queue);
```
spouliot: that's documented as returning `void`
https://developer.apple.com/documentation/coreservices/1444164-fseventstreamsetdispatchqueue?language=objc

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +409,39 @@ public void Invalidate ()
...
+		static extern ulong FSEventStreamGetDeviceBeingWatched (IntPtr handle);
+
+		public ulong GetDevice ()
+		{
+			CheckDisposed ();
+			return FSEventStreamGetDeviceBeingWatched (handle);
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetExclusionPaths (IntPtr handle, IntPtr pathsToExclude);
```
spouliot: new in macOS 10.9 so it should have availability attributes

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,30 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
...
+		public FSEventStream (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice, 
+			ulong sinceWhen, TimeSpan latency, FSEventStreamCreateFlags flags)
+		{
+			if (pathsToWatchRelativeToDevice == null) {
+				throw new ArgumentNullException ("pathsToWatchRelativeToDevice");
+			}
+
+			eventsCallback = new FSEventStreamCallback (EventsCallback);
+
+			handle =  FSEventStreamCreateRelativeToDevice (
```
rolfbjarne: Extremely minor:

```suggestion
			handle = FSEventStreamCreateRelativeToDevice (
```

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,30 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
 		{
 		}
 
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern IntPtr FSEventStreamCreateRelativeToDevice (IntPtr allocator, FSEventStreamCallback callback, 
+		IntPtr context, ulong deviceToWatch, IntPtr pathsToWatchRelativeToDevice, 
+		ulong sinceWhen, double latency, FSEventStreamCreateFlags flags);
+
+		public FSEventStream (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice,
```
rolfbjarne: The `allocator` is optional, so you can provide an overload without it, that just calls this version:

```csharp
public FSEventStream (ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice, ulong sinceWhen, TimeSpan latency, FSEventStreamCreateFlags flags)
    : this (null, deviceToWatch, pathsToWatchRelativeToDevice, sinceWhen, latency, flags)
{
}
```

src/CoreServices/FSEvents.cs
```diff
@@ -307,6 +331,31 @@ public void ScheduleWithRunLoop (NSRunLoop runLoop)
 			ScheduleWithRunLoop (runLoop.GetCFRunLoop (), CFRunLoop.ModeDefault);
 		}
 
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern void FSEventStreamUnscheduleFromRunLoop (IntPtr handle,
+			IntPtr runLoop, IntPtr runLoopMode);
+
+		public void UnscheduleWithRunLoop (CFRunLoop runLoop, NSString runLoopMode)
```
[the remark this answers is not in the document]
Therzok: Good catch!

src/CoreServices/FSEvents.cs
mandel-macaque: 'nameof'

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +409,39 @@ public void Invalidate ()
...
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetExclusionPaths (IntPtr handle, IntPtr pathsToExclude);
+
+		public bool SetExclusionPaths (NSArray pathsToExclude)
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException ("pathsToExclude");
+			}
```
mandel-macaque: Use nameof (pathToExclude).

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +409,39 @@ public void Invalidate ()
...
+			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.Handle);
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetDispatchQueue (IntPtr handle, DispatchQueue queue);
+
+		public bool SetDispatchQueue (DispatchQueue queue)
+		{
+			if (queue == null) {
+				throw new ArgumentNullException ("queue");
```
mandel-macaque: Use nameof (queue).

Therzok:
We also need to do a perf pass to the bindings. Allocating an array for every native callback is going to create a lot of GC garbage and may not be wanted.

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,30 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
...
+
+			eventsCallback = new FSEventStreamCallback (EventsCallback);
+
+			handle =  FSEventStreamCreateRelativeToDevice (
+				allocator ==  null ? IntPtr.Zero : allocator.Handle,
+				eventsCallback, IntPtr.Zero, deviceToWatch, pathsToWatchRelativeToDevice.Handle,
+				sinceWhen, latency.TotalSeconds, flags);
+			
+			if (handle == IntPtr.Zero) {
+				throw new Exception ("Unable to create FSEventStream");
```
mandel-macaque: This exception is too general! Use InvalidOperationException or something more precise. https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=netframework-4.8

prmallic:
Still have to add tests

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +417,40 @@ public void Invalidate ()
...
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException (nameof (pathsToExclude));
+			}
+			CheckDisposed ();
+			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.Handle);
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern void FSEventStreamSetDispatchQueue (IntPtr handle, DispatchQueue queue);
```
mandel-macaque: Hmmm this looks VERY VERY funny. Is a exter C call that takes https://developer.apple.com/documentation/dispatch/dispatch_queue_t?language=objc as a second parameter. I would expect it to be an IntPtr, so you cannot pass the DispatchQueue object, but its handle.

prmallic: I think that mapping is taken care of by the runtime (https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/objective-c-libraries?tabs=macos)

spouliot: @mandel-macaque good catch :)

@prmallic indeed it's documented... but for **ObjC** bindings. What you're doing are p/invokes, i.e. **C** bindings, which is defined by .net (not our own generator)

prmallic:
Tests are incorrect and incomplete right now. Working on it.

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,169 @@
...
+		[Test]
+		public void FSEventFileChangedTest ()
+		{
+			string path = "FSEventStreamTests/TempFileToWatch.txt";
+			Assert.IsTrue (File.Exists (path));;
+
+			File.AppendAllText (path, "Hello World!");
+			//watchEventCalled.WaitOne (Int32.MaxValue);
+			//waitHandle.WaitOne ();
+			Assert.IsTrue (watchedFileChanged);
```
prmallic: @mandel-macaque @spouliot I'm having a bit of trouble getting this working. I'm setting the variable `watchedFileChanged` in the callback here - `fsEventStream.Events += FsEventStream_WatchEvents;`, and checking if its set here, but it doesn't get triggered. maybe i'm missing something obvious here

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		//[Test]
+		public void UnscheduleFromRunLoopTest ()
+		{
+			fsEventStream.UnscheduleFromRunLoop (CFRunLoop.Current);
```
prmallic: Also, i'm trying to understand the implications of `UnscheduleFromRunLoop` here, and how to test it. Would unscheduling from the current loop here mean that the event won't be received?

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		//[Test]
+		public void SetDispatchQueueTest ()
+		{
+			using (var fsEventStreamDispatchQueue = new FSEventStream (st_dev, pathsToWatchRelativeToDevice, ulong.MaxValue, TimeSpan.MinValue, FSEventStreamCreateFlags.None)) {
+				fsEventStreamDispatchQueue.Events += FsEventStream_ExcludeWatchEvents;
+				fsEventStreamDispatchQueue.ScheduleWithRunLoop (CFRunLoop.Main);
+
+				fsEventStreamDispatchQueue.SetDispatchQueue (null); // shouldn't recieve events after this
```
prmallic: I'm not really sure how to test `SetDispatchQueue` apart from setting it to `null` to stop receiving events

tests/apitest/src/CoreServices/FSEventStream.cs
Therzok: That is correct

tests/apitest/src/CoreServices/FSEventStream.cs
rolfbjarne: It's because you need to yield control to the current thread's runloop, so that it can process the FSEventStream events. Luckily we already have a way to do this in our tests:

```csharp
TestRuntime.RunAsync (TimeSpan.FromSeconds (10), async () => { }, () => watchedFileChanged);
```

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,169 @@
...
+			//waitHandle.WaitOne ();
+			Assert.IsTrue (watchedFileChanged);
+		}
+
+		//[Test]
+		public void UnscheduleFromRunLoopTest ()
+		{
+			fsEventStream.UnscheduleFromRunLoop (CFRunLoop.Current);
+			File.AppendAllText ("FSEventStreamTests/TempFileToWatch.txt", "Hello World!");
+			watchEventCalled.WaitOne (Int32.MaxValue);
```
rolfbjarne: The same here: you must call TestRuntime.RunAsync (but with a short timeout, since it will take that long every time).

prmallic:
A lot of thanks to @mandel-macaque for helping in finding the issue with the CFArray release and with the test!

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,44 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
...
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern IntPtr FSEventStreamCreateRelativeToDevice (IntPtr allocator, FSEventStreamCallback callback, 
+			IntPtr context, ulong deviceToWatch, IntPtr pathsToWatchRelativeToDevice, 
+			ulong sinceWhen, double latency, FSEventStreamCreateFlags flags);
+
+		private void FSEventStreamInit (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice, 
+			ulong sinceWhen, TimeSpan latency, FSEventStreamCreateFlags flags)
+		{
+			if (pathsToWatchRelativeToDevice == null) {
+				throw new ArgumentNullException (nameof (pathsToWatchRelativeToDevice));
```
mandel-macaque: What happens if it is empty?

src/CoreServices/FSEvents.cs
```diff
@@ -207,8 +245,8 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
 			}
 
 			var events = new FSEvent[numEvents];
-			var pathArray = new CFArray (eventPaths);
-
+			using var pathArray = new CFArray (eventPaths);
```
mandel-macaque: To clarify to other reviewers, if the pathArray is not release after the call, we had issues with the array since it is an Array the follows the Get rule.

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +417,49 @@ public void Invalidate ()
...
+
+		[Mac (10, 9)]
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetExclusionPaths (IntPtr handle, IntPtr pathsToExclude);
+
+		public bool SetExclusionPaths (NSArray pathsToExclude)
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException (nameof (pathsToExclude));
+			}
```
mandel-macaque: What happens if it is empty ?

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +417,49 @@ public void Invalidate ()
...
+			CheckDisposed ();
+			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.GetHandle());
+		}
+
+		public bool SetExclusionPaths (string [] pathsToExclude)
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException (nameof (pathsToExclude));
+			}
```
mandel-macaque: Same, is an empty array valid?

tests/apitest/apitest.csproj
```diff
@@ -31,6 +31,7 @@
     <IncludeMonoRuntime>false</IncludeMonoRuntime>
     <UseSGen>false</UseSGen>
     <PlatformTarget>x86</PlatformTarget>
+    <AOTMode>None</AOTMode>
```
mandel-macaque: Your IDE added something new.

tests/apitest/apitest.csproj
```diff
@@ -47,6 +48,7 @@
     <UseSGen>false</UseSGen>
     <LinkMode>SdkOnly</LinkMode>
     <PlatformTarget>x86</PlatformTarget>
+    <AOTMode>None</AOTMode>
```
mandel-macaque: Same, probably added by the IDE.

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,177 @@
...
+			dir = Directory.CreateDirectory (dirPath);
+
+			using (FileStream fileToWatch = File.Create (Path.Combine (dirPath, "TempFileToWatch.txt")))
+			using (FileStream fileToExclude = File.Create (Path.Combine (dirPath, "TempFileToExclude.txt"))) { }
+		}
+
+		[TestFixtureTearDown]
+		public void Destroy ()
+		{
+			Directory.Delete ("FSEventStreamTests", true);
```
mandel-macaque: Looks like we are leaving stuff behind. Should it be dirPath?

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,177 @@
...
+			exludedFileChanged = false;
+
+			fsEventStream = new FSEventStream (st_dev, pathsToWatchRelativeToDevice, FSEvent.SinceNowId, TimeSpan.FromSeconds (1), FSEventStreamCreateFlags.WatchRoot | FSEventStreamCreateFlags.FileEvents | FSEventStreamCreateFlags.NoDefer);
+			Assert.IsNotNull (fsEventStream, "Null");
+		}
+
+		[TearDown]
+		public void EndFSEventStream ()
+		{
+			fsEventStream.Stop ();
```
mandel-macaque: Does this work if the test does not start the stream? We do not start it in the SetUp AFAIK. Could, if we ever add a test that does not start it, be an issue. Just double check.

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,177 @@
...
+			fsEventStream.Invalidate ();
+			fsEventStream.Dispose ();
+		}
+
+		[Test]
+		public void GetDeviceBeingWatchedTest ()
+			=> Assert.AreEqual (st_dev, fsEventStream.GetDevice (), "Device ID");
```
mandel-macaque: I probably be replying to myself, but if this test does not crash, ignore the previous comment.

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,177 @@
...
+		[Test]
+		public void SetExclusionPathsTest ()
+		{
+			var watchDirPath = Path.Combine (dirPath, "WatchDir");
+			var watchDir = Directory.CreateDirectory (watchDirPath);
+			var unWatchDirPath = Path.Combine (dirPath, "UnwatchDir");
+			var unWatchDir = Directory.CreateDirectory (unWatchDirPath);
+
+			// Excluding the unWatchDirPath from the watcher so any event inside it doesn't get triggered
+			Assert.IsTrue (fsEventStream.SetExclusionPaths (new [] { Path.Combine ("private", unWatchDirPath.Substring (1)) }), "SetExclusionPaths failed");
```
mandel-macaque: That private looks funny, ask the stream what is the root in order to create the path.

tests/apitest/src/CoreServices/FSEventStream.cs
```diff
@@ -0,0 +1,177 @@
...
+
+			using (FileStream fileToWatch = File.Create (Path.Combine (watchDirPath, "TempFileToWatch.txt")))
+			using (FileStream fileToExclude = File.Create (Path.Combine (unWatchDirPath, "TempFileToExclude.txt"))) {
+				fileToWatch.Close ();
+				fileToExclude.Close ();
```
mandel-macaque: Funny pattern. Create an close it, never use the stream again, but we have a using block increasing indentation. we can unindent the bellow code right? It does not look to be using the create streams.

src/CoreServices/FSEvents.cs
```diff
@@ -154,9 +154,11 @@ public void Dispose ()
 
 		protected virtual void Dispose (bool disposing)
 		{
-			if (handle != IntPtr.Zero) {
-				FSEventStreamRelease (handle);
-				handle = IntPtr.Zero;
+			if (disposing) {
```
Therzok: Why only when disposing? We'll never release the unmanaged handle otherwise?

mandel-macaque:
Timeouts are related to the 32b dialog in the bot. Should be fixed.

mandel-macaque:
build

src/CoreServices/FSEvents.cs
prmallic: This follows the disposing pattern - https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

prmallic:
Known failure - https://github.com/xamarin/maccore/issues/581

src/CoreServices/FSEvents.cs
Therzok: Yes, but in this case, an IntPtr does not have a finalizer.

[SafeHandle-derived](https://source.dot.net/#System.Private.CoreLib/SafeHandle.cs,69) classes have a finalizer run by default, so what that means is:
a) your class is finalized, then safehandle is finalized
b) your class is disposed, then your safehandle is disposed

With the current code, it's just a plain IntPtr. What it means for the finalizer case is that when `FSEventStream` is finalized, _nothing happens_. Which means that the native handle is never released, so every FSEventStream instance that is not explicitly disposed _will leak_.

Therzok: See the pattern which doesn't use a safe-handle, specifically:
```csharp
   // Protected implementation of Dispose pattern.
   protected virtual void Dispose(bool disposing)
   {
      if (disposed)
         return; 
      
      if (disposing) {
         // Free any other managed objects here.
         //
      }
      
      // Free any unmanaged objects here.
      // <---- FSEventStreamRelease here
      disposed = true;
   }

```

src/CoreServices/FSEvents.cs
```diff
@@ -199,6 +199,43 @@ public FSEventStream (string [] pathsToWatch, TimeSpan latency, FSEventStreamCre
 		{
 		}
 
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern IntPtr FSEventStreamCreateRelativeToDevice (IntPtr allocator, FSEventStreamCallback callback, 
+			IntPtr context, ulong deviceToWatch, IntPtr pathsToWatchRelativeToDevice, 
+			ulong sinceWhen, double latency, FSEventStreamCreateFlags flags);
+
+		private void FSEventStreamInit (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice,
```
rolfbjarne: Style: we usually don't use `private`, it's redundant.

```suggestion
	void FSEventStreamInit (CFAllocator allocator, ulong deviceToWatch, NSArray pathsToWatchRelativeToDevice, 
```

src/CoreServices/FSEvents.cs
```diff
@@ -307,16 +344,35 @@ public void ScheduleWithRunLoop (NSRunLoop runLoop)
...
+		public void UnscheduleFromRunLoop (NSRunLoop runLoop) => UnscheduleFromRunLoop (runLoop.GetCFRunLoop (), CFRunLoop.ModeDefault);
+
 		[DllImport (Constants.CoreServicesLibrary)]
 		static extern IntPtr FSEventStreamCopyPathsBeingWatched (IntPtr handle);
 
 		public string [] PathsBeingWatched {
 			get {
 				CheckDisposed ();
-				var cfarray = new CFArray (FSEventStreamCopyPathsBeingWatched (handle), true);
+				using var cfarray = new CFArray (FSEventStreamCopyPathsBeingWatched (handle));
```
rolfbjarne: This looks wrong, according to the `Copy` rule you own the return value from `FSEventStreamCopyPathsBeingWatched `, I think it should be like this:

```suggestion
				using var cfarray = new CFArray (FSEventStreamCopyPathsBeingWatched (handle), true);
```

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +416,47 @@ public void Invalidate ()
...
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern bool FSEventStreamSetExclusionPaths (IntPtr handle, IntPtr pathsToExclude);
+
+		public bool SetExclusionPaths (NSArray pathsToExclude)
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException (nameof (pathsToExclude));
+			}
+			CheckDisposed ();
+			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.GetHandle());
```
rolfbjarne:
```suggestion
			return FSEventStreamSetExclusionPaths (handle, pathsToExclude.GetHandle ());
```

src/CoreServices/FSEvents.cs
```diff
@@ -360,6 +416,47 @@ public void Invalidate ()
...
+		}
+
+		public bool SetExclusionPaths (string [] pathsToExclude)
+		{
+			if (pathsToExclude == null) {
+				throw new ArgumentNullException (nameof (pathsToExclude));
+			}
+			CheckDisposed ();
+			using NSArray nsPathsToExclude = NSArray.FromStrings (pathsToExclude);
+			return FSEventStreamSetExclusionPaths (handle, nsPathsToExclude.GetHandle());
```
rolfbjarne:
```suggestion
			return FSEventStreamSetExclusionPaths (handle, nsPathsToExclude.GetHandle ());
```

spouliot:
Tests fails on 10.15
https://jenkins.internalx.com/blue/organizations/jenkins/macios/detail/master/1107/pipeline/223/

Issue #7513, opened 2019-11-28
[xcode11.3] Test failure: Xamarin.ApiTest.ApiTest.AlwaysOptimizable: All methods calling optimizable API must be optimizable


rolfbjarne:
Test failures:

* Xamarin.ApiTest.ApiTest.AlwaysOptimizable(macOSMobile): All methods calling optimizable API must be optimizable

    > The method Network.NWParameters Network.NWParameters::CreateCustomIP(System.Byte,System.Action`1<Network.NWProtocolOptions>) calls System.Void ObjCRuntime.BlockLiteral::SetupBlockUnsafe(System.Delegate,System.Delegate), but it does not have a [BindingImpl (BindingImplOptions.Optimizable)] attribute.

* Xamarin.ApiTest.ApiTest.AlwaysOptimizable(macOSFull): All methods calling optimizable API must be optimizable
    
    > The method Network.NWParameters Network.NWParameters::CreateCustomIP(System.Byte,System.Action`1<Network.NWProtocolOptions>) calls System.Void ObjCRuntime.BlockLiteral::SetupBlockUnsafe(System.Delegate,System.Delegate), but it does not have a [BindingImpl (BindingImplOptions.Optimizable)] attribute.

See http://xamarin-storage/jenkins/xamarin-macios/xcode11.3/bcf3899893c0ee1a0bbcd814164a08423f0d543a/21/jenkins-results/tests/index.html for https://github.com/xamarin/xamarin-macios/commit/bcf3899893c0ee1a0bbcd814164a08423f0d543a.

mandel-macaque:
Hmmm expected this to be closed after the PR was merged.

PR #7530, opened 2019-12-02
[Network] Add missing BindingImplAttr.

mandel-macaque:
Fixes: https://github.com/xamarin/xamarin-macios/issues/7513

### DIFF
--- a/src/Network/NWParameters.cs
+++ b/src/Network/NWParameters.cs
@@ -183,6 +183,7 @@ namespace Network {
 		unsafe static extern IntPtr nw_parameters_create_custom_ip (byte custom_ip_protocol_number, BlockLiteral *configure_ip);
 
 		[NoWatch, NoTV, NoiOS, Mac (10,15)]
+		[BindingImpl (BindingImplOptions.Optimizable)]
 		public unsafe static NWParameters CreateCustomIP (byte protocolNumber, Action<NWProtocolOptions> configureCustomIP = null)
 		{
 			var ipHandler = new BlockLiteral ();
--- a/tests/monotouch-test/Network/NWParametersTest.cs
+++ b/tests/monotouch-test/Network/NWParametersTest.cs
@@ -199,6 +199,21 @@ namespace MonoTouchFixtures.Network {
 			}
 		}
 
+#if MONOMAC
+		[Test]
+		public void CreateCustomIP ()
+		{
+			TestRuntime.AssertXcodeVersion (11, 0);
+			byte ipVersion = 10;
+			var setUpProtocol = CreateConfigureProtocolHandler ();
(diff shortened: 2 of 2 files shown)


src/Network/NWParameters.cs
```diff
@@ -183,6 +183,7 @@ public unsafe static NWParameters CreateUdp (Action<NWProtocolOptions> configure
 		unsafe static extern IntPtr nw_parameters_create_custom_ip (byte custom_ip_protocol_number, BlockLiteral *configure_ip);
 
 		[NoWatch, NoTV, NoiOS, Mac (10,15)]
+		[BindingImpl (BindingImplOptions.Optimizable)]
 		public unsafe static NWParameters CreateCustomIP (byte protocolNumber, Action<NWProtocolOptions> configureCustomIP = null)
```
rolfbjarne: Is there a test for this method?

spouliot: it might not need to be a _complete_ test, i.e. the linker can only try to optimize it if it's used (otherwise it's just removed).

rolfbjarne: My thoughts were that the implementation of this method is very different from most other optimizable methods, so it would be nice to have a test to make sure it works even after optimization.

spouliot: make sense, even if its been proven harder to get good network test :)

mandel-macaque: +1 BUT I prefer to land that in a diff PR, you need to give me time to understand what https://developer.apple.com/documentation/network/3141919-nw_parameters_create_custom_ip?language=objc means with a 'a custom IP protocol.'

tests/monotouch-test/Network/NWParametersTest.cs
```diff
@@ -199,6 +199,21 @@ public void CreateSecureTcpTestDoNotSetUpTls ()
...
 
+#if MONOMAC
+		[Test]
+		public void CreateCustomIP ()
+		{
+			TestRuntime.AssertXcodeVersion (11, 0);
+			byte ipVersion = 10;
+			var setUpProtocol = CreateConfigureProtocolHandler ();
+			using (var parameters = NWParameters.CreateCustomIP (ipVersion, setUpProtocol))
+			using (var endpoint = NWEndpoint.Create ("wwww.google.com", "80")) {
```
rolfbjarne: Are the 4 `w`'s intentional?

spouliot: `:wq!` 🤣

mandel-macaque: the api does not care, will be need to change once we merge with master to use the respires.

PR #7531, opened 2019-12-02
[Localization] Add generator error messages

wachaudh:
I will be making multiple PRs for localization changes, hopefully one per tool. This PR is copying what android does for localization for more context see: https://github.com/xamarin/xamarin-android/pull/3900/ and https://github.com/xamarin/monodroid/wiki/Localization


src/Resources.Designer.cs
```diff
@@ -0,0 +1,638 @@
+﻿//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//     Runtime Version:4.0.30319.42000
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
```
spouliot: how is this generated ?
do we need to add it inside the repo ? (or can the build generate it on the fly)
how to we _property_ add new error codes in it ? (should be added to the README.md)

src/Resources.resx
```diff
@@ -0,0 +1,890 @@
...
+
+	<data name="BI0068" xml:space="preserve">
+		<value>Invalid value for target framework: {0}.</value>
+		<comment>
+		</comment>
+	</data>
+
+	<data name="BI0070" xml:space="preserve">
+		<value>
+				Invalid target framework: {0}. Valid target frameworks are: {1}.
```
spouliot: are the extra spaces inside the `<value/>` significant ?

src/generator.cs
```diff
@@ -378,7 +378,7 @@ public GeneratedType (Type t, GeneratedTypes root)
 			// protected against a StackOverflowException - bug #19751
 			// it does not protect against large cycles (but good against copy/paste errors)
 			if (Parent == Type)
-				throw new BindingException (1030, true, "{0} cannot have [BaseType(typeof({1}))] as it creates a circular dependency", Type, Parent);
+				throw new BindingException (1030, true, bgen.Resources.BI1030, Type, Parent);
```
spouliot: I wonder if we could look them up using reflection (based on the error code) ?

wachaudh:
> Not sure how the whole "fix a typo" process would work ?
> I suspect (hope) there's not need to update all `.xlf` files but, beside that, I would not know how to proceed.

It works by editing `Resources.resx` in visual studio. The xlf and and designeer.cs will be updated automatically. There is documentation in the android pr ( https://github.com/xamarin/xamarin-android/pull/3900 ) that does a better job of explaining the process  ). I will port the documentation eventually.

src/Resources.Designer.cs
wachaudh: It's generated by visual studio when there are updates to `Resources.resx`.  For now, I am including it in the repo because that's what the android folks are doing however I will investigate it further.

src/generator.cs
mandel-macaque: I think we definitely could like:
```csharp
internal static string GetErrorMessage (errorCode) {
// reflection stuff yo find the property
}
```
and just change the binding exception to something even lazier for us to write:
```csharp
public BindingException (int error, bool, params object[] content) : this (error, bool, GetErrorMessage (error), content);
```

And just call the new constructor with the error code, that calls the reflection and returns the correct string. then we move to:

```csharp
throw new BindingException (1030, true, Type, Parent); // already has the translated message.
```

src/Resources.resx
wachaudh: No I don't think so. I will remove them.

src/generator.cs
spouliot: yup
```
throw new BindingException (1030, true, bgen.Resources.BI1030 ...
```
is just a typo magnet :)

src/Resources.resx
wachaudh: Not going to remove the spaces because they look fine in vsmac. 

Screen Shot 2019-12-03 at 10 45 56 AM

src/generator.cs
brendanzagaeski: Oo cool idea. I might borrow that for Xamarin.Android too. One little complication I can think of for the Xamarin.Android case is that sometimes we have multiple separate strings for a single message code. At the moment, the approach we have is to add suffixes on the message code property name like `XA0000_Files` and `XA0000_Directories`. There might still be a clean way to use reflection for a scenario like that. I'll have to muse on it a bit.

brendanzagaeski: Oh, one way to handle the multiple strings scenario would be to allow an optional additional suffix argument in the new override method. That could work nicely. Thanks for the inspiration on this!

src/Resources.resx
spouliot: as long as they don't affect the output (on terminal or the IDE) then it's fine by me :)

src/Resources.Designer.cs
```diff
@@ -0,0 +1,638 @@
...
+                return resourceCulture;
+            }
+            set {
+                resourceCulture = value;
+            }
+        }
+        
+        internal static string BI0000 {
+            get {
+                return ResourceManager.GetString("BI0000", resourceCulture);
```
stephen-hawley: Hand hint: you can use `nameof ()` instead of a string for the symbol (easier refactoring), combine that with the short-hand for a get-only property and every property becomes a one-liner:
```
internal static string BI0000 => ResourceManager.GetString (nameof (BI0000), resourceCulture);
```

spouliot: That file is (.net, not our) generated code https://github.com/xamarin/xamarin-macios/pull/7531#discussion_r352859870
I'm pretty sure the generator is older than the syntax you're suggesting :)

src/Resources.resx
```diff
@@ -129,13 +129,15 @@
...
+		<value> Could not compile the API bindings. 
+		{0}
+		</value>
 		<comment>
 		</comment>
 	</data>
 
 	<data name="BI0026" xml:space="preserve">
-		<value>Could not parse the command line argument '{0}': {1}</value>
+		<value>Could not parse the command line argument '--warnaserror': {0}</value>
```
spouliot: not sure it's a good change
it makes it too easy for someone to translate `--warnaserror` by mistake
and the same error message could be used for something else (later) and would require updating translations

src/btouch.cs
```diff
@@ -103,7 +103,7 @@ string GetAttributeLibraryPath ()
 			} else if (target_framework == TargetFramework.Xamarin_Mac_2_0_Mobile) {
 				return Path.Combine (GetSDKRoot (), "lib", "bgen", "Xamarin.Mac-mobile.BindingAttributes.dll");
 			} else {
-				throw ErrorHelper.CreateError (1043, "Internal error: unknown target framework '{0}'.", target_framework);
+				throw ErrorHelper.CreateError (1053, target_framework);
```
spouliot: why has the number changed ?
`git grep 1053 src` returns me nothing

src/btouch.cs
```diff
@@ -132,7 +132,7 @@ IEnumerable<string> GetLibraryDirectories ()
 			} else if (target_framework == TargetFramework.Xamarin_Mac_2_0_Mobile) {
 				yield return Path.Combine (GetSDKRoot (), "lib", "mono", "Xamarin.Mac");
 			} else {
-				throw ErrorHelper.CreateError (1043, "Internal error: unknown target framework '{0}'.", target_framework);
+				throw ErrorHelper.CreateError (1053, target_framework);
```
spouliot: same

src/error.cs
```diff
@@ -11,51 +11,63 @@
...
-		this (code, error, null, message, args)
+	public BindingException (int code, bool error, params object[] args) : 
+		this (code, error, null, args)
 	{
 	}
 
-	public BindingException (int code, bool error, Exception innerException, string message, params object[] args) : 
-		base (String.Format (message, args), innerException)
+	public BindingException (int code, bool error, Exception innerException, params object[] args) : 
+		base (String.Format (GetMessage(code), args), innerException)
```
spouliot: minor/style: missing space before `(`

src/error.cs
```diff
@@ -11,51 +11,63 @@
...
+	static string GetMessage (int code)
+	{
+		Type resourceType = Type.GetType ("bgen.Resources");
+		string errorCode = string.Format ("BI{0:0000}", code);
+		PropertyInfo prop = resourceType.GetProperty (errorCode, BindingFlags.NonPublic |
+		    BindingFlags.Static |
+		    BindingFlags.GetProperty);
+		var errorMessage = (String) prop.GetValue (null);
+		return errorMessage;
+	}
```
spouliot: That needs some null checks - e.g. if there's a typo and the message does not exists then the customer will get an error that's even harder to debug (for them and us).

e.g. if `prop` is null you'll end with ` BI0000` because of an NRE
and not with the original error code (without a description)

Maybe we should have a default message to ask for an issue to be filled when this happens, e.g.

```BIxxxx: Sorry the error message for this code (xxxx) could not be found. Please report this missing message using GitHub...`

File: src/btouch.cs (hunk 6 above)
Comment (wachaudh) [reply src/btouch.cs]: According to `generator-errors.md ` the error message matches 1053, not 1043. I'm assuming whoever put the error code as 1043 in btouch.cs made a mistake.
Screen Shot 2019-12-04 at 1 49 14 PM

File: src/Resources.resx (hunk 5 above)
Comment (wachaudh) [reply src/Resources.resx]: Okay I can change it back. I changed it because that's the message that was being used in btouch.cs, which is different from how it's documented in generator-errors.md.

Screen Shot 2019-12-04 at 1 54 26 PM

**Comment by wachaudh** [comment]
build

File: src/error.cs [hunk 10]
```diff
@@ -11,51 +11,65 @@
...
 using ProductException=BindingException;
 
 // Error allocation: the errors are listed (and documented) in $(TOP)/docs/website/generator-errors.md
 
 public class BindingException : Exception {
 	
-	public BindingException (int code, bool error, string message) :
-		base (message)
+	public BindingException (int code, bool error) :
+		base (GetMessage(code))
```
rolfbjarne:
```suggestion
		base (GetMessage (code))
```

src/generator.cs
```diff
@@ -5191,9 +5182,10 @@ void PrintMethodAttributes (MemberInformation minfo)
 			PrintPlatformAttributes (minfo.method.DeclaringType);
 		}
 
-		// in theory we could check for `minfo.is_ctor` but some manual bindings are using methods for `init*`
-		if (AttributeManager.HasAttribute<DesignatedInitializerAttribute> (mi))
+		foreach (var di in AttributeManager.GetCustomAttributes<DesignatedInitializerAttribute> (mi)) {
```
rolfbjarne: Is this change where you change the check from a single attribute to multiple attributes intentional? IIRC @spouliot did the reverse not long ago.

src/generator-attribute-manager.cs
```diff
@@ -198,7 +198,7 @@ static IEnumerable<System.Attribute> ConvertOldAttributes (CustomAttributeData a
 					arr [a] = (Type) typed_values [a].Value;
 				value = arr;
 			} else if (arg.TypedValue.ArgumentType.IsArray) {
-				throw ErrorHelper.CreateError (1059, "Internal error: failed to instantiate mock attribute '{0}' (unknown type for the named argument #{1} ({2}). Please file a bug report (https://github.com/xamarin/xamarin-macios/issues/new) with a test case.", attribType.FullName, i + 1, arg.MemberName);
+				throw ErrorHelper.CreateError (1073, attribType.FullName, i + 1, arg.MemberName);
```
whitneyschmidt: Could you explain why the error # changing from `1059` to `1073` here? It looks like error # is retained ^ :)

src/generator.cs
wachaudh: No, it wasn't intentional. It happened because I was copying stuff from an older file. I thought I had fixed it. Should be fixed now.

src/generator-attribute-manager.cs
wachaudh: Code 1059 is used for a different error message in other places. The error message for 1059 is : `Found * * attributes on the member *. At most one was expected.`

src/generator-filters.cs
```diff
@@ -191,7 +191,7 @@ void GenerateProperties (Type type)
 			// we can skip the name when it's identical to a protocol selector
 			if (name == null) {
 				if (export == null)
-					throw new BindingException (1072, true, $"Missing [CoreImageFilterProperty] attribute on {0} property {1}", type.Name, p.Name);
+					throw new BindingException (1074, true, type.Name, p.Name);
```
mandel-macaque: Same comments as the ones from @spouliot why do we change here the number?

src/generator-filters.cs
```diff
@@ -288,7 +288,7 @@ void GenerateFilterGetter (string propertyType, string propertyName)
 			print ("return NSArray.ArrayFromHandle<CIVector> (handle);");
 			break;
 		default:
-			throw new BindingException (1018, true, "Unimplemented CoreImage property type {0}", propertyType);
+			throw new BindingException (1075, true, propertyType);
```
mandel-macaque: Same, diff error number.

src/generator-filters.cs
```diff
@@ -357,7 +357,7 @@ void GenerateFilterSetter (string propertyType, string propertyName)
 			print ("}");
 			break;
 		default:
-			throw new BindingException (1018, true, "Unimplemented CoreImage property type {0}", propertyType);
+			throw new BindingException (1075, true, propertyType);
```
mandel-macaque: Same diff error number.

src/generator.cs
```diff
@@ -1358,7 +1357,7 @@ string GetToBindAsWrapper (MethodInfo mi, MemberInformation minfo = null, Parame
 		}
 
 		if (originalType.IsByRef)
-			throw new BindingException (1048, true, $"Unsupported type 'ref/out {originalType.Name.Replace ("&", string.Empty)}' decorated with [BindAs]");
+			throw new BindingException (1080, true, originalType.Name.Replace ("&", string.Empty));
```
mandel-macaque: Same diff error number.

src/generator.cs
```diff
@@ -1978,7 +1975,7 @@ string MakeSig (string send, bool stret, MethodInfo mi, bool aligned, EnumMode e
 		try {
 			sb.Append (ParameterGetMarshalType (new MarshalInfo (this, mi) { IsAligned = aligned, EnumMode = enum_mode } ));
 		} catch (BindingException ex) {
-			throw new BindingException (ex.Code, ex.Error, ex,  "{0} in method `{1}'", ex.Message, mi.Name);
+			throw new BindingException (1078, ex.Error, ex, ex.Message, mi.Name);
```
mandel-macaque: Why do we go from ex.Code to 1078? Is ex.Code always the same, if it is not, we are going to get the wrong message.

src/generator.cs
```diff
@@ -1993,7 +1990,7 @@ string MakeSig (string send, bool stret, MethodInfo mi, bool aligned, EnumMode e
 			try {
 				sb.Append (ParameterGetMarshalType (new MarshalInfo (this, mi, pi) { EnumMode = enum_mode }).Replace (' ', '_'));
 			} catch (BindingException ex) {
-				throw new BindingException (ex.Code, ex.Error, ex, "{0} in parameter `{1}' from {2}.{3}", ex.Message, pi.Name.GetSafeParamName (), mi.DeclaringType, mi.Name);
+				throw new BindingException (1079, ex.Error, ex, ex.Message, pi.Name.GetSafeParamName (), mi.DeclaringType, mi.Name);
```
mandel-macaque: Same

src/generator.cs
```diff
@@ -2026,7 +2023,7 @@ void RegisterMethod (bool need_stret, MethodInfo mi, string method_name, bool al
 			try {
 				b.Append (ParameterGetMarshalType (new MarshalInfo (this, mi, pi) { EnumMode = enum_mode }, true));
 			} catch (BindingException ex) {
-				throw new BindingException (ex.Code, ex.Error, ex, "{0} in parameter {1} of {2}.{3}", ex.Message, pi.Name.GetSafeParamName (), mi.DeclaringType, mi.Name);
+				throw new BindingException (1079, ex.Error, ex, ex.Message, pi.Name.GetSafeParamName (), mi.DeclaringType, mi.Name);
```
mandel-macaque: Same

src/generator.cs
```diff
@@ -2829,9 +2826,7 @@ void GenerateStrongDictionaryTypes ()
 							getter = "{1} (ulong?) GetNUIntValue ({0})";
 							setter = "SetNumberValue ({0}, {1}value)";
 						} else {
-							throw new BindingException (1031, true,
-										    "Limitation: can not automatically create strongly typed dictionary for " +
-										    "({0}) the value type of the {1}.{2} property", pi.PropertyType, dictType, pi.Name);
+							throw new BindingException (1033, true, pi.PropertyType, dictType, pi.Name);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -2883,9 +2876,7 @@ void GenerateStrongDictionaryTypes ()
 							getter = "GetNativeValue<" + pi.PropertyType +"> ({0})";
 							setter = "SetNativeValue ({0}, value)";
 						} else {
-							throw new BindingException (1031, true,
-										    "Limitation: can not automatically create strongly typed dictionary for " +
-										    "({0}) the type of the {1}.{2} property", pi.PropertyType, dictType, pi.Name);
+							throw new BindingException (1033, true, pi.PropertyType, dictType, pi.Name);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -4112,7 +4103,7 @@ void GenerateTypeLowering (MethodInfo mi, bool null_allowed_override, EnumMode e
 					} else if (isNSObject || isINativeObjectSubclass) {
 						by_ref_init.AppendFormat ("{0} == null ? IntPtr.Zero : {0}.Handle;\n", pi.Name.GetSafeParamName ());
 					} else {
-						throw ErrorHelper.CreateError (99, $"Internal error: don't know how to create ref/out (input) code for {mai.Type} in {mi}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+						throw ErrorHelper.CreateError (88, mai.Type, mi);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -4129,7 +4120,7 @@ void GenerateTypeLowering (MethodInfo mi, bool null_allowed_override, EnumMode e
 					} else if (isArrayOfString) {
 						by_ref_processing.AppendFormat ("{0} = NSArray.StringArrayFromHandle ({0}Value);\n", pi.Name.GetSafeParamName ());
 					} else {
-						throw ErrorHelper.CreateError (99, $"Internal error: don't know how to create ref/out code for array {mai.Type} in {mi}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+						throw ErrorHelper.CreateError (88, mai.Type, mi);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -4141,7 +4132,7 @@ void GenerateTypeLowering (MethodInfo mi, bool null_allowed_override, EnumMode e
 						by_ref_processing.AppendFormat ("if ({0}Value != ({0} == null ? IntPtr.Zero : {0}.Handle))\n\t", pi.Name.GetSafeParamName ());
 					by_ref_processing.AppendFormat ("{0} = Runtime.GetINativeObject<{1}> ({0}Value, {2}, {3});\n", pi.Name.GetSafeParamName (), RenderType (elementType), isForcedType ? "true" : "false", isForcedType ? isForcedOwns : "false");
 				} else {
-					throw ErrorHelper.CreateError (99, $"Internal error: don't know how to create ref/out (output) code for {mai.Type} in {mi}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+					throw ErrorHelper.CreateError (88, mai.Type, mi);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -5012,7 +5003,7 @@ string GetAsyncTaskType (AsyncMethodInfo minfo)
 			return FormatType (minfo.type, attr.ResultType);
 
 		Console.WriteLine ("{0}", minfo.MethodInfo.GetParameters ().Last ().ParameterType);
-		throw new BindingException (1023, true, "Async method {0} with more than one result parameter in the callback by neither ResultTypeName or ResultType", minfo.mi);
+		throw new BindingException (1077, true, minfo.mi);
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -5887,24 +5878,24 @@ void GenerateProtocolTypes (Type type, string class_visibility, string TypeName,
...
+					exceptions.Add (ErrorHelper.CreateError (1067, type.FullName, gr.Key, @readonly.DeclaringType.FullName, writeonly.DeclaringType.FullName,
+						@readonly, writeonly));
 					continue;
 				} else if (@readonly != null) {
 					bestMatch = @readonly;
 				} else if (writeonly != null) {
 					bestMatch = writeonly;
 				} else {
-					exceptions.Add (ErrorHelper.CreateError (99, $"Internal error: property {properties [0]} doesn't have neither a getter nor a setter."));
+					exceptions.Add (ErrorHelper.CreateError (89, properties [0]));
```
mandel-macaque: Diff error number.

src/generator.cs
```diff
@@ -7478,7 +7467,7 @@ string FindSelector (Type type, MethodInfo mi)
 			currentType = bta.BaseType;
 		}
 		while (currentType != null);
-		throw new BindingException (1035, true, "Unable to find selector for {0} on {1} on self or base class", mi, type);
+		throw new BindingException (1076, true, mi, type);
```
mandel-macaque: Diff error number.

tests/generator/ErrorTests.cs
```diff
@@ -421,7 +421,7 @@ interface FooObject {
 	}
 }");
 			bgen.AssertExecuteError ("build");
-			bgen.AssertError (1048, "Unsupported type 'ref/out NSValue' decorated with [BindAs]");
+			bgen.AssertError (1080, "Unsupported type 'ref/out NSValue' decorated with [BindAs]");
```
mandel-macaque: Diff error number.

tests/generator/ErrorTests.cs
```diff
@@ -445,7 +445,7 @@ interface FooObject {
...
```
mandel-macaque: Same.

src/generator-filters.cs
wachaudh: The error code 1018 is being used with a different error message in other parts of the generator.  The message for 1018 should be: `No [Export] attribute on property *.*`.  I also changed the error codes for other existing exceptions  because they were using the same error code for different messages.

src/generator.cs
```diff
@@ -5872,9 +5863,9 @@ void GenerateProtocolTypes (Type type, string class_visibility, string TypeName,
 					// Check that all the duplicates use the same selector, and if not show a warning.
 					exportAttributes [i] = GetExportAttribute (properties [i]);
 					if (i > 0 && exportAttributes [i].Selector != exportAttributes [0].Selector) {
-						ErrorHelper.Warning (1068, $"The type '{type.FullName}' is trying to inline the property '{gr.Key}' from the protocols '{properties [0].DeclaringType.FullName}' and '{properties [i].DeclaringType.FullName}'," +
-							$" and the inlined properties use different selectors ({properties [0].DeclaringType.Name}.{properties [0].Name} uses '{exportAttributes [0].Selector}', and {properties [i].DeclaringType.Name}.{properties [i].Name} uses '{exportAttributes [i].Selector}'.");
-					}
+						ErrorHelper.Warning (1068, type.FullName, gr.Key, properties [0].DeclaringType.FullName, properties [i].DeclaringType.FullName,
+							properties [0].DeclaringType.Name, properties [0].Name, exportAttributes [0].Selector, properties [i].DeclaringType.Name, properties [i].Name, exportAttributes [i].Selector);
+				}
```
spouliot: minor: ^ incorrect alignment

src/generator.cs
```diff
@@ -5841,8 +5832,8 @@ void GenerateProtocolTypes (Type type, string class_visibility, string TypeName,
 			foreach (var gr in duplicateMethodsGroupedBySelector) {
 				var distinctMethodsBySignature = gr.GroupBy ((v) => v.Signature).Select ((v) => v.First ()).ToArray ();
 				if (distinctMethodsBySignature.Length > 1) {
-					exceptions.Add (ErrorHelper.CreateError (1069, $"The type '{type.FullName}' is trying to inline the methods binding the selector '{gr.Key}' from the protocols '{distinctMethodsBySignature [0].Method.DeclaringType.FullName}' and '{distinctMethodsBySignature [1].Method.DeclaringType.FullName}'," +
-$" using methods with different signatures ('{distinctMethodsBySignature [0].Method.ToString ()}' vs '{distinctMethodsBySignature [1].Method.ToString ()}')."));
+					exceptions.Add (ErrorHelper.CreateError (1069, type.FullName, gr.Key, distinctMethodsBySignature [0].Method.DeclaringType.FullName, distinctMethodsBySignature [1].Method.DeclaringType.FullName,
+	distinctMethodsBySignature [0].Method.ToString (), distinctMethodsBySignature [1].Method.ToString ()));
```
spouliot: minor: ^ incorrect alignment

src/generator-filters.cs
wachaudh: All of you other comments about diff error codes have the same answer so I've marked them as resolved.

src/error.cs
```diff
@@ -11,51 +11,65 @@
...
+	static string GetMessage (int code)
+	{
+		Type resourceType = Type.GetType ("bgen.Resources");
+		string errorCode = string.Format ("BI{0:0000}", code);
+		PropertyInfo prop = resourceType.GetProperty (errorCode, BindingFlags.NonPublic |
+				BindingFlags.Static |
+				BindingFlags.GetProperty);
+
+		var errorMessage = prop == null ? String.Format (bgen.Resources._default, errorCode) :
+					(String) prop.GetValue (null);
```
prmallic:
```suggestion
					prop.GetValue (null) as string;
```
To avoid cast exceptions?

src/Resources.Designer.cs
```diff
@@ -0,0 +1,716 @@
...
+        
+        private static System.Resources.ResourceManager resourceMan;
+        
+        private static System.Globalization.CultureInfo resourceCulture;
+        
+        [System.Diagnostics.CodeAnalysis.SuppressMessageAttribute("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
+        internal Resources() {
+        }
+        
+        [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Advanced)]
```
prmallic: I think this file isn't code-formatted to our guidelines. Edit->Format->Format Document on Visual Studio does the trick

src/generator-filters.cs
wachaudh: All of the comments about error numbers have the same reason so I've marked them as resolved.

src/Resources.Designer.cs
wachaudh: This file is auto-generated.

src/error.cs
rolfbjarne: Cast exceptions are usually easier to diagnose than null reference exceptions, so I think the current code is fine.

PR #7532, opened 2019-12-02
[Foundation] Ensure that we do not iterate over a collection accessed by other threads.

mandel-macaque:
Make sure we do no iterate ove the dictionary while other threads are
accessing it. Lock, get the sources, then cancel.

Fixes: https://github.com/xamarin/xamarin-macios/issues/7386

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -206,11 +206,14 @@ namespace Foundation {
 			// iteration. We split the operation in two, get all the diff cancelation sources, then try to cancel each of them
 			// which will do the correct lock dance. Note that we could be tempted to do a RemoveAll, that will yield the same
 			// runtime issue, this is dull but safe. 
-			var sources = new List <TaskCompletionSource<HttpResponseMessage>> (inflightRequests.Count);
-			foreach (var r in inflightRequests.Values) {
-				sources.Add (r.CompletionSource);
+			List <TaskCompletionSource<HttpResponseMessage>> sources = null; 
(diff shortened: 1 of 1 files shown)


src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -206,11 +206,14 @@ void BackgroundNotificationCb (NSNotification obj)
...
-				sources.Add (r.CompletionSource);
+			List <TaskCompletionSource<HttpResponseMessage>> sources = null; 
+			lock (inflightRequestsLock) { // just lock when we iterate
+				sources = new List <TaskCompletionSource<HttpResponseMessage>> (inflightRequests.Count);
+				foreach (var r in inflightRequests.Values) {
+					sources.Add (r.CompletionSource);
+				}
 			}
-			sources.ForEach (source => { source.TrySetCanceled (); });
+			sources?.ForEach (source => { source.TrySetCanceled (); });
```
spouliot: `sources` should never be `null` here

mandel-macaque: Agreed, but I have 0 trust ;)

spouliot: Just like me :) but adding `?` means you're hiding a problem (if any) which will only make diagnosing it much harder than it needs to be.

IOW my lack of trust makes me rather have this _should-never-be-null_ value throw an NRE. I rather have a bug I understand than one I don't ;-)

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/2069

mandel-macaque:
@monojenkins backport d16-5

PR #7553, opened 2019-12-07
Avoid multiple lookups in arrays

Therzok:


mandel-macaque:
Hmm I don't like some of the failures in the BCL, will rerun in case.

mandel-macaque:
build

mandel-macaque:
Much nicer results, I'll wait for the device test results, but they should not be a blocker for this.

mandel-macaque:
Nebula had issues running the tvOS tests.

mandel-macaque:
Tests look good. We have had issues with 32b for a long time and this PR does not make a change that would affect those.

PR #7555, opened 2019-12-07
[Runtime] Switch delegates list to use a GCHandle, to reduce in-memor…

Therzok:
…y usage

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -36,7 +36,6 @@ namespace ObjCRuntime {
 		static Dictionary<Type, ConstructorInfo> intptr_ctor_cache;
 		static Dictionary<Type, ConstructorInfo> intptr_bool_ctor_cache;
 
-		static List <object> delegates;
 		static List <Assembly> assemblies;
 		static Dictionary <IntPtr, WeakReference> object_map;
 		static object lock_obj;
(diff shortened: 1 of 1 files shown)


Therzok:
I don't think the reduction is _huge_. There are 35 objects queried with `GetFunctionPointer`. So the list would most likely be backed by an array of 64 items internally. Just something that stood out to me, as we never use the `delegates` list.

rolfbjarne:
Are you certain this actually saves memory? How much memory does each GCHandle need? Also I seem to remember that mono's GCHandle implementation wasn't very efficient, if there are many of them they can cause trouble for the GC. Do you know how this change would affect GC performance?

spouliot:
Like @rolfbjarne I'm unconvinced this saves memory, as you're now allocating `GCHandle`. Let's not claim such thing in commit messages without numbers to back.
Also it's not clear this is net positive change (even if some memory is saved).

mandel-macaque:
Added do not merge until we get the answers ;)

Therzok:
I did a benchmark on GC run times by using a long array held by static reference and gchandles.

https://github.com/Therzok/BenchPlayground/blob/409e98dc4f9a53758ad79c5a08b5dbebc36fb25b/BenchPlayground/Benchmarks/runtime/GCHandleVsReference.md

It seems you are right @rolfbjarne. It seems both runtimes have a higher GC impact due to GCHandles. closing the PR.

PR #7558, opened 2019-12-08
Use sizeof instead of Marshal.Sizeof

Therzok:


Therzok:
I've not migrated `public` methods to unmanaged constraint. The internal users don't use any kind of special marshal attributes.

Regardless, I could add tests.

Therzok:
Any suggestion of where to put the tests?

spouliot:
@Therzok by default we add them to `monotouch-test` which is executed under most possible configurations.

src/AudioUnit/AudioUnit.cs
```diff
@@ -644,7 +644,7 @@ public unsafe AudioUnitParameterInfo[] GetParameterList (AudioUnitScopeType scop
 			}
 
 			var info = new AudioUnitParameterInfo [data.Length];
-			size = (uint) Marshal.SizeOf (typeof (AudioUnitParameterInfoNative));
+			size = (uint)sizeof (AudioUnitParameterInfoNative);
```
Therzok: This is internal, so I'm not sure how to test that.

Therzok:
I've updated everything except for `AudioUnitParameterNative`, as that's internal

src/AudioUnit/AudioUnit.cs
Therzok: But the definition definitely should return the same values:
https://github.com/xamarin/xamarin-macios/blob/b8f9dccdcc09f43a44d591fd0eaf9ba954ea2752/src/AudioUnit/AudioUnit.cs#L235

spouliot:
build

spouliot:
failure is a known issue (HE0038) https://github.com/xamarin/maccore/issues/581

PR #7567, opened 2019-12-10
[introspection] Update according to changes that seem to have occurred in macOS 10.15.2.

rolfbjarne:

PR #7546, opened 2019-12-05
[monotouch-test] Simplify NetworkReachabilityTest.CtorIPAddressPair test.

rolfbjarne:
Don't assert anything about the flags, it's a never ending game to keep up with OS versions.

Fixes https://github.com/xamarin/maccore/issues/2047 for good.

PR #7568, opened 2019-12-10
Bump for Xcode 11.3 final

VincentDondain:


rolfbjarne:
Test failures are unrelated

* Xamarin.Mac tests on 10.15:
    * introspection: will be fixed with #7567.
    * xammac tests: will be fixed with #7546.
* Publish failure: xamarin-storage is out of disk space. This also prevent any tests to run, but since the tests ran successfully on the public Jenkins instance, it should be good.

PR #7591, opened 2019-12-12
[Tests] Improve error when we have network issues in ImageCaptioningTest.GetCaption

mandel-macaque:
The error does give a description letting us know that the url was
invalid, that is, we could not reach it.

Fixes: https://github.com/xamarin/maccore/issues/2088

### DIFF
--- a/tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
+++ b/tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
@@ -32,7 +32,12 @@ namespace MonoTouchFixtures.MediaAccessibility {
 			using (NSUrl url = new NSUrl (NetworkResources.MicrosoftUrl)) {
 				var s = MAImageCaptioning.GetCaption (url, out var e);
 				Assert.Null (s, "remote / return value");
-				Assert.Null (e, "remote / no error"); // weird should be an "image on disk"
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
```diff
@@ -32,7 +32,12 @@ public void GetCaption ()
 			using (NSUrl url = new NSUrl (NetworkResources.MicrosoftUrl)) {
 				var s = MAImageCaptioning.GetCaption (url, out var e);
 				Assert.Null (s, "remote / return value");
-				Assert.Null (e, "remote / no error"); // weird should be an "image on disk"
+				if (e != null && e.Description.Contains ("Invalid url:")) {
+					Assert.Null (e, "remote / no internet connection"); // could not connect to the network, fail and add a nice reason
```
wachaudh:
```suggestion
					Assert.Null (e, "remote / invalid url"); // could not connect to the network, fail and add a nice reason
```

spouliot: Since `e != null` no need to `Assert.Null (e, ...` just `Assert.Fail`
For the message maybe `"Ignore this failure when network is down"` just to be very clear that it's an expected failure only in this specific case

mandel-macaque:

PR #7601, opened 2019-12-16
[msbuild] Remove xbuild-specific code since we don't support building with xbuild anymore.

rolfbjarne:


rolfbjarne:
Test failure is unrelated

PR #7604, opened 2019-12-16
[DevOps] Convert our device tests into a pool agnostic yaml pipeline

VincentDondain:
- Needed for the DDFun migration of our tests
- Don't use SSH because it's restricted on the new bots
- Use brand new DevOps feature to clone multiple repos via HTTPS
- Update names of all tasks (clearer)
- Simplify and re-use code
- Update all paths because our cloning strategy is different compared to old pipeline
- Update xharness to print "experimental" titles in GitHub comments
- Update xharness to not add the Html link for tests from the DDFun pipeline
- Fix all timeouts to be relevant to the tests length
- Bump maccore to use the right certificates and profiles
- Bump maccore to target the right pipeline
- Bump maccore to not use ssh in scripts

Notes:
- Right now the xharness test output can only be seen when downloading the zip file locally (better output soon)
- Related [maccore PR](https://github.com/xamarin/maccore/pull/2095)
- This was pushed to origin so the tests should run and be reported in this PR


mk/xamarin.mk
```diff
@@ -7,8 +7,8 @@ MONO_BRANCH    := $(shell cd $(MONO_PATH) 2> /dev/null && git symbolic-ref --sho
 endif
 
 ifdef ENABLE_XAMARIN
-NEEDED_MACCORE_VERSION := 92433b7757c57de089b3edb0e7060d90363273f9
-NEEDED_MACCORE_BRANCH := master
+NEEDED_MACCORE_VERSION := 009fe1df33c98a5ab2d20107a0385b8aeb6089d0
+NEEDED_MACCORE_BRANCH := ddfun-device-tests
```
spouliot: can't be merged like this

VincentDondain: Yes you're right, I added to related PR obviously maccore needs to be merged first and this code updated. Early feedback welcome (;

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
...
+  - checkout: maccore
+
+  - bash: ./xamarin-macios/tools/devops/system-info.sh
+    displayName: System info
+
+  - bash: cd xamarin-macios && ./configure --enable-xamarin
+    displayName: 'Enable Xamarin'
+
+  - script: |
+     set -x
```
rolfbjarne: Minor, but I've used a two-character indentation in other yml files, a single space looks rather puny: https://github.com/xamarin/xamarin-macios/blob/d5135cce25e60419b70093b825d292765cc2a842/tools/devops/build-samples.yml#L136-L138

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
...
+     ./xamarin-macios/system-dependencies.sh $CONFIGURE_ARGS
+    displayName: 'Provision dependencies'
+    timeoutInMinutes: 240
+
+  - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
+    displayName: 'Install XI PKG'
+    inputs:
+      type: InlineScript
+      script: |
+       #!/bin/bash -ex
```
rolfbjarne: I think bigger scripts would be better as external .sh files, preferably in tools/devops/

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
...
+     set +e # Don't care about errors here
+
+     pkill -9 mlaunch
+     pkill -9 -f mono.*xharness.exe
+     pkill -9 -f ssh.*rsync.*xamarin-storage
+
+     exit 0
+    displayName: 'Process cleanup'
+
+  - task: CmdLine@1
```
rolfbjarne: All the script tasks can be simplified to use bash like this:

```suggestion
  - bash: security set-key-partition-list -S apple-tool:,apple: -s -k $(OSX_KEYCHAIN_PASS) login.keychain
```

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
...
+    continueOnError: true
+    condition: succeededOrFailed()
+
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish Artifact: HtmlReport'
+    inputs:
+      pathtoPublish: '$(Build.ArtifactStagingDirectory)/HtmlReport-$(Build.BuildId).zip'
+      artifactName: HtmlReport
+    continueOnError: true
+    condition: succeededOrFailed()
```
rolfbjarne: Minor: eol at end of file

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
...
+    displayName: 'Process cleanup'
+
+  - task: CmdLine@1
+    displayName: 'Remove security UI-prompt (http://stackoverflow.com/a/40039594/183422)'
+    inputs:
+      filename: security
+      arguments: ' set-key-partition-list -S apple-tool:,apple: -s -k $(OSX_KEYCHAIN_PASS) login.keychain'
+    timeoutInMinutes: 1
+
+  - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
```
rolfbjarne: Or a multi-line version:

```yml
  - bash: |
      set -e
      set -x
      echo 1234567890 > ~/.mlaunch-verbosity
    displayName: 'Make mlaunch verbose'
```

tools/devops/device-tests.yml
```diff
@@ -0,0 +1,236 @@
+resources:
```
rolfbjarne: I think we should put this file in tools/devops too.

tools/devops/device-tests.yml
VincentDondain: Tots agree, let me clean that up!

mandel-macaque: yep, same opinion. I wish we could have a language that can be tested easily.

tools/devops/device-tests.yml
VincentDondain: What do you mean? `device-tests.yml` is in `tools/devops`. Are you talking about moving the `repositories` section into its own file so it can be re-used? I don't know if that works (:

rolfbjarne: Never mind, for some reason I thought you'd put this file in jenkins/. My bad.

tools/devops/add-summaries.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+
+set +x
+
+X="#vso"
+echo help
```
rolfbjarne: 🆘

tools/devops/install-xi-pkg.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash -ex
+
+set -e
+set -x
```
rolfbjarne: These two `set`s are already in the shebang (as `-ex`), so you can remove them.

tools/devops/run-tests.sh
```diff
@@ -0,0 +1,15 @@
+set -x
+set -e
+
+P=$(cat tmp.p)
+
+if grep ios-64 xamarin-macios/tests/xharness/Jenkins.cs 2>&1 > /dev/null; then
```
rolfbjarne: You don't need the `ios-64` check here, you know that this check will always turn out true (it was needed when this code was in Azure Devops, because then it could execute for old branches, but that won' happen when the code is in git).

tools/devops/run-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+   export TESTS_EXTRA_ARGUMENTS="--label=run-ios-64-tests"
+else
+   export TESTS_EXTRA_ARGUMENTS="--label=run-ios-tests"
+fi
+
+make -C xamarin-macios/builds download -j || true
+make -C xamarin-macios/builds downloads -j || true
```
rolfbjarne: The `downloads` target doesn't exist, so this line can be removed.
```suggestion
```

tools/devops/run-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+
+P=$(cat tmp.p)
+
+if grep ios-64 xamarin-macios/tests/xharness/Jenkins.cs 2>&1 > /dev/null; then
+   export TESTS_EXTRA_ARGUMENTS="--label=run-ios-64-tests"
+else
+   export TESTS_EXTRA_ARGUMENTS="--label=run-ios-tests"
+fi
+
+make -C xamarin-macios/builds download -j || true
```
rolfbjarne: We know the `download` target exists, so we don't have to ignore the result:

```suggestion
make -C xamarin-macios/builds download -j
```

tools/devops/run-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+make -C xamarin-macios/builds downloads -j || true
+make -C xamarin-macios/builds .stamp-mono-ios-sdk-destdir -j || true
```
rolfbjarne: We know the `.stamp-mono-ios-sdk-destdir` target exists, so no need to ignore the result:

```suggestion
make -C xamarin-macios/builds .stamp-mono-ios-sdk-destdir -j
```

tools/devops/add-summaries.sh
VincentDondain: I can't tell if you're joking or you want me to remove that :P help!

tools/devops/run-tests.sh
VincentDondain: Ooooooooooh you're riiiiiiiiight!

jenkins/vsts-device-tests-set-status.sh
```diff
@@ -44,8 +44,6 @@ if test -z "$DEVICE_TYPE"; then
 	DEVICE_TYPE="iOS/tvOS"
 fi
 
-P=$(cat tmp.p)
```
VincentDondain: @rolfbjarne I don't know why that's breaking things :P 

Now it works: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3341819

Before it failed: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3338569

VincentDondain: Well ok so I guess it fails at `cat: tmp.p: No such file or directory` but I thought I saw other builds with that command failing and it still worked 🤷‍♂

VincentDondain:
HE0038 known issues: https://github.com/xamarin/maccore/issues/581

VincentDondain:
✅ Ok so all comments were addressed
✅  Set GitHub Status is fixes: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3341819
✅ Tests fully ran on the DDFun pool here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3338569 (only 14 failures)

🎉 I believe this is ready to merge 🎉

PR #7605, opened 2019-12-16
[Tests] Ensure that if the TCP connection cannot be done, test are run.

mandel-macaque:
TouchUnit uses Console.Out as the writer when the TCP connection cannot
be performed. The BCL tests were not doing so and were crashing.

With this change, when ran locally, if the TCP connection is not done,
we will use Console.Out as TouchUnit and the file will be parsed by
xharness correctly.

### DIFF
--- a/tests/bcl-test/templates/iOSApp/ViewController.cs
+++ b/tests/bcl-test/templates/iOSApp/ViewController.cs
@@ -17,6 +17,7 @@ using NUnit.Framework.Internal.Filters;
 
 namespace BCLTests {
 	public partial class ViewController : UIViewController {
+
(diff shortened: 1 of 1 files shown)


tests/bcl-test/templates/iOSApp/ViewController.cs
```diff
@@ -17,7 +17,6 @@
 
 namespace BCLTests {
 	public partial class ViewController : UIViewController {
-
```
spouliot: minor: white space noise

mandel-macaque:
@monojenkins backport d16-5

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1907

Issue #7509, opened 2019-11-28
Incorrect changes from int to (unsigned) long in the runtime


rolfbjarne:
The native `long` and `unsigned long` are 64-bit on a 64-bit OS and 32-bit on a 32-bit OS.

This means that this change is incorrect: https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-ef40c5787a0475f88e10461efe5f3c01L55-R55

Any (unsigned) int -> (unsigned) long commits must be fully reviewed again, there are several other places where there are bugs:

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-ef40c5787a0475f88e10461efe5f3c01L206-R206

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-5ca7e9836bcee437c268d6065e2c6eaeL26-R26 (breaking API change)

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-a41cda5f508168c4aafab2b33741eed4L384-R384 (breaking API change)

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-ba72c2a1b404212ae58242dc25660defL12-R13 (breaking API change - this one actually causes the link all tests to crash when run on a device in release mode)

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-6ddca81e8497599c1b02b643c8ed5128L29-R29 (breaking API change)

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-ae96dff46e2db1f319d0331d9beed8d2L184-R184 (breaking API change)

https://github.com/xamarin/xamarin-macios/commit/2e5b64893f694bb8b65d5626e556d3805d81428e#diff-0689f45915b28af398ed3ac8f5d98545L38-R38 (breaking API change)

rolfbjarne:
This also causes https://github.com/xamarin/maccore/issues/2072.

spouliot:
@mandel-macaque it needs to be backported to d16-5 since we still have test failures in that branch

spouliot:
[...]

PR #7529, opened 2019-12-02
[Runtime] Move some of the changes back to int.

mandel-macaque:
Some of the fixes done for the warnings have breaking changes. Move back
to int and ensure that we do not have any compilation errors (we are
using -Werror).

Fixes: https://github.com/xamarin/xamarin-macios/issues/7509

PR #7606, opened 2019-12-16
[d16-5][Runtime] Move some of the changes back to int. (#7529)

mandel-macaque:
Some of the fixes done for the warnings have breaking changes. Move back
to int and ensure that we do not have any compilation errors (we are
using -Werror).

Fixes: https://github.com/xamarin/xamarin-macios/issues/7509

### DIFF
--- a/runtime/monotouch-debug.m
+++ b/runtime/monotouch-debug.m
@@ -381,9 +381,9 @@ monotouch_set_connection_mode (const char *mode)
 }
 
 void
-monotouch_set_monodevelop_port (long port)
+monotouch_set_monodevelop_port (int port)
 {
-	monodevelop_port = port;
+	monodevelop_port = (long) port;
 }
 
 void
--- a/runtime/delegates.t4
+++ b/runtime/delegates.t4
@@ -52,7 +52,7 @@
 
 		new XDelegate ("MonoObject *", "IntPtr", "xamarin_get_block_wrapper_creator",
 			"MonoObject *", "IntPtr", "method",
-			"unsigned long", "uint", "parameter"
+			"int", "int", "parameter"
 		) {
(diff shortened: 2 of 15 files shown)


runtime/monotouch-debug.m
```diff
@@ -381,9 +381,9 @@ -(void) URLSession: (NSURLSession *) session task: (NSURLSessionTask *) task did
 }
 
 void
-monotouch_set_monodevelop_port (long port)
+monotouch_set_monodevelop_port (int port)
 {
-	monodevelop_port = port;
+	monodevelop_port = (long) port;
```
whitneyschmidt: My understanding is that we are trying to remove long/unsigned long types because their size changes depending on the platform and somehow that is a breaking change?

It doesn't seem like there is any harm in casting an `int` to a `long`, but why are we doing this here when everywhere else we are removing the `long` type?

spouliot: > remove long/unsigned long types

Not quite. They (`[u]int/[u]long`) are all fine individually... until you mix them up. 
* `long` -> `int` can overflow;
* `sign/unsigned` changes can be issues too, e.g. `-1` to unsigned;

The goal is not to remove anything specific but to be consistent (correctness) in their usage.

> It doesn't seem like there is any harm in casting an int to a long

Right, it is generally not an issue... unless mapped to memory, e.g. change a structure member from `int` to `long` can (alignment dependent) change its size or element position, which might be defined elsewhere (e.g. mono or in managed code).

An earlier PR enabled the compiler to spot those changes (i.e. emit warnings) and also turn warnings into errors. 

However there were issues in the previous fixes (the compiler does not see everything, having no warning/error does not mean the code works). @rolfbjarne listed them inside https://github.com/xamarin/xamarin-macios/issues/7509

mandel-macaque:

Issue #6328, opened 2019-06-17
[mono 2019-06] mscorlib tests fail on iOS and tvOS


mandel-macaque:
The following tests fail on an iOS 64 device:

> [FAIL] System.Text.Tests.StringBuilderTests.Append_CharPointer_Null_ThrowsNullReferenceException 
> [FAIL] System.Text.Tests.StringBuilderTests.Append_CharPointer_Null_ThrowsNullReferenceException Test name: System.Text.Tests.StringBuilderTests.Append_CharPointer_Null_ThrowsNullReferenceException 
> [FAIL] System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable 
> [FAIL] System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable Test name: System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable 

And on tvOS we have the following failures:

> [FAIL] System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable 
> [FAIL] System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable Test name: System.Reflection.Tests.MethodBaseTests.Test_GetCurrentMethod_Inlineable 

Failures only happen on Release (probably related to the linker).

mandel-macaque:
Still happens, mono issue related to this is https://github.com/mono/mono/issues/14761

rolfbjarne:
We don't run the BCL tests anymore.

PR #7615, opened 2019-12-17
[Tests] Reenable ignored test after they pass on device.

mandel-macaque:
Ran the tests on a iOS 64b device and passed. Looks like test were
fixed, probably by an update in netcore but mono did not close the
issue.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6328


mandel-macaque:
FYI device test will be ran for this PR.

mandel-macaque:
Bummer, seems that Release is broken. Closing.

PR #7619, opened 2019-12-18
[d16-5] Merge Xcode 11.3 into d16-5

dalexsoto:


spouliot:
API diff breakage are fine (version numbers only)

Make.versions
```diff
@@ -43,5 +43,5 @@
 #   line changed in git).
 #
 
-IOS_PACKAGE_VERSION=13.14.0.$(IOS_COMMIT_DISTANCE)
-MAC_PACKAGE_VERSION=6.14.0.$(MAC_COMMIT_DISTANCE)
+IOS_PACKAGE_VERSION=13.14.1.$(IOS_COMMIT_DISTANCE)
```
VincentDondain: @spouliot history master: do we do that for preview releases? Wouldn't we increase the min version numbers for Xcode bumps **after** this is released to the public?

spouliot: yes we (often [1]) did for significant changes, e.g. a new mono or xcode or service releases (but the later became much harder due to process changes)

Makes it easier to spot when those happens, 13.14.0.100 and 13.14.0.101 having a different Xcode is hard to guess.

[1] we likely forgot a few times too :)

dalexsoto:
1 Known failure: HE0038

PR #7621, opened 2019-12-18
[msbuild] Use package references.

rolfbjarne:
One of many steps towards building for netstandard 2.0.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks/Xamarin.MacDev.Tasks.csproj
+++ b/msbuild/Xamarin.MacDev.Tasks/Xamarin.MacDev.Tasks.csproj
@@ -38,10 +38,9 @@
     <Reference Include="Microsoft.Build.Utilities.Core" HintPath="$(MSBuildBinPath)\Microsoft.Build.Utilities.Core.dll" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
-    <Reference Include="Mono.Cecil, Version=0.9.5.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756">
-      <Private>True</Private>
-      <HintPath Condition=" Exists('..\..\packages\Mono.Cecil.0.9.5.0\lib\net40\Mono.Cecil.dll') ">..\..\packages\Mono.Cecil.0.9.5.0\lib\net40\Mono.Cecil.dll</HintPath>
-    </Reference>
+    <PackageReference Include="Mono.Cecil" Version="0.9.5.0">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\external\Xamarin.MacDev\Xamarin.MacDev\Xamarin.MacDev.csproj">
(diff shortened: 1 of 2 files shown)


msbuild/Xamarin.MacDev.Tasks/Xamarin.MacDev.Tasks.csproj
```diff
@@ -38,10 +38,9 @@
     <Reference Include="Microsoft.Build.Utilities.Core" HintPath="$(MSBuildBinPath)\Microsoft.Build.Utilities.Core.dll" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
-    <Reference Include="Mono.Cecil, Version=0.9.5.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756">
-      <Private>True</Private>
-      <HintPath Condition=" Exists('..\..\packages\Mono.Cecil.0.9.5.0\lib\net40\Mono.Cecil.dll') ">..\..\packages\Mono.Cecil.0.9.5.0\lib\net40\Mono.Cecil.dll</HintPath>
-    </Reference>
+    <PackageReference Include="Mono.Cecil" Version="0.9.5.0">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
spouliot: We should be using the latest. This already caused an issue with XF using something newer, but it can only be loaded once into memory for all msbuild tasks.

I guess it could use ILRepack too... but the best solution would be not to use Cecil at all in msbuild tasks since it's slow (to load every [1] assemblies) for extracting files. `mtouch` already has everything loaded (no extra cost) and could do the same (and already does similar extraction).

[1] it was updated not to load everything a few years back, assuming well-known assemblies (like the SDK) would not contain such resources.

rolfbjarne: > guess it could use ILRepack too

We already use ILRepack to merge Mono.Cecil.dll into the task assembly we ship, so in-process assembly version conflict is not an issue.

I agree we should move this logic to mtouch, but that can be done separately from the move to netstandard, it does not need to block this PR.

spouliot: If ILRepack works (include Cecil) then yes, let's postpone :)

@kzu do you have a link to the original issue wrt Mono.Cecil ? that was discussed a few weeks back on #ios

rolfbjarne: Conversation: https://xamarinhq.slack.com/archives/C03CCJNR7/p1574712036425800, but it doesn't link to any XF bugs.

kzu: @joj could you link the Cecil issue we faced several days ago that affected the debugger and was related to some Cecil mismatch?

Therzok: Is a typesystem needed? Otherwise System.Reflection.Metadata might be the cheaper option.

rolfbjarne: @joj said: "My commit was this and the problem was nrefactory but I didn't link an issue there https://github.com/xamarin/debugger-vs/commit/dc2f8878da54a75e72171c47d09326ed81ecac97" (https://xamarinhq.slack.com/archives/C03CCJNR7/p1576858598064900)

Which means it was not related to our MSBuild tasks' Cecil usage.

PR #7625, opened 2019-12-18
[xharness] Support the split of test assemblies from the BCL.

mandel-macaque:
Some of the test assemblies are too large and will be splitted by mono so
that they can be compiled for iOS 32b. In that case, we are using the
following pattern

`
test_assembly_name.dll
`

becomes

`
test_assembly_name.part1.dll
test_assembly_name.part2.dll
`

Perse the only change we need to add to make our life easier is to be
able to mantain a single .ignore file since mantaining more .ignore
files is hard and more error prone. This change simply checks if we are
working with a splitted dll and ensures that the correct .ignore files
are added to the BCL test application.

### DIFF
--- a/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
+++ b/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
@@ -15,7 +15,7 @@ namespace BCLTestImporter {
 
 		static string NUnitPattern = "monotouch_*_test.dll"; 
 		static string xUnitPattern = "monotouch_*_xunit-test.dll";
-		internal static string splitPattern = ".part";
+		internal static readonly string splitPattern = ".part";
 		internal static readonly string ProjectGuidKey = "%PROJECT GUID%";
 		internal static readonly string NameKey = "%NAME%";
 		internal static readonly string ReferencesKey = "%REFERENCES%";
@@ -441,8 +441,9 @@ namespace BCLTestImporter {
 				// we could be looking at a splitted assembly, if that is the case, lets pass the name of the dll without the 'part{number}.dll
 				// so that we have all the ignores in a single file
 				var assemblyName = assembly;
-				if (assembly.Contains (splitPattern)) {
-					var index = assembly.IndexOf (splitPattern, StringComparison.Ordinal);
+				var index = assembly.IndexOf (splitPattern, StringComparison.Ordinal);
+
+				if (index != -1) {
 					assemblyName = assembly.Substring (0, index) + ".dll";
 				}
 				foreach (var platformFile in GetIgnoreFileNames (assemblyName, platform)) {
--- a/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
+++ b/tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
@@ -442,7 +442,7 @@ namespace BCLTestImporter {
 				// so that we have all the ignores in a single file
 				var assemblyName = assembly;
 				if (assembly.Contains (splitPattern)) {
-					var index = assembly.IndexOf(splitPattern, StringComparison.Ordinal);
+					var index = assembly.IndexOf (splitPattern, StringComparison.Ordinal);
 					assemblyName = assembly.Substring (0, index) + ".dll";
 				}
 				foreach (var platformFile in GetIgnoreFileNames (assemblyName, platform)) {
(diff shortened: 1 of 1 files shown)


tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -437,8 +438,15 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
 			}
 			// do we have ignores per files and not the project name? Add them
 			foreach (var (assembly, hintPath) in assemblies) {
-				foreach (var platformFile in GetIgnoreFileNames (assembly, platform)) {
-					var commonAssemblyIgnore = Path.Combine (templateDir, GetCommonIgnoreFileName (assembly, platform));
+				// we could be looking at a splitted assembly, if that is the case, less pass the name of the dll without the 'part{number}.dll
+				// so that we have all the ignores in a single file
+				var assemblyName = assembly;
+				if (assembly.Contains (splitPattern)) {
+					var index = assembly.IndexOf(splitPattern, StringComparison.Ordinal);
```
spouliot: minor: space before `(`

spouliot: that can be simplified, just to the `IndexOf`
i.e. there's no need to call `Contains` first, `IndexOf` will return `-1` if the content is not present

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -15,6 +15,7 @@ public class BCLTestProjectGenerator {
 
 		static string NUnitPattern = "monotouch_*_test.dll"; 
 		static string xUnitPattern = "monotouch_*_xunit-test.dll";
+		internal static string splitPattern = ".part";
```
spouliot: why not `redline` like the other constants that follows ?

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -437,8 +438,15 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
 			}
 			// do we have ignores per files and not the project name? Add them
 			foreach (var (assembly, hintPath) in assemblies) {
-				foreach (var platformFile in GetIgnoreFileNames (assembly, platform)) {
-					var commonAssemblyIgnore = Path.Combine (templateDir, GetCommonIgnoreFileName (assembly, platform));
+				// we could be looking at a splitted assembly, if that is the case, less pass the name of the dll without the 'part{number}.dll
```
prmallic:
```suggestion
				// we could be looking at a splitted assembly, if that is the case, lets pass the name of the dll without the 'part{number}.dll
```

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
prmallic:
```suggestion
					var index = assembly.IndexOf (splitPattern, StringComparison.Ordinal);
```

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -15,6 +15,7 @@ public class BCLTestProjectGenerator {
 
 		static string NUnitPattern = "monotouch_*_test.dll"; 
 		static string xUnitPattern = "monotouch_*_xunit-test.dll";
+		internal static readonly string splitPattern = ".part";
```
wachaudh:
```suggestion
		internal static readonly string SplitPattern = ".part";
```

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -437,8 +438,16 @@ internal static IEnumerable<string> GetIgnoreFiles (string templatePath, string
...
-				foreach (var platformFile in GetIgnoreFileNames (assembly, platform)) {
-					var commonAssemblyIgnore = Path.Combine (templateDir, GetCommonIgnoreFileName (assembly, platform));
+				// we could be looking at a splitted assembly, if that is the case, lets pass the name of the dll without the 'part{number}.dll
+				// so that we have all the ignores in a single file
+				var assemblyName = assembly;
+				var index = assembly.IndexOf (splitPattern, StringComparison.Ordinal);
+
+				if (index != -1) {
+					assemblyName = assembly.Substring (0, index) + ".dll";
+				}
```
wachaudh: remove curly braces per mono style

tools/bcl-test-importer/BCLTestImporter/BCLTestProjectGenerator.cs
wachaudh: make it pascal case like the other internal static readonly strings

mandel-macaque:
@monojenkins backport d16-5

PR #7633, opened 2019-12-20
[mtouch] Fix case where `copyfile` could fail silently

spouliot:
Copying the framework could fail (error 260) and the failure was never
reported so the build succeeded - but without updating (completely) the
framework.

The exact reason it fails is unknown :( but we can recover from it by
deleting the target and copying (everything) back to the expected
(target) location.

Build logs will now indicate when this fails and will try to recover
before reporting a build error. Best case it works :) worse case we'll
be aware something is wrong (which is better than ignoring)

This is half the fix from https://github.com/xamarin/xamarin-macios/pull/7544
which was reverted due to the second half causing a regression (under
investigation).

Reference: https://github.com/xamarin/xamarin-macios/issues/7514 (partial fix)

### DIFF
--- a/tools/common/FileCopier.cs
+++ b/tools/common/FileCopier.cs
@@ -78,19 +78,17 @@ namespace Xamarin.Bundler {
 		public static void UpdateDirectory (string source, string target)
 		{
 			// first chance, try to update existing content inside `target`
-			int rv = TryUpdateDirectory (source, target);
-			if (rv == 0)
+			if (TryUpdateDirectory (source, target, out var err))
 				return;
 
 			// 2nd chance, nuke `target` then copy everything
-			var err = Marshal.GetLastWin32Error (); // might not be very useful since the callback signaled an error (CopyFileResult.Quit)
 			Log (1, "Could not update `{0}` content (error #{1} : {2}), trying to overwrite everything...", target, err, strerror (err));
 			Directory.Delete (target, true);
-			if (TryUpdateDirectory (source, target) != 0)
-				throw CreateError (1022, "Could not copy the directory '{0}' to '{1}': {2}", source, target, strerror (Marshal.GetLastWin32Error ()));
+			if (!TryUpdateDirectory (source, target, out err))
+				throw CreateError (1022, "Could not copy the directory '{0}' to '{1}': (error #{2} : {3})", source, target, err, strerror (err));
 		}
 
-		static int TryUpdateDirectory (string source, string target)
+		static bool TryUpdateDirectory (string source, string target, out int errno)
 		{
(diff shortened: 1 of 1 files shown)


tools/common/FileCopier.cs
```diff
@@ -77,23 +77,36 @@ enum CopyFileWhat {
...
 			// Mono's File.Copy can't handle symlinks (the symlinks are followed instead of copied),
 			// so we need to use native functions directly. Luckily OSX provides exactly what we need.
 			IntPtr state = copyfile_state_alloc ();
 			try {
 				CopyFileCallbackDelegate del = CopyFileCallback;
 				copyfile_state_set (state, CopyFileState.StatusCB, Marshal.GetFunctionPointerForDelegate (del));
-				int rv = copyfile (source, target, state, CopyFileFlags.Data | CopyFileFlags.Recursive | CopyFileFlags.Nofollow | CopyFileFlags.Clone);
-				if (rv != 0)
-					throw CreateError (1022, "Could not copy the directory '{0}' to '{1}': {2}", source, target, strerror (Marshal.GetLastWin32Error ()));
+				return copyfile (source, target, state, CopyFileFlags.Data | CopyFileFlags.Recursive | CopyFileFlags.Nofollow | CopyFileFlags.Clone);
```
rolfbjarne: A minor improvement might be to return `bool` from this function and have `errno` as an out parameter:

```csharp
static bool TryUpdateDirectory (string source, string target, out int errno)
{
    ...
    var rv = copyfile (source, target, state, CopyFileFlags.Data | CopyFileFlags.Recursive | CopyFileFlags.Nofollow | CopyFileFlags.Clone);
    if (rv != 0) {
        errno =  Marshal.GetLastWin32Error ();
        return false;
    }
    ...
}

public static void UpdateDirectory (string source, string target)
{
	if (!Directory.Exists (target))
		Directory.CreateDirectory (target);
	// first chance, try to update existing content inside `target`
	if (TryUpdateDirectory (source, target, out var errno))
		return;

	// 2nd chance, nuke `target` then copy everything
	Log (1, "Could not update `{0}` content (error #{1} : {2}), trying to overwrite everything...", target, errno, strerror (errno));
	Directory.Delete (target, true);
	if (!TryUpdateDirectory (source, target, out errno))
		throw CreateError (1022, "Could not copy the directory '{0}' to '{1}': {2}", source, target, strerror (errno));
}
```

This makes sure `Marshal.GetLastWin32Error` is called right after the P/Invoke, and it's also feels a bit more like C#.

spouliot:
known failure (HE0038) on watchOS https://github.com/xamarin/maccore/issues/581

PR #7636, opened 2019-12-20
[msbuild] Don't leave changes behind if a test fails.

rolfbjarne:


mandel-macaque:

Issue #5811, opened 2019-03-27
[CFNetwork] Add missing bindings for the constants required to set a local proxy


mandel-macaque:
The only way to set a proxy is with a NSDictionary with the following constants: 

* kCFStreamPropertyHTTPProxyHost 
* kCFStreamPropertyHTTPProxyPort
* kCFStreamPropertyHTTPSProxyHost
* kCFStreamPropertyHTTPSProxyPort

And would be nice to have HTTPEnable and HTTPSEnabled too. The best way to do so would be to expose a smart dict on NSURlSessionConfiguration. 

Right now we have to do something like:

```CSharp
var lib = Dlfcn.dlopen (Constants.CFNetworkLibrary, 0);
var kCFStreamPropertyHTTPProxyHost = Dlfcn.GetStringConstant (lib, "kCFStreamPropertyHTTPProxyHost");
var kCFStreamPropertyHTTPProxyPort = Dlfcn.GetStringConstant (lib, "kCFStreamPropertyHTTPProxyPort");
var kCFStreamPropertyHTTPSProxyHost = Dlfcn.GetStringConstant (lib, "kCFStreamPropertyHTTPSProxyHost");
var kCFStreamPropertyHTTPSProxyPort = Dlfcn.GetStringConstant (lib, "kCFStreamPropertyHTTPSProxyPort");
var httpEnable = new NSString ("HTTPEnable");
var httpsEnable = new NSString ("HTTPSEnable");
var proxyConfig = new NSDictionary (
			httpEnable, new NSNumber (true),
			kCFStreamPropertyHTTPProxyHost, new NSString ("myhost"),
			kCFStreamPropertyHTTPProxyPort, new NSNumber (8888),
			httpsEnable, new NSNumber (true),
			kCFStreamPropertyHTTPSProxyHost, new NSString ("myhttpsHost"),
			kCFStreamPropertyHTTPSProxyPort, new NSNumber (8888));
var config = NSUrlSessionConfiguration.DefaultSessionConfiguration; // gets a copy, not the same instance!!!!!!
config.ConnectionProxyDictionary = proxyConfig;
var handler = new NSUrlSessionHandler (config);
var client = new HttpClient (handler);
```

spouliot:
agreed a smart dictionary seems the best option

PR #7640, opened 2019-12-20
[Foundation] Provide a strong dictionary for proxy configuration.

mandel-macaque:
Although we provide the correct property to set the proxy configuration,
we do not add a strong dict that will make it nicer for users. This
fixes that API.

Fixes: https://github.com/xamarin/xamarin-macios/issues/5811


src/foundation.cs
```diff
@@ -7378,6 +7378,41 @@ partial interface NSUrlSessionDownloadTask {
 		void Cancel (Action<NSData> resumeCallback);
 	}
 	
+	[Internal]
+	[Static]
+	[NoWatch]
+	partial interface ProxyConfigurationDictionaryKeys {
```
spouliot: is it really `partial` ?
if it is then your `[Static]` and `[Internal]` might overreach

mandel-macaque: Good point, will remove partial.

mandel-macaque:

PR #7643, opened 2019-12-21
Fix NSUuid constructor. Pinning is only for the duration of the fixed…

Therzok:
… statement


mandel-macaque:
build

mandel-macaque:
Know issue: https://github.com/xamarin/maccore/issues/581

PR #7647, opened 2019-12-23
[msbuild] Avoid lines starting with a whitespace inside generated response files. Fix #7645

### LINKED ISSUE #7645: CommandLineArgumentBuilder adds extra spaces at beginning of (some) lines
Note: it's not a big deal, even less if we supported the full .rsp spec (see #7644) [1] but it does affect how mtouch will parse options later.

[1] except if we wanted to add comments (lines must starts with `#` and a space would break this)

What happens is that calls to `AddQuoted` behave correctly while calls to `AddLine` adds an extra space at the start of new lines.

This is likely due to recent changes regarding quoting and the fact that we did not care (long ago) to have each option into a separate line.

### Steps to Reproduce

1. Build a project 
2. Look inside the generated `response-file.rsp` file

### Expected Behavior

No line starting with whitespaces.

### Actual Behavior

```
...
-r=/Users/poupou/git/xamarin/xamarin-macios/builds/downloads/ios-release-Darwin-8f396bbb408b5758fccb8602030b9fa5293ce718/ios-bcl/monotouch/tests/Xunit.NetCore.Extensions.dll
 --target-framework=Xamarin.iOS,v1.0
--root-assembly=/Users/poupou/git/xamarin/xamarin-macios/tests/xharness/tmp-test-dir/mscorlib/bin/mscorlib/iPhoneSimulator/Debug-unified/com.xamarin.bcltests.mscorlib.exe
 -v -v -v -v
```

### Environment

master

### References
This is a bit related to https://github.com/xamarin/xamarin-macios/issues/7514 and https://github.com/xamarin/xamarin-macios/issues/7644

spouliot:
Since the "quote" refactoring there's no real difference between `Add*`
and `Add*Line` methods, everything is now line-oriented.

However `Add` and `AddQuoted` behaved differently, which meant that `Add`
would add a whitespace before a new argument in the `StringBuilder` was
not empty (as a separator needed _when_ different lines were not used).

That resulted in response files that had some lines starting with a white
space. That was not an issue (spec [1] wise) but it broke some (not so
correct) assumptions in `mtouch` and could also break anything that added
comments (which must start with `#`) in the future.

This brings consistency in the output whether quoted/non-quoted arguments
are used.

Reference:
[0] https://github.com/xamarin/xamarin-macios/issues/7645
[1] https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/response-file-compiler-option

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
+++ b/msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
@@ -242,9 +242,12 @@ namespace Xamarin.Mac.Tasks
 
 			actualArgs.AddQuoted ($"@{responseFile}");
 
-			if (!string.IsNullOrWhiteSpace (ExtraArguments))
+			if (!string.IsNullOrWhiteSpace (ExtraArguments)) {
+				actualArgs.Add (" ");
 				actualArgs.Add (ExtraArguments);
+			}
 
+			actualArgs.Add (" ");
 			var verbosity = VerbosityUtils.Merge (ExtraArguments, (LoggerVerbosity) Verbosity);
 			// for compatibility with earlier versions nothing means one /v
 			actualArgs.AddLine (verbosity.Length > 0 ? verbosity : "/verbose");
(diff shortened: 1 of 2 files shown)


spouliot:
@prmallic it should get clearer with some other commits :) but https://github.com/xamarin/xamarin-macios/blob/1c3bd7f11477a2b9b36e612f86c4f793861aec2c/tools/common/cache.cs#L219 is the main issue (assumes not per-line, no space, arguments)

spouliot:
failure is a known, watchOS deploy, issue -> https://github.com/xamarin/maccore/issues/581

PR #7662, opened 2019-12-27
[xharness] Only run 'dont link' on release configuration

spouliot:
The 32bits **debug** binaries are now too big for Apple's native linker
to process, which gives us (non useful) build errors on the bots.

This will still run the release builds configuration of the tests since
they are smaller and still within the limits of the tooling.


mandel-macaque:

mandel-macaque:
@monojenkins backport d16-5

PR #7670, opened 2020-01-02
[Introspection] Allow introspection to test CAMetalLayer when xcode 11 and simulator.

mandel-macaque:
The order of the check matters, we only skip with a sim && not on
xcode11.

The other whay around will do the wrong short circuit. Since if we are
not on Xcode11 we will always skip, including device, which is not
correct.

Fixes: https://github.com/xamarin/xamarin-macios/issues/6243.

### DIFF
--- a/tests/introspection/iOS/iOSApiCtorInitTest.cs
+++ b/tests/introspection/iOS/iOSApiCtorInitTest.cs
@@ -157,7 +157,7 @@ namespace Introspection {
 
 			// Metal is not available on the (iOS8) simulator
 			case "CAMetalLayer":
-				return  (Runtime.Arch == Arch.SIMULATOR) && !TestRuntime.CheckXcodeVersion (11, 0);
+				return (Runtime.Arch == Arch.SIMULATOR) && !TestRuntime.CheckXcodeVersion (11, 0);
 
 #if !XAMCORE_2_0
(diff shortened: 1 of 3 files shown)


mandel-macaque:
Please do test this in an older version of the OS, I do not have any old machine (at the office at least).

tests/introspection/iOS/iOSApiCtorInitTest.cs
```diff
@@ -157,7 +157,7 @@ protected override bool Skip (Type type)
 
 			// Metal is not available on the (iOS8) simulator
 			case "CAMetalLayer":
-				return (Runtime.Arch == Arch.SIMULATOR);
+				return  (Runtime.Arch == Arch.SIMULATOR) && !TestRuntime.CheckXcodeVersion (11, 0);
```
wachaudh:
```suggestion
				return (Runtime.Arch == Arch.SIMULATOR) && !TestRuntime.CheckXcodeVersion (11, 0);
```

mandel-macaque:
Screenshot 2020-01-05 at 18 25 39
Above are the results on Mojave 10.14.5

mandel-macaque:
Known issues: https://github.com/xamarin/maccore/issues/581

PR #7683, opened 2020-01-06
[Device tests] Bump maccore to fix device tests.

mandel-macaque:
Bump maccore to get a fix in the launch.sh script that pings the device
tests.

### DIFF
--- a/mk/xamarin.mk
+++ b/mk/xamarin.mk
@@ -7,7 +7,7 @@ MONO_BRANCH    := $(shell cd $(MONO_PATH) 2> /dev/null && git symbolic-ref --sho
 endif
 
 ifdef ENABLE_XAMARIN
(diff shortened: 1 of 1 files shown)


mk/xamarin.mk
```diff
@@ -7,7 +7,7 @@ MONO_BRANCH    := $(shell cd $(MONO_PATH) 2> /dev/null && git symbolic-ref --sho
 endif
 
 ifdef ENABLE_XAMARIN
-NEEDED_MACCORE_VERSION := 113cf2ad61ee0ecf9affde6ebae7804744f47933
+NEEDED_MACCORE_VERSION := b2d37c77cc34558a67fd328cf886ca00ba413db4
```
prmallic: Should this also be backported to other branches?

mandel-macaque: I need to check the DDFun stuff, AFAIK no, but we can take a look at the tvOS jobs and decide.

PR #7685, opened 2020-01-07
[Master][Bump] Bump mono to match android.

mandel-macaque:
Commits are:
* [domain] Don't add NULL runtime to runtimes list (#18265) https://github.com/mono/mono/commit/dc5902c243ec0b033b21ba8af3bc2ca5035bec7c
* [eglib] Don't define _FORTIFY_SOURCE on Cygwin builds (#18312)  https://github.com/mono/mono/commit/5f0704ce372a1dcd575538351a0ce0c0b229f57e
* Socket.BeginMConnect() should not attempt connections on unsupported address families. https://github.com/mono/mono/commit/99e3aed6ec366383372f75af07994162f247d9ff
* [2019-10] [metadata] Size 0 Blob heap is ok when resolving assembly refs (#18338) https://github.com/mono/mono/commit/2faea92a92469322e29edc8f0fdeac24bd316fd0
* [FileSystemWatcher] Fix wrong behaviour when move multiple files out of watched folder on mac (#41963) https://github.com/mono/mono/commit/df42020fe6f7a0a1b5fa59b481fb603d88d72348

Complete diff is: https://github.com/mono/mono/compare/bfcac9bd309a05bcc1820e9ff1cb167708e4d6f4...df42020fe6f7a0a1b5fa59b481fb603d88d72348


mandel-macaque:
Related android branch: https://github.com/xamarin/xamarin-android/pull/4083

mk/mono.mk
```diff
@@ -1,4 +1,4 @@
-NEEDED_MONO_VERSION := bfcac9bd309a05bcc1820e9ff1cb167708e4d6f4
+NEEDED_MONO_VERSION := df42020fe6f7a0a1b5fa59b481fb603d88d72348
 NEEDED_MONO_BRANCH := 2019-10
```
whitneyschmidt: This doesn't need to be bumped too?

mandel-macaque: no, we only care about the hash. Master is atm in the same mono as d16-5 but we have the mono integration branch for the bump to 2019-12.

Bumping branches in mono (rather than a hash) is more dangerous and that is why we have the mono monthly integration branch, we use it to find possible mono issues or issues with our code and then merge it to master.

Issue #7446, opened 2019-11-14
DispatchData does not have a very user friendly API


mandel-macaque:
The class does not have very user friendly methods as well as could use some updates to be able to handler ReadOnlySpan.

Useful methods to add to the current API:

* `DispatchData FromReadOnlySpan (ReadOnlySpan<byte>)`
* `DispatchData FromReadOnlySpan (ReadOnlySpan<byte>, int start, int length)`
*  `byte [] ToArray ()`
* `ReadOnlySpan <byte> ToSpan ()`

Therzok:
The start/length overloads are pretty much redundant with ReadOnlySpan.

You can do `obj.AsSpan(start, length)` or `span.Slice(start, length)` at almost no cost.

Therzok:
Also, maybe provide `in Span<T>` support.

spouliot:
not sure what's the most common name in the BCL update but `FromReadOnlySpan` has no verb and we generally don't use that pattern

Therzok:
BCL usually uses overloaded `From`, since there is overloadability in parameter type.

PR #7690, opened 2020-01-08
[CoreFoundation] Add some helper methods to make DispatchData easier to use.

mandel-macaque:
Add the following methods:

1. FromReadOnlySpan - Allows to create DispatchData from Spans.
2. ToArray - Return a managed copy of the data.

Fixes: https://github.com/xamarin/xamarin-macios/issues/7446

### DIFF
--- a/src/CoreFoundation/DispatchData.cs
+++ b/src/CoreFoundation/DispatchData.cs
@@ -88,6 +88,18 @@ namespace CoreFoundation {
 			return new DispatchData (dd, owns: true);
 		}
 
+		// create a dd using the data of the span 
+		public static DispatchData FromReadOnlySpan (ReadOnlySpan<byte> content)
+		{
+			unsafe {
+				fixed (byte* ptr = content) {
+					// set the block that will be used to clean the data once we are done with it
+					var dd = dispatch_data_create ((IntPtr)ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
+					return new DispatchData (dd, owns: true);
+				}
+			}
+		}
+
 		[DllImport (Constants.libcLibrary)]
 		extern static nuint dispatch_data_get_size (IntPtr handle);
 
@@ -122,6 +134,19 @@ namespace CoreFoundation {
 		{
 			return new DispatchData (dispatch_data_create_subrange (Handle, offset, size), owns: true);
 		}
+
+		// copies the dispatch data to a managed array
+		public byte [] ToArray ()
+		{
+			IntPtr bufferAddress = IntPtr.Zero;
+			nuint bufferSize = 0;
+			using DispatchData dataCopy = CreateMap (out bufferAddress, out bufferSize);
+
+			byte[] managedArray = new byte[(int) bufferSize];
+			Marshal.Copy (bufferAddress, managedArray, 0, (int) bufferSize);
+			return managedArray;
+		}
+
 #endif
 	}
 }
--- a/src/CoreFoundation/DispatchData.cs
+++ b/src/CoreFoundation/DispatchData.cs
@@ -94,7 +94,7 @@ namespace CoreFoundation {
 			unsafe {
 				fixed (byte* ptr = content) {
 					// set the block that will be used to clean the data once we are done with it
-					var dd = dispatch_data_create ((IntPtr)ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
+					var dd = dispatch_data_create ((IntPtr) ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
 					return new DispatchData (dd, owns: true);
 				}
 			}
--- a/src/CoreFoundation/DispatchData.cs
+++ b/src/CoreFoundation/DispatchData.cs
@@ -93,7 +93,9 @@ namespace CoreFoundation {
 		{
 			unsafe {
 				fixed (byte* ptr = content) {
-					// set the block that will be used to clean the data once we are done with it
+					// As per the documentation for dispatch_data_create, seems the data the pointer points to is copied 
+					// internally when specifying DISPATCH_DATA_DESTRUCTOR_DEFAULT as the destructor,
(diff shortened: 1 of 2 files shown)


tests/monotouch-test/CoreFoundation/DispatchDataTest.cs
```diff
@@ -0,0 +1,55 @@
...
+			using (var dd = DispatchData.FromByteBuffer (testData)) {
+				var ddString = Encoding.UTF8.GetString (dd.ToArray ());
+				Assert.AreEqual (testString, ddString);
+			}
+		}
+
+		[Test]
+		public void FromReadOnlySpanTest ()
+		{
+			//var readOnlySpan = new ReadOnlyMemory <byte> (testData);
```
wachaudh: remove comment

wachaudh:
```suggestion
```

src/CoreFoundation/DispatchData.cs
```diff
@@ -88,6 +88,18 @@ public static DispatchData FromBuffer (IntPtr buffer, nuint size)
 			return new DispatchData (dd, owns: true);
 		}
 
+		// create a dd using the data of the span 
+		public static DispatchData FromReadOnlySpan (ReadOnlySpan<byte> content)
+		{
+			unsafe {
+				fixed (byte* ptr = content) {
+					// set the block that will be used to clean the data once we are done with it
+					var dd = dispatch_data_create ((IntPtr)ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
```
wachaudh:
```suggestion
					var dd = dispatch_data_create ((IntPtr) ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
```

src/CoreFoundation/DispatchData.cs
```diff
@@ -88,6 +88,18 @@ public static DispatchData FromBuffer (IntPtr buffer, nuint size)
...
 		}
 
+		// create a dd using the data of the span 
+		public static DispatchData FromReadOnlySpan (ReadOnlySpan<byte> content)
+		{
+			unsafe {
+				fixed (byte* ptr = content) {
+					// set the block that will be used to clean the data once we are done with it
+					var dd = dispatch_data_create ((IntPtr) ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
+					return new DispatchData (dd, owns: true);
```
rolfbjarne: I think this leaks memory, because nobody will release the pointer passed to dispatch_data_create.

src/CoreFoundation/DispatchData.cs
```diff
@@ -88,6 +88,18 @@ public static DispatchData FromBuffer (IntPtr buffer, nuint size)
 			return new DispatchData (dd, owns: true);
 		}
 
+		// create a dd using the data of the span 
+		public static DispatchData FromReadOnlySpan (ReadOnlySpan<byte> content)
+		{
+			unsafe {
+				fixed (byte* ptr = content) {
+					// set the block that will be used to clean the data once we are done with it
+					var dd = dispatch_data_create ((IntPtr) ptr, (nuint) content.Length, IntPtr.Zero, destructor: IntPtr.Zero);
```
prmallic: Could you explain the use of `IntPtr.Zero` for the queue and destructor parameters?

src/CoreFoundation/DispatchData.cs
```diff
@@ -122,6 +134,19 @@ public DispatchData CreateSubrange (nuint offset, nuint size)
...
 		}
+
+		// copies the dispatch data to a managed array
+		public byte [] ToArray ()
+		{
+			IntPtr bufferAddress = IntPtr.Zero;
+			nuint bufferSize = 0;
+			using DispatchData dataCopy = CreateMap (out bufferAddress, out bufferSize);
+
+			byte[] managedArray = new byte[(int) bufferSize];
```
whitneyschmidt: maybe create `bufferSize` as an `int` since it looks like you are casting here and below?

tests/monotouch-test/CoreFoundation/DispatchDataTest.cs
```diff
@@ -0,0 +1,54 @@
...
+		public void FromReadOnlySpanTest ()
+		{
+			var readOnlySpan = new ReadOnlySpan<byte> (testData);
+			using (var dd = DispatchData.FromReadOnlySpan (readOnlySpan)) {
+				var data = dd.ToArray ();
+				var ddString = Encoding.UTF8.GetString (dd.ToArray ());
+				Assert.AreEqual (testString, ddString);
+			}
+		}
+
```
whitneyschmidt: extra line?

src/CoreFoundation/DispatchData.cs
rolfbjarne: Sorry, I phrased that wrong, what I meant is that the pointer is a pointer to stack memory, and as such it will be invalid as soon as this function returns, but the DispatchData still contains it and may use it later.

However, after re-reading the documentation for `dispatch_data_create`, it seems the data the pointer points to is copied internally when specifying DISPATCH_DATA_DESTRUCTOR_DEFAULT as the destructor, and DISPATCH_DATA_DESTRUCTOR_DEFAULT=NULL, which is what you're passing, which means the code is OK. A comment might be nice though for the next people wondering if `dispatch_data_create` keeps the pointer around or not.

src/CoreFoundation/DispatchData.cs
mandel-macaque: The info is in the comment from @rolfbjarne https://github.com/xamarin/xamarin-macios/pull/7690#discussion_r364950115

mandel-macaque:
@rolfbjarne will literally copy what you just said ;)

src/CoreFoundation/DispatchData.cs
mandel-macaque: Cant be done, since the C API needs it to be a nuint :/

mandel-macaque:

PR #7694, opened 2020-01-10
[msbuild] Adjust test to work locally by being less demanding.

rolfbjarne:
These tests were checking if mtouch is called (or not) by checking for the
full installed path to mtouch. When running the tests locally, we're using the
locally built mtouch, and thus the path is different and the tests failed.

Instead check for a smaller substring that is present in both cases.


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/581)

PR #7696, opened 2020-01-10
[DevOps] Add cleanup step to device-tests.yml

VincentDondain:
This was recommended by QA / DDFun because they were running into issues re-installing VSMac on the machines we ran the pipeline on.

### DIFF
--- a/tools/devops/cleanup.sh
+++ b/tools/devops/cleanup.sh
@@ -1,0 +1,6 @@
+sudo rm -rf /Applications/Visual\ Studio*
+rm -rf ~/Library/Caches/VisualStudio
+rm -rf ~/Library/Logs/VisualStudio
+rm -rf ~/Library/VisualStudio
+rm -rf ~/Library/Preferences/Xamarin/
+rm -rf ~/Library/Caches/com.xamarin.provisionator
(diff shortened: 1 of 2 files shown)


tools/devops/cleanup.sh
```diff
@@ -0,0 +1,7 @@
+sudo find  $(dirname $(dirname "$(System.DefaultWorkingDirectory)")) -type d -name 'r*' -not -name $(basename $(dirname "$(System.DefaultWorkingDirectory)")) -maxdepth 1 -exec rm -rf {} +
```
rolfbjarne: What's this line supposed to find and delete?

VincentDondain: First time I read it I thought it was just the source folder but `-name 'r*'`?? I'm not sure so I'm deleting it. I'll ask the person who sent it to me but I think what they really wanted gone was VSM.

VincentDondain: It also doesn't work as is because of the environment variable, I'm not attached to this (:

VincentDondain:
[Simulator tests](https://github.com/xamarin/xamarin-macios/pull/7696#issuecomment-573770212) fail on known HE0038 issue: https://github.com/xamarin/maccore/issues/581.

Device tests issues are irrelevant to merging this PR (CI change) however we can see that the cleanup step of the [DDFun run](https://dev.azure.com/devdiv/DevDiv/_build/index?buildId=3380499) is working as expected.

PR #7698, opened 2020-01-10
[msbuild] Convert project files to sdk-style project files.

rolfbjarne:
This also requires a few adjustments in other places as well, mostly because msbuild places the build output in a different directory now.

### DIFF
--- a/msbuild/.gitignore
+++ b/msbuild/.gitignore
@@ -1,3 +1,4 @@
+.failed-stamp
 .build-stamp
 .stamp-test-xml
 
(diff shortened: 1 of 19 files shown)


kzu:
You can avoid changing the previous output paths with:

```
    <!-- We always target a single framework, so we don't need the added subdir -->
    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
    <!-- We never emit runtime-specific assemblies, so opt-out of this. -->
    <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
```

rolfbjarne:
@kzu: 

> You can avoid changing the previous output paths with:

In most cases it's desired because I'll start building for netstandard2.0 as well pretty soon (in a different PR). But it's still good to know the possibility exists! I wish these build knobs were a bit better documented...

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/581)

PR #7699, opened 2020-01-10
[DevOps] Fix Set GitHub Status

VincentDondain:
In the DDFun pipeline case `cat: tmp.p: No such file or directory` so the `P` variable cannot be set and that's failing vsts-device-tests-set-status.sh

### DIFF
--- a/jenkins/vsts-device-tests-set-status.sh
+++ b/jenkins/vsts-device-tests-set-status.sh
@@ -44,8 +44,6 @@ if test -z "$DEVICE_TYPE"; then
 	DEVICE_TYPE="iOS/tvOS"
 fi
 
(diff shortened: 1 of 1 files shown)


mandel-macaque:
Does this mean we are not getting html reports in DDFun?

VincentDondain:
@mandel-macaque no HTML report for DDFun because we cannot upload via SSH on those bots. You can access the tests results via the build artifacts in the summary tab of VSTS.

VincentDondain:
Simulator tests are green: https://github.com/xamarin/xamarin-macios/pull/7699#issuecomment-573210803
DDFun status is being updated / reported.
Device tests failures are irrelevant to merging this PR (CI change).

PR #7710, opened 2020-01-14
[Localization] mtouch/mmp C#

wachaudh:
Contains localization changes for both mtouch/mmp c# errors.


src/ObjCRuntime/Registrar.cs
```diff
@@ -87,9 +87,9 @@ static class Shared {
 		public static ProductException GetMT4127 (TMethod impl, List<TMethod> ifaceMethods)
 		{
 			var msg = new System.Text.StringBuilder ();
-			msg.Append ("Cannot register more than one interface method for the method '");
+			msg.Append (mtouch.mtouchErrors.MT4127_a);
```
rolfbjarne: This code should be changed to be a single message, for some translations the order of the formatting arguments can change, and that would not be possible to translate like it's implemented now.

tools/mtouch/Assembly.cs
```diff
@@ -90,8 +90,8 @@ HashSet<string> ComputeDependencies (List<Exception> warnings)
 				}
 
 				if (!found) {
-					warnings.Add (new MonoTouchException (3005, false, "The dependency '{0}' of the assembly '{1}' was not found. Please review the project's references.",
-					                                      ar.FullName, AssemblyDefinition.FullName));
+					warnings.Add (new MonoTouchException (3005, false, mtouch.mtouchErrors.MT3005,
+														  ar.FullName, AssemblyDefinition.FullName));
```
rolfbjarne: Minor: you can make this a single line now.

tools/mtouch/MtouchErrors.Designer.cs
```diff
@@ -0,0 +1,2102 @@
...
+
+namespace mtouch {
+    using System;
+    using System.Reflection;
+    
+    
+    [System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [System.Diagnostics.DebuggerNonUserCodeAttribute()]
+    [System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
+    internal class mtouchErrors {
```
rolfbjarne: I think it would be better to choose a namespace and class name that can be shared between mtouch and mmp (this makes it easier to reference an error message from shared code):

```csharp
namespace Xamarin.Bundler {
    internal class Errors {
    }
}
```

wachaudh: I considered this but used separate namespaces/class names because I thought it would be better to separate mtouch and mmp errors. The localized mtouch and mmp errors can already be referenced in shared code, they're separated just under separate namespaces/classes/resx files.

chamons: You can use #if to separate if you want, but having to #if two different namespaces is rather unpleasant.

wachaudh: Using #if isn't necessary. Shared errors can be directly referenced between mtouch and mmp the way both projects are currently set up.

src/ObjCRuntime/Registrar.cs
wachaudh: do the order of arguments being appended in the loop matter?

rolfbjarne: All the English text should be contained within a single string. This means that the text here has to change (the enumeration and string concatenation loop below doesn't work), maybe by creating multiple exceptions:

    MT4127: Cannot register more than one interface method for the method '{0}'. This method implements {0} interface methods.
    MT4???: The method '{0}' implements the interface method '{1}'.
    MT4???: The method '{0}' implements the interface method '{1}'.
    MT4???: The method '{0}' implements the interface method '{1}'.

tools/mtouch/mtouch.Designer.cs
```diff
@@ -0,0 +1,2090 @@
...
+
+namespace Errors {
+    using System;
+    using System.Reflection;
+    
+    
+    [System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [System.Diagnostics.DebuggerNonUserCodeAttribute()]
+    [System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
+    public class mtouch {
```
rolfbjarne: This must be an `internal` class, since it ends up in the class library we ship (Xamarin.iOS.dll). 

Also we end up adding this whole file to the class library, when most of it is not referenced. I think we should come up with a different solution for the code that's included the class library, one that doesn't end up adding all these properties to the library.

One solution that comes to mind is to add extra metadata to the entries in the resx to mark those error messages used in Xamarin.iOS.dll, and implement a custom generator to generate the corresponding code.

spouliot: hmm... maybe that's a step too far ? i.e. does it make sense to translate *runtime* errors ? my understanding is that we were asked to localize build errors/warnings. c.c. @chamons  

Even if those errors share the same number space (no dupes) they are not `mtouch/mmp` errors, they are runtime errors - and others are not localized (e.g. all exception messages).

If we want them localized then it might be better to do this into a separate PR since the process will be different - at least for building (e.g. some extra tool like @rolfbjarne suggested). Strings are large and we do not want to include unneeded ones into final applications, let alone several translations for [un]needed ones :-)

rolfbjarne: @spouliot the thing is that the code is shared between mtouch/mmp and the runtime. That works fine while using string constants, but it gets more complicated now. A potential solution would be an adaptation of the previous solution I suggested, where the custom generator I proposed would write out properties that just return the corresponding string constant (so no need to adapt the src/ build to consume the resx).

wachaudh: wouldn't a secondary resx file that just contains errors  ONLY used by src/ObjCRuntime/Registrar.c and src/ObjCRuntime/RuntimeOptions.cs  (and other runtime code) work? There might be some tiny duplication of errors but it's a lot simpler then writing an additional tool.

spouliot: ok, I thought it also contained the parts that are runtime only...

we do have a `#if MTOUCH || MMP` in a few places... so we could have the platform assemblies ship with a subset (generated or manual)

not ideal... but, at least for now, the autogenerated file is part of the repo :|

tools/linker/CoreTypeMapStep.cs
```diff
@@ -142,6 +142,7 @@ bool RequiresDynamicRegistrar (AssemblyDefinition assembly, bool warnIfRequired)
 
 		void Show2107 (AssemblyDefinition assembly, MemberReference mr)
 		{
+			//todo: localize mmp error 2107
```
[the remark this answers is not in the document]
wachaudh: It's an mmp error, will be added in the mmp PR after this is merged. I'll remove the comment through

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
```diff
@@ -61,6 +61,7 @@ protected override void Process (MethodDefinition method)
 					} else {
 						msg = $"The method {method.DeclaringType.FullName}.{method.Name} contains a '{eh.HandlerType}' exception clause, which is currently not supported when compiling for bitcode. This method will throw an exception if called.";
 					}
+					//todo: localize this error
```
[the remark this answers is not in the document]
wachaudh: Also mmp related. I'll remove the comment.

tools/mtouch/MtouchErrors.Designer.cs
spouliot: Not sure about _confusion_ but I agree it would be easier for _us_ :)

Now we're doing this for our users... so a single file means error duplication or code-overloads is much harder to get thru and confuse our beloved developers.

it also simplify the translator works which in turns could become confusion sources if one translation (for mtouch) differs from another (for mmp).

tools/common/Application.cs
```diff
@@ -229,7 +229,7 @@ public static bool IsUptodate (IEnumerable<string> sources, IEnumerable<string>
 			string max_s = null;
 
 			if (sources.Count () == 0 || targets.Count () == 0)
-				ErrorHelper.Error (1013, "Dependency tracking error: no files to compare. Please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new with a test case.");
+				ErrorHelper.Error (1013, Errors.mtouch.MT1013);
```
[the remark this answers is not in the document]
wachaudh: So just too clarify, we want just one resx (and designer.cs) file that contains all mtouch, mmp and shared errors? And too reference errors in common code as Errors.MX.... ?

tools/common/Target.cs
```diff
@@ -145,7 +145,7 @@ public void GatherFrameworks ()
 			}
 
 			if (asm == null)
-				throw ErrorHelper.CreateError (99, $"Internal error: could not find the product assembly {Driver.GetProductAssembly (App)} in the list of assemblies referenced by the executable. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+				throw ErrorHelper.CreateError (99, Errors.mtouch.MT0099_E, Driver.GetProductAssembly (App));
```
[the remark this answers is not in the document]
wachaudh: Not all existing error 99s follow this pattern that require a stack trace and additional data. For example, `99, "Internal error: Can't rebuild a task that hasn't completed. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new)."` Should these be changed to new error codes?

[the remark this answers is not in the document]
rolfbjarne: I think it's fine to leave any unique parts of MT0099 untranslated. MT0099 is used for impossible conditions, consistency checks, and other error conditions we couldn't predict, which means it's hard to provide a actionable error message. This is why we just ask the user to file a bug: it's the only thing he really needs to know.

This means that all MT0099 errors should become something like `Internal error: {0}. Please file a bug ...`, which should be translated, and the `{0}` argument should not.

tools/mtouch/mtouch.Designer.cs
rolfbjarne: A secondary resx file isn't a bad idea, but the split could be different: one with errors for mtouch+mmp, and one with errors for mtouch+mmp+runtime. That way there's no duplication. We'd still have to fix the src/ build to embed the resx in the assembly, and we'd have to check the size increase (if any) to see if it's acceptable.

Long-term it might be a good idea to move all user-visible strings from the assemblies themselves and into external resource files, which might help decrease the final executable size (unless the AOT compiler already puts string data in the .aotdata files).

@spouliot what do you think?

tools/common/Target.cs
```diff
@@ -421,7 +421,7 @@ internal string GenerateReferencingSource (string reference_m, IEnumerable<Symbo
 					sb.AppendLine ($"\tvalue = [{symbol.ObjectiveCName} class];");
 					break;
 				default:
-					throw ErrorHelper.CreateError (99, $"Internal error: invalid symbol type {symbol.Type} for {symbol.Name}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+						throw ErrorHelper.CreateError(99, Errors.MT0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```
wachaudh:
```suggestion
					throw ErrorHelper.CreateError(99, Errors.MT0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```

tools/mtouch/Target.cs
```diff
@@ -251,8 +250,8 @@ public void LinkWithDynamicLibrary (Abi abi, string path)
 						// When using simlauncher, we copy the executable directly to the target directory.
 						// When not using the simlauncher, but still building for the simulator, we write the executable to a arch-specific app directory (if building for both 32-bit and 64-bit), or just the app directory (if building for a single architecture)
 						if (Abis.Count != 1)
-							throw ErrorHelper.CreateError (99, $"Internal error: expected exactly one abi for a simulator architecture, found: {string.Join (", ", Abis.Select ((v) => v.ToString ()))}");
-						executables.Add (Abis [0], Path.Combine (TargetDirectory, App.ExecutableName));
+							throw ErrorHelper.CreateError (99, Errors.MT0099, $"expected exactly one abi for a simulator architecture, found: {string.Join(", ", Abis.Select((v) => v.ToString()))}");
+							executables.Add (Abis [0], Path.Combine (TargetDirectory, App.ExecutableName));
```
wachaudh:
```suggestion
						executables.Add (Abis [0], Path.Combine (TargetDirectory, App.ExecutableName));
```

src/ObjCRuntime/Registrar.cs
```diff
@@ -1543,6 +1544,7 @@ void VerifyTypeInSDK (ref List<Exception> exceptions, TType type, ObjCMethod par
 					if (attrib.Version <= sdk)
 						break;
 
+					//todo: localize msg
 					string msg = "The type '{0}' (used as {1} {2}) is not available in {3} {4} (it was introduced in {3} {5}){6} Please build with a newer {3} SDK (usually done by using the most recent version of Xcode).";
```
mandel-macaque: Please create an issue, that way we won't forget.

tools/common/Target.cs
```diff
@@ -404,7 +404,7 @@ internal string GenerateReferencingSource (string reference_m, IEnumerable<Symbo
 					sb.AppendLine ($"@interface {symbol.ObjectiveCName} : NSObject @end");
 					break;
 				default:
-					throw ErrorHelper.CreateError (99, $"Internal error: invalid symbol type {symbol.Type} for {symbol.Name}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+						throw ErrorHelper.CreateError(99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```
mandel-macaque: Mono style:
```suggestion
						throw ErrorHelper.CreateError (99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```

tools/common/Target.cs
```diff
@@ -421,7 +421,7 @@ internal string GenerateReferencingSource (string reference_m, IEnumerable<Symbo
 					sb.AppendLine ($"\tvalue = [{symbol.ObjectiveCName} class];");
 					break;
 				default:
-					throw ErrorHelper.CreateError (99, $"Internal error: invalid symbol type {symbol.Type} for {symbol.Name}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+					throw ErrorHelper.CreateError(99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```
mandel-macaque: Mono style:
```suggestion
					throw ErrorHelper.CreateError (99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```

tools/linker/CoreTypeMapStep.cs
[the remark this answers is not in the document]
mandel-macaque: Add an issue so that we do not forget.

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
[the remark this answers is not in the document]
mandel-macaque: Same, add an issue so we do not forget. One for all should be ok if you have a list or point to the file lines.

tools/mmp/driver.cs
```diff
@@ -453,13 +453,13 @@ static void Main2 (string [] args)
...
-				throw new MonoMacException (1404, true, "Target framework '{0}' is invalid.", userTargetFramework);
+				throw new MonoMacException (1404, true, Errors.MM1404, userTargetFramework);
 
 			if (Registrar == RegistrarMode.PartialStatic && App.LinkMode != LinkMode.None)
-				throw new MonoMacException (2110, true, "Xamarin.Mac 'Partial Static' registrar does not support linking. Disable linking or use another registrar mode.");
+				throw new MonoMacException (2110, true, Errors.MM2110);
 
 			if (IsClassic)
-				throw ErrorHelper.CreateError (143, "Projects using the Classic API are not supported anymore. Please migrate the project to the Unified API.");
+				throw ErrorHelper.CreateError(143, Errors.MM0143);
```
mandel-macaque: Mono style:
```suggestion
				throw ErrorHelper.CreateError (143, Errors.MM0143);
```

tools/mtouch/Target.cs
```diff
@@ -772,7 +770,7 @@ public void ManagedLink ()
 
 						var ad = output_assemblies.SingleOrDefault ((AssemblyDefinition v) => v.Name.Name == next);
 						if (ad == null)
-							throw ErrorHelper.CreateError (99, "Internal error {0}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).", $"The assembly {next} was referenced by another assembly, but at the same time linked out by the linker.");
+							throw ErrorHelper.CreateError(99, Errors.MX0099, $"The assembly {next} was referenced by another assembly, but at the same time linked out by the linker");
```
mandel-macaque: Mono style:
```suggestion
							throw ErrorHelper.CreateError (99, Errors.MX0099, $"The assembly {next} was referenced by another assembly, but at the same time linked out by the linker");
```

src/ObjCRuntime/Registrar.cs
wachaudh: I've added all mmp errors to this PR now including this and the other ones you commented on.

tools/mtouch/Application.cs
```diff
@@ -1394,25 +1393,25 @@ void Initialize ()
...
 					if (DeploymentTarget < new Version (2, 0))
-						throw ErrorHelper.CreateError (65, "Xamarin.iOS only supports embedded frameworks when deployment target is at least 2.0 (current deployment target: '{0}'; embedded frameworks: '{1}')", DeploymentTarget, string.Join (", ", Frameworks.ToArray ()));
+						throw ErrorHelper.CreateError (65, Errors.MT0065_A, DeploymentTarget, string.Join (", ", Frameworks.ToArray ()));
 					break;
 				case ApplePlatform.TVOS:
 					// All versions of tvOS support extensions
 					break;
 				default:
-					throw ErrorHelper.CreateError (71, "Unknown platform: {0}. This usually indicates a bug in Xamarin.iOS; please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new with a test case.", Platform);
+				throw ErrorHelper.CreateError (71, Errors.MX0071, Platform, "Xamarin.iOS");
```
wachaudh:
```suggestion
					throw ErrorHelper.CreateError (71, Errors.MX0071, Platform, "Xamarin.iOS");
```

tools/mtouch/Application.cs
```diff
@@ -1951,24 +1950,18 @@ public static void ProcessNativeLinkerOutput (Target target, string output, IEnu
 
 						var symbol = line.Replace (", referenced from:", "").Trim ('\"', ' ');
 						if (symbol.StartsWith ("_OBJC_CLASS_$_", StringComparison.Ordinal)) {
-							errors.Add (new MonoTouchException (5211, error, 
-																"Native linking failed, undefined Objective-C class: {0}. The symbol '{1}' could not be found in any of the libraries or frameworks linked with your application.",
-							                                    symbol.Replace ("_OBJC_CLASS_$_", ""), symbol));
+							errors.Add (new MonoTouchException (5211, error, Errors.MT5211,
+																symbol.Replace ("_OBJC_CLASS_$_", ""), symbol));
```
wachaudh: style

tools/mtouch/Application.cs
```diff
@@ -1951,24 +1950,18 @@ public static void ProcessNativeLinkerOutput (Target target, string output, IEnu
...
-									"Please verify that all the necessary frameworks have been referenced and native libraries linked.",
+								errors.Add (new MonoTouchException (5214, error, Errors.MT5214,
 									symbol, member.DeclaringType.FullName, member.Name));
 							} else {
-								errors.Add (new MonoTouchException (5210, error, 
-							                                    "Native linking failed, undefined symbol: {0}. " +
-																"Please verify that all the necessary frameworks have been referenced and native libraries are properly linked in.",
-							                                    symbol));
+								errors.Add (new MonoTouchException (5210, error, Errors.MT5210,
+																symbol));
```
wachaudh: style

tools/mtouch/mtouch.cs
```diff
@@ -1656,6 +1656,7 @@ public static bool IsSymlink (string file)
 			stat buf;
 			var rv = lstat (file, out buf);
 			if (rv != 0)
+				//todo: fix this
 				throw new Exception (string.Format ("Could not lstat '{0}': {1}", file, Marshal.GetLastWin32Error ()));
```
wachaudh: forgot this

src/ObjCRuntime/Registrar.cs
```diff
@@ -293,7 +294,7 @@ public void VerifyAdoptedProtocolsNames (ref List<Exception> exceptions)
 							str = "}}";
 							ap = ap.Insert (idx, "}");
 						}
-						AddException (ref exceptions, new ProductException (4177, true, $"The 'ProtocolType' parameter of the 'Adopts' attribute used in class '{Registrar.GetTypeFullName (Type)}' contains an invalid character. Value used: '{ap}' Invalid Char: '{str}'"));
+						AddException (ref exceptions, new ProductException (4177, true, $"The 'ProtocolType' parameter of the 'Adopts' attribute used in class '{Registrar.GetTypeFullName(Type)}' contains an invalid character. Value used: '{ap}' Invalid Char: '{str}'"));
```
wachaudh: forgot this

src/ObjCRuntime/Registrar.cs
```diff
@@ -1508,7 +1509,7 @@ bool VerifyNonGenericMethod (ref List<Exception> exceptions, TType declaringType
 				return true;
 
 			AddException (ref exceptions, CreateException (4113, method, 
-				"The registrar found a generic method: '{0}'. Exporting generic methods is not supported, and will lead to random behavior and/or crashes", 
+				Xamarin.Bundler.Errors.MT4113,
```
wachaudh: style

src/ObjCRuntime/Registrar.cs
```diff
@@ -2191,7 +2192,7 @@ ObjCType RegisterTypeUnsafe (TType type, ref List<Exception> exceptions)
 					if (ca != null) {
 						if (!IsINativeObject (GetPropertyType (property))) {
 							AddException (ref exceptions, CreateException (4139, property,
-								"The registrar cannot marshal the property type '{0}' of the property '{1}.{2}'. Properties with the [Connect] attribute must have a property type of NSObject (or a subclass of NSObject).",
+								Xamarin.Bundler.Errors.MT4139,
 								GetTypeFullName (GetPropertyType (property)), GetTypeFullName (type), GetPropertyName (property)));
 							continue;
```
wachaudh: style

tools/common/Application.cs
```diff
@@ -453,13 +453,13 @@ public void RunRegistrar ()
 					resolvedAssemblies.Add (rootName, lastAssembly);
 					Driver.Log (3, "Loaded {0}", lastAssembly.MainModule.FileName);
 				} catch (Exception ex) {
-					ErrorHelper.Warning (9, ex, "Error while loading assemblies: {0}: {1}", rootName, ex.Message);
+					ErrorHelper.Warning(9, ex, Errors.MT0009_A, rootName, ex.Message);
```
wachaudh:
```suggestion
					ErrorHelper.Warning (9, ex, Errors.MT0009_A, rootName, ex.Message);
```

tools/common/Assembly.cs
```diff
@@ -338,9 +338,7 @@ string ExtractNativeLibrary (AssemblyDefinition assembly, NativeReferenceMetadat
 			}
 
 			if (!File.Exists (path))
-				ErrorHelper.Warning (1302, "Could not extract the native library '{0}' from '{1}'. " +
-				"Please ensure the native library was properly embedded in the managed assembly " +
-				"(if the assembly was built using a binding project, the native library must be included in the project, and its Build Action must be 'ObjcBindingNativeLibrary').",
+				ErrorHelper.Warning (1302, Errors.MT1302,
 					metadata.LibraryName, path);
```
wachaudh: style

tools/common/Assembly.cs
```diff
@@ -360,16 +358,14 @@ string ExtractFramework (AssemblyDefinition assembly, NativeReferenceMetadata me
 			}
 
 			if (!File.Exists (zipPath)) {
-				ErrorHelper.Warning (1302, "Could not extract the native framework '{0}' from '{1}'. " +
-					"Please ensure the native framework was properly embedded in the managed assembly " +
-					"(if the assembly was built using a binding project, the native framework must be included in the project, and its Build Action must be 'ObjcBindingNativeFramework').",
+				ErrorHelper.Warning (1302, Errors.MT1302,
 					metadata.LibraryName, zipPath);
```
wachaudh: style

tools/common/Target.cs
```diff
@@ -404,7 +404,7 @@ internal string GenerateReferencingSource (string reference_m, IEnumerable<Symbo
 					sb.AppendLine ($"@interface {symbol.ObjectiveCName} : NSObject @end");
 					break;
 				default:
-					throw ErrorHelper.CreateError (99, $"Internal error: invalid symbol type {symbol.Type} for {symbol.Name}. Please file a bug report with a test case (https://github.com/xamarin/xamarin-macios/issues/new).");
+						throw ErrorHelper.CreateError (99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```
wachaudh:
```suggestion
					throw ErrorHelper.CreateError (99, Errors.MX0099, $"invalid symbol type {symbol.Type} for symbol {symbol.Name}");
```

tools/linker/CoreOptimizeGeneratedCode.cs
```diff
@@ -828,6 +828,7 @@ int ProcessSetupBlock (MethodDefinition caller, Instruction ins)
 				while (prev.OpCode.Code == Code.Nop)
 					prev = prev.Previous; // Skip any nops.
 				if (prev.OpCode.StackBehaviourPush != StackBehaviour.Push1) {
+					//todo: localize mmp error 2106
 					ErrorHelper.Show (ErrorHelper.CreateWarning (Options.Application, 2106, caller, ins, "Could not optimize the call to BlockLiteral.{2} in {0} at offset {1} because the previous instruction was unexpected ({3})", caller, ins.Offset, mr.Name, prev));
```
wachaudh: forgot this

tools/mmp/driver.cs
```diff
@@ -474,7 +474,7 @@ static void Main2 (string [] args)
 			if (IsUnifiedFullXamMacFramework)
 				IsUnifiedCount++;
 			if (IsUnifiedCount != 1)
-				throw ErrorHelper.CreateError (99,"Internal error: IsClassic/IsUnified/IsUnifiedMobile/IsUnifiedFullSystemFramework/IsUnifiedFullXamMacFramework logic regression");
+				throw ErrorHelper.CreateError (99, Errors.MX0099 ,"IsClassic / IsUnified / IsUnifiedMobile / IsUnifiedFullSystemFramework / IsUnifiedFullXamMacFramework logic regression");
```
wachaudh:
```suggestion
				throw ErrorHelper.CreateError (99, Errors.MX0099 ,"IsClassic/IsUnified/IsUnifiedMobile/IsUnifiedFullSystemFramework/IsUnifiedFullXamMacFramework logic regression");
```

tools/mmp/driver.cs
```diff
@@ -1562,8 +1558,8 @@ static void CopyDependencies (IDictionary<string, List<MethodDefinition>> librar
 				if (sb.Count > 0) {
 					sb.Add (library);
 					int ret = XcodeRun ("install_name_tool", sb);
-					if (ret != 0)
-						throw new MonoMacException (5310, true, "install_name_tool failed with an error code '{0}'. Check build log for details.", ret);
+					if (ret != 0)
```
wachaudh:
```suggestion
					if (ret != 0)
```

tools/mtouch/Application.cs
```diff
@@ -427,13 +427,13 @@ void SelectAssemblyBuildTargets ()
...
-							throw ErrorHelper.CreateError (102, "The assemblies '{0}' and '{1}' have the same target name ('{2}'), but different targets ('{3}' and '{4}').",
+							throw ErrorHelper.CreateError (102, Errors.MT0102,
 														   assemblies [0].Identity, assemblies [1].Identity, assemblies [0].BuildTargetName, assemblies [0].BuildTarget, assemblies [1].BuildTarget);
 					}
 
 					// Check that static objects must consist of only one assembly
 					if (assemblies.Length != 1 && assemblies [0].BuildTarget == AssemblyBuildTarget.StaticObject)
-						throw ErrorHelper.CreateError (103, "The static object '{0}' contains more than one assembly ('{1}'), but each static object must correspond with exactly one assembly.",
+						throw ErrorHelper.CreateError (103, Errors.MT0103,
 													   assemblies [0].BuildTargetName, string.Join ("', '", assemblies.Select ((a) => a.Identity).ToArray ()));
```
wachaudh: style

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,56 @@ void DetectCodeSharing ()
...
 				if (!CompareLists (AotOtherArguments, appex.AotOtherArguments)) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the other arguments to the AOT compiler are different between the container app ({StringUtils.FormatArguments (AotOtherArguments)}) and the extension ({StringUtils.FormatArguments (appex.AotOtherArguments)}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_e, StringUtils.FormatArguments(AotOtherArguments), StringUtils.FormatArguments(appex.AotOtherArguments)));
 					continue;
 				}
 
 				if (IsLLVM != appex.IsLLVM) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"LLVM is not enabled or disabled in both the container app ({IsLLVM}) and the extension ({appex.IsLLVM}).");
-					continue;
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_f, IsLLVM, appex.IsLLVM));
```
wachaudh: where did the `continue;` go? fix this

tools/linker/CoreTypeMapStep.cs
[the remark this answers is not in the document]
wachaudh: Added this error all of with all other mmp errors.

tools/linker/MonoTouch.Tuner/RemoveBitcodeIncompatibleCodeStep.cs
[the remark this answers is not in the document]
wachaudh: Added this error and all other mmp errors to this PR.

tools/mtouch/mtouch.csproj
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+  <Import Project="..\..\packages\XliffTasks.1.0.0-beta.20059.1\build\XliffTasks.props" Condition="Exists('..\..\packages\XliffTasks.1.0.0-beta.20059.1\build\XliffTasks.props')" />
+  <Import Project="..\..\packages\XliffTasks.1.0.0-beta.19606.1\build\XliffTasks.props" Condition="Exists('..\..\packages\XliffTasks.1.0.0-beta.19606.1\build\XliffTasks.props')" />
```
rolfbjarne: Why two import statements? Isn't one enough?

src/ObjCRuntime/Registrar.cs
```diff
@@ -87,9 +88,7 @@ static class Shared {
 		public static ProductException GetMT4127 (TMethod impl, List<TMethod> ifaceMethods)
 		{
 			var msg = new System.Text.StringBuilder ();
-			msg.Append ("Cannot register more than one interface method for the method '");
-			msg.Append (impl.DeclaringType.FullName).Append ('.').Append (impl.Name);
-			msg.Append ("' (which is implementing '");
+			msg.Append (String.Format(Errors.MT4127, impl.DeclaringType.FullName, impl.Name));
 			for (int i = 0; i < ifaceMethods.Count; i++) {
 				if (i > 0)
 					msg.Append (i < ifaceMethods.Count - 1 ? "', '" : "' and '");
```
rolfbjarne: This logic needs to be reworked/refactored a little bit so that we don't end up adding English text (`and`) to translated messages.

tools/mtouch/errors.resx
```diff
@@ -0,0 +1,2623 @@
...
+	
+	<data name="MM2106_B" xml:space="preserve">
+		<value>Could not optimize the call to BlockLiteral.{2} in {0} because the type of the value passed as the first argument (the trampoline) is {1}, which makes it impossible to compute the block signature.
+		</value>
+		<comment>
+		</comment>
+	</data>
+	
+	<data name="MM2106_C" xml:space="preserve">
+		<value>Could not optimize the call to BlockLiteral.SetupBlock in {0} at offset {1} because no [UserDelegateType] attribute could be found o
```
rolfbjarne: I think the messages should be sorted by number only (so that you get the MT*, MM*, and MX* messages all together), because a very common question is `What's the next available error number`? If the messages are split in three sections, we'd have to look in three sections every time. Half the time (at least) someone would forget, and try to use a number that's already in use. Having all the messages ordered by number alone should make that much less probable.

tools/mtouch/mtouch.csproj
wachaudh: Not sure, I added the nuget using VSMac. I'll fix it.

src/ObjCRuntime/Registrar.cs
wachaudh: The full message is: 

`Cannot register more than one interface method for the method '{0}' (which is implementing '{1}'  and '{2}' and '{3}' and ...  '{x}').`

The for loop is creating this part:

`'{1}'  and '{2}' and '{3}' and ...  '{x}'`

Would changing `and`s to commas fix this? 

So the new message would be:

`Cannot register more than one interface method for the method '{0}' (which is implementing: '{1}', '{2}', '{3}', ...  '{x}').`

rolfbjarne: I don't know if that would work with languages that aren't left-to-right (Chinese, Arabic, etc.)

My suggestion would be to emit multiple errors:

> MT4127: Cannot register more than one interface method for the method '{0}'.
> MT41??: The method '{0}' is implementing '{1}'.
> MT41??: The method '{0}' is implementing '{1}'.
> MT41??: The method '{0}' is implementing '{1}'.
> ...

wachaudh: I've implemented the multiple errors solution. All of the error codes are `MT4127`, not sure if we want to change `The method '{0}' is implementing '{1}'` to different error codes since they're related.

tools/mtouch/errors.resx
wachaudh: There are some cases where the same error code being used for different error messages in mtouch and mmp. Do we want to handle these differently?

Example:

`MT0001: '-devname' was provided without any device-specific action`

`MM0001: This version of Xamarin.Mac requires Mono {0} (the current Mono version is {1}). Please update the Mono.framework from http://mono-project.com/Downloads`

rolfbjarne: No, just put them next to each other in the file.

src/ObjCRuntime/Registrar.cs
rolfbjarne: Yes, I think it's best to use a different error code since the text is different.

src/ObjCRuntime/Registrar.cs
```diff
@@ -2316,7 +2313,8 @@ ObjCType RegisterTypeUnsafe (TType type, ref List<Exception> exceptions)
 					List<TMethod> impls;
 					if (method_map != null && method_map.TryGetValue (method, out impls)) {
 						if (impls.Count != 1) {
-							AddException (ref exceptions, Shared.GetMT4127 (method, impls));
+							foreach (var err in Shared.GetMT4127(method, impls))
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1047,21 +1047,21 @@ void DetectCodeSharing ()
 				// We can probably lift this requirement (at least partially) at some point,
 				// but for now it makes our code simpler.
 				if (assembly_build_targets.Count != appex.assembly_build_targets.Count) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the --assembly-build-target options are different between the container app ({FormatAssemblyBuildTargets ()}) and the extension ({appex.FormatAssemblyBuildTargets ()}).");
+					ErrorHelper.Warning (113, Errors.MT0113,appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_b, FormatAssemblyBuildTargets(), appex.FormatAssemblyBuildTargets()));
```
spouliot: minor: missing spaces after `(` (multiples) and after `,`

tools/mtouch/Application.cs
```diff
@@ -1038,7 +1038,7 @@ void DetectCodeSharing ()
 				}
 
 				if (BitCodeMode != appex.BitCodeMode) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the bitcode options differ between the container app ({appex.BitCodeMode}) and the extension ({BitCodeMode}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_a, appex.BitCodeMode, BitCodeMode));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1047,21 +1047,21 @@ void DetectCodeSharing ()
...
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the --assembly-build-target options are different between the container app ({FormatAssemblyBuildTargets ()}) and the extension ({appex.FormatAssemblyBuildTargets ()}).");
+					ErrorHelper.Warning (113, Errors.MT0113,appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_b, FormatAssemblyBuildTargets(), appex.FormatAssemblyBuildTargets()));
 					continue;
 				}
 
 				foreach (var key in assembly_build_targets.Keys) {
 					Tuple<AssemblyBuildTarget, string> appex_value;
 					if (!appex.assembly_build_targets.TryGetValue (key, out appex_value)) {
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the --assembly-build-target options are different between the container app ({FormatAssemblyBuildTargets ()}) and the extension ({appex.FormatAssemblyBuildTargets ()}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_b, FormatAssemblyBuildTargets(), appex.FormatAssemblyBuildTargets()));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1047,21 +1047,21 @@ void DetectCodeSharing ()
...
 						applicable = false;
 						break;
 					}
 
 					var value = assembly_build_targets [key];
 					if (value.Item1 != appex_value.Item1 || value.Item2 != appex_value.Item2) {
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the --assembly-build-target options are different between the container app ({FormatAssemblyBuildTargets ()}) and the extension ({appex.FormatAssemblyBuildTargets ()}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_b, FormatAssemblyBuildTargets(), appex.FormatAssemblyBuildTargets()));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
 				
 				// No I18N assemblies can be included
 				if (appex.I18n != Mono.Linker.I18nAssemblies.None) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the I18N assemblies are different between the container app ({I18n}) and the extension ({appex.I18n}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_c, I18n, appex.I18n));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 					continue;
 				}
 
 				// All arguments to the AOT compiler must be identical
 				if (AotArguments != appex.AotArguments) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the arguments to the AOT compiler are different between the container app ({AotArguments}) and the extension ({appex.AotArguments}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_d, AotArguments, appex.AotArguments));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 					continue;
 				}
 
 				if (!CompareLists (AotOtherArguments, appex.AotOtherArguments)) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the other arguments to the AOT compiler are different between the container app ({StringUtils.FormatArguments (AotOtherArguments)}) and the extension ({StringUtils.FormatArguments (appex.AotOtherArguments)}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_e, StringUtils.FormatArguments(AotOtherArguments), StringUtils.FormatArguments(appex.AotOtherArguments)));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 					continue;
 				}
 
 				if (IsLLVM != appex.IsLLVM) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"LLVM is not enabled or disabled in both the container app ({IsLLVM}) and the extension ({appex.IsLLVM}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_f, IsLLVM, appex.IsLLVM));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the managed linker settings are different between the container app ({LinkMode}) and the extension ({appex.LinkMode}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_g, LinkMode, appex.LinkMode));
 					continue;
 				}
 
 				if (LinkMode != LinkMode.None) {
 					var linkskipped_same = !LinkSkipped.Except (appex.LinkSkipped).Any () && !appex.LinkSkipped.Except (LinkSkipped).Any ();
 					if (!linkskipped_same) {
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the skipped assemblies for the managed linker are different between the container app ({string.Join (", ", LinkSkipped)}) and the extension ({string.Join (", ", appex.LinkSkipped)}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_h, string.Join(", ", LinkSkipped), string.Join(", ", appex.LinkSkipped)));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 						continue;
 					}
 
 					if (Definitions.Count > 0) {
-						ErrorHelper.Warning (112, "Native code sharing has been disabled because {0}", $"the container app has custom xml definitions for the managed linker ({string.Join (", ", Definitions)}).");
+						ErrorHelper.Warning (112, Errors.MT0112, String.Format(Xamarin.Bundler.Errors.MT0112_c, string.Join(", ", Definitions)));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 						continue;
 					}
 
 					if (appex.Definitions.Count > 0) {
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the extension has custom xml definitions for the managed linker ({string.Join (", ", appex.Definitions)}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_i, string.Join(", ", appex.Definitions)));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 						continue;
 					}
 				}
 
 				if (UseInterpreter != appex.UseInterpreter) {
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the interpreter settings are different between the container app ({(UseInterpreter ? "Enabled" : "Disabled")}) and the extension ({(appex.UseInterpreter ? "Enabled" : "Disabled")}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_j, (UseInterpreter ? "Enabled" : "Disabled"), (appex.UseInterpreter ? "Enabled" : "Disabled")));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1072,57 +1072,57 @@ void DetectCodeSharing ()
...
 					continue;
 				} else if (UseInterpreter) {
 					var appAssemblies = new HashSet<string> (InterpretedAssemblies);
 					var appexAssemblies = new HashSet<string> (appex.InterpretedAssemblies);
 					if (!appAssemblies.SetEquals (appexAssemblies)) {
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the interpreted assemblies are different between the container app ({(InterpretedAssemblies.Count == 0 ? "all assemblies" : string.Join (", ", InterpretedAssemblies))}) and the extension ({(appex.InterpretedAssemblies.Count == 0 ? "all assemblies" : string.Join (", ", appex.InterpretedAssemblies))}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_k, (InterpretedAssemblies.Count == 0 ? "all assemblies" : string.Join(", ", InterpretedAssemblies)), (appex.InterpretedAssemblies.Count == 0 ? "all assemblies" : string.Join(", ", appex.InterpretedAssemblies))));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1132,12 +1132,12 @@ void DetectCodeSharing ()
 					var matching = abis.FirstOrDefault ((v) => (v & Abi.ArchMask) == (abi & Abi.ArchMask));
 					if (matching == Abi.None) {
 						// Example: extension has arm64+armv7, while the main app has only arm64.
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the container app does not build for the ABI {abi} (while the extension is building for this ABI).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_l, abi));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1132,12 +1132,12 @@ void DetectCodeSharing ()
...
 						applicable = false;
 						break;
 					} else if (matching != abi) {
 						// Example: extension has arm64+llvm, while the main app has only arm64.
-						ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the container app is building for the ABI {matching}, which is not compatible with the extension's ABI ({abi}).");
+						ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_m, matching, abi));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1150,7 +1150,7 @@ void DetectCodeSharing ()
 							return "default";
 						return v.Value ? "true" : "false";
 					};
-					ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the remove-dynamic-registrar optimization differ between the container app ({bool_tostr (appex.Optimizations.RemoveDynamicRegistrar)}) and the extension ({bool_tostr (Optimizations.RemoveDynamicRegistrar)}).");
+					ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_n, bool_tostr(appex.Optimizations.RemoveDynamicRegistrar), bool_tostr(Optimizations.RemoveDynamicRegistrar)));
```
spouliot: minor: missing space after `(`

tools/mtouch/Application.cs
```diff
@@ -1165,7 +1165,7 @@ void DetectCodeSharing ()
 							continue; // appex references an assembly the main app doesn't. This is fine.
 						if (asm.FullPath != kvp.Value.FullPath && !Cache.CompareFiles (asm.FullPath, kvp.Value.FullPath, true)) {
 							applicable = false; // app references an assembly with the same name as the main app, but from a different location and not identical. This is not fine.
-							ErrorHelper.Warning (113, "Native code sharing has been disabled for the extension '{0}' because {1}", appex.Name, $"the container app is referencing the assembly '{asm.Identity}' from '{asm.FullPath}', while the extension references a different version from '{kvp.Value.FullPath}'.");
+							ErrorHelper.Warning (113, Errors.MT0113, appex.Name, String.Format(Xamarin.Bundler.Errors.MT0113_n, asm.Identity, asm.FullPath));
```
spouliot: minor: missing space after `(`

tools/mtouch/mtouch.csproj
spouliot: there are still two of them
please don't make comments as **resolved** since it means we can't double check if any action was done 
iow let the original commenter resolve his comments, based on your changes

tools/mtouch/mtouch.csproj
```diff
@@ -452,8 +461,20 @@
...
+      <CustomToolNamespace>Xamarin.Bundler</CustomToolNamespace>
+      <LogicalName>Errors.mtouch.resources</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
   <Target Name="AfterBuild">
     <!-- This makes sure that just building the csproj will install the updated mtouch.exe, so that tests get it without having to 'make mtouch' manually -->
     <Exec Command="make mtouch" />
   </Target>
+  <Import Project="..\..\packages\XliffTasks.1.0.0-beta.19606.1\build\XliffTasks.targets" Condition="Exists('..\..\packages\XliffTasks.1.0.0-beta.19606.1\build\XliffTasks.targets')" />
+  <Import Project="..\..\packages\XliffTasks.1.0.0-beta.20059.1\build\XliffTasks.targets" Condition="Exists('..\..\packages\XliffTasks.1.0.0-beta.20059.1\build\XliffTasks.targets')" />
```
spouliot: same here, does not make sense to have both

wachaudh: I tried removing one and XliffTasks stopped working. I added XliffTasks using VSMac and it added two, I think it needs both.

src/ObjCRuntime/Registrar.cs
```diff
@@ -1507,8 +1506,7 @@ bool VerifyNonGenericMethod (ref List<Exception> exceptions, TType declaringType
 			if (!IsGenericMethod (method))
 				return true;
 
-			AddException (ref exceptions, CreateException (4113, method, 
-				"The registrar found a generic method: '{0}'. Exporting generic methods is not supported, and will lead to random behavior and/or crashes", 
+			AddException (ref exceptions, CreateException (4113, method, Xamarin.Bundler.Errors.MT4113,
```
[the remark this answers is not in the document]
wachaudh: nope

tools/mtouch/Target.cs
```diff
@@ -113,19 +113,18 @@ public void SelectMonoNative ()
...
 				break;
 			default:
-				throw ErrorHelper.CreateError (71, "Unknown platform: {0}. This usually indicates a bug in Xamarin.iOS; please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new with a test case.", App.Platform);
+				throw ErrorHelper.CreateError (71, Errors.MX0071, App.Platform, "Xamarin.iOS");
 			}
 		}
 
-		public string GetLibNativeName ()
-		{
+		public string GetLibNativeName () {
```
spouliot: minor: incorrect formatting change

PR #7718, opened 2020-01-15
[VSTS] Bump maccore and ensure that the status can be posted.

mandel-macaque:
Bump maccore so that a status is posted when the device tests cannot be
scheduled. Update the JenkinsFile to make sure that the required token
is passed to the launch.sh script.

This follows PR https://github.com/xamarin/maccore/pull/2123


mandel-macaque:

PR #7726, opened 2020-01-20
[jenkins] Fix a couple of issues.

rolfbjarne:
* Fix fetching author's name + email for failure reporting by running git in
  the directory where the repo actually is.
* Don't fail the build if we fail to generate/upload artifacts.json, instead
  show the failure everywhere and continue as if nothing happened. There seems
  to be a random issue with GitHub where a network request fails, and there's
  no need for that to fail the entire build.


rolfbjarne:
Test failure is unrelated

* monotouch-test/iOS 64-bit: https://github.com/xamarin/maccore/issues/2128

PR #7732, opened 2020-01-21
[Jenkins] Fix JenkinsFile groovy.

mandel-macaque:
Fixes compilation error: https://github.com/xamarin/xamarin-macios/commit/f8951f092e331c585f39f956b31a80d8a967de41


mandel-macaque:
Merging even before is done because we needed to test if it was compilable. No need to wait.

Issue #7733, opened 2020-01-21
Obj-C blocks are not being generated correctly in registrar.h


SotoiGhost:
### Steps to Reproduce

1. Open the test case
2. Build the iOS project on a device

### Expected Behavior

Have a successful build. 🎉 

### Actual Behavior

Getting build errors 🔥 like:

```.
/obj/iPhone/Debug/device-builds/iphone11.2-13.3/mtouch-cache/registrar.h(82,82): Error GF247FC6C: interface type 'NSString' cannot be passed by value; did you forget * in 'NSString'? (GF247FC6C) (MessagingTest.iOS)
```

This is the generated code in registrar.h

```
-(void) retrieveFCMTokenForSenderID:(NSString *)p0 completion:(void (^)(NSString, id))p1;
```

It seems that this only happens on Obj-C blocks.

Here are the methods defined in the `ApiDefinition.cs`:

```csharp
// typedef void(^FIRMessagingFCMTokenFetchCompletion)(NSString * _Nullable FCMToken, NSError* _Nullable error) FIR_SWIFT_NAME(MessagingFCMTokenFetchCompletion);
delegate void MessagingFcmTokenFetchCompletionHandler([NullAllowed] string fcmToken, [NullAllowed] NSError error);

// - (void)retrieveFCMTokenForSenderID:(nonnull NSString *)senderID completion:(nonnull FIRMessagingFCMTokenFetchCompletion) completion FIR_SWIFT_NAME(retrieveFCMToken(forSenderID:completion:));
[Async]
[Export("retrieveFCMTokenForSenderID:completion:")]
void RetrieveFcmToken(string senderId, MessagingFcmTokenFetchCompletionHandler completion);
```

_**P.S:** This works on stable channel. Changing to preview channel hits the errors_ 

### Environment

```
=== Visual Studio Enterprise 2019 for Mac (Preview) ===

Version 8.5 Preview (8.5 build 1543)
Installation UUID: 48b5fb7e-9587-45b0-b10e-58dc44b4c3fe
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.14.1.17 (d16-5 / d30880059)

	Package version: 608000087

=== Mono Framework MDK ===

Runtime:
	Mono 6.8.0.87 (2019-10/df42020fe6f) (64-bit)
	Package version: 608000087

=== Roslyn (Language Service) ===

[...]

PR #7735, opened 2020-01-22
[registrar] Fixes NSString trampoline code generation in static registrar

dalexsoto:
Fixes xamarin/xamarin-macios#7733

This was introduced as a side effect of commit 8425129, we
used to generate `id foo` instead of the full block signature
in the trampoline code used by the static registrar, this is
the reason we never caught this condition before.

Added registrar test.

### DIFF
--- a/tests/mtouch/RegistrarTest.cs
+++ b/tests/mtouch/RegistrarTest.cs
@@ -1843,6 +1843,32 @@ class C : NSObject {
 				mtouch.AssertWarning (4179, $"The registrar found the abstract type 'SomeNativeObject' in the signature for 'C.M2'. Abstract types should not be used in the signature for a member exported to Objective-C.", "testApp.cs", 21);
 			}
 		}
+
+		// https://github.com/xamarin/xamarin-macios/issues/7733
+		[Test]
+		public void GHIssue7733 ()
+		{
+			var str1 = @"
+public delegate void ACompletionHandler (string strArg, NSError error);
+
+[Preserve]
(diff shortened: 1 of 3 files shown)


tests/mtouch/RegistrarTest.cs
```diff
@@ -1843,6 +1843,32 @@ public SomeNativeObject M2 ()
...
+		[Test]
+		public void GHIssue7733 ()
+		{
+			var str1 = @"
+public delegate void ACompletionHandler (string strArg, NSError error);
+
+[Preserve]
+public class WidgetObject : NSObject {
+	[Export (""doSomeWork:completion:"")]
+	public virtual void DoWork (string who, ACompletionHandler completion)
```
rolfbjarne: I've usually put positive registrar tests (code that's supposed to work, not code that's supposed to show an error) in tests/monotouch-test/ObjCRuntime/RegistrarTest, because that's much faster to execute (this test adds a whole project build to the tests, while in tests/monotouch-test/ObjCRuntime/RegistrarTest the additional work is just to compile a few more lines of C#).

dalexsoto: Thanks, will move them 👍

dalexsoto:
@rolfbjarne feedback applied, thanks!

dalexsoto:
build

dalexsoto:
@monojenkins backport d16-5

dalexsoto:
Unrelated random tests failures including Known failure: HE0038

PR #7745, opened 2020-01-23
[XHarness] If in CI, no matter which, use the Xml output.

mandel-macaque:
We used to test only on Jenkins, and if the build version had jenkins on
it, rather than doing so, just check if we are in the CI by looking if
BUILD_REVISION is present in the env.


mandel-macaque:

PR #7748, opened 2020-01-23
[msbuild] Remove dependency on Mono.Posix.

rolfbjarne:
Mono.Posix comes with a native library, and that makes IL merging much more
complicated, especially when Mono.Posix comes from a nuget. So remove the
Mono.Posix dependency, and instead write the 2 P/Invokes we need manually.


rolfbjarne:
Test failures are unrelated.

* framework-test/watchOS 32-bits: https://github.com/xamarin/maccore/issues/581

PR #7749, opened 2020-01-23
[msbuild] Don't manually quote, instead rely on CommandLineArgumentBuilder's logic.

rolfbjarne:
This fixes an issue related to single quotes and that the fact that mono's behavior
regarding quoting them has changed (by not using single quotes).

This becomes an issue when running with dotnet.


rolfbjarne:

PR #7752, opened 2020-01-23
[Harness] Remove some of the Wrench logs that do not longer make sense.

mandel-macaque:


rolfbjarne:
We have code that parses the `@MonkeyWrench` lines even when not running on Wrench, so this is not as trivial as it looks:

https://github.com/xamarin/xamarin-macios/blob/197557078fcfece834e80e5072a986dc90a8eba6/jenkins/Jenkinsfile#L230-L248

https://github.com/xamarin/xamarin-macios/blob/197557078fcfece834e80e5072a986dc90a8eba6/tests/xharness/Jenkins.cs#L2737-L2759

mandel-macaque:
Arg, closing then

PR #7755, opened 2020-01-24
[Xharness] Add workaround when we could not share the internet in bots.

mandel-macaque:
Looks like we have issues with the internet sharing in the VSTS bots,
this means that now that we always try to parse the XML on CI, we get an
exception, catch it and do not show the results.

The workaround simply tries to read the xml, if possible, we will parse
it, else deal with the text only log. The fix will show the results, but
it is a workaround for a configuration issue in the CI.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -336,18 +336,14 @@ namespace xharness
 
 		// test if the file is valid xml, or at least, that can be read. The reader will throw an 
 		// exception if it cannot read the file. This is not ideal :/
-		bool IsValidXml (string filePath)
+		bool IsXml (string filePath)
 		{
 			if (!File.Exists (filePath))
 				return false;
 
-			try {
-				using (var stream = new StreamReader (filePath)) 
-				using (var reader = XmlReader.Create (stream)) {
-					return true;
-				}
-			} catch (Exception e) {
-				return false;
+			using (var stream = File.OpenText (filePath)) {
+				var firstLine = stream.ReadLine ();
+				return firstLine.StartsWith("<", StringComparison.Ordinal);
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


tests/xharness/AppRunner.cs
```diff
@@ -334,6 +334,23 @@ public async Task<ProcessExecutionResult> UninstallAsync ()
...
+		// test if the file is valid xml, or at least, that can be read. The reader will throw an 
+		// exception if it cannot read the file. This is not ideal :/
+		bool IsValidXml (string filePath)
+		{
+			if (!File.Exists (filePath))
+				return false;
+
+			try {
+				using (var stream = new StreamReader (filePath)) 
+				using (var reader = XmlReader.Create (stream)) {
```
spouliot: double check that does not use/create an XmlResolver that will try to do network access

tests/xharness/AppRunner.cs
```diff
@@ -336,18 +336,14 @@ public async Task<ProcessExecutionResult> UninstallAsync ()
...
-			try {
-				using (var stream = new StreamReader (filePath)) 
-				using (var reader = XmlReader.Create (stream)) {
-					return true;
-				}
-			} catch (Exception e) {
-				return false;
+			using (var stream = File.OpenText (filePath)) {
+				var firstLine = stream.ReadLine ();
+				return firstLine.StartsWith("<", StringComparison.Ordinal);
```
spouliot: minor: missing space before `(`

mandel-macaque:
build

mandel-macaque:
build

tests/xharness/AppRunner.cs
```diff
@@ -334,6 +334,19 @@ public async Task<ProcessExecutionResult> UninstallAsync ()
...
+		// test if the file is valid xml, or at least, that can be read. The reader will throw an 
+		// exception if it cannot read the file. This is not ideal :/
+		bool IsXml (string filePath)
+		{
+			if (!File.Exists (filePath))
+				return false;
+
+			using (var stream = File.OpenText (filePath)) {
+				var firstLine = stream.ReadLine ();
+				return firstLine.StartsWith ("<", StringComparison.Ordinal);
```
dalexsoto: heh, I love this ❤️

mandel-macaque:

PR #7761, opened 2020-01-27
[src] Add missing availability attributes

spouliot:
Minor, it's less helpful (inside the IDE) and it also mean generating
code for 32bits that can't be executed successfully at runtime.


spouliot:
known issue HE0038 https://github.com/xamarin/maccore/issues/581

PR #7786, opened 2020-02-03
[mtouch] Auto-change settings where they're honored for extensions. Fixes #7780.

### LINKED ISSUE #7780: Incorrect mscorlib.dll used for extensions when the interpreter is enabled
# Steps to Reproduce

1. Create iOS app and add an iOS extension.
2. Select device configuration.
3. Check 'Enable the Mono interpreter' for both.
4. Make sure all other project options are identical as well.
5. Build.

### Expected Behavior

No MT0113 warning.

### Actual Behavior

> warning MT0113: Native code sharing has been disabled for the extension 'actext' because the container app is referencing the assembly 'mscorlib' from '/work/maccore/master/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/lib/mono/Xamarin.iOS/repl/mscorlib.dll', while the extension references a different version from '/work/maccore/master/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/lib/mono/Xamarin.iOS/mscorlib.dll'.

### Environment

xamarin-macios/master: 00985a55e2ed8d5fb2f329683ccc44c90be7ce29

### Build Logs

https://gist.github.com/rolfbjarne/9001ed51815fcd752921df0726b4b6e9

### Example Project (If Possible)

[mt0113test-a0e31ae.zip](https://github.com/xamarin/xamarin-macios/files/4146596/mt0113test-a0e31ae.zip)

rolfbjarne:
When building extensions, we first store all the mtouch arguments in a file
when msbuild builds the extension, and then when msbuild builds the main
project, we load those arguments again and actually build the extension at the
same time as we build the main app.

As such, it's important to make sure that when we reload the extension
arguments we end up with the exact same build configuration as the first time.

Unfortunately that was not the case regarding the interpreter: we
automatically set the 'EnableRepl' value in Main according to whether the
interpreter was enabled or not, but Main is not called after re-loading the
arguments when building extensions.

Fix this by moving the logic that automatically sets 'EnableRepl' to somewhere
that is executed when re-loading arguments when building extensions.

Fixes https://github.com/xamarin/xamarin-macios/issues/7780.

### DIFF
--- a/tests/mtouch/MTouch.cs
+++ b/tests/mtouch/MTouch.cs
@@ -1619,6 +1619,7 @@ public class B : A {}
 		[TestCase ("", "System.dll", "the interpreted assemblies are different between the container app (all assemblies) and the extension (System.dll).")]
 		[TestCase ("mscorlib.dll", "System.dll", "the interpreted assemblies are different between the container app (mscorlib.dll) and the extension (System.dll).")]
 		[TestCase ("mscorlib.dll", "mscorlib.dll,System.dll", "the interpreted assemblies are different between the container app (mscorlib.dll) and the extension (mscorlib.dll, System.dll).")]
+		[TestCase ("-all", "-all", null)]
 		public void MT0113_interpreter (string app_interpreter, string appex_interpreter, string msg)
 		{
 			using (var extension = new MTouchTool ()) {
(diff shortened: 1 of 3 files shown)


rolfbjarne:

PR #7791, opened 2020-02-04
Bump XI/XM versions after d16-6 branching

spouliot:


spouliot:
@rolfbjarne we can revisit if/when we run out of numbers ;-)

spouliot:
That's 14x HE0038 https://github.com/xamarin/maccore/issues/581 :(
and two timedout tests (on watchOS too) -> https://github.com/xamarin/maccore/issues/2142

PR #7794, opened 2020-02-04
[XHarness] Refactor xml code to make the AppRunner class cleaner.

mandel-macaque:
There is a lot of logic related to xml in the AppRunner making it more
complicated than needed. Move the methods to a hlper class.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -34,13 +34,6 @@ namespace xharness
 		TodayExtension,
 	}
 
-	public enum XmlResultType
-	{
-		TouchUnit,
-		NUnit,
-		xUnit,
-	}
-
 	public class AppRunner
 	{
 		public Harness Harness;
@@ -364,7 +357,6 @@ namespace xharness
 			if (Harness.InCI && XmlResultParser.IsXml (path)) {
 				(string resultLine, bool failed, bool crashed) parseResult = (null, false, false);
 				crashed = false;
-				XmlResultType xmlType;
 				try {
 					XmlResultParser.XmlResultType xmlType = XmlResultParser.GetXmlType (path);
 					var newFilename = XmlResultParser.GetXmlFilePath (path, xmlType);
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -353,9 +353,8 @@ namespace xharness
 			// going to check if we are in CI, but also if the listener_log is valid.
 			var path = Path.ChangeExtension (test_log_path, "xml");
 			XmlResultParser.CleanXml (test_log_path, path);
-			XmlResultParser.Jargon xmlType;
 
-			if (Harness.InCI && XmlResultParser.IsValidXml (path, out xmlType)) {
+			if (Harness.InCI && XmlResultParser.IsValidXml (path, out var xmlType)) {
 				(string resultLine, bool failed, bool crashed) parseResult = (null, false, false);
 				crashed = false;
 				try {
@@ -369,6 +368,7 @@ namespace xharness
 					var tmpFile = Path.Combine (Path.GetTempPath (), Guid.NewGuid ().ToString ());
 					(parseResult.resultLine, parseResult.failed) = XmlResultParser.GenerateHumanReadableResults (path, tmpFile, xmlType);
 					File.Copy (tmpFile, test_log_path, true);
+					File.Delete (tmpFile);
 
 					// we do not longer need the tmp file
 					Logs.AddFile (path, "Test xml");
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -341,219 +341,6 @@ namespace xharness
 			}
 		}
 
-		// test if the file is valid xml, or at least, that can be read. The reader will throw an 
-		// exception if it cannot read the file. This is not ideal :/
-		bool IsXml (string filePath)
-		{
-			if (!File.Exists (filePath))
-				return false;
-
-			using (var stream = File.OpenText (filePath)) {
-				string line;
-				while ((line = stream.ReadLine ()) != null) {
-					if (line.Contains ("ping"))
-						continue;
-					return line.StartsWith ("<", StringComparison.Ordinal);
-				}
-			}
-			return false;
-		}
-
-		XmlResultType GetXmlType (StreamReader stream)
-		{
-			var resultType = XmlResultType.TouchUnit; // default
-			// more fun, the first like of the stream, is a ping from the application to the tcp server, and that will not be parsable as
-			// xml, advance the reader one line.
-			string line;
-			while ((line = stream.ReadLine ()) != null) {
-				if (line.Contains ("TouchUnitTestRun")) {
-					resultType = XmlResultType.TouchUnit;
-					break;
-				}
-				if (line.Contains ("nunit-version"))
-					resultType = XmlResultType.NUnit;
-				if (line.Contains ("xUnit")) {
-					resultType = XmlResultType.xUnit;
-					break;
-				}
-			}
-			
-			// we want to reuse the stream (and we are sync)
-			stream.BaseStream.Position = 0;
-			stream.DiscardBufferedData ();
-			return resultType;
-		}
-
-		(string resultLine, bool failed) ParseTouchUnitXml (StreamReader stream, StreamWriter writer)
-		{
-			long total, errors, failed, notRun, inconclusive, ignored, skipped, invalid;
-			total = errors = failed = notRun = inconclusive = ignored = skipped = invalid = 0L;
-			
-			using (var reader = XmlReader.Create (stream)) {
-				while (reader.Read ()) {
-					if (reader.NodeType == XmlNodeType.Element && reader.Name == "test-results") {
-						total = long.Parse (reader ["total"]);
-						errors = long.Parse (reader ["errors"]);
-						failed = long.Parse (reader ["failures"]);
-						notRun = long.Parse (reader ["not-run"]);
-						inconclusive = long.Parse (reader ["inconclusive"]);
-						ignored = long.Parse (reader ["ignored"]);
-						skipped = long.Parse (reader ["skipped"]);
-						invalid = long.Parse (reader ["invalid"]);
-					}
-					if (reader.NodeType == XmlNodeType.Element && reader.Name == "TouchUnitExtraData") {
-						// move fwd to get to the CData
-						if (reader.Read ())
-							writer.Write (reader.Value);
-					}
-				}
-			}
-			var passed = total - errors - failed - notRun - inconclusive - ignored - skipped - invalid;
-			var resultLine = $"Tests run: {total} Passed: {passed} Inconclusive: {inconclusive} Failed: {failed + errors} Ignored: {ignored + skipped + invalid}";
-			return (resultLine, total == 0 || errors != 0 || failed != 0);
-		}
-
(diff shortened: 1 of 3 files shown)


tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,267 @@
...
+using System.Xml;
+
+namespace xharness {
+	public static class XmlResultParser {
+
+		public enum XmlResultType {
+			TouchUnit,
+			NUnit,
+			xUnit,
+		}
```
spouliot: minor: inner (public) types requires more typing to use and are less discoverable

In this case it also duplicate (part of) the name

either
* move it outside the type (and keep the name) - my preference
or
* rename it to something smaller, e.g. `Jargon` because `Type` alone would be confusing

so you don't have to write `XmlResultParser.XmlResultType` but `XmlResultParser.Jargon`

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,267 @@
...
+				while ((line = stream.ReadLine ()) != null) { // special case when get got the tcp connection
+					if (line.Contains ("ping"))
+						continue;
+					return line.StartsWith ("<", StringComparison.Ordinal);
+				}
+			}
+			return false;
+		}
+
+		public static XmlResultType GetXmlType (string filePath)
```
spouliot: can't this be merged inside `IsXml` ?
we're parsing (part of) the same file twice

instead you could have `bool IsValidXml (string path, out XType type)`

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,267 @@
...
+		{
+
+			using (var streamReaderTmp = new StreamReader (path)) {
+				var fileName = Path.GetFileName (path);
+				var newFilename = "";
+				switch (xmlType) {
+				case XmlResultType.TouchUnit:
+				case XmlResultType.NUnit:
+					newFilename = path.Replace (fileName, $"nunit-{fileName}");
+					break;
```
spouliot: you could return from here (and below) and remove local

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,267 @@
...
+			string resultLine = $"Tests run: {total} Passed: {passed} Inconclusive: {inconclusive} Failed: {failed + errors} Ignored: {ignored + skipped + invalid}";
+			writer.WriteLine (resultLine);
+
+			return (resultLine, total == 0 | errors != 0 || failed != 0);
+		}
+
+		public static string GetXmlFilePath (string path, XmlResultType xmlType)
+		{
+
+			using (var streamReaderTmp = new StreamReader (path)) {
```
spouliot: what's the point of `streamReaderTmp` ?
there's cheaper ways to check if the file exists...

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,267 @@
...
+			}
+		}
+
+		public static void CleanXml (string source, string destination)
+		{
+			using (var reader = new StreamReader (source))
+			using (var writer = new StreamWriter (destination)) {
+				string line;
+				while ((line = reader.ReadLine ()) != null) {
+					if (line.StartsWith ("ping", StringComparison.InvariantCulture) || line.Contains ("TouchUnitTestRun") || line.Contains ("NUnitOutput") || line.Contains ("<!--")) {
```
spouliot: `StringComparison.InvariantCulture` is much more costly than `Ordinal`

tests/xharness/XmlResultParser.cs
mandel-macaque: Looks like a left over when I was cleaning it, will remove it.

tests/xharness/AppRunner.cs
```diff
@@ -572,57 +352,28 @@ void CleanXml (string source, string destination)
...
-						}
-						File.Copy (tmpFile, test_log_path, true);
-					}
+					var newFilename = XmlResultParser.GetXmlFilePath (path, xmlType);
+
+					// rename the path to the correct value
+					File.Move (path, newFilename);
+					path = newFilename;
+
+					// write the human readable results in a tmp file, which we later use to step on the logs
```
rolfbjarne: Could we instead add both files to the html report? Replacing/rewriting files ends up being confusing.

mandel-macaque: Both are added. Take a look at: https://github.com/xamarin/xamarin-macios/pull/7794/files/46600f2ba5ddc491bc58530e93b4d4509f00af04#diff-03773861bcef485bf343300f31b60b0eR374

We step on the xml written by the log and add it as the human readable version, later we add a new xml log. The final html report looks like this:

Screenshot 2020-02-04 at 14 11 07

Clicking on 'Test Log' shows the human readable ones. Xml shows the xml (useful for debugging or importing somewhere).

rolfbjarne: This still seems more complex than needed to me. Why not just write out the human readable version to a new file and add that, without stepping on anything we already have?

Also, when is `tmpFile` deleted? The usual way to create a log file is to use `Logs.CreateFile`, which will create a file next to the other logs, and which will also be deleted whenever those are deleted (during a `git clean` for instance).

tests/xharness/AppRunner.cs
```diff
@@ -572,57 +352,28 @@ void CleanXml (string source, string destination)
 			// from the TCP connection, we are going to fail when trying to read it and not parse it. Therefore, we are not only
 			// going to check if we are in CI, but also if the listener_log is valid.
 			var path = Path.ChangeExtension (test_log_path, "xml");
-			CleanXml (test_log_path, path);
+			XmlResultParser.CleanXml (test_log_path, path);
+			XmlResultParser.Jargon xmlType;
 
-			if (Harness.InCI && IsXml (test_log_path)) {
+			if (Harness.InCI && XmlResultParser.IsValidXml (path, out xmlType)) {
```
spouliot: if you use `out var xmlType`
then you can remove the local declaration

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,257 @@
...
+			xUnit,
+			Text,
+		}
+
+		// test if the file is valid xml, or at least, that can be read it.
+		public static bool IsValidXml (string path, out Jargon type)
+		{
+			type = Jargon.Text;
+			if (!File.Exists (path))
+				return false;
```
spouliot: should not `type` be something like `Invalid` or `Missing` ?

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,257 @@
...
+			bool isXml = false;
+			using (var stream = File.OpenText (path)) {
+				string line;
+				while ((line = stream.ReadLine ()) != null) { // special case when get got the tcp connection
+					if (line.Contains ("ping"))
+						continue;
+					if (line.Contains ("TouchUnitTestRun")) {
+						type = Jargon.TouchUnit;
+						isXml = true;
+						break;
```
spouliot: why not return `true` here ? (and in other cases too)
less code, nothing (now or later) can't happen outside the `while` to affect the logic
no need for `isXml` local either
easier to debug a `false` case since a breakpoint to the last `return` will always be false.

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,257 @@
...
+				while (reader.Read ()) {
+					if (reader.NodeType == XmlNodeType.Element && reader.Name == "test-results") {
+						total = long.Parse (reader ["total"]);
+						errors = long.Parse (reader ["errors"]);
+						failed = long.Parse (reader ["failures"]);
+						notRun = long.Parse (reader ["not-run"]);
+						inconclusive = long.Parse (reader ["inconclusive"]);
+						ignored = long.Parse (reader ["ignored"]);
+						skipped = long.Parse (reader ["skipped"]);
+						invalid = long.Parse (reader ["invalid"]);
```
spouliot: potential for exceptions, `TryParse` would be safer
I guess it depends on how much do you trust the XML input ;-)

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,257 @@
...
```
spouliot: same, `TryParse`

tests/xharness/XmlResultParser.cs
```diff
@@ -0,0 +1,257 @@
...
+		{
+			long total, errors, failed, notRun, inconclusive, ignored, skipped, invalid;
+			total = errors = failed = notRun = inconclusive = ignored = skipped = invalid = 0L;
+			using (var reader = XmlReader.Create (stream)) {
+				while (reader.Read ()) {
+					if (reader.NodeType == XmlNodeType.Element && reader.Name == "assembly") {
+						total += long.Parse (reader ["total"]);
+						errors += long.Parse (reader ["errors"]);
+						failed += long.Parse (reader ["failed"]);
+						skipped += long.Parse (reader ["skipped"]);
```
spouliot: same, `TryParse`

tests/xharness/XmlResultParser.cs
spouliot: unresolving - that was not fixed!

please don't resolve other comments, i.e. let the commenter resolve
otherwise there's a risk that the comment will not be addressed and missed

tests/xharness/AppRunner.cs
mandel-macaque: The TestLog is badly formatted xml (include the ping and other stuff). We clean it and create the xml file. At that point the Test Log is no very useful for a human, so we step on it with the human readable version. 

It is true that I forgot to delete the tmp file.

rolfbjarne: > At that point the Test Log is no very useful for a human

It's useful for unconfusing humans who wonder why they don't get the same results / behavior when running tests locally vs running tests on CI :smile:

mandel-macaque: @rolfbjarne but this just happens on CI where we have xml, not locally. We could set to have xml ALWAYS, which is another option to be the same

tests/xharness/XmlResultParser.cs
```diff
@@ -148,10 +145,14 @@ public static bool IsValidXml (string path, out Jargon type)
 			using (var reader = XmlReader.Create (stream)) {
 				while (reader.Read ()) {
 					if (reader.NodeType == XmlNodeType.Element && reader.Name == "assembly") {
-						total += long.Parse (reader ["total"]);
-						errors += long.Parse (reader ["errors"]);
-						failed += long.Parse (reader ["failed"]);
-						skipped += long.Parse (reader ["skipped"]);
+						long.TryParse (reader ["total"], out var asseblyCount);
+						total += asseblyCount;
```
spouliot: minor: typo `asseblyCount` vs `assemblyCount`

tests/xharness/AppRunner.cs
rolfbjarne: > but this just happens on CI where we have xml, not locally.

Yes, that's fine. The confusing part is that replacing files with different content ends up being confusing, especially when something goes wrong and only some files are replaced and then you have several files that are named similarly but with wildly different content.

Keeping all the files around avoids this.

mandel-macaque: Ok, I now understand, I created https://github.com/xamarin/xamarin-macios/issues/7802 lets keep it there, get it to work and I'll do the clean up later. I want to get the CI ready ASAP.

mandel-macaque:

PR #7796, opened 2020-02-04
[VSTS] Update ddfun pipeline to upload xml results.

mandel-macaque:
Add two new steps that will ensure that the xml is uploaded to vsts.


mandel-macaque:
All are https://github.com/xamarin/maccore/issues/581

Issue #7793, opened 2020-02-04
[mtouch] test are failing after localization


spouliot:
### Steps to Reproduce

1. run mtouch tests (e.g. `make runner`)

### Expected Behavior

No failures.

### Actual Behavior

```
Xamarin.MTouch.MT0025(iOS): The error 'MT0025: No SDK version was provided. Please add --sdk=X.Y to specify which iOS SDK should be used to build your application.' was not found in the output:
    Message #1 did not match:
        actual: 'No SDK version was provided. Please add `--sdk=X.Y` to specify which iOS SDK should be used to build your application.'
        expected: 'No SDK version was provided. Please add --sdk=X.Y to specify which iOS SDK should be used to build your application.'
Xamarin.MTouch.MT0025(watchOS): The error 'MT0025: No SDK version was provided. Please add --sdk=X.Y to specify which watchOS SDK should be used to build your application.' was not found in the output:
    Message #1 did not match:
        actual: 'No SDK version was provided. Please add `--sdk=X.Y` to specify which watchOS SDK should be used to build your application.'
        expected: 'No SDK version was provided. Please add --sdk=X.Y to specify which watchOS SDK should be used to build your application.'
Xamarin.MTouch.MT0025(tvOS): The error 'MT0025: No SDK version was provided. Please add --sdk=X.Y to specify which tvOS SDK should be used to build your application.' was not found in the output:
    Message #1 did not match:
        actual: 'No SDK version was provided. Please add `--sdk=X.Y` to specify which tvOS SDK should be used to build your application.'
        expected: 'No SDK version was provided. Please add --sdk=X.Y to specify which tvOS SDK should be used to build your application.'
Xamarin.MTouch.MT0073: The error 'MT0073: Xamarin.iOS .* does not support a deployment target of 3.1 for iOS .the minimum is 7.0.. Please select a newer deployment target in your project's Info.plist.' was not found in the output:
    The message 'Unknown platform: 13.15.1. This usually indicates a bug in 3.1; please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new with a test case.' did not match the pattern 'Xamarin.iOS .* does not support a deployment target of 3.1 for iOS .the minimum is 7.0.. Please select a newer deployment target in your project's Info.plist.'.
[...]

PR #7798, opened 2020-02-04
[mtouch] Fix tests broken by localization

wachaudh:
Fixes https://github.com/xamarin/xamarin-macios/issues/7793.

### DIFF
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -329,7 +329,7 @@ namespace Xamarin.Bundler {
 
 			if (DeploymentTarget != null) {
 				if (DeploymentTarget < Xamarin.SdkVersions.GetMinVersion (Platform))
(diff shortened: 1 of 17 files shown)


spouliot:
@wachaudh bots still does not have devices
`Not run: 211, Invalid: 0, Ignored: 211, Skipped: 0`
so make sure you tested locally before merging :)

rolfbjarne:
build

dalexsoto:
@rolfbjarne I think he needs to create a branch in xamarin-macios and not be in his own fork to run device tests, right?

rolfbjarne:
@dalexsoto no, Sebastien is referring to the fact that we don't enable device support for PR builds by default, which means that we can't (and don't) run tests that need device support (such as tests that build for device). That's fixed by adding the `enable-device-build` label, which I did earlier today, and re-kicked a test run.

mandel-macaque:
Failing tests are known:

* https://github.com/xamarin/maccore/issues/581
* https://github.com/xamarin/maccore/issues/1440

PR #7803, opened 2020-02-05
[msbuild] Make DeviceSpecificOutputPath an absolute path.

rolfbjarne:
It might be absolute at least sometimes when building with dotnet, and making
it always an absolute path simplifies the code.

This also had an affect on AppBundleDir, AppBundlePath, and a few other app
bundle variables, which do not need to be absolutified anymore because they're
based on DeviceSpecificOutputPath.


kzu:
We'll ned to make sure nothing breaks on Windows side from this when creating directories and copying stuff over to the mac during a build.. @emaf

emaf:
Hey @rolfbjarne, have you tried this from VS on Windows? I'm not really sure this is going to work since all the tasks receiving `AppBundleDir` or `DeviceSpecificOutputPath` as parameter will end up using full Windows paths on the Mac. For instance the CodeSign task: https://github.com/xamarin/xamarin-macios/blob/master/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets#L2014 and https://github.com/xamarin/xamarin-macios/blob/master/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignVerifyTaskBase.cs#L48.

rolfbjarne:
@emaf no, I haven't tested on VS on Windows, but I sounds like these changes aren't going to work. However, I think I'll try the opposite approach: make sure DeviceSpecificOutputPath is always a relative path, that sounds like it should be much safer.

emaf:
oh, that's great, we should be fine with that change

rolfbjarne:
Superseded by #7832.

PR #7819, opened 2020-02-07
[jenkins] Switch to use the catalina bot group

spouliot:


spouliot:
^ we can see it was picked up ^
it will be again after merge (no need to wait on above)

PR #7824, opened 2020-02-07
[Harness] Just try to generate the report when is NUnit.

mandel-macaque:
Use the xml parsing helper methods to decide if the xml can be parsed
and if we will be able to generate the report. That way we avoid an
exception that makes the CI noise.


mandel-macaque:
@monojenkins backport d16-6
@monojenkins backport d16-5
@monojenkins backport Xcode11.4

mandel-macaque:

PR #7825, opened 2020-02-07
[tests] Add cecil-based unit tests

spouliot:
Only one to start... it's been discussed before but we generally
found other ways to do them. Let's continue to pick the best place
but we now have more options :)

### DIFF
--- a/tests/cecil-tests/Makefile
+++ b/tests/cecil-tests/Makefile
@@ -5,7 +5,8 @@ include $(TOP)/Make.config
 all-local::
 
 build:
-	$(SYSTEM_MSBUILD) -restore
+	nuget restore
+	$(SYSTEM_MSBUILD)
 
 run-tests: build
 	$(SYSTEM_MONO) --debug $(TOP)/packages/NUnit.ConsoleRunner.3.9.0/tools/nunit3-console.exe $(abspath $(TOP)/tests/cecil-tests/bin/Debug/cecil-tests.dll)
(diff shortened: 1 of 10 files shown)


tests/cecil-tests/Test.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+using NUnit.Framework;
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+#nullable enable
```
mandel-macaque: Fancy!

tests/cecil-tests/Test.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Cecil.Tests {
+
+	[TestFixture]
+	public class Test {
+
+		[TestCase ("/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/32bits/Xamarin.iOS.dll")]
```
rolfbjarne: This references the system install, not the locally built one.

You can compute the local path by including `tests/common/Configuration.cs` in the project, and then do something like this:

```csharp
var localPath = Path.Combine (Configuration.IOS_DESTDIR, assemblyName.SubString (1));
```

Adjust as required for Xamarin.Mac (MAC_DESTDIR).

tests/cecil-tests/Makefile
```diff
@@ -0,0 +1,16 @@
+TOP=../..
+
+include $(TOP)/Make.config
+
+all-local::
+
+build:
+	msbuild
```
rolfbjarne: This is more quiet and also honors V=1 when doing `make V=1`:

```suggestion
	$(SYSTEM_MSBUILD)
```

PR #7832, opened 2020-02-10
[msbuild] Make sure DeviceSpecificOutputPath is a relative path.

rolfbjarne:
It might be absolute at least sometimes when building with dotnet, and making
sure it's always a relative path simplifies the code.


rolfbjarne:
@chamons no, because XM doesn't have a device-specific output path.

Issue #7826, opened 2020-02-07
[Harness] Add the ability to generate reports for xUnit and NUnit3.


mandel-macaque:
Following https://github.com/xamarin/xamarin-macios/pull/7824 we are not generating reports for those two types of formats. Add support for them and then check the jargon to decide what to do.

PR #7835, opened 2020-02-10
[XHarness] Do write the TestReport when we have xUnit results.

mandel-macaque:
Add code to write the TestReport when we have xunit results. Also
refactored code for the NUnit format to not use LINQ, it will use less
memory and in the future we can move to async (not now since it might
raise other problems).

fixes: https://github.com/xamarin/xamarin-macios/issues/7826

### DIFF
--- a/tests/xharness/Extensions.cs
+++ b/tests/xharness/Extensions.cs
@@ -126,5 +126,11 @@ namespace xharness
 			var rnd = new Random ((int) DateTime.Now.Ticks);
 			return collection.OrderBy (v => rnd.Next ());
 		}
+
+		public static string AsHtmlFormat (this string inString)
+		{
+			var rv = System.Web.HttpUtility.HtmlEncode (inString);
+			return rv.Replace ("\t", "&nbsp;&nbsp;&nbsp;&nbsp;").Replace ("\n", "<br/>\n");
+		}
 	}
 }
--- a/tests/xharness/Extensions.cs
+++ b/tests/xharness/Extensions.cs
@@ -127,7 +127,7 @@ namespace xharness
 			return collection.OrderBy (v => rnd.Next ());
 		}
 
-		public static string AsHtmlFormat (this string inString)
+		public static string AsHtml (this string inString)
 		{
 			var rv = System.Web.HttpUtility.HtmlEncode (inString);
 			return rv.Replace ("\t", "&nbsp;&nbsp;&nbsp;&nbsp;").Replace ("\n", "<br/>\n");
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -2100,7 +2100,7 @@ namespace xharness
 								writer.WriteLine ($"Known failure: {test.KnownFailure} <br />");
 
 							if (!string.IsNullOrEmpty (test.FailureMessage)) {
-								var msg = HtmlFormat (test.FailureMessage);
+								var msg = test.FailureMessage.AsHtmlFormat ();
 								var prefix = test.Ignored ? "Ignored" : "Failure";
 								if (test.FailureMessage.Contains ('\n')) {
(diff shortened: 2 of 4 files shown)


tests/xharness/Jenkins.cs
```diff
@@ -2241,45 +2241,21 @@ <li>Toggle visibility
...
 												writer.WriteLine ("</div>");
 											}
 										} catch (Exception ex) {
-											writer.WriteLine ("<span style='padding-left: 15px;'>Could not parse log file: {0}</span><br />", HtmlFormat (ex.Message));
+											writer.WriteLine ("<span style='padding-left: 15px;'>Could not parse log file: {0}</span><br />", ex.Message.AsHtmlFormat ());
 										}
 									} else if (log.Description == "NUnit results" || log.Description == "XML log") {
 										try {
 											if (File.Exists (log.FullPath) && new FileInfo (log.FullPath).Length > 0) {
 												if (XmlResultParser.IsValidXml (log.FullPath, out var jargon)) {
```
spouliot: missed it in the previous PR :(
iirc `IsValidXml` already include the checks from the previous line, right ?

tests/xharness/Jenkins.cs
```diff
@@ -2321,7 +2297,7 @@ <li>Toggle visibility
 					if (runningTests.Any ()) {
 						writer.WriteLine ($"<h3>{runningTests.Count ()} running tests:</h3>");
 						foreach (var test in runningTests) {
-							writer.WriteLine ($"<a href='#test_{test.TestName}'>{test.TestName} ({test.Mode})</a> {test.Duration.ToString ()} {HtmlFormat ("\n\t" + test.ProgressMessage)}<br />");
+							writer.WriteLine ($"<a href='#test_{test.TestName}'>{test.TestName} ({test.Mode})</a> {test.Duration.ToString ()} {"\n\t" + test.ProgressMessage.AsHtmlFormat ()}<br />");
```
spouliot: Not sure if that really change something in the web view but...

That's not quite the same:
`HtmlFormat ("\n\t" + test.ProgressMessage)`
versus:
`"\n\t" + test.ProgressMessage.AsHtmlFormat ()`

The latter won't format as HTML the `"\n\t"` while the former would.

tests/xharness/Extensions.cs
```diff
@@ -126,5 +126,11 @@ public static IEnumerable<T> Shuffle<T> (this IEnumerable<T> collection)
 			var rnd = new Random ((int) DateTime.Now.Ticks);
 			return collection.OrderBy (v => rnd.Next ());
 		}
+
+		public static string AsHtmlFormat (this string inString)
```
spouliot: minor: I'd argue that `AsHtml` is better ;-)

tests/xharness/Jenkins.cs
```diff
@@ -2241,45 +2241,21 @@ <li>Toggle visibility
 											if (errors.Count > 0) {
 												writer.WriteLine ("<div style='padding-left: 15px;'>");
 												foreach (var error in errors)
-													writer.WriteLine ("{0} <br />", HtmlFormat (error));
+													writer.WriteLine ("{0} <br />",  error.AsHtmlFormat ());
```
spouliot: minor: extra `' '` before `error`

tests/xharness/Jenkins.cs
mandel-macaque: Yes, we should not have an exception. I left it just in case until we do not see it for some time and then kill it. Could be removed.

tests/xharness/Jenkins.cs
mandel-macaque: better change it to be sure is the same output.

mandel-macaque:
Failures are not related, but they are a good example of the PR working ;)

mandel-macaque:
@monojenkins xcode11.4
@monojenkins d16-6

mandel-macaque:
@monojenkins backport d16-6

PR #7836, opened 2020-02-10
[Jenkins] Do not notarize on PRs.

mandel-macaque:
It was added as a backup and has not been needed. With the updates to
catalina it makes builds fail on CI.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -501,11 +501,7 @@ timestamps {
                             if (bundleZip.length > 0)
                                 bundleZipFilename = bundleZip [0].name
 
-                            if (isPr) {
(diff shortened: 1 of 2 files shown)


dalexsoto:
@mandel-macaque how do we fail? I am not entirely sure this is the right approach specially when we build pkgs for specific workarounds.

spouliot:
If I understood correctly the non-notarized version was the _backup_. IOW we signed, attached to PR status, then notarized and (again) attached to PR status.

So the backup was to keep the non-notarized version (in case something broke).
While the goal was to have the notarized (ready to use by everyone) version on the PR status.

It seems the (Catalina specific) issue was verifying the signature before the notarization - so that might be just the part to remove (as I assume we're already verifying the notarized version).

mandel-macaque:
@monojenkins backport d16-6

PR #7837, opened 2020-02-11
[XHarness] Use static variables for common test names.

mandel-macaque:
One of the reason why we lost the TestReport was because the xml added
from the new code did not use the same name for the log as the expected
one (https://github.com/xamarin/xamarin-macios/pull/7835/files#diff-03773861bcef485bf343300f31b60b0eR374).

If we are going to be using the description as a way to decide what is
going to be done with it, lets use static vars so that we do not have
logs added with the wrong name.


mandel-macaque:
@monojenkins backport d16-6

mandel-macaque:

PR #7839, opened 2020-02-11
[Xcode11.4][Jenkins] Do not notarize on PRs. (#7836)

mandel-macaque:
It was added as a backup and has not been needed. With the updates to
catalina it makes builds fail on CI.


mandel-macaque:
monojenkins failed with the auto backport :/

PR #7844, opened 2020-02-11
[XHarness] Add logs to VSTS test runs.

mandel-macaque:
As per the documentation of the VSTS test uploader (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#attachments-support)
the tests support the addition of attachments. We want to be able to
access the logs of the different tests runs, this is achieved the
following way:

1. Move TouchUnit to use NUnit V3 which allows to add attachments.
2. Once the tests are ran, add the attachments to the following nodes:
   1. The very first test-suite. This will allow to have the logs for
   succesul tests.
   2. Add logs to failing tests. Reduces the number of clicks to be done
   to access to the logs when a test case fails.
3. Modify the assembly name of the test-suit to match the name of the
application. This ensures two things.
   1. We have a consistent name for the file column in VSTS, that can be
   used to see recurrent failing tests.
   2. The name is more readable, since if not, it will contain the UUID
   of the device.

Logs are not added to succesful tests because it will have the following
problems:

* Larger data storage usage.
* Longer upload time. The addtion of the logs per tests (succesful or
failed) was tested and resulted in an upload time LONGER than 6 hours
for all TouchUnit, NUnit blc tests and xUnit bcl tests.

In order for this to be useful, the task in the pipeline SHOULD NOT
merge test runs. We should have a test run PER application so that we do
not mix the logs.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -366,7 +366,7 @@ namespace xharness
 					var testRunName = $"{appName} {Variation}";
 					if (xmlType == XmlResultParser.Jargon.NUnitV3) {
 						// add the attachments and write in the new filename
-						XmlResultParser.UpdateMissingData (path, newFilename, testRunName, Directory.GetFiles(Logs.Directory));
+						XmlResultParser.UpdateMissingData (path, newFilename, testRunName, Directory.GetFiles (Logs.Directory));
 					} else {
 						// rename the path to the correct value
 						File.Move (path, newFilename);
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -360,8 +360,16 @@ namespace xharness
 				try {
 					var newFilename = XmlResultParser.GetXmlFilePath (path, xmlType);
 
-					// rename the path to the correct value
-					File.Move (path, newFilename);
+					// at this point, we have the test results, but we want to be able to have attachments in vsts, so if the format is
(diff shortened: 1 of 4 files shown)


mandel-macaque:
The UI will be something similar to the following:
Screenshot 2020-02-11 at 07 42 31

tests/xharness/AppRunner.cs
```diff
@@ -360,8 +360,16 @@ public async Task<ProcessExecutionResult> UninstallAsync ()
...
 					var newFilename = XmlResultParser.GetXmlFilePath (path, xmlType);
 
-					// rename the path to the correct value
-					File.Move (path, newFilename);
+					// at this point, we have the test results, but we want to be able to have attachments in vsts, so if the format is
+					// the right one (NUnitV3) add the nodes. ATM only TouchUnit uses V3.
+					var testRunName = $"{appName} ({configuration})";
+					if (xmlType == XmlResultParser.Jargon.NUnitV3) {
+						// add the attachments and write in the new filename
+						XmlResultParser.UpdateMissingData (path, newFilename, testRunName, Directory.GetFiles(Logs.Directory));
```
rolfbjarne:
```suggestion
						XmlResultParser.UpdateMissingData (path, newFilename, testRunName, Directory.GetFiles (Logs.Directory));
```

mandel-macaque:
@monojenkins backport 16-6
@monojenkins d16-5

mandel-macaque:
@monojenkins backport d16-6

PR #7846, opened 2020-02-11
[XHarness] Add NUnit3 support to the old bcl tests.

mandel-macaque:
Move the old bcl tests based on NUnit to use the NUnit V3 output so that
we can add logs to the vsts imported tests.

Follow up of PR: https://github.com/xamarin/xamarin-macios/pull/7844


mandel-macaque:
All failures are https://github.com/xamarin/maccore/issues/581 that watch is not ticking..

PR #7851, opened 2020-02-11
[XHarness] Generate TestReports for NUnit V3.

mandel-macaque:
Now that we have NUnit V3 xml results, do generate the test report for
those correctly.


mandel-macaque:
@monojenkins backport d16-6

PR #7857, opened 2020-02-12
[AppKit] Update bindings to Xcode 11.4 Beta 1

dalexsoto:
diff: https://github.com/xamarin/xamarin-macios/wiki/AppKit-macOS-xcode11.4-beta1

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -5210,6 +5210,10 @@ namespace AppKit {
 		[Export ("maxDate", ArgumentSemantic.Copy)]
 		NSDate MaxDate { get; set; }
 
+		[NoiOS, Mac (10,15,4)]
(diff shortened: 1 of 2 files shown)


src/appkit.cs
```diff
@@ -5210,6 +5210,10 @@ interface NSDatePicker {
 		[Export ("maxDate", ArgumentSemantic.Copy)]
 		NSDate MaxDate { get; set; }
 
+		[NoiOS, Mac (10,15,4)]
```
spouliot: minor: you do not need `[NoiOS]` in `appkit.cs`

dalexsoto:
```suggestion
		[Mac (10,15,4)]
```

PR #7861, opened 2020-02-12
[CarPlay] Update bindings to Xcode 11.4 Beta 1

dalexsoto:
diff: https://github.com/xamarin/xamarin-macios/wiki/CarPlay-iOS-xcode11.4-beta1


src/carplay.cs
```diff
@@ -940,5 +940,72 @@ interface CPAlertTemplate {
...
+
+	interface ICPTemplateApplicationDashboardSceneDelegate { }
+
+	[NoWatch, NoTV, NoMac, iOS (13,4)]
+	[Protocol, Model (AutoGeneratedName = true)]
+	[BaseType (typeof (NSObject))]
+	interface CPTemplateApplicationDashboardSceneDelegate : UISceneDelegate {
+
+		[Export ("templateApplicationDashboardScene:didConnectDashboardController:toWindow:")]
+		void DidConnectDashboardController (CPTemplateApplicationDashboardScene templateApplicationDashboardScene, CPDashboardController dashboardController, UIWindow window);
```
mandel-macaque: Do we need the DashboardController in the name? Maybe DidConnect? It is the second param.
```suggestion
		void DidConnect (CPTemplateApplicationDashboardScene templateApplicationDashboardScene, CPDashboardController dashboardController, UIWindow window);
```

src/carplay.cs
```diff
@@ -940,5 +940,72 @@ interface CPAlertTemplate {
...
+
+		[Export ("templateApplicationDashboardScene:didDisconnectDashboardController:fromWindow:")]
+		void DidDisconnectDashboardController (CPTemplateApplicationDashboardScene templateApplicationDashboardScene, CPDashboardController dashboardController, UIWindow window);
```
mandel-macaque: Same as the previous comment.
```suggestion
		void DidDisconnect (CPTemplateApplicationDashboardScene templateApplicationDashboardScene, CPDashboardController dashboardController, UIWindow window);
```

src/carplay.cs
dalexsoto: I think I like `DidConnectDashboardController` best because it differentiates with the `DashboardScene` that connected a `DashboardController`

src/carplay.cs
dalexsoto: same as above `DidDisconnectDashboardController` of a `DashboardScene`

src/carplay.cs
dalexsoto: Also very easy to search as is https://www.google.com/search?client=safari&rls=en&q=DidConnectDashboardController&ie=UTF-8&oe=UTF-8

PR #7866, opened 2020-02-12
[Xcode11.4][XHarness] Add logs to VSTS test runs. (#7844)

mandel-macaque:
As per the documentation of the VSTS test uploader (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#attachments-support)
the tests support the addition of attachments. We want to be able to
access the logs of the different tests runs, this is achieved the
following way:

1. Move TouchUnit to use NUnit V3 which allows to add attachments.
2. Once the tests are ran, add the attachments to the following nodes:
   1. The very first test-suite. This will allow to have the logs for
   succesul tests.
   2. Add logs to failing tests. Reduces the number of clicks to be done
   to access to the logs when a test case fails.
3. Modify the assembly name of the test-suit to match the name of the
application. This ensures two things.
   1. We have a consistent name for the file column in VSTS, that can be
   used to see recurrent failing tests.
   2. The name is more readable, since if not, it will contain the UUID
   of the device.

Logs are not added to succesful tests because it will have the following
problems:

* Larger data storage usage.
* Longer upload time. The addtion of the logs per tests (succesful or
failed) was tested and resulted in an upload time LONGER than 6 hours
for all TouchUnit, NUnit blc tests and xUnit bcl tests.

In order for this to be useful, the task in the pipeline SHOULD NOT
merge test runs. We should have a test run PER application so that we do
not mix the logs.

Co-Authored-By: Rolf Bjarne Kvinge <rolf@xamarin.com>

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -496,7 +496,7 @@ namespace xharness
 			if (useXmlOutput) {
 				args.Append (" -setenv=NUNIT_ENABLE_XML_OUTPUT=true");
 				args.Append (" -setenv=NUNIT_ENABLE_XML_MODE=wrapped");
-				args.Append ("-setenv=NUNIT_XML_VERSION=nunitv3");
+				args.Append (" -setenv=NUNIT_XML_VERSION=nunitv3");
 			}
 
 			if (Harness.InCI) {
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -39,6 +39,7 @@ namespace xharness
 		public Harness Harness;
 		public string ProjectFile;
 		public string AppPath;
+		public string Variation;
 
 		public TestExecutingResult Result { get; private set; }
 		public string FailureMessage { get; private set; }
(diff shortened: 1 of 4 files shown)


mandel-macaque:
Again, monojekins does not like me lately.

tests/xharness/AppRunner.cs
```diff
@@ -487,6 +496,7 @@ public async Task<int> RunAsync ()
 			if (useXmlOutput) {
 				args.Append (" -setenv=NUNIT_ENABLE_XML_OUTPUT=true");
 				args.Append (" -setenv=NUNIT_ENABLE_XML_MODE=wrapped");
+				args.Append ("-setenv=NUNIT_XML_VERSION=nunitv3");
```
spouliot: missing space ? is that not concatenated with the previous one ?

dalexsoto: heh good catch!

mandel-macaque: @spouliot nice! yeah, xcode11.4 does not have the new args add code, I missed that

mandel-macaque: @spouliot missing space comes from a manual backport, that was the line that was giving the conflict. Original PR is ok, does not use a string builder.

PR #7869, opened 2020-02-12
[GameKit] Update bindings to Xcode 11.4 Beta 1

dalexsoto:
diff: https://github.com/xamarin/xamarin-macios/wiki/GameKit-iOS-xcode11.4-beta1

### DIFF
--- a/src/gamekit.cs
+++ b/src/gamekit.cs
@@ -689,6 +689,9 @@ namespace GameKit {
 		GKLeaderboardTimeScope TimeScope { get; set; }
 	}
 
+	[Watch (6,2), TV (13,4), Mac (10,15,4), iOS (13,4)]
+	delegate void GKFetchItemsForIdentityVerificationSignatureCompletionHandler (NSUrl publicKeyUrl, NSData signature, NSData salt, ulong timestamp, NSError error);
+
 	[Watch (3,0)]
 	[BaseType (typeof (GKPlayer))]
 	interface GKLocalPlayer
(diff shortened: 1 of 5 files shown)


src/gamekit.cs
```diff
@@ -791,11 +794,20 @@ interface GKLocalPlayer
 		[Export ("unregisterAllListeners")]
 		void UnregisterAllListeners ();
 
+		[Deprecated (PlatformName.iOS, 13, 4, message: "Use 'FetchItemsForIdentityVerificationSignature' along with 'GKPlayer.TeamPlayerId' instead.")]
+		[Deprecated (PlatformName.TvOS, 13, 4, message: "Use 'FetchItemsForIdentityVerificationSignature' along with 'GKPlayer.TeamPlayerId' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, 4, message: "Use 'FetchItemsForIdentityVerificationSignature' along with 'GKPlayer.TeamPlayerId' instead.")]
+		[Deprecated (PlatformName.WatchOS, 6, 2, message: "Use 'FetchItemsForIdentityVerificationSignature' along with 'GKPlayer.TeamPlayerId' instead.")]
```
spouliot: why the `along with 'GKPlayer.TeamPlayerId'` ?
that's not something used as an argument
how does that differ from the old API ?

dalexsoto: Looks like I am not at paraphrasing 

```
API_DEPRECATED("API deprecated. Use fetchItemsForIdentityVerificationSignature: and the teamPlayerID value to verify a user identity.", ios(7.0, 13.5), tvos(9.0, 13.5), macosx(10.10, 10.15.5), watchos(3.0,6.5));
```

This is because in code you use the callback params (publicKeyUrl, signature, salt, and timestamp) along with TeamPlayerId to create a payload and verify the user against a web service.

I guess that goes beyond the deprecation API message and can remove if you think is best.

spouliot: I'd rather stick with our default `Use 'X' instead.` unless the information is complete enough to avoid a trip to the API documentation (ours or Apple).

dalexsoto: @spouliot fixed

PR #7870, opened 2020-02-12
[Harness] Add support for NUnit V3 in the new bcl tests.

mandel-macaque:
Add the required changes in the xunit runner to write NUnit V3 xml
result files so that we can have attachments in VSTS.


tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
```diff
@@ -828,7 +830,13 @@ public override void WriteResultsToFile (TextWriter writer, Jargon jargon)
 						writer.WriteLine ($"{e}");
 					}
 					break;
-
+				case Jargon.NUnitV3:
+					try {
+						Transform_Results ("NUnit3Xml.xslt", assembliesElement, xmlWriter);
+					} catch (Exception e) {
+						writer.WriteLine ($"{e}");
```
spouliot: not sure if the compiler is smart enough to remove the `String.Format` call but it's easier on the eye to just
```csharp
writer.WriteLine (e);
```

PR #7871, opened 2020-02-12
[mtouch] Fix looking up satellites for root assemblies without a path.

rolfbjarne:
This may happen when a root assembly is in the current directory.

This fixes the bug-13945 test in the scripted tests.


dalexsoto:
@monojenkins backport d16-6

dalexsoto:
@rolfbjarne is this also needed in xcode11.4? or the realpath workaround Is enough? 

If this is the real fix lets bring it to xcode11.4

rolfbjarne:
@dalexsoto 

> @rolfbjarne is this also needed in xcode11.4?

No, this broke in ff2d5bb94f9b481be879ee719e21a89fccf56c83, which is not in xcode11.4.

> or the realpath workaround Is enough?

This is unrelated to the realpath workaround.

PR #7874, opened 2020-02-12
[Harness] Add code that will generate custom xml results with failures.

mandel-macaque:
There are two cases that are not currently considered in VSTS that we
need to report:

1. The application could not be built.
2. The application crashed.

In those cases we do not have xml to be imported by VSTS. What we are
going to do is create 'fake' test runs that will encapsulate a build
error or a crash to be imported by VSTS with the needed attachments.

This code just writes custom errors. The generated results are like the
following:

* [NUnit V2](https://gist.github.com/mandel-macaque/7207f6acb5a90895adf84f40a05e0b21)
* [NUnit V3](https://gist.github.com/mandel-macaque/99898c0570cde2e54e38d8af5991f824)
* [xUnit](https://gist.github.com/mandel-macaque/4591cc06d3db0e46de6a0bf886804f13)

### DIFF
--- a/tests/xharness/XmlResultParser.cs
+++ b/tests/xharness/XmlResultParser.cs
@@ -522,5 +522,211 @@ namespace xharness {
 
 			doc.Save (destination);
 		}
+
+		static void WriteAttributes (XmlWriter writer, params (string name, string data) [] attrs)
+		{
+			foreach (var (name, data) in attrs ) {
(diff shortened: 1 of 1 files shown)


mandel-macaque:
@spouliot right now, nothing, I don't want to make too complicated PRs with lots of code that is later hard to review, there are two smaller ones that add the calls (cause the have some logic in them).

PR #7877, opened 2020-02-12
[msbuild] Tweak the tests.

rolfbjarne:
This is a series of commits that tweaks the msbuild tests a little bit.

This PR is probably best reviewed commit-by-commit.

* [msbuild] Fix the MyiOSFrameworkBinding test project to have the same
  assembly name as the project itself.
* [msbuild] Unify output path.
* [msbuild] Simplify test project configuration a bit.
* [msbuild] Update test projects to use 64-bit architectures.
* [msbuild] Clean up project files a bit.
* [msbuild] Add the extra argument for the AppWithExtraArgumentThatOverrides
  test project to all configurations.
* [msbuild] Match configuration between MyWatchKit2Extension and
  MyWatchKit2IntentsExtension.

### DIFF
--- a/msbuild/tests/Xamarin.iOS.Tasks.Tests/ProjectsTests/TVOS/TVApp.cs
+++ b/msbuild/tests/Xamarin.iOS.Tasks.Tests/ProjectsTests/TVOS/TVApp.cs
@@ -2,18 +2,18 @@
 
 namespace Xamarin.iOS.Tasks
 {
-	[TestFixture("TV", "iPhone")]
-	[TestFixture("TVSimulator", "iPhoneSimulator")]
+	[TestFixture ("iPhone")]
+	[TestFixture ("iPhoneSimulator")]
 	public class TVAppTests : ExtensionTestBase
 	{
-		public TVAppTests (string bundlePath, string platform) : base(bundlePath, platform)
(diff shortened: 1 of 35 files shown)


msbuild/tests/Xamarin.iOS.Tasks.Tests/ProjectsTests/TVOS/TVApp.cs
```diff
@@ -2,18 +2,18 @@
 
 namespace Xamarin.iOS.Tasks
 {
-	[TestFixture("TV", "iPhone")]
-	[TestFixture("TVSimulator", "iPhoneSimulator")]
```
mandel-macaque: SO, the file is called TVOS/TVApp, and we are deleting the TV and TVSimulator, is that correct?

rolfbjarne: This is best explained by viewing the relevant commit only: https://github.com/xamarin/xamarin-macios/pull/7877/commits/b1fcce426f6055b76f87426a606763a139dc6042, in short it's because I changed the OutputPath in the csproj to be iPhone[Simulator] instead of TV[Simulator].

PR #7878, opened 2020-02-12
Make the smcs script obsolete, and don't use it in tests anymore.

rolfbjarne:
Don't remove the entire script, because I believe there's code out there that
checks for the existence of the smcs script to determine whether Xamarin.iOS
is installed or not.


rolfbjarne:
Test failure is because the test does a string search for `mcs` of build output, and this branch is contains `mcs`, thus the test finds the string and fails.

Issue #7888, opened 2020-02-13
[VSTS] xUnit tests fail to be imported because they are missing an ID.


mandel-macaque:
The xUnit xml does not provide a unique id for a test, which means that the xslt does not add an id attribute. This results in an issue with the test importer that prints the following warning:
```
##[warning]Failed to publish test results: AutomatedTestName must be specified for automated test runs when neither TestPointId nor TestCaseId has been specified..
```
resulting in incomplete test results.

PR #7889, opened 2020-02-13
[XHarness] Add ids to the xunit to nunit transformed tests.

mandel-macaque:
The NUnit3 xml requires an id for the tests, but xunit does not provide
one. Add an extension object to the xslt that will return a very lame id
so that the test uploader does not complain.

fixes: https://github.com/xamarin/xamarin-macios/issues/7888

### DIFF
--- a/tests/bcl-test/templates/common/TestRunner.xUnit/NUnit3Xml.xslt
+++ b/tests/bcl-test/templates/common/TestRunner.xUnit/NUnit3Xml.xslt
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="UTF-8" ?>
-<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+﻿<?xml version="1.0" encoding="UTF-8" ?>
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:myGenerator="urn:hash-generator">
   <xsl:output cdata-section-elements="message stack-trace"/>
 
   <xsl:template match="/">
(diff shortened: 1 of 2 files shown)


tests/bcl-test/templates/common/TestRunner.xUnit/NUnit3Xml.xslt
```diff
@@ -1,13 +1,13 @@
...
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:myGenerator="urn:hash-generator">
   <xsl:output cdata-section-elements="message stack-trace"/>
 
   <xsl:template match="/">
     <xsl:apply-templates/>
   </xsl:template>
 
   <xsl:template match="assemblies">
-    <test-run name="Test results" inconclusive="0" asserts="0" > 
+    <test-run name="Test results" inconclusive="0" asserts="0" id="2">
```
mandel-macaque: NUnit3 docs say it is harcoded to 2 :/

tests/bcl-test/templates/common/TestRunner.xUnit/XUnitTestRunner.cs
```diff
@@ -8,13 +8,24 @@
...
+	{
+		public int GenerateHash (string name)
+		{
+			var sum = 0;
+			foreach (var c in name) {
+				sum += c;
+			}
+			return sum;
+		}
+	}
```
mandel-macaque: Lame, suggestions welcome, has to be an int and positive.

mandel-macaque:
@rolfbjarne no idea, I think is better to try and think of a good hash that can be computed and always returns the same value.

mandel-macaque:
build

PR #7900, opened 2020-02-14
[d16-5] Replace removed dsymutil `-t` switch with `-num-threads`

dalexsoto:


dalexsoto:
Known failure: HE0038

PR #7906, opened 2020-02-14
[Harness] If app does not launch or crash, create a test failure.

mandel-macaque:
Make sure that when we are running on VSTS we do get test failures when
the application cannot be launched or crashes.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -859,8 +859,20 @@ namespace xharness
 								crash_reason = node?.SelectSingleNode ("reason")?.InnerText;
 							}
 						}
-						if (crash_reason != null)
+						if (crash_reason != null) {
+							// if in CI, do write an xml error that will be picked as a failure by VSTS
+							if (Harness.InCI) {
+								// VSTS does not provide a nice way to report build errors, create a fake
+								// test result with a failure in the case the build did not work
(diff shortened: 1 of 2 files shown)


tests/xharness/AppRunner.cs
```diff
@@ -859,8 +859,20 @@ public async Task<int> RunAsync ()
...
+							if (Harness.InCI) {
+								// VSTS does not provide a nice way to report build errors, create a fake
+								// test result with a failure in the case the build did not work
+								var crashXmlTmp = Logs.Create ($"nunit-crash-{Harness.Timestamp}.tmp", "Build Log tmp");
+								var crashLogXml = Logs.Create ($"nunit-crash-{Harness.Timestamp}.xml", Log.XML_LOG);
+								XmlResultParser.GenerateFailure (crashXmlTmp.FullPath, "AppCrash", $"App crashed {crash_reason}.", crash_reports.Log.FullPath, XmlResultParser.Jargon.NUnitV3);
+								// add the required attachments and the info of the application that failed to install
+								var crash_logs = Directory.GetFiles (Logs.Directory).Where (p => !p.Contains ("nunit")); // all logs but ourself
+								XmlResultParser.UpdateMissingData (crashXmlTmp.FullPath, crashLogXml.FullPath, $"{appName} {Variation}", crash_logs);
+							}
```
wachaudh: I think it would be a good idea to abstract this into a function since most of it is the same instead of copy/pasting 3 times.

mandel-macaque: I promise a PR with that.

PR #7907, opened 2020-02-14
[Devops] Remove redundant zip step.

mandel-macaque:
PublishBuildArtifacts@1 will already zip and publish the artifact, there
is no need for the zip step, it is redundant.


spouliot:
@mandel-macaque you should run it on the internal bots before merging (just in case)

mandel-macaque:
Works, but not quite, and this step might be re-added in the matrix work. Closing for now

Issue #7904, opened 2020-02-14
MM0000: Unexpected error -> when using `--xml=${ProjectDir}` and `-v -v -v -v`


spouliot:
Split from https://github.com/xamarin/xamarin-macios/issues/7897

### Steps to Reproduce

1. Create a XM project
2. Add a linker definition file (.xml)
3. Add `--xml=file` to the additional mmp arguments
4. Add `-v -v -v -v` to the additional mmp arguments

### Expected Behavior

Success

### Actual Behavior

`MM0000`

https://github.com/xamarin/xamarin-macios/issues/7897#issuecomment-586413915

spouliot:
c.c. @tipa

spouliot:
```
MMP: Error MM2004: Extra linker definitions file 'mac/${ProjectDir}/linker.xml' could not be located. (MM2004)
```

I cannot duplicate the `MM0000` but it looks like `${ProjectDir}` is **not** expanded correctly, so what's given to `mmp` is incorrect.

@tipa try adding the full path (or a relative one) to the XML file.

tipa:
Maybe you can reproduce the issue with this example project?
[LinkerProblem.zip](https://github.com/xamarin/xamarin-macios/files/4207197/LinkerProblem.zip)

It seems to work when using a relative Path like this: `--xml=LinkDescription.xml`
Is the mmp command always executed in the current project directory or could the relative path lead to problems when building via e.g. Azure Pipelines?

spouliot:
I do not get the `MM0000` with the attached project - but the stack trace suggest it was an incorrect message format (and could have been fixed). I'll try again later with an older XM version.


With XI a `.csproj` with 
```
<MtouchExtraArgs>-v -v -xml=${ProjectDir}/extra-linker-defs.xml</MtouchExtraArgs>
```
result in the argument
```
 '-xml=/Users/poupou/git/d16-6/xamarin-macios/tests/linker/ios/link sdk/extra-linker-defs.xml'
```

but with XM (attached) `.csproj`
```
<MonoBundlingExtraArgs>--xml=${ProjectDir}/LinkDescription.xml</MonoBundlingExtraArgs>
```
results in 
```
/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/mmp @/Users/poupou/Downloads/LinkerProblem/LinkerProblem/obj/Debug/response-file.rsp   --xml=${ProjectDir}/LinkDescription.xml  -q
```
[...]

PR #7908, opened 2020-02-14
[mtouch/mmp] Share verbosity code.

rolfbjarne:
Share the logging/verbosity code between mtouch and mmp.

### DIFF
--- a/tools/common/Driver.cs
+++ b/tools/common/Driver.cs
@@ -23,6 +23,8 @@ namespace Xamarin.Bundler {
 	public partial class Driver {
 		static void AddSharedOptions (Application app, Mono.Options.OptionSet options)
 		{
+			options.Add ("v|verbose", "Specify how verbose the output should be. This can be passed multiple times to increase the verbosity.", v => Verbosity++);
+			options.Add ("q|quiet", "Specify how quiet the output should be. This can be passed multiple times to increase the silence.", v => Verbosity--);
 			options.Add ("sdkroot=", "Specify the location of Apple SDKs, default to 'xcode-select' value.", v => sdk_root = v);
 			options.Add ("no-xcode-version-check", "Ignores the Xcode version check.", v => { min_xcode_version = null; }, true /* This is a non-documented option. Please discuss any customers running into the xcode version check on the maciosdev@ list before giving this option out to customers. */);
 			options.Add ("warnaserror:", "An optional comma-separated list of warning codes that should be reported as errors (if no warnings are specified all warnings are reported as errors).", v =>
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Test failures are unrelated:

* MTouch tests/NUnit: https://github.com/xamarin/maccore/issues/2154

tools/common/Driver.cs
```diff
@@ -166,8 +168,51 @@ static void AddSharedOptions (Application app, Mono.Options.OptionSet options)
...
+
+			Console.WriteLine (value);
+		}
+
+		public static void Log (int min_verbosity, string format, params object [] args)
+		{
+			if (min_verbosity > Verbosity)
+				return;
+
+			Console.WriteLine (format, args);
```
spouliot: This is incorrect and already a bug https://github.com/xamarin/xamarin-macios/issues/7904

We can end up with a `format` that has `{x}` inside it and no `args` - since a file could be named `{x}.cs` (even if the issue above is related to a different issue).

My local fix for this is
```diff
diff --git a/tools/mmp/driver.cs b/tools/mmp/driver.cs
index 59fcf750c..84d5a6ab6 100644
--- a/tools/mmp/driver.cs
+++ b/tools/mmp/driver.cs
@@ -1041,7 +1041,10 @@ namespace Xamarin.Bundler {
                        if (min_verbosity > verbose)
                                return;
 
-                       Console.WriteLine (format, args);
+                       if (args.Length > 0)
+                               Console.WriteLine (format, args);
+                       else
+                               Console.WriteLine (format);
                }
 
                static string GenerateMain ()
```

but since you're moving/unifying the code it might be better to fix it inside this PR.

tools/mtouch/mtouch.cs
```diff
@@ -629,7 +589,7 @@ public static string GenerateMain (Target target, IEnumerable<Assembly> assembli
 					sw.WriteLine ("\txamarin_init_mono_debug = {0};", app.PackageManagedDebugSymbols ? "TRUE" : "FALSE");
 					sw.WriteLine ("\txamarin_executable_name = \"{0}\";", assembly_name);
 					sw.WriteLine ("\tmono_use_llvm = {0};", enable_llvm ? "TRUE" : "FALSE");
-					sw.WriteLine ("\txamarin_log_level = {0};", verbose.ToString (CultureInfo.InvariantCulture));
+					sw.WriteLine ("\txamarin_log_level = {0};", Verbosity.ToString (CultureInfo.InvariantCulture));
```
spouliot: since it's an `int` there's not point in supplying a `CultureInfo` here

spouliot:
I'll merge _as-is_ since @rolfbjarne is not here and the test failures are unrelated
I'll do another PR for the things I mentioned in the PR

tools/common/Driver.cs
spouliot: Fix (and test) in https://github.com/xamarin/xamarin-macios/pull/7941

tools/mtouch/mtouch.cs
spouliot: my bad, it's fine - current culture could insert a space (for thousands), should not happen but...

PR #7909, opened 2020-02-14
[mmp] Use relative path in mmp script.

rolfbjarne:
Makes relocating mmp easier.


spouliot:
Failures are legit (not random/unrelated) and needs fixing before merging.

PR #7916, opened 2020-02-17
[VSTS] Ensure that certain tasks do not run when not needed.

mandel-macaque:
If the bot could no get the provisioning profiles installed, there is no
reason for certain tasks to run since they are all going to fail. This
adds A LOT of noise in the pipeline for the monitoring person to check
when there is no reason.


mandel-macaque:
@monojenkins backport d16-6

PR #7917, opened 2020-02-17
[VSTS] Propagate debug to the make rules

mandel-macaque:
If the pipeline is set on debug, to be able to see any issues, propagate
that to the make calls so that we can also get the information there,
else debugging issues with the pipelines + make is really hard.


mandel-macaque:

PR #7920, opened 2020-02-17
Backport ci fixes d16-6

mandel-macaque:
Backport all the CI fixes present in master.


mandel-macaque:
Bump to TouchUnit is needed so that it can uses NUnit v3 xml format.

spouliot:
@mandel-macaque not sure what's the best way to review this. It's likely ok - but I would not see if some commits are missing. Maybe you can add a link to a gist that shows the diff from master ?

mandel-macaque:
@spouliot will do that, I'll add a list of the commits and those that I picked.

rolfbjarne:
I'm not sure we should backport these testing fixes everywhere (in particular to release branches) until they're all complete and verified to be working over time. There are many fixes, and more keep getting added, which makes the whole thing rather confusing and hard to keep track of.

mandel-macaque:
@spouliot as requested, here is a diff of the tests folder, which is the only thing we are changing: https://gist.github.com/mandel-macaque/22b51f03f84163e99327aec09a61634b

mandel-macaque:
Closing, merging this is going to be a nightmare.

PR #7921, opened 2020-02-17
Backport ci fixes xcode11.4

mandel-macaque:


mandel-macaque:
Bump to TouchUnit is needed so that it can uses NUnit v3 xml format.

spouliot:
The target `xcode11.4` branch is closed

I don't think it should be updated for `d16-5-xcode11.4` since too much other things are happening in `master`, making backport difficult. Better to have 2 (than 3) ways to monitor

PR #7925, opened 2020-02-17
[mtouch][mmp] Add a `--warn-on-type-ref=X` option

spouliot:
Using this option it's possible to test for the presence of a type
reference in both pre-linked and post-linked assemblies.

This makes it possible to detect if
* a 3rd party assemblies are using some specific type you would like to avoid;
* a type reference has been removed during the build (e.g. linker)

Notes:
* Custom attributes are encoded differently and not included in the assembly type references metadata.
* Assembly that define a type `X` do not have a reference (but the definition) of the type (and won't be reported).

If either the pre or post-linked warnings are not useful then it's possible
 to add `-nowarn:150x` to exclude the results.

E.g.
* `-nowarn:1502` would not report references in pre-linked assemblies;
* `-nowarn:1503` would not report references in post-linked assemblies;

Finally `-warnaserror:150x` can be used to stop a build that would not
satisfy either the pre or post-linked condition.

* `-warnaserror:1502` would not report references in pre-linked assemblies;
* `-warnaserror:1503` would not report references in post-linked assemblies;

_side note_ this was first done on `d16-6` to ease backports as `master`
has multiple changes for localization. A similar PR will be done for
`master` once merged.

### DIFF
--- a/tools/linker/ScanTypeReferenceStep.cs
+++ b/tools/linker/ScanTypeReferenceStep.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using Mono.Cecil;
+using Mono.Linker.Steps;
+using Xamarin.Bundler;
+
+#if MTOUCH
+using ProductException = Xamarin.Bundler.MonoTouchException;
+#else
+using ProductException = Xamarin.Bundler.MonoMacException;
+#endif
+
+namespace Xamarin.Linker.Steps {
+
+	abstract public class ScanTypeReferenceStep : BaseStep {
+		protected readonly List<string> lookfor;
+
+		protected ScanTypeReferenceStep (List<string> list)
+		{
+			lookfor = list;
+		}
+
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			foreach (var module in assembly.Modules) {
+				foreach (var name in lookfor) {
+					if (module.HasTypeReference (name))
+						Report (name, assembly);
+				}
+			}
+		}
+
+		protected abstract void Report (string typeName, AssemblyDefinition assembly);
+	}
+
+	public class PreLinkScanTypeReferenceStep : ScanTypeReferenceStep {
+
+		public PreLinkScanTypeReferenceStep (List<string> list) : base (list)
+		{
+		}
+
+		protected override void Report (string typeName, AssemblyDefinition assembly)
+		{
+			ErrorHelper.Show (new ProductException (1502, false, "One or more reference(s) to type '{0}' already exists inside '{1}' before linking", typeName, assembly));
+		}
+	}
+
+	public class PostLinkScanTypeReferenceStep : ScanTypeReferenceStep {
+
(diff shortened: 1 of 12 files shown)


tools/linker/ScanTypeReferenceStep.cs
```diff
@@ -0,0 +1,59 @@
...
+
+	public class PreLinkScanTypeReferenceStep : ScanTypeReferenceStep {
+
+		public PreLinkScanTypeReferenceStep (List<string> list) : base (list)
+		{
+		}
+
+		protected override void Report (string typeName, AssemblyDefinition assembly)
+		{
+			ErrorHelper.Show (new ProductException (1502, false, "One or more reference(s) to type '{0}' already exists inside '{1}' before linking", typeName, assembly));
```
wachaudh: The error string belongs in Errors.resx in mmp.csproj. Errors.resx should only be modified in VSMac because VSMac automatically updates Errors.Desiginer.cs. mmp.csproj  should also be built again to update the xlf files.

tools/linker/ScanTypeReferenceStep.cs
```diff
@@ -0,0 +1,59 @@
...
+
+	public class PostLinkScanTypeReferenceStep : ScanTypeReferenceStep {
+
+		public PostLinkScanTypeReferenceStep (List<string> list) : base (list)
+		{
+		}
+
+		protected override void Report (string typeName, AssemblyDefinition assembly)
+		{
+			ErrorHelper.Show (new ProductException (1503, false, "One or more reference(s) to type '{0}' still exists inside '{1}' after linking", typeName, assembly));
```
wachaudh: Error string belongs in Errors.resx.

tools/linker/ScanTypeReferenceStep.cs
spouliot: @wachaudh you did not read my commit message

> side note this was first done on d16-6 to ease backports as master
> has multiple changes for localization. A similar PR will be done for
> master once merged.

and, now that I have your attention, where is the process for above documented ?

tools/linker/ScanTypeReferenceStep.cs
spouliot: same

tools/linker/ScanTypeReferenceStep.cs
wachaudh: working on it 😄

tools/mmp/driver.cs
```diff
@@ -354,6 +354,10 @@ static void Main2 (string [] args)
 					}
 				},
 				{ "link-prohibited-frameworks", "Natively link against prohibited (rejected by AppStore) frameworks", v => { LinkProhibitedFrameworks = true; } },
+				{ "warn-on-type-ref=", "Warn if any of the comma-separated types is referenced by assemblies - both before and after linking", v => {
```
rolfbjarne: @spouliot it would be nice to document the format for the type string (`Namespace:Type`? `Namespace.Type`? Or just `Type`? What about nested types?)

spouliot:
@monojenkins backport to xcode11.4

PR #7928, opened 2020-02-18
[Harness] Create an xml crash file when we do not have a result file.

mandel-macaque:
On devices that cannot reach the host via TCP we do not have a log, this
means that in the if statement needs to have a case for it.

The main problem is that when the device cannot connect to the host, we
do not get a log OR a crash reason from the crash logs. It makes sense
not to have a crash reason, because the app did not crash. In these
sitations, we have to create a xml crash report (since we really do not
know if we can parse the file) that will tell vsts that there was an
issue. Adding the main log will let the monitoring person see the
results of the test run.


mandel-macaque:
Having he crash xml is very important, else we have a mismatch between what we report to vsts (everything went ok) and the reality.

PR #7935, opened 2020-02-18
[WebKit] Update bindings for Xcode 11.4 beta 1

whitneyschmidt:


tests/xtro-sharpie/macOS-WebKit.todo
```diff
@@ -1,21 +1 @@
 !missing-protocol-conformance! WKWebView should conform to NSTextFinderClient (defined in 'WKNSTextFinderClient' category)
```
whitneyschmidt: Any pointers on what needs to be done to make `WKWebView` conform? I noted that AppKit has several ignores like `tests/xtro-sharpie/macOS-AppKit.ignore:!incorrect-protocol-member! NSTextFinderClient::scrollRangeToVisible: is OPTIONAL and should NOT be abstract` but this seems to be a different case...

spouliot: `NSTextView` conforms to `NSTextFinderClient` so it should be similar (binding wise).

The main difference will be that this is for macOS only (so `#if MONOMAC` will be needed in order to compile this for XI)

The other warnings are likely old binding mistakes or because Apple changed them from `@required` to `@optional` (both happens).

dalexsoto:
@whitneyschmidt you might want to update the bindings to Beta 2 before merging this PR

stephen-hawley:
👀

tests/xtro-sharpie/macOS-WebKit.ignore
```diff
@@ -593,3 +593,28 @@
...
+!missing-selector! WKWebView::createWebArchiveDataWithCompletionHandler: not bound
+!missing-selector! WKWebView::findString:withConfiguration:completionHandler: not bound
+!missing-selector! WKWebView::pageZoom not bound
+!missing-selector! WKWebView::printOperationWithPrintInfo: not bound
+!missing-selector! WKWebView::setPageZoom: not bound
+!missing-type! WKFindConfiguration not bound
+!missing-type! WKFindResult not bound
+!missing-type! WKPDFConfiguration not bound
+
+# TODO mediaType is not available in 11.4 beta2 even though it is marked as such
```
whitneyschmidt: Not sure whether we normally create GH issues for this/the TODO above? I will create issues if that's the standard practice :)

spouliot: In general we don't need issue if it's in the `.ignore` file - since it's already a list of issues

However mismatch between Apple's header and implementation should be filled with Apple (and you create a shadow issue in maccore for tracking). That way whenever it's fixed we'll know what needs to be done on our side to close the loop.

src/wkwebkit.cs
```diff
@@ -749,32 +748,51 @@ interface WKWebView
 		[Export ("handlesURLScheme:")]
 		bool HandlesUrlScheme (string urlScheme);
 
-		[Mac (10, 15), iOS (13, 0)]
+		// TODO
+		// 11.4 beta2:
+		// WK_CLASS_AVAILABLE(macos(WK_MAC_TBA), ios(WK_IOS_TBA))
```
spouliot: there's a newer beta of the OS (but not a newer Xcode). Have you tried with a beta 3 device (or macOS) ?

tests/xtro-sharpie/macOS-WebKit.ignore
mandel-macaque: @whitneyschmidt example of such an issue: https://github.com/xamarin/maccore/issues/2065

whitneyschmidt:
Closing this as many of the `TBA` bindings have been completely removed in beta3.

PR #7936, opened 2020-02-19
[Harness] If there is no BuildTask, do not add logs.

mandel-macaque:
xharness does not only allow to run tests in CI, but also helps to run
specific project. That feature is used for the mtouch tests. In that
scenario, we do not have a build task and therefore we will not have the
build logs.

fixes: https://github.com/xamarin/maccore/issues/2154


mandel-macaque:
build

mandel-macaque:
build

PR #7937, opened 2020-02-19
[Harness] Add extra labels to allow to select a smaller set of tests.

mandel-macaque:
It will be nice to be able to have a smaller set of tests to be ran by
xharness. In this commit we add a set of labels to be able to divide the
test in the following groups:

* Xamarin - all tests BUT monotouch.
* Monotouch - just monotouch tests.
* Old BCL - The BCL tests that are based on NUnit.
* New BCL - The BCL tests that are based on xUnit.
* mscorlib - Just the test that exercise mscorlib (and the different
groups, mscorlib 1, mscorlib 2 etc..)

This will allow to parallelize the execution of the full test suite in
different agents in VSTS.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -40,6 +40,11 @@ namespace xharness
 		public bool IncludeXtro;
 		public bool IncludeCecil;
 		public bool IncludeDocs;
+		public bool IncludeNewBCL;
+		public bool IncludeOldBCL;
+		public bool IncludeMscorlib;
+		public bool IncludeXamarin = true;
+		public bool IncludeMonotouch = true;
 
 		public bool CleanSuccessfulTestRuns = true;
 		public bool UninstallTestApp = true;
@@ -219,8 +224,19 @@ namespace xharness
 		{
 			if (!project.IsExecutableProject)
 				return false;
+			
+			if (project.IsBclTest) {
+				if (!project.IsNewBclTest)
+					return IncludeBcl || IncludeOldBCL;
+				if (project.IsMscorlib) 
+					return IncludeMscorlib;
+				return IncludeBcl || IncludeNewBCL;
+			}
 
-			if (!IncludeBcl && project.IsBclTest)
+			if (!IncludeMonotouch && project.IsMonotouch)
+				return false;
+
+			if (!IncludeXamarin && !project.IsBclTest && !project.IsMonotouch)
 				return false;
 
 			if (Harness.IncludeSystemPermissionTests == false && project.Name == "introspection")
@@ -799,6 +815,9 @@ namespace xharness
 			SetEnabled (labels, "mtouch", ref IncludeMtouch);
 			SetEnabled (labels, "mmp", ref IncludeMmpTest);
 			SetEnabled (labels, "bcl", ref IncludeBcl);
+			SetEnabled (labels, "new-bcl", ref IncludeNewBCL);
+			SetEnabled (labels, "old-bcl", ref IncludeOldBCL);
+			SetEnabled (labels, "mscorlib", ref IncludeMscorlib);
 			SetEnabled (labels, "btouch", ref IncludeBtouch);
 			SetEnabled (labels, "mac-binding-project", ref IncludeMacBindingProject);
 			SetEnabled (labels, "ios-extensions", ref IncludeiOSExtensions);
@@ -817,6 +836,9 @@ namespace xharness
 			SetEnabled (labels, "mac", ref IncludeMac);
 			SetEnabled (labels, "ios-msbuild", ref IncludeiOSMSBuild);
 			SetEnabled (labels, "ios-simulator", ref IncludeSimulator);
+			SetEnabled (labels, "xamarin", ref IncludeXamarin);
+			SetEnabled (labels, "monotouch", ref IncludeMonotouch);
+
 			bool inc_permission_tests = false;
 			if (SetEnabled (labels, "system-permission", ref inc_permission_tests))
 				Harness.IncludeSystemPermissionTests = inc_permission_tests;
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -40,10 +40,10 @@ namespace xharness
 		public bool IncludeXtro;
 		public bool IncludeCecil;
 		public bool IncludeDocs;
-		public bool IncludeNewBCL;
-		public bool IncludeOldBCL;
+		public bool IncludeBCLxUnit;
+		public bool IncludeBCLNUnit;
 		public bool IncludeMscorlib;
-		public bool IncludeXamarin = true;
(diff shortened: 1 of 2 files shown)


spouliot:
> This will allow to parallelize the execution of the full test suite in different agents in VSTS.

Are all the tests executable from a package ?
We do not want parallelization to require each bot to build XI and XM.

Also not sure we need **labels** to match how the bots would be splitting the tests. Labels are used (right now) to override the default test logic (add/remove some).

> Xamarin - all tests BUT monotouch.
> Monotouch - just monotouch tests.

Not sure it's worth splitting.
* A change that affect one pretty much needs to be tested against the other.
* Need time (to run) vs time to provision a bot. If we end up with more time spent provisioning then we're slowing everything;

> Old BCL - The BCL tests that are based on NUnit.

Not sure it's worth splitting (with New BCL) but, if so, let's call a cat a cat - `BCL-NUnit`

> New BCL - The BCL tests that are based on xUnit.

Not sure it's worth splitting (with New BCL) but, if so, let's call a dog a dog - `BCL-xUnit`. We don't want the "New BCL" to become the *New iPad*

> mscorlib - Just the test that exercise mscorlib (and the different groups, mscorlib 1, mscorlib 2 etc..)

Not sure it's worth splitting (with BCL) - if there's a mono bump then all BCL (and mscorlib) should be run. Again it does down to timings and bots availability (which changes over time) so I don't like this being tied to labels.

mandel-macaque:
The split is not meant to be used in PRs or builds, but on device tests. The XI and XM packages are not rebuilt for each split. The overall CI works as follows:

```
- Merge 
  |-------> Jenkins - Creates XI and XM pkg
                       |
                       ---------> Single pipeline
                                           |
                                           ------> iOS stage
                                                         |
                                                         -------> Xamarin ----> |
                                                         -------> Monotouch --> |
                                                         ...                    |
                                           ------> tvOS stage                   |
                                                        |                       |
                                                        --------> Xamarin ----> |
                                                        --------> Monotocuh --> |
                                                                                |
                                                                                ----> Fan in and publish result
```
Each of the stages has a number of jobs, the jobs do not compile the XI or XM packages, those package are compiled by jenkins in the initial step, uploaded and the location of the pkgs is passed to the pipeline via two runtime variables. It is true that each of the jobs has to provision and install the dependencies, but provisionator is smart enough to cache all that information, meaning that we will only hit the penalty or re-downloading Xcode in a very few cases, for example, the times of a cached execution are:

- Add provisioning profiles: 18 seconds
- Provision dependencies: 3 minutes 
- Install XI: 1 minute

Once the above is done, most if not all of the time of the agent is spent compiling, deploying and executing the tests.

The split is done to fulfill two goals:

- try to block an agent the smaller amount of time possible.
- ensure that we do not have more than 100 test runs per agent (which is a limitation added by the publishing of the xml test results, is a limit imposed by the task if we want to keep attachments usable).

The current execution times of the sliced tests on well behave agents are:

- Xamarin: ~ 1 hour
- Monotouch: ~ 45 mins
- xUnit BCL: ~1 hours
- NUnit: ~ 1 hour
- mscorlib: ~1 hour

In an ideal world, we could do everything in under 2 hours, but agent availability plays a part, with the current capacity we have we are seeing results in around 2 hours to 2 hours 30 min.

The entire pipeline execution represent A SINGLE commit, and if any of the jobs fails, the result sent to the monitoring person will be a failure. There is no possible way to have a mismatch between the XI executed in a job and the one if another job, since that information is the same for all.

The default behavior should be left as it is, what is more, we should not let PRs use ANY of this labels, since they will be skipping tests they should not (although the can use the to run extra tests if wanted). The usage of stages has just be done to simplify the yaml templating and to add a logic grouping to the jobs, that is, all jobs for the same devices are executed in the same stage.

Naming: I suck at naming, I was going to call my dogs Mac  & Cheese, I'm happy to use any name (I used group 1, group2, etc.. in the BCL tests apps 😆 )

mandel-macaque:
@spouliot forgot to add a reference build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3483060&view=results

tests/xharness/TestProject.cs
```diff
@@ -94,6 +94,14 @@ public TestProject GetTodayExtension ()
 			}
 		}
 
+		public bool IsMonotouch => Name.Contains ("monotouch");
```
stephen-hawley: will the case ever change and will there ever be naming conflicts that would break this?
For example `monotouchless`

mandel-macaque: No, but in that case, it would not hurt us. Will just run the two tests in the same job.

tests/xharness/Jenkins.cs
```diff
@@ -40,6 +40,11 @@ public class Jenkins
 		public bool IncludeXtro;
 		public bool IncludeCecil;
 		public bool IncludeDocs;
+		public bool IncludeNewBCL;
```
rolfbjarne: I prefer Sebastien's naming (BCL-nunit and BCL-xunit)

tests/xharness/Jenkins.cs
```diff
@@ -219,8 +224,19 @@ bool IsIncluded (TestProject project)
 		{
 			if (!project.IsExecutableProject)
 				return false;
+			
+			if (project.IsBclTest) {
+				if (!project.IsNewBclTest)
+					return IncludeBcl || IncludeOldBCL;
+				if (project.IsMscorlib) 
+					return IncludeMscorlib;
+				return IncludeBcl || IncludeNewBCL;
```
rolfbjarne: This means the mscorlib tests are not included when only `IncludeBcl` is true. Is that expected?

tests/xharness/Jenkins.cs
```diff
@@ -219,8 +224,19 @@ bool IsIncluded (TestProject project)
...
+			}
 
-			if (!IncludeBcl && project.IsBclTest)
+			if (!IncludeMonotouch && project.IsMonotouch)
+				return false;
+
+			if (!IncludeXamarin && !project.IsBclTest && !project.IsMonotouch)
```
rolfbjarne: `!project.IsBclTest` is redundant, because `IsBclTest` will always be false at this point.

rolfbjarne: I don't like the name `Xamarin` for all tests except monotouch-test, it's very unintuitive. What about something more awkward, but understandable, like `NonMonoTouch`, so you'd end up with `IncludeNonMonotouch`, `IsNonMonotouch`, etc?

tests/xharness/Jenkins.cs
mandel-macaque: Yes, I wanted the mscorlib tests to be included explicitly.

tests/xharness/Jenkins.cs
mandel-macaque: yeah, no problem, as I said, I'm terrible at naming :)

PR #7941, opened 2020-02-19
[mtouch][mmp] Fix possible exception when logging strings with `{x}`

spouliot:
We can end up with a format that has `{x}` inside it and no argument. If
this is logged (which depends on verbosity level) then we end up with an
uncaught exceptions and either `MT0000` or `MM0000` errors.

This happens in #7904 due to other, unrelated, issues (a variable is not
expanded) but this can happen in other circumstances, e.g. a file could
be named `{x}.cs`.

The easy fix is to use the right `Console.WriteLine` overload if there
are no arguments provided. This is always good since it avoid an non
required call to `String.Format`.

Fixes https://github.com/xamarin/xamarin-macios/issues/7904

### DIFF
--- a/tests/common/BundlerTest.cs
+++ b/tests/common/BundlerTest.cs
@@ -328,5 +328,37 @@ class Issue4072Session : NSUrlSession {
 				bundler.AssertWarningCount (1);
 			}
 		}
+
+		[Test]
+#if __MACOS__
+		[TestCase (Profile.macOSMobile)]
+#else
+		[TestCase (Profile.iOS)]
+#endif
+		public void XmlUnexpandedVariable_7904 (Profile profile)
+		{
+			using (var bundler = new BundlerTool ()) {
+				var code = @"
+ using System;
+ using Foundation;
+ using ObjCRuntime;
+ class T {
+	static void Main ()
+	{
+		Console.WriteLine (typeof (NSObject));
+	}
+ }
+ ";
+				bundler.Profile = profile;
+				bundler.CreateTemporaryCacheDirectory ();
+				bundler.CreateTemporaryApp (profile, code: code);
+				// typo is intentional since `{ProjectDir}` should be expanded (different issue)
+				bundler.CustomArguments = new [] { "--xml=${xProjectDir}/linker.xml", "-v", "-v", "-v", "-v" };
+				bundler.AssertExecuteFailure ();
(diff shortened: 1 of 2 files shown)


spouliot:
```
14:58:02 AppRunner.cs(375,22): error CS1503: Argument 1: cannot convert from 'string[]' to 'System.Collections.Generic.IEnumerable<xharness.Log>' [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/xharness.csproj]
14:58:02 AppRunner.cs(377,23): error CS1503: Argument 1: cannot convert from 'string[]' to 'System.Collections.Generic.IEnumerable<xharness.Log>' [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/xharness.csproj]
14:58:02 AppRunner.cs(41,17): warning CS0649: Field 'AppRunner.AppPath' is never assigned to, and will always have its default value null [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/xharness/xharness.csproj]
```
PR to fix this is inflight, will re-kick build afterward

mandel-macaque:
@monojenkins rebase

rolfbjarne:
The analysis isn't quite right (the fix should still be OK, but unnecessary).

The problem is that this line:

https://github.com/xamarin/xamarin-macios/blob/0d8fe219c727fc68d495c26823070b510a4aa474/tools/common/Driver.cs#L478

ends up calling a Log overload that take format arguments, as evidenced by the stack trace:

```
at Xamarin.Bundler.Driver.Log (System.String format, System.Object[] args) [0x00001] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tools/mmp/driver.cs:1023 
at Xamarin.Bundler.Driver.LogArguments (System.String[] arguments, System.Int32 indentation) [0x00028] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tools/common/Driver.cs:478 
```

The real fix for the the problem is to provide a Log overload that doesn't take any format arguments - which happened in PR #7908, because mtouch already had such an overload, and sharing code with mmp brought the fix there too: https://github.com/xamarin/xamarin-macios/blob/96d5d3e02b3febb291e28af9f9ee58324ee61a81/tools/common/Driver.cs#L192-L195

Anyways, as I mentioned before, this change is fine.

PR #7947, opened 2020-02-20
[Harness] Provide device name in those errors where is device specific.

mandel-macaque:
In order to simplify the monitoring job add the device name to the
following failures:

* Installation
* Launch
* Tcp Connection

All the above are most of the time due to a misconfigured device. The
device name is useful information for the monitoring person to be able
to reach IT and address the issue.


mandel-macaque:
buld

PR #7950, opened 2020-02-21
[Harness] Do not add the variation two times to failure messages.

mandel-macaque:
AppName should only be the app name since we are passing the varation as
a parameter too. Else we end up with $"{appname} {variation} {variation}".


mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1083

PR #7951, opened 2020-02-21
[Harness] Add start-time to work around a bug in the publishing tool.

mandel-macaque:
The publishing tool is a little fragile. In run
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3490479&view=logs&j=67d14776-f827-5fe4-2625-2db4b5987fd1&t=fa262eec-9d97-5ba4-b4cc-a9292beecd8f
I noticed that valid test runs with a failing test (launch issues) were
not being uploaded.

I found out that the reason is a flaw in the logic on the parser of the
publishing tool. The tool assumes, that if there is no start-time, there
are no test results (do remember that NUnitV3 is schemaless we don't
know exactly what attrs are compulsory).

The culprint is line: https://dev.azure.com/mseng/AzureDevOps/_git/AzureDevOps?path=%2FTa%2FTasks%2FPublishTestResults%2FParser%2FNUnitResultParser.cs&version=GBmaster&line=473&lineEnd=473&lineStartColumn=63&lineEndColumn=64&lineStyle=plain

Basically:

```csharp
 if (testRunNode?.Attributes?["start-time"] != null) {
   // import test data
 }

 // do nothing interesting since there is no data
```

This commit fixes it by setting the start time as the current one, we
dont care since it is a failure xml result.

### DIFF
--- a/tests/xharness/XmlResultParser.cs
+++ b/tests/xharness/XmlResultParser.cs
@@ -627,6 +627,7 @@ namespace xharness {
 
 		static void GenerateNUnitV3Failure (XmlWriter writer, string title, string message, StreamReader stderr)
 		{
+			var date = DateTime.Now;
 			writer.WriteStartElement ("test-run");
 			// defualt values for the crash
 			WriteAttributes (writer,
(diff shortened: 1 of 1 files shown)


mandel-macaque:

tests/xharness/XmlResultParser.cs
```diff
@@ -640,7 +640,8 @@ static void GenerateNUnitV3Failure (XmlWriter writer, string title, string messa
 				("inconclusive", "0"),
 				("skipped", "0"),
 				("asserts", "1"),
-				("date", XmlConvert.ToString (DateTime.Now, "yyyy-MM-dd"))
+				("run-date", XmlConvert.ToString (DateTime.Now, "yyyy-MM-dd")),
+				("start-time", DateTime.Now.ToString ("HH:mm:ss"))
```
spouliot: minor: you can (bots are good at this) have a `run-date of `X` for `X @ 23:59:59` and a a very early (past) `start-time`, e.g. `00:00:00`, for `X+1 @ 00:00:00`. Depending on how the data is gathered/analyzed this could generate illogical data. Using the same instance of `DateTime.Now` ensure both would be in sync.

mandel-macaque: will do.

PR #7955, opened 2020-02-21
[mtouch][mmp] Share TimeStampStep code in the linker pipelines

spouliot:
This makes the `--time` options more useful to see inside the linker's
pipeline steps.

This was used only in `mtouch` before (and missed some steps [1]).
Now it's shared between both tools.

[1] not sure why... merge issue/backport maybe ?

### DIFF
--- a/tools/mmp/Tuning.cs
+++ b/tools/mmp/Tuning.cs
@@ -110,61 +110,62 @@ namespace MonoMac.Tuner {
 		{
 			var pipeline = new Pipeline ();
 
-			pipeline.AppendStep (options.LinkMode == LinkMode.None ? new LoadOptionalReferencesStep () : new LoadReferencesStep ());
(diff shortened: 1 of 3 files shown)


tools/mmp/Tuning.cs
```diff
@@ -110,62 +110,63 @@ static Pipeline CreatePipeline (LinkerOptions options)
...
 
 			// expect that changes can occur until it's all saved back to disk
 			if (options.WarnOnTypeRef.Count > 0)
-				pipeline.AppendStep (new PostLinkScanTypeReferenceStep (options.WarnOnTypeRef));
+				pipeline.Append (new PostLinkScanTypeReferenceStep (options.WarnOnTypeRef));
 
 			return pipeline;
 		}
 
+
```
wachaudh:
```suggestion
```

wachaudh: noise^

PR #7961, opened 2020-02-24
[Harness] Use a tcp tunnel to communicate when running iOS tests.

mandel-macaque:
The tcp connection is very fragile and has been giving a lot of problems
in the labs. With the landing of
https://github.com/xamarin/maccore/pull/2158 and
https://github.com/spouliot/Touch.Unit/pull/56 we can use a tcp tunnel
that will communicate with the device via the usb cable. This means that
we will not longer have communication issues with iOS devices in the
labs.

The tcp tunnel is only used on iOS devices, not simulator, not tvOS or
watchOS.


tests/xharness/SimpleTcpListener.cs
```diff
@@ -23,17 +32,24 @@ public override void Initialize ()
 			server.Start ();
 
 			if (Port == 0)
-				Port = ((IPEndPoint) server.LocalEndpoint).Port;
+				Port = ((IPEndPoint)server.LocalEndpoint).Port;
```
spouliot: minor: removing space breaks coding style

tests/xharness/SimpleTcpListener.cs
```diff
@@ -50,6 +66,55 @@ protected override void Start ()
 				}
 			}
 		}
+		
+		void StartTcpTunnel ()
+		{
+			while (!TunnelInitialized) ; // do nothing until the tunnel is ready
```
spouliot: ^ that's CPU heavy, sleep a few ms ?

tests/bcl-test/templates/common/TestRunner.Core/TcpTextWriter.cs
```diff
@@ -62,25 +64,41 @@ static string SelectHostName (string[] names, int port)
 			return result;
 		}
 
-		public TcpTextWriter (string hostName, int port)
+		public TcpTextWriter (string hostName, int port, bool isTunnel = true)
```
spouliot: Tunnel defaults to `true` here - but it default to `false` somewhere else (confusing to use)

tests/bcl-test/templates/common/TestRunner.Core/TcpTextWriter.cs
```diff
@@ -62,25 +64,41 @@ static string SelectHostName (string[] names, int port)
...
 			UIApplication.SharedApplication.NetworkActivityIndicatorVisible = true;
 #endif
 
 			try {
-				client = new TcpClient (HostName, port);
+				if (isTunnel) {
+					server = new TcpListener (IPAddress.Any, Port);
+					server.Start ();
+					client = server.AcceptTcpClient ();
+					// block until we have the ping from the client side
```
spouliot: block until... what's the caller/harness timeout ?
do we want to wait that long ?
will it report something useful to diagnose ?

tests/bcl-test/templates/common/TestRunner.Core/TcpTextWriter.cs
```diff
@@ -62,25 +64,41 @@ static string SelectHostName (string[] names, int port)
...
+						if (message.Contains ("ping"))
+							break;
+					}
+				} else {
+					client = new TcpClient (HostName, port);
+				}
 				writer = new StreamWriter (client.GetStream ());
-			}
-			catch {
+			} catch {
```
spouliot: ^ breaking correct style

tests/xharness/AppRunner.cs
```diff
@@ -887,7 +923,7 @@ public async Task<int> RunAsync ()
 					FailureMessage = $"Launch failure";
 					if (Harness.InCI)
 						XmlResultParser.GenerateFailure (Logs, "launch", appName, Variation, $"AppLaunch on {device_name}", $"{FailureMessage} on {device_name}", main_log.FullPath, XmlResultParser.Jargon.NUnitV3);
-				} else if ((!File.Exists (listener_log.FullPath) || string.IsNullOrEmpty (crash_reason)) && Harness.InCI) {
+				} else if ((crashed && !File.Exists (listener_log.FullPath) || (!crashed && string.IsNullOrEmpty (crash_reason)) && Harness.InCI)) {
```
spouliot: `crashed` is not assigned (not used) elsewhere in the PR

tests/xharness/SimpleTcpListener.cs
```diff
@@ -10,11 +10,20 @@ namespace xharness
 	public class SimpleTcpListener : SimpleListener
 	{
 		byte[] buffer = new byte [16 * 1024];
+		bool useTcpTunnel = true;
+		public bool TunnelInitialized = false;
 		TcpListener server;
+		TcpClient client;
 
+		public SimpleTcpListener (bool tunnel = false) : base ()
```
spouliot: here tunnel default to false (see earlier when it's true)

tests/xharness/SimpleTcpListener.cs
mandel-macaque: I hate vsmac :/

tests/xharness/SimpleTcpListener.cs
mandel-macaque: Hmm I think I can do better, this was my very first attempt, I can do a asynchronous call, will require some work here and there, but will make everything nicer. Rather than wait seconds or do a thread.sleep, I can do a async completion source that will return the status and then move on.

tests/bcl-test/templates/common/TestRunner.Core/TcpTextWriter.cs
mandel-macaque: hmm you make me use my brain too much... if harness dies, we will have issues. We could do:
```csharp
server = new TcpListener (IPAddress.Any, Port);
server.Server.ReceiveTimeout = 5000;
```
That sets the socket receive timeout in milliseconds.

tests/xharness/AppRunner.cs
mandel-macaque: Not in the diff, but it is set here: https://github.com/xamarin/xamarin-macios/pull/7961/files#diff-03773861bcef485bf343300f31b60b0eL799

Noticed that we needed to check for that value, else we had some crashes wrongly reported. Can be in a diff PR.

tests/xharness/SimpleTcpListener.cs
```diff
@@ -50,6 +66,55 @@ protected override void Start ()
...
 			}
 		}
+		
+		void StartTcpTunnel ()
+		{
+			while (!TunnelInitialized) ; // do nothing until the tunnel is ready
+			bool processed;
+			try {
+				int timeout = 100;
+				var watch = new System.Diagnostics.Stopwatch ();
```
rolfbjarne: This watch is never started.

tests/xharness/AppRunner.cs
```diff
@@ -756,8 +762,38 @@ public async Task<int> RunAsync ()
...
+						tcpArgs.Add ("-v");
+						AddDeviceName (args);
+
+						// use a cancelation token, later will be used to kill the tcp tunnel proces
+						CancellationTokenSource tunnelCancelation = new CancellationTokenSource ();
+						main_log.WriteLine ($"Starting tcp tunnel between mac port: {listener.Port} and devie port {listener.Port}.");
+						var tunnelbackLog = new CallbackLog ((line) => {
+							main_log.WriteLine ($"The tcp tunnel output is {line}");
+							if (line.Contains ("Tcp tunnel started on device")) {
+								main_log.Write ($"Tcp tunnel create on port {listener.Port}");
```
wachaudh:
```suggestion
								main_log.Write ($"Tcp tunnel created on port {listener.Port}");
```

tests/xharness/AppRunner.cs
```diff
@@ -41,6 +41,9 @@ class AppRunner
 		public string AppPath;
 		public string Variation;
 		public BuildToolTask BuildTask;
+		public bool UseTcpTunnel;
+
```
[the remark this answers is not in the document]
mandel-macaque: Yep, but I need all the other branches, I could have add this as a draft PR to be honest.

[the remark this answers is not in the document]
mandel-macaque: comment was for https://github.com/xamarin/xamarin-macios/pull/7961#pullrequestreview-363602174

tests/xharness/SimpleTcpListener.cs
spouliot: It does not have to be complex - but saving energy is good and green :are

tests/xharness/AppRunner.cs
spouliot: ok, if same-PR then make sure it's part of the PR/commit description. Would be easy to miss otherwise

tests/bcl-test/templates/common/TestRunner.Core/TcpTextWriter.cs
```diff
@@ -62,22 +64,40 @@ static string SelectHostName (string[] names, int port)
 			return result;
 		}
 
-		public TcpTextWriter (string hostName, int port)
+		public TcpTextWriter (string hostName, int port, bool isTunnel = false)
```
mandel-macaque: FYI matches the default in Touch.Unit: https://github.com/spouliot/Touch.Unit/pull/56

tests/xharness/AppRunner.cs
mandel-macaque: Added separate PR to unblock another PR: https://github.com/xamarin/xamarin-macios/pull/7979 will merge with master once it lands.

mandel-macaque:
Unrelated test failure: https://github.com/xamarin/maccore/issues/1605

tests/xharness/SimpleTcpListener.cs
```diff
@@ -50,6 +67,58 @@ protected override void Start ()
...
+				}
+				do {
+					client.ReceiveBufferSize = buffer.Length;
+					processed = Processing (client);
+				} while (!AutoExit || !processed);
+			} catch (Exception e) {
+				var se = e as SocketException;
+				if (se == null || se.SocketErrorCode != SocketError.Interrupted)
+					Console.WriteLine ("[{0}] : {1}", DateTime.Now, e);
+			} finally {
```
dalexsoto: Nit + question, we log anything but interrupted and ignore any other exception is this what you intended? 

```suggestion
			} catch (SocketException e) when (e.SocketErrorCode != SocketError.Interrupted) {
				Console.WriteLine ("[{0}] : {1}", DateTime.Now, e);
			} catch (Exception e) {
				// ??
			} finally {
```

mandel-macaque: Yes, at it comes from the previous implementation.

mandel-macaque:
hmm interesting, testing locally the device can create the tcp tunnel, we might have some weird setup in the lab, or something diff. Will need to check the devices :/

mandel-macaque:
build

mandel-macaque:
Looks like the tunnel works, but if we try to make lots of them, the device goes a little funny. I'll address this to see if we can create a single tunnel for all tests. So that we only have to create it once and have the connection done once using the C api. Later do use the manage code to send the messages.

chamons:
Marked do not merge until xharness regressions are nailed down.

mandel-macaque:
Closing in favour of https://github.com/xamarin/xamarin-macios/pull/8217

PR #7969, opened 2020-02-24
[mmp] Don't ignore failures to execute pkg-config and simplify/improve logic.

rolfbjarne:
* Show proper errors if pkg-config fails to execute.
* Extract logic to verify system mono version into its own function.
* Extract logic to execute pkg-config into its own function.

### DIFF
--- a/docs/website/mmp-errors.md
+++ b/docs/website/mmp-errors.md
@@ -776,6 +776,14 @@ You can silence this warning by adding `--nowarn=5221` to the **Additional mmp a
 
 ### MM5312: pkg-config failed with an error code '{code}'. Check build log for details.
 
+<a name="MM5313" />
+
+### MM5313: Could not find pkg-config. Please install the Mono.framework from https://mono-project.com/Downloads
+
+<a name="MM5314" />
+
+### MM5314: Failed to execute pkg-config: '{0}'. Check build log for details.
+
 <!-- MM6xxx: mmp internal tools -->
 <!-- MM7xxx: reserved -->
 
--- a/docs/website/mmp-errors.md
+++ b/docs/website/mmp-errors.md
@@ -740,10 +740,7 @@ You can silence this warning by adding `--nowarn=5221` to the **Additional mmp a
 
 ### MM53xx: other tools
 
-<a name="MM5301" />
-
-#### MM5301: pkg-config could not be found. Please install the Mono.framework from http://mono-project.com/Downloads
-
+<!-- 5301 used by mtouch -->
 <!-- 5302 used by mtouch -->
 <!-- 5303 used by mtouch -->
 <!-- 5304 used by mtouch -->
--- a/tools/mmp/driver.cs
+++ b/tools/mmp/driver.cs
@@ -1321,7 +1321,7 @@ namespace Xamarin.Bundler {
 			var sb = new StringBuilder ();
 			int rv;
 			try {
-				rv = RunCommand (pkg_config, new string [] { option, "mono-2" }, env, sb);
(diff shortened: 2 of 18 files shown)


tools/mmp/driver.cs
```diff
@@ -1311,6 +1308,29 @@ static int Compile ()
...
+			if (!File.Exists (pkg_config))
+				throw ErrorHelper.CreateError (5313, Errors.MX5313);
+
+			if (!IsUnifiedFullSystemFramework && !force_system_mono)
+				env = new [] { "PKG_CONFIG_PATH", Path.Combine (FrameworkLibDirectory, "pkgconfig") };
+
+			var sb = new StringBuilder ();
+			int rv;
+			try {
+				rv = RunCommand (pkg_config, new string [] { option, "mono-2" }, env, sb);
```
mandel-macaque: minor, compiler knows is an array or strings :)

tools/mtouch/xlf/Errors.cs.xlf
```diff
@@ -3002,6 +3002,30 @@
         <note>
 		</note>
       </trans-unit>
+      <trans-unit id="MX5312">
```
filipnavara: OT: Does anyone actually translate those? I didn't even know the translation files are there.

rolfbjarne: > Does anyone actually translate those?

Not yet, but soon.

> I didn't even know the translation files are there.

It's very new (just a few weeks ago), which is why the translations aren't there yet. Step 1 is to make it possible to translate, step 2 is to actually translate :smile:

docs/website/mmp-errors.md
```diff
@@ -776,6 +776,14 @@ You can silence this warning by adding `--nowarn=5221` to the **Additional mmp a
 
 ### MM5312: pkg-config failed with an error code '{code}'. Check build log for details.
 
+<a name="MM5313" />
+
+### MM5313: Could not find pkg-config. Please install the Mono.framework from https://mono-project.com/Downloads
```
spouliot: dupe of

`#### MM5301: pkg-config could not be found. Please install the Mono.framework from http://mono-project.com/Downloads`

sadly localization makes this harder to spot inside PR :(

tools/mmp/driver.cs
```diff
@@ -1063,32 +1081,11 @@ static int Compile ()
...
-				RunCommand (pkg_config, new [] { "--cflags", "mono-2" }, env, cflagsb);
-				RunCommand (pkg_config, new [] { "--variable=libdir", "mono-2" }, env, libdirb);
-				var versionFile = "/Library/Frameworks/Mono.framework/Versions/Current/VERSION";
-				if (File.Exists (versionFile)) {
-					mono_version.Append (File.ReadAllText (versionFile));
-				} else {
-					RunCommand (pkg_config, new [] { "--modversion", "mono-2" }, env, mono_version);
-				}
-			} catch (Win32Exception e) {
-				throw new MonoMacException (5301, true, e, Errors.MM5301);
```
spouliot: that was my hint something could be missing (was actually duped) in the newer code

tools/mtouch/xlf/Errors.cs.xlf
wachaudh: People have been working on step 2 for a few weeks so hopefully there will be something soon.

docs/website/mmp-errors.md
rolfbjarne: I ended up removing the MM5301 error message, because it's not used anymore in mmp, and the same error code is used for something else in mtouch.

PR #7971, opened 2020-02-25
[Xharness] We support diff xml outputs. Allow to pass the desired one.

mandel-macaque:
We support different outputs, lets add the avility for the caller to
decide which one to use. We default to NUnit V3 due to or dependency to
it in VSTS.


tests/xharness/Program.cs
```diff
@@ -72,6 +72,15 @@ public static int Main (string[] args)
 				{ "periodic-command-arguments=", "Arguments to the command to execute periodically.", (v) => harness.PeriodicCommandArguments = v },
 				{ "periodic-interval=", "An interval (in minutes) between every attempt to execute the periodic command.", (v) => harness.PeriodicCommandInterval = TimeSpan.FromMinutes (double.Parse (v)) },
 				{ "include-system-permission-tests:", "If tests that require system permissions (which could cause the OS to launch dialogs that hangs the test) should be executed or not. Default is to include such tests.", (v) => harness.IncludeSystemPermissionTests = ParseBool (v, "include-system-permission-tests") },
+				{ "xml-jargon:", "The xml format to be used for test results. Values can be nunitv2, nunitv3, xunit,", (v) =>
+					{
+						if (Enum.TryParse<Jargon> (v, out var jargon))
+							harness.XmlJargon = jargon;
+						else
+							harness.XmlJargon = Jargon.NUnitV3; // default
```
rolfbjarne: I would show an error here instead, that way invalid values won't actually work.

spouliot: or at least a warning, mentioning something else (than the user-supplied value) is being used

but since it's an internal (as in not-shipped) tool I think an error is simpler - and we want to know (not learn later) that things are not like expected.

tests/xharness/XmlResultParser.cs
```diff
@@ -6,15 +6,16 @@
...
-	public static class XmlResultParser {
 
-		public enum Jargon {
-			TouchUnit,
-			NUnitV2,
-			NUnitV3,
-			xUnit,
-			Missing,
-		}
+	public enum Jargon {
```
spouliot: My original comment (on the PR this `enum` was introduced) mentioned I disliked internal types ;-) because they are even harder to name than other types.

If you unnest the `enum` then it becomes less clear what the usage of the type is... (less self-documenting), e.g. the argument you're adding is `xml-jargon`. Maybe `Xml[Result]Jargon`, the prefix will help for intellisense.

mandel-macaque: Funny thing is, I was thinking about your comment when I was doing the PR... he already mentioned it hehe

PR #7974, opened 2020-02-25
[xharness] Pass 'disable_omit_fp' to mono for debug builds.

rolfbjarne:
This may produce better stack traces when crashes occur, which could be
helpful for tracking down https://github.com/xamarin/maccore/issues/643.

### DIFF
--- a/tests/xharness/Jenkins.cs
+++ b/tests/xharness/Jenkins.cs
@@ -3352,7 +3352,10 @@ namespace xharness
 					}
 					if (!Harness.GetIncludeSystemPermissionTests (Platform, false))
 						proc.StartInfo.EnvironmentVariables ["DISABLE_SYSTEM_PERMISSION_TESTS"] = "1";
-					proc.StartInfo.EnvironmentVariables ["MONO_DEBUG"] = "no-gdb-backtrace";
+					var mono_debug = "no-gdb-backtrace";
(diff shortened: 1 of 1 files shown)


tests/xharness/Jenkins.cs
```diff
@@ -3352,7 +3352,10 @@ protected override async Task RunTestAsync ()
 					}
 					if (!Harness.GetIncludeSystemPermissionTests (Platform, false))
 						proc.StartInfo.EnvironmentVariables ["DISABLE_SYSTEM_PERMISSION_TESTS"] = "1";
-					proc.StartInfo.EnvironmentVariables ["MONO_DEBUG"] = "no-gdb-backtrace";
+					var mono_debug = "no-gdb-backtrace";
+					if (BuildTask.ProjectConfiguration == "Debug")
+						mono_debug += ",disable_omit_fp";
```
spouliot: is that meant to be temporary to debug the mentioned issue ?

I see the debug-ability value to disabling (omit the frame pointer) but afaik `mmp` never sets it so we would be testing using a configuration different from what developers are using, potentially having different (hard to find) issues. If useful why not make it make it default (for DEBUG) in `mmp` instead ?

rolfbjarne:

rolfbjarne:
Superseded by #7981.

PR #7976, opened 2020-02-25
[tests] Add `.cctor` execution to introspection

spouliot:
This will spot cases like https://github.com/xamarin/xamarin-macios/issues/7959
where a type (static) constructor can fail at runtime, leading to
`TypeLoadException`

Ideally it's covered by it's own tests but it's better covered twice
than never :)

### DIFF
--- a/tests/introspection/ApiTypeTest.cs
+++ b/tests/introspection/ApiTypeTest.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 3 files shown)


mandel-macaque:
Crash is due to: https://github.com/xamarin/xamarin-macios/issues/7978

dalexsoto:
Backport to Xcode 11.4!! since we have new bindings there

spouliot:
@monojenkins backport to xcode11.4

PR #7985, opened 2020-02-26
[Harness] Move BCLTestImport sources to their rightful place.

mandel-macaque:
We initially added a cmd to do the bcl test import generation that has
never been used.

We are moving the bcl test importer code to xharness and creating a
NUnit test project to run the tests that we already added. Unit tests
pass.


mandel-macaque:
First tests we have for xharness! \o/

PR #8009, opened 2020-02-28
[mtouch] Add `force-rejected-types-removal` optimization

spouliot:
This optimization can be enabled when it's not possible to use the
managed linker (e.g. **Don't link**) or when the managed linker cannot
remove references to deprecated types that would cause an application
to be rejected by Apple.

References to the existing types will be renamed, e.g. `UIWebView` to
`DeprecatedWebView`, in every assemblies.

The type definition is also renamed (for validity) and all custom
attributes on the types and their members will be removed.
Code inside the members will be replaced with a
`throw new NotSupportedException ();`.

The msbuild test app `MyReleaseBuild` has been updated to test that the
optimization is working as expected (device builds are slow so reusing
this test has little impact in test time).

Basically the test ensure that `UIWebView` is used and cannot be removed
by the compiler (optimization) or the managed linker (since it's
referenced). Since the optimization is enabled then we can `grep` then
final `.app` directory to ensure there's no mention of `UIWebView` inside
any of the files that would be submitted.

The application can be run, by itself, and will turn green if OK, red if
`DeprecatedWebView` can't be found (skeleton replacement for `UIWebView`)
or orange if a `NotSupportedException` is thrown.

Finally introspection tests have been updated to skip over the deprecated
(and renamed) types. It should not be an issue right now, since this
optimization is not enabled by default, but it made testing easier.

### DIFF
--- a/docs/website/optimizations.md
+++ b/docs/website/optimizations.md
@@ -787,3 +787,6 @@ Code inside the members will be replaced with a
 
 The default behavior can be overridden by passing
 `--optimize=[+|-]force-rejected-types-removal` to `mtouch`.
+
+The exact list of types might change over time and is best read directly from
+the [source code](https://github.com/xamarin/xamarin-macios/blob/master/tools/linker/RemoveRejectedTypesStep.cs).
--- a/docs/website/optimizations.md
+++ b/docs/website/optimizations.md
@@ -769,3 +769,21 @@ a product assembly and does not automagically mark all `NSObject`
 subclasses. This allows additional removes and optimizations to be 
 applied to the assembly, including the ability to remove `UIWebView` if
 nothing else in the application requires it.
+
+## Force Rejected Types Removal
+
+This optimization can be enabled when it's not possible to use the 
+managed linker (e.g. **Don't link**) or when the managed linker cannot
+remove references to deprecated types that would cause an application
+to be rejected by Apple.
+
+References to the existing types will be renamed, e.g. `UIWebView` to 
+`DeprecatedWebView`, in every assemblies.
+
+The type definition is also renamed (for validity) and all custom 
+attributes on the types and their members will be removed. 
+Code inside the members will be replaced with a
+`throw new NotSupportedException ();`.
+
+The default behavior can be overridden by passing
+`--optimize=[+|-]force-rejected-types-removal` to `mtouch`.
--- a/tools/linker/MonoTouch.Tuner/RemoveCodeBase.cs
+++ b/tools/linker/MonoTouch.Tuner/RemoveCodeBase.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using Mono.Linker;
+using Mono.Tuner;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Xamarin.Linker;
+
+namespace MonoTouch.Tuner {
+
+	abstract public class RemoveCodeBase : ExceptionalSubStep {
+		MethodDefinition get_nse_def;
+
+		protected RemoveCodeBase ()
+		{
+		}
+
+		protected MethodReference NotSupportedException { get; private set; }
+
+		protected override void Process (AssemblyDefinition assembly)
+		{
+			// only one definition exists (in mscorlib.dll)
+			if (get_nse_def == null) {
+				var corlib = context.GetAssembly ("mscorlib");
+				var nse = corlib.MainModule.GetType ("System", "NotSupportedException");
+				foreach (var m in nse.Methods) {
+					// no need to check HasMethods because we know there are (and nothing is removed at this stage)
+					if (m.Name != "LinkedAway")
+						continue;
+					get_nse_def = m;
+					break;
+				}
+			}
+
+			// import the method into the current assembly
+			// i.e. a different reference exists (and must be used) in every assembly
+			NotSupportedException = assembly.MainModule.ImportReference (get_nse_def);
+		}
+
+		protected void ProcessMethods (TypeDefinition type)
+		{
+			if (type.HasMethods) {
+				MethodDefinition static_ctor = null;
+				foreach (MethodDefinition method in type.Methods) {
+					if (method.IsConstructor && method.IsStatic)
+						static_ctor = method;
+					else
+						ProcessMethod (method);
+				}
+				if (static_ctor != null)
+					type.Methods.Remove (static_ctor);
+			}
+		}
+
+		new protected virtual void ProcessMethod (MethodDefinition method)
+		{
+			ProcessParameters (method);
+
+			if (!method.HasBody)
+				return;
+
+			var body = new MethodBody (method);
+
+			var il = body.GetILProcessor ();
+			il.Emit (OpCodes.Call, NotSupportedException);
+			il.Emit (OpCodes.Throw);
+
+			method.Body = body;
+		}
+
+		static void ProcessParameters (MethodDefinition method)
+		{
+			if (!method.HasParameters)
+				return;
+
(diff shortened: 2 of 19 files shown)


docs/website/optimizations.md
```diff
@@ -769,3 +769,21 @@ a product assembly and does not automagically mark all `NSObject`
...
+References to the existing types will be renamed, e.g. `UIWebView` to 
+`DeprecatedWebView`, in every assemblies.
+
+The type definition is also renamed (for validity) and all custom 
+attributes on the types and their members will be removed. 
+Code inside the members will be replaced with a
+`throw new NotSupportedException ();`.
+
+The default behavior can be overridden by passing
+`--optimize=[+|-]force-rejected-types-removal` to `mtouch`.
```
dalexsoto: We should put a list of the removed types here or a link to where it lives in our source code

tools/linker/MonoTouch.Tuner/RemoveCodeBase.cs
```diff
@@ -0,0 +1,79 @@
...
+						static_ctor = method;
+					else
+						ProcessMethod (method);
+				}
+				if (static_ctor != null)
+					type.Methods.Remove (static_ctor);
+			}
+		}
+
+		new protected virtual void ProcessMethod (MethodDefinition method)
```
dalexsoto: `new protected virtual` BOLD! 

/me hides

docs/website/optimizations.md
spouliot: a link is a good idea - since it would be too easy to get out of sync

tools/linker/MonoTouch.Tuner/RemoveCodeBase.cs
spouliot: heh, just moving code (but it might be mine from old times)

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,95 @@
...
+			base.Process (assembly);
+
+			var updated = false;
+			if (Profile.IsProductAssembly (assembly)) {
+				updated = true;
+				foreach (var remove in ReferencesToBeRemoved) {
+					var type = assembly.MainModule.GetType (remove);
+					if (type == null)
+						continue;
+					type.Name = type.Name.Replace ("UI", "Deprecated");
```
rolfbjarne: This only works for types named `UI*`. Maybe instead remove the first two characters? Or maybe make `ReferencesToBeRemoved` a hash table where the lookup value is the string to search for?

The whole purpose of this code falls down if this replacement fails, so it might make sense to detect if something wasn't replaced as expected (for instance check that pre- and post-replacement-strings are not identical) and show a warning in that case.

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,95 @@
...
+					Context.AddLinkedAwayType (type);
+					// remove [Register], [Protocol] and [ProtocolMembers] attributes
+					type.CustomAttributes.Clear ();
```
rolfbjarne: You remove all attributes, but the comment sounds like you're only removing a few.

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,95 @@
...
+					if (!type.IsInterface) {
+						// remove IL and custom attributes (e.g. [Export])
+						ProcessMethods (type);
+					}
+				}
+			} else if (Profile.IsSdkAssembly (assembly)) {
+				// note: we know the BCL does not include references to UIWebView
```
rolfbjarne: This might not be correct when something new is added to the list of types to remove. Not sure what to do here though. One solution would be to (again) make `ReferencesToBeRemoved` a hash table with a boolean lookup value that states whether the type can be found in SDK assemblies or not.

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,95 @@
...
+
+		protected override void ProcessMethod (MethodDefinition method)
+		{
+			if (method.HasCustomAttributes)
+				method.CustomAttributes.Clear ();
+
+            // don't break IDisposable
+//			if (method.Name != "Dispose") {
+				// deal with method `EnsureUIWebViewDelegate`
+				method.Name = method.Name.Replace ("UIWebView", "Deprecated");
```
rolfbjarne: Indentation is wrong here.

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,95 @@
...
+			// order is important for nested types
+			"UIKit.UIWebView/_UIWebViewDelegate",
+			"UIKit.UIWebView/UIWebViewAppearance",
+			"UIKit.IUIWebViewDelegate",
+			"UIKit.UIWebView",
+			"UIKit.UIWebViewDelegate",
+			"UIKit.UIWebViewDelegate_Extensions",
+			"UIKit.UIWebViewDelegateWrapper",
+			"UIKit.UIWebViewNavigationType",
+		};
```
rolfbjarne: Would it make sense to make this list user-configurable?

Maybe something like: `--optimize=force-rejected-types-removal=type1,type2,...`

That might make it easier for us the next time Apple deprecates something (we can just tell customers the mtouch argument they have to use to get around the requirement until we get an update out with new entries in this list).

tools/linker/RemoveRejectedTypesStep.cs
spouliot: The list are the one that needs to be removed (at least so far)
I'll update the comment

tools/linker/RemoveRejectedTypesStep.cs
spouliot: That was by conscious design :) this is not a general purpose tool to remove code. There's way too many things that could go wrong (it was not as easy as expected, nor do I think it covers every possible cases) and the lure to abuse might be too strong ;-)

tools/linker/RemoveRejectedTypesStep.cs
spouliot: It was not meant as a general purpose mechanism (since it's not clear if the next stone to hit us would fit the same pattern). However it means one less place to update if we supply to alternative name.

tools/linker/RemoveRejectedTypesStep.cs
spouliot: This situation is very uncommon - it's the first time, after a dozen years, that a required removal affect our public API. All other times we only had to change our internals to avoid the _rejected_ API.

I don't think we need the extra complexity. Worse case we can revisit the code if/when that situation change but my guess is that, because of naming convention, the public BCL types are unlikely to be a trigger.

tools/linker/RemoveRejectedTypesStep.cs
spouliot: hmm... List<T> guaranteed order (so inner types could be processed before their declared types). That's not true for a hash table or a (standard) dictionary.

tools/linker/RemoveRejectedTypesStep.cs
spouliot: forgot that part - at some point _of code_ I had issue for `Dispose` :)

tools/linker/RemoveRejectedTypesStep.cs
rolfbjarne: You could also create a list of tuples instead.

spouliot: yeah, that's what I'm doing. Time to move `mtouch` out of .net 4.6 :)

PR #8013, opened 2020-02-28
[CFNetworkHandler] Stop using internal API and instead copy-paste a modified version of it.

rolfbjarne:
Conceptually worse, but pragmatically better. Especially since in .NET 5 we
won't have access to internal API anymore.


mandel-macaque:
We have done this before, looks good.

PR #8017, opened 2020-02-28
[bgen] No need to verify that System.Nullable`1 comes from mscorlib.

rolfbjarne:
Instead verify that it's not a nested type, which should be more than good
enough. If someone else defines a System.Nullable`1, they'll probably have
much bigger problems.

This solves an issue with .NET 5 where System.Nullable`1 doesn't actually come
from mscorlib.

### DIFF
--- a/src/generator-typemanager.cs
+++ b/src/generator-typemanager.cs
@@ -157,7 +157,7 @@ public class TypeManager {
 			return null;
 
 		var gt = type.GetGenericTypeDefinition ();
-		if (gt.Assembly != CorlibAssembly)
+		if (gt.IsNested)
 			return null;
 
 		if (gt.Namespace != "System")


src/generator-typemanager.cs
```diff
@@ -157,7 +157,7 @@ public Type GetUnderlyingNullableType (Type type)
 			return null;
 
 		var gt = type.GetGenericTypeDefinition ();
-		if (gt.Assembly != CorlibAssembly)
+		if (gt.IsNested)
```
spouliot: I agree there's no need to check for the assembly, but is the nested check needed ? or could we just skip it ?

rolfbjarne: It's probably not needed. Then again it's a very cheap check (performance wise), and it rules out a remote possibility if someone does something unwise.

PR #8018, opened 2020-02-28
[bgen] Refactor a bit to not use target framework string comparison when we can use an enum.

rolfbjarne:
More future-proof, because new and different target frameworks will be
introduced in the future with .NET 5.

Also: fewer string comparisons is always good.


rolfbjarne:
Test failures are unrelated

* monotouch-test/tvOS - simulator/Release (all optimizations): https://github.com/xamarin/maccore/issues/2165
* MSBuild tests/iOS (net461): https://github.com/xamarin/maccore/issues/2137

PR #8021, opened 2020-03-01
[Harness] Refactor a little the code for logging to simplify testing.

mandel-macaque:
Move to use interfaces, that will let us later add tests that will
verify that all the correct logging is performed. As an example, added a
test for XmlResultParser that ensures that the failures are correctly
generated. The test uses Moq to pass the different paths to be used and
later be able to verify the wirtten xml.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -12,6 +12,7 @@ using System.Xml;
 using System.Xml.Xsl;
 using Xamarin;
 using Xamarin.Utils;
+using xharness.Logging;
 
 namespace xharness
 {
@@ -99,15 +100,15 @@ namespace xharness
 			set { log_directory = value; }
 		}
 
-		Logs logs;
-		public Logs Logs {
+		ILogs logs;
+		public ILogs Logs {
 			get {
 				return logs ?? (logs = new Logs (LogDirectory));
 			}
 		}
 
-		Log main_log;
-		public Log MainLog {
+		ILog main_log;
+		public ILog MainLog {
 			get { return main_log; }
 			set { main_log = value; }
 		}
@@ -476,9 +477,9 @@ namespace xharness
 		public async Task<int> RunAsync ()
 		{
 			CrashReportSnapshot crash_reports;
-			Log device_system_log = null;
-			Log listener_log = null;
-			Log run_log = main_log;
+			ILog device_system_log = null;
+			ILog listener_log = null;
+			ILog run_log = main_log;
 
 			Initialize ();
 
@@ -929,7 +930,7 @@ namespace xharness
 	class AppInstallMonitorLog : Log {
 		public override string FullPath => copy_to.FullPath;
 
-		Log copy_to;
+		ILog copy_to;
 		CancellationTokenSource cancellation_source;
 		
 		public bool CopyingApp;
@@ -951,7 +952,7 @@ namespace xharness
 			}
 		}
 
-		public AppInstallMonitorLog (Log copy_to)
+		public AppInstallMonitorLog (ILog copy_to)
 				: base (copy_to.Logs, $"Watch transfer log for {copy_to.Description}")
 		{
 			this.copy_to = copy_to;
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -12,9 +12,9 @@ using System.Xml;
 using System.Xml.Xsl;
 using Xamarin;
 using Xamarin.Utils;
-using xharness.Logging;
+using Xharness.Logging;
 
-namespace xharness
+namespace Xharness
 {
 	public enum AppRunnerTarget
 	{
@@ -394,7 +394,7 @@ namespace xharness
 					File.Delete (tmpFile);
 
 					// we do not longer need the tmp file
-					Logs.AddFile (path, Log.XML_LOG);
+					Logs.AddFile (path, LogType.XmlLog.ToString ());
 					return parseResult;
 
 				} catch (Exception e) {
@@ -550,7 +550,7 @@ namespace xharness
 			args.Add ($"-argument=-app-arg:-transport:{transport}");
 			args.Add ($"-setenv=NUNIT_TRANSPORT={transport}");
 
-			listener_log = Logs.Create ($"test-{mode}-{Harness.Timestamp}.log", Log.TEST_LOG, timestamp: !useXmlOutput);
+			listener_log = Logs.Create ($"test-{mode}-{Harness.Timestamp}.log", LogType.TestLog.ToString (), timestamp: !useXmlOutput);
 
 			SimpleListener listener;
 			switch (transport) {
@@ -651,7 +651,7 @@ namespace xharness
 					var log = new CaptureLog (Logs, sim.SystemLog, entire_file: Harness.Action != HarnessAction.Jenkins)
 					{
 						Path = Path.Combine (LogDirectory, sim.Name + ".log"),
-						Description = isCompanion ? Log.COMPANION_SYSTEM_LOG : Log.SYSTEM_LOG,
+						Description = isCompanion ? LogType.CompanionSystemLog.ToString () : LogType.SystemLog.ToString (),
 					};
 					log.StartCapture ();
 					Logs.Add (log);
--- a/tests/xharness/Log.cs
+++ b/tests/xharness/Log.cs
@@ -3,134 +3,19 @@ using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.IO;
 using System.Text;
+using xharness.Logging;
 
 namespace xharness
 {
-	public abstract class Log : TextWriter
-	{
-		public Logs Logs { get; private set; }
-		public string Description;
-		public bool Timestamp = true;
 
-		public static string XML_LOG = "XML log";
-		public static string NUNIT_RESULT = "NUnit results";
-		public static string SYSTEM_LOG = "System log";
-		public static string COMPANION_SYSTEM_LOG = "System log (companion)";
-		public static string BUILD_LOG = "Build log";
-		public static string TEST_LOG = "Test log";
-		public static string EXTENSION_TEST_LOG = "Extension test log";
-		public static string EXECUTION_LOG = "Execution log";
-
-		protected Log (Logs logs)
-		{
-			Logs = logs;
-		}
-
-		protected Log (Logs logs, string description)
-		{
-			Logs = logs;
-			Description = description;
-		}
-
-		public abstract string FullPath { get; }
-
-		internal protected virtual void WriteImpl (string value)
-		{
-			Write (Encoding.UTF8.GetBytes (value));
-		}
-
-		public virtual void Write (byte [] buffer)
-		{
-			Write (buffer, 0, buffer.Length);
-		}
-
-		public virtual void Write (byte[] buffer, int offset, int count)
-		{
-			throw new NotSupportedException ();
-		}
-
-		public virtual StreamReader GetReader ()
-		{
-			throw new NotSupportedException ();
-		}
-
-		public override void Write (char value)
-		{
-			WriteImpl (value.ToString ());
-		}
-
-		public override void Write (string value)
-		{
-			if (Timestamp)
-				value = DateTime.Now.ToString ("HH:mm:ss.fffffff") + " " + value;
-			WriteImpl (value);
-		}
-
-		public override void WriteLine (string value)
-		{
-			Write (value + "\n");
-		}
-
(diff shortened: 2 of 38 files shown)


tests/xharness/Process_Extensions.cs
```diff
@@ -56,17 +57,26 @@ public static Task<bool> PollForExitAsync (int pid, TimeSpan timeout)
...
+		public static Task<ProcessExecutionResult> RunAsync (this Process process, ILog log, bool append = true, TimeSpan? timeout = null, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null, bool? diagnostics = null)
 		{
 			return RunAsync (process, log, log, log, timeout, environment_variables, cancellation_token, diagnostics);
 		}
 
-		public static async Task<ProcessExecutionResult> RunAsync (this Process process, Log log, TextWriter StdoutStream, TextWriter StderrStream, TimeSpan? timeout = null, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null, bool? diagnostics = null)
+		public static Task<ProcessExecutionResult> RunAsync (this Process process, ILog log, ILog stdoutLog, ILog stderrLog, TimeSpan? timeout = null, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null, bool? diagnostics = null)
+		{
+			if (stdoutLog is TextWriter StdoutStream && stderrLog is TextWriter StderrStream) {
+				return RunAsync (process, log, StdoutStream, StderrStream, timeout, environment_variables, cancellation_token, diagnostics);
```
premun: I don't understand why we need this cast. Seems like the params for the method below could also be ILog and nothing would break?

tests/xharness/Log.cs
```diff
@@ -3,134 +3,19 @@
...
-			public override void Write (byte [] buffer, int offset, int count)
-			{
-				foreach (var log in logs)
-					log.Write (buffer, offset, count);
-			}
-		}
-	}
-
-	public class LogFile : Log
+	public class LogFile : Log, ILogFile
```
premun: I think this file should be called LogFile and probably be moved to Logging directory too

tests/xharness/Xharness.Tests/Xharness.Tests.csproj
```diff
@@ -159,6 +165,21 @@
     <Compile Include="BCLTestImporter\Tests\TestAssemblyDefinitionTest.cs" />
     <Compile Include="BCLTestImporter\Tests\TestProjectDefinitionTest.cs" />
     <Compile Include="Tests\XmlResultParserTests.cs" />
+    <Compile Include="..\Logging\ILogs.cs">
```
premun: Any reason why not to reference the tested project instead? (just curious)

premun: Any reason why not to reference the tested project but instead link all the files? I opened the solution and there were all the source files in the test project as links

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace xharness.Logging {
+
+	public abstract class Log : TextWriter, ILog {
+		public ILogs Logs { get; private set; }
+		public string Description { get; set; }
+		public bool Timestamp { get; set; } = true;
+
+		public static string XML_LOG = "XML log";
+		public static string NUNIT_RESULT = "NUnit results";
```
premun: Do you need these fields when you have also the enum in `ILog` and the corresponding `ToString` extension?

tests/xharness/Logging/ILog.cs
```diff
@@ -0,0 +1,62 @@
...
+			case LogType.BUILD_LOG:
+				return "Build log";
+			case LogType.TEST_LOG:
+				return "Test log";
+			case LogType.EXTENSION_TEST_LOG:
+				return "Extension test log";
+			case LogType.EXECUTION_LOG:
+				return "Execution log";
+			default:
+				throw new ArgumentException ($"Unknoe tyme for {nameof (type)}");
```
premun: >Unknoe [](start = 35, length = 6)

Unknown type

premun: >nameof (type) [](start = 52, length = 13)

I believe `nameof(type)` will output the string `type`

tests/xharness/Logging/ILog.cs
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+	}
+	public interface ILog : IDisposable {
+		ILogs Logs { get; }
+		string Description { get; set; }
```
premun: >string [](start = 2, length = 6)

Based on the usage I've seen, this also might be `LogType` and we could get the Description from that always

tests/xharness/Logging/ILogs.cs
```diff
@@ -0,0 +1,24 @@
...
+		ILogFile Create (string filename, string name, bool? timestamp = null);
+
+		// Adds an existing file to this collection of logs.
+		// If the file is not inside the log directory, then it's copied there.
+		// 'path' must be a full path to the file.
+		ILogFile AddFile (string path);
+
+		// Adds an existing file to this collection of logs.
+		// If the file is not inside the log directory, then it's copied there.
+		// 'path' must be a full path to the file.
```
premun: Not sure what is the Xamarin custom on documenting methods, but maybe it would be better to do XML docs?

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+
+		protected Log (ILogs logs, string description)
+		{
+			Logs = logs;
+			Description = description;
+		}
+
+		public abstract string FullPath { get; }
+
+		public virtual void WriteImpl (string value)
```
premun: >WriteImpl [](start = 22, length = 9)

This should rather be protected and not in the interface as it's internal implementation of the class.
You might even make it abstract on this level as well as the `Write` methods and require implementation from the non-abstract implementations of this class

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+		{
+			Write (Encoding.UTF8.GetBytes (value));
```
premun: >Encoding.UTF8 [](start = 10, length = 13)

I think this should use the `Encoding` property here

tests/xharness/Logging/Logs.cs
```diff
@@ -0,0 +1,69 @@
...
+				File.Copy (path, newPath, true);
+				path = newPath;
+			}
+
+			var log = new LogFile (this, name, path, true);
+			Add (log);
+			return log;
+		}
+
+		// Create an empty file in the log directory and return the full path to the file
```
premun: >// [](start = 2, length = 2)

I wouldn't duplicate docs as it is hard to maintain both and the usually divert over time

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+		{
+		}
+
+		public override Encoding Encoding {
+			get {
+				return System.Text.Encoding.UTF8;
+			}
+		}
+
+		public static Log CreateAggregatedLog (params ILog [] logs)
```
premun: >Log [](start = 16, length = 3)

`ILog`?

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+			Write (value + "\n");
+		}
+
+		public void WriteLine (StringBuilder value)
+		{
+			Write (value.ToString () + "\n");
+		}
+		public override void WriteLine (string format, params object [] args)
+		{
+			Write (string.Format (format, args) + "\n");
```
premun: >"\n" [](start = 41, length = 4)

`Environment.NewLine`

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+		class AggregatedLog : Log {
+			ILog [] logs;
+
+			public AggregatedLog (params ILog [] logs)
+				: base (null)
+			{
+				this.logs = logs;
+			}
+
+			public override string FullPath => throw new NotImplementedException ();
```
premun: >NotImplementedException [](start = 48, length = 23)

`NotSupportedException`

tests/xharness/Logging/Log.cs
```diff
@@ -0,0 +1,125 @@
...
+		{
+			Write (Encoding.UTF8.GetBytes (value));
+		}
+
+		public virtual void Write (byte [] buffer)
+		{
+			Write (buffer, 0, buffer.Length);
+		}
+
+		public virtual void Write (byte [] buffer, int offset, int count)
```
premun: >virtual [](start = 9, length = 7)

Better to make it abstract

tests/xharness/Process_Extensions.cs
mandel-macaque: It will not compile. Needs to be casted. I did check without it.

tests/xharness/Log.cs
mandel-macaque: Yes, that is coming in a second PR. I try to minimise the number of lines to make it simpler for reviews.

tests/xharness/Logging/Log.cs
mandel-macaque: I added the enums, forgot to change it, will update!

tests/xharness/Xharness.Tests/Xharness.Tests.csproj
mandel-macaque: Step by step. The main project is not a lib, but an exe. Once testing is done, I'll add it as a ref.

tests/xharness/Logging/ILogs.cs
mandel-macaque: Moving code, wrote by rolf

tests/xharness/Logging/Log.cs
mandel-macaque: Very good point

tests/xharness/Logging/Log.cs
mandel-macaque: 100% agree, but it is used by one of the other classes. I need to find a nice way not to do it like this. Will come in another PR. But you are right

tests/xharness/Process_Extensions.cs
premun: I suggest changing the signature below to `ILog` and remove this one. It probably doesn't compile because the signatures will be the same or is there something inside the method below that needs `TextWriter` and is not ok with `ILog`?

tests/xharness/Logging/ILog.cs
```diff
@@ -0,0 +1,62 @@
...
+	// defines common log types
+	public enum LogType {
+		XML_LOG,
+		NUNIT_RESULT,
+		SYSTEM_LOG,
+		COMPANION_SYSTEM_LOG,
+		BUILD_LOG,
+		TEST_LOG,
+		EXTENSION_TEST_LOG,
+		EXECUTION_LOG,
```
spouliot: not sure why we ended up with all caps here
might be a good time to fix that if this gets re-used

tests/xharness/Logging/ILog.cs
```diff
@@ -0,0 +1,62 @@
...
+			case LogType.COMPANION_SYSTEM_LOG:
+				return "System log (companion)";
+			case LogType.BUILD_LOG:
+				return "Build log";
+			case LogType.TEST_LOG:
+				return "Test log";
+			case LogType.EXTENSION_TEST_LOG:
+				return "Extension test log";
+			case LogType.EXECUTION_LOG:
+				return "Execution log";
```
spouliot: with better enum value names then the original `ToString` should do a good enough job
e.g.
`EXECUTION_LOG` -> `ExecutionLog`

tests/xharness/AppRunner.cs
```diff
@@ -12,6 +12,7 @@
 using System.Xml.Xsl;
 using Xamarin;
 using Xamarin.Utils;
+using xharness.Logging;
```
spouliot: I know the original one did not but it should start with a capital letter  :)

tests/xharness/Process_Extensions.cs
mandel-macaque: Correct, I can remove and either do the casting in the call or in the other method. I added the other one 'cause the indentation went crazy. I'll do it in the caller since if will be cleaner.

premun: I am trying to say that I think there doesn't need to be any casting at all :)

tests/xharness/AppRunner.cs
mandel-macaque: easy fix

PR #8032, opened 2020-03-03
[XHarness] Add tests for all the logging classes.

mandel-macaque:
Add tests for all the classes in the new name space XHarness.Logging.
Tests ensure that logs are correctly written and in the correct stream.

### DIFF
--- a/tests/xharness/Logging/CallbackLog.cs
+++ b/tests/xharness/Logging/CallbackLog.cs
@@ -10,7 +10,7 @@ namespace Xharness.Logging {
 			OnWrite = onWrite;
 		}
 
-		public override string FullPath => throw new NotImplementedException ();
+		public override string FullPath => throw new NotSupportedException ();
 
 		public override void WriteImpl (string value)
 		{
--- a/tests/xharness/Logging/CallbackLog.cs
+++ b/tests/xharness/Logging/CallbackLog.cs
@@ -2,7 +2,7 @@
 namespace Xharness.Logging {
 	// A log that forwards all written data to a callback
 	public class CallbackLog : Log {
-		public Action<string> OnWrite;
+		readonly Action<string> OnWrite;
 
 		public CallbackLog (Action<string> onWrite)
 			: base (null)
--- a/tests/xharness/Logging/CallbackLog.cs
+++ b/tests/xharness/Logging/CallbackLog.cs
@@ -1,0 +1,20 @@
+﻿using System;
+namespace xharness.Logging {
+	// A log that forwards all written data to a callback
+	public class CallbackLog : Log {
+		public Action<string> OnWrite;
+
+		public CallbackLog (Action<string> onWrite)
+			: base (null)
+		{
+			OnWrite = onWrite;
+		}
+
+		public override string FullPath => throw new NotImplementedException ();
+
+		public override void WriteImpl (string value)
+		{
+			OnWrite (value);
+		}
+	}
+}
--- a/tests/xharness/Logging/CallbackLog.cs
+++ b/tests/xharness/Logging/CallbackLog.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-namespace xharness.Logging {
+namespace Xharness.Logging {
 	// A log that forwards all written data to a callback
 	public class CallbackLog : Log {
 		public Action<string> OnWrite;
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -136,4 +136,3 @@ namespace Xharness.Logging {
 		}
 	}
 }
-
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -18,11 +18,9 @@ namespace Xharness.Logging {
 		public CaptureLog (ILogs logs, string path, string capture_path, bool entire_file = false)
 			: base (logs)
 		{
-			if (path == null)
-				throw new ArgumentNullException (nameof (path));
 
-			Path = path;
-			CapturePath = capture_path;
+			Path = path ?? throw new ArgumentNullException (nameof (path));
+			CapturePath = capture_path ?? throw new ArgumentNullException (nameof (path));
 			this.entire_file = entire_file;
 		}
 
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -1,113 +1,7 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics.Contracts;
 using System.IO;
-using System.Text;
-using xharness.Logging;
 
-namespace xharness
-{
-
-	public class LogFile : Log, ILogFile
-	{
-		object lock_obj = new object ();
-		public string Path { get; private set; }
-		FileStream writer;
-		bool disposed;
-
-		public LogFile (ILogs logs, string description, string path, bool append = true)
-			: base (logs, description)
-		{
-			Path = path;
-			if (!append)
-				File.WriteAllText (path, string.Empty);
-		}
-
-		public override void Write (byte [] buffer, int offset, int count)
-		{
-			try {
-				// We don't want to open the file every time someone writes to the log, so we keep it as an instance
-				// variable until we're disposed. Due to the async nature of how we run tests, writes may still
-				// happen after we're disposed, in which case we create a temporary stream we close after writing
-				lock (lock_obj) {
-					var fs = writer;
-					if (fs == null) {
-						fs = new FileStream (Path, FileMode.Append, FileAccess.Write, FileShare.Read);
-					}
-					fs.Write (buffer, offset, count);
-					if (disposed) {
-						fs.Dispose ();
-					} else {
-						writer = fs;
-					}
-				}
-			} catch (Exception e) {
-				Console.WriteLine ($"Failed to write to the file {Path}: {e.Message}.");
-				return;
-			}
-		}
-
-		public override void Flush()
-		{
-			base.Flush();
-
-			if (writer != null && !disposed)
-				writer.Flush ();
-		}
-
-		public override string FullPath {
-			get {
-				return Path;
-			}
-		}
-
-		public override StreamReader GetReader ()
-		{
-			return new StreamReader (new FileStream (Path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite));
-		}
-
-		protected override void Dispose (bool disposing)
-		{
-			base.Dispose (disposing);
-
-			if (writer != null) {
-				writer.Dispose ();
-				writer = null;
-			}
-
-			disposed = true;
-		}
-
-	}
-
-	// A log that writes to standard output
-	public class ConsoleLog : Log
-	{
-		StringBuilder captured = new StringBuilder ();
-
-		public ConsoleLog ()
-			: base (null)
-		{
-		}
-
-		public override void WriteImpl (string value)
-		{
-			captured.Append (value);
-			Console.Write (value);
-		}
-
-		public override string FullPath {
-			get {
-				throw new NotSupportedException ();
-			}
-		}
-
-		public override StreamReader GetReader ()
-		{
-			var str = new MemoryStream (System.Text.Encoding.UTF8.GetBytes (captured.ToString ()));
-			return new StreamReader (str, System.Text.Encoding.UTF8, false);
-		}
-	}
+namespace xharness.Logging {
 
 	// A log that captures data written to a separate file between two moments in time
 	// (between StartCapture and StopCapture).
@@ -233,25 +127,6 @@ namespace xharness
 			get {
 				return Path;
 			}
-		}
-	}
-
-	// A log that forwards all written data to a callback
-	public class CallbackLog : Log
-	{
-		public Action<string> OnWrite;
-
-		public CallbackLog (Action<string> onWrite)
-			: base (null)
-		{
-			OnWrite = onWrite;
-		}
-
-		public override string FullPath => throw new NotImplementedException ();
-
-		public override void WriteImpl (string value)
-		{
-			OnWrite (value);
 		}
 	}
 }
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.IO;
 
-namespace xharness.Logging {
+namespace Xharness.Logging {
 
 	// A log that captures data written to a separate file between two moments in time
 	// (between StartCapture and StopCapture).
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -12,7 +12,7 @@ namespace Xharness.Logging {
 
 		long startPosition;
 		long endPosition;
-		bool entire_file;
+		bool entireFile;
 		bool started;
 
 		public CaptureLog (ILogs logs, string path, string capture_path, bool entire_file = false)
@@ -21,12 +21,12 @@ namespace Xharness.Logging {
 
 			Path = path ?? throw new ArgumentNullException (nameof (path));
 			CapturePath = capture_path ?? throw new ArgumentNullException (nameof (path));
-			this.entire_file = entire_file;
+			this.entireFile = entire_file;
 		}
 
 		public void StartCapture ()
 		{
-			if (entire_file)
+			if (entireFile)
 				return;
 
 			if (File.Exists (CapturePath))
@@ -36,14 +36,14 @@ namespace Xharness.Logging {
 
 		public void StopCapture ()
 		{
-			if (!started && !entire_file)
+			if (!started && !entireFile)
 				throw new InvalidOperationException ("StartCapture most be called before StopCature on when the entire file will be captured.");
 			if (!File.Exists (CapturePath)) {
 				File.WriteAllText (Path, $"Could not capture the file '{CapturePath}' because it doesn't exist.");
 				return;
 			}
 
-			if (entire_file) {
+			if (entireFile) {
 				File.Copy (CapturePath, Path, true);
 				return;
 			}
@@ -55,7 +55,7 @@ namespace Xharness.Logging {
 
 		void Capture ()
 		{
-			if (startPosition == 0 || entire_file)
+			if (startPosition == 0 || entireFile)
 				return;
 
 			if (!File.Exists (CapturePath)) {
--- a/tests/xharness/Logging/CaptureLog.cs
+++ b/tests/xharness/Logging/CaptureLog.cs
@@ -7,8 +7,8 @@ namespace Xharness.Logging {
 	// (between StartCapture and StopCapture).
 	public class CaptureLog : Log
 	{
-		public string CapturePath { get; private set; }
-		public string Path { get; private set; }
+		public readonly string CapturePath;
+		public readonly string Path;
 
 		long startPosition;
 		long endPosition;
@@ -72,7 +72,8 @@ namespace Xharness.Logging {
 			var capturedLength = 0L;
 
 			if (length < 0) {
-				// The file shrank?
+				// The file shrank? lets copy the entire file in this case, which is better than nothing
+				File.Copy (CapturePath, Path, true);
 				return;
 			}
 
--- a/tests/xharness/Logging/ConsoleLog.cs
+++ b/tests/xharness/Logging/ConsoleLog.cs
@@ -2,7 +2,7 @@
 using System.IO;
 using System.Text;
 
-namespace xharness.Logging {
+namespace Xharness.Logging {
 	// A log that writes to standard output
 	public class ConsoleLog : Log {
 		StringBuilder captured = new StringBuilder ();
--- a/tests/xharness/Logging/ConsoleLog.cs
+++ b/tests/xharness/Logging/ConsoleLog.cs
@@ -26,8 +26,8 @@ namespace Xharness.Logging {
 
 		public override StreamReader GetReader ()
 		{
-			var str = new MemoryStream (System.Text.Encoding.UTF8.GetBytes (captured.ToString ()));
-			return new StreamReader (str, System.Text.Encoding.UTF8, false);
+			var str = new MemoryStream (Encoding.GetBytes (captured.ToString ()));
+			return new StreamReader (str, Encoding, false);
 		}
 	}
 }
--- a/tests/xharness/Logging/ConsoleLog.cs
+++ b/tests/xharness/Logging/ConsoleLog.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.IO;
+using System.Text;
+
+namespace xharness.Logging {
+	// A log that writes to standard output
+	public class ConsoleLog : Log {
+		StringBuilder captured = new StringBuilder ();
+
+		public ConsoleLog ()
+			: base (null)
+		{
+		}
+
+		public override void WriteImpl (string value)
+		{
+			captured.Append (value);
+			Console.Write (value);
+		}
+
+		public override string FullPath {
+			get {
+				throw new NotSupportedException ();
+			}
+		}
+
+		public override StreamReader GetReader ()
+		{
+			var str = new MemoryStream (System.Text.Encoding.UTF8.GetBytes (captured.ToString ()));
+			return new StreamReader (str, System.Text.Encoding.UTF8, false);
+		}
+	}
+}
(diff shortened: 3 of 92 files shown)


tests/xharness/Xharness.Tests/Logging/Tests/CaptureLogTest.cs
```diff
@@ -0,0 +1,106 @@
...
+					}
+				}
+			}
+		}
+
+		[Test]
+		public void CaptureMissingFileTest ()
+		{
+			using (var captureLog = new CaptureLog (logs.Object, capturePath, filePath, false)) {
+				Assert.AreEqual (capturePath, captureLog.Path);
```
spouliot: add a unique message to show in case of error

tests/xharness/Xharness.Tests/Logging/Tests/CaptureLogTest.cs
```diff
@@ -0,0 +1,106 @@
...
+		{
+			Assert.Throws<InvalidOperationException> (() => {
+				using (var captureLog = new CaptureLog (logs.Object, capturePath, filePath, false)) {
+					captureLog.StopCapture ();
+				}
+			});
+		}
+
+		[Test]
+
```
spouliot: minor: ^ extra empty line between attribute and method

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+	[TestFixture]
+	public class LogFileTest {
+
+		string path;
+		string description;
+
+		Mock<ILogs> logs;
+
+		[SetUp]
+		public void SetUpt ()
```
spouliot: minor: typo (`t`)

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			if (File.Exists (path))
+				File.Delete (path);
+			path = null;
+			description = null;
+			logs = null;
+		}
+
+		[Test]
+		public void ConstrutorTest ()
```
spouliot: minor: typo (missing `c`)

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			using (var log = new LogFile (logs.Object, description, path)) {
+				Assert.AreEqual (description, log.Description, "description");
+				Assert.AreEqual (path, log.FullPath, "path");
+				Assert.AreSame (logs.Object, log.Logs, "logs");
+			}
+		}
+
+		[Test]
+		public void ConstrutorNullILogsTest ()
```
spouliot: copy/paste :)

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			Assert.DoesNotThrow (() => {
+				using (var log = new LogFile (null, description, path)) ;
+			});
+		}
+
+		[Test]
+		public void ConstrutorNullPathTest ()
```
spouliot: copy/paste :)

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			Assert.Throws<ArgumentNullException> (() => { var log = new LogFile (logs.Object, description, null); });
+		}
+
+		[Test]
+		public void ConstrutorNullDescritionTest ()
```
spouliot: copy/paste :) + missing `p` in Description

tests/xharness/Logging/CallbackLog.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+namespace Xharness.Logging {
+	// A log that forwards all written data to a callback
+	public class CallbackLog : Log {
+		public Action<string> OnWrite;
```
premun: >public [](start = 2, length = 6)

Can be `private readonly` field

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
+﻿using System;
+using System.IO;
+
+namespace Xharness.Logging {
+
+	// A log that captures data written to a separate file between two moments in time
+	// (between StartCapture and StopCapture).
+	public class CaptureLog : Log
+	{
+		public string CapturePath { get; private set; }
```
premun: Can be `private readonly` field

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+		public string Path { get; private set; }
```
premun: >Path [](start = 16, length = 4)

Should be made `readonly` too, no private setter needed

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+
+		long startPosition;
+		long endPosition;
+		bool entire_file;
+		bool started;
+
+		public CaptureLog (ILogs logs, string path, string capture_path, bool entire_file = false)
+			: base (logs)
+		{
+			if (path == null)
```
premun:
```suggestion
			Path = path ?? throw new ArgumentNullException(nameof(path));
```

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				throw new ArgumentNullException (nameof (path));
+
+			Path = path;
```
premun:
```suggestion
			Path = path ?? throw new ArgumentNullException(nameof(path));
```

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+			CapturePath = capture_path;
```
premun:
```suggestion
			CapturePath = capture_path ?? throw new ArgumentNullException(nameof(capture_path));
```

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				return;
+
+			if (File.Exists (CapturePath))
+				startPosition = new FileInfo (CapturePath).Length;
+			started = true;
+		}
+
+		public void StopCapture ()
+		{
+			if (!started && !entire_file)
```
premun: I think it would be better to not store state of the class in two boolean variables but rather have a `CaptureState` enum with values `NotStarted`, `EntireFile` and `Started`. Because for example `started && entire_file` is not a valid state and the code should not let that happen.

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				File.Copy (CapturePath, Path, true);
+				return;
+			}
+
+			endPosition = new FileInfo (CapturePath).Length;
+
+			Capture ();
+		}
+
+		void Capture ()
```
premun:
```suggestion
		private void Capture ()
```

But I would also maybe rename this to `Finalize` or `Flush` or something..

tests/xharness/Logging/Logs.cs
```diff
@@ -54,6 +57,8 @@ public ILogFile AddFile (string path, string name)
 		// Create an empty file in the log directory and return the full path to the file
 		public string CreateFile (string path, string description)
```
premun: >description [](start = 48, length = 11)

`description` is not used

tests/xharness/Logging/CallbackLog.cs
```diff
@@ -0,0 +1,20 @@
...
+
+		public CallbackLog (Action<string> onWrite)
+			: base (null)
+		{
+			OnWrite = onWrite;
+		}
+
+		public override string FullPath => throw new NotImplementedException ();
```
premun:
```suggestion
		public override string FullPath => throw new NotSupportedException ();
```

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+	{
+		public string CapturePath { get; private set; }
+		public string Path { get; private set; }
+
+		long startPosition;
+		long endPosition;
+		bool entire_file;
+		bool started;
+
+		public CaptureLog (ILogs logs, string path, string capture_path, bool entire_file = false)
```
premun: I think it would be worth naming `path` something more descriptive, such as `target_path` so that it's more clear. `capture_path` could be also `source_path` but that is up to you, this one is ok too.

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+	// A log that captures data written to a separate file between two moments in time
+	// (between StartCapture and StopCapture).
+	public class CaptureLog : Log
+	{
+		public string CapturePath { get; private set; }
+		public string Path { get; private set; }
+
+		long startPosition;
+		long endPosition;
+		bool entire_file;
```
premun: I think you should stick with camelCase rather than snake_case just like rest of the code base

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				throw new InvalidOperationException ("StartCapture most be called before StopCature on when the entire file will be captured.");
+			if (!File.Exists (CapturePath)) {
+				File.WriteAllText (Path, $"Could not capture the file '{CapturePath}' because it doesn't exist.");
+				return;
+			}
+
+			if (entire_file) {
+				File.Copy (CapturePath, Path, true);
+				return;
+			}
```
premun: I think it should be responsibility of just one method to write the output file like the `Capture`/`Finalize` one.
I suggest for entire_file we just set start to 0 and end to Length and let either `Capture` not even know we have something like entire_file OR `Capture` can do the `Copy`

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+			if (startPosition == 0 || entire_file)
+				return;
+
+			if (!File.Exists (CapturePath)) {
+				File.WriteAllText (Path, $"Could not capture the file '{CapturePath}' because it does not exist.");
+				return;
+			}
+
+			var currentEndPosition = endPosition;
+			if (currentEndPosition == 0)
```
premun: >if (currentEndPosition == 0) [](start = 3, length = 28)

This would happen for empty file only, no? So we can create that empty file and be done here.
Or if this is for entire_file mode, than this is handled already above and shouldn't happen here.

This is example of the impossible state of the class too

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				currentEndPosition = new FileInfo (CapturePath).Length;
+			
+			var length = (int) (currentEndPosition - startPosition);
+			var currentLength = new FileInfo (CapturePath).Length;
+			var capturedLength = 0L;
+
+			if (length < 0) {
+				// The file shrank?
```
premun: What if someone has overriden the file during the test run and it was long before and is short now? We should just copy the file I guess?

Definitely do something other than just return. At least throw would be better but I would copy the file

tests/xharness/Logging/CaptureLog.cs
```diff
@@ -0,0 +1,141 @@
...
+				}
+			}
+		}
+
+		public override StreamReader GetReader ()
+		{
+			if (File.Exists (CapturePath)) {
+				return new StreamReader (new FileStream (CapturePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite));
+			} else {
+				return new StreamReader (new MemoryStream ());
```
premun: I think this might be misleading. If we expect the file to be created later than this logger (perfectly valid), we should return `null` here to let the user know that there is nothing to capture yet.
Because if someone gets the empty `MemoryStream` that will always be empty, they will hold onto it and never see any content.
While this class will start capturing from the real file if it is created after someone called `GetReader()` it will have a different source.

tests/xharness/Logging/ConsoleLog.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		public override string FullPath {
+			get {
+				throw new NotSupportedException ();
+			}
+		}
+
+		public override StreamReader GetReader ()
+		{
+			var str = new MemoryStream (System.Text.Encoding.UTF8.GetBytes (captured.ToString ()));
```
premun: >System.Text.Encoding [](start = 31, length = 20)

This should reference the `Encoding` field

tests/xharness/Logging/ConsoleLog.cs
```diff
@@ -0,0 +1,33 @@
...
+			return new StreamReader (str, System.Text.Encoding.UTF8, false);
```
premun:
```suggestion
			var str = new MemoryStream (Encoding.GetBytes (captured.ToString ()));
			return new StreamReader (str, Encoding, false);
```

tests/xharness/Logging/ConsoleLog.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.Text;
+
+namespace Xharness.Logging {
+	// A log that writes to standard output
+	public class ConsoleLog : Log {
+		StringBuilder captured = new StringBuilder ();
+
+		public ConsoleLog ()
+			: base (null)
```
premun: >null [](start = 11, length = 4)

I think the parent abstract class should not have the `Logs` field. It is not using it in any way in its implementation and it is obvious from this call that it is not needed there. Just in the implementations

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+		public string Path { get; private set; }
+		FileStream writer;
+		bool disposed;
+
+		public LogFile (ILogs logs, string description, string path, bool append = true)
+			: base (logs, description)
+		{
+			if (path == null)
+				throw new ArgumentNullException (nameof (path));
+			Path = path;
```
premun:
```suggestion
			Path = path ?? throw new ArgumentNullException (nameof (path));
```

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+			if (!append)
```
premun: >(! [](start = 6, length = 2)

Not sure this should happen in ctor really. If this blows up (e.g. path invalid or not existing) it will throw from `new` which is a bad practice.
I would rather handle that in `Write` by making the `FileMode` depend on `append` and then switch `append` after first write.

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+				return;
+			}
+		}
+
+		public override void Flush ()
+		{
+			base.Flush ();
+
+			if (writer != null && !disposed)
+				writer.Flush ();
```
premun: If you are worried about multi-threading, you might need to lock around this too.
You can also omit the writer check and do `writer?.Flush ();`

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		public override string FullPath {
+			get {
+				return Path;
+			}
+		}
```
premun: nit: You are using arrow methods already elsewhere

```suggestion
		public override string FullPath => Path;
```

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+
+		protected override void Dispose (bool disposing)
+		{
+			base.Dispose (disposing);
+
+			if (writer != null) {
+				writer.Dispose ();
+				writer = null;
+			}
```
premun:
```suggestion
			lock (lock_obj) {
				writer?.Dispose ();
				writer = null;
			}
```

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+			: base (logs, description)
+		{
+			if (path == null)
+				throw new ArgumentNullException (nameof (path));
+			Path = path;
+			if (!append)
+				File.WriteAllText (path, string.Empty);
+		}
+
+		public override void Write (byte [] buffer, int offset, int count)
```
premun: This should check disposed first and, in case it is true ,throw OR out warning and return

tests/xharness/Logging/LogFile.cs
```diff
@@ -0,0 +1,77 @@
...
+					}
+					fs.Write (buffer, offset, count);
+					if (disposed) {
+						fs.Dispose ();
+					} else {
+						writer = fs;
+					}
+				}
+			} catch (Exception e) {
+				Console.WriteLine ($"Failed to write to the file {Path}: {e.Message}.");
```
premun:
```suggestion
				Console.Error.WriteLine ($"Failed to write to the file {Path}: {e.Message}.");
```

tests/xharness/Xharness.Tests/Logging/Tests/CallbackLogTest.cs
```diff
@@ -0,0 +1,29 @@
...
+			Action<string> cb = (d) => {
+				called = true;
+				data = d;
+			};
+
+			var log = new CallbackLog (cb);
+			log.Write (message);
+			Assert.IsTrue (called, "called");
+			Assert.IsNotNull (data, "data");
+			StringAssert.EndsWith (message, data, "message"); // take time stamp into account
```
premun: Should this be a TODO?

tests/xharness/Xharness.Tests/Logging/Tests/CallbackLogTest.cs
```diff
@@ -0,0 +1,29 @@
...
+			string data = null;
+
+			Action<string> cb = (d) => {
+				called = true;
+				data = d;
+			};
+
+			var log = new CallbackLog (cb);
+			log.Write (message);
+			Assert.IsTrue (called, "called");
```
premun: FYI: The assert message is in case of a failed assert, so here it would be `"Callback was not called"`

tests/xharness/Xharness.Tests/Logging/Tests/CaptureLogTest.cs
```diff
@@ -0,0 +1,105 @@
...
+		{
+			filePath = Path.GetTempFileName ();
+			capturePath = Path.GetTempFileName ();
+			logs = new Mock<ILogs> ();
+			File.Delete (filePath);
+			File.Delete (capturePath);
+		}
+
+		[TearDown]
+		public void TeatDown ()
```
premun: nit: Tea**r**down

tests/xharness/Xharness.Tests/Logging/Tests/CaptureLogTest.cs
```diff
@@ -0,0 +1,105 @@
...
+				using (var writer = new StreamWriter (filePath)) {
+					foreach (var line in logLines)
+						writer.WriteLine (line);
+				}
+				captureLog.StopCapture ();
+				// get the stream and assert we do have the correct lines
+				using (var captureStream = captureLog.GetReader ()) {
+					string line;
+					while ((line = captureStream.ReadLine ()) != null) {
+						Assert.Contains (line, logLines, line);
```
premun:
```suggestion
						Assert.Contains (line, logLines, "Lines not captured");
```

tests/xharness/Xharness.Tests/Logging/Tests/CaptureLogTest.cs
```diff
@@ -0,0 +1,105 @@
...
+					captureLog.StopCapture ();
+				}
+			});
+		}
+
+		[Test]
+		public void CaptureWrongOrderEntirePath ()
+		{
+			Assert.DoesNotThrow (() => {
+				using (var captureLog = new CaptureLog (logs.Object, capturePath, filePath, true)) {
```
premun: I think this should also verify contents of the target file as we are not really testing that functionality anywhere

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
```diff
@@ -0,0 +1,130 @@
...
+			Assert.Throws<ArgumentNullException> (() => { var log = new LogFile (logs.Object, description, null); });
+		}
+
+		[Test]
+		public void ConstructorNullDescriptionTest ()
+		{
+			Assert.DoesNotThrow (() => {
+				using (var log = new LogFile (logs.Object, null, path)) ;
+			});
+		}
```
premun: Not sure these tests bring any actual value (apart from code coverage points)

premun:
> 			Write (Encoding.UTF8.GetBytes (value));

```suggestion
   Write (Encoding.GetBytes (value));
```

---
Refers to: tests/xharness/Logging/Log.cs:27 in 8c9a73c. [](commit_id = 8c9a73c17749d79b97b18aa0db2c0a7cb5db385e, deletion_comment = False)

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: private is the default in c#. Because it is, we do not add it since in the mono style it is not recommended because it will 'clutter' the code.

tests/xharness/Logging/Logs.cs
mandel-macaque: It is:
```csharp
using (var rv = new LogFile (this, description, Path.Combine (Directory, path), false)) {
```

tests/xharness/Xharness.Tests/Logging/Tests/CallbackLogTest.cs
mandel-macaque: We use them to identify what assert failed, but is not bad to add a more decent message.

tests/xharness/Xharness.Tests/Logging/Tests/LogFileTest.cs
mandel-macaque: Is mainly code coverage. Make sure that if we start making it needed, that test fail and we know we change something that will break the code.

tests/xharness/Logging/Logs.cs
premun: Oh, sorry, my bad

tests/xharness/Logging/CaptureLog.cs
premun: `internal` is the default

tests/xharness/Logging/ConsoleLog.cs
mandel-macaque: Agreed, but lets not touch that until we verify it with @rolfbjarne

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: old code. But is a good time to unify.

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: From the docs:

> Class members, including nested classes and structs, can be public, protected internal, protected, internal, private protected, or private. Class and struct members, including nested classes and structs, have private access by default.

[Source](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers)

Since this is a member, not a class, it is private by default. Also, [the mono style guide](https://www.mono-project.com/community/contributing/coding-guidelines/#baroque-coding)

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: Would not set it private, we might use it, but we can remove the property and make it readonly.

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: No, lets imaging we have several task/processes writing to that file (assuming no share violations). If at the point of stop we have an empty file, but when Capture gets called a flush happens, we could be in a situation in which the file has content but our information is outdated. @rolfbjarne is there a reason why you added this?

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: Copying the file seems a reasonable approach.

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: @rolfbjarne does sound reasonable, input?

tests/xharness/Logging/CaptureLog.cs
premun: Oops, my bad, had no idea it's different for members

tests/xharness/Logging/CaptureLog.cs
premun: I would make it public when we need to use it? Maybe it's not going to happen..

tests/xharness/Logging/CaptureLog.cs
mandel-macaque: No problem :)

PR #8036, opened 2020-03-04
[Harness] Increase the timeout on iOS32b devices.

mandel-macaque:
Double the timeout (via a multiplier) when running tests on a iOS 32b
device. This is done because devices are slow and test keep timing out.


mandel-macaque:
Closing, the issue was the pipeline timeout, not the test run timeout.

PR #8037, opened 2020-03-04
[Localization] Rework mlaunch

wachaudh:
Rework the mlaunch localization to fix https://github.com/xamarin/xamarin-macios/issues/8030. This PR relies on https://github.com/xamarin/maccore/pull/2171.


localization/Xamarin.Localization.Mlaunch/Base.cs
```diff
@@ -0,0 +1,14 @@
...
+[assembly: InternalsVisibleTo("mlaunch")]
+namespace Xamarin.Localization.Mlaunch
+{
+	public class Base
+	{
+		public Base()
+		{
+		}
+	}
+}
```
spouliot: is the type really needed ?

if so then make it internal, not public.

wachaudh:
Moved to maccore.

localization/Xamarin.Localization.Mlaunch/Base.cs
wachaudh: It was autogenerated. I removed it.

PR #8047, opened 2020-03-05
[mtouch][mmp] Share error.cs and make it safer against FormatException

spouliot:
Beside the obvious code sharing [1] this makes the error reporting code
safer against `FormatException`.

Even with code review this happened a few times in the past - we get
a string that can't format correctly in some error handling code. This
means the `FormatException` must be fixed before we can see the
underlying root issue (delaying bug fixes and support).

Recent changes made for localization makes it harder to review the
original strings (and even harder for translated strings) for
correctness. This means we can end up with something like

```
String.Format ("{0} {1}", "foo");
```

and easily miss it (since they split between different files) if there
are several changes inside a PR.

Finally sprinkle some unit tests :)

[1] we should move everything to use `ProductException` but that will be
for another PR...


tools/common/error.cs
```diff
@@ -0,0 +1,89 @@
...
+			base (Format (message, args), innerException)
+		{
+			Code = code;
+			Error = error || ErrorHelper.GetWarningLevel (code) == ErrorHelper.WarningLevel.Error;
+		}
+#else
+	public class MonoMacException : Exception {
+		public const string Prefix = "MM";
+
+		public MonoMacException (int code, string message, params object[] args) :
```
rolfbjarne: One way to make this safer is to add an overload without the `params` (for every overload with `params`, so three more constructors). This way passing a verbatim string works, because it won't end up calling string.Format.

spouliot: good idea 👍

tools/common/error.cs
```diff
@@ -0,0 +1,89 @@
...
+			try {
+				return String.Format (message, args);
+			}
+			catch {
+				var sb = new StringBuilder (message);
+				sb.Append (". String.Format failed! Arguments were:");
+				foreach (var arg in args) {
+					sb.Append (" \"").Append (arg).Append ('"');
+				}
+				sb.Append (". Please file an issue to report this incorrect error handling.");
```
wachaudh: Should this message be localized?

tests/mtouch/ErrorTest.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace Xamarin.Test.Bundler {
+
+	[TestFixture]
+	public class ErrorTest {
+
+		[Test]
+		public void Ctor2 ()
+		{
+			var e = new ProductException (0, "valid");
+			Assert.That (e.ToString (), Is.EqualTo ($"warning {ProductException.Prefix}0000: valid"), "ToString-1");
```
stephen-hawley: Why use `Assert.That (..., Is.EqualTo...)` instead of `Assert.AreEqual ()`?

spouliot: it's the newer (now quite old) fluent syntax
it also was the only one part aof NUnitLite when we started monotouch
and the older API, like `AreEqual`, came back in later versions (and eventually NUnitLite was merged back w/Nunit 3)

PR #8048, opened 2020-03-05
[msbuild] Pass the entire TargetFrameworkMoniker to tasks instead of just TargetFrameworkIdentifier/TargetFrameworkVersion.

rolfbjarne:
When .NET 5 comes, the TargetFrameworkMoniker will change, and we need the
entire moniker to distinguish between various platforms.

So change our msbuild code to consume the entire TargetFrameworkMoniker, so
that we have all the information we need when we need it.

Also redirect everything through an intermediate
_ComputedTargetFrameworkMoniker property, so that the target framework can be
overridden without affecting any other code. This becomes necessary during the
initial implementation phase, because we don't have a .NET version to test
with yet that can give us the new target frameworks. Eventually it should be
possible to remove this intermediate variable


msbuild/Xamarin.MacDev.Tasks.Core/PlatformFramework.cs
```diff
@@ -31,27 +31,14 @@ namespace Xamarin.MacDev.Tasks
...
-			case "MonoTouch":
-				return ApplePlatform.iOS;
-			case "Xamarin.WatchOS":
-				return ApplePlatform.WatchOS;
-			case "Xamarin.TVOS":
-				return ApplePlatform.TVOS;
-			default:
-				throw new InvalidOperationException ("Unknown TargetFrameworkIdentifier: " + targetFrameworkIdentifier);
-			}
+			return TargetFramework.Parse (targetFrameworkMoniker).Platform;
```
spouliot: does `Parse` always return something (e.g. never null) or throws ?

rolfbjarne: `TargetFramework` is a struct, so `Parse` returns a struct (an empty struct in case the input couldn't be parsed).

PR #8049, opened 2020-03-05
[AutomaticAssessmentConfiguration] Update bindings to Xcode 11.4 Beta 3

dalexsoto:
Diff: https://github.com/xamarin/xamarin-macios/wiki/AutomaticAssessmentConfiguration-iOS-xcode11.4-beta3


src/automaticassessmentconfiguration.cs
```diff
@@ -0,0 +1,81 @@
...
+	public enum AEAssessmentErrorCode : long {
+		Unknown = 1
+	}
+	
+	[Unavailable (PlatformName.MacCatalyst)]
+	[Mac (10,15,4), iOS (13,4)]
+	[Advice ("This API is not available when using UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	interface AEAssessmentConfiguration : NSCopying {
+
```
spouliot: hmm... does not look very helpful _yet_

dalexsoto: Agreed but nothing much we can do right now heh

spouliot: Did I _type_ that out loud ?
I was thinking of 🍎

PR #8052, opened 2020-03-05
Add dotnet as a configure option and provision it.

rolfbjarne:
Because building the .NET bits makes the build slower, and they're not very
useful quite yet, it's opt-in for developers, but enabled by default on CI.

Also provision the .NET we want to use (for everybody).

### DIFF
--- a/system-dependencies.sh
+++ b/system-dependencies.sh
@@ -74,6 +74,11 @@ while ! test -z $1; do
 			unset IGNORE_SIMULATORS
 			shift
 			;;
+		--provision-dotnet)
+			PROVISION_DOTNET=1
+			unset IGNORE_DOTNET
+			shift
+			;;
 		--provision-all)
 			PROVISION_MONO=1
 			unset IGNORE_MONO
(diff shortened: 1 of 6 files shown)


system-dependencies.sh
```diff
@@ -999,6 +1011,41 @@ function check_simulators ()
...
+	DOTNET_FILENAME=$(basename "$DOTNET_URL")
+
+	if test -d "$DOTNET_INSTALL_DIR"; then
+		ok "Found dotnet $DOTNET_VERSION (exactly $DOTNET_VERSION is required)."
+		return
+	fi
+
+	if test -z "$PROVISION_DOTNET"; then
+		fail "You must install dotnet $DOTNET_VERSION. You can download it from ${COLOR_BLUE}$DOTNET_URL${COLOR_RESET}."
+		fail "Alternatively you can ${COLOR_MAGENTA}export IGNORE_DOTNET=1${COLOR_RED} to skip this check."
```
stephen-hawley: should this include a message about the `--ignore-dotnet` option?

rolfbjarne: You'll usually see this message after running `make`, and then it's must harder to pass the `--ignore-dotnet` option to this script (you have to edit the makefiles) than to export an environment variable.

PR #8055, opened 2020-03-06
[Metal] Update bindings to Xcode 11.4 Beta 3

dalexsoto:


src/Metal/Defs.cs
```diff
@@ -365,8 +365,10 @@ public struct MTLTextureSwizzleChannels {
 		public MTLTextureSwizzle Alpha;
 	}
 
-#if IOS
+#if IOS || MONOMAC || COREBUILD
```
spouliot: is `COREBUILD` really needed ?

PR #8056, opened 2020-03-06
[jenkins] Fix provisioning old macOS bots.

rolfbjarne:
On old macOS bots we only want to provision Mono, so change the logic so that
we first ignore everything, then re-enable Mono.

This way we don't have to add new ignore flags here every time we add
something new to the provisioning script.

This fixes an issue where the script detects that dotnet isn't installed, and
shows an error, when trying to run tests on older macOS bots in CI.

### DIFF
--- a/jenkins/prepare-packaged-macos-tests.sh
+++ b/jenkins/prepare-packaged-macos-tests.sh
@@ -66,7 +66,7 @@ cd mac-test-package
 COUNTER=0
 while [[ $COUNTER -lt 5 ]]; do
 	EC=0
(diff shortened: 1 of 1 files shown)


jenkins/prepare-packaged-macos-tests.sh
```diff
@@ -66,7 +66,7 @@ cd mac-test-package
 COUNTER=0
 while [[ $COUNTER -lt 5 ]]; do
 	EC=0
```
stephen-hawley: style issue: when I have to read a shell script, it means that my day has already gone worse than I anticipated. I prefer shell variables to be as self-describing as possible. It took a couple google searches to find that the intent of `EC`. How about `EXIT_CODE` or `ERROR_CODE`?
As a side note, the `|| EC=$?` is clever.

rolfbjarne: I agree about variable names, I'll have that in mind for the future, since it would be out of scope for this PR.

PR #8061, opened 2020-03-06
[Harness] Split out Targets, Utils and remove external links

premun:
Removes links to external `StringUtils` and `PlistExtensions` and creates `Targets` and `Utilities` namespaces


tests/xharness/xharness.csproj
```diff
@@ -70,87 +70,78 @@
...
+    <Compile Include="BCLTestImporter\RegisterTypeGenerator.cs" />
+    <Compile Include="CrashReportSnapshot.cs" />
+    <Compile Include="DeviceLogCapturer.cs" />
+    <Compile Include="Execution\IProcessManager.cs" />
+    <Compile Include="Execution\ProcessManager.cs" />
+    <Compile Include="GitHub.cs" />
+    <Compile Include="Harness.cs" />
+    <Compile Include="Jenkins\Jenkins.cs" />
+    <Compile Include="Jenkins\Resource.cs" />
+    <Compile Include="Jenkins\Resources.cs" />
```
mandel-macaque: Looks good. But what is going on with these lines.. I'm sure we already had the BCLTest* etc.. something being changed by the IDE?

mandel-macaque:
Looks like something in the refactor broke the tests :/

tests/xharness/Targets/MacTarget.cs
```diff
@@ -56,36 +53,16 @@ protected override void CalculateName ()
...
-				return "{A3F8F2AB-B479-4A4A-A458-A89E7DC349F1};" + LanguageGuid;
-			}
-		}
 
-		protected override string BindingsProjectTypeGuids {
-			get {
-				return "{810C163F-4746-4721-8B8E-88A3673A62EA}";
-			}
-		}
+		protected override string ProjectTypeGuids => "{A3F8F2AB-B479-4A4A-A458-A89E7DC349F1};" + LanguageGuid;
```
rolfbjarne: These type of changes make the PR absolutely horrible to review, because I spend an extraordinary amount of time wondering what changed, just to figure out that nothing actually did. Please do not refactor code just to refactor code.

tests/xharness/xharness.csproj
premun: They are now sorted alphabetically. Should I try to revert this? Might be handy later on..

tests/xharness/Targets/MacTarget.cs
premun: We are refactoring the code as part of this effort though.

There are no functional changes, just a bit heavier IDE auto-formatting, so shouldn't be that hard to review... but if you really feel strong about this, I can revert the contents. Just that it's never going to be refactored after this and it makes the code way shorter to read in the future so will save time later (for a small price now)

tests/xharness/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+			if (f.IndexOfAny (mustQuoteCharactersProcess) == -1)
+				return f;
+
+			var s = new StringBuilder ();
+
+			s.Append ('"');
+			foreach (var c in f) {
+				if (c == '"') {
+					s.Append ('\\');
+					s.Append (c).Append (c);
```
premun: @rolfbjarne can you check this line? It was copied over from tools but through unit tests I discovered that this line does something weird with the output and I am not sure it is 100% correct.
```x"yz```
gets turned into
```x\"""yz```
(both then wrapped in "")
I am not sure this shouldn't have been appending `\` instead or something but I don't really understand the use case.

File: tests/xharness/Targets/MacTarget.cs (hunk 2 above)
Comment (rolfbjarne) [reply tests/xharness/Targets/MacTarget.cs]: > There are no functional changes

The 32 failed tests show that something functional changed.

> but if you really feel strong about this, I can revert the contents

Yes, please.

> Just that it's never going to be refactored after this

If you want to refactor code, do it in a separate PR, and make it as clear as possible (PR description / commit message) that there should only be no functional changes.

It's always much easier to review two (or more!) small PRs than a single big one.

Comment (premun) [reply tests/xharness/Targets/MacTarget.cs]: Reverted.

> It's always much easier to review two (or more!) small PRs than a single big one.

It is easy but not across time zones when on a tight schedule and having changes heavily depend on each other because files are moved around a lot. But I agree it's better to have a single purpose PRs

File: tests/xharness/Utilities/StringUtils.cs (hunk 3 above)
Comment (rolfbjarne) [reply tests/xharness/Utilities/StringUtils.cs]: Yes, that does look strange. That said, if unit tests like this work:

```csharp
using System.Diagnostics;
var p = new Process ();
p.StartInfo.RedirectStandardOutput = true;
p.StartInfo.UseShellExecute = false;
p.StartInfo.FileName = "/bin/echo";                                                                                                                                                                                                                                                                                             
var input = "\"""'"; // some complex input
p.StartInfo.Arguments = "-e " + StringUtils.QuoteForProcess (input);
p.Start ();
var output = p.StandardOutput.ReadToEnd ();
Assert.AreEqual (input, output, "echo");
```

then any changes should be good.

tests/xharness/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+		{
+			return FormatArguments ((IList<string>) arguments);
+		}
+
+		public static string FormatArguments (IList<string> arguments)
+		{
+			return string.Join (" ", QuoteForProcess (arguments));
+		}
+
+		private static string [] QuoteForProcess (params string [] array)
```
rolfbjarne: The Mono Coding Guidelines says to not use `private` (because it's redundant).

```suggestion
	static string [] QuoteForProcess (params string [] array)
```

tests/xharness/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+
+				s.Append (c);
+			}
+			s.Append (shellQuoteChar);
+
+			return s.ToString ();
+		}
+
+		// Quote input according to how System.Diagnostics.Process needs it quoted.
+		private static string QuoteForProcess (string f)
```
rolfbjarne:
```suggestion
		static string QuoteForProcess (string f)
```

tests/xharness/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+					s.Append (c);
+				}
+				s.Append (c);
+			}
+			s.Append ('"');
+
+			return s.ToString ();
+		}
+
+		private static string [] QuoteForProcess (IList<string> arguments)
```
rolfbjarne:
```suggestion
		static string [] QuoteForProcess (IList<string> arguments)
```

tests/xharness/xharness.csproj
mandel-macaque: Nah, not big problem. I was curious

tests/xharness/Targets/MacTarget.cs
mandel-macaque: @premun I have made the same mistake, I will try to avoid fast fingers ;)

PR #8062, opened 2020-03-07
[d16-5-xcode11.4] Merge xcode11.4 into d16-5

dalexsoto:


dalexsoto:
Known failure: HE0038

tools/linker/RemoveRejectedTypesStep.cs
```diff
@@ -0,0 +1,97 @@
...
+
+		public DerivedLinkContext Context {
+			get { return context as DerivedLinkContext; }
+		}
+
+		public override SubStepTargets Targets {
+			get { return SubStepTargets.Assembly; }
+		}
+
+		protected override string Name { get; } = " Removing Rejected Type";
```
whitneyschmidt: Extra space at beginning of `Removing Rejected Type`?

mandel-macaque: To late for this, since is a cherry-pick/backport. We should fix that in the correct location :/

spouliot: Good catch :-)
It's only shown on on errors, obviously we hope never to throw unexpectedly here, and won't affect execution (nor would it be a breaking change)

PR #8064, opened 2020-03-09
[Harness] Refactor simulators and devices listing.

mandel-macaque:
Moved the implementations under Hardware. Add the following to simplify
the testing of the AppRunner:

1. Add interfaces for the listing/loading of devices and simulators.
2. Add a special class for mlaunch args. That way we are sure that
arguments are correctly passed and we do not have typos.
3. Added tests for the new classes and for the device and devices
classes.
4. Added tests for the management of the TCC db.

Tests for the SimDevice and Simulators will come in a following PR. The
only reason for it is that I realize that the commit was getting to
large.


mandel-macaque:
build

tests/xharness/Collections/ILoadAsync.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using Xharness;
+using Xharness.Logging;
+
+namespace xharness.Collections{
+	public interface ILoadAsync {
+		Task LoadAsync (ILog log, bool include_locked, bool force);
+		IHarness Harness { get; set; }
+	}
+}
```
rolfbjarne: Really minor: missing final newline.

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Xharness.Execution {
+	// mlaunch is really important and has a lot of arguments that are known but
+	// used to be passed as strings. This class allows to add arguments without
+	// knowing the exact string and will also validate that an argument that
+	// needs a value does contain the valuw
```
rolfbjarne:
```suggestion
	// needs a value does contain the value
```

tests/xharness/Xharness.Tests/Hardware/Tests/TCCDatabaseTests.cs
```diff
@@ -0,0 +1,143 @@
...
+		[TestCase ("com.apple.CoreSimulator.SimRuntime.tvOS-8-1", 2)]
+		[TestCase ("com.apple.CoreSimulator.SimRuntime.watchOS-5-1", 3)]
+		[TestCase ("com.apple.CoreSimulator.SimRuntime.watchOS-4-1", 2)]
+		public void GetTCCFormatTest (string runtime, int expected)
+		{
+			Assert.AreEqual (expected, database.GetTCCFormat (runtime));
+		}
+
+		[Test]
+		public void GetTCCFormatUnknonwTest ()
```
rolfbjarne:
```suggestion
		public void GetTCCFormatUnknownTest ()
```

tests/xharness/Xharness.Tests/Samples/devices.xml
```diff
@@ -0,0 +1,243 @@
...
+    <IsValid>True</IsValid>
+    <ModelNumber>MC980</ModelNumber>
+    <Name>Manuel de la Pena Saenz’s iPad</Name>
+    <PasswordProtected>False</PasswordProtected>
+    <ProductionSOC>1</ProductionSOC>
+    <ProductType>iPad2,1</ProductType>
+    <ProductVersion>9.3.5</ProductVersion>
+    <SerialNumber>DN6FV2LSDKPJ</SerialNumber>
+    <TotalDataCapacity>29631791104</TotalDataCapacity>
+    <UDID>56f4fd946620062d8f87205aebc7eb4fc9fb565d</UDID>
```
rolfbjarne: Might want to use dummy data for UDID/DeviceIdentifier and serial numbers.

tests/xharness/Hardware/Device.cs
```diff
@@ -0,0 +1,144 @@
...
+		public bool IsLocked { get; set; }
+
+		public string UDID { get { return DeviceIdentifier; } set { DeviceIdentifier = value; } }
+
+		public string OSVersion => ProductVersion;
+
+		// Add a speed property that can be used to sort a list of devices according to speed.
+		public int DebugSpeed => InterfaceType?.ToLowerInvariant () switch
+		{
+			"usb" => 0,
```
rolfbjarne: Please don't refactor code when moving it around, it makes everything much harder to review. For instance: the comments that explained why this order in particular disappeared.

tests/xharness/Collections/BlockingEnumerableCollection.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace xharness.Collections {
+	// This is a collection whose enumerator will wait enumerating until 
+	// the collection has been marked as completed (but the enumerator can still
+	// be created; this allows the creation of linq queries whose execution is
+	// delayed until later).
+	internal class BlockingEnumerableCollection<T> : IEnumerable<T> where T : class {
```
premun: I think this could use test but no hard feelings around that

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
...
+		ListSim,
+		OutputFormat,
+		ListExtraData,
+	}
+
+	struct MlaunchArgument {
+		public MlaunchArgumentType Type { get; set; }
+		public string Value { get; set; }
+
+		public MlaunchArgument ((MlaunchArgumentType type, string value) data)
```
premun: >(MlaunchArgumentType type, string value) [](start = 26, length = 40)

This one maybe doesn't have to be a tuple

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
...
+
+		public MlaunchArgument (MlaunchArgumentType type)
+		{
+			Type = type;
+			Value = null;
+			if (!IsValid (out var reason))
+				throw new ArgumentException (reason);
+		} 
+
+		bool IsValid (out string reason)
```
premun: >IsValid [](start = 7, length = 7)

This one could be called `Validate` and throw right away. No out needed?

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
...
+			default:
+				break;
+			}
+			return true;
+		}
+
+		// return the string needed to be used in the cmd
+		public string AsCommandLineArgument () => Type switch
+		{
+			MlaunchArgumentType.SdkRoot => $"--sdkroot={Value}",
```
premun: >{Value} [](start = 46, length = 7)

Escaping is taken care of on a different layer?

tests/xharness/Hardware/SimDevice.cs
```diff
@@ -0,0 +1,132 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+using Xharness;
+using Xharness.Execution;
+using Xharness.Logging;
+
+namespace xharness.Hardware {
```
premun: >Hardware [](start = 19, length = 8)

Not sure this namespace shouldn't be `Devices`. Not sure I would look for Simulator in Hardware?

tests/xharness/Hardware/SimDevice.cs
```diff
@@ -0,0 +1,132 @@
...
+		public ITCCDatabase TCCDatabase { get; set; } = new TCCDatabase ();
+
+
+		public string UDID { get; set; }
+		public string Name { get; set; }
+		public string SimRuntime { get; set; }
+		public string SimDeviceType { get; set; }
+		public string DataPath { get; set; }
+		public string LogPath { get; set; }
+		public string SystemLog { get { return Path.Combine (LogPath, "system.log"); } }
```
premun:
```suggestion
		public string SystemLog => Path.Combine (LogPath, "system.log");
```

tests/xharness/Hardware/SimDevice.cs
```diff
@@ -0,0 +1,132 @@
...
+
+		public IHarness Harness;
+
+		public bool IsWatchSimulator { get { return SimRuntime.StartsWith ("com.apple.CoreSimulator.SimRuntime.watchOS", StringComparison.Ordinal); } }
```
premun:
```suggestion
		public bool IsWatchSimulator => SimRuntime.StartsWith ("com.apple.CoreSimulator.SimRuntime.watchOS", StringComparison.Ordinal);
```

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Xharness.Execution {
+	// mlaunch is really important and has a lot of arguments that are known but
+	// used to be passed as strings. This class allows to add arguments without
+	// knowing the exact string and will also validate that an argument that
+	// needs a value does contain the valuw
+	public enum MlaunchArgumentType {
```
premun: >MlaunchArgumentType [](start = 13, length = 19)

I checked all the usages of Harness.MlaunchPath and it seems there are way more many usages and more arguments.

Is the plan to migrate those piece by piece later or they are left out unintentionally?

tests/xharness/Execution/MLaunchArguments.cs
```diff
@@ -0,0 +1,102 @@
...
+		public MlaunchArgument ((MlaunchArgumentType type, string value) data)
+		{
+			Type = data.type;
+			Value = data.value;
+			if (!IsValid (out var reason)) 
+				throw new ArgumentException (reason);
+		}
+
+		public MlaunchArgument (MlaunchArgumentType type)
+		{
```
premun: To be honest, it would be better to have a class for each argument class accepting exactly what it needs as value (maybe there's none, maybe it needs an array, maybe just a string..)

There would be `abstract MlaunchArgument` class and the argument classes would implement some `ToCommandLine` method. Validation would happen in ctor of each of those.

It might also help in future because instead of providing the tuple (Type, Value) you would have the value named so you know what it is, something like:
`new ListSimArgument(deviceId: value)` whereas now you actually cannot tell what the argument is (is it device ID, is it iOS version...)

But again, no hard feelings around this, just the only thing I mind is that now the code lets you create invalid argument on compile time by letting you (not) supply value.

tests/xharness/Execution/ProcessManager.cs
```diff
@@ -59,6 +59,12 @@ public Task<ProcessExecutionResult> RunAsync (Process process, ILog log, TimeSpa
 			return RunAsync (process, log, log, log, timeout, environment_variables, cancellation_token, diagnostics);
 		}
 
+		public Task<ProcessExecutionResult> RunAsync (Process process, MlaunchArguments args, ILog log, TimeSpan? timeout = null, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null, bool? diagnostics = null)
```
premun: This could also set the mlaunch executable path to the process info

tests/xharness/Hardware/Simulators.cs
```diff
@@ -0,0 +1,416 @@
...
+				try {
+					using (var process = new Process ()) {
+						process.StartInfo.FileName = Harness.MlaunchPath;
+						var arguments = new MlaunchArguments (
+							(MlaunchArgumentType.SdkRoot, Harness.XcodeRoot),
+							(MlaunchArgumentType.ListSim, tmpfile)
+						);
+						process.StartInfo.Arguments = arguments.AsCommandLine ();
+						log.WriteLine ("Launching {0} {1}", process.StartInfo.FileName, process.StartInfo.Arguments);
+						var rv = await ProcessManager.RunAsync (process, log, timeout: TimeSpan.FromSeconds (30));
```
premun: Not sure if part of this change but later this should also be taken care of by the `ProcessManager`

tests/xharness/Hardware/Device.cs
mandel-macaque: My bad, will re-add comment.

tests/xharness/AppRunner.cs
```diff
@@ -9,7 +9,7 @@
 using System.Threading.Tasks;
 using System.Xml;
 using Xamarin;
-using Xamarin.Utils;
+using xharness.Hardware;
```
premun: nit

```suggestion
using Xharness.Hardware;
```

tests/xharness/Execution/MLaunchArguments.cs
premun: I can see now in other usages that we might benefit from this more by having specialized constructors such as the `-argument=-app-arg:-something` one

tests/xharness/Hardware/Simulators.cs
mandel-macaque: yes, will come later. Had to cut the PR to not be a gazillion lines :/

tests/xharness/Execution/ProcessManager.cs
mandel-macaque: Yep

tests/xharness/Execution/MLaunchArguments.cs
mandel-macaque: yes. this was a first implementation. I think it will evolve until we find an API that will cover all the cases. At least is much more secure than using strings.

tests/xharness/Execution/MLaunchArguments.cs
mandel-macaque: Correct, jus moving the ones I'm touching atm.

tests/xharness/Execution/MLaunchArguments.cs
mandel-macaque: true, no need

PR #8065, opened 2020-03-09
[xharness] Handle UTF8 text better.

rolfbjarne:
* Set UTF8 as the encoding for process output.
* Serve plain text as UTF-8 plain text from our test server.

This makes emojiis render as emojiis 😁😁😁


rolfbjarne:
build

PR #8075, opened 2020-03-10
[FileProvider] Update bindings to Xcode 11.4 Beta 3

dalexsoto:
We got the deprecated attributes! Long live our test suites!

### DIFF
--- a/src/fileprovider.cs
+++ b/src/fileprovider.cs
@@ -490,6 +490,8 @@ namespace FileProvider {
 
 #if XAMCORE_4_0 // Not available in mac
 		[NoMac]
+#elif MONOMAC
+		[Obsolete ("'IsTrashed' is not available in macOS and will be removed in the future.")]
 #endif
 		[Export ("isTrashed")]
 		bool IsTrashed ();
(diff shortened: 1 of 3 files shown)


src/fileprovider.cs
```diff
@@ -488,6 +488,9 @@ interface NSFileProviderItem {
 		[Export ("favoriteRank", ArgumentSemantic.Copy)]
 		NSNumber GetFavoriteRank ();
 
+#if XAMCORE_4_0 // Not available in mac
+		[NoMac]
+#endif
```
spouliot: let's add an `[Obsolete]` message (on macOS) so people won't think it should work
*and* won't be surprised if the member is removed later (at `XAMCORE_4_0` time)

dalexsoto: Good call 👍

dalexsoto:
```suggestion
#if XAMCORE_4_0 // Not available in mac
		[NoMac]
#elif MONOMAC
		[Obsolete ("'IsTrashed' is not available in macOS and will be removed in the future.")]
#endif
```

dalexsoto: Fixed, thanks @spouliot

PR #8076, opened 2020-03-10
[mmp] Only iOS has dual bitness now, no need to check for Xamarin.Mac.

rolfbjarne:


rolfbjarne:

PR #8081, opened 2020-03-10
[StoreKit] Update bindings for Xcode 11.4 beta 3

whitneyschmidt:
Most of these changes are to add StoreKit as a new framework available on WatchOS.

It would be great to get extra close feedback on this one since it's my first time adding a new WatchOS framework :)


src/storekit.cs
```diff
@@ -205,7 +218,7 @@ interface SKPaymentQueue {
 		[Export ("cancelDownloads:")]
 		void CancelDownloads (SKDownload [] downloads);
 
-		[Mac (10, 15), iOS (13, 0)]
+		[Mac (10, 15), iOS (13, 0)] // NoWatch??
```
whitneyschmidt: Should this explicitly be labeled as `NoWatch`?

src/storekit.cs
```diff
@@ -350,7 +372,7 @@ interface SKPaymentTransaction {
 		string TransactionIdentifier { get; }
 
 #if !MONOMAC
-		[Availability (Deprecated = Platform.iOS_7_0, Message = "Use 'NSBundle.AppStoreReceiptUrl' instead.")]
+		[Availability (Deprecated = Platform.iOS_7_0, Message = "Use 'NSBundle.AppStoreReceiptUrl' instead.")] // NoWatch??
```
whitneyschmidt: Same as above - should this be `[NoWatch]`?

src/storekit.cs
```diff
@@ -450,7 +480,7 @@ interface SKProductsRequestDelegate {
 		void ReceivedResponse (SKProductsRequest request, SKProductsResponse response);
 	}
 
-#if !MONOMAC
+#if !MONOMAC && !WATCH
```
whitneyschmidt: `!WATCH` b/c `UIViewController` is not available on Watch.

src/storekit.cs
```diff
@@ -617,11 +647,13 @@ interface SKCloudServiceController {
 		NSString StorefrontCountryCodeDidChangeNotification { get; }
 	}
 
+#if !WATCH
```
whitneyschmidt: `UIViewController` is not available on Watch

src/storekit.cs
```diff
@@ -760,7 +793,7 @@ interface SKProductSubscriptionPeriod {
 		SKProductPeriodUnit Unit { get; }
 	}
 
-	[iOS (11,2), TV (11,2), Mac (10,13,2)]
+	[Watch (4,2), iOS (11,2), TV (11,2), Mac (10,13,2)]
```
whitneyschmidt: Should this be `Watch(6,2)`? Apple docs say that it's been available since `Watch(4,2)`, but we are only adding it now...

src/storekit.cs
whitneyschmidt: Or is there a preference for doing this w/ macros vs at the top of the file?

src/StoreKit/NativeMethods.cs
```diff
@@ -5,7 +5,7 @@
 namespace StoreKit {
 
 	partial class SKReceiptRefreshRequest {
-		[iOS (7,1), Mac (10,14)]
+		[Watch (6,2), iOS (7,1), Mac (10,14)]
```
spouliot: probably not needed
since StoreKit is new in watchOS 6.2 then you only have to decorate all (available) types with the attributes
IOW you don't need to add them on members

src/storekit.cs
```diff
@@ -24,6 +24,7 @@ partial interface SKDownload {
 
 		[iOS (12,0)]
 		[TV (12,0)]
+		[Watch (6,2)]
```
spouliot: move to type-level

src/storekit.cs
```diff
@@ -132,34 +140,39 @@ interface SKMutablePayment {
 		[Availability (Deprecated = Platform.iOS_5_0, Message = "Use 'PaymentWithProduct (SKProduct)' after fetching the list of available products from 'SKProductRequest' instead.")]
 		SKMutablePayment PaymentWithProduct (string identifier);
 
+		[Watch (6, 2)]
```
spouliot: move to type level

it's important because now that StoreKit is included (in `Xamarin.WatchOS.dll`) it means that the lack of attribute correspond to the earliest supported version (i.e. watchOS 2.0).

src/storekit.cs
spouliot: yes (unless the headers says otherwise)

It often happens Apple will put macros for `watchOS(unavailable)` on an API even if the framework is not available. We normally don't add those (since it only increase the metadata size without any benefits)

src/storekit.cs
spouliot: yes, even more if it's deprecated
in such case it might be present but we don't want to encourage people to use it :)

src/storekit.cs
```diff
@@ -617,11 +647,13 @@ interface SKCloudServiceController {
 		NSString StorefrontCountryCodeDidChangeNotification { get; }
 	}
 
+#if !WATCH
 	[iOS (10,1)]
 	[NoTV] // __TVOS_PROHIBITED
 	[BaseType (typeof(UIViewController))]
 	interface SKCloudServiceSetupViewController
 	{
+#endif // !WATCH
```
spouliot: ^ I think your `#endif` should be after the end of the type
(or I'm missing something)

src/storekit.cs
spouliot: likely a typo from Apple
the framework was not available before 6.2 - so the API was not

src/storekit.cs
mandel-macaque: yes, [NoWatch] should be used.

src/storekit.cs
```diff
@@ -785,14 +818,15 @@ interface SKProductDiscount {
 		[Export ("paymentMode")]
 		SKProductDiscountPaymentMode PaymentMode { get; }
 
+		[Watch (6,2)]
```
spouliot: not needed (once type level is fixed)

src/storekit.cs
mandel-macaque: yes, same, use NoWatch, easier to add it later.

src/storekit.cs
mandel-macaque: it needs to be tested to make sure, if we did not add it there most be a reason. Add it to 6,0 which is safer than adding an API to a later version and then not being able to remove it.

src/storekit.cs
```diff
@@ -852,13 +888,17 @@ interface SKStorefront {
...
 	[Protocol]
 	[Model (AutoGeneratedName = true)]
 	[BaseType (typeof(NSObject))]
 	interface SKPaymentQueueDelegate {
 		[Export ("paymentQueue:shouldContinueTransaction:inStorefront:")]
 		bool ShouldContinueTransaction (SKPaymentQueue paymentQueue, SKPaymentTransaction transaction, SKStorefront newStorefront);
+
+		[NoWatch, NoMac, iOS (13,4)]
+		[Export ("paymentQueueShouldShowPriceConsent:")]
+		bool PaymentQueueShouldShowPriceConsent (SKPaymentQueue paymentQueue);
```
spouliot: `PaymentQueue` prefix comes from the protocol and the fact that there's only one parameter
It's not needed (like the API above)

src/storekit.cs
mandel-macaque: You can create an empty interface for UIViewController at the top under Watch, then use NoWatch, once it available (if ever) you can find all the interfaces in the same place.

tools/common/Frameworks.cs
```diff
@@ -391,6 +391,7 @@ public static Frameworks GetwatchOSFrameworks (Application app)
 				{ "PushKit", "PushKit", 6 },
 				{ "SoundAnalysis", "SoundAnalysis", 6 },
 				{ "CoreMedia", "CoreMedia", 6 },
+				{ "StoreKit", "StoreKit", 6 }
```
spouliot: 6,2 ;-)

src/storekit.cs
spouliot: whatever is the simplest / shortest
sometimes you only need one `#if` but a dozen `using`
other times you can avoid a dozen `#if` with a single `using`

whitneyschmidt:
Closing and moving this PR to the `d16-5-xcode11.4` branch.

PR #8083, opened 2020-03-11
[Harness] Separate WrenchLog

premun:
This will later allow us to remove some dependencies on the Harness class


premun:
This branch is on top of a master that was reverted and will make this change later as part of some other refactoring as it is not big.

PR #8085, opened 2020-03-11
[Harness] Move utilities out from Harness.cs, refactor AppRunner fields

premun:
Move all the extension methods to a class. After this refactor, we will
be able to DI the manager in the other classes and assert that the
processes are called with the correct parameters without the need of
launching them.


premun:
This branch has digressed a lot and will be fixed in #8117 and another follow-up PR for Harness

PR #8086, opened 2020-03-11
[Harness] Refactor simulators and devices listing.

mandel-macaque:
This is a revert of the revert of https://github.com/xamarin/xamarin-macios/commit/e53b686ef63705f2381a83adfb3a5bcfac88eb73 that has the same contents as https://github.com/xamarin/xamarin-macios/pull/8064 but without the merge of the commits that caused the issue.

Original PR comments:

Moved the implementations under Hardware. Add the following to simplify
the testing of the AppRunner:

Add interfaces for the listing/loading of devices and simulators.
Add a special class for mlaunch args. That way we are sure that
arguments are correctly passed and we do not have typos.
Added tests for the new classes and for the device and devices
classes.
Added tests for the management of the TCC db.
Tests for the SimDevice and Simulators will come in a following PR. The
only reason for it is that I realize that the commit was getting to
large.


tests/xharness/Collections/BlockingEnumerableCollection.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace xharness.Collections {
```
spouliot: `X` ^

tests/xharness/Collections/IAsyncEnumerable.cs
```diff
@@ -0,0 +1,7 @@
+﻿using System.Threading.Tasks;
+
+namespace xharness.Collections {
```
spouliot: `X`

tests/xharness/Collections/ILoadAsync.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using Xharness;
+using Xharness.Logging;
+
+namespace xharness.Collections{
```
spouliot: `X`

tests/xharness/Collections/ILoadAsync.cs
```diff
@@ -0,0 +1,10 @@
...
+	public interface ILoadAsync {
+		Task LoadAsync (ILog log, bool include_locked, bool force);
```
premun: >include_locked [](start = 33, length = 14)

Here it says `include_locked` but in the `Devices` implementation it is `remove_locked` and in `Simulators` impl it is `include_locked`.
In Jenkins it is that used as `ILoadAsync` which might then cause different behaviour based on what is inside the `ILoadAsync` pointer

premun: cc @rolfbjarne

---
In reply to: [390979607](https://github.com/xamarin/xamarin-macios/pull/8086#discussion_r390979607) [](ancestors = 390979607)

premun: I will open an issue to track this and we can merge here

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1956

PR #8087, opened 2020-03-11
[linker] Fix some strings (no effect on execution)

spouliot:
* Extra whitespace in step name. Found by Whitney reviewing a backport;
* Two errors not using (localizable) resources strings. I likely forgot to add the file to the original PR.

### DIFF
--- a/tools/linker/RemoveRejectedTypesStep.cs
+++ b/tools/linker/RemoveRejectedTypesStep.cs
@@ -26,7 +26,7 @@ namespace Xamarin.Linker {
 			get { return SubStepTargets.Assembly; }
 		}
 
-		protected override string Name { get; } = " Removing Rejected Type";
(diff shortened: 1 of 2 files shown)


spouliot:
tvOS failure (only seen in XML)
```xml
<test-case id="1126" name="RedirectionWithAuthorizationHeaders(System.Net.Http.NSUrlSessionHandler)" fullname="MonoTests.System.Net.Http.MessageHandlerTest.RedirectionWithAuthorizationHeaders(System.Net.Http.NSUrlSessionHandler)" result="Inconclusive" label="Inconclusive" time="00:00:00.2150000" asserts="0">
<reason>
<message>
<![CDATA[
Response from httpbin does not contain headers, therefore we cannot ensure that if the authoriation is present.
]]>
</message>
</reason>
</test-case>
</test-suite>
```

filed as https://github.com/xamarin/maccore/issues/2174

PR #8090, opened 2020-03-12
[Harness] Move the templates src to be embedded resources.

mandel-macaque:
Move the entire templates dir as a embedded resource. The idea is to have a class that will generate the required scaffold app to run bcl without the need of any external file. The next step is to make all the code that generates the code to be in the template class and simply choose a template class to be used for the tests.


tests/xharness/BCLTestImporter/Templates/Managed/XamariniOSTemplate.cs
```diff
@@ -48,5 +74,60 @@ Stream GetTemplateStream (string templateName)
 		public Stream GetProjectTemplate (Platform platform) => GetTemplateStream (projectTemplateMatches [platform]);
 
 		public Stream GetProjectTemplate (WatchAppType appType) => GetTemplateStream (watchOSProjectTemplateMatches [appType]);
+
+		void BuildSrcTree (string srcOuputPath)
+		{
+			// loop over the known paths, and build them accodringly
```
rolfbjarne:
```suggestion
			// loop over the known paths, and build them accordingly
```

Issue #7534, opened 2019-12-03
Redundant dispatch to UI thread for thread-safe objects finalizers


Therzok:
### Steps to Reproduce

1. Create a thread-safe object, i.e. `NSString`
2. Don't dispose it
3. Observe that finalizer queues [disposal on the UI thread](https://github.com/xamarin/xamarin-macios/blob/bc6735374b1bb31d58e228304ea89815c57004a1/src/Foundation/NSObject2.cs#L781-L792)

### Expected Behavior

Thread-safe types don't need to be released on the UI thread.

### Possible Problem

NSString is not sealed, thus subclasses can become thread-unsafe. One example of this is NSMutableString.

Therzok:
I'm actually wondering if derived classes actually come into play. Due to the finalizer only being run on the current object, it should not matter. `NSMutableString` still doesn't have any UI thread dependency.

Releasing the managed ref on `NSString` or `NSMutableString` even though they are subclassed and include managed objects, it will not trigger the managed ref removal on any of the new fields added.

So, it's all a matter of figuring out how to automate most of the annotations. Would it be framework-level? Should it be applied whenever ThreadSafeAttribute is applied?

I suppose that the generator could just add another [flag to NSObject](https://github.com/xamarin/xamarin-macios/blob/afb11d65cc26f0c629957120f07be35525d1f014/src/Foundation/NSObject2.cs#L76-L83) when ThreadSafe is annotated and then not dispatch to the UI thread.

Therzok:
I would like to try and prototype this for NSString/NSMutableString. I guess something like a new Attribute and `IsCustomType == false` is enough to decide whether to dispatch to the UI thread for ref-counting. `ThreadSafeAttribute` is most likely subset of the `ThreadIndependentAttribute` (naming things is hard).

@rolfbjarne @mandel-macaque @chamons I have a few questions:

❗️How does the runtime handle classes marked with ThreadSafe that are subclassed? Is there a bug hiding there that may warrant a fix?

❓ Not sure why NSString - or anything at all in Foundation - is not marked ThreadSafe.

❔Is checking for `IsCustomType` enough to judge whether it's a user type?

❔This is a really tricky problem, but I guess it's possible to propagate the attribute properties on subclasses, but would probably need some analysis on the objects it uses on the finalizer.

Therzok:
[...]

PR #8092, opened 2020-03-12
Implement a way to have non-UI thread finalization possible.

Therzok:
Updated the NSObjectFlags to be a short, and that way we can fit new
values in at the code of possibly slightly increased memory usage,
but it might be covered by padding anyway.

Added NSObjectFlagsRequiresUIThread which is then used in the generator
to signal whether the UI thread is required for object finalization.

Also added a feature flag to enable this behaviour, while keeping the
old behaviour for compatibility.

Not sure what to do about user types, so it keeps the old behaviour for
now.

(The benchmark has the same issue that @slluis reported in: https://github.com/xamarin/xamarin-macios/issues/13101)

Benchmark implementation: https://github.com/Therzok/xamarin-macios/blob/ui-thread-gen-with-benchmarks/tests/dotnet/GCBenchmark/ViewController.cs#L14

The feature flag is enabled in Main.cs. The tests below were run on intel macOS.

Benchmark with feature enabled: https://gist.github.com/Therzok/99eff9483b8a5f7b33a4d8a6f0a2225c

Benchmark without feature enabled: https://gist.github.com/Therzok/2541a9c308b9506cdd84f291bee81844

Benchmarks wrapped in NSAutoreleasePool by using DangerousRetain().DangerousRelease(): 
https://gist.github.com/Therzok/09bdc573e68c737d8d2f6fc41c5f8cd1

Fixes https://github.com/xamarin/xamarin-macios/issues/7534


src/Foundation/NSObject2.cs
```diff
@@ -81,6 +81,8 @@ enum Flags : byte {
 			HasManagedRef = 32,
 			// 64, // Used by SoM
 			IsCustomType = 128,
+			// TODO: Which version would generate less code, RequiresUIThread or the negated version?
+			RequiresUIThread = 256,
```
Therzok: How can I check the size increase here?

Therzok:
A bit overzealous, it's placing it in a lot of places

src/Foundation/NSObject2.cs
rolfbjarne: There is probably a lot more API that is thread-safe than not thread-safe, since only these frameworks are considered UI frameworks and get the UI thread check:

https://github.com/xamarin/xamarin-macios/blob/16a84bd7f7abe6d05ca9dc8abd0e97333d14fecf/src/generator.cs#L684-L710

so my guess is that the way you implemented is would be the smallest size increase.

src/Foundation/NSObject2.cs
```diff
@@ -72,7 +72,7 @@ public partial class NSObject
 
 		// This enum has a native counterpart in runtime.h
 		[Flags]
-		enum Flags : byte {
+		enum Flags : ushort {
```
rolfbjarne: Any changes to this enum, especially its size, require several changes in the runtime code as well (runtime.m/runtime.h/etc).

spouliot:
Speaking of 
```
public static bool CheckForIllegalCrossThreadCalls = true; 
 public static bool CheckForEventAndDelegateMismatches = true; 
```
They are, in hindsight, a bad design decision - location wise.

It does not matter for most people, except if you have a custom way to start the runtime (like vs4m does). The issue is that you're calling into `NSApplication` to set them... which calls its `.cctor` and everything the app needs might not yet be loaded by the OS (if they are called too soon).

There's a fix (might not cover all possible cases) in `master` for that.

src/Foundation/NSObject2.cs
Therzok: I suppose we can't break xamarin_set_nsobject_flags, so can I add a Flags2 enum?

rolfbjarne: Yes, that's not a bad idea. The new enum should be internal though.

Therzok:
Added feature flag.

runtime/xamarin/runtime.h
```diff
@@ -149,6 +149,10 @@ enum NSObjectFlags {
 	NSObjectFlagsIsCustomType = 128,
 };
 
+enum NSObjectFlags2 {
```
Therzok: I suppose that if the new enum should be private/internal, this is not the right place for it.

rolfbjarne: That's fine: https://github.com/xamarin/xamarin-macios/blob/0f12c101e0da3968fedb540ee75070f7f0ed5f82/runtime/xamarin/runtime.h#L4-L5

spouliot:
Let's measure first (like run the PR on the samples) as the impact might not be noticeable - considering it affects only some (mostly UI) types and the linker will remove a bunch of them.

Even if noticeable we might get better results/saving by tweeking other things (even more with .net5) but we can file it as an enhancement.

Therzok:
If there is any sample that creates a lot of garbage objects, the impact should be pretty noticeable.

rolfbjarne:
OK, I've scheduled a sample test run, let's see what happens.

Therzok:
@rolfbjarne The sample test will run with the UI thread check disabled though.

Therzok:
I can push a change to flip it

Therzok:
Oh, that's to check the size impact on the samples.

spouliot:
@Therzok the sample tests do not execute the samples. They build them (multiple configurations), making sure tooling is fine (no regression) and also record the build times and app sizes for analysis.

Therzok:
What else is needed here?

rolfbjarne:
The samples didn't trigger, I just did it manually.

rolfbjarne:
App size increases: https://gist.github.com/rolfbjarne/4deeb04b96f16ba697f06cbd0a9709e7

On average each app increased 1807 bytes, and every app had a size increase.

That's between 51b6b707e44627961ab42db9968e0274a9a88f4b and 7472778506e7b954a41bfb483c2a84e56d4ae5c0, there are a few other changes (https://github.com/xamarin/xamarin-macios/compare/51b6b707e44627961ab42db9968e0274a9a88f4b...7472778506e7b954a41bfb483c2a84e56d4ae5c0), but most are unrelated (and non-shipping) changes. I couldn't find any other explanation for the size increase.

Therzok:
Is that an acceptable increase?

spouliot:
> On average each app increased 1807 bytes, and every app had a size increase.

Nice table. The distribution looks weird. Most are around 250 bytes with a few over 16kb.

Makes me wonder if some samples are build with the managed linker being disabled.

> Is that an acceptable increase?

That always depends on what it gives back. If you can't measure the positive side then yes, then what we ship is a visible *and* measurable negative hit.

> I couldn't find any other explanation for the size increase.

Agreed. There's not much that should have a size impact beside the PR code and the mono bump.

https://github.com/mono/mono/compare/b33fa1d00ed3b3c16c9abf8c623bb089c6ddc669..2ff898859b8b75f123a5c895959a7e82dc829ea5

> A bit overzealous, it's placing it in a lot of places

Indeed, the PR diff made me think this was done on constructors only. Why not make this decision at the type level ? that would only require changing the constructors [1].

Once you covered them (ctors) then the additional `RequiresUIThread = true;` are no-ops - since the flag is never set back to `false` under any condition. If also feels weird that calling some members would affect how we dispose the instance.

[1] the chaining constructors, e.g. `.ctor(IntPtr)`, needs to set this too - but they are much fewer than the members. The final increase should be a lot smaller (or at least similar to the smallest ones).

So, even if we decide to add linker support (not sure it's needed with the above), it seems the current code is not optimal - and that should be fixed first.

Therzok:
Yeah, it's done for constructors only in the latest version.

spouliot:
Oh, so it looks like I clicked on an older generator diff and my reading of the diff was not too far off :) 

This one looks better: https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/12969//Generator_20Diff

However it does not handle the chaining constructors I mentioned. So

* if a native instance is given to us we'll be calling `.ctor(IntPtr)` and the field won't be set to `true`;

* if someone (manually) subclass a type then the `.ctor(NSObjectFlag)` won't set the field to `true`

It also looks like we're missing the `global::UIKit.UIApplication.EnsureUIThread ();` for the default `init` and `initWithCoder` (less common) that are generated elsewhere. Since you are piggybacking on the check then your field won't be `true` for both uses.

Therzok:
> if a native instance is given to us we'll be calling .ctor(IntPtr) and the field won't be set to true;

Will try and see how I can fix that. Doesn't it automatically get it from the MonoObject struct?

> if someone (manually) subclass a type then the .ctor(NSObjectFlag) won't set the field to true

Oh, that's going to be tricky. How can we solve that? I used `IsDirectBinding` to check if it's a XamMac binding, I suppose `IsCustomUserType` should be used?

> That always depends on what it gives back. If you can't measure the positive side then yes, then what we ship is a visible and measurable negative hit.

We kind of can measure the positive. In large applications, the GC time and overall application memory use over time will be decreased noticeable - since objects won't survive first collection automatically.

spouliot:
If you receive an instance from ObjC (mono is unaware) all you get is a pointer and we call (thru a bunch of things) the `.ctor(IntPtr)`.

That's generated code (so easy to fix). `.ctor(NSObjectFlag)` is also generated code and does not require additional knowledge.

E.g. If you subclass `NSView` then you'll chain to `: base (NSObjectFlag.Empty)` so that generated code must include setting the field to `true` (and you already know it's required for `NSView` itself, so it also applies to all subclasses)

> We kind of can measure the positive. 

That be extra nice for the release notes. Saying things are better, without numbers, is too often disregarded or quickly overlooked :)

Therzok:
Just saw there's a PKG associated with the PR, now I can actually test it :D

spouliot:
I'm working on the generator right now (nullability) so I'll take care of the missing `EnsureUIThread` calls -> https://github.com/xamarin/xamarin-macios/issues/8249

spouliot:
@Therzok yes :) but, in your case, you also need the `.ctor(NSObjectFlag)` for subclasses

Therzok:
Rudimentary benchmark for now, will work on improving it:
https://github.com/Therzok/XamMacIOSUIThreadCheck/blob/1db20bef310d2ef0b37b13db757b7794bbd4eb13/MacUIThreadCHeck/AllocatorBenchmark.cs#L15-L98

Output run in Debug: https://gist.github.com/Therzok/4bda5d2274a7ae99fd3a1c2ba9faed31

We see a sizeable decrease in number of live NSString objects and memory usage for a scenario where we allocate `50000 NSString` objects every second, each on a UI thread callback.

Therzok:
I'll create a test matrix, for all the permutations: FeatureFlag=on/off, NSStringCount=0,50000, NSViewCount=0,50000, Config=Debug,Release and Platform=mac/iphonesim

Therzok:
I will resurrect this PR once we're able to bump the Xamarin.Mac we build VSMac against. I suppose the conflicts are due to moving more NSObject data into managed.

Therzok:
I've updated this PR. Will create some benchmarks for macOS (I don't have iOS simulators installed or an iOS device to benchmark those - but I can provide the benchmark implementation)

rolfbjarne:
This doesn't build:

```
coreclr-bridge.m:298:39: error: implicit conversion loses integer precision: 'enum NSObjectFlags' to 'uint8_t' (aka 'unsigned char') [-Werror,-Wimplicit-int-conversion]
        res = xamarin_gc_toggleref_callback (flags, handle, NULL, NULL);
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ^~~~~
1 error generated.
make[1]: *** [.libs/mac/coreclr-bridge-dotnet-coreclr.x86_64.o] Error 1
```

Therzok:
Updated PR description with benchmark numbers, something's weird with the numbers, the GC is highly indeterministic, even with induced collections.

Therzok:
Updated again! The numbers still don't make sense, even though I am sure the objects are tied to the autoreleasepools now: https://gist.github.com/Therzok/09bdc573e68c737d8d2f6fc41c5f8cd1

Overall, the GC time is dramatically reduced, even though the NSView version should have not changed at all.

Therzok:
@rolfbjarne ping

Therzok:
With a more recent version of xammac, the test app crashes, I'll try and figure out what is causing the crashes.

Therzok:
I don't think this PR will work, I need to redo the changes since the enums were moved to managed afaik.

I'll look into it this week.

src/Foundation/NSObject2.cs
```diff
@@ -186,6 +187,11 @@ struct objc_super {
 			}
 		}
 
+		internal bool RequiresUIThread {
```
Therzok, line 190: Good news is that we now have GC metrics we're monitoring in CI, so once we insert it, we can validate how much this improves things.

PR #8093, opened 2020-03-12
[Harness] Make Harness properties immutable

premun:
Making most of the properties immutable or at least not mutable from outside of Harness.

This is a first step that will enable easier refactoring of AppRunner and Harness where we will be able to do the same for AppRunner and move lot of the initialization to the constructor.
We can then handle the state of the class better when we know what is and what is not changing during the lifetime, This will enable us to rewrite some of the actions such as Initialize, Install or Uninstall in a way that they can be unit tested before we refactor them properly.


tests/xharness/Program.cs
```diff
@@ -90,16 +115,44 @@ public static int Main (string [] args)
 			var input = os.Parse (args);
 			if (input.Count > 0)
 				throw new Exception (string.Format ("Unknown arguments: {0}", string.Join (", ", input.ToArray ())));
-			if (harness.Action == HarnessAction.None)
+			
+			if (action == HarnessAction.None)
 				showHelp ();
```
premun: @rolfbjarne should we `return 1` here?

tests/xharness/Program.cs
```diff
@@ -1,82 +1,107 @@
...
-				{ "target=", "Where to run the project ([ios|watchos|tvos]-[device|simulator|simulator-32|simulator-64]).", (v) => harness.Target = v.ParseAsAppRunnerTarget () },
-				{ "configuration=", "Which configuration to run (defaults to Debug).", (v) => harness.Configuration = v },
-				{ "logdirectory=", "Where to store logs.", (v) => harness.LogDirectory = v },
-				{ "logfile=", "Where to store the log.", (v) => harness.LogFile = v },
-				{ "timeout=", "Timeout for a test run (in minutes). Default is 10 minutes.", (v) => harness.Timeout = double.Parse (v) },
+				{ "target=", "Where to run the project ([ios|watchos|tvos]-[device|simulator|simulator-32|simulator-64]).", (v) => target = v.ParseAsAppRunnerTarget () },
+				{ "configuration=", "Which configuration to run (defaults to Debug).", (v) => configuration = v },
+				{ "logdirectory=", "Where to store logs.", (v) => logDirectory = v },
+				{ "logfile=", "Where to store the log.", (v) => logFile = v },
+				{ "timeout=", $"Timeout for a test run (in minutes). Default is {timeout} minutes.", (v) => timeout = double.Parse (v) },
```
premun: This timeout was actually 15 (default of Harness)

tests/xharness/Program.cs
rolfbjarne: Yes, I think that's fine.

tests/xharness/Harness.cs
```diff
@@ -80,52 +80,104 @@ public class Harness : IHarness
...
+		public string PeriodicCommand { get; }
+		public string PeriodicCommandArguments { get; }
+		public TimeSpan PeriodicCommandInterval { get; }
 		// whether tests that require access to system resources (system contacts, photo library, etc) should be executed or not
 		public bool? IncludeSystemPermissionTests { get; set; }
 
-		public Harness ()
-		{
+		public Harness (IProcessManager processManager,
+			HarnessAction action,
```
rolfbjarne: A constructor with 26 required arguments is not a particularly pretty sight...

I think it would be much better to separate all the command-line argument data into a separate class, and just pass an instance of that class to the constructor.

premun: I am trying to solve following problem - someone initializing the class and forgetting to provide an argument.
If I enforce arguments in ctor, I can make them readonly and know that the state of my object is always correct.
I am not sure how else to enforce this other than method arguments.
If I make a wrapper argument class, this class will have to have this ctor and I haven't helped myself much?

I can do that though, if you think it's better.

rolfbjarne: The idea of a wrapper class is something like this:

```csharp
class HarnessArguments {
    public bool AutoConf;
    public string Configuration;
    ...
}

class Harness {
    public bool AutoConf { get; }

    public Harness (HarnessArguments arguments)
    {
        AutoConf = arguments.AutoConf;

        // maybe verify that the arguments provided make sense.
    }
}
```

this way you can still make all the Harness fields readonly if you wish.

IMHO enforcing that all the arguments have been provided doesn't make much sense, because it's never the case that all arguments are needed, some are needed some times, and some are needed other times.

tests/xharness/Program.cs
premun: Ah, `showHelp` already calls `Environment.Exti(0)`..

mandel-macaque:
@premun code looks good and the test failure you got is known. I would take @rolfbjarne and land it :)

tests/xharness/Harness.cs
premun: I understand the idea.. I just don't like the fact that now if we add a new field, we must look for all places where we instantiate the arguments class and we won't know on compile time that we forgot to set a field somewhere. While this is perfectly solvable on compile time rather than runtime... I think here we really should aim for maximum correctness not visual appearance of the code and let the code tell us as early as possible something is wrong.

Anyway, I added the class..

tests/xharness/Harness.cs
```diff
@@ -31,16 +31,42 @@ public interface IHarness {
...
-		public HarnessAction Action { get; set; }
-		public int Verbosity { get; set; }
+		public HarnessAction Action { get; }
+		public int Verbosity { get; }
 		public ILog HarnessLog { get; set; }
-		public bool UseSystem { get; set; } // if the system XI/XM should be used, or the locally build XI/XM.
-		public HashSet<string> Labels { get; } = new HashSet<string> ();
-		public XmlResultJargon XmlJargon { get; set; } = XmlResultJargon.NUnitV3;
-		public IProcessManager ProcessManager { get; set; } = new ProcessManager ();
+		public bool UseSystem { get; } // if the system XI/XM should be used, or the locally build XI/XM.
```
stephen-hawley: `UseSystem` is a vague name. Is there a reason why it couldn't be named `UseSystemXamarinIOSMac` or with the opposite connotation `UseLocalXamarinIOSMac`?

tests/xharness/Harness.cs
```diff
@@ -80,52 +106,80 @@ public class Harness : IHarness
...
+			Action = action;
+
+			AutoConf = configuration.AutoConf;
+			Configuration = configuration.Configuration ?? throw new ArgumentNullException (nameof (configuration));
+			DryRun = configuration.DryRun;
+			EnvironmentVariables = configuration.EnvironmentVariables ?? new Dictionary<string, string> ();
+			IncludeSystemPermissionTests = configuration.IncludeSystemPermissionTests;
+			IOSTestProjects = configuration.IOSTestProjects;
+			JenkinsConfiguration = configuration.JenkinsConfiguration;
+			Labels = configuration.Labels ?? new HashSet<string> ();
```
stephen-hawley: I worry about this - the `Labels` property is set to the object in the configuration. If the configuration changes, so does this since they share the same object. Maybe it would be better to set the `Labels` to a default empty set and use `if (configuration.Labels != null) Labels.UnionWith (configuration.Labels);` I see the same issue with `EnvironmentVariables` as well.

tests/xharness/Harness.cs
premun: I would not rename fields as part of this refactoring, if possible

tests/xharness/Harness.cs
premun: I am very much against having the `HarnessConfiguration` wrapper because it is misleading on what it does when we change it.

I was considering referencing the `HarnessConfiguration` inside `Harness` and then have the other properties point to it.

But to be honest, I would just not have it at all and have everything handed over in ctor like on the previous iteration of this PR.

premun: But a good point otherwise

tests/xharness/Harness.cs
premun: Although, this wouldn't hurt

premun: Done

mandel-macaque:
@premun issues are known: https://github.com/xamarin/maccore/issues/1956 and https://github.com/xamarin/maccore/issues/581

PR #8098, opened 2020-03-12
[Harness] Fix bug in TCCDb editing that is missing using the watchapp id.

mandel-macaque:
The wrong var name was used and the table was not correctly updated.
Updated tests to take into account the new \n in the sql.


mandel-macaque:
Failure is unrelated. tests pass.

PR #8099, opened 2020-03-12
[runtime] Abort if the call to Runtime.Initialize fails.

rolfbjarne:
Also print out something to the system log. In theory
xamarin_process_managed_exception should also end up terminating the process
in one way or another, but when something goes wrong it tend to go very, very
wrong, so this makes sure some info is printed somewhere.


rolfbjarne:
Api diff is failing due to https://github.com/xamarin/maccore/issues/2176.

PR #8104, opened 2020-03-12
[StoreKit] Update bindings for Xcode 11.4 beta 3

whitneyschmidt:
Original PR against `xcode11.4` branch can be found here: https://github.com/xamarin/xamarin-macios/pull/8081


src/storekit.cs
```diff
@@ -76,21 +82,23 @@ partial interface SKDownload {
 		void DeleteContentForProduct (string productId);
 #endif
 
-		[Mac (10,14)]
+		[Mac (10, 14)]
```
mandel-macaque: Not needed space change.

src/storekit.cs
```diff
@@ -76,21 +82,23 @@ partial interface SKDownload {
...
 		[Field ("SKDownloadTimeRemainingUnknown")]
 		double TimeRemainingUnknown { get; }
 
-		[Mac (10,11)]
+		[Mac (10, 11)]
```
mandel-macaque: Not needed space change.

src/storekit.cs
```diff
@@ -132,34 +140,39 @@ interface SKMutablePayment {
...
 		NSData RequestData { get; set; }
 
-		[iOS (7,0), Mac (10, 9)]
+		[Watch (6, 2), iOS (7,0), Mac (10, 9)]
 		[NullAllowed] // by default this property is null
 		[Export ("applicationUsername", ArgumentSemantic.Copy)][New]
 		string ApplicationUsername { get; set; }
 
-		[iOS (8,3), Mac (10,14)]
+		[Watch (6, 2), iOS (8,3), Mac (10,14)]
```
mandel-macaque: We should be consistent between [Watch (6, 2)] and [Watch (6,2)]. I know we are not very, but in this case you are using both in the same PR.

src/frameworks.sources
```diff
@@ -2029,6 +2029,7 @@ WATCHOS_FRAMEWORKS =        \
 	PassKit                 \
 	PushKit			\
 	Security                \
+	StoreKit                \
```
mandel-macaque: Looking at the complete file, StoreKit is present in all platforms, Mac, iOS, tvOS and watchOS. At this point will should remove it f rom all the subsets and add it to COMMON_FRAMEWORKS

src/storekit.cs
```diff
@@ -132,34 +140,39 @@ interface SKMutablePayment {
 		[Availability (Deprecated = Platform.iOS_5_0, Message = "Use 'PaymentWithProduct (SKProduct)' after fetching the list of available products from 'SKProductRequest' instead.")]
 		SKMutablePayment PaymentWithProduct (string identifier);
 
+		[Watch (6, 2)]
```
spouliot: those should be on the type
because without a [Watch] on the type we'll assume it's been there since watchOS 2.0 (and that's not correct)

src/storekit.cs
```diff
@@ -24,6 +27,7 @@ partial interface SKDownload {
 
 		[iOS (12,0)]
 		[TV (12,0)]
+		[Watch (6, 2)]
```
spouliot: move to type (see other message)

src/storekit.cs
whitneyschmidt: Ah, got it. I will go through all the APIs and mark them as appropriate :)

src/storekit.cs
```diff
@@ -393,12 +419,14 @@ interface SKRequestDelegate {
 	[Mac (10,9)]
 	[BaseType (typeof (SKRequest))]
```
mandel-macaque: As @spouliot mentioned with others, move the [Watch (6,2)] to the top, just need to add one and the wraps should not need any attr decoration.

mandel-macaque:
@whitneyschmidt looks like there is a introspection issue with SKProductStorePromotionController, looking at the docs, is not present in the watchos

src/storekit.cs
```diff
@@ -27,18 +31,22 @@ partial interface SKDownload {
 		[Export ("state")]
 		SKDownloadState State { get; }
 #if MONOMAC
+		[NoWatch] // not necessary b/c #if MONOMAC, so this will not even compile for ios/watch/tvos?
```
whitneyschmidt: It seems like this is not necessary because we have `#if MONOMAC` and it was not marked as `[NoTV][NoiOS]` before?

src/storekit.cs
```diff
@@ -239,10 +259,12 @@ interface SKProduct {
 		string ProductIdentifier { get; }
 
 #if MONOMAC
+		[NoWatch] // not necessary b/c #if MONOMAC, so this will not even compile for ios/watch/tvos?
 		[Deprecated (PlatformName.MacOSX, 10,15, message: "Use 'IsDownloadable' instead.")]
 		[Export ("downloadable")]
 		bool Downloadable { get; }
 #elif !XAMCORE_4_0
+		[NoWatch] // IS THIS RIGHT?
```
whitneyschmidt: Not sure about this one, but since it is `Obsolete` it seems like we would not want users to have access to it starting in `[Watch (6, 2)]`?

src/storekit.cs
```diff
@@ -6,7 +6,7 @@
 //   Miguel de Icaza
 //
 // Copyright 2009, Novell, Inc.
-// Copyright 2012 Xamarin Inc.
+// Copyright 2020 Xamarin Inc.
```
spouliot: don't remove, replace - just append
in this case it should be a new line for 2020 / Microsoft

src/storekit.cs
spouliot: no, it's not required
it would tell, in Xamarin.Mac.dll, that this is not available on watchOS - but that's not very useful

src/storekit.cs
spouliot: no, it's not needed

some times when Apple enable an existing framework in a new platform then bring everything, including deprecated API :(
Since this is replaced by something newer/better, that is always available on the new platform, then we can reduce warnings and let intellisense provide better choice to the developers

if xtro complains (because it's in the header) then add a line in watchOS-StoreKit.ignore

whitneyschmidt:
build

spouliot:
@whitneyschmidt please don't issue a new `build` without checking the logs first. This was a known HE0038 issue so it could be merged (with a note).

PR #8108, opened 2020-03-13
Fix API comparison. Fixes #xamarin/maccore@2176.

rolfbjarne:
* Make everything in src/ build into a directory dependent upon BUILD_DIR.
  That way the API comparison can successfully build into a different
  directory.
* Use an absolute path as the temporary build directory. This works better
  when the path is used in directories with different depth from the root.

Fixes https://github.com/xamarin/maccore/issues/2176.

The API comparison will still fail for this PR/commit, the fix won't take
effect until the next commit.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,7 +2,7 @@ TOP=..
 include $(TOP)/Make.config
 
 BUILD_DIR=build
-DOTNET_BUILD_DIR=build/dotnet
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Not sure why the html report wasn't posted here, but it succeeded:

https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/13005/Test_20Report/

API comparison failed (as expected), it will start working in the next commit/PR.

PR #8109, opened 2020-03-13
Fix path to dont link project in Xamarin.Mac.sln.

rolfbjarne:


rolfbjarne:
API comparison failure will be fixed with PR #8108.

PR #8117, opened 2020-03-15
[Harness] Refactor and unit-test AppRunner

premun:
- Makes dependencies of AppRunner injectable to allow testing - this needs some factories to be added
- Revise life-cycle of AppRunner properties/fields:
  - Make as many readonly and private as possible - initialize early as it's already happening anyway
  - Strong type fields (RunMode, DeviceClass..)
- Add unit tests for AppRunner initialization, `InstallAsync` and `UninstallAsync`


tests/xharness/Harness.cs
```diff
@@ -57,8 +53,78 @@ public class HarnessConfiguration {
 		public XmlResultJargon XmlJargon { get; set; } = XmlResultJargon.NUnitV3;
 	}
 
-	public class Harness : IHarness
-	{
+	public interface IHarness {
```
premun: This will be trimmed down in a follow-up PR

tests/xharness/Harness.cs
```diff
@@ -871,7 +948,8 @@ public async Task<ILogFile> SymbolicateCrashReportAsync (ILogs logs, ILog log, I
 			var symbolicated = logs.Create (Path.ChangeExtension (name, ".symbolicated.log"), $"Symbolicated crash report: {name}", timestamp: false);
 			var environment = new Dictionary<string, string> { { "DEVELOPER_DIR", Path.Combine (XcodeRoot, "Contents", "Developer") } };
 			var rv = await ProcessManager.ExecuteCommandAsync (symbolicatecrash, new [] { report.Path }, symbolicated, TimeSpan.FromMinutes (1), environment);
-			if (rv.Succeeded) {;
+			if (rv.Succeeded) {
+				;
```
rolfbjarne:
```suggestion
```

tests/xharness/UnsupportedOperationException.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Xharness {
+	public class UnsupportedOperationException : Exception {
```
rolfbjarne: Sounds like this is the same as `InvalidOperationException`, could we use that exception type instead of creating a new one?

premun: I am using it in tests to distinguish that proper behavior happened. With `InvalidOperationException`  I was worried some other either ours or not even ours code would throw the same. I also wasn't going to base the unit test on comparing exception message. What do you think?

rolfbjarne: That seems like a rather remote possibility, that you'd get `InvalidOperationException` for a different cause than the expected one. I still think it's best to not create a new exception type, because the best practices for exceptions say ["Introduce a new exception class only when a predefined one doesn't apply"](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#use-the-predefined-net-exception-types).

premun: I reverted to `InvalidOperationException`

PR #8118, opened 2020-03-16
[msbuild] The Mmp/Mtouch tasks are passed the expected TargetFrameworkMoniker, so no need to have any custom logic here.

rolfbjarne:
We can even move it all to the shared base class, and remove an unused variable (`UseXamMacFullFramework`).

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
+++ b/msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
@@ -28,8 +28,6 @@ namespace Xamarin.Mac.Tasks
 
 		public bool IsXPCService { get; set; }
 
-		public bool UseXamMacFullFramework { get; set; }
-
 		public string ApplicationName { get; set; }
 		public string Architecture { get; set; }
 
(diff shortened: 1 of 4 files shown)


msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
```diff
@@ -537,7 +537,6 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 			ApplicationName="$(_AppBundleName)"
 			MainAssembly="$(OutputPath)$(TargetName)$(TargetExt)"
 			TargetFrameworkMoniker="$(_ComputedTargetFrameworkMoniker)"
-			UseXamMacFullFramework="$(UseXamMacFullFramework)"
```
spouliot: was the `UseXamMacFullFramework` removal intentional ?
it's not mentioned in the commit message and this is specific to XM

msbuild/Xamarin.Mac.Tasks.Core/Tasks/MmpTaskBase.cs
```diff
@@ -28,8 +28,6 @@ public class MmpTaskBase : BundlerToolTaskBase
 
 		public bool IsXPCService { get; set; }
 
-		public bool UseXamMacFullFramework { get; set; }
```
spouliot: was the `UseXamMacFullFramework` removal intentional ?
it's not mentioned in the commit message and this is specific to XM

rolfbjarne: Yes, because it's not used anymore. I've updated the PR description and will add it to the commit message.

rolfbjarne:
Test failure is due to https://github.com/xamarin/maccore/issues/2177.

PR #8122, opened 2020-03-16
[Harness] Move all test project generation to the template object.

mandel-macaque:
At this point the BCLTestProjectGenerator class just represents a configuration class that knows about the project that are particular to xamarin-macios. All the logic that creates the projects has been moved to the XamariniOSTemplate which creates the project using the information passed by the config class. 

At this stage, we have the ability to generate test projects by simply passing the info classes, which can be built either by a class (like the generator one) or a command line. We have fully decoupled the project generation from Xharness.

Tests have been added or moved to the correct location.


tests/xharness/BCLTestImporter/Templates/Managed/XamariniOSTemplate.cs
```diff
@@ -75,16 +104,25 @@ Stream GetTemplateStream (string templateName)
 
 		public Stream GetProjectTemplate (WatchAppType appType) => GetTemplateStream (watchOSProjectTemplateMatches [appType]);
 
+		public Stream GetRegisterTypeTemplate () => GetTemplateStream (registerTemplateResourceName);
+
 		void BuildSrcTree (string srcOuputPath)
 		{
-			// loop over the known paths, and build them accordingly
+			// loop over the known paths, and build them accodringly
```
mandel-macaque: Hmm merge failure here.

mandel-macaque:
```suggestion
			// loop over the known paths, and build them accordingly
```

tests/xharness/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -201,508 +134,61 @@ public class BclTestProjectInfo {
...
-				Path.Combine (rootPath, "xunit-excludes.txt")
+			OutputDirectoryPath = outputDirectory ?? throw new ArgumentNullException (nameof (outputDirectory));
+			AssemblyLocator = new AssemblyLocator ();
+			ProjectFilter = new ProjectFilter { AssemblyLocator = AssemblyLocator, IgnoreFilesRootDir = outputDirectory };
+			TemplatedProject = new XamariniOSTemplate {
+				AssemblyLocator = AssemblyLocator,
+				OutputDirectoryPath = outputDirectory,
+				IgnoreFilesRootDirectory = outputDirectory,
+				ProjectFilter = ProjectFilter,
+				//public Func<string, Guid> GuidGenerator { get; set; }
```
mandel-macaque:
```suggestion
```
Remove comment (also I can trigger a build with the run-bcl-tests tag ;)

tests/xharness/Utilities/ProjectFileExtensions.cs
```diff
@@ -538,14 +539,15 @@ public static void FixInfoPListInclude (this XmlDocument csproj, string suffix,
 			var import = GetInfoPListNode (csproj, false);
 			if (import != null) {
 				var value = import.Attributes ["Include"].Value;
-				var fname = System.IO.Path.GetFileName (value.Replace ('\\', '/'));
+				var unixValue = value.Replace ('\\', '/');
```
premun: >'/' [](start = 41, length = 3)

This could also be `Environment.DirectorySeparator` or what is the exact name

mandel-macaque: hmm I can give it a try, but if it is ran on Windows, it won't be a unix path, right?

tests/xharness/BCLTestImporter/BCLTestProjectGenerator.cs
```diff
@@ -201,508 +134,60 @@ public class BclTestProjectInfo {
...
-		{
-			switch (appType) {
-				case WatchAppType.App:
-					return Path.Combine (rootDir, "Info-watchos-app.plist");
-				default:
-					return Path.Combine (rootDir, "Info-watchos-extension.plist");
+		public string MacMonoSDKPath {
+			get {
+				var locator = AssemblyLocator as AssemblyLocator;
+				return locator?.MacMonoSDKPath;
```
premun: I think it's safer to make the property of type `AssemblyLocator` and forget about the interface since we are not even injecting it at this point?

tests/xharness/BCLTestImporter/Templates/IProjectFilter.cs
```diff
@@ -0,0 +1,49 @@
...
+	/// </summary>
+	public interface IProjectFilter {
+
+		/// <summary>
+		/// True if a specific dlls should be ignored from a project in a platform.
+		/// </summary>
+		/// <param name="platform">The platform whose project is going to be generated.</param>
+		/// <param name="assembly">The assembly that is part of the test project.</param>
+		/// <returns>True if the dll was to be excluded, false otherwhise.</returns>
+		bool ExcludeDll (Platform platform, string assembly);
```
premun: >ExcludeDll [](start = 7, length = 10)

I think the naming here might be better when something like `ShouldExclude` or `ShouldBeExcluded` or `IsDllExcluded`...
`Exclude` only hints that there will be a side effect of some sort.

But no hard feelings about this, up to you.

tests/xharness/BCLTestImporter/Templates/IProjectFilter.cs
```diff
@@ -0,0 +1,49 @@
...
+		/// <summary>
+		/// Returns the list of ignore files that will be added to a project so that certain tests are ignored. This
+		/// is useful when we know that there are flacky tests that we want to ignore but they have not yet been ignored
+		/// by the project that builds the test dlls.
+		/// </summary>
+		/// <param name="projectName">The name of the project being generated.</param>
+		/// <param name="assemblies">The assemblies information.</param>
+		/// <param name="platform">The plaform targented by the generated project.</param>
+		/// <returns>An enumerable with all the ignore text files to use.</returns>
+		IEnumerable<string> GetIgnoreFiles (string projectName, List<(string assembly, string hintPath)> assemblies, Platform platform);
```
premun: >List [](start = 58, length = 4)

`IEnumerable`?

tests/xharness/Utilities/ProjectFileExtensions.cs
premun: Yeah, I guess I am just mentioning it because unit tests then fail for me on Windows but that is not a big issue

tests/xharness/BCLTestImporter/BCLTestProjectGenerator.cs
mandel-macaque: We have it, but in this case, this class is specific for the xamarin-macios project. I don't want the interface to have the mono sdk paths in the general case when running on helix, does not make much sense

mandel-macaque:
Known issues: 

* https://github.com/xamarin/maccore/issues/1956 
* https://github.com/xamarin/maccore/issues/1488

PR #8125, opened 2020-03-16
[Harness] Move utilities out of Harness.cs

premun:
This is re-introduction of #8083 and #8085 on top of the newly changed master and making the Harness class more single responsibility so that it can be more easily replaced

Just moving code around (out of `Harness.cs`), no functional changes.


mandel-macaque:
@premun some of the failures are know flaky tests. The 

```
The command "make -j8 -C /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/test-libraries"
```

Has not only happened in this branch. But does look that we might have broken something in the refactoring. We should take a closer look in a diff branch.

mandel-macaque:
There are several PRs with the same issue https://github.com/xamarin/maccore/issues/2179 so merging and will fix in another commit.

PR #8126, opened 2020-03-17
[NetworkExtension] Update bindings to Xcode 11.4 Beta 3

dalexsoto:
https://github.com/xamarin/xamarin-macios/wiki/NetworkExtension-iOS-xcode11.4-beta1
https://github.com/xamarin/xamarin-macios/wiki/NetworkExtension-iOS-xcode11.4-beta3


src/networkextension.cs
```diff
@@ -963,13 +1014,45 @@ interface NETunnelProviderManager
...
-	
+
+		[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
+		[Static]
+		[Export ("forPerAppVPN")]
+		NETunnelProviderManager CreatePerAppVpn ();
+
+		[return: NullAllowed]
+		[Export ("copyAppRules")]
+		NEAppRule[] CopyAppRules ();
```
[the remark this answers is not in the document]
dalexsoto: From my point of view it does not feel right to have this as a property, feels like copyWithZone:’ish, also swift has this as a method, here is the doc description 

> Create a copy of the app rules currently set in the configuration.

src/networkextension.cs
```diff
@@ -146,10 +161,26 @@ interface NEAppProxyFlow {
...
 
+		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
+		[NullAllowed, Export ("networkInterface", ArgumentSemantic.Copy)]
+		OS_nw_interface WeakNetworkInterface { get; set; }
+
+		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
+		NWInterface NetworkInterface {
+			[Wrap ("Runtime.GetINativeObject<NWInterface> (WeakNetworkInterface, true)")]
+			get;
+			[Wrap ("WeakNetworkInterface = value.Handle")]
```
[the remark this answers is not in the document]
dalexsoto:
```suggestion
			[Wrap ("WeakNetworkInterface = value != null ? value.Handle : IntPtr.Zero")]
```

src/networkextension.cs
```diff
@@ -146,10 +161,26 @@ interface NEAppProxyFlow {
...
 
+		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
+		[NullAllowed, Export ("networkInterface", ArgumentSemantic.Copy)]
+		OS_nw_interface WeakNetworkInterface { get; set; }
+
+		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
+		NWInterface NetworkInterface {
+			[Wrap ("Runtime.GetINativeObject<NWInterface> (WeakNetworkInterface, true)")]
+			get;
+			[Wrap ("WeakNetworkInterface = value != null ? value.Handle : IntPtr.Zero")]
```
spouliot: the `GetHandle ()` extension method would have done the same ;-)

src/networkextension.cs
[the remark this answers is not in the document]
spouliot: k, was not sure if `Copy` was a verb (`Copy` + `AppRules`) or qualifier of `AppRules`, i.e. `CopyApp` + `Rules`

rolfbjarne: I wonder if needs a `[return: Release ()]`?

tests/xtro-sharpie/common-NetworkExtension.ignore
```diff
@@ -1 +1 @@
-!missing-release-attribute-on-return-value! NetworkExtension.NEAppRule[] NetworkExtension.NETunnelProviderManager::get_AppRules()'s selector's ('copyAppRules') Objective-C method family ('copy') indicates that the native method returns a retained object, and as such a '[return: Release]' attribute is required.
+!missing-release-attribute-on-return-value! NetworkExtension.NEAppRule[] NetworkExtension.NETunnelProviderManager::CopyAppRules()'s selector's ('copyAppRules') Objective-C method family ('copy') indicates that the native method returns a retained object, and as such a '[return: Release]' attribute is required.
```
rolfbjarne: Looks like it does...

src/networkextension.cs
dalexsoto: You can’t do that because is a managed array there is no Handle

tests/xtro-sharpie/common-NetworkExtension.ignore
dalexsoto: We return a managed array

src/networkextension.cs
dalexsoto: Ahhh I always forget about it heh

src/networkextension.cs
rolfbjarne: We should fix the generator if it complains, because otherwise we'll leak.

src/networkextension.cs
```diff
@@ -146,10 +161,26 @@ interface NEAppProxyFlow {
...
+		[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
+		[Export ("setMetadata:")]
+		void SetMetadata (OS_nw_parameters parameters);
 	
 		[Export ("metaData")]
 		NEFlowMetaData MetaData { get; }
 
+		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
+		[NullAllowed, Export ("networkInterface", ArgumentSemantic.Copy)]
+		OS_nw_interface WeakNetworkInterface { get; set; }
```
spouliot: It might be lack of caffeine (I'm working on that right now) but I don't recall using the `Weak` pattern for an `IntPtr`, which is what `OS_nw_interface` ends up to be, inside `INativeObject`.

src/networkextension.cs
dalexsoto: Yeah you are right, we are orphaning the native array, I'll look into it

spouliot: the generator code _should_ cope with that, if the attribute is present, and return the same managed array.

dalexsoto: It does not, I'll fix it

rolfbjarne: I guess it could be done in another PR though, since it's not new API.

src/networkextension.cs
dalexsoto: This is because I am providing a more concrete property that Wraps the IntPtr version, maybe I should make `WeakNetworkInterface` internal or teach the generator about [NativeObject](https://github.com/xamarin/xamarin-macios/blob/master/src/CoreFoundation/NativeObject.cs) objects.
```csharp
		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
		[NullAllowed, Export ("networkInterface", ArgumentSemantic.Copy)]
		OS_nw_interface WeakNetworkInterface { get; set; }

		[NoWatch, NoTV, Mac (10,15,4), iOS (13,4)]
		NWInterface NetworkInterface {
			[Wrap ("Runtime.GetINativeObject<NWInterface> (WeakNetworkInterface, false)")]
			get;
			[Wrap ("WeakNetworkInterface = value != null ? value.Handle : IntPtr.Zero")]
			set;
		}
```

dalexsoto: @spouliot  Thinking about it, maybe I should do both, make the property internal and in a separated PR teach the generator about [NativeObject](https://github.com/xamarin/xamarin-macios/blob/master/src/CoreFoundation/NativeObject.cs) objects.

src/networkextension.cs
dalexsoto: Agreed

src/networkextension.cs
```diff
@@ -146,10 +161,26 @@ interface NEAppProxyFlow {
 	
 		[Export ("closeWriteWithError:")]
 		void CloseWrite ([NullAllowed] NSError error);
+
+		[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
+		[Export ("setMetadata:")]
+		void SetMetadata (OS_nw_parameters parameters);
```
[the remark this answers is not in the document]
dalexsoto: Looks like they are unrelated, also swift has the [setMetadata](https://developer.apple.com/documentation/networkextension/neappproxyflow/3543399-setmetadata) method using this protocol [OS_nw_parameters](https://developer.apple.com/documentation/network/os_nw_parameters?language=objc) which is defined as `typedef NSObject<OS_nw_parameters> *nw_parameters_t;` so maybe this might be a better(?) binding?

```csharp
[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
[Export ("setMetadata:")]
void SetMetadata (OS_nw_parameters parameters);
```

@spouliot what do you think?

src/networkextension.cs
spouliot: I'm not sure we need (at least _right now_) special generator support...

We normally use `Weak` when there are other _easy_ ways to use the API, e.g. `NSDictionary` (instead of strong dictionary) and `NSObject` for delegates (since you implement the members on a different type).

I'm not sure those applies to `IntPtr` and, as usual, it's much easier to expose it later than to remove it :)

src/networkextension.cs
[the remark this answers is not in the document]
dalexsoto: Or just expose `OS_nw_parameters` protocol with a better C# name and just bind it as` IOS_nw_parameters` 🤔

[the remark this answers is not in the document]
dalexsoto: @mandel-macaque I would also like your take on this one since you bound `Network` :)

src/networkextension.cs
dalexsoto: So making the weak internal and just exposing the wrapper it is

src/networkextension.cs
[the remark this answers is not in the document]
spouliot: hmm... we have API that returns `nw_parameters_t` instances

https://developer.apple.com/documentation/network/2976839-nw_parameters_create_secure_tcp?language=objc

so why not use `NWParameters` ?

It's (again) mixing `NSObject`/`INativeObject`, which requires some additional code glue, but it's the right thing to do, i.e.

* `void SetMetadata (NSObject parameters);` would not work with an instance of `NWParameters`

* `void SetMetadata (IntPtr parameters);` would _work_ with an instance of `NWParameters`, but would be _needlessly_ ugly `SetMetadata (x.Handle);` and error prone (no type check).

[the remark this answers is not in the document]
dalexsoto: @spouliot because I am dumb and missed the `NWParameters` parameters class 🙈, I'll use that instead

src/networkextension.cs
```diff
@@ -162,13 +163,19 @@ interface NEAppProxyFlow {
...
 		void CloseWrite ([NullAllowed] NSError error);
 
+		[Internal]
 		[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
 		[Export ("setMetadata:")]
-		void SetMetadata (OS_nw_parameters parameters);
+		void SetMetadata (OS_nw_parameters nwparameters);
+
+		[NoWatch, NoTV, NoiOS, Mac (10,15,4)]
+		[Wrap ("SetMetadata (parameters.Handle)")]
```
spouliot: NRE ;-) use `GetHandle ()`

dalexsoto: lol, too much cmd+z to file https://github.com/xamarin/xamarin-macios/issues/8127 🤣

dalexsoto:
```suggestion
		[Wrap ("SetMetadata (parameters.GetHandle ())")]
```

dalexsoto:
@spouliot / @rolfbjarne I think I have resolved your feedback but the generator change for `[return: Release]` when returning managed arrays which I filled here https://github.com/xamarin/xamarin-macios/issues/8127

PR #8129, opened 2020-03-17
[Harness] Make CrashSnapshotReporter injectable and add unit tests

premun:
Another dependency of AppRunner that will need to be refactored out to make AppRunner testable


tests/xharness/CrashSnapshotReporter.cs
```diff
@@ -0,0 +1,206 @@
...
+		readonly bool isDevice;
+		readonly string deviceName;
+		readonly Func<string> tempFileProvider;
+		readonly string symbolicateCrashPath;
+
+		HashSet<string> initialCrashes;
+
+		public CrashSnapshotReporter (IProcessManager processManager,
+									  ILog log,
+								      ILogs logs,
```
mandel-macaque: Funny alignment, spaces and tabs?

tests/xharness/Execution/ProcessManager.cs
```diff
@@ -17,7 +17,12 @@ public ProcessManager ()
 		{
 		}
 
-		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename, IList<string> args, ILog log, TimeSpan timeout, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null)
+		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
+																	   IList<string> args,
```
mandel-macaque: Funny alignment.

tests/xharness/Execution/ProcessManager.cs
```diff
@@ -26,6 +31,20 @@ public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
 			}
 		}
 
+		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
+																	   MlaunchArguments args,
+																	   ILog log,
+																	   TimeSpan timeout,
+																	   Dictionary<string, string> environment_variables = null,
```
mandel-macaque: Same.

tests/xharness/AppRunner.cs
```diff
@@ -747,7 +749,7 @@ public async Task<int> RunAsync ()
 			if (!success.Value) {
 				int pid = 0;
 				string crash_reason = null;
-				foreach (var crash in crash_reports.Logs) {
+				foreach (var crash in Logs) {
```
rolfbjarne: This will make xharness process every log as if it was a crash report, and that's quite wasteful, some of the logs can be quite big. It's better to just process logs we know might be crash reports.

premun: I was thinking the same when writing this  but my refactor uncovered possible problems like these because if you look at the original:
1. `Logs` is passed to the ctor above on line 434
2. I was able to make `Logs` readonly inside `CrashReportSnapshot` because it is only assigned in ctor so it never changes
3. Then we read the value back here but it's the same `Logs` we pass in so I didn't change anything and we already do what you are saying. Or I am wrong somehow but I let the code show me this while refactoring

tests/xharness/Execution/ProcessManager.cs
premun: Weird, the tab size in PR review is different than in Code on GH. I guess you said it should be 8 but I see 4 everywhere (and the way the code is structured)

tests/xharness/AppRunner.cs
rolfbjarne: The Logs field/property itself may not have changed, but there can have been any number of Log instances added to the Logs instance after it was created.

premun: I understand that. But do you want me to fix it or leave it working like it is (was)?

rolfbjarne: Please fix it to only iterate over actual crash reports instead of every log.

premun: Fixed

tests/xharness/AppRunner.cs
```diff
@@ -423,15 +426,16 @@ public bool TestsSucceeded (AppInformation appInfo, string test_log_path, bool t
...
 			ILog listener_log = null;
 			ILog run_log = MainLog;
 
 			if (!isSimulator)
 				FindDevice ();
 
-			crash_reports = new CrashReportSnapshot (harness, MainLog, Logs, isDevice: !isSimulator, deviceName);
+			var crashLogs = new Logs (Logs.Directory);
+
+			ICrashSnapshotReporter crash_reports = snapshotReporterFactory.Create (MainLog, crashLogs, isDevice: !isSimulator, deviceName);
```
rolfbjarne: Did you confirm that crash reports show up in the html report now? It looks like they're just added into a separate Logs instance, which no other code knows about.

premun: You're right. I added a fix and will verify locally

premun:
@rolfbjarne tested that crash logs propagate back:
image

We can merge this (I think the failed tests are known issues)

tests/xharness/CrashSnapshotReporter.cs
```diff
@@ -0,0 +1,206 @@
...
+		readonly string mlaunchPath;
+		readonly bool isDevice;
+		readonly string deviceName;
+		readonly Func<string> tempFileProvider;
+		readonly string symbolicateCrashPath;
+
+		HashSet<string> initialCrashes;
+
+		public CrashSnapshotReporter (IProcessManager processManager,
+										ILog log,
```
mandel-macaque: Funny alignment.

tests/xharness/Execution/ProcessManager.cs
```diff
@@ -17,7 +17,12 @@ public ProcessManager ()
 		{
 		}
 
-		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename, IList<string> args, ILog log, TimeSpan timeout, Dictionary<string, string> environment_variables = null, CancellationToken? cancellation_token = null)
+		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
+																		IList<string> args,
```
mandel-macaque: Funny alignment.

tests/xharness/Execution/ProcessManager.cs
```diff
@@ -26,6 +31,20 @@ public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
 			}
 		}
 
+		public async Task<ProcessExecutionResult> ExecuteCommandAsync (string filename,
```
mandel-macaque: Same, alignment.

tests/xharness/Xharness.Tests/Tests/CrashReportSnapshotTests.cs
```diff
@@ -0,0 +1,122 @@
...
+			Directory.CreateDirectory (tempXcodeRoot);
+
+			// Create fake symbolicate binary
+			Directory.CreateDirectory (symbolicatePath);
+			File.Create (Path.Combine (symbolicatePath, "symbolicatecrash"));
+		}
+
+		[TearDown]
+		public void TearDown () {
+			Directory.Delete (tempXcodeRoot, true);
```
mandel-macaque: Should we also delete 'symbolicatePath'

premun: It is under the tempXcodeRoot so will be deleted with it

rolfbjarne:
Test failure is unrelated: https://github.com/xamarin/maccore/issues/2177

PR #8137, opened 2020-03-18
[mtouch/mmp] Improve target framework code.

rolfbjarne:
* Unify target framework code between mtouch and mmp.
* Simplify the code in mmp: have three possible valid target frameworks for
  most of code, and add special code to handle setting any other valid target
  frameworks to redirect to one of those three valid target frameworks (and
  warn if given any of those valid, but not "main", target frameworks). Any
  other code can then depend on the target framework having exactly one of
  those specific values, which means we can make IsUnified* variables
  convenience properties instead.
* Unify a bit more of the argument parsing code between mtouch and mmp, since
  that made a few other things easier.
* Add TargetFramework.IsValidFramework to have one validation implementation.
* Move the implementation of TargetFramework.MonoFrameworkDirectory to mmp
  itself, it's not really related to the target framework.
* Remove Driver.IsUnified and IsClassic from mmp, they're not used anymore.
* Formally deprecate --xamarin-[full|system]-framework in mmp, they've really been deprecated for many years.
* Remove LinkerOptions.TargetFramework, it's not used anymore.
* Get rid of mmp's userTargetFramework fried, it's duplicated with the
  targetFramework field.
* Add a few tests, and tweak others a bit.

Breaking changes:

* Both mtouch and mmp require --target-framework now. The only direct
  consumers should be the MSBuild tasks, which already pass --target-framework
  all the time. This simplifies code, and removes assumptions.

### DIFF
--- a/tools/common/Driver.cs
+++ b/tools/common/Driver.cs
@@ -19,6 +19,12 @@ using Xamarin.MacDev;
 using Xamarin.Utils;
 using ObjCRuntime;
 
+#if MTOUCH
+using ProductException = Xamarin.Bundler.MonoTouchException;
+#else
+using ProductException=Xamarin.Bundler.MonoMacException;
+#endif
+
 namespace Xamarin.Bundler {
 	public partial class Driver {
 		public static int Main (string [] args)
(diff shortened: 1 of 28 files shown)


tools/common/Driver.cs
```diff
@@ -235,54 +272,110 @@ public static void Log (int min_verbosity, string format, params object [] args)
...
-			if (Array.IndexOf (TargetFramework.ValidFrameworks, targetFramework.Value) == -1)
-				throw ErrorHelper.CreateError (70, Errors.MT0070, targetFramework.Value, string.Join (" ", TargetFramework.ValidFrameworks.Select ((v) => v.ToString ()).ToArray ()));
+			bool show_0090 = false;
+#if MONOMAC
+			if (!TargetFramework.IsValidFramework (targetFramework)) {
+				// For historic reasons this is messy.
+				// If the TargetFramework we got isn't any of the one we accept, we have to do some fudging.
+				bool force45From40UnifiedSystemFull = false;
+
+				if (references.Any ((v) => Path.GetFileName (v) == "XamMac.dll"))
```
spouliot: there should not be any usage  of `XamMac.dll` *classic/32bits) anymore

tools/common/Driver.cs
```diff
@@ -235,54 +272,110 @@ public static void Log (int min_verbosity, string format, params object [] args)
...
+					// Xamarin.Mac,Version=vX.Y,Profile=Full => Xamarin.Mac,Version=v4.5,Profile=Full
+					TargetFramework = TargetFramework.Xamarin_Mac_4_5_Full;
+				} else if (TargetFramework.Identifier == TargetFramework.Xamarin_Mac_4_5_System.Identifier
+						 && TargetFramework.Profile == TargetFramework.Xamarin_Mac_4_5_System.Profile) {
+					// Xamarin.Mac,Version=vX.Y,Profile=System => Xamarin.Mac,Version=v4.5,Profile=System
+					TargetFramework = TargetFramework.Xamarin_Mac_4_5_System;
+				} else {
+					// This is a total hack. Instead of passing in an argument, we walk the references looking for
+					// the "right" Xamarin.Mac and assume you are doing something
+					// Skip it if xamarin-full-framework or xamarin-system-framework passed in
```
spouliot: did not the commit message mention those `xamarin-[full|system]-framework` were removed ?

tools/common/Driver.cs
rolfbjarne: Exactly, the error we're showing is:

> Projects using the Classic API are not supported anymore. Please migrate the project to the Unified API.

tools/common/Driver.cs
rolfbjarne: They're deprecated, not removed, so they'll still work. But while the comment is still technically correct, it's still outdated, so I'll update it.

tools/common/Driver.cs
spouliot: make sense - localization/using id does not help :| maybe a comment ?

PR #8138, opened 2020-03-18
[Harness] Move Timestamp and TTY utilities to Helpers

premun:
- Removing dependencies on Harness
- Making TTY utilities injectable and thus code using it testable

No functional changes

### DIFF
--- a/tests/xharness/Utilities/Helpers.cs
+++ b/tests/xharness/Utilities/Helpers.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Text;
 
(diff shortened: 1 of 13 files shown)


tests/xharness/Utilities/Helpers.cs
```diff
@@ -25,5 +26,10 @@ public static Guid GenerateStableGuid (string seed = null)
 			}
 			return new Guid (bytes);
 		}
+
+		public static string Timestamp => $"{DateTime.Now:yyyyMMdd_HHmmss}";
+
+		[DllImport ("/usr/lib/libc.dylib")]
+		extern internal static IntPtr Ttyname (int filedes);
```
rolfbjarne:
```suggestion
		extern internal static IntPtr ttyname (int filedes);
```

rolfbjarne: And this would be better exposed with a managed wrapper:

```csharp
[DllImport ("/usr/lib/libc.dylib")]
extern static IntPtr ttyname (int filedes);

internal static string GetTerminalName (int filedescriptor)
{
     return Marshal.PtrToStringAuto (ttyname (filedescriptor));
}
```

mandel-macaque:

PR #8139, opened 2020-03-18
[Msbuild] Do not error if the localizations are out of sync.

mandel-macaque:
The translations code is quite fragile at the moment. When building a
package it is throwing errors but does not do the same when building
xamarin-macios. This has happened several times with the mono-2020-02
integration branch:

* https://jenkins.internalx.com/job/macios/job/PR-7795/24/display/redirect
* https://jenkins.internalx.com/job/macios/job/PR-7795/25/display/redirect

This change makes sure that we do not raise an error, unblocks mono
2020-02 and an issue will be added to make sure that we re-enable the
feature once the code is more stable.

### DIFF
--- a/msbuild/Xamarin.Localization.MSBuild/Xamarin.Localization.MSBuild.csproj
+++ b/msbuild/Xamarin.Localization.MSBuild/Xamarin.Localization.MSBuild.csproj
@@ -5,6 +5,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <UpdateXlfOnBuild>true</UpdateXlfOnBuild>
+    <ErrorOnOutOfDateXlf>false</ErrorOnOutOfDateXlf>
   </PropertyGroup>
   <ItemGroup>
     <EmbeddedResource Update="MSBStrings.resx">


rolfbjarne:
I'd like to point out that the error explicitly says:

> it is strongly discouraged to set UpdateXlfOnBuild=true in official/CI build environments as they should not modify source code during the build.

mandel-macaque:
@rolfbjarne this is something for @wachaudh to look at. My knowledge on that is limited. I'm trying to unblock mono 2020-02, but I do agree with you.

wachaudh:
The `UpdateXlf` target generates a set of  *.lastrun files. The .lastrun files are used by the xliff-tasks nuget package to check if the translation files (the ones with the *.xlf extension) are out of date with the *.resx file. The .lastrun files are wiped out on a git clean so the nuget package complains that translations are out of date even when they aren't. I had the `UpdateXlfOnBuild` project property (provided by xliff-tasks) set to true so that it calls the `UpdateXlf` target on every build to generate the .lastrun files. However, there is a bug with the xliff-tasks nugget package and it seems to ignore `UpdateXlfOnBuild` in projects with multiple target frameworks. To workaround the issue I hooked into the `BeforeBuild` msbuild target to force the `UpdateXlf` target to run on every build.

.lastrun file paths: 
./obj/Debug/netstandard2.0/Xamarin.Localization.MSBuild.xlf/UpdateXlf.lastrun
./obj/Debug/net461/Xamarin.Localization.MSBuild.xlf/UpdateXlf.lastrun

mandel-macaque:
Known issues: https://github.com/xamarin/maccore/issues/1956 and https://github.com/xamarin/maccore/issues/1488

PR #8143, opened 2020-03-18
[linker] Update the list of opcodes we can't compute a constant value for.

rolfbjarne:
Teach the linker about more opcodes it can't compute a constant value for.
This way a debug version of mmp/mtouch doesn't print out these opcodes for
someone to check.

This code is only executed when debugging mtouch/mmp in the IDE, so no change
to released bits.


rolfbjarne:
Test failures are unrelated, all due to https://github.com/xamarin/maccore/issues/1956.

PR #8149, opened 2020-03-18
[Harness] Create a more general TestAssemblyDefinition.

mandel-macaque:
XHarness TestAssemblyDefinition is very particular on the way we deal
with the mono test assemblies, in the general case, it only contains the
assembly name, but in the xamarin-macios it calculates the name of the
assembly using the platform.

Create an interface so that the command line was its own implementaiton
and move as much as possible to use the AssemblyLocator class. Later we
can add two simpler implementations for the more general cases.

Move the AssemblyLocator out of the Templates namespace since it makes
more sense.


tests/xharness/BCLTestImporter/Xamarin/AssemblyLocator.cs
```diff
@@ -26,5 +47,35 @@ public string GetAssembliesRootLocation (Platform platform)
...
+			case Platform.iOS:
+			case Platform.TvOS:
+			case Platform.WatchOS:
+				return Path.Combine (downloadsPath, "ios-bcl", platformPathMatch [platform], "tests");
+			}
+
+			return null;
+		}
+
+		public string GetHintPathForRefenreceAssembly (string assembly, Platform platform)
```
premun: >GetHintPathForRefenreceAssembly [](start = 16, length = 31)

GetHintPathForReferenceAssembly

tests/xharness/BCLTestImporter/BCLTestProjectDefinition.cs
```diff
@@ -10,11 +10,12 @@ namespace Xharness.BCLTestImporter {
 	/// Class that defines a bcl test project. A bcl test project by definition is the combination of the name
 	/// of the project and a set on assemblies to be tested.
 	/// </summary>
-	public partial struct BCLTestProjectDefinition {
+	public partial class BCLTestProjectDefinition {
```
premun: Does it need to be partial? Is there somewhere another piece?

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/1956

PR #8152, opened 2020-03-19
[Harness] Increase wait time for crash logs

premun:
Manual testing showed that crash logs don't appear fast enough


mandel-macaque:
Test failure is known and under investigation.

PR #8153, opened 2020-03-19
[d16-6-xcode11.4] merge d16-5-xcode11.4 into d16-6-xcode11.4

dalexsoto:


Make.versions
```diff
@@ -43,5 +43,5 @@
 #   line changed in git).
 #
 
-IOS_PACKAGE_VERSION=13.18.0.$(IOS_COMMIT_DISTANCE)
-MAC_PACKAGE_VERSION=6.18.0.$(MAC_COMMIT_DISTANCE)
+IOS_PACKAGE_VERSION=13.18.1.$(IOS_COMMIT_DISTANCE)
```
dalexsoto: @spouliot should we go `13.18.1` or just bump to `13.20.0` and bump master to `.21`?

tools/mtouch/Makefile
```diff
@@ -158,6 +158,8 @@ SIMLAUNCHER_FRAMEWORKS =  \
 	-weak_framework QuickLookThumbnailing		\
 	-weak_framework SoundAnalysis			\
 	-weak_framework VisionKit			\
+						\
```
whitneyschmidt: is the extra line here needed?

dalexsoto: We separate frameworks added by different Xcode releases so not really needed but follows the pattern of the file

Make.versions
spouliot: hard to be sure before _all the things_ gets released
Since d16-6 is still in preview then it makes sense to bump the 3rd digit (as you did)
It won't cause issues for release notes (like the d16-5 backport)

Make.config
```diff
@@ -355,7 +354,8 @@ JENKINS_RESULTS_DIRECTORY ?= $(abspath $(TOP)/jenkins-results)
 # Clone files instead of copying them on APFS file systems. Much faster.
 CP:=$(shell df -t apfs / >/dev/null 2>&1 && echo "cp -c" || echo "cp")
 
-XCODE_ARCHIVE_VERSION=xcode-$(XCODE_PRODUCT_BUILD_VERSION)
+# WORKAROUND, should be  xcode-$(XCODE_PRODUCT_BUILD_VERSION) but we are stuck due to xcode 11.4 beta 1 issue: https://github.com/xamarin/xamarin-macios/issues/7872
+XCODE_ARCHIVE_VERSION=xcode-11C29
```
mandel-macaque: Do we need this still?? @akoeplinger maybe you do know?

mandel-macaque: Got confirmation from @akoeplinger online, looks good.

akoeplinger: Yes, still needed.

PR #8154, opened 2020-03-19
[Harness] Create DeviceLogCapturer factory and inject it

premun:
Another step to make `AppRunner.RunAsync` testable


mandel-macaque:
Test failure is a known issue.

PR #8124, opened 2020-03-16
🚧 Diagnosis in progress 🚧 DO NOT MERGE. 🚧

rolfbjarne:
Trying to figure out what's happening with https://github.com/xamarin/maccore/issues/2177.

PR #8156, opened 2020-03-19
[Harness] Generalize XmlResultParser.

mandel-macaque:
First step to try and clean up the parsing of results and ensure that we
are doing the correct thing. We have several issues there. We want to
move all the complicated logic out of AppRunner and test the different
outputs and the full result matrix.

This commit first allows to use an instance class for the parsing, later
we will move things out. Step by step


mandel-macaque:
Build issue is known and under investigation: https://github.com/xamarin/xamarin-macios/pull/8124

tests/xharness/Harness.cs
```diff
@@ -601,6 +603,7 @@ int Install ()
 					new DeviceLoaderFactory (this, ProcessManager),
 					new CaptureLogFactory (),
 					new DeviceLogCapturerFactory (ProcessManager, XcodeRoot, MlaunchPath),
+					new XmlResultParser (),
```
premun: Not sure we shouldn't use the `resultParser` field here?

Issue #8147, opened 2020-03-18
[tvos] MT0000 running install_name_tool


spouliot:
> 76 tests failed, 74 tests passed.

Many failures, yet some worked :|

Works locally :(

```
01:09:22.1725810 MTOUCH : error MT0000: Unexpected error - Please file a bug report at https://github.com/xamarin/xamarin-macios/issues/new [/Users/xamarinqa/agent/_work/11/s/tests/xharness/tmp-test-dir/monotouch-test2411/monotouch-test-tvos.csproj]
01:09:22.1729550   System.ComponentModel.Win32Exception (0x80004005): ApplicationName='xcrun: error: unable to read SDK settings for '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' (TaskId:415)
01:09:22.1731870   /Applications/Xcode113.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool (TaskId:415)
01:09:22.1735680 xcrun : error : unable to read SDK settings for '/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'', CommandLine='-change /System/Library/PrivateFrameworks/IOSurface.framework/IOSurface /System/Library/Frameworks/IOSurface.framework/IOSurface /Users/xamarinqa/agent/_work/11/s/tests/xharness/tmp-test-dir/monotouch-test2411/obj/iPhone/Debug-tvos/mtouch-cache/arm64/monotouchtest', CurrentDirectory='', Native error= Cannot find the specified file [/Users/xamarinqa/agent/_work/11/s/tests/xharness/tmp-test-dir/monotouch-test2411/monotouch-test-tvos.csproj]
01:09:22.1736760     at System.Diagnostics.Process.StartWithCreateProcess (System.Diagnostics.ProcessStartInfo startInfo) [0x0029f] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-12/external/bockbuild/builds/mono-x64/mcs/class/System/System.Diagnostics/Process.cs:778  (TaskId:415)
01:09:22.1738010     at System.Diagnostics.Process.Start () [0x0003a] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-12/external/bockbuild/builds/mono-x64/mcs/class/referencesource/System/services/monitoring/system/diagnosticts/Process.cs:2006  (TaskId:415)
[...]

PR #8166, opened 2020-03-20
[mtouch/mmp] Add better diagnostics if xcrun returns invalid data.

rolfbjarne:
Ref: #8147.


rolfbjarne:
Test failures are both unrelated and due to https://github.com/xamarin/maccore/issues/1956.

PR #8168, opened 2020-03-20
[tests][monotouch-test] Update UrlProtocolTest

spouliot:
The _nice_ thing about `NSURLProtocol` is that it does not force the use
of the network - but we inflicted this upon us on tests.

So this updates the test to avoid the network, which should make it
more reliable and not so much influenced by the URL used or the OS
versions where it's executed.

This does not cover as much callback as before. However this it does
test that the NSUrlProtocol is registered correctly (what we care) and
the rest is largely only related the the OS (which is not the target
of our unit tests).

Hopefully fix https://github.com/xamarin/maccore/issues/1956


spouliot:
@dalexsoto hopefully makes your life easier for the next sprint :)

dalexsoto:
@spouliot thanks, should we backport this one to our xcode11.4 branches?

dalexsoto:
@monojenkins backport d16-6-xcode11.4

dalexsoto:
@monojenkins backport d16-5-xcode11.4

PR #8199, opened 2020-03-25
[d16-6-xcode11.4] Merge d16-5-xcode11.4 into d16-6-xcode11.4

dalexsoto:


spouliot:
maintaining the branch will only lead to additional, yet avoidable, mistakes
is there any reason why we're not merging this into `d16-6` ?

dalexsoto:
After an awesome discussion with @spouliot this is getting merged and then `d16-6-xcode11.4` is getting merged into `d16-6`

Issue #8160, opened 2020-03-20
[xharness] Device Listing One or more errors occurred. (Must specify valid information for parsing in the string.))


spouliot:
### Steps to Reproduce

1.  `make runner`
2. See that devices are not found (top)
3. Devices are visible/works inside VS4M

### Expected Behavior

xharness should see all devices
or minimally parse the ones it can

### Actual Behavior

```
Device Listing One or more errors occurred. (Must specify valid information for parsing in the string.))
```

### Environment

master

### Build Logs

clicking on the message (link) shows:

```
09:11:27.8558370 Launching /Users/poupou/git/master/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/bin/mlaunch 
09:11:27.8575670 /Users/poupou/git/master/xamarin-macios/_ios-build/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/bin/mlaunch --sdkroot=/Applications/Xcode113.app --listdev=/var/folders/zw/d0tk01n90mq3w4rm5h2_n9bc0000gn/T/tmp499717a3.tmp --output-format=XML
09:11:50.8194860 Result:
09:11:50.8208900 <?xml version="1.0" encoding="utf-8"?>
<MTouch Version="0" Revision="0">
  <SdkRoot>/Applications/Xcode113.app/Contents/Developer</SdkRoot>
  <Device>
    <ActivationState />
    <AmountDataAvailable />
    <BatteryCurrentCapacity />
    <BluetoothAddress />
    <BuildVersion />
    <ChipID />
    <CPUArchitecture />
    <DeveloperStatus />
    <DeviceClass />
    <DeviceColor />
    <DeviceEnclosureColor />
    <DeviceIdentifier>ebfd5a244e8ee0aec4169846487a9b4edc3c574e</DeviceIdentifier>
    <DevToolsAvailable />
    <EnableWifiDebugging />
    <HardwareModel />
    <InterfaceType>Wifi</InterfaceType>
    <InternationalMobileEquipmentIdentity />
    <IsPaired>False</IsPaired>
    <IsUsableForDebugging />
    <IsValid>True</IsValid>
    <ModelNumber />
    <Name>Office</Name>
    <PasswordProtected />
    <ProductionSOC />
    <ProductType />
    <ProductVersion />
    <SerialNumber />
    <TotalDataCapacity />
    <UDID />
    <UniqueChipID />
    <WatchCompanionCapability />
[...]

PR #8201, opened 2020-03-25
[Harness] If empty nodes are found, do not crash.

mandel-macaque:
Looks like we have cases in which we get empty nodes when listing
devices, which makes the parsing of the nodes fail and crash.

Move the code to a diff method, try to parse the data, if there is an
exception (missing data) return null. If null is returned, do not add
it.

Fixes: https://github.com/xamarin/xamarin-macios/issues/8160


mandel-macaque:
Why we get empty nodes from apple is unclear, might be a bug in their side :/

tests/xharness/Xharness.Tests/Hardware/Tests/DevicesTest.cs
```diff
@@ -72,7 +72,7 @@ public void LoadAsyncProcessErrorTest (bool timeout)
 
 			MlaunchArgument sdkRootArg = passedArguments.Where (a => a is SdkRootArgument).FirstOrDefault();
 			Assert.IsNotNull (sdkRootArg, "sdk arg missing");
-			AssertArgumentValue (sdkRootArg, sdkPath, "sdk arg wrong");
+			AssertArgumentValue (sdkRootArg, $"--sdkroot {sdkPath}", "sdk arg wrong");
```
spouliot: is this related to the change ?

mandel-macaque: No, but the tests failed, there was a commit that should have updated this.

tests/xharness/Hardware/Devices.cs
```diff
@@ -62,6 +62,31 @@ Task ILoadAsync.LoadAsync (ILog log, bool include_locked, bool force)
...
+					BuildVersion = deviceNone.SelectSingleNode ("BuildVersion")?.InnerText,
+					ProductVersion = deviceNone.SelectSingleNode ("ProductVersion")?.InnerText,
+					ProductType = deviceNone.SelectSingleNode ("ProductType")?.InnerText,
+					InterfaceType = deviceNone.SelectSingleNode ("InterfaceType")?.InnerText,
+					IsUsableForDebugging = usable == null ? (bool?) null : ((bool?) (usable == "True")),
+				};
+				bool.TryParse (deviceNone.SelectSingleNode ("IsLocked")?.InnerText, out var locked);
+				device.IsLocked = locked;
+				return device;
+			} catch {
```
rolfbjarne: A solution that doesn't involve throwing and catching exceptions is much better. One example is that exceptions tracing can be used to figure out real problems.

tests/xharness/Xharness.Tests/Hardware/Tests/DevicesTest.cs
spouliot: That should be a separate PR or, minimally, mentioning in the commit message.
Also it should link to the original commit and check why this was merged if not working.

mandel-macaque: @spouliot issue with these tests, they are not in our CI, xharness is not running its own tests, might be a good idea to add that.

tests/xharness/Hardware/Devices.cs
mandel-macaque: Fixed, the TryParse should work correctly, if the DeviceClass cannot be parsed, return null.

tests/xharness/Xharness.Tests/Hardware/Tests/DevicesTest.cs
spouliot: Tests that don't run don't exists :(
Let's move that unrelated fix into a PR that enable the tests :)

premun: I opened a PR to fix these and couple of others that are failing: https://github.com/xamarin/xamarin-macios/pull/8208

mandel-macaque: @premun thx!!!! I'll revert that change then!

PR #8203, opened 2020-03-26
[Harness] Remove Harness as a dependency

premun:
- Move Xcode and mlaunch paths to `ProcessManager` so it is not passed around as well and then sent back to `ProcessManager`
- Shrink `IHarness` API to what is needed so it's easier to remove it later
- Remove `IHarness` from as many classes as possible where it was necessary only for the mlaunch/xcode paths


tests/xharness/Execution/ProcessManager.cs
```diff
@@ -7,84 +7,100 @@
...
-			CancellationToken? cancellation_token = null)
+			Dictionary<string, string> environmentVariables = null,
+			CancellationToken? cancellationToken = null)
 		{
-			using (var p = new Process ()) {
-				p.StartInfo.FileName = filename;
-				p.StartInfo.Arguments = args.AsCommandLine ();
-				return await RunAsync (p, log, timeout, environment_variables, cancellation_token);
-			}
+			return await RunAsync (new Process(), args, log, timeout, environmentVariables, cancellationToken);
```
mandel-macaque: We are leaking the Process here! Or at least, it will not be disposed. Either re-add the using block, or move too the new using pattern.

premun: Good catch

rolfbjarne:
Pending test run on internal jenkins.

PR #8207, opened 2020-03-26
[xcode11.4] Merge d16-5-xcode11.4 into master

dalexsoto:


Make.config
```diff
@@ -92,24 +92,23 @@ MAX_SHARPIE_VERSION=3.5.99
 MIN_SHARPIE_URL=https://download.visualstudio.microsoft.com/download/pr/693cb4a2-5455-4841-904f-a2936d2781f4/3890cb5f74ea75ff4474152f1f2b8008/objectivesharpie-3.5.22.pkg
 
 # Minimum OSX versions for building XI/XM
-MIN_OSX_BUILD_VERSION=10.14
+MIN_OSX_BUILD_VERSION=10.15
```
mandel-macaque: I  think we should aim for 10.15.3 since Xcode 11.4 requires that min version.

dalexsoto: I do not think we have a choice since the SDK identifier is just 10.15

image

dalexsoto: or maybe we use `OSX_SDK_VERSION` for that purpose so you could be right :P 

@rolfbjarne / @spouliot thoughts?

spouliot: While no guarantee, this has not been an issue in the past
Most Xcode version has a 3 digits version requirement for macOS

rolfbjarne: This is not the SDK version, that's specified below with `OSX_SDK_VERSION`. `MIN_OSX_BUILD_VERSION` checks the current OS version during the build, so bumping it to 10.15.3 is perfectly fine.

Make.config
```diff
@@ -92,24 +92,23 @@ MAX_SHARPIE_VERSION=3.5.99
...
 # Minimum OSX version for executing XI/XM tooling.
 MIN_OSX_VERSION_FOR_IOS=10.11
 MIN_OSX_VERSION_FOR_MAC=10.11
 
-# WARNING: Do **not** use versions higher than the available Xcode SDK or else we will have issues with mtouch (See https://github.com/xamarin/xamarin-macios/issues/7705)
```
rolfbjarne: Seems like this warning should stay in there.

Make.config
dalexsoto: Thanks Rolf, Release notes says 10.15.2 so I’ll use that https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes

```suggestion
MIN_OSX_BUILD_VERSION=10.15.2
```

Make.config
```diff
@@ -92,24 +92,23 @@ MAX_SHARPIE_VERSION=3.5.99
...
-IOS_SDK_VERSION=13.2
+IOS_SDK_VERSION=13.4
```
dalexsoto:
```suggestion
# WARNING: Do **not** use versions higher than the available Xcode SDK or else we will have issues with mtouch (See https://github.com/xamarin/xamarin-macios/issues/7705)
IOS_SDK_VERSION=13.4
```

dalexsoto:
@rolfbjarne addressed feedback, thanks!

PR #8209, opened 2020-03-27
[siminstaller] Update parsing xml because Apple changed their schema.

rolfbjarne:
Apparently file sizes can be real numbers now, because in the manifest of
downloadable simulators there's one entry where the fileSize is a real:

    <key>fileSize</key>
    <real>2049324382</real>

when every other file size in the manifest is an integer.

So handle real file sizes too, and fix file size parsing to not throw
exceptions if the number can't be parsed. Maybe we'll get imaginary file sizes
too one day...

Fixes this exception:

    Unhandled Exception:
    System.NullReferenceException: Object reference not set to an instance of an object
      at xsiminstaller.MainClass.Main (System.String[] args) [0x00617] in /Users/rolf/work/maccore/d16-6/xamarin-macios/tools/siminstaller/Program.cs:194


rolfbjarne:
@monojenkins backport master

PR #8212, opened 2020-03-27
[Harness] Remove AppRunner references from TestReporter

premun:
Partial revert of #8183 where introduced references to `AppRunner` to a the `TestReporter` class that will be moved out to a shared NuGet library and thus cannot reference `AppRunner`.


tests/xharness/AppRunner.cs
```diff
@@ -437,6 +448,15 @@ public async Task<int> RunAsync ()
...
 			// check the final status, copy all the required data
 			(Result, FailureMessage) = await testReporter.ParseResult ();
+			
+			var crashLogWaitTime = 0;
+			if (!testReporter.Success.Value)
+				crashLogWaitTime = 5;
+			if (Result == TestExecutingResult.Crashed)
+				crashLogWaitTime = 30;
+
+			await crashReporter.EndCaptureAsync (TimeSpan.FromSeconds (crashLogWaitTime));
```
mandel-macaque: I'm worried about this change. EndCaptureAsync, is it collecting the crash logs? f that is case, we probably want to do it before the test reporter  parses the result so that the crash logs are correctly added, right?

premun: Reverted

PR #8218, opened 2020-03-28
[Harness] Remove not needed classes from the BCLTestImporter.

mandel-macaque:
Remove two classes that are not really needed and a method. We moved to
named tuples to make things simpler. Unfortunally C# does not support:

```csharp
using MyNamedTuple = (string Name, double value, bool answer);
```

But we have to live with this. Makes the namespace simpler and removes
confusion with the already present project classes that are more widely
used.


mandel-macaque:
@premun named tuple made it nice to remove the class, I did not like the overlap of names, with it out of the way, and the info too, we are in a good shape. If we notice that the named tuple needs to be moved to a POCO, it should not be too much work.

Issue #7191, opened 2019-10-08
Xamarin.IOS version mismatch Latest VS Mac (13.2.0.47) & Latest VS 2019 Windows (13.2.0.42)


mbullock976:
Hi

My Visual Studio for Mac was updated to now use Xamarin.IOS 13.2.0.47, however the latest Visual Studio 2019 for Windows 16.3.1 which was released today only supports 13.2.0.42. 

This means I cannot upgrade windows to .47 and don't know how or if I should downgrade Mac to .42

If I try and pair to Mac from windows VS 2019, I get a versions mismatch which tries to install .42 over the .47 on my Mac but it fails to install without a reason.

Please help

Thank you

mbullock976:
Hi

Ratherthan waiting to Microsoft to release a new update to VS 2019 for windows that supports Xamarin.ios 13.2.0.47

I instead downgrade my Mac from 13.2.0.47 to 13.2.0.42 by downloading the pkg file on my Mac and installing it. Now my VS Mac is showing Xamarin.IOS 13.2.0.42 and I'm able to pair to mimic again from windows

bnserig:
Why VS 2019 Windows version is not released at the same time ?

mbullock976:
I have no idea why. Microsoft released new update for VS 2019 today v16.3.1 but this does not support the latest xamarin iOS release

I could wait for Microsoft to release the compatible version but instead chose to downgrade my Mac

I’ve switched off auto check updates in my visual studio for Mac now

dalexsoto:
Hello @bnserig and @mbullock976 👋 

Sorry for the trouble, VS Windows should now have the same version as VSMac.

biapar:
The problem happened again.

CMD73:
You can download the exact matching version on the Mac and install manually, as per this older post: https://stackoverflow.com/questions/41991122/xamarin-studio-for-mac-standalone-xamarin-ios-installer

https://dl.xamarin.com/MonoTouch/Mac/xamarin.ios-13.10.0.17.pkg

This seemed to work for me, and I was able to pair the Mac in VS2019.

PR #8219, opened 2020-03-28
[Mono] Bump mono to 2020-02 becc5fb to match android

mandel-macaque:
Commits:

* mono/mono@beccf5ba438: [2020-02] Use GArray instead of GList when getting custom attributes from an image (#19102)
* mono/mono@5bea89d861f: [llvm] Disable running the llvm verifier by default, it was enabled by mistake, and it takes a long time. (#7191  * mono/mono@4b0424cd688: [2020- 02][debugger] Implementing step through multithreaded code. (#19343)
* mono/mono@2444b38b3c5: [2020-02] [dim] Class overriding interface method that was already override by another interface (#19121)
* mono/mono@93a2f4ea395: Fix suspend_policy that will be sent to debugger-libs, because on debugger-libs we want to resume the VM only when it's suspended, but we were sending the wrong suspend_policy in the case of EVENT_KIND_VM_START and mono is started with suspend=y. (#19330)
* mono/mono@0565f6e2c0e: Do not access unloading domains in debugger (case 1013579) (#19322)
* mono/mono@b7196c45e86: [jit] Initialize correct class for tls fields (#19300) (#19313)
* mono/mono@88c1989895f: [2020-02] [merp] Add breadcrumb for StackHash (#19242)
* mono/mono@d2be7389b11: [2020-02] [corlib] Suppress finalization of underlying console FileStreams (#19164)

Full diff: https://github.com/mono/mono/compare/a7e29665d246d359a778bf4d692b4cb5fd2f3548...beccf5ba4386148534467bff03fabe92da6d9705


mandel-macaque:
Related android branch https://github.com/xamarin/xamarin-android/pull/4479

PR #8227, opened 2020-03-29
[Harness] Add a new known failure to help the monitoring person.

mandel-macaque:
Looks like mono is not investing much time at the moment fixing
https://github.com/mono/mono/issues/18560

The above error happens always on 32b device tests. Added a new known
failure for the monitoring to use in their task so that they do not have
to goole the issue and they know that is knonw already.

### DIFF
--- a/tests/xharness/Jenkins/Jenkins.cs
+++ b/tests/xharness/Jenkins/Jenkins.cs
@@ -1795,6 +1795,22 @@ namespace Xharness.Jenkins
 			}
 			return false;
 		}
+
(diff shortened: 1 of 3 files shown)


mandel-macaque:
After the PR the monitoring person will get the following:

Screenshot 2020-03-29 at 14 44 53

spouliot:
@mandel-macaque please run the PR against internal bots before merging

> The above error happens always on 32b device tests. 

^ the tests that were executed does not run that config, so this code has not been tested,

PR #8233, opened 2020-03-30
[Harness] Move code to a shared library

premun:
- Code that will later be moved to the `dotnet/xharness` repo is first moved to an isolated project before it will be extracted to the new repo and NuGetified
- New project for the shared code and new test project for tests accompanying the moved code are created
- Only Factories are left behind that are used only by XHarness


premun:
Will clean the branch first

PR #8237, opened 2020-03-30
[generator] Remove unreachable code

spouliot:
An earlier `wrap != null` condition ends with a `return` so any check or
code assuming `wrap == null` must always be executed, while any `wrap != null`
will never be.


spouliot:
Easier to review when hiding whitespaces
https://github.com/xamarin/xamarin-macios/pull/8237/files?diff=unified&w=1

PR #8241, opened 2020-03-30
[generator] Remove unneeded null checks in event args

spouliot:
The calls to `Runtime.GetNSObject<T>` and `NSArray.ArrayFromHandle<T>`
already do the null / `IntPtr.Zero` check.

This generates a bit less code (always a good thing) and reduce the
number of nullability warnings.


spouliot:
failure is a known random issue -> https://github.com/xamarin/maccore/issues/2160

PR #8253, opened 2020-03-31
[devops] Fix provision-xcode unhandled exception

VincentDondain:
In cases where a `Make.config` doesn't exist, there's an unhandled exception and the user friendly one cannot be reached.

Arguably, an edge case not applicable to xamarin-macios I discovered when using that code in an other context (where obviously I don't have a Make.config :P). Still this is making the code more robust (;


VincentDondain:
Known HE0038: https://github.com/xamarin/maccore/issues/581

Issue #8271, opened 2020-04-02
[Harness] Harness throws an exception when there are no simulators or runtasks 


mandel-macaque:
When executing xharness we noticed that the following line throws and exception:

https://github.com/xamarin/xamarin-macios/blob/master/tests/xharness/Jenkins/TestTasks/AggregatedRunSimulatorTask.cs#L66

In the case when there are no tasks we get exceptions like the following: http://xamarin-storage/jenkins/xamarin-macios/PR-8269/7bbc0c974ca7201576c261cdc60a9351293533ae/1/jenkins-results/tests/index.html

That should not happen and we should ignore the test to the very least.

rolfbjarne:
https://github.com/xamarin/xamarin-macios/blob/5aa2b1c34fdd544ae42a20f28ee3b51b9703abde/tests/xharness/Jenkins/TestTasks/TestTask.cs#L35-L40

PR #8272, opened 2020-04-02
[Harness] Ensure we do not throw an exception if we do not have tasks.

mandel-macaque:
The First extension will throw an exception if the list is empty. Use
FirstOrDefault, check if the devices are null and return the test as
Ignored.

fixes: https://github.com/xamarin/xamarin-macios/issues/8271

### DIFF
--- a/tests/xharness/Jenkins/TestTasks/AggregatedRunSimulatorTask.cs
+++ b/tests/xharness/Jenkins/TestTasks/AggregatedRunSimulatorTask.cs
@@ -65,7 +65,7 @@ namespace Xharness.Jenkins.TestTasks {
 
 				var devices = executingTasks.FirstOrDefault ()?.Simulators; 
 				if (devices == null) { 
-					ExecutionResult = TestExecutingResult.Ignored;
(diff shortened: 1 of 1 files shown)


tests/xharness/Jenkins/TestTasks/AggregatedRunSimulatorTask.cs
```diff
@@ -63,7 +63,11 @@ protected override async Task ExecuteAsync ()
 					await task.SelectSimulatorAsync ();
 				}
 
-				var devices = executingTasks.First ().Simulators;
+				var devices = executingTasks.FirstOrDefault ()?.Simulators; 
+				if (devices == null) { 
+					ExecutionResult = TestExecutingResult.Ignored;
```
rolfbjarne, line 68: No simulator means the test should fail with TestExecutingResult.DeviceNotFound.

mandel-macaque: @rolfbjarne fixed.

Issue #8146, opened 2020-03-18
[xharness] incorrect attachments in azdo ?


spouliot:
https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3559942&view=ms.vss-test-web.build-test-results-tab&runId=11890976&resultId=749457&paneView=attachments shows App Build [xUnit] Mono SystemCoreXunit Part 1 failed twice

1st time I don’t get attachments
2nd time I get attachments that suggest execution

so something is really off - either the reported result or the attachment

### Expected Behavior

no execution attachment on something that reported as not built

### Actual Behavior

* App Build [xUnit] Mono SystemCoreXunit Part 1
* with attachment logs showing execution

### Environment

master

spouliot:
There can be huge difference in timestamps between the attached logs
https://github.com/xamarin/xamarin-macios/commit/ec420e674004c187f4beb7119aab45b215c7a765

![Screen Shot 2020-03-20 at 2 31 36 PM](https://user-images.githubusercontent.com/260465/77195163-946dba80-6ab7-11ea-85c3-c1a2e8642fcf.png)

mandel-macaque:
Taking a look at this one atm.

PR #8273, opened 2020-04-02
[Harness] Make sure TestsTasks do not share an id.

mandel-macaque:
Threading is hard, part 2. We could have a situation in which two tasks
that are instatiated async use the same id. The id is later used to
build the logs directory etc.. When they share the id we end up in a
situation in which the logs overlap.

fixes: https://github.com/xamarin/xamarin-macios/issues/8146

### DIFF
--- a/tests/xharness/Jenkins/TestTasks/TestTask.cs
+++ b/tests/xharness/Jenkins/TestTasks/TestTask.cs
@@ -12,8 +12,9 @@ using Microsoft.DotNet.XHarness.iOS.Shared.Utilities;
 namespace Xharness.Jenkins.TestTasks {
 	public abstract class TestTask
 	{
+		static object counterLock = new object ();
 		static int counter;
-		public readonly int ID = counter++;
+		public readonly int ID;
 
 		bool? supports_parallel_execution;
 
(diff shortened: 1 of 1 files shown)


tests/xharness/Jenkins/TestTasks/TestTask.cs
```diff
@@ -32,6 +33,16 @@ public abstract class TestTask
 		public bool BuildOnly;
 		public string KnownFailure;
 
+		public TestTask ()
+		{
+			// no, Interlocked.Increment is not going to do the samething, it is a two step, set and increment,
```
rolfbjarne, line 38: I don't understand why using Interlocked.Increment wouldn't work:

```csharp
ID = Interlocked.Increment (ref counter);
```

mandel-macaque: I could have read the lock wrong. But I don't like locks in constructors and are a bad practice. I move to Increment and if we see it happening again we know where is the issue.

tests/xharness/Jenkins/TestTasks/TestTask.cs
```diff
@@ -32,6 +33,16 @@ public abstract class TestTask
 		public bool BuildOnly;
 		public string KnownFailure;
 
+		public TestTask ()
+		{
+			// no, Interlocked.Increment is not going to do the samething, it is a two step, set and increment, 
+			// while Interlocked.Increment is just atomic for the increment. You might think that Interlocked.Ingrement
```
dalexsoto, line 39: Don't spam a new build for this but typo! 😁 

Definitely `Interlocked.Ingrement` would not work but `Increment` may.

mandel-macaque: I'm going to spam builds because I'm going to  implement what @rolfbjarne suggested and see what happens ;)

tests/xharness/Jenkins/TestTasks/TestTask.cs
mandel-macaque: @rolfbjarne fixed.

PR #8275, opened 2020-04-03
[d16-5] Update azure-pipelines.yml

dalexsoto:
We need to use the Catalina pool now that Xcode 11.4 is stable


dalexsoto:
Known failure: HE0038

PR #8278, opened 2020-04-03
[Harness] Allow to create a process manager with default tool paths

premun:
- `XcodeRoot` will be detected using `xcode-select` and `mlaunch` will have a hardcoded default

### DIFF
--- a/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Execution/ProcessManager.cs
+++ b/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Execution/ProcessManager.cs
@@ -373,8 +373,8 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared.Execution {
 			process.StartInfo.Arguments = "-p";
 
 			var log = new MemoryLog ();
-			var stdout = new MemoryLog ();
-			var stderr = new MemoryLog ();
+			var stdout = new MemoryLog () { Timestamp = false };
+			var stderr = new ConsoleLog ();
 			var timeout = TimeSpan.FromSeconds (30);
 
 			var result = RunAsyncInternal (process, log, stdout, stderr, timeout).GetAwaiter ().GetResult ();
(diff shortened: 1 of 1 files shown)


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Execution/ProcessManager.cs
```diff
@@ -345,5 +365,31 @@ static List<int> GetChildrenPS (ILog log, int pid)
...
+				throw new Exception ("Failed to detect Xcode path from xcode-select!");
+
+			// Something like /Applications/Xcode114.app/Contents/Developers
+			var xcodeRoot = stdout.ToString ().Trim ();
+
+			if (string.IsNullOrEmpty (xcodeRoot))
+				throw new Exception ("Failed to detect Xcode path from xcode-select!");
+
+			// We need /Applications/Xcode114.app only
+			return Path.GetDirectoryName(Path.GetDirectoryName(xcodeRoot));
```
spouliot, line 392: minor: missing spaces before the `(`

premun: True but this code will be moved and reformatted next week so I would let it go in this time

tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Execution/ProcessManager.cs
```diff
@@ -345,5 +365,31 @@ static List<int> GetChildrenPS (ILog log, int pid)
...
+
+			var log = new MemoryLog ();
+			var stdout = new MemoryLog () { Timestamp = false };
+			var stderr = new ConsoleLog ();
+			var timeout = TimeSpan.FromSeconds (30);
+
+			var result = RunAsyncInternal (process, log, stdout, stderr, timeout).GetAwaiter ().GetResult ();
+
+			if (!result.Succeeded)
+				throw new Exception ("Failed to detect Xcode path from xcode-select!");
```
stephen-hawley, line 383: You have two different failure conditions that throw identical exceptions. Future you will be happier if you can distinguish the two cases.

premun: Good point. Too bad it got merged already. However if this fails, we will have details in stderr (which is a ConsoleLog) so I guess we will be able to debug

PR #8296, opened 2020-04-06
[Foundation] As a default, do not use the workaround in NSUrlSessionHandler for the threadpool.

mandel-macaque:
When an application was moved the the background, the thread pool from
mono would be left in an unknonw state, making applications to stale
(https://xamarin.github.io/bugzilla-archives/58/58633/bug.html#c7).

This work was added in https://github.com/xamarin/xamarin-macios/pull/5463

We now set the default behaviour to skip the workaround to see if the
new provided mono works as expected. We do not fully remove the
workaround because we need some real world testing.

If the new ThreadPool from mono does not work as expected we do provide
a property to re-add the workaround. The BypassBackgroundSessionCheck
can be set to false to allow users get it back.

The following is an example usage of the API:

```csharp
// create your own handler instance rather than using the one provided by default.
var handler = new NSUrlSessionHandler() {
  BypassBackgroundSessionCheck = false, // readd the hack
};
var httpClient = new HttpClient (handler); // use the handler with the hack
```

This is a partial fix for https://github.com/xamarin/xamarin-macios/issues/7080

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -334,7 +334,7 @@ namespace Foundation {
 		// in certain cases the user does that on purpose (BeingBackgroundTask) and wants to be able
 		// to use the network. In those cases, which are few, we want the developer to explicitly 
 		// bypass the check when there are not request in flight 
-		bool bypassBackgroundCheck;
+		bool bypassBackgroundCheck = true;
 
 		public bool BypassBackgroundSessionCheck {
 			get {


mandel-macaque:
Adding do not merge, we have to wait for the d16-7 branch to be created.

mandel-macaque:
@whitneyschmidt yes, we have Remove workaround for ThreadPool not working #7080 for that, this PR is not closing it :)

mandel-macaque:
d16-7 branched, landing.

PR #8312, opened 2020-04-08
[xharness] Don't overwrite variations.

rolfbjarne:
This means different variations of tests won't show up as identical in the
html report.

Before:

Before

After:

After


rolfbjarne:
Test failure is unrelated

* monotouch-test/iOS 64-bits: https://github.com/xamarin/maccore/issues/1265

PR #8320, opened 2020-04-08
Create a general rule to create makefile fragments for project files.

rolfbjarne:
The generator project file still needs some custom logic, because we can't put
the generated makefile fragment next to the csproj (that would break the
API/generator diff).


rolfbjarne:
Test failure is unrelated

PR #8322, opened 2020-04-08
[Harness] Remove launchTimeout parameter from TestReporter, move factories

premun:
This parameter is not needed - only used for logging - and it was quite confusing with the other timeouts.

The change also moves one factory to a shared library and two from shared library back to XHarness.

No functional changes.

### DIFF
--- a/tests/xharness/AppRunner.cs
+++ b/tests/xharness/AppRunner.cs
@@ -269,7 +269,6 @@ namespace Xharness {
 				harness.XmlJargon,
 				deviceName,
 				testReporterTimeout,
-				harness.LaunchTimeout,
 				buildTask?.Logs?.Directory,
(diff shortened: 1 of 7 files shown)


premun:
@rolfbjarne I changed it and made it print out the exact actual time it took since it is using the `StopWatch` now so if a 30 second timeout is hit, it will print out 30 seconds. It still prints out correctly whether it was the launch or the test timeout that was hit.

`launchTimeout` wasn't used in any functional way inside that class and it was confusing because you would think it was from the API.

It is used one level up here:
https://github.com/xamarin/xamarin-macios/blob/f891076bd831750ae10042e770f144aa9e9286ee/tests/xharness/AppRunner.cs#L276-L279

Issue #8329, opened 2020-04-09
NSScreen.Screens crashed when not executed in the UIThread.


mandel-macaque:
VSMac has the following failing test: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-environment-logs&releaseId=637823&environmentId=3470218

Looking at the AppKit generated code we have:

```csharp
public static NSScreen[] Screens {
    Export ("screens", ArgumentSemantic.Copy)]
     get {
        global::AppKit.NSApplication.EnsureUIThread ();
        NSScreen[] ret;
        ret = NSArray.ArrayFromHandle<NSScreen>(global::ObjCRuntime.Messaging.IntPtr_objc_msgSend (class_ptr, selScreensHandle));
         return ret;
    }

}
```

Yet, the following swift code, that calls the property from a background thread works:
```swift
import Cocoa
import AppKit

DispatchQueue.global(qos: .background).async {
    print("This is run on the background queue")

    print(Thread.current)
    var screens = NSScreen.screens
    print (screens.count)
}
```

We should allow VSMac to get the screens array without the need of dispatching the execution of the property to the main thread.

PR #8330, opened 2020-04-09
[AppKit] Allow static properties to execute on non-UI threads.

mandel-macaque:
VSMac has failing tests when they query the NSScrees.Screens property
which the following swift code shows that it can be executed in a diff
thread:

```swift
import Cocoa
import AppKit

DispatchQueue.global(qos: .background).async {
    print("This is run on the background queue")

    print(Thread.current)
    var screens = NSScreen.screens
    print (screens.count)
}
```

Fixes: https://github.com/xamarin/xamarin-macios/issues/8329

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -13152,14 +13152,18 @@ namespace AppKit {
 
 	[BaseType (typeof (NSObject))]
 	partial interface NSScreen {
+
+		[ThreadSafe]
 		[Static]
 		[Export ("screens", ArgumentSemantic.Copy)]
 		NSScreen [] Screens { get; }
 
+		[ThreadSafe]
 		[Static]
 		[Export ("mainScreen")]
 		NSScreen MainScreen { get; }
 
+		[ThreadSafe]
 		[Static]
 		[Export ("deepestScreen")]
 		NSScreen DeepestScreen { get; }
--- a/tests/apitest/apitest.csproj
+++ b/tests/apitest/apitest.csproj
@@ -156,6 +156,7 @@
     <Compile Include="..\common\TestRuntime.cs">
       <Link>shared\TestRuntime.cs</Link>
     </Compile>
+    <Compile Include="src\AppKit\NSScreen.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\external\guiunit\src\framework\GuiUnit_NET_4_5.csproj">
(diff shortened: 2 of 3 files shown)


mandel-macaque:
@spouliot ran an application that does the same thing on xcode 11.4 with no issues. Is there any way to show it works (logs or something) here?

spouliot:
@mandel-macaque 
1. Enable MTC in your Xcode project;
2. Test your native sample with a selector that is not thread safe, e.g. `[NSWIndow init]`
3. Execute it

```
11  libsystem_pthread.dylib             0x000000010034b49b start_wqthread + 15
2020-04-09 09:14:59.338321-0400 mta[75054:965899] [reports] Main Thread Checker: UI API called on a background thread: -[NSWindow init]
PID: 75054, TID: 965899, Thread name: (none), Queue name: com.apple.root.default-qos, QoS: 0
```

4. You should get something like the above and know your project is setup correctly.
5. Try with each of the selector you _think_ should be thread safe

Also I don't think we should be adding the tests. They don't prove anything (at least not the _safety_ part we want) and even if a future MTC warns about them it's unlikely the tests would fail :(

mandel-macaque:
@spouliot I'm adding the tests in case anyone ever decides to change the attr or we screw it up someway. They prove we do call the selector correctly, do not throw and exception and get reasonable values.

mandel-macaque:
@spouliot as requested, the sample that gets called is the following:
```swift
override func viewDidLoad() {
        super.viewDidLoad()

        // Do any additional setup after loading the view.
        DispatchQueue.global(qos: .background).async {
            let window = NSWindow ()
            window.title = "Foo"
            print("This is run on the background queue")

            print(Thread.current)
            var screens = NSScreen.screens
            print (screens.count)
        }
    }
```
Output with the MTC complaining about NSWindow:
```
ain Thread Checker: UI API called on a background thread: -[NSWindow init]
PID: 8754, TID: 10065425, Thread name: (none), Queue name: com.apple.root.background-qos, QoS: 9
Backtrace:
```
Once we remove the NSWindow code, there is not complains from MTC.
```
This is run on the background queue
<NSThread: 0x60000171aa00>{number = 2, name = (null)}
1
2020-04-09 10:19:40.646574-0400 TestScreens[8766:10066384] Metal API Validation Enabled
```

mandel-macaque:
The internal checks is because I was creating a pkg for vsmac to test and get their tests fixed. Merging.

mandel-macaque:
@monojenkins backport d16-6

mandel-macaque:
@monojenkins backport d16-7

PR #8335, opened 2020-04-09
[xtro][bindings] Add xtro support for nullability and fix some bindings

spouliot:
Enabling this will ensure that future bindings (and Xcode updates that
change nullability information) are spotted right away.

The binding fixes are **not** complete, i.e. what was done was mostly to
debug the xtro rule and find corner cases. The backlog will be
_ignored_ so the builds won't fail.

Note: No _real_ need to review the 2nd commit, i.e. the `.ignore` files


spouliot:

PR #8336, opened 2020-04-09
[linker] Better workaround for the queue

spouliot:


spouliot:
I might have found the root issue

PR #8360, opened 2020-04-13
[tests] NugetRestore should only restore, not build.

rolfbjarne:
"msbuild /restore" will run nuget restore, then build. "msbuild /t:Restore"
will just run the Restore target, which should just restore.

This becomes significant when we later try to do "msbuild", and expect
warnings to show up. If we previously built the project unintentionally, those
warnings won't show up because nothing will actually be built.


rolfbjarne:
build

Issue #8364, opened 2020-04-13
NUnitExecuteTask do not generate HtmlReports.


mandel-macaque:
All tests based on the NUnitExecuteTask are not generating html reports, which make reading the test results harder. There have been several occurrences of this, for example in the generator tests. When running the task we get the following exception:

```
12:01:25.1819190 Failed to produce HTML report: System.ArgumentNullException: Value cannot be null.
Parameter name: output
  at System.Xml.XmlWriterSettings.CreateWriter (System.IO.TextWriter output) [0x000b0] in /Users/builder/jenkins/workspace/build-package-osx-mono/2020-02/external/bockbuild/builds/mono-x64/mcs/class/referencesource/System.Xml/System/Xml/Core/XmlWriterSettings.cs:643 
  at System.Xml.XmlWriter.Create (System.IO.TextWriter output, System.Xml.XmlWriterSettings settings) [0x0000a] in /Users/builder/jenkins/workspace/build-package-osx-mono/2020-02/external/bockbuild/builds/mono-x64/mcs/class/referencesource/System.Xml/System/Xml/Core/XmlWriter.cs:637 
  at Xharness.Jenkins.TestTasks.NUnitExecuteTask.RunTestAsync () [0x00367] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/xharness/Jenkins/TestTasks/NUnitExecuteTask.cs:144 
```

Issue #8347, opened 2020-04-10
Generator test fail in master 


mandel-macaque:
Generator tests started failing: 

http://xamarin-storage/jenkins/xamarin-macios/master/48b813e8a37a2af272b9cf13be70fb498725ab1f/1415/jenkins-results/tests/index.html

I cannot pin point a commit because we started having issues with xamarin-storage which means that we have to dig throw the build logs. Will post here once I get one.

mandel-macaque:
So after so digging I have the following:

The issue happens because when looking at the attributes of a method/class we are finding a NullableContextAttribute. To be more precise we get to https://github.com/xamarin/xamarin-macios/blob/master/src/generator-attribute-manager.cs#L198 which does not have a special case of that attribute, that returns null, and therefore we get an internal error in the tests.

looking at the liner sources, I though we would not have the attribute: https://github.com/xamarin/xamarin-macios/blob/b279401727fcc608868ecdd82297eb1744ba0f96/tools/linker/MobileRemoveAttributes.cs#L42 

So there is something fishy, either the linker is not doing its job, or maybe the generator tests are using a dll without the attrs removed?

spouliot:
@mandel-macaque most of the tests are not using the linker (or at least not always). Same for our normal code, i.e. linking is optional. In this case the generator is almost always (99+%) working on non-linked code (since only apps are linked and the generator works on assemblies).

spouliot:
Console/Makefile and xharness results differ :(

```
Run Settings
    DisposeRunners: True
    WorkDirectory: /Users/poupou/git/d16-6/xamarin-macios/tests/generator
    ImageRuntimeVersion: 4.0.30319
    ImageTargetFrameworkName: .NETFramework,Version=v4.6.1
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False
    NumberOfTestWorkers: 8

Test Run Summary
  Overall result: Passed
  Test Count: 132, Passed: 132, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2020-04-14 13:47:16Z
    End time: 2020-04-14 13:48:01Z
    Duration: 44.516 seconds

Results (nunit2) saved as /Users/poupou/git/d16-6/xamarin-macios/tests/generator/TestResult.xml
```

PR #8367, opened 2020-04-13
[Harness] Fix HTML report for NUnitTestTasks.

mandel-macaque:
PR https://github.com/xamarin/xamarin-macios/pull/8184 removed the
inheritance with TextWriter, therefore the `as` will return null and we
will not generate the Html report. In this particular case, we do not
need an ILog, we just use it to get a path to the correct location,
therefore, we can create the file using the full path and pass it to the
xslt.

Fixes: https://github.com/xamarin/xamarin-macios/issues/8364


mandel-macaque:
After this PR we get:

Screenshot 2020-04-13 at 11 52 46

That contains a html report as it follows:

Screenshot 2020-04-13 at 11 55 57

mandel-macaque:
Known failures: https://github.com/xamarin/xamarin-macios/issues/8347 and the show the PR works.

spouliot:

Issue #8363, opened 2020-04-13
Generator tests were not run when changes in the generator ocurred


mandel-macaque:
There are several PRs that have modified the generator and have not triggered the generator tests:

* https://github.com/xamarin/xamarin-macios/pull/8333
* https://github.com/xamarin/xamarin-macios/pull/8301

Which meant that we have a number of generator tests failing in master: https://github.com/xamarin/xamarin-macios/issues/8347

If the generator sources are modified, the generator tests most be ran.

PR #8371, opened 2020-04-13
[Harness] Ensure that the generator tests are set when we modify the sources.

mandel-macaque:
fixes: https://github.com/xamarin/xamarin-macios/issues/8363

### DIFF
--- a/tests/xharness/Jenkins/Jenkins.cs
+++ b/tests/xharness/Jenkins/Jenkins.cs
@@ -782,7 +782,7 @@ namespace Xharness.Jenkins {
 
 		void SetEnabled (IEnumerable<string> files, string [] prefixes, string testname, ref bool value)
 		{
-			MainLog.WriteLine ($"Checking if test {testname} should be enabled accoding to the modified files.");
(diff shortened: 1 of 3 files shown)


mandel-macaque:
@spouliot as you can see in the previous commits, test run when we do change the generator sources. Issue was that the timestamp was added when executing the git diff and the compare would not longer work. Added some logging for the next time.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -782,7 +782,9 @@ void SelectTestsByModifiedFiles (int pull_request)
 
 		void SetEnabled (IEnumerable<string> files, string [] prefixes, string testname, ref bool value)
 		{
+			MainLog.WriteLine ($"Checking if test {testname} should be enabled accoding to the modified files.");
```
rolfbjarne, line 785:
```suggestion
			MainLog.WriteLine ($"Checking if test {testname} should be enabled according to the modified files.");
```

spouliot:

PR #8376, opened 2020-04-13
[ImageIO] Update bindings for Xcode11

whitneyschmidt:
Sample: https://github.com/xamarin/private-samples/pull/58


src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ IntPtr block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+        public virtual int CGAnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
+            if (url == null)
+                throw new ArgumentNullException ("url");
```
mandel-macaque, line 34:
```suggestion
                throw new ArgumentNullException (nameof (url));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (block == null)
+                throw new ArgumentNullException ("block");
```
mandel-macaque, line 36:
```suggestion
                throw new ArgumentNullException (nameof (block));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+using ObjCRuntime;
+
+namespace ImageIO {
+
+    public partial class CGImageAnimation {
+
+        public delegate void CGImageSourceAnimationBlock (nint arg0, CGImage arg1, out bool done /*check*/ );
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern unsafe int CGAnimateImageAtURLWithBlock (/* CFURLRef */ IntPtr url, /* CFDictionaryRef _iio_Nullable */ IntPtr options, /* CGImageSourceAnimationBlock */ IntPtr block);
```
mandel-macaque, line 24: I believe we can pass the Block as a ref like we do here: https://github.com/xamarin/xamarin-macios/blob/master/src/Network/NWConnection.cs#L203

Would change the call you do later in:
```csharp
ret = CGAnimateImageAtURLWithBlock (url.Handle, options == null ? IntPtr.Zero : options.Handle, (IntPtr) block_ptr_block);
```
to
```csharp
ret = CGAnimateImageAtURLWithBlock (url.Handle, options == null ? IntPtr.Zero : options.Handle, ref block_block);
```

Once you have done that, we do not longer need the unsafe block:
```csharp
                BlockLiteral block_block;
                block_block = new BlockLiteral ();
                block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, block);

                try {
                    ret = CGAnimateImageAtURLWithBlock (url.Handle, options == null ? IntPtr.Zero : options.Handle, ref block_block);
                } finally {
                    block_block.CleanupBlock ();
                }
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ IntPtr block);
```
mandel-macaque, line 27: Same as with the CGAnimateImageAtURLWithBlock call, we can use ref BlockLiteral AFAIK and we get the pointers out of our way.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            return ret;
+        }
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+        public virtual int CGAnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
+            if (data == null)
+                throw new ArgumentNullException ("data");
```
mandel-macaque, line 61:
```suggestion
                throw new ArgumentNullException (nameof (data));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (block == null)
+                throw new ArgumentNullException ("block");
```
mandel-macaque, line 63:
```suggestion
                throw new ArgumentNullException (nameof (block));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+        //
+        // This class bridges native block invocations that call into C#
+        //
+        static internal class SDCGImageSourceAnimationBlock {
+            static internal readonly DCGImageSourceAnimationBlock Handler = Invoke;
+
+            [MonoPInvokeCallback (typeof (DCGImageSourceAnimationBlock))]
+            static unsafe void Invoke (IntPtr block, nint arg0, IntPtr arg1, [System.Runtime.InteropServices.MarshalAs (System.Runtime.InteropServices.UnmanagedType.I1)] out bool done) {
+                var descriptor = (BlockLiteral * ) block;
+                var del = (CGImageSourceAnimationBlock) (descriptor -> Target);
```
mandel-macaque, line 92: I believe we can do nicer code like in https://github.com/xamarin/xamarin-macios/blob/master/src/Network/NWConnection.cs#L195

```csharp
var del = BlockLiteral.GetTarget< CGImageSourceAnimationBlock> (block);
```
And we get all the pointers out of the way.

src/imageio.cs
```diff
@@ -2356,4 +2368,17 @@ interface CGImageAuxiliaryDataInfo {
...
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	interface CGImageAnimationOptions {
+		[Field ("kCGImageAnimationDelayTime")]
+		NSString DelayTime { get; }
+
+		[Field ("kCGImageAnimationLoopCount")]
+		NSString LoopCount { get; }
+
+		[Field ("kCGImageAnimationStartIndex")]
+		NSString StartIndex { get; }
```
mandel-macaque, line 2382: Is StartIndex an NSNumber? Could we use BindAs?

src/imageio.cs
```diff
@@ -2356,4 +2368,17 @@ interface CGImageAuxiliaryDataInfo {
...
 	}
+
+	[Partial]
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	interface CGImageAnimationOptions {
+		[Field ("kCGImageAnimationDelayTime")]
+		NSString DelayTime { get; }
+
+		[Field ("kCGImageAnimationLoopCount")]
+		NSString LoopCount { get; }
```
mandel-macaque, line 2379: Is it an NSNumber of some kind? Can we use BindAs?

src/imageio.cs
```diff
@@ -2356,4 +2368,17 @@ interface CGImageAuxiliaryDataInfo {
 		NSData Data { get; set; }
 		NSDictionary DataDescription { get; set; }
 	}
+
+	[Partial]
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	interface CGImageAnimationOptions {
+		[Field ("kCGImageAnimationDelayTime")]
+		NSString DelayTime { get; }
```
mandel-macaque, line 2376: Is it an NSNumber of some kind? Can we use BindAs?

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        //
+        // This class bridges native block invocations that call into C#
+        //
+        static internal class SDCGImageSourceAnimationBlock {
+            static internal readonly DCGImageSourceAnimationBlock Handler = Invoke;
+
+            [MonoPInvokeCallback (typeof (DCGImageSourceAnimationBlock))]
+            static unsafe void Invoke (IntPtr block, nint arg0, IntPtr arg1, [System.Runtime.InteropServices.MarshalAs (System.Runtime.InteropServices.UnmanagedType.I1)] out bool done) {
```
mandel-macaque, line 90: Lets not use the generator variable names, I'm sure we know what arg0 and arg1 are. And using meaningful names will make our future lifes better.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+using CoreFoundation;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+
+namespace ImageIO {
+
+    public partial class CGImageAnimation {
+
+        public delegate void CGImageSourceAnimationBlock (nint arg0, CGImage arg1, out bool done /*check*/ );
```
rolfbjarne, line 21: According to the documentation, the parameters should be named:

```suggestion
        public delegate void CGImageSourceAnimationBlock (nint index, CGImage image, out bool stop /*check*/ );
```

src/ImageIO/CGImageAnimation.cs
rolfbjarne, line 24: Missing availability attributes.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern unsafe int CGAnimateImageAtURLWithBlock (/* CFURLRef */ IntPtr url, /* CFDictionaryRef _iio_Nullable */ IntPtr options, /* CGImageSourceAnimationBlock */ IntPtr block);
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ IntPtr block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+        public virtual int CGAnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
```
rolfbjarne, line 32: Any particular reason this is virtual?

Also I think we can remove the `CG` prefix:

```suggestion
        public int AnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+
+            return ret;
+        }
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+        public virtual int CGAnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
```
rolfbjarne, line 59: Same: non-virtual and remove the `CG` prefix:

```suggestion
        public int AnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern unsafe int CGAnimateImageAtURLWithBlock (/* CFURLRef */ IntPtr url, /* CFDictionaryRef _iio_Nullable */ IntPtr options, /* CGImageSourceAnimationBlock */ IntPtr block);
+
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ IntPtr block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
```
spouliot, line 31: remove `BindingImplOptions.GeneratedCode | ` as this is manual, not generated, code

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+            int ret;
+            unsafe {
+                BlockLiteral * block_ptr_block;
+                BlockLiteral block_block;
+                block_block = new BlockLiteral ();
+                block_ptr_block = & block_block;
+                block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, block);
+
+                try {
+                    ret = CGAnimateImageAtURLWithBlock (url.Handle, options == null ? IntPtr.Zero : options.Handle, (IntPtr) block_ptr_block);
```
spouliot, line 47: you can use the simpler `options.GetHandle ()` to replace `options == null ? IntPtr.Zero : options.Handle` as it will handle the null case.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,128 @@
...
+            int ret;
+            unsafe {
+                BlockLiteral * block_ptr_block;
+                BlockLiteral block_block;
+                block_block = new BlockLiteral ();
+                block_ptr_block = & block_block;
+                block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, block);
+
+                try {
+                    ret = CGAnimateImageDataWithBlock(data.Handle, options == null ? IntPtr.Zero : options.Handle, (IntPtr) block_ptr_block);
```
spouliot, line 74: minor: missing space before `(`
same comment about `GetHandle ()`

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed.

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: updated! Thank you for the refresher on `unsafe`/`IntPtr`/passing by reference :)

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: done

src/imageio.cs
whitneyschmidt: From Apple documentation: `const CFStringRef kCGImageAnimationStartIndex;`

I think this means that we should leave it as an NSString?

src/imageio.cs
whitneyschmidt: Likewise, from the docs: `const CFStringRef kCGImageAnimationLoopCount;`

src/imageio.cs
whitneyschmidt: Apple docs: `const CFStringRef kCGImageAnimationDelayTime;`

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: much better, thank you!

whitneyschmidt:
> This is also missing tests: all manually written bindings should have tests in monotouch-test (and this is even more important when dealing with manually written block bindings).

Does this apply to manual bindings with a sample? I have a sample (to be published soon) that tests both of the manual blocks, but if we need tests as well I am happy to write them :)

src/imageio.cs
mandel-macaque: Yet we have in the header:
```
/* The value of this key overrides the `loop count' specified by the image
+ * Value is a CFNumber; kCFNumberPositiveInfinity may be used.
+ */
+IMAGEIO_EXTERN const CFStringRef kCGImageAnimationLoopCount  IMAGEIO_AVAILABLE_STARTING(10.15, 13.0);
```
So it is a CFNumber.

src/imageio.cs
mandel-macaque: From the headers:
```
/* Starts the animation at the given index. Defaults to 0
+ * Value is a CFNumber
+ */
+IMAGEIO_EXTERN const CFStringRef kCGImageAnimationStartIndex  IMAGEIO_AVAILABLE_STARTING(10.15, 13.0);
```
So it is a CFNumber.

src/imageio.cs
mandel-macaque: From the headers:
```
+/* The value of this key overrides the `delay time' specified by the image
+ * Value is a CFNumber of kCFNumberDoubleType.
+ */
+IMAGEIO_EXTERN const CFStringRef kCGImageAnimationDelayTime  IMAGEIO_AVAILABLE_STARTING(10.15, 13.0);
```
To it is a CFNumber.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+        internal sealed class NIDCGImageSourceAnimationBlock : TrampolineBlockBase {
+            DCGImageSourceAnimationBlock invoker;
+
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+            public unsafe NIDCGImageSourceAnimationBlock (BlockLiteral * block) : base (block) {
+                invoker = block -> GetDelegateForBlock<DCGImageSourceAnimationBlock> ();
+            }
+
+            [Preserve (Conditional = true)]
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
```
mandel-macaque, line 112: Is not generated code, we have to remove the flag.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+                    del (index, new CoreGraphics.CGImage (image), out stop);
+                else
+                    stop = default (bool);
+            }
+        } /* class SDCGImageSourceAnimationBlock */
+
+        internal sealed class NIDCGImageSourceAnimationBlock : TrampolineBlockBase {
+            DCGImageSourceAnimationBlock invoker;
+
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
```
mandel-macaque, line 106: Not generated code, we have to remove the flag.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+            [Preserve (Conditional = true)]
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+            public unsafe static CGImageSourceAnimationBlock Create (IntPtr block) {
+                if (block == IntPtr.Zero)
+                    return null;
+                var del = (CGImageSourceAnimationBlock) GetExistingManagedDelegate (block);
+                return del ?? new NIDCGImageSourceAnimationBlock ((BlockLiteral * ) block).Invoke;
+            }
+
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
```
mandel-macaque, line 120: Same.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+            public unsafe NIDCGImageSourceAnimationBlock (BlockLiteral * block) : base (block) {
+                invoker = block -> GetDelegateForBlock<DCGImageSourceAnimationBlock> ();
+            }
+
+            [Preserve (Conditional = true)]
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+            public unsafe static CGImageSourceAnimationBlock Create (IntPtr block) {
+                if (block == IntPtr.Zero)
+                    return null;
+                var del = (CGImageSourceAnimationBlock) GetExistingManagedDelegate (block);
```
mandel-macaque, line 116: Cant we get use the GetTarget method with the pointer here?

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+                return del ?? new NIDCGImageSourceAnimationBlock ((BlockLiteral * ) block).Invoke;
+            }
+
+            [BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
+            unsafe void Invoke (nint arg0, CGImage arg1, out bool done) {
```
mandel-macaque, line 121: arg0 and arg1 are not very reader friendly.

mandel-macaque:
@whitneyschmidt CFNumber is “toll-free bridged” with its Cocoa Foundation counterpart, NSNumber. For a list of them: https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFDesignConcepts/Articles/tollFreeBridgedTypes.html

rolfbjarne:
> > This is also missing tests: all manually written bindings should have tests in monotouch-test (and this is even more important when dealing with manually written block bindings).
> 
> Does this apply to manual bindings with a sample? I have a sample (to be published soon) that tests both of the manual blocks, but if we need tests as well I am happy to write them :)

Yes, because it's rather unlikely someone will run the sample if they want to modify this code - the sample is to show how to use the API, not to verify that it doesn't regress.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace ImageIO {
+
+    public partial class CGImageAnimation {
+
+        public delegate void CGImageSourceAnimationBlock (nint index, CGImage image, out bool stop);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern unsafe int CGAnimateImageAtURLWithBlock (/* CFURLRef */ IntPtr url, /* CFDictionaryRef _iio_Nullable */ IntPtr options, /* CGImageSourceAnimationBlock */ ref BlockLiteral block);
```
rolfbjarne, line 26: You don't need `unsafe` anymore:

```suggestion
        static extern int CGAnimateImageAtURLWithBlock (/* CFURLRef */ IntPtr url, /* CFDictionaryRef _iio_Nullable */ IntPtr options, /* CGImageSourceAnimationBlock */ ref BlockLiteral block);
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        //
+        // This class bridges native block invocations that call into C#
+        //
+        static internal class SDCGImageSourceAnimationBlock {
+            static internal readonly DCGImageSourceAnimationBlock Handler = Invoke;
+
+            [MonoPInvokeCallback (typeof (DCGImageSourceAnimationBlock))]
+            static void Invoke (IntPtr block, nint index, IntPtr image, [System.Runtime.InteropServices.MarshalAs (System.Runtime.InteropServices.UnmanagedType.I1)] out bool stop) {
+                var del = BlockLiteral.GetTarget<CGImageSourceAnimationBlock>(block);
```
rolfbjarne, line 95:
```suggestion
                var del = BlockLiteral.GetTarget<CGImageSourceAnimationBlock> (block);
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ ref BlockLiteral block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.Optimizable)]
+        public int AnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block) {
```
rolfbjarne, line 36: Our coding style says the starting brace for methods should be on a new line.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,131 @@
...
+using System;
+using System.Runtime.InteropServices;
+using CoreFoundation;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+
+namespace ImageIO {
+
+    public partial class CGImageAnimation {
```
rolfbjarne, line 19: It looks like this file doesn't follow our coding standards, because it seems to be using spaces to indent instead of tabs.

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed.

src/imageio.cs
whitneyschmidt: Good catch :) Changed to `NSNumber`

src/imageio.cs
mandel-macaque: ok, I'm stupid, is a key, I can't parse english: 'The **value** of this **key**'. To the value if a double.

src/imageio.cs
mandel-macaque: same, I'm an idiot.

src/imageio.cs
mandel-macaque: ditto, stupid manuel

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: I might be missing something, but I don't think this is possible. `GetTarget<CGImageSourceAnimationBlock>` returns type `CGImageSourceAnimationBlock`. 

We cannot convert from this to a `BlockLiteral`, which is needed by the constructor. The constructor will still need to call `GetDelegateForBlock` as well as the base constructor (`TrampolineBlockBase`).

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			stop = true;
+			testValue = 1;
+		}
+
+		[TestFixtureSetUp]
+		public void Init ()
+		{
+			TestRuntime.AssertXcodeVersion (11, 4);
+
+			imageUrl = NSBundle.MainBundle.GetUrlForResource ("hack", "gif"); // why not just initialize in declaration?
```
spouliot, line 57: Unless you need this one specifically let's reuse some existing tests resources. That minimize the .app size (so it builds and deploy faster).

e.g. `NSBundle.MainBundle.GetUrlForResource ("xamarin2", "png")`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			imageData = NSData.FromUrl(imageUrl);
```
spouliot, line 58: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		public void InitPerTest()
+		{
+			TestRuntime.AssertXcodeVersion (11, 4);
+			testValue = -1;
+		}
+
+		[TearDown]
+		public void Dispose()
+		{
+			TestRuntime.AssertXcodeVersion (11, 4); // this seems cleaner, but isn't strictly necessary?
```
spouliot, line 72: not needed
also there's no harm, if executed, to execute the following code

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			testValue = -1;
+		}
+
+		[Test]
+		public void AnimateImageWithUrl ()
+		{
+			Assert.AreEqual(-1, testValue);
```
spouliot, line 79: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
+
+			bool done = false;
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage(imageUrl, null, MyBlockSetValueZero);
```
spouliot, line 84: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+				await taskCompletionSource.Task;
+				done = true;
+			},
+				() => done);
+			Assert.AreEqual (0, testValue);
```
spouliot, line 89: whenever there's more than one assert you must add a label to all assert
sometimes the debugging symbols are missing (or broken) and the _correct_ assert failing is hard to diagnose from the logs. Having the label makes them all unique (reduce confusion/errors)

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			},
+				() => done);
+			Assert.AreEqual (1, testValue);
+		}
+
+		[Test]
+		public void AnimateImageWithUrlNullUrl()
+		{
+			NSUrl nullUrl = null;
+			Assert.Throws<ArgumentNullException>(() => imageAnimation.AnimateImage(nullUrl, null, MyBlockSetValueZero));
```
spouliot, line 163: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithUrlNullBlock()
+		{
+			Assert.Throws<ArgumentNullException>(() => imageAnimation.AnimateImage(imageUrl, null, /* CGImageSourceAnimationBlock */ null));
```
spouliot, line 169: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithDataNullData()
+		{
+			NSData nullData = null;
+			Assert.Throws<ArgumentNullException>(() => imageAnimation.AnimateImage(nullData, null, MyBlockSetValueZero));
```
spouliot, line 176: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithDataNullBlock()
+		{
+			Assert.Throws<ArgumentNullException>(() => imageAnimation.AnimateImage(imageData, null, /* CGImageSourceAnimationBlock */ null));
```
spouliot, line 182: minor: missing space before `(`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
spouliot: minor: you do not need a local variable, you can just cast the null, e.g. `(NSData)null`

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
mandel-macaque: AFAIK, needs to be a gif and not a png :/

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
+
+			bool done = false;
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage (imageUrl, null, MyBlockSetValueZero);
+				await taskCompletionSource.Task;
+				done = true;
+			},
+				() => done);
+			Assert.AreEqual (0, testValue);
```
mandel-macaque, line 122: Add a message to the assert since we have several.

mandel-macaque, line 122: Add a message to the assert since we have several.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		[Test]
+		public void AnimateImageWithData ()
+		{
+
+			Assert.AreEqual (-1, testValue);
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
+
+			bool done = false;
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage(imageData, null, MyBlockSetValueZero);
```
mandel-macaque, line 101: minor, missing space:
```suggestion
				imageAnimation.AnimateImage (imageData, null, MyBlockSetValueZero);
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			Assert.AreEqual (0, testValue);
+
+			done = false; // reset after first call
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage (imageUrl, null, MyBlockSetValueOne);
+				await taskCompletionSource.Task;
+				done = true;
+			},
+				() => done);
+			Assert.AreEqual (1, testValue);
```
mandel-macaque, line 131: same

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithDataChangeBlock ()
+		{
+			Assert.AreEqual (-1, testValue);
```
mandel-macaque, line 137: Same

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			Assert.AreEqual (0, testValue);
+
+			done = false; // reset after first call
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage (imageData, null, MyBlockSetValueOne);
+				await taskCompletionSource.Task;
+				done = true;
+			},
+				() => done);
+			Assert.AreEqual (1, testValue);
```
mandel-macaque, line 156: Same

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+		public void AnimateImageWithUrlChangeBlock ()
+		{
+			Assert.AreEqual (-1, testValue);
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
+
+			bool done = false;
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				imageAnimation.AnimateImage (imageUrl, null, MyBlockSetValueZero);
+				await taskCompletionSource.Task;
+				done = true;
```
mandel-macaque, line 119: This code is repeated in all of the tests, you can do better than that. Move it to a reusable function that takes the appropriate parameters and reuse the code.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
mandel-macaque, line 137: Same

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			},
+				() => done);
+			Assert.AreEqual (0, testValue);
+		}
+
+		[Test]
+		public void AnimateImageWithData ()
+		{
+
+			Assert.AreEqual (-1, testValue);
```
mandel-macaque, line 96: I don't believe this assert gives a lot of value, and it is done in very test. Should be -1  already thanks to the setup. If not, setup or tear down are wrong.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
mandel-macaque, line 101: minor, missing space:
```suggestion
				imageAnimation.AnimateImage (imageData, null, MyBlockSetValueZero);
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
mandel-macaque: AFAIK, needs to be a gif and not a png :/

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+namespace MonoTouchFixtures.ImageIO {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class CGImageAnimationTest {
+
+		NSUrl imageUrl;
+		NSData imageData;
+		int testValue;
+		CGImageAnimation imageAnimation;
```
mandel-macaque, line 38: Looks like we are not calling Dispose on this object, would be nice to clean after the tests in the TearDown.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+using NUnit.Framework;
+
+namespace MonoTouchFixtures.ImageIO {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class CGImageAnimationTest {
+
+		NSUrl imageUrl;
+		NSData imageData;
```
mandel-macaque, line 36: Call Dispose on this guys in the tear down.

whitneyschmidt: Added dispose calls for `NSUrl` and `NSData`.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: Removed 👍

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: It needs to be a .gif for animation :/

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: added labels to all asserts - thank you!

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: fixed 👍

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,186 @@
...
+			},
+				() => done);
+			Assert.AreEqual (0, testValue);
+		}
+
+		[Test]
+		public void AnimateImageWithUrlChangeBlock ()
+		{
+			Assert.AreEqual (-1, testValue);
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
```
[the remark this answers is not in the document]
whitneyschmidt: I moved this to a shared function. I think it's best not to move it to per-test setup since we have all of the null exception tests which do not need a `TaskCompletionSource` object.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: Okay, okay. Moved to a private shared function 👍

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+using Foundation;
+using ObjCRuntime;
+
+namespace ImageIO
+{
+
+    public partial class CGImageAnimation
+    {
+
+        public delegate void CGImageSourceAnimationBlock (nint index, CGImage image, out bool stop);
```
dalexsoto, line 23: Let's not use `Block` in our public API since it really does not mean much for a .NET developer, you can suffix stuff like this with `Handler` instead.

```suggestion
        public delegate void CGImageSourceAnimationHandler (nint index, CGImage image, out bool stop);
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern int CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationBlock _Nonnull */ ref BlockLiteral block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.Optimizable)]
+        public int AnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block)
```
dalexsoto, line 38: Same, let's avoid `Block` as a suffix or parameter name.

```suggestion
        public int AnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationHandler handler)
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+#if IOS && ARCH_32
+            throw new PlatformNotSupportedException ();
+#else
+            if (url == null)
+                throw new ArgumentNullException (nameof (url));
+            if (block == null)
+                throw new ArgumentNullException (nameof (block));
```
dalexsoto, line 46:
```suggestion
            if (handler == null)
                throw new ArgumentNullException (nameof (handler));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            int ret;
+            BlockLiteral block_block;
+            block_block = new BlockLiteral ();
+            block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, block);
```
dalexsoto, line 51:
```suggestion
            block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, handler);
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+
+            return ret;
+#endif
+        }
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.Optimizable)]
+        public int AnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationBlock block)
```
dalexsoto, line 66:
```suggestion
        public int AnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationHandler handler)
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+#if IOS && ARCH_32
+            throw new PlatformNotSupportedException ();
+#else
+            if (data == null)
+                throw new ArgumentNullException (nameof (data));
+            if (block == null)
+                throw new ArgumentNullException (nameof (block));
```
dalexsoto, line 74:
```suggestion
            if (handler == null)
                throw new ArgumentNullException (nameof (handler));
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            int ret;
+            BlockLiteral block_block;
+            block_block = new BlockLiteral ();
+            block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, block);
```
dalexsoto, line 79:
```suggestion
            block_block.SetupBlockUnsafe (SDCGImageSourceAnimationBlock.Handler, handler);
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+using UIKit;
+#endif
+using CoreGraphics;
+using ImageIO;
+#else
+using MonoTouch.CoreGraphics;
+using MonoTouch.Foundation;
+using MonoTouch.ImageIO;
+using MonoTouch.UIKit;
+#endif
```
dalexsoto, line 26: are the MonoTouch namespaces needed? IIRC we no longer run tests using classic API

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+			testValue = -1;
+		}
+
+		[TearDown]
+		public void Dispose ()
+		{
+			testValue = -1;
+		}
+
+		private void CallAnimateImage (bool useUrl, CGImageAnimation.CGImageSourceAnimationBlock block)
```
dalexsoto, line 81:
```suggestion
		private void CallAnimateImage (bool useUrl, CGImageAnimation.CGImageSourceAnimationHandler handler)
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+		{
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
+			bool done = false;
+
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (30), async () => {
+				if (useUrl) {
+					imageAnimation.AnimateImage (imageUrl, null, block);
+				} else {
+					imageAnimation.AnimateImage (imageData, null, block);
```
dalexsoto, line 90:
```suggestion
					imageAnimation.AnimateImage (imageUrl, null, handler);
				} else {
					imageAnimation.AnimateImage (imageData, null, handler);
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+
+		[TearDown]
+		public void Dispose ()
+		{
+			testValue = -1;
+		}
+
+		private void CallAnimateImage (bool useUrl, CGImageAnimation.CGImageSourceAnimationBlock block)
+		{
+			TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool> ();
```
dalexsoto, line 83: Nit:
```suggestion
			var taskCompletionSource = new TaskCompletionSource<bool> ();
```

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            try {
+                ret = CGAnimateImageAtURLWithBlock (url.Handle, options.GetHandle (), ref block_block);
+            } finally {
+                block_block.CleanupBlock ();
+            }
+
+            return ret;
```
spouliot, line 59: minor (no need to change) but you can `return` in the `try` block and the `finally` clause will still be executed. e.g.
```csharp
try {
    return CGAnimateImageAtURLWithBlock (...);
} finally {
    block_block.CleanupBlock ();
}
```
and get rid of `ret` local :)

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        //
+        // This class bridges native block invocations that call into C#
+        //
+        static internal class SDCGImageSourceAnimationBlock
+        {
+            static internal readonly DCGImageSourceAnimationBlock Handler = Invoke;
+
+            [MonoPInvokeCallback (typeof (DCGImageSourceAnimationBlock))]
+            static void Invoke (IntPtr block, nint index, IntPtr image, [System.Runtime.InteropServices.MarshalAs (System.Runtime.InteropServices.UnmanagedType.I1)] out bool stop)
```
spouliot, line 99: `System.Runtime.InteropServices` is already in your `using` clauses so it can be removed.
I'm lazy and rather not scroll horizontally when reviewing ;-)

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+                if (block == IntPtr.Zero)
+                    return null;
+                var del = (CGImageSourceAnimationBlock) GetExistingManagedDelegate (block);
+                return del ?? new NIDCGImageSourceAnimationBlock ( (BlockLiteral *) block).Invoke;
+            }
+
+            [BindingImpl (BindingImplOptions.Optimizable)]
+            void Invoke (nint index, CGImage image, out bool stop)
+            {
+                invoker (BlockPointer, index, image == null ? IntPtr.Zero : image.Handle, out stop);
```
spouliot, line 132: `image == null ? IntPtr.Zero : image.Handle` -> `image.GetHandle`

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+        } /* class NIDCGImageSourceAnimationBlock */
+
+        [UnmanagedFunctionPointerAttribute (CallingConvention.Cdecl)]
+        [UserDelegateType (typeof (CGImageSourceAnimationBlock))]
+        internal delegate void DCGImageSourceAnimationBlock (IntPtr block, nint index, IntPtr imageHandle, [System.Runtime.InteropServices.MarshalAs (System.Runtime.InteropServices.UnmanagedType.I1)] out bool stop);
```
spouliot, line 138: unneeded `System.Runtime.InteropServices` (twice)

src/ImageIO/Enums.cs
```diff
@@ -60,4 +61,14 @@ public enum CGImagePropertyPngFilters {
 		Average = 0x40,
 		Paeth = 0x80
 	}
+
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	public enum CGImageAnimationStatus
```
spouliot, line 66: where is that used ?

the animation API you added are also only on iOS and macOS, but the enum is marked as available on all platforms

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		[TestFixtureSetUp]
+		public void Init ()
+		{
+			TestRuntime.AssertXcodeVersion (11, 4);
+
+			imageUrl = NSBundle.MainBundle.GetUrlForResource ("hack", "gif"); // why not just initialize in declaration?
+			imageData = NSData.FromUrl (imageUrl);
+			imageAnimation = new CGImageAnimation ();
```
spouliot, line 59: there's nothing in the file to exclude tvOS or watchOS so it means

* the API is available everywhere - but it's missing the attributes (making available in watchOS 2.0 and tvOS 9.0, which is unlikely to be correct)

* the tests are not executed

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+			Assert.AreEqual (0, testValue, "first block called with data");
+
+			CallAnimateImage ( /* useUrl */ false, MyBlockSetValueOne);
+			Assert.AreEqual (1, testValue, "second block called with data");
+		}
+
+		[Test]
+		public void AnimateImageWithUrlNullUrl ()
+		{
+			Assert.Throws<ArgumentNullException>( () => imageAnimation.AnimateImage ( (NSUrl) null, null, MyBlockSetValueZero), "null url");
```
spouliot, line 135: minor: space before `(`, not after

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithUrlNullBlock ()
+		{
+			Assert.Throws<ArgumentNullException>( () => imageAnimation.AnimateImage (imageUrl, null, /* CGImageSourceAnimationBlock */ null), "null block called with url");
```
spouliot, line 141: minor: space before `(`, not after

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithDataNullData ()
+		{
+			Assert.Throws<ArgumentNullException>( () => imageAnimation.AnimateImage ( (NSData) null, null, MyBlockSetValueZero), "null data");
```
spouliot, line 147: minor: space before `(`, not after

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		[Test]
+		public void AnimateImageWithDataNullBlock ()
+		{
+			Assert.Throws<ArgumentNullException>( () => imageAnimation.AnimateImage (imageData, null, /* CGImageSourceAnimationBlock */ null), "null block called with data");
```
spouliot, line 153: minor: space before `(`, not after

src/ImageIO/CGImageAnimation.cs
dalexsoto, line 38: This returns an `int` which in reality it is an `OSStatus` but I wonder if we can do even better, there is `CGImageAnimationStatus` inside `CGImageAnimation.h` I think it is better to return `CGImageAnimationStatus` 

@spouliot what do you think?

src/ImageIO/CGImageAnimation.cs
dalexsoto: This returns an `int` which in reality it is an `OSStatus` but I wonder if we can do even better, there is `CGImageAnimationStatus` inside `CGImageAnimation.h` I think it is better to return `CGImageAnimationStatus` 

@spouliot what do you think?

src/ImageIO/Enums.cs
dalexsoto: @spouliot  I think it should be the return type for both `AnimateImage` members please see my comment on them 😄

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: fixed! :)

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: @dalexsoto @spouliot this seems like a good idea. One more question: the enum looks like this:

```
public enum CGImageAnimationStatus
	{
		ParameterError = -22140,
		CorruptInputImage = -22141,
		UnsupportedFormat = -22142,
		IncompleteInputImage = -22143,
		AllocationFailure = -22144,
	}
```

It seems like `None = 0` should be added, thoughts?

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: removed MonoTouch namespaces! Lmk if any other namespaces here look like they are not needed, these using are copied from another ImageIO test.

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
whitneyschmidt: @spouliot - I've added Watch and TV attributes to `CGImageAnimation.cs` to prevent the API from being included in the min versions of these platforms. Let me know if I understood your comment correctly :)

src/ImageIO/CGImageAnimation.cs
spouliot: Good point! but looks at `OSStatus` enum to see how `0` was named (for consistency).

src/ImageIO/Enums.cs
whitneyschmidt: I thought that the animation API is available on Watch and TV as well? It appears in the diffs for those device types as well as in the xtro output for all device types. Am I missing something? It's also marked as available across all in the Apple Docs, but I know we can't trust those...

src/ImageIO/CGImageAnimation.cs
whitneyschmidt: Added `Ok = 0` (thanks for the FYI on `enums` @dalexsoto!). 

Looks like we use `Ok` normally in our code: https://github.com/xamarin/xamarin-macios/blob/5eb593f4ba86ef2c3dd9554db1ab5a9cbb9df923/src/VideoToolbox/VTDefs.cs#L17-L18

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+				() => done);
+
+			tcs = null;
+		}
+
+		[Test]
+		public void AnimateImageWithUrl ()
+		{
+			CallAnimateImage ( /* useUrl */ true, MyHandlerSetValueZero);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: handler called with url");
```
mandel-macaque, line 104: Why not AssertEqual? 
```suggestion
			Assert.AreEqual (CGImageAnimationStatus.Ok, status, "status ok: handler called with url");
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.AreEqual (0, testValue, "handler called with url");
+		}
+
+		[Test]
+		public void AnimateImageWithData ()
+		{
+			CallAnimateImage ( /* useUrl */ false, MyHandlerSetValueZero);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: handler called with data");
```
mandel-macaque, line 112: Same
```suggestion
			Assert.AreEqual (CGImageAnimationStatus.Ok, status, "status ok: handler called with data");
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.AreEqual (0, testValue, "handler called with data");
+		}
+
+		[Test]
+		public void AnimateImageWithUrlChangeHandler ()
+		{
+			CallAnimateImage ( /* useUrl */ true, MyHandlerSetValueZero);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: first handler called with url");
```
mandel-macaque, line 120: Same
```suggestion
			Assert.AreEqual (CGImageAnimationStatus.Ok, status, "status ok: first handler called with url");
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.AreEqual (0, testValue, "first handler called with url" );
+
+			CallAnimateImage ( /* useUrl */ true, MyHandlerSetValueOne);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: second handler called with url");
```
mandel-macaque, line 124: Same
```suggestion
			Assert.AreEqual (CGImageAnimationStatus.Ok, status, "status ok: second handler called with url");
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.AreEqual (1, testValue, "second handler called with url");
+		}
+
+		[Test]
+		public void AnimateImageWithDataChangeHandler ()
+		{
+			CallAnimateImage ( /* useUrl */ false, MyHandlerSetValueZero);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: first handler called with data");
```
mandel-macaque, line 132: Same
```suggestion
			Assert.IsTrue (CGImageAnimationStatus.Ok, status, "status ok: first handler called with data");
```

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			Assert.AreEqual (0, testValue, "first handler called with data");
+
+			CallAnimateImage ( /* useUrl */ false, MyHandlerSetValueOne);
+			Assert.IsTrue (status == CGImageAnimationStatus.Ok, "status ok: second handler called with data");
```
mandel-macaque, line 136: Same
```suggestion
			Assert.AreEqual (CGImageAnimationStatus.Ok, status, "status ok: second handler called with data");
```

src/ImageIO/Enums.cs
spouliot: Apple headers can be confusing and are not consistent. Assume that if headers exists for a platform that:
* the API is available
* that it's available starting with this version of Xcode (and related SDK OS versions)
even if no availability macro for the platform is present in the headers.

Introspection, xtrospection and submission tests should be able to find mistakes.

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,143 @@
...
+#if IOS && ARCH_32
+            throw new PlatformNotSupportedException ();
+#else
+            if (url == null)
+                throw new ArgumentNullException (nameof (url));
+            if (handler == null)
+                throw new ArgumentNullException (nameof (handler));
+
+            BlockLiteral block_block;
+            block_block = new BlockLiteral ();
```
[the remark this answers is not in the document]
spouliot: iirc `handler` was named `block` in earlier commits

src/imageio.cs
```diff
@@ -2356,4 +2368,17 @@ interface CGImageAuxiliaryDataInfo {
 		NSData Data { get; set; }
 		NSDictionary DataDescription { get; set; }
 	}
+
+	[Partial]
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	interface CGImageAnimationOptions {
```
spouliot, line 2374: where is `CGImageAnimationOptions` used ?
It looks like it should be a strong dictionary and have overload in some new API

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,143 @@
...
+        [Introduced (PlatformName.WatchOS, 6, 0, PlatformArchitecture.All)]
+        [DllImport (Constants.ImageIOLibrary)]
+        static extern /* OSStatus */ CGImageAnimationStatus CGAnimateImageDataWithBlock ( /* CFDataRef _Nonnull */ IntPtr data, /* CFDictionaryRef _Nullable */ IntPtr options, /* CGImageSourceAnimationHandler _Nonnull */ ref BlockLiteral block);
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [Introduced (PlatformName.TvOS, 13, 0, PlatformArchitecture.All)]
+        [Introduced (PlatformName.WatchOS, 6, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.Optimizable)]
+        public CGImageAnimationStatus AnimateImage (NSUrl url, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationHandler handler)
```
spouliot, line 44: ^ that `NSDictionary` is likely mapped to the new options type
and that would be exposed as a strong dictionary

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Runtime.InteropServices;
+using CoreFoundation;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+
+namespace ImageIO
+{
+
+    public partial class CGImageAnimation
```
spouliot, line 20: This is a new type right ?

Since there's nothing for the instance then it makes no sense to have a way to instantiate it.

Make this a static type and all it's members static.

src/ImageIO/CGImageAnimation.cs
spouliot: Also since it seems to be a new type - which is not a `INativeObject` and does not require an `Handle` then this member should be `static`

src/ImageIO/CGImageAnimation.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+#endif
+        }
+
+        [Introduced (PlatformName.MacOSX, 10, 15, PlatformArchitecture.All)]
+        [Introduced (PlatformName.iOS, 13, 0, PlatformArchitecture.All)]
+        [Introduced (PlatformName.TvOS, 13, 0, PlatformArchitecture.All)]
+        [Introduced (PlatformName.WatchOS, 6, 0, PlatformArchitecture.All)]
+        [BindingImpl (BindingImplOptions.Optimizable)]
+        public CGImageAnimationStatus AnimateImage (NSData data, NSDictionary options, [BlockProxy (typeof (NIDCGImageSourceAnimationBlock))] CGImageSourceAnimationHandler handler)
```
spouliot, line 71: same, should be `static`

src/ImageIO/CGImageAnimation.cs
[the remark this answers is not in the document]
whitneyschmidt: fixed! good memory @spouliot 💯

tests/monotouch-test/ImageIO/CGImageAnimationTest.cs
```diff
@@ -0,0 +1,165 @@
...
+			imageUrl = NSBundle.MainBundle.GetUrlForResource ("hack", "gif");
+			imageData = NSData.FromUrl (imageUrl);
+			imageAnimation = new CGImageAnimation ();
+		}
+
+		[TestFixtureTearDown]
+		public void Cleanup ()
+		{
+			imageUrl.Dispose ();
+			imageData.Dispose ();
```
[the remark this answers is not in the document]
whitneyschmidt: Ah, good catch! Thank you :)

src/imageio.cs
whitneyschmidt: I don't see it used anywhere - the tests and sample pass in `null` for `NSDictionary options` and I don't see anything in the diffs. I can still rewrite it as a strong dictionary if you think that's the right thing to do.

spouliot: yes, we don't have to expose the `NSString` keys alone (whenever possible)
a strong dictionary allows us to hide the keys and make the API easier to use
header (or web) documentation might be helpful (@mandel-macaque already commented on them, and how they should look in a strong dictionary)

src/ImageIO/CGImageAnimation.cs
```diff
@@ -17,7 +17,7 @@
 namespace ImageIO
 {
 
-    public partial class CGImageAnimation
+    public static partial class CGImageAnimation // should remove partial? or leave in case we need to add more to this class in other files later?
```
spouliot, line 20: not a big deal... but whenever I see `partial` I must grep to see what else could be present
so **not** having it _by default_ can be helpful

src/imageio.cs
whitneyschmidt: Added `[StrongDictionary]`.

Based on the diff:

```
+/* The value of this key overrides the `delay time' specified by the image
+ * Value is a CFNumber of kCFNumberDoubleType.
+ */
+IMAGEIO_EXTERN const CFStringRef kCGImageAnimationDelayTime  IMAGEIO_AVAILABLE_STARTING(10.15, 13.0);
```

DelayTime is of type double:
`double DelayTime { get; set; }`

^ FYI @mandel-macaque

src/imageio.cs
```diff
@@ -2356,4 +2368,28 @@ interface CGImageAuxiliaryDataInfo {
...
+		[Field ("kCGImageAnimationStartIndex")]
+		NSString StartIndexKey { get; }
+	}
+
+	[Mac (10, 15), iOS (13, 0), TV (13, 0), Watch (6, 0)]
+	[StrongDictionary ("CGImageAnimationOptionsKeys")]
+	interface CGImageAnimationOptions {
+		double DelayTime { get; set; }
+
+		NSNumber LoopCount { get; set; }
```
dalexsoto, line 2391: I looked for code in GitHub that uses `kCGImageAnimationLoopCount` and it is used as follows

```objective-c

@property (nonatomic) NSUInteger loopCount;
...
options[(NSString *)kCGImageAnimationLoopCount] = @(self.loopCount);
```

```suggestion
		nuint LoopCount { get; set; }
```

From our [docs](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#strong-dictionary-types) `nuint -> NSUInteger stored in an NSNumber` so we will handle this for the user

src/imageio.cs
```diff
@@ -2356,4 +2368,28 @@ interface CGImageAuxiliaryDataInfo {
...
+
+		NSNumber StartIndex { get; set; }
```
dalexsoto, line 2393: same for `kCGImageAnimationLoopCount` it is used as follows

```objective-c

@property (nonatomic) NSUInteger beginningFrameIndex;
...
options[(NSString *) kCGImageAnimationStartIndex] = @(self.beginningFrameIndex);
```

```suggestion
		nuint StartIndex { get; set; }
```

src/imageio.cs
whitneyschmidt: done! :)

dalexsoto:
@whitneyschmidt one last request, could you replace `tests/monotouch-test/Resources/hack.gif` with this? I like the file but it is just too large I just made this quick gif for you 
color

spouliot:

PR #8377, opened 2020-04-14
[mmp] Explicitly resolve assemblies from the GAC / system mono.

rolfbjarne:
Cecil has a fall-back mode where it looks in the GAC / system mono for
assemblies when failing to find them elsewhere. This is not the expected
behavior when using Xamarin.Mac in the Full/XM mode, because then we should
only resolve to assemblies shipped with Xamarin.Mac.

Unfortunately doing so will break apps (our own tests break), so instead
change our resolution to be explicit about where we find assemblies, and if we
find assemblies in the GAC / system mono when we're not supposed to, then show
a warning.

Also add a fall-back mechanism, where we use the old logic instead, in case
the new logic is not 100% compatible with the old one.

This showed up when I tried to port mmp to dotnet, because then Cecil stopped
looking in the GAC / system mono for assemblies (Cecil has a special case when
running on Mono to look in Mono's GAC), and tests started failing.

### DIFF
--- a/tools/common/CoreResolver.cs
+++ b/tools/common/CoreResolver.cs
@@ -103,8 +103,7 @@ namespace Xamarin.Bundler {
 			catch (Exception e) {
 				throw new ProductException (9, true, e, Errors.MX0009, fileName);
 			}
-			CacheAssembly (assembly);
-			return assembly;
+			return CacheAssembly (assembly);
 		}
 
 		public AssemblyDefinition CacheAssembly (AssemblyDefinition assembly)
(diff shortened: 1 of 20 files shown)


tools/common/CoreResolver.cs
```diff
@@ -103,7 +103,13 @@ public virtual AssemblyDefinition Load (string fileName)
 			catch (Exception e) {
 				throw new ProductException (9, true, e, Errors.MX0009, fileName);
 			}
-			cache.Add (name, assembly);
+			CacheAssembly (assembly);
+			return assembly;
```
spouliot, line 107: minor: since you made `CacheAssembly` return the assembly definition you can do

```csharp
return CacheAssembly (assembly);
```

PR #8382, opened 2020-04-14
[generator] Handle (ignore) C#8 special nullability attributes

spouliot:
C# 8 nullability attributes are special (injected into assemblies) and
not meant to be used from C# source code.

We do not **use** them (we generated them) so existing attributes can
be ignored (filtered) by the generator.

Fix https://github.com/xamarin/xamarin-macios/issues/8347


spouliot:
build

spouliot:
build

spouliot:

PR #8387, opened 2020-04-15
[msbuild] Build bindings with the latest (stable) version of C#

spouliot:
We now require C# 8 for nullability support. However we allow custom code
to be included inside binding projects and we should not support anything
(stable) that the C# compiler (installed separately) allow, so `latest`
it is.


spouliot:
The failure (two mmp failures) are unrelated
* https://github.com/xamarin/xamarin-macios/issues/8399
* https://github.com/xamarin/xamarin-macios/issues/8400

spouliot:

PR #8391, opened 2020-04-15
[mtouch] Handle a failure to launch the native linker better by showing better error messages.

rolfbjarne:
dotnet will throw a Win32Exception if the command line is too long, so handle
that scenario. Also handle any other Win32Exceptions and show a better error
message.

This PR is best reviewed by [ignoring whitespace changes](https://github.com/xamarin/xamarin-macios/pull/8391/files?diff=unified&w=1).

### DIFF
--- a/tools/mtouch/BuildTasks.mtouch.cs
+++ b/tools/mtouch/BuildTasks.mtouch.cs
@@ -341,7 +341,7 @@ namespace Xamarin.Bundler
 							int arg_max;
 							if (int.TryParse (getconf_output.ToString ().Trim (' ', '\t', '\n', '\r'), out arg_max)) {
 								if (cmd_length > arg_max) {
-									linker_errors.Add (ErrorHelper.CreateWarning (5217, Errors.MT5217, cmd_length));
+									linker_errors.Add (ErrorHelper.CreateError (5217, Errors.MT5217, cmd_length));
 								} else {
 									Driver.Log (3, $"Linker failure is probably not due to command-line length (actual: {cmd_length} limit: {arg_max}");
 								}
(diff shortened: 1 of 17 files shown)


tools/mtouch/BuildTasks.mtouch.cs
```diff
@@ -307,50 +307,59 @@ protected override async Task ExecuteAsync ()
...
 						}
-					} else {
-						Driver.Log (3, "Failed to execute 'getconf ARG_MAX'\n{0}", getconf_output);
 					}
 				}
+			} catch (System.ComponentModel.Win32Exception wex) {
+				/* This means we failed to execute the linker, not that the linker itself returned with a failure */
+				if (wex.NativeErrorCode == 7 /* E2BIG = Too many arguments */ )
+					linker_errors.Add (ErrorHelper.CreateWarning (5217, wex, Errors.MT5217, cmd_length));
+				linker_errors.Add (ErrorHelper.CreateError (5222, wex, Errors.MX5222, wex.Message));
```
spouliot, line 360: do we need to emit two errors ?

rolfbjarne: You mean make 5217 an error as well (as opposed to a warning)? It currently says: `Native linking possibly failed because the linker command line was too long ({0} characters).`

spouliot: So I mis-read the `CreateWarning` :) likely because the comment (and message) are referring to an error :)

Still I would expect a single error as
* it might be the only thing (the error) reported in an issue;
* the warning only happens if things fails (it's an error condition); and 
* there's a single problem/error in the build (not two and I hate C++ compilers giving me tons of warnings/errors for a single, often unrelated, cause)

rolfbjarne:

PR #8394, opened 2020-04-15
[mmp/mtouch] Share more code.

rolfbjarne:
* Rearrange files in Xamarin.Mac a bit to ease code sharing between mmp and
  mtouch, by putting mono's static and dynamic libraries in
  /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/Sdks/Xamarin.macOS.sdk
  to match how Xamarin.iOS does it.

* Don't use 'usr' as an intermediate directory. This removes another special
  case.

* Share many of the functions and properties that return specific directories,
  and document (as comments) what each function/property is supposed to
  return.


rolfbjarne:
Test failures are unrelated:

* mmptest: https://github.com/xamarin/xamarin-macios/issues/8400 and https://github.com/xamarin/xamarin-macios/issues/8399

Issue #8395, opened 2020-04-15
-[NSScreen maximumExtendedDynamicRangeColorComponentValue]: unrecognized selector sent to instance 0x7ff453f09bf0


spouliot:
https://github.com/xamarin/xamarin-macios/commit/e6f64d7b3878598f6968dc43fcd633f777908c2b

all old (pre 10.15) macOS tests runs are failing with

```
[2020-04-15T07:05:43.795Z] 2020-04-15 03:05:43.615 apitest[48797:14f27] -[NSScreen maximumExtendedDynamicRangeColorComponentValue]: unrecognized selector sent to instance 0x7ff453f09bf0
[2020-04-15T07:05:43.795Z] 2020-04-15 03:05:43.618 apitest[48797:14f27] 
[2020-04-15T07:05:43.795Z] Unhandled Exception:
[2020-04-15T07:05:43.795Z] Foundation.ObjCException: NSInvalidArgumentException: -[NSScreen maximumExtendedDynamicRangeColorComponentValue]: unrecognized selector sent to instance 0x7ff453f09bf0
[2020-04-15T07:05:43.795Z]   at (wrapper managed-to-native) ObjCRuntime.Messaging.nfloat_objc_msgSend(intptr,intptr)
[2020-04-15T07:05:43.795Z]   at AppKit.NSScreen.get_MaximumExtendedDynamicRangeColorComponentValue () [0x0000c] in /Library/Frameworks/Xamarin.Mac.framework/Versions/6.21.0.45/src/Xamarin.Mac/AppKit/NSScreen.g.cs:379 
[2020-04-15T07:05:43.795Z]   at Xamarin.Mac.Tests.NSSCreenTests+<>c__DisplayClass10_0.<MaximumExtendedDynamicRangeColorComponentValueNoMainThread>b__0 () [0x00001] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/apitest/src/AppKit/NSScreen.cs:206 
[2020-04-15T07:05:43.795Z]   at System.Threading.ThreadHelper.ThreadStart_Context (System.Object state) [0x00014] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/referencesource/mscorlib/system/threading/thread.cs:74 
[...]

PR #8397, opened 2020-04-15
[AppKit] Ignore failing test on older OS X versions.

mandel-macaque:
The API should be present, yet we have an exception thrown when using
the selector. On 10.15 we do not have issues, so we skip in older
versions. Might be related to the fact that we use headless mac minis.

fixes: https://github.com/xamarin/xamarin-macios/issues/8395


mandel-macaque:
build

Issue #8408, opened 2020-04-16
[Tests] Mac tests fail with maximumReferenceExtendedDynamicRangeColorComponentValue


mandel-macaque:
This properties seem to just work on 10.15 when we are in a headless server. My strong feeling is that they change the implementation of the underlying OpenGL or other graphic API and in older versions of mac os x the selectors throw when we are using a virtual screen.

PR #8409, opened 2020-04-16
[Tests] Ignore failing tests in older versions of Mac OS X.

mandel-macaque:
Something must be different between the virtual screen created in a
headless system in older versions of Mac OS X and 10.15. Everything
works on 10.15 older versions throw an exception.

fixes: https://github.com/xamarin/xamarin-macios/issues/8408


mandel-macaque:
@spouliot I cannot test since I don't have and older mac :/ and in catalina everything works

rolfbjarne:
> I cannot test since I don't have and older mac :/ and in catalina everything works

You can push to origin to run everything on internal jenkins.

PR #8410, opened 2020-04-16
[mtouch/mmp] Simplify namespace handling now that we don't support Classic anymore.

rolfbjarne:
Also consolidate and sort the list of namespaces.


tools/linker/ObjCExtensions.cs
```diff
@@ -11,205 +8,80 @@
...
-		public static string PrintCore { get; private set; }
-		public static string ExternalAccessory { get; private set; }
-		public static string ImageCaptureCore { get; private set; }
-		public static string iTunesLibrary { get; private set; }
-#else
-		public static string Registrar { get; private set; }
-
-		public static string UIKit { get; private set; }
-#endif
+		public const string Accounts = "Accounts";
```
spouliot, line 11: might be better to use
```csharp
public const string Accounts = nameof (Accounts);
```
to protect us from typos.

PR #8411, opened 2020-04-16
Extract the ApplePlatform and LinkMode enums into their own files.

rolfbjarne:
I also stopped excluding the LinkMode.Platform value from XI, which makes it a bit easier to write shared code.


rolfbjarne:

PR #8417, opened 2020-04-17
[Foundation] Add an overload to NSString.FromHandle that takes an 'owns' parameter.

rolfbjarne:
This matches our API elsewhere.


rolfbjarne:

PR #8423, opened 2020-04-17
[Devops] If we fail to provision dependencies skip certain tasks.

mandel-macaque:
xamarin-storage can not be reachable or a network issue might happen
when we try to provision the dependencies. In that case, skip those
tasks that are skipped when we cannot provision the certs.

The provisioning dependencies will only execute in the provisioning
profiles was successful, therefore, the if statement is not stepping in
any value that was set by the profiles step.


mandel-macaque:
@whitneyschmidt 1000% agree, the only reason I did not add it is because I was going to ask you to do it ;)

mandel-macaque:
@rolfbjarne if provisioning profiles fails, provisioning dependencies wont be executed. 

The default condition of a step is "condition: succeeded()" as per the docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml

Since the provisoning step has the following:
```yml
- bash: |
      set -x
      set -e

      ./xamarin-macios/system-dependencies.sh --ignore-all --provision-xcode --provision-xamarin-studio --provision-mono --provision-7z
    displayName: 'Provision dependencies'
    timeoutInMinutes: 240
```
It means that the default condition applies. Therefore, if we fail in the previous step, this step is never executed. Since it is never executed the value of the env var is never chanced and left to be failure. Only if we set the condition of the provisioning dependencies to 'always()' or 'succeededOrFailed ()' will the env be stepped over, but that would be an issue in the step definition.

mandel-macaque:
@rolfbjarne removed

rolfbjarne:
@mandel-macaque `set -e` is there to detect problems when they occur (it's quite hard to figure out why something fails in a script when the script doesn't fail...). A much better solution would be to do handle the return error like this:

```shell
EC=0
./xamarin-macios/system-dependencies.sh --ignore-all --provision-xcode --provision-xamarin-studio --provision-mono --provision-7z || EC=$?
if [ $EC -eq 0 ]; then
    echo "##vso[task.setvariable variable=ProvisioningStatus]success"
else
    echo "##vso[task.setvariable variable=ProvisioningStatus]failure"
fi
exit $EC
```

tools/devops/device-tests-common.yml
```diff
@@ -58,26 +58,33 @@ jobs:
 
   - bash: |
       set -x
-      set -e
```
rolfbjarne, line 61: The `set -e` can return now :)

PR #8425, opened 2020-04-17
[mtouch] Always enable experimental-xforms-product-type

spouliot:
Not an experiment anymore - it works as expected.

This half-remove the optimization option (it must remain there to avoid
breaking all projects that have it defined) but it will always be `true`
so `Xamarin.Forms.Platform.iOS.dll` will **always** be considered as SDK
code by the linker.

Fix https://github.com/xamarin/xamarin-macios/issues/8407


spouliot:

spouliot:
Failure is a known error (HE0038) -> https://github.com/xamarin/maccore/issues/581

PR #8446, opened 2020-04-23
[Harness] Add support to create tunnels.

mandel-macaque:
Add support to create tunnels in case the devices cannot connect to
the host. This option is false by default, which means that unless told
otherwise xharness will try to se a tcp connection over the WiFi.

We are not setting it as default because the devices in DDFun will have
access to an unrestricted network. Nevertheless after this PR we will
create a new one with the following logic:

1. Try to use the tcp connection using the network.
2. If devices cannot connect to the host via the network, fall back to
the tcp tunnel.

This change executes a tunnel process per test application, most of the
cases out of the 150 test application we execute, most of them (maybe 98%
but most % are made up) will pass, and just a few of them will fail. The
reason is that there is a daemon in the OS that gets underwater.

Rather than tryingt o find a hacky way to re-use the tunnel, lets KISS
and if we need to hack that, do it as an enhancement.


tests/xharness/Harness.cs
```diff
@@ -219,6 +221,7 @@ public Harness (IResultParser resultParser, HarnessAction action, HarnessConfigu
 				SdkRoot = config ["XCODE_DEVELOPER_ROOT"] ?? configuration.SdkRoot;
 
 			processManager = new ProcessManager (XcodeRoot, MlaunchPath);
+			TunnelBore = new TunnelBore (processManager);
```
spouliot, line 224: what's the cost of creating this if `UseTcpTunnel` turns out to be `false` ?

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -602,6 +607,8 @@ Task<IEnumerable<AppleTestTask>> CreateRunDeviceTasksAsync ()
 						devices: devices,
 						buildTask: build64,
 						processManager: processManager,
+						tunnelBore: tunnelBore,
+						useTcpTunnel: Harness.UseTcpTunnel,
```
spouliot, line 611: maybe this could be merged ?

tests/xharness/TestTasks/RunDevice.cs
```diff
@@ -103,6 +106,7 @@ public async Task RunTestAsync ()
 						timeoutMultiplier: testTask.TimeoutMultiplier,
 						variation: testTask.Variation,
 						buildTask: testTask.BuildTask);
+					testTask.Runner.UseTcpTunnel = useTcpTunnel && testTask.Device.DevicePlatform == DevicePlatform.iOS || testTask.Device.DevicePlatform == DevicePlatform.tvOS;
```
spouliot, line 109: Looks like you're silently ignoring the requested use of the tunnel.
Would not it be better to warn/error earlier that the configuration does not make sense ?

tests/xharness/TestTasks/RunDevice.cs
```diff
@@ -175,6 +179,7 @@ public async Task RunTestAsync ()
 								timeoutMultiplier: testTask.TimeoutMultiplier,
 								variation: testTask.Variation,
 								buildTask: testTask.BuildTask);
+							todayRunner.UseTcpTunnel = useTcpTunnel && testTask.Device.DevicePlatform == DevicePlatform.iOS || testTask.Device.DevicePlatform == DevicePlatform.tvOS;
```
spouliot, line 182: same

tests/xharness/TestTasks/RunTest.cs
```diff
@@ -36,6 +39,7 @@ public class RunTest {
 			this.testTask = testTask ?? throw new ArgumentNullException (nameof (testTask));
 			this.BuildTask = buildTask ?? throw new ArgumentNullException (nameof (buildTask));
 			this.ProcessManager = processManager ?? throw new ArgumentNullException (nameof (processManager));
+			this.TunnelBore = tunnelBore; // can be null since some tests do not need tunnel
```
spouliot, line 42: there's a lot of other places where there's a `null` check...
I think this and `UseTcpTunnel` should be merged, e.g. `UseTcpTunnel` is only a getter that checks if the `TunnelBore` is non-null.

tests/xharness/Harness.cs
mandel-macaque: It is an empty dictionary. But we can simply not create it, and if it is null it wont be executed.

mandel-macaque:
@monojenkins backport xcode12

PR #8450, opened 2020-04-23
[Bump VMac] Bump max version of VSMac to current 8.6.

mandel-macaque:
Bump the vmac version to simplify dogfooding.


mandel-macaque:
That failure is from a testing pipeline. To be ignored.

PR #8455, opened 2020-04-24
[mtouch] Fix typo in simlauncher.mm.

rolfbjarne:


rolfbjarne:
> I guess `simlauncher` is only built in debug ?

Correct. This happens if you build a project for Simulator|Release, which it seems nobody does.

PR #8459, opened 2020-04-27
[Dependencies] Use brewfile to track the dependencies.

mandel-macaque:
brew bundle is a 'Bundler for non-Ruby dependencies from Homebrew,
Homebrew Cask and the Mac App Store.' which allows to keep the project
dependencies in a single file.

Dependencies can be tracked in the Brewfile, this allows new developers,
or those moving machines, to install all the project dependencies with:

```bash
brew bundle
```

Read docs with more information at https://github.com/Homebrew/homebrew-bundle


mandel-macaque:
@rolfbjarne wanted to talk about this in the weekly, I forgot to add it as a draft (was an example).

PR #8461, opened 2020-04-27
[mtouch/mmp] Build and run mtouch and mmp using dotnet.

rolfbjarne:
Compatibility: these tools now require dotnet 3.1 (or later) to be installed
before they'll work.

There are a few seemingly unrelated changes: these were required because some
of the tests would fail otherwise.

### DIFF
--- a/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/TestProject.cs
+++ b/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/TestProject.cs
@@ -16,6 +16,7 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared {
 		public string Name;
 		public bool IsExecutableProject;
 		public bool IsNUnitProject;
+		public bool IsDotNetProject;
 		public string [] Configurations;
 		public Func<Task> Dependency;
 		public string FailureMessage;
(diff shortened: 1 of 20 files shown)


rolfbjarne:
It might be possible to not require dotnet, by using the system mono and something like this in the bin script:

```shell
...
# Run with mono if we can find mono.
if type mono >/dev/null 2>&1; then
	echo Using mono
	exec mono $MONO_OPTIONS "$MONOTOUCH_PREFIX/lib/mtouch/mtouch.dll" "$@"
else
	echo Using dotnet
	exec "$MONOTOUCH_PREFIX/lib/mtouch/mtouch" "$@"
fi
```

spouliot:
@rolfbjarne it then becomes a testing issue, each runtime with a range of versions :(

Right now we still need mono for what ? `cil-strip` comes to mind

rolfbjarne:
build

akoeplinger:
Did you look at multi-targetting the tools and only using the netcore version for .net 5?

tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/TestProject.cs
```diff
@@ -88,7 +89,18 @@ public async Task CreateCopyAsync (ITestTask test = null)
 				doc.SetNode ("DocumentationFile", "bin\\$(Configuration)\\nunitlite.xml");
 			}
 			doc.ResolveAllPaths (original_path);
-
+			if (doc.IsDotNetProject ()) {
```
mandel-macaque, line 92: I'll have to move this to dotnet/xharness, created an issue not to forget: https://github.com/dotnet/xharness/issues/128

tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Utilities/ProjectFileExtensions.cs
```diff
@@ -195,7 +195,7 @@ public static void SetIntermediateOutputPath (this XmlDocument csproj, string va
 			// Make sure there's a top-level version too.
 			var property_group = csproj.SelectSingleNode("/*/*[local-name() = 'PropertyGroup' and not(@Condition)]");
 
-			var intermediateOutputPath = csproj.CreateElement ("IntermediateOutputPath", MSBuild_Namespace);
+			var intermediateOutputPath = csproj.CreateElement ("IntermediateOutputPath", csproj.GetNamespace ());
```
mandel-macaque, line 198: Same, issue: https://github.com/dotnet/xharness/issues/128

rolfbjarne:
Closing, this will probably be done a bit differently in the end.

PR #8472, opened 2020-04-29
Remove the LaunchProfiles capability for the CPS integration

kzu:
Implements https://work.azdo.io/1112733 as a workaround for the conflicts between
the built-in launchsettings.json-based .NET Core debugger and our Mono debugger.


msbuild/dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -28,6 +28,10 @@
 	<ItemGroup>
 		<ProjectCapability Include="Apple" />
 		<ProjectCapability Include="Mobile" />
+
+		// See https://work.azdo.io/1112733
```
spouliot, line 32: ^ as much as I like `//` that does not look legit here

PR #8483, opened 2020-04-30
Use the correct return type for the xamarin_get_delegate_for_block_parameter function.

rolfbjarne:
This code is old, and it seems the int* is a left-over from the 32-bit days.
Using the correct type makes us able to avoid a cast.

Ref: https://github.com/xamarin/maccore/commit/0f83cf16e2258f5f430e2672df199ee78d8f2c94


rolfbjarne:

PR #8485, opened 2020-04-30
[mtouch/mmp] When copying a directory hierarchy, delete the target file if it exists.

rolfbjarne:
The copyfile method does not like copying files over an existing file, and
sometimes it may end up showing an error.

### DIFF
--- a/tools/common/FileCopier.cs
+++ b/tools/common/FileCopier.cs
@@ -123,6 +123,13 @@ namespace Xamarin.Bundler {
 					else if (stage == CopyFileStep.Err) {
 						Log (1, "Could not copy the file '{0}' to '{1}'", source, target);
 						return CopyFileResult.Quit;
+					} else if (stage == CopyFileStep.Start) {
+						if (File.Exists (target) || Directory.Exists (target)) {
+							Log (1, "Deleted target {0}, it's not up-to-date", target);
+							// This callback won't be called for directories, but we can get here for symlinks to directories.
+							// This means that File.Delete should always work (no need to check for a directory to call Directory.Delete)
+							File.Delete (target);
+						}
 					}
 					return CopyFileResult.Continue;
 				} else {


tools/common/FileCopier.cs
```diff
@@ -123,6 +123,13 @@ static CopyFileResult CopyFileCallback (CopyFileWhat what, CopyFileStep stage, I
 					else if (stage == CopyFileStep.Err) {
 						Log (1, "Could not copy the file '{0}' to '{1}'", source, target);
 						return CopyFileResult.Quit;
+					} else if (stage == CopyFileStep.Start) {
+						if (File.Exists (target) || Directory.Exists (target)) {
+							Log (1, "Deleted target {0}, it's not up-to-date", target);
+							// This callback won't be called for directories, but we can get here for symlinks to directories.
+							// This means that File.Delete should always work (no need to check for a directory to call Directory.Delete)
+							File.Delete (target);
+						}
```
mandel-macaque, line 132: I don't thin that it the target is a dir the delete will work. I did a quick tests in the csharp RELP and I get the following exception:


```csharp
var dir = "Mydir"; 
Directory.CreateDirectory(dir); 
File.Delete(dir);
```     

Exception is thrown:
                                                                                                                                               

> System.UnauthorizedAccessException: Access to the path '/Users/mandel/Desktop/Mydir' is denied. ---> System.IO.IOException: Operation not permitted                                     --- End of inner exception stack trace ---                                                                                                                                          at System.IO.FileSystem.DeleteFile (System.String fullPath) [0x0006d] in <dd436d8b1aa7421eb3bcf48b995163c4>:0                                                                        at System.IO.File.Delete (System.String path) [0x00014] in <dd436d8b1aa7421eb3bcf48b995163c4>:0                                                                                      at <InteractiveExpressionClass>.Host (System.Object& $retval) [0x00000] in <7680ce88366e47b0aaff572764d5938d>:0                                                                      at Mono.CSharp.Evaluator.Evaluate (System.String input, System.Object& result, System.Boolean& result_set) [0x00038] in <6f83349cf15e4f8ab74a5f73daa81e78>:0                         at Mono.CSharpShell.Evaluate (System.String input) [0x00000] in <1a1d94369c5b4ba08b4121dc5d616b7f>:0

spouliot: My understanding, from the comment, is that it’s not called for directories.
But it’s called for a symlink - that can point to a directory and the symlink is actually a file (to delete).

mandel-macaque: Ok, in that case, I don't think we need the Exists at all. The File.Delete already handles if the file exists or not. Directory.Exist is just needed to confirm that is a symlink, but since File.Delete will do the right thing, I believe we can remove the if and simply do a File.Delete(target)

tools/common/FileCopier.cs
```diff
@@ -123,6 +123,13 @@ static CopyFileResult CopyFileCallback (CopyFileWhat what, CopyFileStep stage, I
 					else if (stage == CopyFileStep.Err) {
 						Log (1, "Could not copy the file '{0}' to '{1}'", source, target);
 						return CopyFileResult.Quit;
+					} else if (stage == CopyFileStep.Start) {
+						if (File.Exists (target) || Directory.Exists (target)) {
+							Log (1, "Deleted target {0}, it's not up-to-date", target);
+							// This callback won't be called for directories, but we can get here for symlinks to directories.
+							// This means that File.Delete should always work (no need to check for a directory to call Directory.Delete)
+							File.Delete (target);
```
mandel-macaque, line 131:
```suggestion
							File.Delete (target);
```

tools/common/FileCopier.cs
mandel-macaque, line 132:
```suggestion
						Log (1, "Deleted target {0}, it's not up-to-date", target);
						// This callback won't be called for directories, but we can get here for symlinks to directories.
						// This means that File.Delete should always work (no need to check for a directory to call Directory.Delete)
						File.Delete (target);
```

rolfbjarne: The logging statement can end up confusing now, because it will say it deleted something that may or may not have existed.

mandel-macaque: Is not a big issue, We can leave it as it is, is just that File.Exist + File.Delete might give threading issues, but those File.* methods always do :)

PR #8486, opened 2020-04-30
[xcode11.5] Initial bump to support Xcode 11.5 Beta 1

dalexsoto:
This includes changes to NetworkExtension and Intents. ExposureNotification is being done in another repository.

* https://github.com/xamarin/xamarin-macios/wiki/NetworkExtension-iOS-xcode11.5-beta1
* https://github.com/xamarin/xamarin-macios/wiki/Intents-iOS-xcode11.5-beta1


src/networkextension.cs
```diff
@@ -148,6 +148,12 @@ enum NEFilterReportFrequency : long {
 		High,
 	}
 
+	[NoWatch, NoTV, NoiOS, Mac (10,15,5)]
+	[Native]
+	enum NEFilterDataAttribute : long {
```
spouliot, line 153: hmmm... where should that be used ?

tests/xtro-sharpie/iOS-ExposureNotification.todo
```diff
@@ -0,0 +1,68 @@
+!missing-enum! ENAuthorizationStatus not bound
```
spouliot, line 1: might be easier to ask xtro Makefile to skip this framework (in case of future changes, new betas or Xcode)

src/networkextension.cs
spouliot: headers has some details

> 4-bytes NEFilterDataAttribute

so it should not be `[Native]` or `long` but Apple made it an `NSInteger` :(

dalexsoto: According to docs 

> @param readBytes The data that was read.  For non-UDP/TCP flows, since data may optionally include the IP header, readBytes includes a 4-bytes NEFilterDataAttribute field preceding the user data.  Handler must examine the NEFilterDataAttribute field and handle the data accordingly.

So it is a companion of these two https://github.com/xamarin/xamarin-macios/blob/8d3987b84a489e1ffd6fd825c1207b35f4a9fd32/src/networkextension.cs#L378-L382

tests/xtro-sharpie/iOS-ExposureNotification.todo
dalexsoto: I wonder if we should keep it as a 🦜 I am not sure if just the wiki is enough, OTOH it might be just a short lived framework

dalexsoto:
Fun, they all work locally :/ 

image

dalexsoto:
build

spouliot:
failure in last build is unrelated - it's HE0038 -> https://github.com/xamarin/maccore/issues/581

PR #8499, opened 2020-05-04
[linker] Make the corlib name dynamic.

rolfbjarne:
Use the correct corlib name depending on whether we're targeting .NET or Mono,
since there's no mscorlib.dll anymore in .NET, it's System.Private.CoreLib.dll.

### DIFF
--- a/tools/mtouch/xlf/Errors.pl.xlf
+++ b/tools/mtouch/xlf/Errors.pl.xlf
@@ -2986,6 +2986,14 @@
         <note>
 		</note>
       </trans-unit>
+      <trans-unit id="MX2111">
+        <source>Can not find the corlib assembly '{0}' in the list of loaded assemblies.
+		</source>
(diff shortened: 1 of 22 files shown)


tools/mtouch/xlf/Errors.pl.xlf
```diff
@@ -2986,6 +2986,14 @@
         <note>
 		</note>
       </trans-unit>
+      <trans-unit id="MX2111">
+        <source>Can not find the corlib assembly '{0}' in the list of loaded assemblies.
+		</source>
+        <target state="new">Can not find the corlib assembly '{0}' in the list of loaded assemblies.
+		</target>
+        <note>
+		</note>
```
mandel-macaque, line 2995: Nit: what is going on with the indentation?

rolfbjarne: No idea, this file is generated, and there are other entries with funny indentation as well.

PR #8505, opened 2020-05-04
[mtouch] Fix debug with profiler on app extensions

spouliot:
Always add `libmono-profiler-log.dylib` if profiling is enabled and we
are building with dynamic libraries. The profiler code is not (meant to
be) shipped so it can be added even without a `Frameworks` directory.

This fix debugging too (if profiler is enabled) since the library was
linked (even if it was not included).

Fix https://github.com/xamarin/xamarin-macios/issues/8470

### DIFF
--- a/tools/mtouch/Target.cs
+++ b/tools/mtouch/Target.cs
@@ -1570,8 +1570,7 @@ namespace Xamarin.Bundler
 				case AssemblyBuildTarget.DynamicLibrary:
 					libprofiler = Path.Combine (libmonodir, "libmono-profiler-log.dylib");
 					linker_flags.AddLinkWith (libprofiler);
-					if (App.HasFrameworksDirectory)
-						AddToBundle (libprofiler);
+					AddToBundle (libprofiler);
 					break;
 				case AssemblyBuildTarget.StaticObject:
 					libprofiler = Path.Combine (libmonodir, "libmono-profiler-log.a");


rolfbjarne:
I'm not sure this is the right fix, because the appex should be able to load the dylib from the container app. Does the container have have the dylib? Is profiling even enabled in the container?

spouliot:
@rolfbjarne 

> Does the container have have the dylib? 

no, but that's fine...

> Is profiling even enabled in the container?

no... seems to be a template issue (from a past conversation) but it could happen by mistake too.

The fix let you have different (profiling) settings for the app and the extension, so it works with the current solution unchanged (where only the extension has profiling enabled).

The other way would be to give an error because the settings are different. Still that's not identical to what we're doing (in most cases) since we _generally_ warn and disable code sharing (which does not help in this case).

rolfbjarne:
@spouliot 

> no... seems to be a template issue (from a past conversation) but it could happen by mistake too.

OK, the problem makes sense then (good!)

> The fix let you have different (profiling) settings for the app and the extension, so it works with the current solution unchanged (where only the extension has profiling enabled).

I think the right fix would be to copy the profiler dylib into the container if the container or any extension has profiling enabled, that way the profiler dylib is not duplicated anywhere (if multiple extensions have profiling enabled, but the container app doesn't).

BTW the `HasFrameworksDirectory` property is somewhat badly named: that's true for container apps and watchOS 2 extensions (because watchOS 2 extensions live on the watch and don't have a container app).

spouliot:
I have mixed emotions :) as it seems we're trying to optimize something that should not happen (and inside a situation that is not optimal by any other mean).

> that way the profiler dylib is not duplicated anywhere

In this case we're already in a _degraded_ mode where the application size is not optimal. 

```
/Users/poupou/Downloads/AppWithExtension/MTOUCH: Warning MT0113: Native code sharing has been disabled for the extension 'TodayTest' because the --assembly-build-target options are different between the container app () and the extension (--assembly-build-target:@all=dynamiclibrary). (MT0113) (AppWithExtension)
```

Also since apps don't ship (or there's another, larger problem) with profiling enabled then it's not a problem that would affect the final, release/submitted applications. _In fact the warning is way more troublesome for release, if it also happens on release builds_

The main issue, as I see it, is that debugging does not work (very bad) and the (existing) warning is not helpful to diagnose the issue. The warning is not bad, it's just not mean for this. Also since it's a warning people should expect things (like debug) to work.

rolfbjarne:
> as it seems we're trying to optimize something that should not happen

What is it that should not happen? Enabling profiling in app extensions and not the main app? I don't really see a problem with that.

> In this case we're already in a _degraded_ mode where the application size is not optimal.

Correct, but that doesn't mean there's no benefit to continue trying to keep the app size small, it affects build time and deploy time as well.

> The main issue, as I see it, is that debugging does not work (very bad)

Agreed

> and the (existing) warning is not helpful to diagnose the issue

I don't think it's worth it to add a warning when we can and should handle the case transparently for the user (by just putting the profiler dylib in the container app). IIRC we ran into this exact same some time ago (probably years ago), and for reasons I don't remember we didn't add a unit test.

The MT0113 warning was never meant for this particular case, but it's just another symptom of having different default values in the template for the container app and app extensions. Hopefully this will be all fixed with .NET where the default value will be no value at all in the csproj.

spouliot:
> What is it that should not happen?

_Optimizing_ for size in a case where we already warn it's not possible.

> trying to keep the app size small

If both container and (single) extension settings are different then adding it where defined does not change the size (just the location). 

If there are several extensions then the size impact of having different settings is much larger and already covered by the warning (and tests). Adding extra code in this situation makes it harder to test this properly (both individually and with other options). 

Still it can be done (it's the tests that will take most of the time) but I don't think its the a pressing (or common enough) issue to do this right now, considering the simple fix alternative. c.c. @chamons 

> I don't think it's worth it to add a warning

Oh, but I don't want to! Like I said the _warning is not bad_ and I said it was _not helpful_ because it (currently) breaks debugging (and that the real issue). You should not have to fix warnings for debug to work (unless it's mentioned, but that would not help here).

So once fixed it means my _not helpful_ comment is non applicable anymore, the existing warning does a good job and nothing more is needed :)

PR #8508, opened 2020-05-05
[mtouch] Detect the right build target for the partial static registrar.

rolfbjarne:


rolfbjarne:
Test failures are unrelated

PR #8511, opened 2020-05-05
[mtouch/mmp] Share Driver.GetFrameworks.

rolfbjarne:


tools/common/Frameworks.cs
```diff
@@ -341,7 +342,7 @@ public static Frameworks GetiOSFrameworks (Application app)
 	}
 
 	static Frameworks watch_frameworks;
-	public static Frameworks GetwatchOSFrameworks (Application app)
+	public static Frameworks GetwatchOSFrameworks (bool is_simulator_build)
```
mandel-macaque, line 345: Question, not that is a big deal, why not camelCase?

dalexsoto: Because of consistency and branding!! 

- iOS
- watchOS
- macOS
- tvOS

rolfbjarne: > Question, not that is a big deal, why not camelCase?

Because I can never remember which casing to use for parameters.

Maybe I'll remember for next time now :smile:

spouliot: if speaking of parameters then it should be camelCase as per FXDG
for the method name then @dalexsoto comment is right - looks weird in any case :)

rolfbjarne:
Test failures are unrelated

* monotouch-tests: https://github.com/xamarin/maccore/issues/581

PR #8512, opened 2020-05-05
[mtouch/mmp] Remove IStaticRegistrar, it's no longer needed.

rolfbjarne:
It was created to support Xamarin.Mac/Classic, but that's not supported
anymore, so we can just remove this interface.


mandel-macaque:
@rolfbjarne I DOS attack xharness with interfaces, while you remove them from other parts.. we keep it valanced ;)

rolfbjarne:
Test failure is unrelated

PR #8513, opened 2020-05-05
[BGTasks] Do not make BGTask abstract.

mandel-macaque:
The class in the API is abstract, the problem is that when the
application goes to the background, when it gets back we try to
instantiate an abstract class probably because Apple returns a internal
type that we do not know about.

fixes: https://github.com/xamarin/xamarin-macios/issues/7456

### DIFF
--- a/src/backgroundtasks.cs
+++ b/src/backgroundtasks.cs
@@ -48,7 +48,6 @@ namespace BackgroundTasks {
 		NSDate EarliestBeginDate { get; set; }
 	}
 
-	[Abstract]
 	[TV (13,0), NoWatch, NoMac, iOS (13,0)]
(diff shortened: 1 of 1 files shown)


mandel-macaque:
We probably want to back port this to at least 16-7.

rolfbjarne:
> We probably want to back port this to at least 16-7.

It would be nice to get someone to try a package to see if this fix actually fixes the problem.

mandel-macaque:
@rolfbjarne @spouliot we have a repo case, so I can take it for a spin before we do the backport.

mandel-macaque:
@monojenkins backport d16-7

mandel-macaque:
@monojenkins backport d16-6

PR #8515, opened 2020-05-05
[Harness] Refactor the selection of the tests out of Jenkins.cs

mandel-macaque:
Move the test selection logic out. Tests will land after we refactor the
Github class, which we depend on, to not be static, so that we can test
the different paths.

PS: Added a label to show that things do work, we just moved the logic to a diff file.

### DIFF
--- a/tests/xharness/Jenkins/TestSelector.cs
+++ b/tests/xharness/Jenkins/TestSelector.cs
@@ -21,59 +21,59 @@ namespace Xharness.Jenkins {
 		ILog MainLog => jenkins.MainLog;
 		Harness Harness => jenkins.Harness;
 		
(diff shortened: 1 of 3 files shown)


tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -0,0 +1,285 @@
...
+	public class TestSelector {
+
+		#region private vars
+		
+		readonly Jenkins jenkins;
+		readonly IProcessManager processManager;
+		ILog MainLog => jenkins.MainLog;
+		Harness Harness => jenkins.Harness;
+		
+			// We select tests based on a prefix of the modified files.
```
rolfbjarne, line 24: Your indentation is wrong somewhere.

PR #8523, opened 2020-05-06
Remove XAMARIN_APPLETLS, it seems like it's always defined, which means it's not needed.

rolfbjarne:
Also remove outdated comment about Xamarin.Mac/Classic, since XM/Classic is dead.


rolfbjarne:
Test failure is unrelated

PR #8525, opened 2020-05-06
[mtouch/mmp] Pass the path to mscorlib explicitly to the partial static registrar code instead of relying on resolving it successfully.

rolfbjarne:
This makes the code simpler when we have to add support for .NET.

This requires modifying the linker to accept a null FrameworkDirectory.


mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/2217

PR #8528, opened 2020-05-06
[msbuild] add System.Drawing.Common ref for bindings

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-macios/issues/8265

Usage of `System.Drawing` types in iOS binding projects are failing with:

    error CS1069: The type name 'Color' could not be found in the namespace 'System.Drawing'.
    This type has been forwarded to assembly 'System.Drawing.Common, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
    Consider adding a reference to that assembly.

In 3a7bdc0, an `_AddExtraReferences` MSBuild target was added for iOS
and Mac projects, but mistakenly missing from binding projects.

I moved the `_AddExtraReferences` target to `Xamarin.Shared.targets`,
so it will be used for all project types.

I also updated a test, I could merely use `System.Drawing.Color` in
`StructsAndEnums.cs` to reproduce the failure.

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
@@ -183,13 +183,6 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		<RemoveDir Directories="$(IntermediateOutputPath)" />
 	</Target>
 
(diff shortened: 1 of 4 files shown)


mandel-macaque:
Mtouch issue fixed by https://github.com/xamarin/xamarin-macios/pull/8531

PR #8530, opened 2020-05-06
[linker] Directly use Annotations in BaseStep subclasses

spouliot:
It's exposed directly in `BaseStep` and makes porting code easier to net5


mandel-macaque:
Failure not related already fixed by https://github.com/xamarin/xamarin-macios/pull/8531

PR #8531, opened 2020-05-06
[Harness] Do not use the tunnel bore in the mtouch tests.

mandel-macaque:
The change in xamarin/xamarin-macios/@617777ae73ccefdddd5711fe9628243b399492be
forgot to ignore the use of the tunnel in some cases. Is not the root of
the exception, but is the root of the tests failing after that commit.

fixes: https://github.com/xamarin/maccore/issues/2217


mandel-macaque:
MTouch tests did pass \o/

Issue #8533, opened 2020-05-06
[Harness] When reading the stderr from a log use the Reader over the FullPath.


mandel-macaque:
There are several types of logs, no all of the support the FUllPath property (Mainly, the ocnsole log and the memory log). All of the logs do support the GetReader methods, therefore it make sense to move the method that create failures to use the reader over the path to that way minimize the chances of getting a NotSupportedException.

PR #8534, opened 2020-05-06
[Harness] Use readers over paths.

mandel-macaque:
As issue https://github.com/xamarin/xamarin-macios/issues/8533 states,
not all loggers support the FullPath property, all of them do support
the GetReader one. We move the creation of the failures to use the
readers and reduce the chances of getting an exception.

fixes: https://github.com/xamarin/xamarin-macios/issues/8533

### DIFF
--- a/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs
+++ b/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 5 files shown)


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs
```diff
@@ -1,3 +1,4 @@
+#nullable enable
```
spouliot, line 1: ❤️

mandel-macaque: Step by step in xharness ;)

PR #8535, opened 2020-05-06
[linker] Update custom attributes that can be removed

spouliot:
Some are no longer part of the SDK (or converted into new ones
at build time), others were new (and missing).

A full list of attributes and their usage frequency in what we ship can
be seen in https://gist.github.com/spouliot/ca03c6da7d4d75670ca77749350eb8a2

Also update tests: no need to check for removals of stuff that does not
exists anymore.


mandel-macaque:
Issue not related, already fixed by https://github.com/xamarin/xamarin-macios/pull/8531

PR #8536, opened 2020-05-07
[mtouch] Make warning about not generating static registrar code for some simulator frameworks an actual (ignorable) warning.

rolfbjarne:
That way we can ignore this output when generating code for the partial static registrar code:

> The static registrar won't generate code for ... because its framework is not supported in the simulator.


rolfbjarne:
Looks like this will add warnings that will show up by default when using the static registrar in the simulator, so I don't think it's worth it.

PR #8541, opened 2020-05-07
[linker] Don't pass information to linker steps by selectively creating them or using constructors.

rolfbjarne:
Instead use the built-in logic to determine if a linker step should light up,
and use information available in the LinkContext to determine how steps should
behave.

This is required for .NET, where linker steps can't have custom constructors.

Several steps have not been modified, because they're not all required in .NET.


rolfbjarne:
Test failure is unrelated

* monotouch-test: https://github.com/xamarin/maccore/issues/1605

Issue #8344, opened 2020-04-10
No way to sign out of a NSUrlSession


dotMorten:
When making a request to a Basic auth service, there's no way to sign out.
Normally using a new NSUrlSession like the NSUrlSessionHandler creates should do this, but due to an Apple bug (RADAR #21403897), this doesn't actually happen.
To workaround that, you'd need to reset the session, but there's no way to get at that session.

### Steps to Reproduce
1.  Run the following code:
```cs
var handler1 = new NSUrlSessionHandler()
{
	Credentials = new System.Net.NetworkCredential("guest", "guest", "")
};
HttpClient client1 = new HttpClient(handler1);
var s1 = await client1.GetAsync("https://jigsaw.w3.org/HTTP/Basic/");
Debug.Assert(s1.StatusCode == System.Net.HttpStatusCode.OK);


// Workaround:
/*
var field = typeof(NSUrlSessionHandler).GetField("session", BindingFlags.NonPublic | BindingFlags.Instance);
if (field != null)
{
	var session = field.GetValue(handler1) as Foundation.NSUrlSession;
	if (session != null)
	{
                session.InvalidateAndCancel();
		session.Reset(() => { });
	}
}
*/

var handler2 = new NSUrlSessionHandler()
{
	Credentials = null
};
HttpClient client2 = new HttpClient(handler2);
var s2 = await client2.GetAsync("https://jigsaw.w3.org/HTTP/Basic/");
Debug.Assert(s1.StatusCode == System.Net.HttpStatusCode.Unauthorized); // Returns OK when 401 was expected
```

2. Notice that the second request returns OK without a credential, when it should return 401.
3. Uncomment the workaround that resets the session on the old handler, and the code now passes

### Expected Behavior
Second request should return 401.
Due to an Apple bug we must reset the session, but the used session isn't exposed on the NSUrlSessionHandler.

### Actual Behavior
Returns 200, reusing an old credential.

### Suggestion:
Add a constructor that take a session. That'll allow us to also use the shared session.
Secondary, provide a get-only property for the NSUrlSession.

### Environment
```
Microsoft Visual Studio Enterprise 2019
Version 16.5.3
VisualStudio.16.Release/16.5.3+30002.166
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Enterprise

Visual C++ 2019   00435-60000-00000-AA802
Microsoft Visual C++ 2019

.NET Portability Analyzer   1.1.10808.0
[...]

PR #8543, opened 2020-05-07
[Foundation] Do not get in an infinite loop with empty creds.

mandel-macaque:
As described in the issue, when the credentials for the base auth are set
to be empty the request reaches a timeout.

The inner issue is that when we have empty credentials the query that
is performed to the credentials db of the system throws an exception.
This exception is not raised to our code, but simply prints out in stderr.

When this situation occurs we get in an infinite loop in which we keep
sending the same credentials, we get another challenge, but the OS does
not do the right thing, we believe that we have to provide the creds
again (which are the same) and we get another challenge.. this goes on
an on until we reach the timeout.

Fortunately we know the number of failed challenges and the failure
response. With this information we can deduce if we already sent the
credentials, and if we did, do not set them again. In this case, the OS
sees no credentials in out delegate, uses the default handler and
correctly returns a 401.

fixes: https://github.com/xamarin/xamarin-macios/issues/8342
fixes: https://github.com/xamarin/xamarin-macios/issues/8344

The following attached app allows you to test the following:

1. Empty creds to the url provided by the user.
2. Test basic auth works (comment the first creds, uncomment the second ones, uncomment the request to httpbin.org)
3. Test basic auth works with the wrong creds (similar to 2 but uncomment the wrong creds).
[TestiOSHttpClient.zip](https://github.com/xamarin/xamarin-macios/files/4594666/TestiOSHttpClient.zip)

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -757,7 +757,23 @@ namespace Foundation {
 
 				if (sessionHandler.Credentials != null && TryGetAuthenticationType (challenge.ProtectionSpace, out string authType)) {
 					NetworkCredential credentialsToUse = null;
-					if (authType != RejectProtectionSpaceAuthType) {
+					// interesting situation, when we use a credential that we created that is empty, we are not getting the RejectProtectionSpaceAuthType,
+					// nevertheless, we need to check is not the first challange we will continue trusting the 
+					// TryGetAuthenticationType method, but we will also check that the status respose in no a 401
+					// look like we do get an exception from the credentials db:
+					//  TestiOSHttpClient[28769:26371051] CredStore - performQuery - Error copying matching creds.  Error=-25300, query={
+					// class = inet;
+					// "m_Limit" = "m_LimitAll";
+					// ptcl = htps;
+					// "r_Attributes" = 1;
+					// sdmn = test;
+					// srvr = "jigsaw.w3.org";
+					// sync = syna;
+					// }
+					// do remember that we ALWAYS get a challange, so the failure count has to be 1 or more for this check, 1 would be the first time
+					var nsurlRespose = challenge.FailureResponse as NSHttpUrlResponse;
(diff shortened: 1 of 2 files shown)


src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -757,7 +757,23 @@ public override void DidReceiveChallenge (NSUrlSession session, NSUrlSessionTask
...
+					// ptcl = htps;
+					// "r_Attributes" = 1;
+					// sdmn = test;
+					// srvr = "jigsaw.w3.org";
+					// sync = syna;
+					// }
+					// do remember that we ALWAYS get a challange, so the failure count has to be 1 or more for this check, 1 would be the first time
+					var nsurlRespose = challenge.FailureResponse as NSHttpUrlResponse;
+					var responseIsUnauthorized = (nsurlRespose == null) ? false : nsurlRespose.StatusCode == (int) HttpStatusCode.Unauthorized && challenge.PreviousFailureCount > 0;
+					if (authType != RejectProtectionSpaceAuthType && !responseIsUnauthorized) {
```
spouliot, line 776: nitpick: if `authType == RejectProtectionSpaceAuthType` then you're computing all this new stuff for nothing
might be better to nest the new code under the existing `if (authType != RejectProtectionSpaceAuthType) {`

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -757,7 +757,23 @@ public override void DidReceiveChallenge (NSUrlSession session, NSUrlSessionTask
 
 				if (sessionHandler.Credentials != null && TryGetAuthenticationType (challenge.ProtectionSpace, out string authType)) {
 					NetworkCredential credentialsToUse = null;
-					if (authType != RejectProtectionSpaceAuthType) {
+					// interesting situation, when we use a credential that we created that is empty, we are not getting the RejectProtectionSpaceAuthType,
+					// nevertheless, we need to check is not the first challange we will continue trusting the
```
whitneyschmidt, line 761: nit: `challenge` `status response`

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: good point

mandel-macaque:
@monojenkins backport d16-7

PR #8550, opened 2020-05-08
[ObjCRuntime/NSObject] Remove dead Classic code.

rolfbjarne:


src/ObjCRuntime/Runtime.cs
```diff
@@ -835,11 +829,6 @@ internal static ProtocolMemberAttribute GetProtocolMemberAttribute (Type type, s
 		// delegate
 		//
 		[EditorBrowsable (EditorBrowsableState.Never)]
-#if XAMCORE_2_0
-		internal
-#else
-		public 
-#endif
 		static MethodInfo GetBlockWrapperCreator (MethodInfo method, int parameter)
```
Therzok, line 832: Should be internal?

rolfbjarne: Looks like it doesn't need to be (anymore?), so I just made it private.

rolfbjarne:
build

mandel-macaque:
build

Issue #8350, opened 2020-04-11
Using HttpClient synchronously results in memory leak


prollin:
Calling `HttpClient.GetAsync(url).Result` in a loop causes process memory to grow continuously:

```
Task.Run(() =>
{
    var httpClient = new HttpClient();
    httpClient.BaseAddress = new Uri("https://www.google.com/", UriKind.Absolute);
    var requestUri = "images/branding/googlelogo/1x/googlelogo_color_272x92dp.png";
    while (true)
    {
        using (var httpResponseMessage = httpClient.GetAsync(requestUri).Result)
        {
            Console.WriteLine(httpResponseMessage.Content.Headers.ContentLength);
        }
        Task.Delay(10);
    }
});
```

Using async/await seem to solve the issue (but unless I missed something obvious, the synchronous version shouldn't leak):

```
Task.Run(async () =>
{
   var httpClient = new HttpClient();
   httpClient.BaseAddress = new Uri("https://www.google.com/", UriKind.Absolute);
   var requestUri = "images/branding/googlelogo/1x/googlelogo_color_272x92dp.png";
   while (true)
   {
        using (var httpResponseMessage = await httpClient.GetAsync(requestUri).ConfigureAwait(false))
        {
            Console.WriteLine(httpResponseMessage.Content.Headers.ContentLength);
        }
        await Task.Delay(10).ConfigureAwait(false);
    }
});
```

### Steps to Reproduce

1. Compile and deploy attached project to device
2. Launch app through the "Activity" instrument

### Expected Behavior

Process memory doesn't increase continuously

### Actual Behavior

Process memory increases continuously

### Environment

```
Microsoft Visual Studio Professional 2019
Version 16.5.3
VisualStudio.16.Release/16.5.3+30002.166
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Professional

Visual C++ 2019   00435-60000-00000-AA532
Microsoft Visual C++ 2019

.NET Memory Profiler   5.6
.NET Memory Profiler Visual Studio Integration Package

ADL Tools Service Provider   1.0
This package contains services used by Data Lake tools

ASA Service Provider   1.0

ASP.NET and Web Tools 2019   16.5.236.49856
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   16.5.236.49856
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.5.236.49856
Azure App Service Tools v3.0.0

[...]

PR #8555, opened 2020-05-09
[Foundation] Fix memory leak when using sync methods.

mandel-macaque:
The issue is a very interesting one. When using the HttpClient in a sync
way, the memory was leaking, on the async mode it was not. The root of
the leak is the following code:

```csharp
// EVIL HACK: having TrySetResult inline was blocking the request from completing
Task.Run (() => inflight.CompletionSource.TrySetResult (httpResponse));
```

The problem with the above line is that the lambda (a closure since it
captures the context) is leaking the captured variable making the memory
grow. The main problem is that as the // HACK comment mentions, the
TrySetResult locks. The reason for that is that it is contained inside
of a not needed lock.

The correct approach to this is to remove the lock which has no used
because we can use the tcs methods to know if the value was set
(IsCompleted does that). TrySetValue is thread safe and will return
without exceptions even if we could not set the response (either we had
an exception or the value was already set).

fixes: https://github.com/xamarin/xamarin-macios/issues/8350

### DIFF
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -675,9 +675,6 @@ namespace Foundation {
 				if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
 					return;
 
-				if (inflight.CompletionSource.Task.IsCompleted)
-					return;
-
 				inflight.CompletionSource.TrySetResult (inflight.Response);
 			}
 
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -483,7 +483,7 @@ namespace Foundation {
 			var nsrequest = await CreateRequest (request).ConfigureAwait(false);
 			var dataTask = session.CreateDataTask (nsrequest);
 
-			var tcs = new TaskCompletionSource<HttpResponseMessage> ();
+			var tcs = new TaskCompletionSource<HttpResponseMessage> (TaskCreationOptions.RunContinuationsAsynchronously);
 
 			lock (inflightRequestsLock) {
 #if !MONOMAC  && !__WATCHOS__
@@ -672,10 +672,12 @@ namespace Foundation {
 
 			void SetResponse (InflightData inflight)
 			{
-				if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
-					return;
+				lock (inflight.Lock) {
+					if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
+						return;
 
-				inflight.CompletionSource.TrySetResult (inflight.Response);
+					inflight.CompletionSource.TrySetResult (inflight.Response);
+				}
 			}
 
 			[Preserve (Conditional = true)]
--- a/src/Foundation/NSUrlSessionHandler.cs
+++ b/src/Foundation/NSUrlSessionHandler.cs
@@ -672,23 +672,13 @@ namespace Foundation {
 
 			void SetResponse (InflightData inflight)
 			{
-				lock (inflight.Lock) {
-					if (inflight.ResponseSent)
-						return;
+				if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
+					return;
 
-					if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
-						return;
+				if (inflight.CompletionSource.Task.IsCompleted)
+					return;
 
-					if (inflight.CompletionSource.Task.IsCompleted)
-						return;
-
-					var httpResponse = inflight.Response;
-
-					inflight.ResponseSent = true;
-
-					// EVIL HACK: having TrySetResult inline was blocking the request from completing
-					Task.Run (() => inflight.CompletionSource.TrySetResult (httpResponse));
-				}
+				inflight.CompletionSource.TrySetResult (inflight.Response);
 			}
 
 			[Preserve (Conditional = true)]
@@ -829,7 +819,6 @@ namespace Foundation {
 			public HttpResponseMessage Response { get; set; }
 
 			public Exception Exception { get; set; }
-			public bool ResponseSent { get; set; }
 			public bool Errored { get; set; }
 			public bool Disposed { get; set; }
 			public bool Completed { get; set; }


mandel-macaque:
Please read comment https://github.com/xamarin/xamarin-macios/issues/8350#issuecomment-626106244 since it has the profiling files that show that the leak is not longer an issue and the comparison between the handlers.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -672,23 +672,13 @@ public override void DidCompleteWithError (NSUrlSession session, NSUrlSessionTas
...
-
-					var httpResponse = inflight.Response;
+				if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
+					return;
 
-					inflight.ResponseSent = true;
+				if (inflight.CompletionSource.Task.IsCompleted)
+					return;
 
-					// EVIL HACK: having TrySetResult inline was blocking the request from completing
```
rolfbjarne, line 689: I don't understand why this hack was necessary in the first place, exactly what was blocking? If some other code is waiting for the CompletionSource to complete, and executes code when that happens, it's that code's responsibility to not block.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -672,23 +672,13 @@ public override void DidCompleteWithError (NSUrlSession session, NSUrlSessionTas
...
-					if (inflight.CompletionSource.Task.IsCompleted)
-						return;
-
-					var httpResponse = inflight.Response;
+				if (inflight.CancellationTokenSource.Token.IsCancellationRequested)
+					return;
 
-					inflight.ResponseSent = true;
+				if (inflight.CompletionSource.Task.IsCompleted)
+					return;
```
rolfbjarne, line 679: This is not needed, `TrySetResult` will check for this as well (in a thread-safe manner).

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -672,23 +672,13 @@ public override void DidCompleteWithError (NSUrlSession session, NSUrlSessionTas
...
 
-					// EVIL HACK: having TrySetResult inline was blocking the request from completing
-					Task.Run (() => inflight.CompletionSource.TrySetResult (httpResponse));
```
rolfbjarne, line 690: I don't understand why this is leaking. The closure will be collected by the GC when the Task.Run method completes.

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: I was not the person that wrote this. But in the tests, the lock + the TrySetResult blocks meaning that the request would never complete.

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: The closure should be GC, but it is not. Look at https://github.com/xamarin/xamarin-macios/issues/8350#issuecomment-626106244 profiling results. The leak is a memory 
stream from the httpcontent of the response. Removing the use of a lambda, fixes the issue.

The leak only happens when it is used in a sync method, so in async the lambda is correctly GC, in sync is not. I could not pin point the exact reason on why the GC is not doing the job correctly. 

The code was wrong nevertheless.

src/Foundation/NSUrlSessionHandler.cs
```diff
@@ -672,23 +672,10 @@ public override void DidCompleteWithError (NSUrlSession session, NSUrlSessionTas
 
 			void SetResponse (InflightData inflight)
 			{
-				lock (inflight.Lock) {
```
spouliot, line 675: We might not need the lock anymore. It's not clear the (only) other usage is needed.

src/Foundation/NSUrlSessionHandler.cs
spouliot: There has been a few other threading fixes in the past. It's possible that the _original_ condition does not happen anymore, but it's also possible it's just harder to reproduce. How do you intend to test this ?

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: I do agree, and I was tempted to remove it, but did not want to add a more complicated change over the one that fixed the leak.

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: It is already tested, all the tests in https://github.com/xamarin/xamarin-macios/blob/master/tests/monotouch-test/System.Net.Http/MessageHandlers.cs will timeout and fail if the call blocks. The blocking happens both on async and sync calls. There is no need to add a specific test for this.

spouliot: Our test suite has historically not been good to catch threading errors, like missing locks. I'm not so much afraid of the EVIL HACK removal than the rest.

mandel-macaque:
There was a long conversation about this. The root of the locking is similar to the following sample: https://gist.github.com/mandel-macaque/1b89e89b50044bf09701793a9f7e8967 but we fear that removing the lock might bring some issues. 

The TrySetValue locks because the continuations of the task completion source are usually inlined as with the cancelation subscriptions. Another way to remove the lock without using a Task.Run is to create the tcs with TaskCreationsOptions.RunContinuationsAsynchronously.

Not as good with memory, but safer.

mandel-macaque:
build

spouliot:
last failure (no report) seems to be another case of https://github.com/xamarin/xamarin-macios/issues/8569

src/Foundation/NSUrlSessionHandler.cs
mandel-macaque: We are keeping the lock, so no change there. The lock was added if there is more than one thread accessing the inflight data, I don't think it will happen. The task.run was added due to https://github.com/xamarin/xamarin-macios/pull/8555#issuecomment-628114963 but moving the token creation to be async fixes the lock and the memory leak (in managed objects).

spouliot: The change is less risky than it was. Still, with recent changes, I don't think we have the correct setup to test this properly (which goes beyond unit testing). c.c. @chamons

mandel-macaque:
Closing this is stale and the fix was added by a diff PR from @rolfbjarne

PR #8564, opened 2020-05-11
[mtouch/mmp] Share the mono native code.

rolfbjarne:
Also move the mono native code to the Application class from the Target
class, since it's per-Application, not per-Target.


rolfbjarne:
build

PR #8586, opened 2020-05-12
[Devops] Do link xcode to the expected path.

mandel-macaque:
Create symlink to the expected location. For this to work we need https://github.com/xamarin/provisionator/pull/305


mandel-macaque:
@rolfbjarne https://github.com/xamarin/provisionator/blob/master/Provisionator/IO/Symlink.cs#L76

PR #8591, opened 2020-05-13
Improve our error handling code.

rolfbjarne:
* Move much of ErrorHandler.cs into a partial class in ErrorHandler.tools.cs,
  which is referenced by mtouch and mmp (but not our runtime). This makes it
  possible to remove a lot of the ifdefs in the code.
* Add ErrorHandler.runtime.cs for runtime-specific bits, including a simpler
  version of ErrorHandler.Show. In particular this gets rid of the call to
  Environment.Exit, which should never happen at runtime.
* Rename MonoTouchException and MonoMacException to ProductException, which
  allows us to remove even more ifdefs.
* This required moving Application.LoadSymbols and Target.LoadSymbols to
  shared mtouch/mmp code.


rolfbjarne:

PR #8598, opened 2020-05-14
[tests] Fix computing the dependencies for the mtouch test project by including the makefile that contains the rule to compute the dependencies.

rolfbjarne:


rolfbjarne:
build

PR #8600, opened 2020-05-14
[dotnet] Implement 'dotnet run' for macOS apps.

rolfbjarne:


dotnet/Microsoft.macOS.Sdk/targets/Microsoft.macOS.Sdk.targets
```diff
@@ -1,4 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="Xamarin.Shared.Sdk.targets" />
+
+  <PropertyGroup>
+    <RunCommand>open</RunCommand>
+    <RunArguments>$(OutputPath)/$(AssemblyName).app</RunArguments>
+  </PropertyGroup>
```
spouliot, line 8: how does it map to `dotnet run` ? default action ?

rolfbjarne: @spouliot The `dotnet run` command does some magic (restores packages, builds the app, etc), and then runs the "Run" target:

https://github.com/dotnet/sdk/blob/47ba0d486cd742d6d6a5e40ba174cbec1bfbd839/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L755-L758

Issue #8211, opened 2020-03-27
[loc][Query][Xamarin_MaciOS] Localizability 


PDostalek:
Hi,
please help localization team to answer question from translator.
In case translator is not sure about context/meaning/localizability of string we raise a question.

File Name Live Page Extended | String Resource ID | Full Source String | Query | Answer Needed
-- | -- | -- | -- | --
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0; "E0044" | Could not find a valid Xcode app   bundle at '{0}'. Please update your Apple SDK location in Visual Studio's   preferences {1}.\r\n | Please provide the meaning of   placeholder {1}. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0;   "E0094" | Failed   to load {0} log file `{1}`: {2}\r\n | Please,   define placeholders. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0; "E0096" | Unknown target framework   identifier: {0}.\r\n | What is unknown, the platform or   the identifier? |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0;   "E0098" | {0}   failed.\r\n | What   does the placeholder stand for? Please explain. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0; "E0124" | The identity '{0}' doesn't match   any valid certificate/private key pair in the default keychain\r\n | Does "keychain" refer   to the Apple keychain here? We assumed so for now, please advise. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0;   "E0128" | No   valid {0} code signing keys found in keychain. You need to request a   codesigning certificate from https://developer.apple.com.\r\n | What   does the placeholder stand for? Please explain. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0; "E0130" | {0} code signing key '{1}' not   found in keychain.\r\n | Please, define placeholders. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0;   "E0131" | Could   not find any available provisioning profiles for {0} on {1}.\r\n | What   does the placeholder {1} stand for? We assumed a computer or alike for now,   please advise. |  
Xamarin.Localization.MSBuild\xlf\MSBStrings.xlf | 0; "E0133" | No installed provisioning   profiles match the installed {0} {1} signing identities.\r\n | Please, define placeholders. |  
[...]

PR #8602, opened 2020-05-15
[msbuild] add missing localization comments

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-macios/issues/8211
Context: https://github.com/xamarin/xamarin-android/blob/master/Documentation/workflow/Localization.md

A list of MSBuild error codes came back from the translators. These
had no `<comment/>` fields filled out at all. I added these, so it
should be much clearer what the messages actually mean.

* E0044
* E0094
* E0098
* E0124
* E0128
* E0130
* E0131
* E0133
* E0140
* E0156
* E7022_A
* E7057
* M0119
* W0020
* W0022

Error codes that did not appear to be used anymore:

* E0096

I also added a short `README.md` so others can more easily pick this
up.

### DIFF
--- a/msbuild/Xamarin.Localization.MSBuild/MSBStrings.Designer.cs
+++ b/msbuild/Xamarin.Localization.MSBuild/MSBStrings.Designer.cs
@@ -503,12 +503,6 @@ namespace Xamarin.Localization.MSBuild {
             }
         }
         
-        public static string E0096 {
-            get {
-                return ResourceManager.GetString("E0096", resourceCulture);
-            }
-        }
-        
         public static string E0097 {
             get {
                 return ResourceManager.GetString("E0097", resourceCulture);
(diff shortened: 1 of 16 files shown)


rolfbjarne:
FWIW this is the reason there were no empty comment fields: https://github.com/xamarin/xamarin-macios/pull/8157

jonathanpeppers:
@rolfbjarne I think a lot of the messages need actual comments, so the translators have a better chance of knowing what they mean.

I don't think the empty `<comment/>` tag matters, I was comparing some messages:

https://github.com/xamarin/xamarin-macios/blob/ff3d5cbde6341b8b25da7d83c5f9acfd6d59056f/msbuild/Xamarin.Localization.MSBuild/MSBStrings.resx#L430-L433

vs:
https://github.com/xamarin/xamarin-android/blob/42822e0488185cdf4bca7c0bd05b21ad03dfbd7e/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx#L120-L142

👍  I'll just try to add comments as we get feedback like: https://github.com/xamarin/xamarin-macios/issues/8211

wachaudh:
@jonathanpeppers the majority of messages have already been translated. The translators have reached out a few times for clarification about a couple messages. I had just been responding to them directly, not adding comment tags.

PR #8610, opened 2020-05-19
[src] Use an intermediate script file for the compiler passed to bgen.

rolfbjarne:
That way we can pass additional flags to the compiler, without those flags
interferring with the flags to bgen.


rolfbjarne:
Test failure is known: https://github.com/xamarin/maccore/issues/2219

PR #8615, opened 2020-05-19
[dotnet] Switch to net5.0 instead of netcoreapp5.0.

rolfbjarne:
This also requires bumping .NET to a version that supports 'net5.0'.

The actual TFM is still '.NETCoreApp,Version=5.0', it's just the short name
that has changed.


peppy:
Is this in the latest preview releases of `Xamarin.iOS`? Trying to get around https://github.com/dotnet/runtime/issues/35988 but I'm still hitting the same error as reported in that thread.

rolfbjarne:
@peppy no, Xamarin.iOS won't move to the .NET core runtime until .NET 6, which is next fall.

peppy:
Aha I see. That includes preview releases too?

rolfbjarne:
@peppy extremely early previews are already available (see https://github.com/xamarin/net6-samples), but that may require significant changes to your project, and there are very many rough edges still.

PR #8616, opened 2020-05-19
[Harness] Move GetColor method out of Jenkins.

mandel-macaque:
Reduce the logic that is not related to the execution of tests out of
jenkins. Add tests.


mandel-macaque:
@rolfbjarne yes, that is the plan, but I'm doing it step by step. More like, get it out, add tests, move. I want to move all the html generation to its own class and then have those extension methods there.

PR #8628, opened 2020-05-20
[src] Put the implementation assemblies for our product assemblies in the NuGet runtime packages.

rolfbjarne:
This is the current structure:

        Microsoft.iOS.Runtime.<rid>
        └─── data
        │    └─── RuntimeList.xml
        └─── runtimes
        │    └─── <rid>
        │    │    └─── lib
        │    │    │    └─── net5.0
        │    │    │    │    └─── Xamarin.iOS.dll
        │    │    │    │    └─── Xamarin.iOS.pdb

and likewise for tvOS, watchOS and macOS.

Also fix the nuget packaging to include the reference/implementation
assemblies in FrameworkList.xml and RuntimeList.xmls.

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -453,6 +453,15 @@ else
 DOTNET5=$(abspath $(TOP)/builds/downloads/dotnet/$(DOTNET5_VERSION))/dotnet
 endif
 
+DOTNET_IOS_RUNTIME_IDENTIFIERS=ios-arm ios-arm64 ios-x86 ios-x64
+DOTNET_TVOS_RUNTIME_IDENTIFIERS=tvos-arm64 tvos-x64
+DOTNET_WATCHOS_RUNTIME_IDENTIFIERS=watchos-arm watchos-x86
+DOTNET_MACOS_RUNTIME_IDENTIFIERS=osx-x64
+
+DOTNET_IOS_RUNTIME_IDENTIFIERS_32=ios-arm ios-x86
+DOTNET_IOS_RUNTIME_IDENTIFIERS_64=ios-arm64 ios-x64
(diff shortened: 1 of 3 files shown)


Make.config
```diff
@@ -453,6 +453,15 @@ else
 DOTNET5=$(abspath $(TOP)/builds/downloads/dotnet/$(DOTNET5_VERSION))/dotnet
 endif
 
+DOTNET_IOS_RUNTIME_IDENTIFIERS=ios-arm ios-arm64 ios-x86 ios-x64
+DOTNET_TVOS_RUNTIME_IDENTIFIERS=tvos-arm64 tvos-x64
+DOTNET_WATCHOS_RUNTIME_IDENTIFIERS=watchos-arm watchos-x86
```
spouliot, line 458: what about `arm64_32` ?

Make.config
```diff
@@ -453,6 +453,15 @@ else
...
+DOTNET_MACOS_RUNTIME_IDENTIFIERS=osx-x64
+
+DOTNET_IOS_RUNTIME_IDENTIFIERS_32=ios-arm ios-x86
+DOTNET_IOS_RUNTIME_IDENTIFIERS_64=ios-arm64 ios-x64
+DOTNET_WATCHOS_RUNTIME_IDENTIFEIRS_32=watchos-arm watchos-x86
```
spouliot, line 463: typo `IDENTIFEIRS`

Make.config
rolfbjarne: The runtime identifiers for watchOS are still unclear, but my guess is that `watchos-arm` will contain both `armv7k` and `arm64_32` (like `ios-arm` contains both `armv7` and `armv7s`).

PR #8629, opened 2020-05-20
[Make] Allow to disable the compilation of the sample tests.

mandel-macaque:
The are not usually ran by developers, allow them not to build them
since building them means that we download all the samples that results
in around 5Gb of code that they will never be compiled or used.


rolfbjarne:
Wouldn't it be simpler to just not run `make` in `tests/sampletester`? We don't recurse into that directory if `make` is run in any directories higher up.

configure
```diff
@@ -3,35 +3,38 @@
...
-    --disable-packaged-mono Enable/disable using the precompiled version of mono. If disabled mono will be compiled from source.
+    --disable-packaged-mono   Enable/disable using the precompiled version of mono. If disabled mono will be compiled from source.
 
-    --enable-dotnet         Enable building .NET 5 bits.
-    --disable-dotnet        Disable building .NET 5 bits.
+    --enable-dotnet           Enable building .NET 5 bits.
+    --disable-dotnet          Disable building .NET 5 bits.
+
+    --enable-sample-testing
+    --disable-sample-testing  Enable/disable the sample tests. If disable the tests are not compiled or run. Disabling saves HDD space.
```
spouliot, line 37: don't reformat everything in a PR that introduce changes (do it in a separate PR if needed)

mandel-macaque: Is the doc string, adds no logical changes.

mandel-macaque:
@rolfbjarne yes, I don't mind either way. But we have to make sure the pipeline is updated correctly. That is why I added it as an option.

rolfbjarne:
@mandel-macaque we *already* don't recurse into the `tests/sampletester`, so I have no idea how you ran into this problem (unless you manually ran `make` yourself in this directory).

configure
spouliot: Exactly, but hat's not the reason for doing separate PR. They are different, unrelated changes and one might need the be reverted, without requiring the other part to be removed.

mandel-macaque:
@rolfbjarne no idea, just noticed. Close until I see it again.

PR #8632, opened 2020-05-20
[linker] Remove code to deal with ExportedTypes

spouliot:
This is now supported by upstream mono/linker


mandel-macaque:

Issue #8625, opened 2020-05-19
[Harness] Report connection issues in the markdown report.


mandel-macaque:
Add the connection issues that are known in the markdown report rather than set them as crashes. This seems to be misleading to the monitoring person.

PR #8633, opened 2020-05-20
[Harness] When we have a known issue, add it to the markdown.

mandel-macaque:
If we have a known issue, lets add the message and the link to the
markdown so that the monitoring person does not need to look at the
html page.

fixes: https://github.com/xamarin/xamarin-macios/issues/8625

### DIFF
--- a/tests/xharness/Jenkins/ErrorKnowledgeBase.cs
+++ b/tests/xharness/Jenkins/ErrorKnowledgeBase.cs
@@ -7,15 +7,16 @@ using Microsoft.DotNet.XHarness.iOS.Shared.Logging;
 
 namespace Xharness.Jenkins {
 	public class ErrorKnowledgeBase : IErrorKnowledgeBase {
-		static readonly Dictionary<string, string> testErrorMaps = new Dictionary<string, string> {
-			["error HE0038: Failed to launch the app"] = "<a href='https://github.com/xamarin/maccore/issues/581'>HE0038</a>",
+		static readonly Dictionary<string, (string HumanMessage, string IssueLink)> testErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)> {
+			["error HE0038: Failed to launch the app"] = (HumanMessage: "HE0038", IssueLink: "https://github.com/xamarin/maccore/issues/581)"),
+			["Couldn't establish a TCP connection with any of the hostnames"] = (HumanMessage: "Tcp Connection Error: Tests are reported as crashes when they succeeded.", IssueLink: "https://github.com/xamarin/maccore/issues/1741"),
 		};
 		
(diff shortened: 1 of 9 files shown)


tests/xharness/Jenkins/TestData.cs
```diff
@@ -9,7 +9,7 @@ class TestData
 		public string Variation;
 		public string MTouchExtraArgs;
 		public string MonoBundlingExtraArgs; // mmp
-		public string KnownFailure;
+		public (string HumanMessage, string IssueLink)? KnownFailure;
```
whitneyschmidt, line 12: How does `?` work here?

mandel-macaque: Named tuples are value types, not reference types, we add the ? to allow the value to be null, same as with an int.

spouliot: c#8 nullability, it means it can be null (top of file has `#nullable enable`)

mandel-macaque: @spouliot not this one ;) 

But I should add it.

mandel-macaque: @whitneyschmidt for ref, a named tuple is under the hood a https://docs.microsoft.com/en-us/dotnet/api/system.valuetuple?view=netcore-3.1 that is a struct, structs can't be null, so I added the ? to do a nullable one.

tests/xharness/Jenkins/ErrorKnowledgeBase.cs
```diff
@@ -7,15 +7,16 @@
 
 namespace Xharness.Jenkins {
 	public class ErrorKnowledgeBase : IErrorKnowledgeBase {
-		static readonly Dictionary<string, string> testErrorMaps = new Dictionary<string, string> {
-			["error HE0038: Failed to launch the app"] = "<a href='https://github.com/xamarin/maccore/issues/581'>HE0038</a>",
+		static readonly Dictionary<string, (string HumanMessage, string IssueLink)> testErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)> {
+			["error HE0038: Failed to launch the app"] = (HumanMessage: "HE0038", IssueLink: "https://github.com/xamarin/maccore/issues/581)"),
+			["Couldn't establish a TCP connection with any of the hostnames"] = (HumanMessage: "Tcp Connection Error: Tests are reported as crashes when they succeeded.", IssueLink: "https://github.com/xamarin/maccore/issues/1741"),
```
rolfbjarne, line 12: _may_ have succeeded (we don't know if they succeeded or not)

tests/xharness/Jenkins/TestData.cs
spouliot: when you think too much about something then you see it everywhere ;-)

PR #8641, opened 2020-05-21
Fix old-style provisioning by using a non-composed variable for XCODE_URL.

rolfbjarne:
Fixes this:

    Checking system...
        Found OSX 10.15.3 (at least 10.15 is required)
            Downloading Xcode 11.5 from http://xamarin-storage/bot-provisioning/xcodes/$(XCODE_XIP_NAME) to /tmp/x-provisioning...


rolfbjarne:
Merging right away since the build is broken anyway, this way we'll get it tested more faster.

PR #8650, opened 2020-05-21
[corefoundation] Inline Throw in NativeObject.GetCheckedHandle

spouliot:
`Throw` is not virtual and unlikely to be inlined (since it throws)
and is (small) additional metadata for no good reason.

### DIFF
--- a/src/CoreFoundation/NativeObject.cs
+++ b/src/CoreFoundation/NativeObject.cs
@@ -12,7 +12,6 @@ using System;
 using System.Runtime.InteropServices;
 using ObjCRuntime;
 using Foundation;
-using CoreFoundation;
 
 namespace CoreFoundation {
 	//
@@ -78,12 +77,10 @@ namespace CoreFoundation {
 			this.handle = handle;
 		}
 
-		void Throw () => throw new ObjectDisposedException (GetType ().ToString ());
-
 		internal IntPtr GetCheckedHandle ()
 		{
 			if (handle == IntPtr.Zero)
-				Throw ();
+				throw new ObjectDisposedException (GetType ().ToString ());
 			return handle;
 		}
 	}


src/CoreFoundation/NativeObject.cs
```diff
@@ -78,12 +77,10 @@ protected virtual void InitializeHandle (IntPtr handle)
 			this.handle = handle;
 		}
 
-		void Throw () => throw new ObjectDisposedException (GetType ().ToString ());
-
 		internal IntPtr GetCheckedHandle ()
 		{
 			if (handle == IntPtr.Zero)
-				Throw ();
+				throw new ObjectDisposedException (GetType ().ToString ());
```
Therzok, line 83: Yeah, but this prevents GetCheckedHandle from getting inlined.

Therzok: It's a common pattern (in dotnet/runtime) to have helper functions for throwing methods, so the containing methods have more chances to be inlined (and the cost of calling a new method is just in the scenario where the actual exception should be thrown).

mandel-macaque: Thats actually the only reason they do it. I'm with @Therzok on this one.

Therzok: This is why things like [ThrowHelper](https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Linq/src/System/Linq/ThrowHelper.cs#L10-L59) exist.

spouliot: heh, I already have a local `ThrowHelper` for something else ;-)

It's a performance vs size tradeoff (one much easier to measure than the other), which has not until recently been considered by dotnet/runtime (and they should now be considered as equals).

spouliot:
Closing. I'll look at sharing this (along other things) to minimize the size without disabling inlining of the `GetCheckedHandle` method.

src/CoreFoundation/NativeObject.cs
Therzok: Perf vs size has been considered by dotnet/runtime before, but possibly not in this specific scenario.

This looks like a generic enough method to be reused via a ThrowHelper though.

See speed vs size opt here: https://github.com/dotnet/runtime/tree/master/src/libraries/System.Linq/src/System/Linq

Therzok:
On another miscellaneous note, reusing throw helpers does decrease total IL size :D

src/CoreFoundation/NativeObject.cs
spouliot: I'd argue that is was a fix after the fact, not a consideration when the feature coding was done.

Now there's nothing wrong about putting performance first, it made a lot of sense for the original scenarios that netcore targeted. This is also true for mono. There's a lot of code, even in the Xamarin years, where performance alone (without size) was considered.

Expending the scenarios to include mobile means that code examples (like your linke) will become more common (at build time or thru the linker).

Ideally new code will consider both perf/size and not assume existing code did the same :)

PR #8660, opened 2020-05-24
[Harness] Move the resource manager implementation out of Jenkins.

mandel-macaque:
Move the resource management out of Jenkins. Add tests to the new class.

### DIFF
--- a/tests/xharness/Xharness.Tests/Jenkins/ResourceManagerTests.cs
+++ b/tests/xharness/Xharness.Tests/Jenkins/ResourceManagerTests.cs
@@ -1,0 +1,126 @@
+﻿using System.Collections.Generic;
+using System.Resources;
+using Microsoft.DotNet.XHarness.iOS.Shared.Hardware;
+using Xharness.Jenkins;
+using Moq;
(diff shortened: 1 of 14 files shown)


tests/xharness/Xharness.Tests/Jenkins/ResourceManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+			device2.Setup (d => d.UDID).Returns ("device2");
+			device2.Setup (d => d.Name).Returns ("device2");
+
+			var resources = resourceManager.GetDeviceResources (devices);
+
+			// assert that both are present
+			Assert.NotNull (resources, "resources");
+			Assert.AreEqual (2, resources.Count ());
+			Assert.AreEqual (1, resources.Where (r => r.Name == "device1").Count ());
+			Assert.AreEqual (1, resources.Where (r => r.Name == "device2").Count ());
```
spouliot, line 42: you should label each one individually
in case debug symbols (or runtime) is broken and point to the incorrect line (or no line at all)

PR #8667, opened 2020-05-25
[corefoundation] Safer NativeObject

spouliot:
Do not allow calling `CFRetain` or `CFRelease` on a null handle as this
will crash the process, instead use `GetCheckedHandle` so a managed (and
catchable) `ObjectDisposedException` can be thrown.

Make `GetCheckedHandle` public so subclasses (outside the platform
assembly) can use it - it's already recommended in some `[Obsolete]`
attributes (but was not made available).

Move the `ObjectDisposedException` thrower into a new `ThrowHelper`
class. This will allow sharing it (to be re-used in future PR).

Add unit tests for `NativeObject`.

### DIFF
--- a/src/CoreFoundation/NativeObject.cs
+++ b/src/CoreFoundation/NativeObject.cs
@@ -85,7 +85,7 @@ namespace CoreFoundation {
 		public IntPtr GetCheckedHandle ()
 		{
 			if (handle == IntPtr.Zero)
-				ObjCRuntime.ThrowHelper.ThrowObjectDisposedException (GetType ());
+				ObjCRuntime.ThrowHelper.ThrowObjectDisposedException (this);
 			return handle;
 		}
 	}
--- a/src/CoreFoundation/NativeObject.cs
+++ b/src/CoreFoundation/NativeObject.cs
@@ -6,13 +6,12 @@
 //   Alex Soto
 //   Miguel de Icaza
 //
-// Copyrigh 2018 Microsoft Inc
+// Copyright 2018, 2020 Microsoft Corp
 //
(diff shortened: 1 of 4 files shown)


spouliot:
build

spouliot:
Error seems unrelated (no `NativeObject`) but related to `Dispose` so I'll let the bots double check first (and file if unrelated).

src/CoreFoundation/NativeObject.cs
```diff
@@ -78,12 +82,10 @@ protected virtual void InitializeHandle (IntPtr handle)
...
 		}
 
-		void Throw () => throw new ObjectDisposedException (GetType ().ToString ());
-
-		internal IntPtr GetCheckedHandle ()
+		public IntPtr GetCheckedHandle ()
 		{
 			if (handle == IntPtr.Zero)
-				Throw ();
+				ObjCRuntime.ThrowHelper.ThrowObjectDisposedException (GetType ());
```
rolfbjarne, line 88: Would it be better to pass `this` to `ThrowObjectDisposedException` instead of calling `GetType ()`? That would avoid a method call (and a virtual one), which would make the IL shorter and thus more likely to be inlined by the AOT compiler. `ThrowObjectDisposedException` could do `t?.GetType ()?.ToString ()` on the input, and the end result would be the same.

spouliot: Good idea! The (virtual) call will still be made under the same condition (if `handle == `IntPtr.Zero`), but the caller's IL will be shorter (and there will be more callers than callees).

PR #8671, opened 2020-05-26
[ObjCRuntime] Use _NSConcreteStackBlock instead of _NSStackBlock.

rolfbjarne:
_NSConcreteStackBlock is documented in http://clang.llvm.org/docs/Block-ABI-Apple.html.

It's not an actual Objective-C class, so we have to use dlsym to find its pointer.


rolfbjarne:
@monojenkins backport d16-7
@monojenkins backport d16-6

PR #8679, opened 2020-05-26
[objcruntime] Add helper to check for disposed instances

spouliot:
We're handling `ObjectDisposedException` a bit inconsistently in our
bindings.

Some, but not all, manual bindings have checks (not all of them are fully
consistent) but there's none for generated bindings, which can lead to
errors wrt nullability.

**Example**

```
[Test]
public void IncompleteNullabilityCheck ()
{
	NSString s = new NSString ("bonjour");
	s.Dispose ();
	ManagedLayer (s);
}

void ManagedLayer (NSString s)
{
	// this is similar to the generated binding code
	if (s == null)
		throw new ArgumentNullException ("s");
	NativeCode (s.Handle);
}

void NativeCode (IntPtr p)
{
	// let's assume this is native and dereference the pointer
	if (p == IntPtr.Zero)
		Assert.Fail ("boo");
}
```

This shows that we can, _if disposed_, provide `nil` to a native API for
which we know `nil` is not a valid argument. That can crash a process in
a situation that was possible to detect and throw a (catchable) managed
exception.

Adding dispose checks everywhere could be costly (in size) unless we
share that code with the null check (and a few other optimizations could
be applied too). Sharing the code would also ensure more consistency
across all bindings.

**Impact**

Here's the IL for an existing `CGColor` constructor that *already*
does the dispose check.

```
.method public hidebysig specialname rtspecialname
	instance void .ctor (
		class CoreGraphics.CGColorSpace colorspace,
		valuetype System.nfloat[] components
	) cil managed
{
	// Method begins at RVA 0x20c834
	// Code size 82 (0x52)
	.maxstack 3

	IL_0000: ldarg.0
	IL_0001: call instance void [mscorlib]System.Object::.ctor()
	IL_0006: ldarg.2
	IL_0007: brtrue.s IL_0014

	IL_0009: ldstr "components"
	IL_000e: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_0013: throw

	IL_0014: ldarg.1
	IL_0015: brtrue.s IL_0022

	IL_0017: ldstr "colorspace"
	IL_001c: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_0021: throw

	IL_0022: ldarg.1
	IL_0023: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_0028: ldsfld native int [mscorlib]System.IntPtr::Zero
	IL_002d: call bool [mscorlib]System.IntPtr::op_Equality(native int, native int)
	IL_0032: brfalse.s IL_003f

	IL_0034: ldstr "colorspace"
	IL_0039: newobj instance void [mscorlib]System.ObjectDisposedException::.ctor(string)
	IL_003e: throw

	IL_003f: ldarg.0
	IL_0040: ldarg.1
	IL_0041: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_0046: ldarg.2
	IL_0047: call native int CoreGraphics.CGColor::CGColorCreate(native int, valuetype System.nfloat[])
	IL_004c: stfld native int CoreGraphics.CGColor::handle
	IL_0051: ret
} // end of method CGColor::.ctor

Here's the IL code (as committed) using a helper extension method.

```
.method public hidebysig specialname rtspecialname
	instance void .ctor (
		class CoreGraphics.CGColorSpace colorspace,
		valuetype System.nfloat[] components
	) cil managed
{
	// Method begins at RVA 0x210558
	// Code size 46 (0x2e)
	.maxstack 3
	.locals init (
		[0] native int handleof_colorspace
	)

	IL_0000: ldarg.0
	IL_0001: call instance void [mscorlib]System.Object::.ctor()
	IL_0006: ldarg.2
	IL_0007: brtrue.s IL_0014

	IL_0009: ldstr "components"
	IL_000e: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_0013: throw

	IL_0014: ldarg.1
	IL_0015: ldstr "colorspace"
	IL_001a: call native int ObjCRuntime.NativeObjectHelper::GetNonNullHandle(class ObjCRuntime.INativeObject, string)
	IL_001f: stloc.0
	IL_0020: ldarg.0
	IL_0021: ldloc.0
	IL_0022: ldarg.2
	IL_0023: call native int CoreGraphics.CGColor::CGColorCreate(native int, valuetype System.nfloat[])
	IL_0028: stfld native int CoreGraphics.CGColor::handle
	IL_002d: ret
} // end of method CGColor::.ctor
```

So existing checks becomes a lot smaller. Here's another example where
a dispose check is **missing** (on `pattern`). Note that the check for
`colorspace` is not updated so we can see the impact of adding missing
dispose checks in existing code (manual or generated).

```
.method public hidebysig specialname rtspecialname
	instance void .ctor (
		class CoreGraphics.CGColorSpace colorspace,
		class CoreGraphics.CGPattern pattern,
		valuetype System.nfloat[] components
	) cil managed
{
	// Method begins at RVA 0x210660
	// Code size 126 (0x7e)
	.maxstack 4

	IL_0000: ldarg.0
	IL_0001: call instance void [mscorlib]System.Object::.ctor()
	IL_0006: ldarg.1
	IL_0007: brtrue.s IL_0014

	IL_0009: ldstr "colorspace"
	IL_000e: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_0013: throw

	IL_0014: ldarg.1
	IL_0015: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_001a: ldsfld native int [mscorlib]System.IntPtr::Zero
	IL_001f: call bool [mscorlib]System.IntPtr::op_Equality(native int, native int)
	IL_0024: brfalse.s IL_0031

	IL_0026: ldstr "colorspace"
	IL_002b: newobj instance void [mscorlib]System.ObjectDisposedException::.ctor(string)
	IL_0030: throw

	IL_0031: ldarg.2
	IL_0032: brtrue.s IL_003f

	IL_0034: ldstr "pattern"
	IL_0039: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_003e: throw

	IL_003f: ldarg.3
	IL_0040: brtrue.s IL_004d

	IL_0042: ldstr "components"
	IL_0047: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_004c: throw

	IL_004d: ldarg.0
	IL_004e: ldarg.1
	IL_004f: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_0054: ldarg.2
	IL_0055: callvirt instance native int CoreFoundation.NativeObject::get_Handle()
	IL_005a: ldarg.3
	IL_005b: call native int CoreGraphics.CGColor::CGColorCreateWithPattern(native int, native int, valuetype System.nfloat[])
	IL_0060: stfld native int CoreGraphics.CGColor::handle
	IL_0065: ldarg.0
	IL_0066: ldfld native int CoreGraphics.CGColor::handle
	IL_006b: ldsfld native int [mscorlib]System.IntPtr::Zero
	IL_0070: call bool [mscorlib]System.IntPtr::op_Equality(native int, native int)
	IL_0075: brfalse.s IL_007d

	IL_0077: newobj instance void [mscorlib]System.ArgumentException::.ctor()
	IL_007c: throw

	IL_007d: ret
} // end of method CGColor::.ctor
```

into

```
.method public hidebysig specialname rtspecialname
	instance void .ctor (
		class CoreGraphics.CGColorSpace colorspace,
		class CoreGraphics.CGPattern pattern,
		valuetype System.nfloat[] components
	) cil managed
{
	// Method begins at RVA 0x210660
	// Code size 119 (0x77)
	.maxstack 4
	.locals init (
		[0] native int handleof_pattern
	)

	IL_0000: ldarg.0
	IL_0001: call instance void [mscorlib]System.Object::.ctor()
	IL_0006: ldarg.1
	IL_0007: brtrue.s IL_0014

	IL_0009: ldstr "colorspace"
	IL_000e: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_0013: throw

	IL_0014: ldarg.1
	IL_0015: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_001a: ldsfld native int [mscorlib]System.IntPtr::Zero
	IL_001f: call bool [mscorlib]System.IntPtr::op_Equality(native int, native int)
	IL_0024: brfalse.s IL_0031

	IL_0026: ldstr "colorspace"
	IL_002b: newobj instance void [mscorlib]System.ObjectDisposedException::.ctor(string)
	IL_0030: throw

	IL_0031: ldarg.3
	IL_0032: brtrue.s IL_003f

	IL_0034: ldstr "components"
	IL_0039: newobj instance void [mscorlib]System.ArgumentNullException::.ctor(string)
	IL_003e: throw

	IL_003f: ldarg.2
	IL_0040: ldstr "pattern"
	IL_0045: call native int ObjCRuntime.NativeObjectHelper::GetNonNullHandle(class ObjCRuntime.INativeObject, string)
	IL_004a: stloc.0
	IL_004b: ldarg.0
	IL_004c: ldarg.1
	IL_004d: ldfld native int CoreGraphics.CGColorSpace::handle
	IL_0052: ldloc.0
	IL_0053: ldarg.3
	IL_0054: call native int CoreGraphics.CGColor::CGColorCreateWithPattern(native int, native int, valuetype System.nfloat[])
	IL_0059: stfld native int CoreGraphics.CGColor::handle
	IL_005e: ldarg.0
	IL_005f: ldfld native int CoreGraphics.CGColor::handle
	IL_0064: ldsfld native int [mscorlib]System.IntPtr::Zero
	IL_0069: call bool [mscorlib]System.IntPtr::op_Equality(native int, native int)
	IL_006e: brfalse.s IL_0076

	IL_0070: newobj instance void [mscorlib]System.ArgumentException::.ctor()
	IL_0075: throw

	IL_0076: ret
} // end of method CGColor::.ctor
```

so the difference is smaller, but it's still smaller (7 bytes) than the
existing code (that did not check for a disposed instance).

There are many more optimizations that can be applied on top of this.
However the main/original goal is to have a correct and consistent
handling of disposed managed instances. IOW the size reduction is a nice
side-effect of correctness and actual optimizations can come later :)

The logical next step is to update the generator to use the helper
whenever it checks for `null`...

### DIFF
--- a/src/CoreGraphics/CGColor.cs
+++ b/src/CoreGraphics/CGColor.cs
@@ -86,12 +86,9 @@ namespace CoreGraphics {
 		{
 			if (components == null)
 				throw new ArgumentNullException ("components");
-			if (colorspace == null)
-				throw new ArgumentNullException ("colorspace");
-			if (colorspace.handle == IntPtr.Zero)
-				throw new ObjectDisposedException ("colorspace");
+			var handleof_colorspace = colorspace.GetNonNullHandle ("colorspace");
 			
-			handle = CGColorCreate (colorspace.handle, components);
+			handle = CGColorCreate (handleof_colorspace, components);
 		}
 
 #if !XAMCORE_3_0 || MONOMAC
@@ -142,12 +139,11 @@ namespace CoreGraphics {
 				throw new ArgumentNullException ("colorspace");
 			if (colorspace.handle == IntPtr.Zero)
 				throw new ObjectDisposedException ("colorspace");
-			if (pattern == null)
-				throw new ArgumentNullException ("pattern");
 			if (components == null)
 				throw new ArgumentNullException ("components");
+			var handleof_pattern = pattern.GetNonNullHandle ("pattern");
 
-			handle = CGColorCreateWithPattern (colorspace.handle, pattern.Handle, components);
+			handle = CGColorCreateWithPattern (colorspace.handle, handleof_pattern, components);
 			if (handle == IntPtr.Zero)
 				throw new ArgumentException ();
 		}
--- a/src/ObjCRuntime/INativeObject.cs
+++ b/src/ObjCRuntime/INativeObject.cs
@@ -20,6 +20,15 @@ namespace ObjCRuntime {
 		{
 			return self == null ? IntPtr.Zero : self.Handle;
 		}
+
+		static public IntPtr GetNonNullHandle (this INativeObject self, string argumentName)
+		{
+			if (self == null)
+				ThrowHelper.ThrowArgumentNullException (argumentName);
+			if (self.Handle == IntPtr.Zero)
+				ThrowHelper.ThrowObjectDisposedException (self);
+			return self.Handle;
+		}
 	}
 #endif
 }
--- a/src/ObjCRuntime/ThrowHelper.cs
+++ b/src/ObjCRuntime/ThrowHelper.cs
@@ -9,6 +9,12 @@ namespace ObjCRuntime {
 	static class ThrowHelper {
 
 		[DoesNotReturn]
+		static internal void ThrowArgumentNullException (string argumentName)
+		{
+			throw new ArgumentNullException (argumentName);
+		}
+
+		[DoesNotReturn]
 		static internal void ThrowObjectDisposedException (object o)
 		{
 			throw new ObjectDisposedException (o.GetType ().ToString ());

### CODE REVIEW

**Comment by Therzok** [comment]
> The logical next step is to update the generator to use the helper
whenever it checks for null...

Having ArgumentNullException also reuse a helper method would also reduce the size more. This looks great!

**Comment by spouliot** [comment]
> Having ArgumentNullException also reuse a helper method would also reduce the size more

hmm... in what context ?

This PR adds `ThrowHelper.ThrowArgumentNullException` (meant to be used in other places) and `GetNonNullHandle` already calls it.

At _term_ I'd like all (manual and generated) bindings to use

* `GetNonNullHandle`, for `INativeObject` instance - unless `null` is allowed (and that's where `GetHandle` is useful for readability); and
* `ThrowHelper.ThrowArgumentNullException` for reference types that are not `INativeObject`, e.g. `System.String`, arrays...

File: src/CoreGraphics/CGColor.cs [hunk 1]
```diff
@@ -142,12 +139,11 @@ public CGColor (CGColorSpace colorspace, CGPattern pattern, nfloat [] components
 				throw new ArgumentNullException ("colorspace");
 			if (colorspace.handle == IntPtr.Zero)
 				throw new ObjectDisposedException ("colorspace");
-			if (pattern == null)
-				throw new ArgumentNullException ("pattern");
 			if (components == null)
 				throw new ArgumentNullException ("components");
+			var handleof_pattern = pattern.GetNonNullHandle ("pattern");
```
rolfbjarne, line 144:
```suggestion
			var handleof_pattern = pattern.GetNonNullHandle (nameof (pattern));
```

src/CoreGraphics/CGColor.cs
```diff
@@ -86,12 +86,9 @@ public CGColor (CGColorSpace colorspace, nfloat [] components)
 		{
 			if (components == null)
 				throw new ArgumentNullException ("components");
-			if (colorspace == null)
-				throw new ArgumentNullException ("colorspace");
-			if (colorspace.handle == IntPtr.Zero)
-				throw new ObjectDisposedException ("colorspace");
+			var handleof_colorspace = colorspace.GetNonNullHandle ("colorspace");
```
rolfbjarne, line 89:
```suggestion
			var handleof_colorspace = colorspace.GetNonNullHandle (nameof (colorspace));
```

src/CoreGraphics/CGColor.cs
rolfbjarne: I think `pattern_handle` sounds slightly better than `handleof_pattern`, but it's not a big deal.

spouliot: yeah, it's not meant to be committed as-is, e.g. `nameof`

for `handleof_` prefix it's more for generated code (maybe with a few extra `_` prefixed) as the `*_handle` has an higher probability of already existing and clashing. Or maybe it will end up with `__handle_p1` for the handle of argument 1...

Funnily it's named after `nameof`, hence the lack of `_` inside it ;-)

src/ObjCRuntime/INativeObject.cs
```diff
@@ -20,6 +20,15 @@ static public IntPtr GetHandle (this INativeObject self)
 		{
 			return self == null ? IntPtr.Zero : self.Handle;
 		}
+
+		static public IntPtr GetNonNullHandle (this INativeObject self, string argumentName)
+		{
+			if (self == null)
+				ThrowHelper.ThrowArgumentNullException (argumentName);
```
stephen-hawley, line 27: Separate item, but I've found the pattern of `Ex.ThrowOnNull` extremely useful in BTfS:
```
public static T ThrowOnNull<T> (T o, string name, string message = null) where T : class
{
	if (o == null)
		throw new ArgumentNullException (name ?? "<no name supplied>",
					message ?? "");
	return o;
}
```
Maybe be want to consider using it in other places too.

src/ObjCRuntime/INativeObject.cs
```diff
@@ -20,6 +20,15 @@ static public IntPtr GetHandle (this INativeObject self)
...
+			if (self.Handle == IntPtr.Zero)
+				ThrowHelper.ThrowObjectDisposedException (self);
```
stephen-hawley, line 29: We have an argument name - I know it's an existing API, but shouldn't we put that into the exception to make chasing down an error easier?

src/ObjCRuntime/INativeObject.cs
spouliot: This is the pattern used by the dotnet/BCL to minimize size, some details here https://referencesource.microsoft.com/#mscorlib/system/throwhelper.cs

Like @Therzok reminded me in https://github.com/xamarin/xamarin-macios/pull/8650 your pattern would not inline. Also optional arguments so add a cost to callers. Not a big deal for a tool, but that's meant to be used extensively in generated code.

I still need to measure the code size difference (IL and AOT) before moving the PR out of it's draft status. But I need to complete the generator changes first :)

spouliot:
The generated diff is huge. Don't ask Jenkins to display it. Instead download the zip archive (3.9MB)
```
mbp:generator-reference poupou$ l x.diff
-rw-r--r--  1 poupou  staff  56820778  4 Jun 10:14 x.diff
mbp:generator-reference poupou$ wc -l x.diff
  837548 x.diff
```
[x.diff.zip](https://github.com/xamarin/xamarin-macios/files/4730603/x.diff.zip)

As for the size impact, it's about 1% less than before. Reminder that this PR about correctness not a size optimization (even if it can lead to several ones).

# products-assemblies

| File | Before | After | Diff | % |
|------|-------:|------:|-----:|--:|
| Xamarin.iOS (32bits) | 17,012,736 | 16,881,664 | 131,072 | 99.2% |
| Xamarin.iOS (64bits) | 18,406,400 | 18,219,520 | 186,880 | 99.0% |
| Xamarin.TVOS | 12,486,144 | 12,367,872 | 118,272 | 99.1% |
| Xamarin.WatchOS | 7,883,776 | 7,814,656 | 69,120 | 99.1% |
| Xamarin.Mac (full) | 22,087,680 | 21,876,224 | 211,456 | 99.0% |
| Xamarin.Mac (mobile) | 22,090,752 | 21,879,296 | 211,456 | 99.0% |

spouliot:
Size impact on *dont link.app* is about 0.4% less
https://gist.github.com/spouliot/89a5f770d6b398eb87222c2830c07317

Q: Why are `Xamarin.iOS.dll` a bit bigger, and not smaller, than before ?
A: Because release builds strips IL (which we produce less with this PR) and we added new helper API (metadata)

src/ObjCRuntime/ThrowHelper.cs
```diff
@@ -6,10 +6,16 @@
 
 namespace ObjCRuntime {
 
-	static class ThrowHelper {
+	public static class ThrowHelper {
```
rolfbjarne, line 9: Maybe hide this class from intellisense?

```suggestion
	[EditorBrowsable (EditorBrowsableState.Never)]
	public static class ThrowHelper {
```

spouliot:
@rolfbjarne I was able to replicate the failure when I builded/installed `xcode12.2` then ran this test from `main`.
While not identical (its not what happens on bots) it shows that the test is not using the latest build platform assemblies.

rolfbjarne:
> @rolfbjarne I was able to replicate the failure when I builded/installed `xcode12.2` then ran this test from `main`.

I just tried that, but the tests still pass for me 😒 

I'm assuming you're using different directories for each branch?

> While not identical (its not what happens on bots) it shows that the test is not using the latest build platform assemblies.

I created a PR to show the compiler command line if it fails to compile: https://github.com/xamarin/xamarin-macios/pull/9998

spouliot:
> I'm assuming you're using different directories for each branch?

Yes, Both branches have their own directory.

> I created a PR to show the compiler command line if it fails to compile: #9998

Great, I'll update this PR once your gets merged!

rolfbjarne:
build

spouliot:
I guess the generator diff is _corrupting_ the assembly (with the previous one) so it can't be used to test ?

rolfbjarne:
> I guess the generator diff is _corrupting_ the assembly (with the previous one) so it can't be used to test ?

Yeah, that's my theory (and that shouldn't happen). I'll have to investigate a bit.

rolfbjarne:
> I'll have to investigate a bit.

Should be fixed with https://github.com/xamarin/xamarin-macios/pull/10011.

PR #8630, opened 2020-05-20
page through github statuses for manifest URL

cadsit:
Fixes (as seen by [CI](https://dev.azure.com/xamarin/internal/_build/results?buildId=18702&view=logs&j=307a3e1d-f0f2-5a3d-98c5-8c16e3f24b09&t=1ea22bc9-7602-5454-032c-4e9789b26d98)):
```
Unhandled exception. System.Exception: Could not find the manifest for 89d621e37355592d071ce92b9f58e7bcbdf322f0. Is the commit already built by CI?
```

There are so many statuses on the commit that the `manifest` status gets pushed back to a different page.

PR #8680, opened 2020-05-26
satisfy the type-checking gods

cadsit:
I did some last minute cleanup on `utils.csx` in #8630 that I thought was harmless, but wasn't :(


rolfbjarne:
@monojenkins backport d16-7

PR #8683, opened 2020-05-26
[.gitignore] Ignore the json that is created on a device crash on xharness.

mandel-macaque:
Those files can be found when we test crashes locally and should not be
added to the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ tests/bcl-test/SystemXunit.csproj
 global.json
 .idea
 device-tests-provisioning.csx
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -33,3 +33,4 @@ tests/bcl-test/SystemXunit.csproj
 global.json
 .idea
 device-tests-provisioning.csx
+mono_crash.*.json
```
rolfbjarne, line 36: Screenshot 2020-05-27 08 50 33

PR #8689, opened 2020-05-27
[xharness] Always capture stdout/stderr from the simulator to actual files.

rolfbjarne:
This avoids confusion when tests are run locally using 'make runner' (and
would print the simulator's stdout/stderr to the terminal) versus when run on
bots (and would write the simulator's stdout/stderr to separate files that
would show up in the html report).


rolfbjarne:
Test failure is unrelated: https://github.com/xamarin/maccore/issues/581)

PR #8690, opened 2020-05-27
[xharness] Add crash reports to a logs collection we care about.

rolfbjarne:
Add crash reports to a logs collection we care about, instead of to a logs
collection that's promptly forgotten.

This makes sure crash reports actually show up in the html report.


rolfbjarne:
Test failure is unrelated:

* monotouch-test: https://github.com/xamarin/maccore/issues/2219

rolfbjarne:
@monojenkins backport xcode12

Issue #8695, opened 2020-05-27
[Harness] Markdown comment does not use the correct test name.


mandel-macaque:
As seen in the commit here: https://github.com/xamarin/xamarin-macios/pull/8690#issuecomment-634533903 the test is not reported using the name but the task id which does not make it very human readable.

PR #8696, opened 2020-05-27
[Harness] Fix issues in the markdown when reporting failures.

mandel-macaque:
The failures are aggregated tasks, we need to get the grouped task, use the key as
the name of the test and use the rest of the data from the ITestTask.
The reason is that the ITestTask.Name was modified by Jenkins.cs to
group them.

fixes: https://github.com/xamarin/xamarin-macios/issues/8695

### DIFF
--- a/tests/xharness/Jenkins/Reports/MarkdownReportWriter.cs
+++ b/tests/xharness/Jenkins/Reports/MarkdownReportWriter.cs
@@ -17,8 +17,7 @@ namespace Xharness.Jenkins.Reports {
 			writer.WriteLine ();
 
 			foreach (var group in failedTests.GroupBy ((v) => v.TestName)) {
-				var enumerableGroup = group as IEnumerable<ITestTask>;
-				if (enumerableGroup != null) {
+				if (group is IEnumerable<ITestTask> enumerableGroup) {
 					foreach (var test in enumerableGroup) {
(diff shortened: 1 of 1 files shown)


tests/xharness/Jenkins/Reports/MarkdownReportWriter.cs
```diff
@@ -16,19 +16,26 @@ void WriteFailedTestsDetails (IEnumerable<ITestTask> failedTests, StreamWriter w
...
-					writer.Write ($"/{test.Variation}");
-				writer.Write ($": {test.ExecutionResult}");
-				if (!string.IsNullOrEmpty (test.FailureMessage))
-					writer.Write ($" ({test.FailureMessage})");
-				if (test.KnownFailure.HasValue)
-					writer.Write ($" Known issue: [{test.KnownFailure.Value.HumanMessage}]({test.KnownFailure.Value.IssueLink})");
-				writer.WriteLine ();
+			foreach (var group in failedTests.GroupBy ((v) => v.TestName)) {
+				var enumerableGroup = group as IEnumerable<ITestTask>;
+				if (enumerableGroup != null) {
```
spouliot, line 21: looks like a candidate for `if (group is IEnumerable<ITestTask> enumerableGroup) {`

PR #8705, opened 2020-05-28
[Harness] Add a new common build issue for Mac OS X test apps.

mandel-macaque:
The issue is recurrent and happens when we try to build mac os x test
applications in which we cannot find the Xamarin.Mac.CSharp.targets. The
search string does not look for the error MSB4226 becuase that coudl be
due to other reasons, but we do know about the Xamarin.Mac targets.

### DIFF
--- a/tests/xharness/Jenkins/ErrorKnowledgeBase.cs
+++ b/tests/xharness/Jenkins/ErrorKnowledgeBase.cs
@@ -15,6 +15,7 @@ namespace Xharness.Jenkins {
 		
 		static readonly Dictionary<string, (string HumanMessage, string IssueLink)> buildErrorMaps = new Dictionary<string, (string HumanMessage, string IssueLink)> {
 			["error MT5210: Native linking failed, undefined symbol: ___multi3"] = (HumanMessage: "Undefined symbol ___multi3 on Release Mode.", IssueLink:"https://github.com/mono/mono/issues/18560"),
+			["Also, tried to find \"Xamarin / Mac / Xamarin.Mac.CSharp.targets\" in the fallback search path(s)"] = (HumanMessage: "\"Xamarin.Mac.CSharp.targets\" could not be found.", IssueLink: "https://github.com/xamarin/maccore/issues/1488"),
 		};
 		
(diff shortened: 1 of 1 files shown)


rolfbjarne:
If you mean the error text [here](http://xamarin-storage/jenkins/xamarin-macios/d16-7/fb5c5187a2f7475b2131ed25a33ddab6af3b3f90/2/jenkins-results/tests/index.html):

Screenshot 2020-05-28 15 10 35

then that's a red herring, those error messages are not the actual problem causing the test to fail.

In that particular case, there's a test failure:

```
20:42:52.2401730 1) Failed : Xamarin.BundlerTests.MX1502_3(macOSMobile)
20:42:52.2402000 warnings
20:42:52.2402110 Expected: 2
20:42:52.2402200 But was: 3
20:42:52.2402280 Warnings:
20:42:52.2402370 	warning MM1502: One or more reference(s) to type 'Foundation.NSObject' already exists inside 'testApp, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' before linking
20:42:52.2402460 	warning MM1503: One or more reference(s) to type 'Foundation.NSObject' still exists inside 'testApp, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' after linking
20:42:52.2402550 	warning MM1503: One or more reference(s) to type 'Foundation.NSObject' still exists inside 'testApp, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' after linking
20:42:52.2402630   at Xamarin.Tests.Tool.AssertWarningCount (System.Int32 count, System.String message) [0x00000] in <6e83eccc6b54496a8d2dba0f96cc5888>:0 
20:42:52.2402720   at Xamarin.BundlerTests.MX1502_3 (Xamarin.Tests.Profile profile) [0x00000] in <6e83eccc6b54496a8d2dba0f96cc5888>:0 
```

We should instead show the test failures in the html report, and instead show that these particular build errors are to be ignored when diagnosing the problem (they're not actually build errors, it's from "nuget restore" failing to restore packages, which isn't a fatal failure).

mandel-macaque:
@rolfbjarne we could add logic to detect nuget restore issues and not show that problem if preset.

rolfbjarne:
@mandel-macaque that could work too

rolfbjarne:
This error will be fixed with #8845, so there won't be a need for known failure code for it.

PR #8716, opened 2020-05-29
[scenekit] Fix BI1106 warnings (don't expose [Model] but protocol interface)

spouliot:
The warning was skipped (an optimization done too early) on some API.
That's fixed with #8712 but SceneKit needed to be updated to make some
API more usable.


spouliot:
failure is a known, test (network) issue -> https://github.com/xamarin/maccore/issues/2160

spouliot:

PR #2184, opened 2017-06-07
[runtime] set mono_set_crash_chaining

lewurm:
it was introduced a while ago for Xamarin-Android:

https://github.com/mono/mono/commit/8ffc4070a94c5c3ef73b99e836e1b1e5652dffba

https://github.com/xamarin/xamarin-android/blob/8d7cc2b8000f8d1b0329561d368c4611fa2f9279/src/monodroid/jni/monodroid-glue.c#L2840

it potentially helps to get even better crash reports.

PR #8718, opened 2020-05-29
Revert "[runtime] set mono_set_crash_chaining (#2184)"

alexischr:
This reverts commit 15c9275b43f40f504b6f115aa951d0b50e1e5425.

Fixes https://github.com/mono/mono/issues/19860

cc @rolfbjarne

### DIFF
--- a/runtime/exports.t4
+++ b/runtime/exports.t4
@@ -554,10 +554,6 @@
 			"mono_bool", "chain_signals"
 		),
 
-		new Export ("void", "mono_set_crash_chaining",
-			"mono_bool", "chain_signals"
-		),
-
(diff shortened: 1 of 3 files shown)


rolfbjarne:
@alexischr what's `mono_set_crash_chaining` actually supposed to do?

rolfbjarne:
`do-not-merge` until I've figured out if this is the right solution or not (there's an ongoing discussion in https://github.com/mono/mono/issues/19860).

alexischr:
We have merged a different solution to the issue that doesn't require changes on the XM/I side (https://github.com/mono/mono/commit/2ff424be2933f711207a8139c5792ab5d1b495a9)

Issue #8468, opened 2020-04-28
[loc][Query][Xamarin_MaciOS] Localizability


PDostalek:
Hi,
please help localization team to answer question from translator.
In case translator is not sure about context/meaning/localizability of string we raise a question

Id | Full Source   String | String Resource   ID | File Name Live   Page Extended | Extended   Description | Answer
-- | -- | -- | -- | -- | --
47 632 | Xamarin.Mac Unified   API against a full .NET framework does not support linking SDK or All   assemblies. Pass either the `-nolink` or `-linkplatform` flag.\r\n\t\t | 0; "MM2007" | mtouch\xlf\Errors.xlf | Please clarify All   assemblies. Is All localizable? |  
47 633 | The type '{0}' (used as {1} {2}) is not available in {3} {4}   (it was introduced in {3} {5}){6} Please build with a newer {3} SDK (usually   done by using the most recent version of Xcode).\r\n\t\t | 0; "MT4162" | mtouch\xlf\Errors.xlf | Please clarify placeholders, especially {6} and confirm that a   space can be added before it |  
47 634 | Async method {0} with more than one   result parameter in the callback by neither ResultTypeName or   ResultType\r\n\t\t | 0; "BI1077" | xlf\Resources.xlf | Unclear string |  
47 737 | {4} {0} does not support a deployment target of {1} for {3}   (the maximum is {2}). Please select an older deployment target in your   project's Info.plist or upgrade to a newer version of {4}.\r\n\t\t | 0; "MX0074" | mtouch\xlf\Errors.xlf | Could you provide more info for the placeholders? Thanks. |  
47 738 | {4} {0} does not support a deployment   target of {1} for {3} (the minimum is {2}). Please select a newer deployment   target in your project's Info.plist.\r\n\t\t | 0; "MT0073" | mtouch\xlf\Errors.xlf | Could you provide more info for the   placeholders? Thanks. |  
47 739 | The Name parameter of the Registrar attribute on the class   '{0}' ('{3}') contains an invalid character: '{1}' (0x{2}).\r\n\t\t | 0; "MT4146" | mtouch\xlf\Errors.xlf | Could you provide more info for the placeholders? Thanks. |  
47 740 | the container app has custom xml   definitions for the managed linker ({0}).\r\n\t\t | 0; "MT0112_c" | mtouch\xlf\Errors.xlf | Could you provide more info for the   placeholders? Thanks. |  
[...]

PR #8721, opened 2020-06-01
[msbuild] missing localization comments - part 3

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-macios/issues/8468

Added missing `<comment/>` fields for:

* BI1033
* BI1077
* MM2007
* MT0073
* MT0074
* MT0112_c
* MT0113_i
* MT4146
* MT4162

I had to split up the `MT4162` error message, introducing:

* `Errors.MT4162_BaseType` - a base type of
* `Errors.MT4162_Parameter` - a parameter in
* `Errors.MT4162_ReturnType` - a return type in
* `Errors.MT4162_PropertyType` - the property type of

This also removed an argument passed into `string.Format`.

### DIFF
--- a/tools/mtouch/xlf/Errors.ru.xlf
+++ b/tools/mtouch/xlf/Errors.ru.xlf
@@ -167,7 +167,7 @@
 		</source>
         <target state="new">Xamarin.Mac Unified API against a full .NET framework does not support linking SDK or All assemblies. Pass either the `-nolink` or `-linkplatform` flag.
 		</target>
-        <note>
+        <note>Using the linker is not supported in certain situations in Xamarin.Mac, see: https://docs.microsoft.com/xamarin/mac/deploy-test/linker. The following are literal names and should not be translated: Xamarin.Mac Unified API, .NET, SDK, All, `-nolink`, and `-linkplatform`.
 		</note>
       </trans-unit>
       <trans-unit id="MM2009">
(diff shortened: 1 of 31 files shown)


tools/mtouch/xlf/Errors.ru.xlf
```diff
@@ -2223,7 +2231,14 @@
 		</source>
         <target state="new">The type '{0}' (used as {1} {2}) is not available in {3} {4} (it was introduced in {3} {5}){6} Please build with a newer {3} SDK (usually done by using the most recent version of Xcode).
 		</target>
-        <note>
+        <note>This is a message for usage of a type that is not available targeting older platform versions. The following are literal names and should not be translated: SDK, Xcode
+{0} - The full C# type name.
+{1} - One of MT4162_BaseType, MT4162_Parameter, MT4162_ReturnType, or MT4162_PropertyType
```
rolfbjarne, line 2236: This might not be possible to translate properly. There could be languages where these inserted parts consist of multiple words that should go in different parts of the sentence, which would make it impossible because there's a single {1}.

I think a better approach would be to have 4 complete error messages, instead of composing one at runtime.

So:

* MT4162_BaseType: The type '{0}' (used as _a base type_ of {2}) is not available in {3} {4} (it was introduced in {3} {5}){6} Please build with a newer {3} SDK (usually done by using the most recent version of Xcode).
* MT4162_Parameter: The type '{0}' (used as _a parameter in_ {2}) is not available in {3} {4} (it was introduced in {3} {5}){6} Please build with a newer {3} SDK (usually done by using the most recent version of Xcode).
...

PR #8723, opened 2020-06-02
[mtouch/mmp] Add CFNetwork to our list of frameworks for iOS and macOS.

rolfbjarne:
It's already in the list for tvOS and watchOS.

### DIFF
--- a/tools/common/Frameworks.cs
+++ b/tools/common/Frameworks.cs
@@ -124,6 +124,7 @@ public class Frameworks : Dictionary <string, Framework>
 
 					{ "Accounts", 10, 8 },
 					{ "AudioVideoBridging", 10, 8 },
+					{ "CFNetwork", 10, 8 },
 					{ "EventKit", 10, 8 },
 					{ "GameKit", 10, 8 },
 					{ "GLKit", 10, 8 },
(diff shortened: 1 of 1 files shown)


tools/common/Frameworks.cs
```diff
@@ -211,6 +212,7 @@ public static Frameworks GetiOSFrameworks (bool is_simulator_build)
 				{ "AddressBookUI", "AddressBookUI", 3 },
 				{ "AudioToolbox", "AudioToolbox", 3 },
 				{ "AVFoundation", "AVFoundation", 3 },
+				{ "CFNetwork", "CFNetwork", 3 },
```
spouliot, line 215: iirc it was unconditionally added because of the StartWWAN code

rolfbjarne: > iirc it was unconditionally added because of the StartWWAN code

Correct:

https://github.com/xamarin/xamarin-macios/blob/bf34d86323f532eab95a73e8f35885a8f1bc641d/tools/common/CompilerFlags.cs#L161-L162

It's still valuable to have a complete list of frameworks here though (less confusing when you start wondering why a framework isn't listed).

spouliot: yeah, unless the dupe cause an issue in our (or later Apple) logic - which I don't think is a problem.

Alternative would be just to add it as a comment.

PR #8729, opened 2020-06-02
[dotnet] Improve the 'clean' target.

rolfbjarne:
Don't nuke the entire DOTNET_DESTDIR, because that means it will be necessary
to run 'make' in the top-level directory again. Instead only clean files that
the dotnet/ directory will re-build.


rolfbjarne:
Test failures are unrelated

* monotouch-test/tvOS: https://github.com/xamarin/maccore/issues/1605

PR #8730, opened 2020-06-02
[foundation] Do not throw inside NSObject.SuperHandle property getter

spouliot:
It's against the FXDG https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

> ❌ AVOID throwing exceptions from property getters.

and it makes things harder in some corner cases, like when Apple calls
a selector after an instance is disposed, e.g.

https://github.com/xamarin/xamarin-macios/issues/8606#issuecomment-637221248

It's not the only case where we throw in getters but this one is an
immediate and much more common case than others.

### DIFF
--- a/src/Foundation/NSObject2.cs
+++ b/src/Foundation/NSObject2.cs
@@ -390,9 +390,6 @@ namespace Foundation {
 
 		public IntPtr SuperHandle {
 			get {
-				if (handle == IntPtr.Zero)
-					return IntPtr.Zero;
-
 				if (class_handle == IntPtr.Zero)
 					class_handle = ClassHandle;
 
--- a/src/Foundation/NSObject2.cs
+++ b/src/Foundation/NSObject2.cs
@@ -391,7 +391,7 @@ namespace Foundation {
 		public IntPtr SuperHandle {
 			get {
 				if (handle == IntPtr.Zero)
-					throw new ObjectDisposedException (GetType ().Name);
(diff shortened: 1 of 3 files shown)


rolfbjarne:
> and it makes things harder in some corner cases, like when Apple calls
> a selector after an instance is disposed

Are you sure we won't run into a native crash instead now?

spouliot:
> Are you sure we won't run into a native crash instead now?

Pretty sure, `objc_msg[Super]Send` does not send the message when the target is `nil` and returns.

Also we already do this _all the time_ as there's not such check inside the `Handle` property, and it ends up being `IntPtr.Zero` once disposed.

I'll add some tests.

spouliot:
well, I stand corrected, it does not send the message but
* `objc_msgSend` works
* `objc_msgSuperSend` crash :(

as the latter expects to be able to peek into a struct like (our old) without a `null` check *before* it realize there's no receiver for the message :|

```csharp
 public struct objc_super {
       public IntPtr Handle;
       public IntPtr SuperHandle;
}
```

IOW while it does not actually care if the `Handle` value is `null`, i.e. it won't send the message, there must be a valid pointer **and** `SuperHandle` is _super_ important.

spouliot:
@dalexsoto yeah, but it's also how all calls using `Handle` are doing :|

Anyway it looks like macOS does not behave like iOS :( 
That's the final nail - might it never wake again 🧛

PR #8744, opened 2020-06-03
[src] Remove files that only contain Classic code.

rolfbjarne:


src/AVFoundation/AVAudioUnit.cs
```diff
@@ -1,30 +0,0 @@
...
-#if !XAMCORE_2_0
-				return new AudioComponentDescription (_AudioComponentDescription);
-#endif
-			}
-		}
-
-			       
-	}
-}
-
```
spouliot, line 30: This seems to remove the API - but it looks like the file is not included in the build.

rolfbjarne: Yes, that's correct, this file wasn't included in the build.

src/AVFoundation/AVAudioUnit.cs
```diff
@@ -1,30 +0,0 @@
-// 
-// AVAudioUnit.cs: Support for AVAudioUnit
-//
-// Authors: Miguel de Icaza (marek.safar@gmail.com)
```
spouliot, line 4: 🤣

rolfbjarne:
@monojenkins backport d16-7

PR #8746, opened 2020-06-03
[generator] Remove [RetainList] from generator-attributes and docs

spouliot:
It's been long gone from the generator code itself.


dalexsoto:
Ahh right 👍

PR #8752, opened 2020-06-04
[dotnet] Detect, compile and publish Info.plist into the app.

rolfbjarne:
* Automatically detect any property lists in the root project directory, and
  include them into the build.
* Introduce the existing build targets to detect and compile Info.plist into
  the .NET build.
* Add documentation for default inclusion. This document will grow over time
  as more file types are automatically included.
* Add some tests.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets
@@ -1,5 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--  Default inclusion -->
+  <PropertyGroup>
+    <!-- Enable default inclusion behavior unless told otherwise, but default to the value for EnableDefaultItems -->
+    <EnableDefaultXamarinPlistItems Condition=" '$(EnableDefaultXamarinPlistItems)' == '' ">$(EnableDefaultItems)</EnableDefaultXamarinPlistItems>
+  </PropertyGroup>
+
+  <!-- Default plist file inclusion -->
+  <ItemGroup Condition="'$(EnableDefaultXamarinPlistItems)' == 'true' ">
+    <None Include="*.plist" Exclude="$(DefaultItemExcludes);$(DefaultXamarinPlistItemExcludes)">
+      <Link>$([MSBuild]::MakeRelative ('$(MSBuildProjectDirectory)', '%(Identity)'))</Link>
+    </None>
+  </ItemGroup>
+
   <!-- Declare the XI/XM framework bundled with this version of the SDK. See Microsoft.NETCoreSdk.BundledVersions.props -->
   <PropertyGroup>
     <!-- Runtime pack identifiers -->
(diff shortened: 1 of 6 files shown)


dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets
```diff
@@ -1,5 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--  Default inclusion -->
+  <PropertyGroup>
+    <!-- Enable default inclusion behavior unless told otherwise, but default to the value for EnableDefaultItems -->
+    <EnableDefaultXamarinPlistItems Condition=" '$(EnableDefaultXamarinPlistItems)' == '' ">$(EnableDefaultItems)</EnableDefaultXamarinPlistItems>
```
rolfbjarne, line 6: Android is using `EnableDefaultAndroid*Items` as the pattern: https://github.com/xamarin/xamarin-android/blob/97c057d9cf4702c65188d00b6c32e3a4c980c824/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props#L12-L13

This pattern would become somewhat annoying for us to implement (`EnableDefaultiOSPlistItems`, `EnableDefaulttvOSPlistItems`, ...), so I went with `Xamarin` as the platform, although I'm not sure about that either.

Maybe someone else has a better idea?

CC @pjcollins

spouliot: I'd go with `EnableDefaultPlistItems` since `plist` are not Xamarin specific.
As for different names by OS then I guess it's better done with `Condition`.

rolfbjarne: A point I forgot to mention is that there will be various other variables: https://github.com/xamarin/xamarin-macios/blob/84c10593ffed2b527e3a2c1a7f126e6a00368f97/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets#L5-L14, and it would be nice to come up with names nobody else is likely to use. All variables are global in MSBuild 😒

spouliot: Except `Binding` there are more `Apple` than `Xamarin`
We do lack a good umbrella name for iOS, tvOS, watchOS and macOS (the later not mobile) and Xamarin includes too many other things (even if things gets renamed)

pjcollins: We discussed this a bit in our weekly msbuild call this morning (which we should maybe extend the invitation for to include some more maciOS representation?). General consensus from that discussion was that we probably want take a step back from this level of granularity and perhaps only introduce one additional property per TFM profile, if at all? I think we don't want people to disable default globbing, and if we are surfacing an option to do so it should be more of an all or nothing type option on a per sdk/profile basis. We maybe also want PM representation to weigh in a bit here from a design perspective before we continue to extend this pattern?

CC @jonathanpeppers @Redth

rolfbjarne:
build

PR #8765, opened 2020-06-04
[src] Remove Classic code from AVFoundation.

rolfbjarne:


rolfbjarne:
@monojenkins backport d16-7

PR #8768, opened 2020-06-04
[Harness] Move the test selection from several bools to a flag.

mandel-macaque:
Move the test selection from a collection of booleans to a flag. Makes it
easier to add new tests and simplifies the code and tests in certain
areas, for example checking which version of the iOS tests to execute
(32, 64 or both).


tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
...
+	[Flags]
+	public enum TestSelection {
+		Mac = 0,
+		iOS32 = 1,
+		iOS64 = 2,
+		iOS = iOS32 | iOS64,
+		iOSExtensions = 8,
+		iOSMSBuild = 16,
+		tvOS = 32,
+		watchOs = 64,
```
akoeplinger, line 27:
```suggestion
		watchOS = 64,
```

mandel-macaque: I would apply from github, but will probably have a compilation error since is in more places, will push the change.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +227,54 @@ void SelectTestsByLabel (int pullRequest)
...
+				if (jenkins.TestSelection.HasFlag (TestSelection.Docs)) {
+					jenkins.TestSelection &= ~TestSelection.Docs; // could have been enabled by 'run-all-tests', so disable it if we can't run it.
 					MainLog.WriteLine ("Disabled 'docs' tests because the Xamarin-specific parts of the build are not enabled.");
 				}
 			}
 
 			// old simulator tests is also a bit special:
 			// - enabled by default if using a beta Xcode, otherwise disabled by default
-			changed = SetEnabled (labels, "old-simulator", ref jenkins.IncludeOldSimulatorTests);
+			//changed = SetEnabled (labels, "old-simulator", ref jenkins.IncludeOldSimulatorTests);
```
spouliot, line 275: commented code should be removed

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
 using Xharness.Jenkins.Reports;
 
 namespace Xharness.Jenkins {
+	[Flags]
+	public enum TestSelection {
```
dalexsoto, line 19: I have a small itch with your enum, we are limited to 30~ values before we overflow and we are currently using 24 slots if I counted correctly, I guess this is minor and since we are not shipping this it can be changed anytime.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
...
+		iOSExtensions = 8,
+		iOSMSBuild = 16,
+		tvOS = 32,
+		watchOs = 64,
+		Mmp = 128,
+		Mtouch = 256,
+		Btouch = 512,
+		MacBindingProject = 1024,
+		Simulator = 2048,
+		OldSimulator,
```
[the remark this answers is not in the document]
mandel-macaque: Ugh, missed adding the value, nice catch.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
...
 
 namespace Xharness.Jenkins {
+	
+	[Flags]
+	public enum TestSelection : long {
+		MacOS = 0,
+		iOS32 = 1,
+		iOS64 = 2,
+		iOS = iOS32 | iOS64,
+		iOSExtensions = 8,
```
rolfbjarne, line 25: 4 is still available.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
 using Xharness.Jenkins.Reports;
 
 namespace Xharness.Jenkins {
+	
+	[Flags]
+	public enum TestSelection : long {
+		MacOS = 0,
```
rolfbjarne, line 21: This flag will always be set.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -15,6 +15,36 @@
...
+		Xtro = 16384,
+		Cecil = 32768,
+		Docs = 65536,
+		BclXUnit = 131072,
+		BclNUnit = 262144,
+		Bcl = BclXUnit | BclNUnit,
+		Mscorlib = 524288,
+		NonMonotouch = 1048576,
+		Monotouch = 2097152,
+		DotNet = 4194304,
```
rolfbjarne, line 45: Maybe use bit shifts instead, that way we don't have to use higher math to find the next value:

```csharp
MacOS = 1 << 0,
iOS32 = 1 << 1,
iOS64 = 1 << 2,
iOSExtensions = 1 << 3,
...
```

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -82,6 +82,37 @@ class TestSelector {
 			"msbuild",
 			"tests/dotnet",
 		};
+		
+			// disabled by default
```
rolfbjarne, line 86: Indentation is wrong

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -31,34 +61,13 @@ class Jenkins {
 		
 		public bool Populating { get; private set; } = true;
 
+		public TestSelection TestSelection { get; set; } = TestSelection.MacOS | TestSelection.iOS | TestSelection.tvOS |
```
rolfbjarne, line 64:
```suggestion
		public TestSelection TestSelection { get; set; } = DefaultTestSelection;
		const TestSelection DefaultTestSelection =  TestSelection.MacOS | TestSelection.iOS | TestSelection.tvOS |
```

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -82,6 +82,37 @@ class TestSelector {
...
+			["mac-binding-project"] = TestSelection.MacBindingProject,
+			["ios-extensions"] = TestSelection.iOSExtensions,
+			["device"] = TestSelection.Device,
+			["xtro"] = TestSelection.Xtro,
+			["cecil"] = TestSelection.Cecil,
+			["old-simulator"] = TestSelection.OldiOSSimulator,
+			["all"] = TestSelection.All,
+		};
+		
+		static readonly Dictionary<string, TestSelection> enabledByDefault = new Dictionary<string, TestSelection> {
```
rolfbjarne, line 104: You can use `DefaultTestSelection` to compute which tests are enabled by default and which aren't, which allows you to remove code duplication (you're specifying which tests are the default ones in multiple places).

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -114,29 +145,17 @@ void SetEnabled (IEnumerable<string> files, string [] prefixes, string testname,
...
-				MainLog.WriteLine ("Disabled '{0}' tests because the label 'skip-all-tests' is set.", testname);
-				value = false;
+				jenkins.TestSelection |= selection;
 				return true;
-			} else if (labels.Contains ("run-all-tests")) {
-				MainLog.WriteLine ("Enabled '{0}' tests because the label 'run-all-tests' is set.", testname);
-				value = true;
-				return true;
-			}
+			}
```
rolfbjarne, line 158: Whitespace change.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -149,15 +168,21 @@ void SelectTestsByModifiedFiles (int pullRequest)
 			MainLog.WriteLine ("Found {0} modified file(s) in the pull request #{1}.", files.Count (), pullRequest);
 			foreach (var f in files)
 				MainLog.WriteLine ("    {0}", f);
+
+			var modifiedFiles = new Dictionary<string, (string [] Prefixes, TestSelection Selection)> {
+				["mtouch"] = (Prefixes: mtouchPrefixes, Selection: TestSelection.Mtouch),
```
rolfbjarne, line 173: You're mapping string to enum here, which is already done elsewhere, which leaves room for typos.

It should be possible to make something like this work instead:

```csharp
var modifiedFiles = new Dictionary<TestSelection, string []> {
	[TestSelection.Mtouch] = mtouchPrefixes,
	...,
};
```

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +227,53 @@ void SelectTestsByLabel (int pullRequest)
...
+			
+			foreach (var keyValuePair in enabledByDefault) {
+				SetEnabled (labels, keyValuePair.Key, keyValuePair.Value);
+			}
 
-			bool inc_permission_tests = false;
-			if (SetEnabled (labels, "system-permission", ref inc_permission_tests))
-				Harness.IncludeSystemPermissionTests = inc_permission_tests;
+			// special case since we are not working with the TestSelection flag in jenkins
+			Harness.IncludeSystemPermissionTests = labels.Contains ("run-system-permission-tests");
```
rolfbjarne, line 239: This is a change in behavior, adding the label `run-all-tests` won't set this flag now.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +227,53 @@ void SelectTestsByLabel (int pullRequest)
...
+				if (jenkins.TestSelection.HasFlag (TestSelection.Docs)) {
+					jenkins.TestSelection &= ~TestSelection.Docs; // could have been enabled by 'run-all-tests', so disable it if we can't run it.
 					MainLog.WriteLine ("Disabled 'docs' tests because the Xamarin-specific parts of the build are not enabled.");
 				}
 			}
 
 			// old simulator tests is also a bit special:
 			// - enabled by default if using a beta Xcode, otherwise disabled by default
-			changed = SetEnabled (labels, "old-simulator", ref jenkins.IncludeOldSimulatorTests);
 			if (!changed && jenkins.IsBetaXcode) {
```
rolfbjarne, line 275: You're still checking `changed` even if you removed the previous line that set `changed`.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +227,53 @@ void SelectTestsByLabel (int pullRequest)
...
-				jenkins.IncludeOldSimulatorTests = true;
+				jenkins.TestSelection |= TestSelection.OldiOSSimulator;
```
rolfbjarne, line 276: This will now be enabled even if `skip-old-simulator-tests` is set when using a beta Xcode.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -285,25 +296,23 @@ public void SelectTests ()
...
-				jenkins.IncludetvOS = false;
+				jenkins.TestSelection &= ~TestSelection.tvOS;
 			}
 
 			if (!Harness.INCLUDE_MAC) {
 				MainLog.WriteLine ("The macOS build is disabled, so any macOS tests will be disabled as well.");
-				jenkins.IncludeMac = false;
-			}
+				jenkins.TestSelection &= ~TestSelection.MacOS;
+			}
```
rolfbjarne, line 315: Whitespace change.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -202,14 +212,14 @@ Task PopulateTasksAsync ()
...
-				Ignored = !IncludeDotNet,
+				Ignored = !TestSelection.HasFlag (TestSelection.DotNet),
 			};
 			var runDotNetTests = new DotNetTestTask (this, buildDotNetTests, processManager) {
 				TestProject = buildDotNetTestsProject,
 				Platform = TestPlatform.All,
 				TestName = "DotNet tests",
 				Timeout = TimeSpan.FromMinutes (5),
-				Ignored = !IncludeDotNet,
+				Ignored = ! TestSelection.HasFlag (TestSelection.DotNet),
```
rolfbjarne, line 222:
```suggestion
				Ignored = !TestSelection.HasFlag (TestSelection.DotNet),
```

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -150,14 +162,9 @@ void SelectTestsByModifiedFiles (int pullRequest)
...
 			
-			SetEnabled (files, mtouchPrefixes, "mtouch", ref jenkins.IncludeMtouch);
-			SetEnabled (files, mmpPrefixes, "mmp", ref jenkins.IncludeMmpTest);
-			SetEnabled (files, bclPrefixes, "bcl", ref jenkins.IncludeBcl);
-			SetEnabled (files, btouchPrefixes, "btouch", ref jenkins.IncludeBtouch);
-			SetEnabled (files, macBindingProject, "mac-binding-project", ref jenkins.IncludeMacBindingProject);
-			SetEnabled (files, xtroPrefixes, "xtro", ref jenkins.IncludeXtro);
-			SetEnabled (files, cecilPrefixes, "cecil", ref jenkins.IncludeCecil);
-			SetEnabled (files, dotnetPrefixes, "dotnet", ref jenkins.IncludeDotNet);
+			foreach (var selection in selectionPrefixesMap.Keys) {
```
rolfbjarne, line 165: `all` should be processed first, so that doing `skip-all-tests,run-mtouch-tests` runs the mtouch tests (and nothing else). Just iterating over the keys do not guarantee any order, since this is a dictionary.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +209,50 @@ void SelectTestsByLabel (int pullRequest)
...
-			if (SetEnabled (labels, "system-permission", ref inc_permission_tests))
-				Harness.IncludeSystemPermissionTests = inc_permission_tests;
+			// special case since we are not working with the TestSelection flag in jenkins
+			Harness.IncludeSystemPermissionTests = labels.Contains ("run-system-permission-tests") || jenkins.TestSelection.HasFlag (TestSelection.All);
 
 			// docs is a bit special:
 			// - can only be executed if the Xamarin-specific parts of the build is enabled
 			// - enabled by default if the current branch is master (or, for a pull request, if the target branch is master)
-			var changed = SetEnabled (labels, "docs", ref jenkins.IncludeDocs);
+			var changed = SetEnabled (labels, "docs", TestSelection.Docs);
```
rolfbjarne, line 222: `SetEnabled` doesn't return `true` in this case for `skip-all-tests` anymore, which means we will run the documentation tests if the target branch for a PR is `master` even when `skip-all-tests` is set.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +209,50 @@ void SelectTestsByLabel (int pullRequest)
...
 			if (Harness.ENABLE_XAMARIN) {
 				if (!changed) { // don't override any value set using labels
```
rolfbjarne, line 224: `SetEnabled` doesn't return `true` in this case for `skip-all-tests` or `run-all-tests` anymore, so this comment isn't accurate now. This means we will run the documentation tests if the target branch for a PR is `master` even when `skip-all-tests` is set.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +209,50 @@ void SelectTestsByLabel (int pullRequest)
...
+					jenkins.TestSelection &= ~TestSelection.Docs; // could have been enabled by 'run-all-tests', so disable it if we can't run it.
 					MainLog.WriteLine ("Disabled 'docs' tests because the Xamarin-specific parts of the build are not enabled.");
 				}
 			}
 
 			// old simulator tests is also a bit special:
 			// - enabled by default if using a beta Xcode, otherwise disabled by default
-			changed = SetEnabled (labels, "old-simulator", ref jenkins.IncludeOldSimulatorTests);
+			changed = SetEnabled (labels, "old-simulator", TestSelection.OldiOSSimulator);
 			if (!changed && jenkins.IsBetaXcode) {
```
rolfbjarne, line 254: Same: the old iOS simulator tests will be run even if `skip-all-tests` is set (and we're using a beta Xcode).

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +209,54 @@ void SelectTestsByLabel (int pullRequest)
...
-			SetEnabled (labels, "ios-simulator", ref jenkins.IncludeSimulator);
-			SetEnabled (labels, "non-monotouch", ref jenkins.IncludeNonMonotouch);
-			SetEnabled (labels, "monotouch", ref jenkins.IncludeMonotouch);
+			SetEnabled (labels, "all", TestSelection.All);
 
-			bool inc_permission_tests = false;
-			if (SetEnabled (labels, "system-permission", ref inc_permission_tests))
-				Harness.IncludeSystemPermissionTests = inc_permission_tests;
+			// special case since we are not working with the TestSelection flag in jenkins
+			Harness.IncludeSystemPermissionTests = labels.Contains ("run-system-permission-tests") || jenkins.TestSelection.HasFlag (TestSelection.All);
```
rolfbjarne, line 221: What about `run-all-tests,skip-system-permissions-tests`?

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -202,67 +209,54 @@ void SelectTestsByLabel (int pullRequest)
...
-			SetEnabled (labels, "ios-64", ref jenkins.IncludeiOS64);
-			SetEnabled (labels, "ios", ref jenkins.IncludeiOS); // Needs to be set after `ios-32` and `ios-64` (because it can reset them)
-			SetEnabled (labels, "tvos", ref jenkins.IncludetvOS);
-			SetEnabled (labels, "watchos", ref jenkins.IncludewatchOS);
-			SetEnabled (labels, "mac", ref jenkins.IncludeMac);
-			SetEnabled (labels, "ios-msbuild", ref jenkins.IncludeiOSMSBuild);
-			SetEnabled (labels, "ios-simulator", ref jenkins.IncludeSimulator);
-			SetEnabled (labels, "non-monotouch", ref jenkins.IncludeNonMonotouch);
-			SetEnabled (labels, "monotouch", ref jenkins.IncludeMonotouch);
+			SetEnabled (labels, "all", TestSelection.All);
```
rolfbjarne, line 218: `run-all-tests,skip-<any>-tests` won't skip the selected tests.

tests/xharness/Xharness.Tests/Jenkins/TestSelectorTests.cs
```diff
@@ -0,0 +1,165 @@
...
+			tmpPath = Path.GetTempPath ();
+			harness = new Mock<IHarness> (); 
+			resultParser = new Mock<IResultParser> ();
+			tunnelBore = new Mock<ITunnelBore> ();
+			processManager = new Mock<IProcessManager> ();
+			versionControlSystem = new Mock<IVersionControlSystem> ();
+			log = new Mock<ILog> ();
+
+			// defult expectations
+			harness.Setup (h => h.ENABLE_XAMARIN).Returns (true);
```
rolfbjarne, line 40: You need to test with `h.ENABLE_XAMARIN` being both true and false.

tests/xharness/Xharness.Tests/Jenkins/TestSelectorTests.cs
```diff
@@ -0,0 +1,165 @@
...
+					expectedSelection = TestSelection.All;
+					expectedSelection &= ~TestSelection.MacOS;
+					yield return new TestCaseData ("testBranch", 
+						new List<string> { "run-all-tests", "skip-mac-tests"},
+						expectedSelection, true);
+					
+					// skip all ye select one of the tests
+					expectedSelection = TestSelection.MacOS;
+					yield return new TestCaseData ("testBranch",
+						new List<string> {"skip-all-tests", "run-mac-tests"},
```
rolfbjarne, line 82: There are several places that are missing spaces:

```suggestion
						new List<string> { "skip-all-tests", "run-mac-tests" },
```

tests/xharness/Xharness.Tests/Jenkins/TestSelectorTests.cs
```diff
@@ -0,0 +1,165 @@
...
+					yield return new TestCaseData ("testBranch", new List<string> {"run-all-tests"}, expectedSelection, true);
+
+					// all tests selected via tag, but we added some skips, therefore we skip them
+					expectedSelection = TestSelection.All;
+					expectedSelection &= ~TestSelection.MacOS;
+					yield return new TestCaseData ("testBranch", 
+						new List<string> { "run-all-tests", "skip-mac-tests"},
+						expectedSelection, true);
+					
+					// skip all ye select one of the tests
```
rolfbjarne, line 79:
```suggestion
					// skip all and select one of the tests
```

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -31,34 +61,14 @@ class Jenkins {
...
 		public bool Populating { get; private set; } = true;
 
+		public TestSelection TestSelection { get; set; } = DefaultTestSelection;
+		internal const TestSelection DefaultTestSelection =  TestSelection.MacOS | TestSelection.iOS | TestSelection.tvOS |
+		                                                   TestSelection.watchOS | TestSelection.iOSMSBuild |
+		                                                   TestSelection.Simulator | TestSelection.NonMonotouch |
+		                                                   TestSelection.Monotouch;
+
 		public IHarness Harness { get; }
-		public bool IncludeAll;
```
rolfbjarne, line 35: You could have these properties become a getter that computes the value:

```csharp
public bool IncludeAll => (TestSelection & TestSelection.All) == TestSelection.All;
public bool IncludeBcl => (TestSelection & TestSelection.Bcl) == TestSelection.Bcl; 
...
```

that would make the diff significantly smaller, since you don't have to change the places that use these fields.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -201,68 +208,67 @@ void SelectTestsByLabel (int pullRequest)
 			}
 
 			MainLog.WriteLine ($"In total found {labels.Count ()} label(s): {string.Join (", ", labels.ToArray ())}");
+			
+			// first check if the all label is used, we either skip our run all, then check the skips, then run,
```
rolfbjarne, line 212: Not sure why the previous logic in `SetEnabled` to establish this precedence doesn't work anymore? Not changing existing logic seems like the better option as opposed to writing new logic.

tests/xharness/Jenkins/Jenkins.cs
```diff
@@ -31,34 +61,14 @@ class Jenkins {
...
-		public bool IncludeiOSMSBuild = true;
-		public bool IncludeMtouch;
-		public bool IncludeBtouch;
-		public bool IncludeMacBindingProject;
-		public bool IncludeSimulator = true;
-		public bool IncludeOldSimulatorTests;
-		public bool IncludeDevice;
-		public bool IncludeXtro;
-		public bool IncludeCecil;
-		public bool IncludeDocs;
```
rolfbjarne, line 55: You could even add a setter here, that would make the `enable-docs-tests` changes simpler:

```csharp
public bool IncludeDocs {
    get { return (TestSelection & TestSelection.Docs) == TestSelection.Docs; }
    set {
        if (value) {
            TestSelection |= TestSelection.Docs;
        } else {
            TestSelection &= ~TestSelection.Docs;
        }
    }
}
```

tests/xharness/Jenkins/TestSelector.cs
mandel-macaque: I'll revert to that with flags, will leave the tests. To much work for no gain.

tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -203,66 +180,65 @@ void SelectTestsByLabel (int pullRequest)
...
+			SetEnabled (labels, "bcl-nunit", TestSelection.BclNUnit);
+			SetEnabled (labels, "mscorlib", TestSelection.Mscorlib);
+			SetEnabled (labels, "btouch", TestSelection.Btouch);
+			SetEnabled (labels, "mac-binding-project", TestSelection.MacBindingProject);
+			SetEnabled (labels, "ios-extensions", TestSelection.iOSExtensions);
+			SetEnabled (labels, "device", TestSelection.Device);
+			SetEnabled (labels, "xtro", TestSelection.Xtro);
+			SetEnabled (labels, "cecil", TestSelection.Cecil);
+			SetEnabled (labels, "old-simulator", TestSelection.OldiOSSimulator);
+			SetEnabled (labels, "all", TestSelection.All);
```
mandel-macaque, line 196: All should be moved to the very bottom of the checks.

tests/xharness/Xharness.Tests/Jenkins/TestSelectorTests.cs
rolfbjarne: @mandel-macaque as far as I can tell this is still not the case

PR #8769, opened 2020-06-04
[tools] Remove MD5 from tooling

spouliot:
Note: affected tools are not included in the code shipping inside the SDK

Usage of deprecated cryptographic algorithms is not approved anymore,
even if not used for cryptographic purpose.

MD5 was used to create immutable GUID since it's output is 128bits which
is just what a GUID wants as it's input (16 bytes). The same can be
achieved (even if a bit slower) with a newer/longer hash function

https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1128148


spouliot:

PR #8779, opened 2020-06-05
[src] Remove Classic code from the CoreData, CoreFoundation, CoreGraphics, CoreImage and CoreLocation frameworks.

rolfbjarne:


rolfbjarne:
@monojenkins backport d16-7

Issue #8787, opened 2020-06-05
NSPasteboardTests_WriteObjectTests fails with a NRE


mandel-macaque:
The test does not always fail and does not do it in all the OS versions. Looking at the code of the test, the nspastebin is not released, leading to a leak, which might be the root cause of the issue. The issue started to occur after: https://github.com/xamarin/xamarin-macios/commit/5d2598af695c41c32712cd136d02c0e8cdfcac5e

spouliot:
That would not be a classic _memory_ leak since it survive app restart.

>  Because the lifetime of a unique pasteboard is not related to the lifetime of the creating app, you must release a unique pasteboard by calling releaseGlobally to avoid possible leaks.

https://developer.apple.com/documentation/appkit/nspasteboard/1528936-pasteboardwithuniquename?language=objc

Not clear if a reboot can solve the `null` being returned on bots...

rolfbjarne:
I think we should try to reboot the affected bots, they've been up 450+ days.

PR #8788, opened 2020-06-05
[AppKit][Tests] Do not fail when we cannot create a NSPasteboard.

mandel-macaque:
The test has started failign more commonly, but only in a number of old
OS versions and not all the time. The function should return a new
Pasteboard, but it is true that we are not releasing the old ones,
meaning that the pasteboard is left after the app is done as stated by
the apple documenation.

The test has been updated to:

1. Release the pasteboard.
2. Be inconclusive if the pasteboard service could not create a new one.

fixes: https://github.com/xamarin/xamarin-macios/issues/8787


rolfbjarne:
@monojenkins backport d16-7
@monojenkins backport d16-6

PR #8793, opened 2020-06-08
[src] Remove Classic code from the FinderSync, FileProvider[UI], ExternalAccessory, EventKit[UI], DeviceCheck, CoreWlan and CoreVideo frameworks.

rolfbjarne:


rolfbjarne:
@monojenkins backport d16-7

PR #8794, opened 2020-06-08
[src] Remove Classic code from Foundation.

rolfbjarne:


rolfbjarne:
@monojenkins backport d16-7

PR #8830, opened 2020-06-10
[dotnet] Ship libxamarin.dylib and friends.

rolfbjarne:
Add libxamarin[-debug].[a|dylib] to the NuGets.

Also create a DOTNET_PLATFORMS variable in Make.config and use it everywhere
instead of the PLATFORMS variable we were defining in multiple Makefiles.


rolfbjarne:
build

PR #8835, opened 2020-06-10
[xcode11.6] Initial bump to support Xcode 11.6 Beta 1

dalexsoto:
Aside from Xcode 11.6 Beta 1 support and HealthKit changes, this also include a bump in sharpie since the bots (due to d16-7 & master) now have installed a higher version of sharpie which requires a xtro fix so cherry-picked both commits.

Also about the HealthKit enums, they are not used directly in the API but they are needed for some dictionary values to be compared against.


spouliot:
@dalexsoto please don't force push on your PR as this requires reviewing the whole PR (or last commit) entirely instead of just the change(s) you made :)

dalexsoto:
@spouliot yeah my bad, I did not want to squash the whole thing but keep the 3 commits separated, I guess I could have squashed again when green but that would have triggered again yet another build :/

PR #8852, opened 2020-06-15
[xharness] Don't ignore a failure to restore nugets.

rolfbjarne:


mandel-macaque:
Looks like all the failures are related to the fact that we could not restore the nugets.

PR #8854, opened 2020-06-15
[msbuild] Share the DeviceSpecificIntermediateOutputPath property.

rolfbjarne:
We don't technically need this value for Xamarin.Mac, but many of the targets
that can be shared need to use this for Xamarin.iOS, so set this value for
Xamarin.Mac as well. It will always be the default IntermediateOutputPath for
Xamarin.Mac, while for Xamarin.iOS it might be changed in
_ComputeTargetArchitectures.


rolfbjarne:
build

PR #8859, opened 2020-06-15
[runtime] Simplify string code inside launcher by using ObjC

spouliot:
This avoids the use of `strcpy`, `memcpy` and `strlen` which can be
misused. We're already using ObjC code inside the file so we can
leverage higher-level API that makes review the code easier.

### DIFF
--- a/runtime/launcher.m
+++ b/runtime/launcher.m
@@ -276,15 +276,11 @@ static int
 push_env (const char *variable, NSString *str_value)
 {
 	const char *current;
-	NSString *value;
 	
-	if ((current = getenv (variable)) && *current) {
-		value = [NSString stringWithFormat: @"%s:%@", current, str_value];
-	} else {
-		value = str_value;
-	}
+	if ((current = getenv (variable)) && *current)
+		str_value = [str_value stringByAppendingFormat: @":%s", current];
 	
-	return setenv (variable, [value UTF8String], 1);
+	return setenv (variable, [str_value UTF8String], 1);
 }
 #endif
 
--- a/runtime/launcher.m
+++ b/runtime/launcher.m
@@ -53,14 +53,13 @@ init_logdir (void)
 	static int redirected_stdout = -1;
 	static int redirected_stderr = -1;
 	const char *env;
-	size_t dirlen;
-	char *path;
 
 	if ((env = getenv ("MONOMAC_LOGDIR")) != NULL && *env) {
+		NSString *logdir = [NSString stringWithUTF8String: env];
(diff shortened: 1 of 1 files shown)


runtime/launcher.m
```diff
@@ -285,23 +275,16 @@
...
+	NSString *value;
 	
 	if ((current = getenv (variable)) && *current) {
-		char *buf = (char *) malloc (len + strlen (current) + 2);
-		memcpy (buf, value, len);
-		buf[len] = ':';
-		strcpy (buf + len + 1, current);
-		rv = setenv (variable, buf, 1);
-		free (buf);
+		value = [NSString stringWithFormat: @"%s:%@", current, str_value];
```
rolfbjarne, line 282: I think you reversed the order here:

```suggestion
		value = [NSString stringWithFormat: @"%@:%s", str_value, current];
```

spouliot: I don't think so. It does the same, both what I read (from source) and executed. E.g.

```
int main(int argc, char *argv[])
{
	printf ("NEW %s\n", getenv ("NEW"));
	push_env ("NEW", @"Bonjour");
	printf ("NEW %s\n", getenv ("NEW"));

	printf ("TMPDIR %s\n", getenv ("TMPDIR"));
	push_env ("TMPDIR", @"/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/");
	printf ("TMPDIR %s\n", getenv ("TMPDIR"));
}
```

with existing (old) code
```
$ clang t2.m -framework Foundation
$ ./a.out 
NEW (null)
NEW Bonjour
TMPDIR /var/folders/4p/hr7rrbjj2w5d1ky_yrjw1s380000gn/T/
TMPDIR /var/folders/4p/hr7rrbjj2w5d1ky_yrjw1s380000gn/T/:/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/
```

with new ObjC code
```
$ clang t2.m -framework Foundation
$ ./a.out 
NEW (null)
NEW Bonjour
TMPDIR /var/folders/4p/hr7rrbjj2w5d1ky_yrjw1s380000gn/T/
TMPDIR /var/folders/4p/hr7rrbjj2w5d1ky_yrjw1s380000gn/T/:/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/
```

spouliot: Scratch that, source should be `t1.m` in the first one.
Funny thing is that my first version was correct and I inverted it for some reason (I can't see why now).

PR #8874, opened 2020-06-17
[DevOps] Move away from a testing branch and use main.

mandel-macaque:
This is mainly for completeness. If we revert the coming yaml, we should use main.


mandel-macaque:
@monojenkins backport d16-7
@monojenkins backport d16-6

PR #8875, opened 2020-06-17
[msbuild] Unify the CompileAppManifest target.

rolfbjarne:
Also rework the class hierarchy a little bit, so that Xamarin.iOS and
Xamarin.Mac are identical:

    CompileAppManifestTaskBase
    └─── iOS/CompileAppManifestTaskCore
    │    └─── iOS/CompileAppManifest
    └─── Mac/CompileAppManifestTaskCore
        │    └─── Mac/CompileAppManifest


msbuild/Xamarin.MacDev.Tasks.Core/PlatformFramework.cs
```diff
@@ -66,5 +66,27 @@ public static string GetMinimumOSVersionKey (ApplePlatform platform)
 				throw new InvalidOperationException ($"Invalid platform: {platform}");
 			}
 		}
+
+		public static string GetMinimumVersionOperatingSystem (string targetFrameworkMoniker, bool is_simulator)
```
spouliot, line 70: method name mention version but it only returns platform/OS names, no version information

msbuild/Xamarin.Mac.Tasks/Tasks/CompileAppManifest.cs
```diff
@@ -6,50 +6,7 @@
...
-			// Merge with any partial plists generated by the Asset Catalog compiler...
-			MergePartialPlistTemplates (plist);
-
-			CompiledAppManifest = new TaskItem (Path.Combine (AppBundleDir, "Contents", "Info.plist"));
-			plist.Save (CompiledAppManifest.ItemSpec, true, true);
-
-			return !Log.HasLoggedErrors;
-		}
+namespace Xamarin.Mac.Tasks {
+	public class CompileAppManifest : CompileAppManifestTaskCore {
```
spouliot, line 10: semi-unrelated, but it's a recurring issue fixed here
we could use the cecil-based tests to ensure `Core` tasks have the code and the non-`Core` code are empty

msbuild/Xamarin.MacDev.Tasks.Core/PlatformFramework.cs
rolfbjarne: It was supposed to be something like `GetOperatingSystemAsPassedToMinimumVersionString`, but I found that a little bit too long. There's already a `GetOperatingSystem` method in this class that returns different values, so I had to come up with something different. I've added a comment to explain the usage.

rolfbjarne: Actually these methods are unrelated to this PR, so I just removed them, they're only in #8875 now.

rolfbjarne:
Test failures are unrelated

* */watchOS 32-bit: https://github.com/xamarin/maccore/issues/581
* xammac tests: will be fixed with #8886.

msbuild/Xamarin.Mac.Tasks/Tasks/CompileAppManifest.cs
rolfbjarne: > semi-unrelated, but it's a recurring issue fixed here
> we could use the cecil-based tests to ensure `Core` tasks have the code and the non-`Core` code are empty

@spouliot see https://github.com/xamarin/xamarin-macios/pull/8888

PR #8876, opened 2020-06-17
[msbuild] Split out min version argument calculation from MetalTask.

rolfbjarne:


msbuild/Xamarin.MacDev.Tasks.Core/PlatformFramework.cs
```diff
@@ -66,5 +66,27 @@ public static string GetMinimumOSVersionKey (ApplePlatform platform)
 				throw new InvalidOperationException ($"Invalid platform: {platform}");
 			}
 		}
+
+		public static string GetMinimumVersionOperatingSystem (string targetFrameworkMoniker, bool is_simulator)
```
spouliot, line 70: same as other PR
this method has nothing to do with (minimum) versions

rolfbjarne: It was supposed to be something like `GetOperatingSystemAsPassedToMinimumVersionString`, but I found that a little bit too long. There's already a `GetOperatingSystem` method in this class that returns different values, so I had to come up with something different. I've added a comment to explain the usage.

PR #8880, opened 2020-06-17
[runtime] Remove memcpy usage from Runtime.CloneMemory

spouliot:
instead use `Buffer.MemoryCopy`.

Currently only used from `CGDataProvider`. Added unit tests for the
public/indirect, usage of the API (we had none).


rolfbjarne:
@monojenkins backport d16-7

PR #8890, opened 2020-06-18
[runtime] Remove internal memcpy p/invoke

spouliot:
* Replace `memcpy` with `Buffer.MemoryCopy`
* Add cecil-based test to make sure we're not p/invoke into it again (nor any other MS banned API)
* Remove `memcpy` from xtro ignore file

### DIFF
--- a/src/CoreMidi/MidiThruConnectionParams.cs
+++ b/src/CoreMidi/MidiThruConnectionParams.cs
@@ -329,7 +329,7 @@ namespace CoreMidi {
 				if (connectionParams.NumControlTransforms > 0) {
 					unsafe {
 						fixed (void* arrAddr = &Controls[0])
-							Buffer.MemoryCopy ((void*) bufferEnd, arrAddr, controlsSize * connectionParams.NumControlTransforms, controlsSize * connectionParams.NumControlTransforms);
+							Buffer.MemoryCopy (arrAddr, (void*) bufferEnd, controlsSize * connectionParams.NumControlTransforms, controlsSize * connectionParams.NumControlTransforms);
 					}
 				}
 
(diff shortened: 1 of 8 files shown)


src/CoreMidi/MidiThruConnectionParams.cs
```diff
@@ -327,19 +327,19 @@ internal NSData WriteStruct ()
...
 				if (connectionParams.NumMaps > 0) {
 					// Set the destination buffer after controls arr if any
 					bufferEnd = IntPtr.Add (bufferEnd, controlsSize * connectionParams.NumControlTransforms);
-					unsafe { // Lets use memcpy to avoid a for loop
+					unsafe {
 						for (int i = 0; i < connectionParams.NumMaps; i++) {
-							fixed (byte* arrAddr = &Maps[i].Value [0])
-								Runtime.memcpy (bufferEnd, (IntPtr) arrAddr, 128);
+							fixed (void* arrAddr = &Maps[i].Value [0])
+								Buffer.MemoryCopy ((void*) bufferEnd, arrAddr, 128, 128);
```
rolfbjarne, line 342: Buffer.MemoryCopy has the source/destination parameters reversed wrt memcpy, but they're the same order here.

A simpler change would have been to implement `Runtime.memcpy` using `Buffer.MemoryCopy` :smile:

src/CoreMidi/MidiThruConnectionParams.cs
```diff
@@ -327,19 +327,19 @@ internal NSData WriteStruct ()
 				Marshal.StructureToPtr (connectionParams, buffer, false);
 
 				if (connectionParams.NumControlTransforms > 0) {
-					unsafe { // Lets use memcpy to avoid a for loop
-						fixed (MidiControlTransform* arrAddr = &Controls[0])
-							Runtime.memcpy (bufferEnd, (IntPtr) arrAddr, controlsSize * connectionParams.NumControlTransforms);
+					unsafe {
+						fixed (void* arrAddr = &Controls[0])
+							Buffer.MemoryCopy ((void*) bufferEnd, arrAddr, controlsSize * connectionParams.NumControlTransforms, controlsSize * connectionParams.NumControlTransforms);
```
rolfbjarne, line 332: Buffer.MemoryCopy has the source/destination parameters reversed wrt memcpy, but they're the same order here.

src/CoreMidi/MidiThruConnectionParams.cs
spouliot: heh, I wanted the name to disappear :)

PR #8893, opened 2020-06-18
[runtime] Use `strnlen` instead of `strlen`

spouliot:
`strlen` is part of the list of C API banned by Microsoft.

Also rename a local variable in `Runtime.cs` so grepping the source
files won't show `strlen` outside the `tests` subdirectories.


spouliot:
@rolfbjarne that's what Binscope for Mac does (and is required to use by our security tenets). The tool is not public but we have a script to run it, in [maccore](https://github.com/xamarin/maccore/tree/main/tools/binscopemac-runner) and an issue https://github.com/xamarin/maccore/issues/2232 to automate it on internal bots.

I did add the check for p/invokes in managed code, as this was not covered. So we could reuse the list of symbols (from BSfM) in the test you mentioned. It does not remove the requirement, but it would give partial results earlier.

PR #8908, opened 2020-06-19
[dotnet] Make the generated main function work.

rolfbjarne:
With this PR a very basic sample app will build and launch successfully in the simulator.

This PR is best reviewed commit-by-commit; each change is explained by its commit message.


tests/dotnet/MySingleView/AppDelegate.cs
```diff
@@ -7,9 +7,21 @@ namespace MySingleView
 {
 	public partial class AppDelegate : UIApplicationDelegate
 	{
-		public override UIWindow Window {
-			get;
-			set;
+		UIWindow window;
```
spouliot, line 10: not a big deal but it would be better to keep the property (for anyone looking for an example)

tools/dotnet-linker/LinkerConfiguration.cs
```diff
@@ -13,6 +13,7 @@
 namespace Xamarin.Linker {
 	public class LinkerConfiguration {
 		public List<Abi> Abis;
+		public string AssemblyName { get; private set; }
```
spouliot, line 16: The naming is unclear, that's the _root_ or _entry_ assembly right ?

tests/dotnet/MySingleView/AppDelegate.cs
rolfbjarne: I was planning on reverting to the more complex sample again when I get storyboards working.

tools/dotnet-linker/LinkerConfiguration.cs
rolfbjarne: It's the `AssemblyName` MSBuild property of the executable project, so yes, it's the root/entry assembly. I'll add a comment to explain it a bit better.

PR #8910, opened 2020-06-19
[xharness] Set the working directory to the project directory when building projects.

rolfbjarne:
This is particularly important when we have global.json and NuGet.config in
the project directory, because those are looked up starting with the current
directory, so those two need to match.


rolfbjarne:
Test failures are unrelated

* monotouch-test/iOS 32-bits: https://github.com/xamarin/maccore/issues/2219

PR #2248, opened 2017-06-27
[common] Fix StringUtils.Quote to quote backslashes too. Fixes #57768.

rolfbjarne:
Some Quote implementations quoted backslashes, some didn't. When selecting a
common implementation, one of the implementations that didn't quote
backslashes won, and the rest were forgotten. Almost. Except for the MT0106
test, which started failing, thus exposing the winner's deficiencies.

So dethrone the implementation that won and reinstante the importance of the
backslash.

https://bugzilla.xamarin.com/show_bug.cgi?id=57768

PR #8954, opened 2020-06-25
[Tests] Add missing resource needed by tests. Fixes #2248

mandel-macaque:
New test that did not consider that tests are shared with mac os x.

fixes: https://github.com/xamarin/maccore/issues/2248


mandel-macaque:
@spouliot is weird, cause it would have failed even in the PR builds

PR #8958, opened 2020-06-25
[DevOps] Add error handling for incorrect storage path

whitneyschmidt:
Add error handling so that we continue to execute when the following issue occurs:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3841066&view=logs&j=8dce0938-6154-51dc-3f74-a72db5aff65a&t=cc57090b-774f-5444-2e3f-e88cfe4adb9e

```
+ ssh builder@xamarin-storage 'mkdir -p /volume1/storage/jenkins/xamarin-macios/main/00a8a13df86c63ef4d829dc63cd210f62a174bd5/3841066/device-tests'\'''
sh: -c: line 0: unexpected EOF while looking for matching `''
sh: -c: line 1: syntax error: unexpected end of file

##[error]Bash exited with code '1'.
```

If we cannot create the dir, we do not add the rsync call to be executed by xharness. The rest of execution continues as normal.

### DIFF
--- a/tools/devops/device-tests/templates/device-tests.yml
+++ b/tools/devops/device-tests/templates/device-tests.yml
@@ -256,10 +256,11 @@ steps:
         eval $MKSTORAGE || EC=$?
         if [ $EC -eq 0 ]; then
           echo '##vso[task.setvariable variable=XAMARIN_STORAGE_PATH;isOutput=true]$XAMARIN_STORAGE_PATH'
-        else
           echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]false'
+          export TESTS_PERIODIC_COMMAND="--periodic-interval 10 --periodic-command rsync --periodic-command-arguments '-avz -e \"ssh\" $PWD/jenkins-results builder@xamarin-storage:/volume1/storage/$XAMARIN_STORAGE_PATH'"
+        else
+          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]true'
         fi
-        export TESTS_PERIODIC_COMMAND="--periodic-interval 10 --periodic-command rsync --periodic-command-arguments '-avz -e \"ssh\" $PWD/jenkins-results builder@xamarin-storage:/volume1/storage/$XAMARIN_STORAGE_PATH'"
       else
         echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]true'
         echo '##vso[task.setvariable variable=XAMARIN_STORAGE_PATH;isOutput=true]""'
--- a/tools/devops/device-tests/templates/device-tests.yml
+++ b/tools/devops/device-tests/templates/device-tests.yml
@@ -7,12 +7,12 @@
 parameters:
 
 - name: statusContext
-  type: string 
-  default: 'iOS Device Tests' # default context, since we started dealing with iOS devices. 
+  type: string
+  default: 'iOS Device Tests' # default context, since we started dealing with iOS devices.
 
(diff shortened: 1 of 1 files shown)


tools/devops/device-tests/templates/device-tests.yml
```diff
@@ -251,9 +251,14 @@ steps:
     cd $WORKING_DIR/xamarin-macios
     if [[ "$USE_XAMARIN_STORAGE" == "True" ]]; then
       if nc -z xamarin-storage 22 2>/dev/null; then
-        ssh builder@xamarin-storage "mkdir -p /volume1/storage/$XAMARIN_STORAGE_PATH"
-        echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]false'
-        echo '##vso[task.setvariable variable=XAMARIN_STORAGE_PATH;isOutput=true]$XAMARIN_STORAGE_PATH'
+        EC=0
+        MKSTORAGE="ssh builder@xamarin-storage "mkdir -p /volume1/storage/$XAMARIN_STORAGE_PATH""
+        eval $MKSTORAGE || EC=$?
```
mandel-macaque, line 256: works! 🚀

tools/devops/device-tests/templates/device-tests.yml
```diff
@@ -251,9 +251,14 @@ steps:
...
+        if [ $EC -eq 0 ]; then
+          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_PATH;isOutput=true]$XAMARIN_STORAGE_PATH'
```
mandel-macaque, line 258: You forgot to add the variable XAMARIN_STORAGE_FAILED to be false, better to add it in case we have to check the value. Also we need to move the export to inside the if.
```suggestion
          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_PATH;isOutput=true]$XAMARIN_STORAGE_PATH'
          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]false'
          export TESTS_PERIODIC_COMMAND="--periodic-interval 10 --periodic-command rsync --periodic-command-arguments '-avz -e \"ssh\" $PWD/jenkins-results builder@xamarin-storage:/volume1/storage/$XAMARIN_STORAGE_PATH'"
```

tools/devops/device-tests/templates/device-tests.yml
```diff
@@ -251,9 +251,14 @@ steps:
...
+        else
+          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]false'
```
mandel-macaque, line 260: This is the otherway around, we did not get a 0, ergo we could not set the path, so we did fail
```suggestion
          echo '##vso[task.setvariable variable=XAMARIN_STORAGE_FAILED;isOutput=true]true'
```

tools/devops/device-tests/templates/device-tests.yml
```diff
@@ -251,9 +251,14 @@ steps:
...
+        fi
         export TESTS_PERIODIC_COMMAND="--periodic-interval 10 --periodic-command rsync --periodic-command-arguments '-avz -e \"ssh\" $PWD/jenkins-results builder@xamarin-storage:/volume1/storage/$XAMARIN_STORAGE_PATH'"
```
mandel-macaque, line 262: This line needs to be removed since it is moved in the if statement with the eq 0/
```suggestion
```

whitneyschmidt: Makes sense - we only want to execute this if we're able to successfully access/create dir on xamarin-storage.

PR #8963, opened 2020-06-26
[DevOp] Allow $XamarinStorage to be empty.

mandel-macaque:
pwsh will complain if a mandatory parameter is empty. Remove the attr
and allow it to be empty since the if statements do check for it.


mandel-macaque:
@whitneyschmidt is any pwsh, is a pwsh module, not a script, so that you can do Import-Module.

PR #8978, opened 2020-06-30
[Foundation] Fix code to not cause a potential null error when compiling using the .NET 5 BCL.

rolfbjarne:
Fix code to not cause a potential null reference error from csc when compiling
using the .NET 5 BCL:

> Foundation/DictionaryContainer.cs(330,47): error CS8604: Possible null reference argument for parameter 'source' in 'IEnumerable<NSObject> Enumerable.Select<T, NSObject>(IEnumerable<T> source, Func<T, NSObject> selector)'.

And do so by rewriting Linq code, which also makes the code much less memory
hungry and more performant.

### DIFF
--- a/src/Foundation/DictionaryContainer.cs
+++ b/src/Foundation/DictionaryContainer.cs
@@ -27,7 +27,6 @@
 //
 
 using System;
-using System.Linq;
 using System.Runtime.InteropServices;
 
 #if !COREBUILD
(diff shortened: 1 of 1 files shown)


src/Foundation/DictionaryContainer.cs
```diff
@@ -326,8 +325,12 @@ protected void SetArrayValue (NSString key, NSNumber[]? values)
 
 		protected void SetArrayValue<T> (NSString key, T[]? values)
 		{
-			if (NullCheckAndRemoveKey (key, values == null))
-				Dictionary [key] = NSArray.FromNSObjects (values.Select (x => NSObject.FromObject (x)).ToArray ());
+			if (NullCheckAndRemoveKey (key, values == null)) {
+				var nsValues = new NSObject [values!.Length];
```
Therzok, line 329: `Array.ConvertAll` is a neat alternative to `Select().ToArray()`

src/Foundation/DictionaryContainer.cs
```diff
@@ -326,8 +325,12 @@ protected void SetArrayValue (NSString key, NSNumber[]? values)
 
 		protected void SetArrayValue<T> (NSString key, T[]? values)
 		{
-			if (NullCheckAndRemoveKey (key, values == null))
-				Dictionary [key] = NSArray.FromNSObjects (values.Select (x => NSObject.FromObject (x)).ToArray ());
+			if (NullCheckAndRemoveKey (key, values == null)) {
```
Therzok, line 328: Looks to me like the problem is that there's no nullability analysis on values == null. `NullCheckAndRemoveKey` does not mark `values` as non-null.

rolfbjarne: I don't think `NullCheckAndRemoveKey` can mark `values` in any way without changing the code in a much bigger way.

PR #8979, opened 2020-06-30
[DevOps] Release device bots as soon as possible.

mandel-macaque:
The upload of the results takes 2 hours:

* 1 hour uploading to vsts test results.
* 1 hour uploading to vsdrops.

This means that we are blocking a device bot for an extra hours (since
the vsdrops is done in a windows bot) when it is not needed. We could
add both uploads to the vsdrops job, but since steps cannot be ran in
parallel, it means that we re waiting for 2 hours when we really do not
have to.

We create two jobs to parallilize the upload of files and we release the
device bot ASAP. That means that our device bots can be free to get new
jobs (and they are the bottle neck) while the windows bots perform the
different uploads.

To make things easier a template that downloads the artifacts was added
so that there is not much code repetition.

### DIFF
--- a/tools/devops/device-tests/scripts/GitHub.psm1
+++ b/tools/devops/device-tests/scripts/GitHub.psm1
@@ -337,7 +337,7 @@ function New-GitHubSummaryComment {
     $headerSb = [System.Text.StringBuilder]::new()
     $headerSb.AppendLine(); # new line to start the list
     $headerSb.AppendLine("* [Azure DevOps]($vstsTargetUrl)")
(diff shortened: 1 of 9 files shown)


tools/devops/device-tests/scripts/GitHub.psm1
```diff
@@ -311,6 +311,7 @@ function New-GitHubSummaryComment {
         $TestSummaryPath,
 
         [String]
+        [AllowEmptyString()]
```
mandel-macaque, line 314: This will be a merge issues with https://github.com/xamarin/xamarin-macios/pull/8977 but was needed to test the full pipeline.

PR #8982, opened 2020-06-30
[xcode12][pencilkit] Update for beta 1 (add macOS support, new apis for iOS)

whitneyschmidt:


src/pencilkit.cs
```diff
@@ -19,8 +19,10 @@
 using UIView = Foundation.NSObject;
 using UIWindow = Foundation.NSObject;
 using UIUserInterfaceStyle = Foundation.NSObject;
+using UIBezierPath = Foundation.NSObject;
```
whitneyschmidt, line 22: Required for PKStroke constructors.

src/pencilkit.cs
```diff
@@ -30,14 +32,16 @@
 
 namespace PencilKit {
 
-	[iOS (13, 0), NoMac]
+	[iOS (13, 0), Mac (10, 16)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
```
whitneyschmidt, line 36: Not sure what we are doing for Catalyst support - are we adding availability attributes now or later?

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
 		[return: NullAllowed]
 		[Export ("sharedToolPickerForWindow:")]
 		PKToolPicker GetSharedToolPicker (UIWindow window);
+
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[Export ("showsDrawingPolicyControls")]
+		bool ShowsDrawingPolicyControls { get; set; }
+
+		// !missing-null-allowed! 'System.Void PencilKit.PKToolPicker::set_AutosaveName(System.String)' is missing an [NullAllowed] on parameter #0
```
whitneyschmidt, line 315: How do we handle binding `[NullAllowed]` properties in this case?

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		[Export ("ink")]
+		PKInk Ink { get; }
+
+		[Export ("transform")]
+		CGAffineTransform Transform { get; }
+
+		[Export ("path")]
+		PKStrokePath Path { get; }
+
+#if MONOMAC
```
whitneyschmidt, line 375: I thought that the #if/#else wouldn't be needed with the `using` statements at the top of this file, but it looks like c# is unable to distinguish between these two signatures for `Mask` w/o the `using` :(

```
GEN      [ios] generated_sources
error BI0002: bgen: Could not compile the API bindings.
		iosurface.cs(202,4): warning CS0657: 'return' is not a valid attribute location for this declaration. Valid attribute locations for this declaration are 'property'. All attributes in this block will be ignored.
	pencilkit.cs(382,16): error CS0102: The type 'PKStroke' already contains a definition for 'Mask'
```

Is there a nicer way to do this?

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		PKStrokePoint GetObjectAtIndexedSubscript (nuint i);
+
+		[Export ("interpolatedLocationAt:")]
+		CGPoint GetInterpolatedLocation (nfloat parametricValue);
+
+		[Export ("interpolatedPointAt:")]
+		PKStrokePoint GetInterpolatedPoint (nfloat parametricValue);
+
+		// option to add an enum since two of these enumerateInterpolates... share the same signature. Not sure whether this makes sense since only 2/3
+		// have the same sig + these aren't constructors.
```
whitneyschmidt, line 422: Enum vs changing names?

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
+	interface PKInk : NSCopying
+	{
+		[Export ("initWithInkType:color:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string type, UIColor color);
+
+		[Export ("inkType")]
+		string InkType { get; }
```
mandel-macaque, line 332: Looking at the [documentation](https://developer.apple.com/documentation/pencilkit/pkinkingtool/inktype) InkType is an enum. We should use a better type than a string.

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
+	interface PKInk : NSCopying
+	{
+		[Export ("initWithInkType:color:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string type, UIColor color);
```
mandel-macaque, line 329: type is an enum.

src/pencilkit.cs
mandel-macaque: I think you can do better than doing the #if here. On monomac we could simply point to the correct type via a using, similar to what we do here: https://github.com/xamarin/xamarin-macios/blob/main/src/UIKit/NSLayoutManager.cs#L24

src/pencilkit.cs
```diff
@@ -2,9 +2,9 @@
 // PencilKit C# bindings
 //
 // Authors:
-//	TJ Lambert  <t-anlamb@microsoft.com>
+//	Whitney Schmidt  <whschm@microsoft.com>
```
mandel-macaque, line 5: I would not remove TJs copyright. It is his original work, we should just add our names, example: https://github.com/xamarin/xamarin-macios/blob/main/src/Foundation/NSUrlSessionHandler.cs#L6

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[Export ("showsDrawingPolicyControls")]
+		bool ShowsDrawingPolicyControls { get; set; }
+
+		// !missing-null-allowed! 'System.Void PencilKit.PKToolPicker::set_AutosaveName(System.String)' is missing an [NullAllowed] on parameter #0
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[Export ("stateAutosaveName")]
+		string stateAutosaveName { get; set; }
```
rolfbjarne, line 319:
```suggestion
		string StateAutosaveName { get; set; }
```

src/pencilkit.cs
rolfbjarne: I believe it should work to add `[NullAllowed]` to the property itself.

src/pencilkit.cs
rolfbjarne: Yes, you can do this at the top:

```csharp
#if MONOMAC
using BezierPath = AppKit.NSBezierPath;
#else
using BezierPath = UIKit.UIBezierPath;
#endif
```

and then here you can just do:

```csharp
[NullAllowed, Export ("mask")]
BezierPath Mask { get; }
```

and the same for the constructor above as well.

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		[NoMac]
+		[NullAllowed, Export ("mask")]
+		UIBezierPath Mask { get; }
+#endif
+
+		[Export ("renderBounds")]
+		CGRect RenderBounds { get; }
+
+		[Export ("maskedPathRanges")]
+		PKFloatRange[] MaskedPath { get; }
```
rolfbjarne, line 389: I think we should keep the `Ranges` part of the name, it's not a suffix (and the swift version of the API also has it):

```suggestion
		PKFloatRange[] MaskedPathRanges { get; }
```

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		IntPtr Constructor (PKStrokePoint[] controlPoints, NSDate creationDate);
+
+		[Export ("count")]
+		nuint Count { get; }
+
+		[Export ("creationDate")]
+		NSDate CreationDate { get; }
+
+		[Export ("pointAtIndex:")]
+		PKStrokePoint PointAtIndex (nuint i);
```
rolfbjarne, line 410:
```suggestion
		PKStrokePoint GetPoint (nuint index);
```

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+
+		[Export ("objectAtIndexedSubscript:")]
+		PKStrokePoint GetObjectAtIndexedSubscript (nuint i);
```
rolfbjarne, line 413:
```suggestion
		PKStrokePoint GetObject (nuint indexedSubscript);
```

src/pencilkit.cs
rolfbjarne: I think the names are fine as-is.

src/pencilkit.cs
dalexsoto: Yep, do not remove copyright/authors from header files

src/pencilkit.cs
dalexsoto: Is fine to add them now, won't hurt and it is likely to help us in the future

src/pencilkit.cs
```diff
@@ -97,12 +113,20 @@ interface PKCanvasView : PKToolPickerObserver {
...
 		UIGestureRecognizer DrawingGestureRecognizer { get; }
 
+		[Introduced (PlatformName.iOS, 13, 0, message: "Use 'drawingPolicy' property instead.")]
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'drawingPolicy' property instead.")]
 		[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]
 		[Export ("allowsFingerDrawing")]
 		bool AllowsFingerDrawing { get; set; }
+
+		[iOS (14, 0)]
+		[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")] // should this line be included? probably...
```
dalexsoto, line 123: It is ok to have

src/pencilkit.cs
dalexsoto: Docs on string backed enums https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#enum-attributes

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		[Export ("inkType")]
+		string InkType { get; }
+
+		[Export ("color")]
+		UIColor Color { get; }
+	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
```
dalexsoto, line 340: There is a constructor here and the properties are read-only so you need to add `[DisableDefaultCtor]`

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		nfloat LowerBound { get; }
+
+		[Export ("upperBound")]
+		nfloat UpperBound { get; }
+	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
+	interface PKStroke : NSCopying
```
dalexsoto, line 356: same, there are constructors here and the properties are read-only so you need to add [DisableDefaultCtor]

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+	{
+		[NoMac]
+		[Export ("initWithInk:strokePath:transform:mask:")]
+		IntPtr Constructor (PKInk ink, PKStrokePath path, CGAffineTransform transform, [NullAllowed] UIBezierPath mask);
```
dalexsoto, line 360: Use Rolf's suggestion on UIBezierPath vs NSBezierPath differences to have a single constructor

src/pencilkit.cs
dalexsoto: Agreed, as a general rule always keep property names just as the ObjC ones and when in doubt look at the swift name.

src/pencilkit.cs
```diff
@@ -246,9 +300,171 @@ interface PKToolPicker {
...
+		[Export ("maskedPathRanges")]
+		PKFloatRange[] MaskedPath { get; }
+	}
+
+	delegate void PKInterpolatedPointsEnumeratorHandler (PKStrokePoint strokePoint, out bool stop);
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
+	interface PKStrokePath : NSCopying
```
dalexsoto, line 397: same, there is a constructor here and the properties are read-only so you need to add [DisableDefaultCtor]

src/pencilkit.cs
dalexsoto: This object is also a good candidate for a read-only [C# indexer](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/) using `GetObject`, this is additional sugar is not really needed but a nice to have.

src/pencilkit.cs
whitneyschmidt: Is this b/c the default constructor has no parameters, but there's no way to add values for the properties after construction? Trying to understand :)

dalexsoto: @whitneyschmidt correct, normally these kind of objects are the result of some other API calls and if we we are wrong adding a ctor is not a breaking change but removing one is :)

mandel-macaque: @whitneyschmidt precisely, there is no point to create a PKFloatRange() with not data. The default constructor is by default the empty one. You can disable it and use the initWithLowerBound:upperBound: one.

src/pencilkit.cs
```diff
@@ -3,8 +3,9 @@
 //
 // Authors:
 //	TJ Lambert  <t-anlamb@microsoft.com>
+//	Whitney Schmidt  <whschm@microsoft.com>
 //
-// Copyright 2019 Microsoft Corporation All rights reserved.
+// Copyright 2020 Microsoft Corporation All rights reserved.
```
rolfbjarne, line 8: Same for the year, you never remove, only add:

```suggestion
// Copyright 2019, 2020 Microsoft Corporation All rights reserved.
```

src/pencilkit.cs
```diff
@@ -97,12 +115,20 @@ interface PKCanvasView : PKToolPickerObserver {
 		[Export ("drawingGestureRecognizer")]
 		UIGestureRecognizer DrawingGestureRecognizer { get; }
 
+		[Introduced (PlatformName.iOS, 13, 0, message: "Use 'drawingPolicy' property instead.")]
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'drawingPolicy' property instead.")]
```
mandel-macaque, line 119: I would use the c# name in this deprecation message, no the Objc.

src/pencilkit.cs
mandel-macaque: This has not yet been fixed.

src/pencilkit.cs
```diff
@@ -49,8 +55,19 @@ enum PKInkType {
...
+	enum PKCanvasViewDrawingPolicy : ulong
+	{
+		Default,
+		AnyInput,
+		PencilOnly,
+	}
+
 	[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]
 	[iOS (13, 0), NoMac]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
```
spouliot, line 70: you need to remove the `[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]` above since it contradict your change

`PlatformName.UIKitForMac` == `PlatformName.MacCatalyst`, just a new name :)

src/pencilkit.cs
```diff
@@ -97,12 +115,20 @@ interface PKCanvasView : PKToolPickerObserver {
 		[Export ("drawingGestureRecognizer")]
 		UIGestureRecognizer DrawingGestureRecognizer { get; }
 
+		[Introduced (PlatformName.iOS, 13, 0, message: "Use 'drawingPolicy' property instead.")]
```
spouliot, line 118: no need for the `[Introduced]` it's already on the type as `[iOS (13,0)]` and the message does not make sense for the introduction

src/pencilkit.cs
```diff
@@ -97,12 +115,20 @@ interface PKCanvasView : PKToolPickerObserver {
...
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'drawingPolicy' property instead.")]
 		[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]
 		[Export ("allowsFingerDrawing")]
 		bool AllowsFingerDrawing { get; set; }
+
+		[iOS (14, 0)]
+		[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 125: ^ @dalexsoto seems sharpie still use the old platform name for Unavailable

src/pencilkit.cs
```diff
@@ -196,6 +246,7 @@ interface IPKToolPickerObserver {}
 
 	[Unavailable (PlatformName.UIKitForMac), Advice ("This API is not available when using UIKit on macOS.")]
 	[iOS (13, 0), NoMac]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
```
spouliot, line 249: same, remove above `[Unavailable]`

src/pencilkit.cs
```diff
@@ -215,10 +266,15 @@ interface PKToolPickerObserver {
...
```
spouliot, line 269: same

src/pencilkit.cs
```diff
@@ -246,9 +302,167 @@ interface PKToolPicker {
 		[Export ("colorUserInterfaceStyle", ArgumentSemantic.Assign)]
 		UIUserInterfaceStyle ColorUserInterfaceStyle { get; set; }
 
+		[Introduced (PlatformName.iOS, 13, 0, message: "Create individual instances instead.")]
```
spouliot, line 305: no need for `[Introduced]`, it's already on the type

src/pencilkit.cs
```diff
@@ -246,9 +302,167 @@ interface PKToolPicker {
...
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
+	interface PKInk : NSCopying
+	{
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		[Export ("initWithInkType:color:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (/* enum PKInkType */ string type, UIColor color);
```
spouliot, line 332: `/* enum PKInkType */ string type` that looks like it should be an `NSString`
using a `string` force a conversion (time/memory)
and it's possible (should not but known to happen) that the native code will compare pointers (and not the string content) and that would fail

src/pencilkit.cs
```diff
@@ -246,9 +302,167 @@ interface PKToolPicker {
...
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[NullAllowed]
+		[Export ("stateAutosaveName")]
+		string StateAutosaveName { get; set; }
+	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[Introduced (PlatformName.MacCatalyst, 14, 0)]
+	[BaseType (typeof(NSObject))]
```
spouliot, line 326: add `[DisableDefaultCtor]` because
* there's a `[DesignatedInitializer]`
* there's no way to set the properties after creating the instance (immutable) so whatever `init` creates has very low value (if it even works)

src/pencilkit.cs
whitneyschmidt: @spouliot should I go through and update all of the changes that use `PlatformName.UIKitForMac` to use Catalyst instead?

spouliot: @whitneyschmidt for this file ? yes, it can be done inside this PR

For all files ? this should be a separate PR and might conflict with other people work on bindings. It's best to do it gradually :)

src/pencilkit.cs
```diff
@@ -246,9 +297,168 @@ interface PKToolPicker {
...
+		PKStrokePoint GetInterpolatedPoint (nfloat parametricValue);
+
+		[Export ("enumerateInterpolatedPointsInRange:strideByDistance:usingBlock:")]
+		void EnumerateInterpolatedPointsByDistance (PKFloatRange range, nfloat distanceStep, PKInterpolatedPointsEnumeratorHandler enumeratorHandler);
+
+		[Export ("enumerateInterpolatedPointsInRange:strideByTime:usingBlock:")]
+		void EnumerateInterpolatedPointsByTime (PKFloatRange range, double timeStep, PKInterpolatedPointsEnumeratorHandler enumeratorHandler);
+
+		[Export ("enumerateInterpolatedPointsInRange:strideByParametricStep:usingBlock:")]
+		void EnumerateInterpolatedPointsByParametric (PKFloatRange range, nfloat parametricStep, PKInterpolatedPointsEnumeratorHandler enumeratorHandler);
```
rolfbjarne, line 424: Just an idea, but would this be better?

```suggestion
		void EnumerateInterpolatedPointsByParametricStep (PKFloatRange range, nfloat parametricStep, PKInterpolatedPointsEnumeratorHandler enumeratorHandler);
```

It sounds slightly better to me, but depending on how the API is used one or the other would make more sense.

In any case, probably not a big deal.

src/pencilkit.cs
```diff
@@ -246,9 +297,168 @@ interface PKToolPicker {
...
 		UIUserInterfaceStyle ColorUserInterfaceStyle { get; set; }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Create individual instances instead.")]
 		[Static]
 		[return: NullAllowed]
 		[Export ("sharedToolPickerForWindow:")]
 		PKToolPicker GetSharedToolPicker (UIWindow window);
+
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
```
spouliot, line 307: not needed, it's already present on the type

src/pencilkit.cs
```diff
@@ -246,9 +297,168 @@ interface PKToolPicker {
 		[Export ("colorUserInterfaceStyle", ArgumentSemantic.Assign)]
 		UIUserInterfaceStyle ColorUserInterfaceStyle { get; set; }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Create individual instances instead.")]
```
spouliot, line 300: you should exclude it from Catalyst (no point adding deprecated API)

src/pencilkit.cs
```diff
@@ -246,9 +297,168 @@ interface PKToolPicker {
...
 		[Export ("sharedToolPickerForWindow:")]
 		PKToolPicker GetSharedToolPicker (UIWindow window);
+
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[Export ("showsDrawingPolicyControls")]
+		bool ShowsDrawingPolicyControls { get; set; }
+
+		[iOS (14, 0)]
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
```
spouliot, line 312: same

src/pencilkit.cs
whitneyschmidt: I like this. According to the documentation, `Executes a given block using each point in a range with a parametric step.` + same for all of these other `EnumerateInterpolated...` methods here. I think it makes sense given the use to change the name here.

whitneyschmidt:
build

PR #8990, opened 2020-07-02
[Foundation] Do expose server errors.

mandel-macaque:
Creating a good API is hard. The delegate DOES return two different
errors.

1. Client errors in the error variable in the delegate method.
2. Server errros in the error in the task in the delegate method.

We need to expose both of them to the user in case there is an issue in
any of them. An exception should be thrown ig any is not null.

PD: As per apple docs:

> The only errors your delegate receives through the error parameter are
> client-side errors, such as being unable to resolve the hostname or
> connect to the host. To check for server-side errors, inspect the
> response property of the task parameter received by this callback.


mandel-macaque:
@monojenkins backport d16-7
@monojenkins backport d16-8

PR #8995, opened 2020-07-03
[msbuild] Unify XamarinMacFrameworkRoot and MonoTouchSdkRoot into _XamarinSdkRoot.

rolfbjarne:


msbuild/Xamarin.Shared/Xamarin.Shared.props
```diff
@@ -171,13 +183,12 @@ Copyright (C) 2020 Microsoft. All rights reserved.
 	</PropertyGroup>
 
 	<PropertyGroup Condition="'$(IsBindingProject)' == 'true'">
-		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'iOS'">$(MonoTouchSdkRoot)/lib/mono/Xamarin.iOS/Xamarin.iOS.dll</BaseLibDllPath>
-		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'tvOS'">$(MonoTouchSdkRoot)/lib/mono/Xamarin.TVOS/Xamarin.TVOS.dll</BaseLibDllPath>
-		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'watchOS'">$(MonoTouchSdkRoot)/lib/mono/Xamarin.WatchOS/Xamarin.WatchOS.dll</BaseLibDllPath>
+		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'iOS'">$(_XamarinSdkRoot)/lib/mono/Xamarin.iOS/Xamarin.iOS.dll</BaseLibDllPath>
+		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'tvOS'">$(_XamarinSdkRoot)/lib/mono/Xamarin.TVOS/Xamarin.TVOS.dll</BaseLibDllPath>
+		<BaseLibDllPath Condition="'$(OS)' == 'Unix' And '$(_PlatformName)' == 'watchOS'">$(_XamarinSdkRoot)/lib/mono/Xamarin.WatchOS/Xamarin.WatchOS.dll</BaseLibDllPath>
```
spouliot, line 188: it's not really the change itself but...

when would it be possible for `'$(_PlatformName)' == 'iOS'"` to be true and `Condition="'$(OS)' == 'Unix' ` to be false ? 
* when run from Windows ? it seems it just would not set `BaseLibDllPath` which would not work

To me it seems that the `Condition="'$(OS)' == 'Unix'` condition is redundant and could be removed.

rolfbjarne: @spouliot This used to be set to a value that wasn't valid:

https://github.com/xamarin/xamarin-macios/pull/8988/files#diff-803b0f269cbc07d96dbbaf64e2f8a399L34

and since that was obviously never used by anyone, I removed that particular case. I suppose I can remove that condition now too then, but I think that can be done in a different PR.

BTW the origin for the invalid value is old: https://github.com/xamarin/maccore/commit/21d2f93be83#diff-84f944514710010dc768f1fcf92c1273R42

PR #8996, opened 2020-07-03
[Accounts] Update binding for xcode 12 beta 1.

mandel-macaque:
Added a deprecation that is not picked up by xtro. No other changes.

### DIFF
--- a/src/accounts.cs
+++ b/src/accounts.cs
@@ -87,8 +87,8 @@ namespace Accounts {
 		[Async]
 		void RequestAccess (ACAccountType accountType, ACRequestCompletionHandler completionHandler);
 
-		[Deprecated (PlatformName.iOS, 14, 0, message: "Public notification deprecated. None Apple applications should not use this API.")]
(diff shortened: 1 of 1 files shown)


src/accounts.cs
```diff
@@ -87,6 +87,8 @@ interface ACAccountStore {
 		[Async]
 		void RequestAccess (ACAccountType accountType, ACRequestCompletionHandler completionHandler);
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Public notification deprecated. None Apple applications should not use this API.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Public notification deprecated. None Apple applications should not use this API.")]
```
rolfbjarne, line 91: The second sentence is not grammatically correct, and I don't think we need it anyway, so I think we can just remove it:

```suggestion
		[Deprecated (PlatformName.iOS, 14, 0, message: "Public notification is deprecated.")]
		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Public notification is deprecated.")]
```

PR #9001, opened 2020-07-04
[CoreLocation] Add support for Xcode 12 beta 1.

mandel-macaque:
A number of APIs added and deprecated in the same release. We will see
that is that about.

The status property was moved to be a instance property in CLLocationManager.
We use the same approach as in CBManager with the Authorization
property.


src/CoreLocation/CLEnums.cs
```diff
@@ -99,4 +99,12 @@ public enum CLActivityType : long {
 		[iOS (12,0)][TV (12,0)][Watch (5,0)][Mac (10,14)]
 		Airborne,
 	}
+
+	[Native]
+	[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 104:
```suggestion
	[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Deprecated (PlatformName.WatchOS, 7, 0)]
+		[Deprecated (PlatformName.TvOS, 14, 0)]
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[Deprecated (PlatformName.MacOSX, 10, 16)]
+		[Export ("authorizedForPreciseLocation")]
+		bool AuthorizedForPreciseLocation { [Bind ("isAuthorizedForPreciseLocation")] get; }
+
+		[Async]
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 441:
```suggestion
		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Export ("requestTemporaryFullAccuracyAuthorizationWithPurposeKey:completion:")]
+		void RequestTemporaryFullAccuracyAuthorization (string purposeKey, [NullAllowed] Action<NSError> completion);
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 445:
```suggestion
		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Export ("requestTemporaryFullAccuracyAuthorizationWithPurposeKey:")]
+		void RequestTemporaryFullAccuracyAuthorization (string purposeKey);
+
+		// TODO: Apple??? 
+		[Async]
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 451:
```suggestion
		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Deprecated (PlatformName.WatchOS, 7, 0)]
+		[Deprecated (PlatformName.TvOS, 14, 0)]
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[Deprecated (PlatformName.MacOSX, 10, 16)]
+		[Export ("requestTemporaryPreciseLocationAuthorizationWithPurposeKey:completion:")]
+		void RequestTemporaryPreciseLocationAuthorization (string purposeKey, [NullAllowed] Action<NSError> completion);
+
+		// TODO: Apple??? 
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 460:
```suggestion
		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -508,6 +569,11 @@ partial interface CLLocationManagerDelegate
 		[NoWatch][NoTV]
 		[Export ("locationManager:didFinishDeferredUpdatesWithError:"), EventArgs ("NSError", true)]
 		void DeferredUpdatesFinished (CLLocationManager manager, [NullAllowed] NSError error);
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
```
mandel-macaque, line 573:
```suggestion
		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
```

src/corelocation.cs
```diff
@@ -508,6 +569,11 @@ partial interface CLLocationManagerDelegate
...
+		[Export ("locationManagerDidChangeAuthorization:")]
+		void LocationManagerDidChangeAuthorization (CLLocationManager manager);
```
mandel-macaque, line 575:
```suggestion
		void DidChangeAuthorization (CLLocationManager manager);
```

src/CoreLocation/CLLocationManager.cs
```diff
@@ -47,5 +47,22 @@ public static bool IsMonitoringAvailable (Type t)
...
+
+		public static CLAuthorizationStatus Status {
+			get {
+				// because an instance property in iOS 14, tvOS 14, Watch 7 and macOS 10,16
+#if IOS || TVOS
+				if (UIKit.UIDevice.CurrentDevice.CheckSystemVersion (14, 0)) {
+#elif WATCH
+				if (WatchKit.WKInterfaceDevice.CurrentDevice.CheckSystemVersion (7, 0)) {
+#else 
+				if (PlatformHelper.CheckSystemVersion (10, 16)) {
```
rolfbjarne, line 59: This would make things easier if we get new platforms we need to support, since we'll get compilation errors instead of having to wait for someone to run into the potentially incorrect version check and file an issue about it:

```suggestion
#elif MONOMAC
				if (PlatformHelper.CheckSystemVersion (10, 16)) {
#else 
				#error Unknown platform
```

src/CoreLocation/CLLocationManager.cs
```diff
@@ -47,5 +47,24 @@ public static bool IsMonitoringAvailable (Type t)
 			return false;
 		}
 #endif
+
+		public static CLAuthorizationStatus Status {
```
rolfbjarne, line 51: Manual API should have manual tests too!

src/CoreLocation/CLLocationManager.cs
```diff
@@ -47,5 +47,24 @@ public static bool IsMonitoringAvailable (Type t)
...
+			get {
+				// because an instance property in iOS 14, tvOS 14, Watch 7 and macOS 10,16
```
spouliot, line 53: there might be a reason why it was changed to an instance ?
if so then only exposing the static one might be problematic
if not then it might be an apple bug and should be reported

src/corelocation.cs
```diff
@@ -282,8 +287,19 @@ partial interface CLLocationManager {
...
-		CLAuthorizationStatus Status { get; }
+		[Internal]
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("authorizationStatus")]
+		CLAuthorizationStatus _IStatus { get; }
+
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use the instance property instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use the instance property instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use the instance property instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use instance property instead.")]
```
spouliot, line 298: remove messages
they will never be used since it's internal

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
 		[Mac (10,14)]
 		[Export ("requestLocation")]
 		void RequestLocation ();
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("accuracyAuthorization")]
+		CLAccuracyAuthorization AccuracyAuthorization { get; }
+
+		// TODO: Apple???
```
spouliot, line 431: ??? please add this to the main tracking issue
or we're likely to forget it
In general we should not expose new/deprecated API

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Async]
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("requestTemporaryFullAccuracyAuthorizationWithPurposeKey:completion:")]
+		void RequestTemporaryFullAccuracyAuthorization (string purposeKey, [NullAllowed] Action<NSError> completion);
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("requestTemporaryFullAccuracyAuthorizationWithPurposeKey:")]
+		void RequestTemporaryFullAccuracyAuthorization (string purposeKey);
+
+		// TODO: Apple???
```
spouliot, line 449: same

src/corelocation.cs
```diff
@@ -407,6 +423,47 @@ partial interface CLLocationManager {
...
+		[Async]
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Deprecated (PlatformName.WatchOS, 7, 0)]
+		[Deprecated (PlatformName.TvOS, 14, 0)]
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[Deprecated (PlatformName.MacOSX, 10, 16)]
+		[Export ("requestTemporaryPreciseLocationAuthorizationWithPurposeKey:completion:")]
+		void RequestTemporaryPreciseLocationAuthorization (string purposeKey, [NullAllowed] Action<NSError> completion);
+
+		// TODO: Apple???
```
spouliot, line 459: same

src/corelocation.cs
```diff
@@ -485,6 +542,10 @@ partial interface CLLocationManagerDelegate
 		void DidVisit (CLLocationManager manager, CLVisit visit);
 #endif
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'LocationManagerDidChangeAuthorization' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LocationManagerDidChangeAuthorization' instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LocationManagerDidChangeAuthorization' instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LocationManagerDidChangeAuthorization' instead.")]
```
spouliot, line 548: update message, you used a better name `DidChangeAuthorization` for the binding

src/corelocation.cs
whitneyschmidt: This is super weird...the impression I got from WWDC is that all of these location authorization APIs were one of the big new features 🤔

mandel-macaque: ditto, I bet it was a screw up in the headers.

src/CoreLocation/CLLocationManager.cs
mandel-macaque: Looking at the documentation, it makes sense to be using the instance one because the status now depends on the user interaction from the delegate and its method:
```objc
- (void)locationManagerDidChangeAuthorization:(CLLocationManager *)manager;
```
The issue here is that we need to keep the static one or we will have a breaking API change.

rolfbjarne: We'll have to use a different name for the instance version if we want both.

mandel-macaque: Exactly, we could do the following:

```csharp
		[Deprecated (PlatformName.iOS, 14, 0, message: "Use the instance property 'AuthorizationStatus' instead.")]
		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use the instance 'AuthorizationStatus' property instead.")]
		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use the instance property AuthorizationStatus' instead.")]
		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use instance property 'AuthorizationStatus' instead.")]
		[Static]
		[Export ("authorizationStatus")]
		CLAuthorizationStatus Status { get; }

		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
		[Export ("authorizationStatus")]
		CLAuthorizationStatus AuthorizationStatus { get; }
```

That would fix the API breaking change,  I don't think we would want to add a #if XAM_CORE since we will have to support the old versions too. With this we remove the manual code, what do you think?

PR #9006, opened 2020-07-05
[CFNetwork] Add support for xcode 12 beta 1.

mandel-macaque:


src/CFNetwork/CFHTTPMessage.cs
```diff
@@ -85,6 +85,11 @@ static IntPtr GetVersion (Version version)
 					return _HTTPVersion2_0.Handle;
 				// HTTP 2.0 requires OS X 10.11 or later.
 				return _HTTPVersion1_1.Handle;
+			} else if (version.Major == 3 && version.Minor == 0) {
+				if (_HTTPVersion3_0 != null && _HTTPVersion3_0.Handle != IntPtr.Zero)
+					return _HTTPVersion3_0.Handle;
+				// HTTP 3.0 requires OS X 10.16 or later.
+				return _HTTPVersion1_1.Handle;
```
rolfbjarne, line 92: Shouldn't the fallback be HTTP/2 and not HTTP/1.1 here (and then it should check if HTTP/2 is supported first, like the code block just above)?

mandel-macaque: You are right.. will change the entire if to got from max to min

spouliot: yup, looks like you should check for it earlier
so if 3.0 is not present, then 2.0 is checked and if not present 1.1 can be used (e.g. on 10.10)

src/cfnetwork.cs
```diff
@@ -67,6 +67,10 @@ interface CFHTTPMessage {
 		[Internal][Field ("kCFHTTPVersion2_0", "CFNetwork")]
 		NSString _HTTPVersion2_0 { get; }
 
+		[Mac (10, 16), iOS (14,0), TV (14, 0)]
+		[Internal][Field ("kCFHTTPVersion3_0", "CFNetwork")]
+		NSString _HTTPVersion3_0 { get; }
```
spouliot, line 72: use `IntPtr` since it's internal and you only use the `Handle` of the `NSString` anyway
you'll same some memory and remove null checks :)

src/CFNetwork/CFHTTPMessage.cs
```diff
@@ -80,16 +80,20 @@ static IntPtr GetVersion (Version version)
 				return _HTTPVersion1_1.Handle;
```
spouliot, line 80: I wonder if `null` meant _latest_ possible. Seems weird to default to 1.1...

src/CFNetwork/CFHTTPMessage.cs
```diff
@@ -80,16 +80,20 @@ static IntPtr GetVersion (Version version)
...
+			else if (version.Major == 3 && version.Minor == 0) {
+				if (_HTTPVersion3_0 != IntPtr.Zero)
+					// HTTP 3.0 requires OS X 10.16 or later.
+					return _HTTPVersion3_0;
+				else if (_HTTPVersion2_0 != null && _HTTPVersion2_0.Handle != IntPtr.Zero)
+					// HTTP 2.0 requires OS X 10.11 or later.
+					return _HTTPVersion2_0.Handle; 
+				else
+					return _HTTPVersion1_1.Handle;
+			} else if (version.Major == 2 && version.Minor == 0) {
```
spouliot, line 92: better :) but I think it could be simplified

First `.Handle` is not needed for the other, already existing, `NSString` fields. They are internal so this would not be a breaking change.

Next, you can reduce the number of `if` by removing `else`

```
if (== 3.0) {
	// HTTP 3.0 requires OS X 10.16 or later.
    if (_HTTPVersion3_0 != IntPtr.Zero)
 		return _HTTPVersion3_0;
}
if (>= 2.0) {
    // this will catch 3.0 on system where it's not supported and return 2.0
    if (_HTTPVersion2_0 != IntPtr.Zero)
 		return _HTTPVersion2_0;
}
if (>= 1.1) {
   // ...
}
...
```

mandel-macaque: Did not change the other ones because the NSStrings are indeed used and did want to minimize the diff. Do you want me to change all occurrences to use handles and create the native string when needed instead?

spouliot: are they used _many_ time elsewhere ? a quick grep suggest noe
if it's only in this method then it makes sense, otherwise it depends on how large the scope become

mandel-macaque: is not too bad, one small change in another method from one of the other internal vars.

src/CFNetwork/CFHTTPMessage.cs
mandel-macaque: I vote for not touching it. Http 1.0 is ooooold

PR #9035, opened 2020-07-08
[src/usernotificationsui] Update for beta 2 (catalyst attributes)

spouliot:


spouliot:
failure on monotouch-test / watchOS - which platform is not part of the PR
```
Tests run: 1568 Passed: 1544 Inconclusive: 2 Failed: 1 Ignored: 23
```
but there's no `[FAIL]` in the logs

PR #9040, opened 2020-07-08
[DevOps] Move back to self hosted windows pools.

mandel-macaque:
The microsoft hosted images have a limit of 10gb, or logs are getting
close to that size when expanded, therefore we will get into issues.
Move back to the self-hosted pool before we have problems.


whitneyschmidt:
> The microsoft hosted images have a limit of 10gb, or logs are getting
> close to that size when expanded, therefore we will get into issues.
> Move back to the self-hosted pool before we have problems.

Are we reaching out to the team that manages the hosted images re upping the limit?

mandel-macaque:
@whitneyschmidt is a worldwide thing, if you want something bigger you have to use self-hosted ones

PR #9041, opened 2020-07-08
[DevOps] Move to the new download build artifacts tasks.

mandel-macaque:


tools/devops/device-tests/templates/download-artifacts.yml
```diff
@@ -14,27 +14,27 @@ steps:
...
 
 # Download the Html Report that was added by the tests job.
-- task: DownloadPipelineArtifact@2
+- task: DownloadBuildArtifacts@0
   displayName: Download html report 
   inputs:
-    patterns: '**/HtmlReport.zip'
-    allowFailedBuilds: true
-    path: $(System.DefaultWorkingDirectory)/Reports
+    allowPartiallySucceededBuilds: true
```
whitneyschmidt, line 20: Sanity check: this is the same as `allowfailedbuilds`?

mandel-macaque: AFAIK yes

PR #9045, opened 2020-07-09
[Xcode12][Photos] Update for beta1 (no beta2 changes)

whitneyschmidt:


src/photos.cs
```diff
@@ -1072,13 +1072,29 @@ interface PHPhotoLibrary {
...
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[Deprecated (PlatformName.MacOSX, 10, 16)]
+		[Deprecated (PlatformName.TvOS, 14, 0)]
 		[Static, Export ("authorizationStatus")]
 		PHAuthorizationStatus AuthorizationStatus { get; }
 
+		[TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("authorizationStatusForAccessLevel:")]
+		PHAuthorizationStatus authorizationStatus (PHAccessLevel accessLevel);
```
spouliot, line 1084: -> `GetAuthorizationStatus`
needs a verb/action (method) and can't start with a lower case character

src/photos.cs
```diff
@@ -1072,13 +1072,29 @@ interface PHPhotoLibrary {
 		[Export ("sharedPhotoLibrary")]
 		PHPhotoLibrary SharedPhotoLibrary { get; }
 
+		[Deprecated (PlatformName.iOS, 14, 0)]
```
spouliot, line 1075: add message to use `authorizationStatusForAccessLevel:` (managed name)

src/photos.cs
```diff
@@ -1072,13 +1072,29 @@ interface PHPhotoLibrary {
...
 		PHAuthorizationStatus AuthorizationStatus { get; }
 
+		[TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("authorizationStatusForAccessLevel:")]
+		PHAuthorizationStatus authorizationStatus (PHAccessLevel accessLevel);
+
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[Deprecated (PlatformName.MacOSX, 10, 16)]
+		[Deprecated (PlatformName.TvOS, 14, 0)]
```
spouliot, line 1088: add message

whitneyschmidt: How do we write messages when the new API has the same name as the old one, but with different parameters?

spouliot: generally I add the type(s), e.g. `"Use the 'Foo(Bar)' overload instead.`"

src/Photos/Enums.cs
```diff
@@ -378,6 +383,8 @@ public enum PHPhotosError : long {
 		LibraryVolumeOffline = 3114,
 		RelinquishingLibraryBundleToWriter = 3142,
 		SwitchingSystemPhotoLibrary = 3143,
+		[TV (14,0), Mac (10,16), iOS (14,0)]
```
spouliot, line 386: we do not add availability to error enums
that's because they are output only (from API) and it's easier to handle values alone (than adding version checks)

PR #9063, opened 2020-07-10
[identitylookup] Update for Xcode 12 beta 2

spouliot:


src/identitylookup.cs
```diff
@@ -21,7 +21,17 @@ namespace IdentityLookup {
 	public enum ILMessageFilterAction : long {
 		None = 0,
 		Allow = 1,
-		Filter = 2,
+		Junk = 2,
+#if !XAMCORE_4_0
+		[Obsolete ("Use 'Junk' instead.")]
```
whitneyschmidt, line 26: 👏

PR #9070, opened 2020-07-13
[coregraphics] Update for Xcode 12 beta 2

spouliot:


src/CoreGraphics/CGColor.cs
```diff
@@ -309,6 +319,17 @@ static public CGColor CreateGenericGrayGamma2_2 (nfloat gray, nfloat alpha)
...
+
+		[iOS (14,0)][TV (14,0)][Watch (7,0)][Mac (10,16)]
+		[DllImport(Constants.CoreGraphicsLibrary)]
+		static extern /* CGColorRef */ IntPtr CGColorCreateGenericCMYK (nfloat cyan, nfloat magenta, nfloat yellow, nfloat black, nfloat alpha);
+
+		[iOS (14,0)][TV (14,0)][Watch (7,0)][Mac (10,16)]
+		static public CGColor CreateCmyk (nfloat cyan, nfloat magenta, nfloat yellow, nfloat black, nfloat alpha)
+		{
+			var h = CGColorCreateGenericCMYK (cyan, magenta, yellow, black, alpha);
+			return h == IntPtr.Zero ? null : new CGColor (h);
```
rolfbjarne, line 331: Do we own the handle?

src/CoreGraphics/CGContextPDF.cs
```diff
@@ -113,6 +117,12 @@ internal override NSMutableDictionary ToDictionary ()
 				ret.LowlevelSetObject (CFBoolean.FalseHandle, kCGPDFContextAllowsCopying);
 			if (AccessPermissions.HasValue)
 				ret.LowlevelSetObject (NSNumber.FromInt32 ((int) AccessPermissions.Value), kCGPDFContextAccessPermissions);
+			// only set the keys if they exists in the current OS version
+			if ((kCGPDFContextCreateLinearizedPDF != IntPtr.Zero) && CreateLinearizedPdf.HasValue)
+				ret.LowlevelSetObject (CFBoolean.ToHandle (CreateLinearizedPdf.Value), kCGPDFContextCreateLinearizedPDF);
+			// default to kCFBooleanFalse
+			if ((kCGPDFContextCreatePDFA != IntPtr.Zero) && CreatePdfA2u.HasValue && CreatePdfA2u.Value == true)
```
rolfbjarne, line 124: Really minor, this can be simplified as:

```suggestion
			if ((kCGPDFContextCreatePDFA != IntPtr.Zero) && CreatePdfA2u == true)
```

src/CoreGraphics/CGColor.cs
```diff
@@ -131,7 +130,18 @@ public CGColor (string name)
...
+
+		[iOS (14,0)][TV (14,0)][Watch (7,0)]
+		public CGColor (CGConstantColor color)
+		{
+			var constant = color.GetConstant ();
+			if (constant == null)
+				throw new ArgumentNullException (nameof (color));
+			handle = CGColorGetConstantColor (constant.Handle);
+			if (handle == IntPtr.Zero)
+				throw new ArgumentException (nameof (color));
```
dalexsoto, line 142: I wonder if we should provided a little more of information like `ObjectDisposedException` here I know this one in particular does not apply here but maybe adding something more meaningful in the exception message? Like "Could not create a native instance of {nameof (color)}" or something along those lines.

src/CoreGraphics/CGColor.cs
dalexsoto, line 331: Shouldn't we use `Runtime.GetINativeObject<T>` here? I know we have `new CGColor` on other similar ones, I would not die on it but  `¯\_(ツ)_/¯`

spouliot: hmm... I copy/pasted code from similar API. Let me check...

src/CoreGraphics/CGColor.cs
spouliot: Ideally this type, like others, should move to subclass `NativeObject` and we would let the `Handle` setting code to _handle_ it

src/CoreGraphics/CGColor.cs
spouliot: we could but it would not be optimal because `NSObject` is a `INativeObject` so the logic to create an instance from `GetINativeObject<T>` is much more costly (and there's no doubt here about which type should be used).

spouliot: copy/paste is bad
except when it spot something already wrong

src/coregraphics.cs
```diff
@@ -76,6 +76,14 @@ interface CGPDFInfo {
 		[Mac (10,13)][iOS (11,0)][TV (11,0)][Watch (4,0)]
 		[Internal][Field ("kCGPDFContextAccessPermissions")]
 		IntPtr kCGPDFContextAccessPermissions { get; }
+
+		[Mac (10,16)][iOS (14,0)][TV (14,0)][Watch (7,0)]
+		[Field ("kCGPDFContextCreateLinearizedPDF")]
+		IntPtr kCGPDFContextCreateLinearizedPDF { get; }
```
mandel-macaque, line 82: Internal? Or a better name?

src/coregraphics.cs
```diff
@@ -76,6 +76,14 @@ interface CGPDFInfo {
...
+
+		[Mac (10,16)][iOS (14,0)][TV (14,0)][Watch (7,0)]
+		[Field ("kCGPDFContextCreatePDFA")]
+		IntPtr kCGPDFContextCreatePDFA { get; }
```
mandel-macaque, line 86: Internal or a better name?

src/coregraphics.cs
spouliot: good catch, it should have been internal, like the other existing fields

src/coregraphics.cs
spouliot: same, the better name is in the _old/manual_ strong dictionary

PR #9073, opened 2020-07-13
[MTouch] Add a workaround for a failing tests in the Xcode 12 betas.

mandel-macaque:
There is an issue in the linker warnings (https://github.com/xamarin/maccore/issues/2264) which makes our CI red. Added a workaround in the test for the Xcode version until is fixed by Apple.


mandel-macaque:
@spouliot  is the future! everything is unknown until it happens!

PR #9087, opened 2020-07-14
[mediasetup] Update for Xcode 12 beta 2

spouliot:
The framework is not available (binary or even headers) for the iOS
simulator, only devices.
    
As such `tools/mtouch/Makefile` is not updated with a `-weak-framework`
    
Tracked with
* https://github.com/xamarin/maccore/issues/2266
* https://feedbackassistant.apple.com/feedback/8001312


pdonovan:
This still doesn't work with the GM of Xcode 12, VS Mac Version 8.7.6 (build 19) and Xamarin.iOS 13.99.0.251.

Are there any temporary workarounds I can do to get our app building for the simulator again? Even if it means temporarily ripping out some parts (I don't know what MediaSetup is used for)?

praeclarum:
I'm hitting this too when building for simulator.

rolfbjarne:
@praeclarum hitting what exactly? do you have a build log?

pdonovan:
I've attached part of the build log that seems to be pertinent to the problem.
[mediasetuperror.txt](https://github.com/xamarin/xamarin-macios/files/5259708/mediasetuperror.txt)

praeclarum:
Sorry @rolfbjarne, I had an edit that I don't seem to have saved.

There was an issue in the simulator where it kept trying to link with MediaSetup even though the sim doesn't support it.

Fortunately, the stable build released yesterday fixes it. So you can disregard my comment.

PR #9088, opened 2020-07-14
[msbuild] Allow overriding CompileAppManifestTaskBase Execute method

emaf:
If the Execute method of a task is sealed the Windows side tasks won't be able to "inject" the code that executes the tasks remotely.


rolfbjarne:
@monojenkins backport main

Issue #8943, opened 2020-06-24
Xcode 12 final audit checklist


spouliot:
## Stuff we need to review / fix before the final / stable release of the SDK

- [ ] [fileprovider] Remove API to match Xcode 12. Workaround for device (and registrar) builds

> * (change)[https://github.com/xamarin/xamarin-macios/pull/8935/commits/4f2b839fb121d5543ee085b79c995d28af8bc135]

juliuszint:
I am really looking forward to `UICollectionViewDiffableDataSourceSectionSnapshotHandlers` and `UICollectionViewDiffableDataSourceReorderingHandlers` being available. Thank you all for your work.

Dolfik1:
Any update on `UICollectionViewDiffableDataSourceReorderingHandlers`?

keozx:
So `UICollectionViewDiffableDataSourceReorderingHandlers`still has no properties?

rolfbjarne:
Closing, this is mostly done now, and in any case Xcode 12 was a long time ago.

PR #9095, opened 2020-07-15
[foundation] Update for Xcode12 beta 2

spouliot:
Some backlog from Xcode11 still exists and requires generator changes
Tracked in https://github.com/xamarin/xamarin-macios/issues/8943


src/foundation.cs
```diff
@@ -1309,6 +1313,26 @@ interface NSCoder {
...
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Export ("decodeArrayOfObjectsOfClasses:forKey:")]
+		[return: NullAllowed]
+		NSObject[] DecodeArrayOfObjects (NSSet<Class> classes, string key);
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Export ("decodeDictionaryWithKeysOfClass:objectsOfClass:forKey:")]
+		[return: NullAllowed]
+		NSDictionary DecodeDictionary (Class keyCls, Class objectCls, string key);
```
rolfbjarne, line 1330:
```suggestion
		NSDictionary DecodeDictionary (Class keyClass, Class objectClass, string key);
```

src/foundation.cs
```diff
@@ -1309,6 +1313,26 @@ interface NSCoder {
 		[iOS (9,0)][Mac (10,11)]
 		[NullAllowed, Export ("error", ArgumentSemantic.Copy)]
 		NSError Error { get; }
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Export ("decodeArrayOfObjectsOfClass:forKey:")]
+		[return: NullAllowed]
+		NSObject[] DecodeArrayOfObjects (Class cls, string key);
```
rolfbjarne, line 1320: I believe the guidelines say that parameters shouldn't be abbreviated:

```suggestion
		NSObject[] DecodeArrayOfObjects (Class @class, string key);
```

src/foundation.cs
```diff
@@ -2497,6 +2521,30 @@ interface NSKeyedUnarchiver {
 		[Export ("requiresSecureCoding")]
 		bool GetRequiresSecureCoding ();
 
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("unarchivedArrayOfObjectsOfClass:fromData:error:")]
+		[return: NullAllowed]
+		NSObject[] GetUnarchivedArray (Class cls, NSData data, [NullAllowed] out NSError error);
```
rolfbjarne, line 2529:
```suggestion
		NSObject[] GetUnarchivedArray (Class @class, NSData data, [NullAllowed] out NSError error);
```

src/foundation.cs
```diff
@@ -2497,6 +2521,30 @@ interface NSKeyedUnarchiver {
...
+		[Static]
+		[Export ("unarchivedArrayOfObjectsOfClasses:fromData:error:")]
+		[return: NullAllowed]
+		NSObject[] GetUnarchivedArray (NSSet<Class> classes, NSData data, [NullAllowed] out NSError error);
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("unarchivedDictionaryWithKeysOfClass:objectsOfClass:fromData:error:")]
+		[return: NullAllowed]
+		NSDictionary GetUnarchivedDictionary (Class keyCls, Class valueCls, NSData data, [NullAllowed] out NSError error);
```
rolfbjarne, line 2541:
```suggestion
		NSDictionary GetUnarchivedDictionary (Class keyClass, Class valueClass, NSData data, [NullAllowed] out NSError error);
```

src/foundation.cs
spouliot: I'm always ambivalent between _better_ or original parameter names. Been thinking of an xtro rules to check them - but it would be too noisy (and in general less good).

PR #9096, opened 2020-07-15
[monotouch-test] Fix broken UrlRequestTest.Mutability_30744 test.

rolfbjarne:
This test is supposed to assert that setting a header on an immutable
dictionary throws an exception.

The problem is that the tested code always throws an exception: either when
setting the header, or when asserting right after because no exception was
thrown, effectively always passing.

This showed up with current versions of NUnit[Lite], which keeps track of
assertion failures, and fails a test even if those assertion
failures/exceptions are caught in exception handlers.

So remove the whole exception verification logic, and just try to set the
value. We'll see if the test start failing somewhere (I've already tested both
on device and simulator, and no problems found so far).

### DIFF
--- a/tests/monotouch-test/Foundation/UrlRequestTest.cs
+++ b/tests/monotouch-test/Foundation/UrlRequestTest.cs
@@ -38,14 +38,7 @@ namespace MonoTouchFixtures.Foundation {
 				// that would crash on devices
 				// NSInternalInconsistencyException -[__NSCFDictionary setObject:forKey:]: mutating method sent to immutable object
 				if (Runtime.Arch == Arch.SIMULATOR) {
-					bool native_exception = false;
-					try {
-						mur.Headers.SetValueForKey (s3, s1);
-						Assert.Fail ("exception immutability");
-					} catch {
-						native_exception = true;
-					}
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/Foundation/UrlRequestTest.cs
```diff
@@ -38,14 +38,7 @@ public void Mutability_30744 ()
 				// that would crash on devices
 				// NSInternalInconsistencyException -[__NSCFDictionary setObject:forKey:]: mutating method sent to immutable object
 				if (Runtime.Arch == Arch.SIMULATOR) {
```
spouliot, line 40: > I've already tested both on device and simulator

with the condition enabled ?

rolfbjarne: Hmm, good point. Seems like my device testing was totally unnecessary.

Let me remove that condition and see what happens.

rolfbjarne: The test runs fine on device, both iOS and tvOS, with the condition removed, so I just pushed that change.

PR #9099, opened 2020-07-15
[authenticationservices] Update for Xcode 12 beta 2

spouliot:


src/authenticationservices.cs
```diff
@@ -743,4 +778,148 @@ interface ASWebAuthenticationSessionWebBrowserSessionManager {
...
+	[Unavailable (PlatformName.MacCatalyst)][Advice ("This API is not available when using Catalyst on macOS.")]
+	[BaseType (typeof (NSExtensionContext))]
+	[DisableDefaultCtor]
+	interface ASAccountAuthenticationModificationExtensionContext {
+
+		[Export ("getSignInWithAppleUpgradeAuthorizationWithState:nonce:completionHandler:")]
+		void GetSignInWithAppleUpgradeAuthorization ([NullAllowed] string state, [NullAllowed] string nonce, Action<ASAuthorizationAppleIdCredential, NSError> completionHandler);
+
+		[Export ("completeUpgradeToSignInWithAppleWithUserInfo:")]
+		void CompleteUpgradeToSignInWithApple ([NullAllowed] NSDictionary userInfo);
```
mandel-macaque, line 851: Do we need the 'WithApple'?

src/authenticationservices.cs
```diff
@@ -743,4 +778,148 @@ interface ASWebAuthenticationSessionWebBrowserSessionManager {
...
+		void ConvertAccountToSignInWithAppleWithoutUserInteraction (ASCredentialServiceIdentifier serviceIdentifier, ASPasswordCredential existingCredential, [NullAllowed] NSDictionary userInfo);
+
+		[Export ("prepareInterfaceToConvertAccountToSignInWithAppleForServiceIdentifier:existingCredential:userInfo:")]
+		void PrepareInterfaceToConvertAccountToSignInWithApple (ASCredentialServiceIdentifier serviceIdentifier, ASPasswordCredential existingCredential, [NullAllowed] NSDictionary userInfo);
+
+		[Export ("changePasswordWithoutUserInteractionForServiceIdentifier:existingCredential:newPassword:userInfo:")]
+		void ChangePasswordWithoutUserInteraction (ASCredentialServiceIdentifier serviceIdentifier, ASPasswordCredential existingCredential, string newPassword, [NullAllowed] NSDictionary userInfo);
+
+		[Export ("prepareInterfaceToChangePasswordForServiceIdentifier:existingCredential:newPassword:userInfo:")]
+		void PrepareInterfaceToChangePassword (ASCredentialServiceIdentifier serviceIdentifier, ASPasswordCredential existingCredential, string newPassword, [NullAllowed] NSDictionary userInfo);
```
mandel-macaque, line 919: Same, WithApple seems redundant. Definitely not a reason to block the PR.

src/authenticationservices.cs
```diff
@@ -490,6 +506,10 @@ enum ASAuthorizationProviderAuthorizationOperation {
 		[DefaultEnumValue]
 		[Field (null)]
 		None,
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Field ("ASAuthorizationProviderAuthorizationOperationConfigurationRemoved")]
+		Removed,
```
rolfbjarne, line 512: Shouldn't this be `ConfigurationRemoved`?

That's what swift does: https://developer.apple.com/documentation/authenticationservices/asauthorizationproviderauthorizationoperation/3566366-configurationremoved

src/authenticationservices.cs
rolfbjarne: Yes, we need it, because it's the name of a specific way to sign in.

src/authenticationservices.cs
```diff
@@ -743,4 +778,148 @@ interface ASWebAuthenticationSessionWebBrowserSessionManager {
...
+
+	[iOS (14,0)]
+	[NoWatch, NoTV, NoMac]
+	[Unavailable (PlatformName.MacCatalyst)][Advice ("This API is not available when using Catalyst on macOS.")]
+	[BaseType (typeof (NSExtensionContext))]
+	[DisableDefaultCtor]
+	interface ASAccountAuthenticationModificationExtensionContext {
+
+		[Export ("getSignInWithAppleUpgradeAuthorizationWithState:nonce:completionHandler:")]
+		void GetSignInWithAppleUpgradeAuthorization ([NullAllowed] string state, [NullAllowed] string nonce, Action<ASAuthorizationAppleIdCredential, NSError> completionHandler);
```
rolfbjarne, line 848: `[Async]`?

src/authenticationservices.cs
spouliot: The feature name is "Sign in with Apple", so it's an all or nothing

src/authenticationservices.cs
spouliot: same

PR #9103, opened 2020-07-16
[xharness] Fix finding testable types in test assemblies. Fixes #xamarin/maccore@2267.

rolfbjarne:
Apparently a null namespace when using Reflection can be an empty namespace in Cecil.

Fixes https://github.com/xamarin/maccore/issues/2267.


rolfbjarne:
build

PR #9109, opened 2020-07-16
[apitest] Refactor tests to not be async.

rolfbjarne:
The macOS tests will soon be upgraded to use the official NUnit[Lite], and in
that case we can't use async tests, because they end up deadlocking the
processes.

So refactor the two tests this affect to not be async, but instead use other
methods of accomplishing the same thing.

### DIFF
--- a/tests/apitest/src/Foundation/NSObject.cs
+++ b/tests/apitest/src/Foundation/NSObject.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Threading.Tasks;
+using System.Threading;
 using NUnit.Framework;
 
 using AppKit;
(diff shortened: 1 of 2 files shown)


tests/apitest/src/Foundation/NSObject.cs
```diff
@@ -13,17 +13,12 @@ namespace Xamarin.Mac.Tests
...
 			NSApplication.SharedApplication.Invoke (() => hit = true, 1);
-			// Wait for 10 second, then give up
-			for (int i = 0; i < 1000; ++i) {
-				if (hit)
-					return;
-				await Task.Delay (10);
-			}
-			Assert.Fail ("Did not see events after 10 second");
+			TestRuntime.RunAsync (TimeSpan.FromSeconds (10), () => { }, () => hit);
+			Assert.IsTrue (hit, "Did not see events after 10 second");
```
whitneyschmidt, line 21:
```suggestion
			Assert.IsTrue (hit, "Did not see events after 10 seconds");
```

PR #9115, opened 2020-07-17
[tests] Remove common.mk, it doesn't look like this file is used anymore.

rolfbjarne:


rolfbjarne:
Test failures on older macOS versions are all unrelated (and will be fixed in #9116).

PR #9124, opened 2020-07-17
[coreml] Update for Xcode 12 beta 2

spouliot:


src/coreml.cs
```diff
@@ -968,6 +985,18 @@ interface MLUpdateTask {
 		[return: NullAllowed]
 		MLUpdateTask Create (NSUrl modelUrl, IMLBatchProvider trainingData, [NullAllowed] MLModelConfiguration configuration, MLUpdateProgressHandlers progressHandlers, [NullAllowed] out NSError error);
 
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("updateTaskForModelAtURL:trainingData:completionHandler:error:")]
+		[return: NullAllowed]
+		MLUpdateTask Create (NSUrl modelUrl, IMLBatchProvider trainingData, Action<MLUpdateContext> completionHandler, [NullAllowed] out NSError error);
```
whitneyschmidt, line 992: `[Async]` for completion handler?

src/coreml.cs
```diff
@@ -968,6 +985,18 @@ interface MLUpdateTask {
...
+
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Static]
+		[Export ("updateTaskForModelAtURL:trainingData:progressHandlers:error:")]
+		[return: NullAllowed]
```
whitneyschmidt, line 997: '[Async]`?

src/coreml.cs
spouliot: `[Async]` only works when the completion handler is the last argument

src/coreml.cs
spouliot: progress != completion
you don't want to wait on something that's called multiple times

PR #9125, opened 2020-07-17
[CoreMIDI] Add support for Xcode 12 beta 2.

mandel-macaque:


src/coremidi.cs
```diff
@@ -31,7 +31,16 @@
 using CoreGraphics;
 using ObjCRuntime;
 
+using MidiObjectRef = System.Int32;
+
 namespace CoreMidi {
+
+	[Mac (10, 16), iOS (14, 0)]
+	public enum MidiProtocolID {
```
spouliot, line 39: `Id` because it's not an acronym ;-)

src/coremidi.cs
```diff
@@ -177,13 +190,25 @@ interface MidiCIProfileState : NSSecureCoding
 		[Export ("disabledProfiles")]
 		MidiCIProfile[] DisabledProfiles { get; }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message : "Use 'ctr (byte midiChannel, MidiCIProfile[] enabled, MidiCIProfile[] disabled)' instead.")]
```
spouliot, line 193: `.ctor` is how the metadata is named

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
 
 		[NullAllowed, Export ("propertyChangedCallback", ArgumentSemantic.Assign)]
 		MidiCIPropertyChangedHandler PropertyChangedCallback { get; set; }
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("initWithDiscoveredNode:dataReadyHandler:disconnectHandler:")]
+		IntPtr Constructor (MidiCIDiscoveredNode discoveredNode, Action handler, MidiCISessionDisconnectHandler disconnectHandler);
```
spouliot, line 259: `handler` -> `dataReadyHandler` so you don't have to look up the docs :)

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+		[Export ("sendProfile:onChannel:profileData:")]
+		bool SendProfile (MidiCIProfile profile, byte channel, NSData profileSpecificData);
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("deviceInfo")]
+		MidiCIDeviceInfo DeviceInfo { get; }
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("maxSysExSize")]
+		NSNumber MaxSysExSize { get; }
```
spouliot, line 271: do we know the underlying type ? i.e. `[BindAs]`

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("maxPropertyRequests")]
+		NSNumber MaxPropertyRequests { get; }
```
spouliot, line 275: do we know the underlying type ? i.e. `[BindAs]`

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+
+		[Internal]
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("midiDestination")]
+		MidiObjectRef _MidiDestination { get; }
+
+		[Mac (10, 16), iOS (14, 0)]
+		MidiEndpoint MidiDestination {
+			[Wrap ("new MidiEndpoint (_MidiDestination)")]
```
spouliot, line 284: ouch, that returns a new instance each time the property is accessed :(
bad for a watch in the debugger (among other things)

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+		MidiCIProfileSpecificDataHandler ProfileSpecificDataHandler { get; set; }
 	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[BaseType (typeof (NSObject), Name="MIDICIDeviceInfo")]
+	[DisableDefaultCtor]
+	interface MidiCIDeviceInfo : NSSecureCoding
+	{
+		[Export ("manufacturerID")]
+		NSData ManufacturerID { get; }
```
spouliot, line 299: `Id`

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+
+		[Export ("revisionLevel")]
+		NSData RevisionLevel { get; }
+
+		[Internal]
+		[Export ("midiDestination")]
+		MidiObjectRef _MidiDestination { get; }
+
+		MidiEndpoint MidiDestination {
+			[Wrap ("new MidiEndpoint (_MidiDestination)")]
```
spouliot, line 315: same

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+			get;
+		}
+
+		[Internal]
+		[Export ("initWithDestination:manufacturer:family:model:revision:")]
+		IntPtr Constructor (MidiObjectRef midiDestination, NSData manufacturer, NSData family, NSData modelNumber, NSData revisionLevel);
+
+		[Wrap ("this (midiDestination.Handle, manufacturer, family, modelNumber, revisionLevel)")]
```
spouliot, line 323: `midiDestination.Handle` -> NRE if `null`

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+	[BaseType (typeof (NSObject), Name="MIDICIDiscoveredNode")]
+	[DisableDefaultCtor]
+	interface MidiCIDiscoveredNode
+	{
+		[Internal]
+		[Export ("destination")]
+		MidiObjectRef _Destination { get; }
+
+		MidiEndpoint Destination { 
+			[Wrap ("new MidiEndpoint (_Destination)")]
```
spouliot, line 337: same

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+		MidiCIDeviceInfo DeviceInfo { get; }
+
+		[Export ("supportsProfiles")]
+		bool SupportsProfiles { get; }
+
+		[Export ("supportsProperties")]
+		bool SupportsProperties { get; }
+
+		[Export ("maximumSysExSize")]
+		NSNumber MaximumSysExSize { get; }
```
spouliot, line 351: `[BindAs]`

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+	}
+
+	delegate void MidiCIDiscoveryResponseDelegate (MidiCIDiscoveredNode[] arg0);
+
+	[Mac (10, 16), iOS (14, 0)]
+	[BaseType (typeof (NSObject), Name="MIDICIDiscoveryManager")]
+	interface MidiCIDiscoveryManager
```
spouliot, line 358: `[DisableDefaultCtor]` ?
in most case a `sharedInstance` means it's not meant to be user-created

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+		[Export ("handleDataForProfile:onChannel:data:")]
+		void HandleData (MidiCIProfile profile, byte channel, NSData inData);
+	}
+
+	[Mac (10, 16), iOS (14, 0)]
+	[BaseType (typeof (NSObject), Name="MIDICIResponder")]
+	[DisableDefaultCtor]
+	interface MidiCIResponder
+	{
+		// MIDICIInitiatiorMUID is a type alias to NSNumber, we do not have much more info
```
spouliot, line 395: seems to be 32 bits
https://www.midi.org/forum/5674-question-after-reading-the-midi-ci-1-1-spec-re-muids
or even 28 bits is you read further :)

tests/xtro-sharpie/Helpers.cs
```diff
@@ -53,6 +53,7 @@ public static partial class Helpers {
 			{ "UITableViewCellAccessoryType", "UITableViewCellAccessory" },
 			{ "UITableViewCellStateMask", "UITableViewCellState" },
 			{ "WatchKitErrorCode", "WKErrorCode" }, // WebKit already had that name
+			{ "MIDIProtocolID", "MidiProtocolID"},
```
spouliot, line 56: `Id`

src/coremidi.cs
rolfbjarne: I think something like this would be even better:

> "Use the '(byte midiChannel, MidiCIProfile[] enabled, MidiCIProfile[] disabled)' constructor instead."

src/coremidi.cs
```diff
@@ -177,13 +190,25 @@ interface MidiCIProfileState : NSSecureCoding
...
 		MidiCIProfile[] DisabledProfiles { get; }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message : "Use 'ctr (byte midiChannel, MidiCIProfile[] enabled, MidiCIProfile[] disabled)' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message : "Use 'ctr (byte midiChannel, MidiCIProfile[] enabled, MidiCIProfile[] disabled)' instead.")]
 		[Export ("initWithEnabledProfiles:disabledProfiles:")]
 		IntPtr Constructor (MidiCIProfile[] enabled, MidiCIProfile[] disabled);
+
+		[Mac (10, 16), iOS (14, 0)]
+		[Export ("initWithChannel:enabledProfiles:disabledProfiles:")]
+		IntPtr Constructor (byte midiChannelNum, MidiCIProfile[] enabled, MidiCIProfile[] disabled);
```
rolfbjarne, line 200:
```suggestion
		IntPtr Constructor (byte midiChannelNumber, MidiCIProfile[] enabled, MidiCIProfile[] disabled);
```

src/coremidi.cs
```diff
@@ -228,5 +253,172 @@ interface MidiCISession
...
+		bool SupportsProfiles { get; }
+
+		[Export ("supportsProperties")]
+		bool SupportsProperties { get; }
+
+		[Export ("maximumSysExSize")]
+		NSNumber MaximumSysExSize { get; }
+	}
+
+	delegate void MidiCIDiscoveryResponseDelegate (MidiCIDiscoveredNode[] arg0);
```
rolfbjarne, line 354:
```suggestion
	delegate void MidiCIDiscoveryResponseDelegate (MidiCIDiscoveredNode[] discoveredNodes);
```

src/coremidi.cs
dalexsoto: We use the `.ctor (Foo, Bar)` on other places but not opposed to change or to use

> "Use the '.ctor (byte midiChannel, MidiCIProfile[] enabled, MidiCIProfile[] disabled)' overload instead".

src/coremidi.cs
dalexsoto: Not sure about the headers but the [docs](https://developer.apple.com/documentation/coremidi/midicisession/3580337-maxsysexsize) are not very helpful unfortunately 😞 maybe have it now as `NSNumber` and add it to the list https://github.com/xamarin/xamarin-macios/issues/8943 to check at the end of the summer? What do you think @spouliot ?

src/coremidi.cs
dalexsoto: Same, [docs](https://developer.apple.com/documentation/coremidi/midicisession/3580336-maxpropertyrequests) are not very helpful, same idea as above

>  maybe have it now as `NSNumber` and add it to the list #8943 to check at the end of the summer? What do you think @spouliot ?

src/coremidi.cs
dalexsoto: heh nice find

src/coremidi.cs
mandel-macaque: I used ctr because I did see it in other places, I have no preference.. I'll take @rolfbjarne option

src/coremidi.cs
mandel-macaque: So, CoreMIDI is not present in the watch. But I am more worried of having a private copy of the MidiEndpoint, the issue is that the users might do something like calling the Dispose method on the property value and then have issues. 

If we do anything, I would move away from a property, do a GetMidiEndpoint or return the byte.

src/coremidi.cs
spouliot: The MIDI[-CI] specs are more useful

src/coremidi.cs
spouliot: `[BindAs (typeof (ulong))]` like the other one ?

mandel-macaque: I missed it, sorry

PR #9126, opened 2020-07-17
[Xcode12][GameKit] Update for beta 1-2

whitneyschmidt:


src/GameKit/GKLeaderboardSet.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace GameKit {
+	public partial class GKLeaderboardSet {
+
+		public void LoadLeaderboardsWithHandler (GKLeaderboardsHandler completionHandler)
+		{
+#if IOS || TVOS
+			if (UIKit.UIDevice.CurrentDevice.CheckSystemVersion (14, 0))
+#elif WATCH
+			if (WatchKit.WKInterfaceDevice.CurrentDevice.CheckSystemVersion (7, 0))
+#elif MONOMAC
```
spouliot, line 50: there's a `#if !MONOMAC` earlier

src/GameKit/GKLeaderboardSet.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System;
+using Foundation;
+using ObjCRuntime;
+using UIKit;
+
+namespace GameKit {
+	public partial class GKLeaderboardSet {
+
+		public void LoadLeaderboardsWithHandler (GKLeaderboardsHandler completionHandler)
```
spouliot, line 44: remove `WithHandler` since it's implied

spouliot: ideally there would be a `[Async]` version of the API
It will need to be manually added in this case

src/GameKit/GameKit.cs
```diff
@@ -291,11 +291,43 @@ public enum GKInviteRecipientResponse : long {
 		NoAnswer = 5,
 	}
 
-	[Mac (10,13,4), TV (11,3), iOS (11,3)]
+	[iOS (11,3)][Obsoleted (PlatformName.iOS, 14,0, message: "Do not use; this API was removed.")]
+	[Mac (10,13,4)][Obsoleted (PlatformName.MacOSX, 10,16, message: "Do not use; this API was removed.")]
+	[TV (11,3)][Obsoleted (PlatformName.TvOS, 14,0, message: "Do not use; this API was removed.")]
+	[Obsolete]
```
spouliot, line 297: we normally use `[Obsolete]`, with a message, only for our binding _mistakes_
and remove them under `#if !XAMCORE_4_0`
so customers can easily spot what they should avoid for whenever (and if) we do breaking changes

tests/xtro-sharpie/macOS-GameKit.ignore
```diff
@@ -45,3 +45,5 @@
 !missing-null-allowed! 'System.Void GameKit.GKDialogController::set_ParentWindow(AppKit.NSWindow)' is missing an [NullAllowed] on parameter #0
 !missing-null-allowed! 'System.Void GameKit.GKGameSession::SendMessage(System.String,System.String[],Foundation.NSData,GameKit.GKCloudPlayer[],System.Boolean,System.Action`1<Foundation.NSError>)' is missing an [NullAllowed] on parameter #2
 !missing-null-allowed! 'System.Void GameKit.GKMatchmakerViewController::set_DefaultInvitationMessage(System.String)' is missing an [NullAllowed] on parameter #0
+!missing-null-allowed! 'System.String[] GameKit.GKMatch::get_PlayersIDs()' is missing an [NullAllowed] on return type
+!missing-null-allowed! 'System.String[] GameKit.GKVoiceChat::get_PlayerIDs()' is missing an [NullAllowed] on return type
```
spouliot, line 49: same

tests/xtro-sharpie/watchOS-GameKit.ignore
```diff
@@ -79,3 +79,4 @@
 
 # Initial result from new rule missing-null-allowed
 !missing-null-allowed! 'System.String GameKit.GKTurnBasedParticipant::get_PlayerID()' is missing an [NullAllowed] on return type
+!missing-null-allowed! 'System.String GameKit.GKAchievement::get_PlayerID()' is missing an [NullAllowed] on return type
```
spouliot, line 82: same

src/gamekit.cs
```diff
@@ -268,14 +276,27 @@ interface GKSession {
 	}
 
 	[Watch (3,0)]
+	[DisableDefaultCtor]
 	[BaseType (typeof (NSObject))]
 	interface GKLeaderboard {
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' instead.")]
```
spouliot, line 285: are the parameters useful ?

src/gamekit.cs
```diff
@@ -290,21 +311,44 @@ interface GKLeaderboard {
...
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' to obtain scores instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries (GKLeaderboardPlayerScope, GKLeaderboardTimeScope, NSRange, GKEntriesForPlayerScopeHandler)' to obtain scores instead.")]
 		[Export ("localPlayerScore", ArgumentSemantic.Retain)]
 		GKScore LocalPlayerScore { get;  }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Do not instantiate GKLeaderboard directly. Use class method 'LoadLeaderboards' to get the leaderboards instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Do not instantiate GKLeaderboard directly. Use class method 'LoadLeaderboards' to get the leaderboards instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Do not instantiate GKLeaderboard directly. Use class method 'LoadLeaderboards' to get the leaderboards instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Do not instantiate GKLeaderboard directly. Use class method 'LoadLeaderboards' to get the leaderboards instead.")]
```
spouliot, line 338: I think `Use 'LoadLeaderboards' instead.` is enough

src/gamekit.cs
```diff
@@ -376,10 +486,21 @@ interface GKLeaderboardSet : NSCoding, NSSecureCoding {
 		[Async]
 		void LoadLeaderboardSets ([NullAllowed] GKLeaderboardSetsHandler completionHandler);
 
+		[Internal]
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'LoadLeaderboardsWithCompletionHandler' instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadLeaderboardsWithCompletionHandler' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadLeaderboardsWithCompletionHandler' instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadLeaderboardsWithCompletionHandler' instead.")]
```
spouliot, line 493: if `[Internal]` then the message is not needed since it will never be shown to customers

src/gamekit.cs
```diff
@@ -481,8 +602,16 @@ interface GKPlayer : NSSecureCoding {
 		[TV (13,0)][Mac (10,15)][iOS (13,0)]
 		[Export ("scopedIDsArePersistent")]
 		bool ScopedIdsArePersistent { get; }
+
+		[TV (14, 0), NoWatch, Mac (10, 16), iOS (14, 0)]
+		[Field ("GKPlayerIDNoLongerAvailable")]
+		NSString GKPlayerIdNoLongerAvailable { get; }
```
spouliot, line 608: we normally remove the prefix on fields, e.g. `IdNoLongerAvailable` when inside the type `GKPlayer`

src/gamekit.cs
```diff
@@ -803,6 +939,11 @@ interface GKLocalPlayer
 		[Static]
 		[Export ("local")]
 		GKLocalPlayer Local { get; }
+
+		[NoWatch]
+		[TV (14,0), NoWatch, Mac (10,16), iOS (14,0)]
```
spouliot, line 944: `NoWatch` twice present

src/gamekit.cs
```diff
@@ -2456,4 +2629,91 @@ interface GKSessionDelegate {
 		[Export ("session:didFailWithError:")]
 		void FailedWithError (GKSession session, NSError error);
 	}
+
+	[TV (14,0), Mac (10,16), iOS (14,0)]
+	[NoWatch]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 2635: `[DisableDefaultCtor]` since there's a `shared` method
unless docs/headers mention you can create your own ?!?

src/gamekit.cs
```diff
@@ -2456,4 +2629,91 @@ interface GKSessionDelegate {
...
+	interface GKAccessPoint
+	{
+		[Static]
+		[Export ("shared")]
+		GKAccessPoint Shared { get; }
+
+		[Export ("active")]
+		bool Active { [Bind ("isActive")] get; set; }
+
+		[TV (14,0), NoMac, NoiOS]
```
spouliot, line 2645: no need for `[TV (14,0)]` since it's already on the type itself

src/GameKit/GKLeaderboardSet.cs
whitneyschmidt: Unfortunately there's already an API named `LoadLeaderboards` in `GKLeaderboardSet` namespace.

The group of related APIs have this (not great) naming scheme:

1) `LoadLeaderboards` - deprecated, generated
2) `LoadLeaderboardsWithCompletionHandler` - new! generated
3) `LoadLeaderboardsWithHandler` - new! manual wrapper for the above

So we can change the names of 2 & 3 but not 1. 

Ideas for other names very welcome!

spouliot: are the parameter names blocking the use of overloads ?
if so then I don't have any better idea atm

src/gamekit.cs
whitneyschmidt: good question! There are 2 API that are both named `LoadEntries`, but since these are both new and have different parameter lists I think it is okay to assume that users can figure out which one they need on their own.

Removing parameters :)

src/gamekit.cs
whitneyschmidt: Doesn't look like you can create one 👍

src/GameKit/GKLeaderboardSet.cs
whitneyschmidt: Yep, they all have the same signature, `(GKLeaderboardsHandler completionHandler)` 😕

src/GameKit/GKLeaderboardSet.cs
```diff
@@ -0,0 +1,60 @@
...
+//
+// Authors:
+//   Whitney Schmidt (whschm@microsoft.com)
+//
+// Copyright 2020, Microsoft Corp. All rights reserved.
+//
+// The class can be either constructed from a string (from user code)
+// or from a handle (from iphone-sharp.dll internal calls).  This
+// delays the creation of the actual managed string until actually
+// required
```
rolfbjarne, line 12: Is this part copied from somewhere else? It doesn't look like it's relevant to this file.

src/GameKit/GameKit.cs
```diff
@@ -291,11 +291,43 @@ public enum GKInviteRecipientResponse : long {
...
+
+	[TV (14,0), Mac (10,16), iOS (14,0)]
+	[NoWatch]
+	[Native]
+	public enum GKAccessPointLocation : long
+	{
+		TopLeading,
+		TopTrailing,
+		BottomLeading,
+		BottomTrailing
```
rolfbjarne, line 313:
```suggestion
		BottomTrailing,
```

src/gamekit.cs
```diff
@@ -347,13 +399,71 @@ interface GKLeaderboard {
...
+		[Export ("duration")]
+		double Duration { get; }
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("loadEntriesForPlayerScope:timeScope:range:completionHandler:")]
+		void LoadEntries (GKLeaderboardPlayerScope playerScope, GKLeaderboardTimeScope timeScope, NSRange range, [NullAllowed] GKEntriesForPlayerScopeHandler handler);
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("loadEntriesForPlayers:timeScope:completionHandler:")]
+		void LoadEntries (GKPlayer[] players, GKLeaderboardTimeScope timeScope, [NullAllowed] GKEntriesForPlayersHandler handler);
```
rolfbjarne, line 444: `[Async]`?

Also:

```suggestion
		void LoadEntries (GKPlayer[] players, GKLeaderboardTimeScope timeScope, [NullAllowed] GKEntriesForPlayersHandler completionHandler);
```

src/gamekit.cs
```diff
@@ -347,13 +399,71 @@ interface GKLeaderboard {
...
+		[Export ("baseLeaderboardID", ArgumentSemantic.Strong)]
+		string BaseLeaderboardId { get; }
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("duration")]
+		double Duration { get; }
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("loadEntriesForPlayerScope:timeScope:range:completionHandler:")]
+		void LoadEntries (GKLeaderboardPlayerScope playerScope, GKLeaderboardTimeScope timeScope, NSRange range, [NullAllowed] GKEntriesForPlayerScopeHandler handler);
```
rolfbjarne, line 440: `[Async]`?

Also:

```suggestion
		void LoadEntries (GKLeaderboardPlayerScope playerScope, GKLeaderboardTimeScope timeScope, NSRange range, [NullAllowed] GKEntriesForPlayerScopeHandler completionHandler);
```

src/gamekit.cs
```diff
@@ -376,10 +486,21 @@ interface GKLeaderboardSet : NSCoding, NSSecureCoding {
...
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadLeaderboardsWithCompletionHandler' instead.")]
 		[Export ("loadLeaderboardsWithCompletionHandler:")]
 		[Async]
 		void LoadLeaderboards ([NullAllowed] GKLeaderboardsHandler completionHandler);
 
+		[Internal]
+		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Export ("loadLeaderboardsWithHandler:")]
+		[Async]
+		void LoadLeaderboardsWithCompletionHandler ([NullAllowed] GKLeaderboardsHandler handler); // ideas for naming scheme here?
```
rolfbjarne, line 502: Is there any explanation why there's a new method? I could be that we'd have to expose both, and let the customers start using the new one whenever they're ready (if the callback returns different data for instance).

src/gamekit.cs
```diff
@@ -569,6 +698,13 @@ interface GKScore : NSSecureCoding {
 		[Async]
 		void ReportScores (GKScore[] scores, [NullAllowed] GKChallenge[] challenges, [NullAllowed] Action<NSError> completionHandler);
 
+		[NoWatch]
+		[iOS (14,0)][Mac (10,16)][TV (14,0)]
+		[Static]
+		[Async]
+		[Export ("reportLeaderboardScores:withEligibleChallenges:withCompletionHandler:")]
+		void ReportLeaderboardScores (GKLeaderboardScore[] scores, GKChallenge[] challenges, [NullAllowed] Action<NSError> completionHandler);
```
rolfbjarne, line 706: Very minor, but this sounds a bit more descriptive:

```suggestion
		void ReportLeaderboardScores (GKLeaderboardScore[] scores, GKChallenge[] eligibleChallenges, [NullAllowed] Action<NSError> completionHandler);
```

src/GameKit/GameKit.cs
whitneyschmidt: Updated these to `Deprecated` w/ a message to let users know that these APIs no longer exist since Apple removed them. Lmk if there's a better way to do this!

src/GameKit/GKLeaderboardSet.cs
whitneyschmidt: Removed.

src/gamekit.cs
whitneyschmidt: The API docs look exactly the same. The headers are slightly different:

```
/// Loads array with all classic leaderboards and current instances of recurring leaderboards for this leaderboardSet
/// Possible reasons for error:
/// 1. Communications problem
/// 2. Unauthenticated player
- (void)loadLeaderboardsWithHandler:(void(^)(NSArray<GKLeaderboard *> * __nullable leaderboards, NSError * __nullable error))handler API_AVAILABLE(ios(14.0), macos(10.16), tvos(14.0), watchos(7.0));

@end

@interface GKLeaderboardSet (Deprecated)

/// Loads array with all leaderboards for the leaderboardSet
/// Possible reasons for error:
/// 1. Communications problem
/// 2. Unauthenticated player
- (void)loadLeaderboardsWithCompletionHandler:(void(^__nullable)(NSArray<GKLeaderboard *> * __nullable leaderboards, NSError * __nullable error))completionHandler API_DEPRECATED("Use loadLeaderboardsWithHandler: instead.", ios(7.0, 14.0), tvos(9.0, 14.0), macosx(10.10, 10.16), watchos(3.0, 7.0));
```

rolfbjarne: That didn't explain much :/

I think the safer option would be to not hide these API behind a version check, and instead expose both of them, because we don't know the difference between them. It could be that they're not interchangeable.

whitneyschmidt: The documentation does says to use the newer API instead of the deprecated one, but I see how it's better to give users more transparency if they're not the same 👍

https://developer.apple.com/documentation/gamekit/gkleaderboardset/1451810-loadleaderboardswithcompletionha

src/gamekit.cs
whitneyschmidt: Adding `[Async]` here causes:

```
error BI1077: bgen: Async method Void LoadEntries(GameKit.GKLeaderboardPlayerScope, GameKit.GKLeaderboardTimeScope, Foundation.NSRange, GameKit.GKEntriesForPlayerScopeHandler) with more than one result parameter in the callback by neither ResultTypeName or ResultType
```

src/gamekit.cs
whitneyschmidt: Same, error:

```
error BI1077: bgen: Async method Void LoadEntries(GameKit.GKPlayer[], GameKit.GKLeaderboardTimeScope, GameKit.GKEntriesForPlayersHandler) with more than one result parameter in the callback by neither ResultTypeName or ResultType
```

src/gamekit.cs
```diff
@@ -268,14 +276,27 @@ interface GKSession {
...
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadEntries' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries' instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries' instead.")]
 		[Export ("playerScope", ArgumentSemantic.Assign)]
 		GKLeaderboardPlayerScope PlayerScope { get; set;  }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'LoadEntries' to obtain the size of the leaderboard.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadEntries' to obtain the size of the leaderboard.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries' to obtain the size of the leaderboard.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries' to obtain the size of the leaderboard.")]
```
spouliot, line 299: better share the existing string `"Use 'LoadEntries' instead."`

src/gamekit.cs
```diff
@@ -290,21 +311,44 @@ interface GKLeaderboard {
...
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadEntries' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries' instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries' instead.")]
 		[Export ("range", ArgumentSemantic.Assign)]
 		NSRange Range { get; set;  }
 
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'LoadEntries' to obtain scores instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'LoadEntries' to obtain scores instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'LoadEntries' to obtain scores instead.")]
+		[Deprecated (PlatformName.WatchOS, 7, 0, message: "Use 'LoadEntries' to obtain scores instead.")]
```
spouliot, line 324: same

src/gamekit.cs
```diff
@@ -1986,15 +2131,34 @@ interface GKGameCenterViewController
...
 		NSObject WeakDelegate { get; set; }
 
 		[Wrap ("WeakDelegate")]
 		[Protocolize]
 		GKGameCenterControllerDelegate Delegate { get; set;  }
 
-		[NoTV]
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'initWithState' instead")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'initWithState' instead")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'initWithState' instead")]
```
spouliot, line 2160: that looks like a selector and not the managed name

spouliot: also messages should end with `.`

src/gamekit.cs
```diff
@@ -2011,8 +2175,10 @@ interface GKGameCenterViewController
 		[Deprecated (PlatformName.MacOSX, 10, 10, message : "Use 'LeaderboardIdentifier' instead.")]
 		string LeaderboardCategory { get; set; }
 
-		[NoTV]
-		[iOS (7,0)][Mac (10,10)] // Marked 10.9 in header, apple 17612948
+		[Deprecated (PlatformName.iOS, 14, 0, message: "Use 'initWithLeaderboard' instead.")]
+		[Deprecated (PlatformName.TvOS, 14, 0, message: "Use 'initWithLeaderboard' instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 16, message: "Use 'initWithLeaderboard' instead.")]
```
spouliot, line 2180: same, selector vs managed name

src/gamekit.cs
```diff
@@ -347,13 +399,71 @@ interface GKLeaderboard {
...
+		[Export ("baseLeaderboardID", ArgumentSemantic.Strong)]
+		string BaseLeaderboardId { get; }
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("duration")]
+		double Duration { get; }
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("loadEntriesForPlayerScope:timeScope:range:completionHandler:")]
+		void LoadEntries (GKLeaderboardPlayerScope playerScope, GKLeaderboardTimeScope timeScope, NSRange range, [NullAllowed] GKEntriesForPlayerScopeHandler completionHandler);
```
spouliot, line 440: `[Async]` ?

src/gamekit.cs
```diff
@@ -347,13 +399,71 @@ interface GKLeaderboard {
...
+
+		[Watch (7, 0), TV (14, 0), Mac (10, 16), iOS (14, 0)]
+		[Export ("loadEntriesForPlayers:timeScope:completionHandler:")]
+		void LoadEntries (GKPlayer[] players, GKLeaderboardTimeScope timeScope, [NullAllowed] GKEntriesForPlayersHandler completionHandler);
```
spouliot, line 444: `[Async]` ?

whitneyschmidt: Same error as above on adding `[Async]`: `BI1077`

src/gamekit.cs
whitneyschmidt: Adding `[Async]` yields this error:

```
error BI1077: bgen: Async method Void LoadEntries(GameKit.GKPlayer[], GameKit.GKLeaderboardTimeScope, GameKit.GKEntriesForPlayersHandler) with more than one result parameter in the callback by neither ResultTypeName or ResultType
```

PR #9134, opened 2020-07-20
[d16-8][msbuild] Changes GetMinimumOSVersionTaskBase to use ITaskItem for inputs

emaf:
From Visual Studio we need the input files to be  instead of plain strings to identify the files that need to be copied to the Mac


emaf:
@monojenkins backport main

PR #9139, opened 2020-07-21
Add a parameterless constructor for MTKView

mattleibow:
This is possible in Xcode:

```objc
MTKView* t = [[MTKView alloc] init];
```

But we only have the one ctor in C#:

```csharp
public MTKView (CGRect frameRect, IMTLDevice device)
```

This PR adds:

```csharp
public MTKView ()
```


src/metalkit.cs
```diff
@@ -39,6 +39,9 @@ interface MTKModel {
 #endif
 	interface MTKView : NSCoding, CALayerDelegate {
 
+		[Export ("init")]
+		IntPtr Constructor ();
```
spouliot, line 43: It's added by default. If not present it's because someone added `[DisableDefaultCtor]` (see above) and that's the line that should be removed.

We'll need to `git blame` why this was added before merging.

dalexsoto: The fact that it can be created does not mean it is usable per se, have you tried to do more than just `MTKView* t = [[MTKView alloc] init];`?

mattleibow: I'll try it, but I think it works... I know I did create a view with this way when testing. Maybe I was mistaken. I'll confirm.

spouliot: There's a `[NullAllowed]` on the `IMTLDevice` device of the designated initializer and you can, in theory, give it a `CGRect.Empty` so there's nothing (again in theory) that should block the default `init` from working. Of course a test to double check theory means being sure in practice 😸

rolfbjarne: git blame: https://github.com/xamarin/maccore/commit/5b9ab25cf8d807d5c2d8c92c424cb0b12875dfb5#diff-2c98c959c61fe409d8a9d47f642fcc230c3c2a37c760d18a80934cc46658e760R14

that didn't explain much, so we can try just removing the `[DisableDefaultCtor]` and see if the tests complain somewhere.

rolfbjarne (approved):
This works on my device, and introspection doesn't complain on device either (on an iOS 10.3.3), so this is good to go.

rolfbjarne:

PR #9140, opened 2020-07-21
[xharness] Allow for null environment variables to mlaunch.

rolfbjarne:
It's a valid scenario to set a null environment variable.

This happens when changing InCI to always return true (to test locally what's
done on the bots), in which case we try to forward BUILD_REVISION to mlaunch,
but BUILD_REVISION isn't necessarily set. Not forwarding it to mlaunch should
not cause problems in most tests, so just allow this scenario.


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Execution/Mlaunch/Arguments.cs
```diff
@@ -181,10 +181,11 @@ public sealed class SetEnvVariableArgument : MlaunchArgument {
 		public SetEnvVariableArgument (string variableName, object variableValue)
 		{
 			this.variableName = variableName ?? throw new ArgumentNullException (nameof (variableName));
-			this.variableValue = variableValue?.ToString () ?? throw new ArgumentNullException (nameof (variableValue));
 
-			if (variableValue is bool)
-				this.variableValue = this.variableValue.ToLower ();
+			if (variableValue is bool b)
+				this.variableValue = b.ToString ().ToLower ();
```
spouliot, line 186: `ToLowerInvariant ()` ?

PR #9143, opened 2020-07-21
[d16-8][msbuild] Avoids running GetMinimumOSVersion if no Mac is connected

emaf:
`GetMinimumOSVersion` should not be run if there's no Mac available because it depends on `_DetectSdkLocations` to calculate the value of `_SdkVersion`, and this will only happen if there's a Mac connected. This change fixes offline builds (plain IL builds) and Hot Restart builds on Windows.


emaf:
@monojenkins backport main

PR #9145, opened 2020-07-21
[devicecheck] Update for Xcode 12 beta 2

spouliot:


src/devicecheck.cs
```diff
@@ -20,7 +20,10 @@ namespace DeviceCheck {
 	[Native]
 	public enum DCError : long {
 		UnknownSystemFailure,
-		FeatureUnsupported
+		FeatureUnsupported,
```
rolfbjarne, line 23: I wonder how many years it'll take us to add a trailing comma everywhere...

spouliot: likely a few more once sharpie output them correctly ;-)
c.c. @dalexsoto

dalexsoto: heh maybe that would require some Roslyn 🙈

src/devicecheck.cs
```diff
@@ -44,4 +47,30 @@ interface DCDevice {
...
+		[Static]
+		[Export ("sharedService")]
+		DCAppAttestService SharedService { get; }
+
+		[Export ("supported")]
+		bool Supported { [Bind ("isSupported")] get; }
+
+		[Async]
+		[Export ("generateKeyWithCompletionHandler:")]
+		void GenerateKey (Action<NSString, NSError> completionHandler);
```
dalexsoto, line 66: Any particular reason for NSString here?

spouliot: no, missed it :)

PR #9147, opened 2020-07-21
[xharness] Fix compiler warning.

rolfbjarne:


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/XmlResultParser.cs
```diff
@@ -348,7 +348,7 @@ static TrxTests ParseTrxXml (XmlReader reader)
 		{
 			var rv = new TrxTests ();
 			var tests = new Dictionary<string, TrxTest> ();
-			TrxTest lastTest = null;
+			TrxTest? lastTest = null;
```
whitneyschmidt, line 351: I'm curious about what compiler warning is generated for the original null assignment 🤔 ❓

rolfbjarne: This is the warning:

> warning CS8600: Converting null literal or possible null value to non-nullable type.

It's because this file has enabled nullability.

PR #9151, opened 2020-07-21
[metrickit] Update for Xcode 12 beta 2

spouliot:


src/MetricKit/MXMetaData.cs
```diff
@@ -0,0 +1,21 @@
...
+using ObjCRuntime;
+using UIKit;
+
+namespace MetricKit {
+
+	public partial class MXMetaData {
+
+		public virtual NSDictionary DictionaryRepresentation {
+			get {
+				if (PlatformHelper.CheckSystemVersion (14,0))
```
rolfbjarne, line 14: This version check doesn't work on macOS nor watchOS. This framework might not be on neither platform right now, but I think we should consider the possibility that it might be in the future, and do something like:

```suggestion
#if __IOS__
				if (PlatformHelper.CheckSystemVersion (14, 0))
#else
#error Platform-specific logic not implemented
#endif
```

src/MetricKit/MXMetric.cs
```diff
@@ -0,0 +1,21 @@
...
+using ObjCRuntime;
+using UIKit;
+
+namespace MetricKit {
+
+	public partial class MXMetric {
+
+		public virtual NSDictionary DictionaryRepresentation {
+			get {
+				if (PlatformHelper.CheckSystemVersion (14,0))
```
rolfbjarne, line 14: Same

src/MetricKit/MXMetricPayload.cs
```diff
@@ -0,0 +1,21 @@
...
+using ObjCRuntime;
+using UIKit;
+
+namespace MetricKit {
+
+	public partial class MXMetricPayload {
+
+		public virtual NSDictionary DictionaryRepresentation {
+			get {
+				if (PlatformHelper.CheckSystemVersion (14,0))
```
rolfbjarne, line 14: Same

src/MetricKit/MXMetaData.cs
spouliot: It's an iOS only framework but it can't hurt :)

PR #9152, opened 2020-07-22
[Xcode12] HealthKit b1 update

whitneyschmidt:


src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+	{
+
+		[Export ("initWithElectrocardiogram:dataHandler:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (HKElectrocardiogram electrocardiogram, HKElectrocardiogramQueryDataHandler dataHandler);
+	}
+
+	[Watch (7,0), iOS (14,0)]
+	[BaseType (typeof (NSObject))]
+	interface HKElectrocardiogramVoltageMeasurement
```
spouliot, line 3410: `[DisableDefaultCtor]` since it seems only exposed by the delegate (no use to create your own)

src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+	{
+		[Export ("timeSinceSampleStart")]
+		double TimeSinceSampleStart { get; }
+
+		[Export ("quantityForLead:")]
+		[return: NullAllowed]
+		HKQuantity QuantityForLead (HKElectrocardiogramLead lead);
```
spouliot, line 3417: * no verb/action
* extra suffix implied by parameter

-> `GetQuantity`

src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+		nint MajorVersion { get; }
+
+		[Export ("minorVersion")]
+		nint MinorVersion { get; }
+
+		[Export ("patchVersion")]
+		nint PatchVersion { get; }
+
+		[Export ("FHIRRelease", ArgumentSemantic.Strong)]
+		string FHIRRelease { get; }
```
spouliot, line 3435: `FhirRelease`

src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+
+		[Export ("stringRepresentation")]
+		string StringRepresentation { get; }
+
+		[Static]
+		[Export ("versionFromVersionString:error:")]
+		[return: NullAllowed]
+		HKFHIRVersion VersionFromVersionString (string versionString, [NullAllowed] out NSError errorOut);
```
spouliot, line 3443: -> `GetVersion` (same reasons as before)

src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+
+		[Static]
+		[Export ("primaryDSTU2Version")]
+		HKFHIRVersion PrimaryDSTU2Version ();
```
spouliot, line 3447: Add `Get*`
or turn it into a property (look what swift is doing)

src/healthkit.cs
```diff
@@ -3105,4 +3371,84 @@ interface HKHeartbeatSeriesQuery
...
+
+		[Static]
+		[Export ("primaryR4Version")]
+		HKFHIRVersion PrimaryR4Version ();
```
spouliot, line 3451: same

src/healthkit.cs
whitneyschmidt: It makes sense to change these into properties since they're more or less getters - no way to modify, no params, returns a single value.

Can you explain more about what Swift is doing?

I took a look at the docs: https://docs.developer.st.apple.com/documentation/healthkit/hkfhirversion

As well as the objc header code.

src/HealthKit/Enums.cs
```diff
@@ -508,4 +517,88 @@ public enum HKAuthorizationRequestStatus : long
 		ShouldRequest,
 		Unnecessary,
 	}
+
+	[Watch (7,0), iOS (13,6)]
```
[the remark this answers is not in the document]
whitneyschmidt: This is accessible + builds/runs on iOS 13.6 sim in Xcode 11.6.

Release notes also (kinda) confirm to the extent that Apple release notes can "confirm" anything: https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-13_6-release-notes

It looks like the header isn't lying...

```
/*!
 @enum          HKCategoryValueAppetiteChanges
 @abstract      Set of values to indicate the direction of appetite changes.
 */
typedef NS_ENUM(NSInteger, HKCategoryValueAppetiteChanges) {
    HKCategoryValueAppetiteChangesUnspecified = 0,
    HKCategoryValueAppetiteChangesNoChange,
    HKCategoryValueAppetiteChangesDecreased,
    HKCategoryValueAppetiteChangesIncreased,
} API_AVAILABLE(ios(13.6), watchos(7.0));
```

src/HealthKit/Enums.cs
```diff
@@ -508,4 +517,88 @@ public enum HKAuthorizationRequestStatus : long
...
+		Version1 = 1,
+	}
+
+	[Watch (7,0), iOS (14,0)]
+	[Native]
+	public enum HKCategoryValueEnvironmentalAudioExposureEvent : long {
+		MomentaryLimit = 1,
+	}
+
+	[Watch (7,0), iOS (13,6)]
```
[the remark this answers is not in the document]
whitneyschmidt: Builds/runs in Xcode 11.6.

src/HealthKit/Enums.cs
```diff
@@ -508,4 +517,88 @@ public enum HKAuthorizationRequestStatus : long
...
+	[Native]
+	public enum HKCategoryValuePresence : long {
+		Present = 0,
+		NotPresent,
+	}
+
+	[Watch (7,0), iOS (13,6)]
```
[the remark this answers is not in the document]
whitneyschmidt: Builds/runs in Xcode 11.6.

src/healthkit.cs
```diff
@@ -1858,6 +1939,170 @@ enum HKCategoryTypeIdentifier
 		[Watch (6, 0), iOS (13, 0)]
 		[Field ("HKCategoryTypeIdentifierToothbrushingEvent")]
 	 	ToothbrushingEvent,
+
+		[Watch (7, 0), iOS (13, 6)]
```
whitneyschmidt, line 1943: I'm going through and validating all these `13,6` attributes

whitneyschmidt:
Waiting for tests to pass after 11.6 is merged.

src/healthkit.cs
spouliot: > It makes sense...

Yes, those are all good points to make a property over a method 😄 

There are also _less visible_ aspects that are important. One of them is performance, IOW how costly is it to call the code. People expect properties to be _cheap_ but it's hard to know since we don't have the source code for the implementation. E.g.

```csharp
long total = 0;
// with a method most people will realize there's a cost and call it once, out of the loop
byte[] large = foo.GetBlob ();
foreach (byte b in large)
    total += b;
```

```csharp
long total = 0;
// for properties it's more fuzzy since it's unclear if the result is cached or not...
for (int i =0 < i; foo.Blob.Length; i++)
    total += foo.Blob [i];
```

if you end up allocating a new `byte[]` for each iteration then you'll have a performance bottleneck on the property version.

Looking at what Swift does gives us a _hint_ if Apple thinks it's a good candidate to be a property (even if it's not Ibn ObjC) versus a method.

> Can you explain more about what Swift is doing?

`class func primaryDSTU2() -> Self`

TL&DR: It's not a property in Swift either.

Longer story: ObjC did not always had properties and even some _recent_ code/headers did not use them. Looking at Swift gives you a second, often better, opinion of the API.

PR #9160, opened 2020-07-22
[Devops] Use the deployment pool.

mandel-macaque:
Use the pool that is used just for utilities/short-lived tasks.


mandel-macaque:
@monojenkins backport d16-8

PR #9174, opened 2020-07-23
[d16-8][msbuild] Makes the xcarchive folder name culture independent

emaf:
Fixes https://developercommunity.visualstudio.com/content/problem/934833/xamarinforms-cant-create-archive-for-ios.html

Depending on the default language of the running OS this could cause translating AM/PM to symbols that are not supported on folder names on Windows.

For instance, the value of `timestamp` without this change for an user that sets Chinese as it's main language would be `8-18-15 1.31 下午`, where `下午` represent PM.


emaf:
@monojenkins backport main

PR #9179, opened 2020-07-24
[generator] Add nullability `?` when [BindAs] is used with an array type

spouliot:
Fix two false positives in AVFoundation.
Needed for xcode12 too (to be backported)


spouliot:

PR #9188, opened 2020-07-24
[tools][tests] Update to use macOS 11.0 (instead of 10.16)

spouliot:


spouliot:
failure is a random known issue -> https://github.com/xamarin/maccore/issues/2127

PR #9196, opened 2020-07-26
Remove some LINQ usages

Therzok:
Contributes to fixing #9019

### DIFF
--- a/src/AppKit/NSPredicateEditorRowTemplate.cs
+++ b/src/AppKit/NSPredicateEditorRowTemplate.cs
@@ -19,7 +19,7 @@ namespace AppKit
 	public partial class NSPredicateEditorRowTemplate
 	{
 		public NSPredicateEditorRowTemplate (params NSCompoundPredicateType [] compoundTypes)
-			: this (compoundTypes.Select (t => NSNumber.FromUInt32 ((uint)t)).ToArray ())
+			: this (Array.ConvertAll (compoundTypes, t => NSNumber.FromUInt32 ((uint)t)))
 		{
 		}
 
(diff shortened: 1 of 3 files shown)


src/AppKit/NSPredicateEditorRowTemplate.cs
```diff
@@ -8,8 +8,8 @@
 //
 
 using System;
-using System.Linq;
 using System.Collections.Generic;
+using System.Linq;
```
spouliot, line 12: _completely_ ?

dalexsoto: :P 
```suggestion
```

Therzok: Build breaks if I do that, so no. Sorry for the diff noise, but this should be sorted :D This is done in the 2nd commit where I convert _one_ usage to Array.ConvertAll, since there are other ToArray() calls in this file on IEnumerable instances.

spouliot: > Build breaks if I do that, so no.

Yup, I saw that ;-)

> Sorry for the diff noise, but this should be sorted :D 

IMO diff noise is worse than unsorted namespaces :)

The problem is that the PR description (and git commit message) do not match the content of the PR/commit. I like to trust descriptions while scrolling over commits to find regressions. Please update it.

p.s. don't resolve comments, let the commenter resolve them. Too easy to miss some stuff to double check, in particular in large PR (not the case here), when the comments are collapsed

Therzok: The commits do explicitly state what's changed, regardless, I'll move the 2nd commit to a different PR then

spouliot: It might be a translation issue but the change to the `src/AppKit/NSPredicateEditorRowTemplate.cs` file
* does not **remove** the System.Linq reference (even if it does _move_ it);
* is not related to **unused** System.Linq reference (it broke the build when you actually removed it);
* does modify, not only remove, existing code (which is not mentioned)

You can simply renamed the PR to `Reduce/remove System.Linq references` to make it match its content

spouliot:
@Therzok 🎂 ‼️

PR #9202, opened 2020-07-27
[msbuild] Fix automatic stripping of release builds by using the correct variable.

rolfbjarne:


dalexsoto:
Or 16-7?

rolfbjarne:
> That's only required in `main` or was that also in earlier (d16-6) branch ?

Only `main`, this regressed in b46aae7fd68bac7e3cae6d575060f3de6428a3eb, which was after `d16-8` was branched.

PR #9214, opened 2020-07-28
[mediaplayer] Update for Xcode 12 beta 3

spouliot:
tvOS mostly catch up with (much older) iOS changes


src/mediaplayer.cs
```diff
@@ -24,12 +24,13 @@
...
 	[Mac (10,12,2)] // type exists only to expose fields
 	[BaseType (typeof (NSObject))]
-#if IOS || WATCH
+#if !MONOMAC
 #if XAMCORE_4_0
 	[NoWatch] // marked as unavailable in xcode 12 beta 1
 #else
 	[Watch (5,0)]
 	[Obsoleted (PlatformName.WatchOS, 7,0, message: "Removed in Xcode 12")]
+	[TV (14,0)]
```
rolfbjarne, line 33: This should go outside of the #if XAMCORE_4_0 ... #endif, since that seems to be for watchOS only (otherwise the attribute won't apply in XAMCORE_4_0).

PR #8041, opened 2020-03-04
[PassKit] Update bindings to Xcode 11.4 Beta 3

dalexsoto:
Diff: https://github.com/xamarin/xamarin-macios/wiki/PassKit-iOS-xcode11.4-beta3

PR #9218, opened 2020-07-28
[passkit] Update for Xcode 12 beta 4

spouliot:


src/passkit.cs
```diff
@@ -1036,6 +1078,18 @@ interface PKPaymentAuthorizationControllerDelegate {
...
 		void DidSelectPaymentMethod (PKPaymentAuthorizationController controller, PKPaymentMethod paymentMethod, Action<PKPaymentRequestPaymentMethodUpdate> completion);
+
+		[Watch (7,0)][iOS (14,0)]
+		[Export ("paymentAuthorizationController:didRequestMerchantSessionUpdate:")]
+		void DidRequestMerchantSessionUpdate (PKPaymentAuthorizationController controller, Action<PKPaymentRequestMerchantSessionUpdate> handler);
+	
+		[Watch (7,0)][iOS (14,0)]
+		[Export ("presentationWindowForPaymentAuthorizationController:")]
+		[return: NullAllowed]
+#if MONOMAC
```
rolfbjarne, line 1089: I'm assuming this should eventually be abstract for all platforms, so:

```suggestion
#if MONOMAC || XAMCORE_4_0
```

src/passkit.cs
```diff
@@ -1316,4 +1372,379 @@ interface PKSecureElementPass {
 		[NullAllowed, Export ("pairedTerminalIdentifier")]
 		string PairedTerminalIdentifier { get; }
 	}
+
+	[NoWatch, NoTV]
+	[iOS (14,0)]
```
rolfbjarne, line 1377: No `[Mac (...)]` attribute?

src/passkit.cs
```diff
@@ -1316,4 +1372,379 @@ interface PKSecureElementPass {
...
+	[Native]
+	public enum PKAddShareablePassConfigurationPrimaryAction : ulong {
+		Add,
+		Share,
+	}
+
+	[Watch (7,0), iOS (14,0)]
+	[NoTV]
```
rolfbjarne, line 1385: No `[Mac (...)]` attribute?

src/passkit.cs
```diff
@@ -1316,4 +1372,379 @@ interface PKSecureElementPass {
...
+	[Protocol, Model (AutoGeneratedName = true)]
+	[BaseType (typeof (NSObject))]
+	interface PKAddSecureElementPassViewControllerDelegate {
+
+		[Deprecated (PlatformName.iOS, 14,0, message: "Use 'DidFinishAddingSecureElementPasses' instead.")]
+		[Abstract]
+		[Export ("addSecureElementPassViewController:didFinishAddingSecureElementPass:error:")]
+		void DidFinishAddingSecureElementPass (PKAddSecureElementPassViewController controller, [NullAllowed] PKSecureElementPass pass, [NullAllowed] NSError error);
+	
+#if XAMCORE_4_0
```
rolfbjarne, line 1439: This is a new protocol, so you don't need the `#if XAMCORE_4_0` here.

src/passkit.cs
```diff
@@ -1316,4 +1372,379 @@ interface PKSecureElementPass {
...
+		[Abstract]
+#endif
+		[Export ("addSecureElementPassViewController:didFinishAddingSecureElementPasses:error:")]
+		void DidFinishAddingSecureElementPasses (PKAddSecureElementPassViewController controller, [NullAllowed] PKSecureElementPass[] passes, [NullAllowed] NSError error);
+}
+
+	[NoWatch, NoTV, NoMac]
+	[iOS (13,4)]
+	[BaseType (typeof(UIViewController))]
```
rolfbjarne, line 1448:
```suggestion
	[BaseType (typeof (UIViewController))]
```

src/passkit.cs
spouliot: unsure, macOS-side needs to be reviewed (reason for being a draft)

src/passkit.cs
spouliot: no, it's @required on macOS and (still?) @optional on iOS / watchOS
probably because they don't want to introduce a breaking change (can't blame them to do the right thing, even if they don't generally care)

src/passkit.cs
spouliot: no, that's `PKAddSecureElementPassViewControllerDelegate` and was added in iOS 13.4

spouliot:
failure due to https://github.com/xamarin/maccore/issues/2154

src/passkit.cs
```diff
@@ -1316,4 +1434,386 @@ interface PKSecureElementPass {
...
+		[Export ("merchantName")]
+		string MerchantName { get; }
+
+		[Export ("transactionDate", ArgumentSemantic.Strong)]
+		NSDate TransactionDate { get; }
+
+		[Export ("currencyCode")]
+		string CurrencyCode { get; }
+
+		// FIXME ?
```
dalexsoto, line 1655: ? Should we create an issue? or list it in our bucket for review issue

spouliot: I forgot to remove the `FIXME` 
Other places (where `amount` is defined, in PassKit and elsewhere) uses `NSDecimalNumber` which is a subclass of `NSNumber` so we can't bind it as such.

src/passkit.cs
```diff
@@ -738,7 +800,7 @@ interface PKPass : NSSecureCoding, NSCopying {
 		[Export ("localizedValueForFieldKey:")]
 		NSObject GetLocalizedValue (NSString key); // TODO: Should be enum for PKPassLibraryUserInfoKey
 
-#if !XAMCORE_4_0
+#if !XAMCORE_4_0 || !MONOMAC
```
rolfbjarne, line 803: We're not defining XAMCORE_4_0, so adding `|| !MONOMAC` doesn't change anything. Do you mean `&& !MONOMAC`?

src/passkit.cs
rolfbjarne: iOS added it in 13.4, but as far as I can see we're adding the bindings for it in this PR, so we don't need to be compatible with existing bindings. Or am I missing something?

spouliot: > Or am I missing something?

Unlikely :) That was added in a different pass and the bindings have been done over a too-long period. I'll double check the versions since I don't recall missing stuff from 13.4.

spouliot: The availability was added in beta 1 (it did not have any before) so it was present in Xcode 11.5

It was added in Xcode 11.4 beta 3 (it's in the diff) but https://github.com/xamarin/xamarin-macios/pull/8041/files does not have it and xtro did not complain (looks like the specific header file was not (at that time) part of the umbrella header). c.c. @dalexsoto

spouliot:
Test failure is a known issue -> https://github.com/xamarin/maccore/issues/2154

PR #9224, opened 2020-07-29
[UIKit] Update bindings to Xcode 12 Beta 3

dalexsoto:


src/uikit.cs
```diff
@@ -5228,7 +5233,7 @@ interface UIColor : NSSecureCoding, NSCopying
 
 		// UIColor (UIAccessibility) Category
 
-		[Watch (7,0), TV (14,0), Mac (10,16), iOS (14,0)]
+		[Watch (7,0), TV (14,0), Mac (11,0), iOS (14,0)]
```
spouliot, line 5236: I doubt it's in macOS since it's related to `UIColor`
It might be that the category is defined on AppKit's `NSColor` on macOS but it would be bound again

dalexsoto: You are right 

```suggestion
		[Watch (7,0), TV (14,0), iOS (14,0)]
```

dalexsoto:
Breaking changes are expected.

PR #9248, opened 2020-07-30
[msbuild] Fix arguments to error logging for GetMinimumOSVersionTask.

rolfbjarne:
The filename arguments to the error logging must be strings, otherwise the
wrong logging overload is picked, and the resulting error message makes no
sense.


rolfbjarne:
Test failure is unrelated and will be fixed with #9251.

PR #9249, opened 2020-07-30
[HealthKit][Xcode12] Update beta 1 (no b2, b3 changes)

whitneyschmidt:


tests/monotouch-test/HealthKit/CategoryTypeIdentifierTest.cs
```diff
@@ -63,12 +61,13 @@ public void EnumValues_22351 ()
 				case HKCategoryTypeIdentifier.Fainting:
 				case HKCategoryTypeIdentifier.Fatigue:
 				case HKCategoryTypeIdentifier.Fever:
+				case HKCategoryTypeIdentifier.GeneralizedBodyAche:
```
whitneyschmidt, line 64: Removing #if !WATCHOS is correct.

Changing the order of preexisting identifiers is noise...fixing now.

PR #9261, opened 2020-08-03
[AppKit/UIKit] Merge the definitions of NS[Mutable]ParagraphStyle.

rolfbjarne:
* Reduces code duplication.
* Makes the macOS versions thread-safe (the iOS versions have been thread-safe
  for years: https://github.com/xamarin/maccore/commit/2c6a5303a71a897d216de4666ede24cf31a2b335).
* A few parameters names were different in the definitions; I chose to keep the ones in Xamarin.iOS, since they looked better.
* Xamarin.Mac had two methods, SetTextBlocks and SetTextLists, in place of an actual property override (for the mutable setter), this was fixed to be an actual property overload, and compat methods were implemented.
* xtro needed an update to cope with multiple static methods for the same selector.


rolfbjarne:
The breaking changes are only parameter name changes.

PR #9262, opened 2020-08-03
[mtouch] Don't try to copy invalid symbol files.

rolfbjarne:
This solves a rebuild problem if an assembly has an invalid or unsupported symbol
file, where we'd detect that the symbol file exists, and expect it to be copied,
but then the linker would drop it, causing us to always rebuild the app (this is
not the same as when a symbol file is out of date).

This happens for NUnitLite 3.12.0's nunit.framework.dll, which ships with an old-style
pdb.

Also add a warning that is shown when we detect that there's a symbol file, but it
couldn't be loaded for some reason.

### DIFF
--- a/tests/common/ExecutionHelper.cs
+++ b/tests/common/ExecutionHelper.cs
@@ -216,8 +216,13 @@ namespace Xamarin.Tests
 
 		public int WarningCount {
 			get {
-				return messages.Count ((v) => v.IsWarning);
+				return GetWarningCount (messages);
 			}
+		}
+
+		public static int GetWarningCount (List<ToolMessage> messages)
+		{
+			return messages.Count ((v) => v.IsWarning);
 		}
 
 		public bool HasError (string prefix, int number, string message)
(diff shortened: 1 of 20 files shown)


tests/common/ExecutionHelper.cs
```diff
@@ -216,10 +216,15 @@ public bool HasErrorPattern (string prefix, int number, string messagePattern)
 
 		public int WarningCount {
 			get {
-				return messages.Count ((v) => v.IsWarning);
+				return GetWarningCount (messages);
 			}
 		}
 
+		public static int GetWarningCount (List<ToolMessage> messages)
```
stephen-hawley, line 223: Minor: if there is no dependency on `List`, consider using `IEnumerable<ToolMessage>`.

tests/common/ExecutionHelper.cs
```diff
@@ -231,8 +236,13 @@ public bool HasError (string prefix, int number, string message)
...
 		public void AssertWarningCount (int count, string message = "warnings")
 		{
-			if (count != WarningCount)
-				Assert.Fail ($"{message}\nExpected: {count}\nBut was: {WarningCount}\nWarnings:\n\t{string.Join ("\n\t", this.Messages.Where ((v) => v.IsWarning).Select ((v) => v.ToString ()))}");
+			AssertWarningCount (messages, count, message);
+		}
+
+		public static void AssertWarningCount (List<ToolMessage> messages, int count, string message = "warnings")
+		{
+			if (count != GetWarningCount (messages))
```
stephen-hawley, line 244: Why not:
```
Assert.AreEqual (count, GetWarningCount (messages), \nWarnings:\n\t{string.Join ("\n\t", messages.Where ((v) => v.IsWarning).Select ((v) => v.ToString ()))}");
```

rolfbjarne: Because that will create the string to report to the user in all cases, even when the assert doesn't fail, and there can be a significant number of messages in some cases.

In other words: it's a performance improvement.

stephen-hawley: 👍

PR #9264, opened 2020-08-03
[tests] Move from MonoTouch.NUnitLite to Touch.Client.

rolfbjarne:
* Touch.Client is a tiny mobile UI/runner for the official NUnitLite. This
  means we'll stop using a very old fork of NUnitLite to run our tests, and
  instead use the most recent upstream version instead.

* Some test changes were required to use the official NUnitLite, because there
  have been breaking changes there.

* NUnitLite is now referenced using a NuGet, which means there are numerous
  changes to support this.

* We're not using GuiUnit anymore, so that dependency has been removed.


rolfbjarne:
Test failures are unrelated, they're all due to https://github.com/xamarin/maccore/issues/1265.

Api comparison failed because this PR this this the 0.1% of the cases where how the api comparison is done doesn't work (there's a source-code breaking change in Touch.Unit, which means the previous commit before this PR doesn't build with the updated Touch.Unit).

PR #9265, opened 2020-08-03
[GameController][Xcode12] Updates through beta 4

whitneyschmidt:


src/gamecontroller.cs
```diff
@@ -142,11 +166,11 @@ partial interface GCControllerDirectionPad {
 	[Deprecated (PlatformName.TvOS, 10, 0, message: "Use 'GCExtendedGamepad' instead.")]
 	[iOS (7,0)]
 	[Mac (10,9)]
-	[BaseType (typeof (NSObject))]
+	[BaseType (typeof (GCPhysicalInputProfile))]
 	[DisableDefaultCtor] // return nil handle -> only exposed as getter
 	partial interface GCGamepad {
 
-		[NullAllowed]
+		//[NullAllowed] This seems like it's a breaking change...
```
whitneyschmidt, line 173: Is this a breaking change? 🤔 It doesn't seem like it. Any code that assumes that a null may be returned will still work here - we're restricting the return value to a subset of what it used to be.

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+	[TV (14,0), Mac (11,0), iOS (14,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface GCDeviceHaptics
+	{
+		[Export ("supportedLocalities", ArgumentSemantic.Strong)]
+		NSSet<NSString> SupportedLocalities { get; }
+
+		// Dependency on Xcode 12 CoreHaptics bindings
```
whitneyschmidt, line 694: FYI @mandel-macaque.

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+		GCControllerButtonInput PaddleButton2 { get; }
+
+		[NullAllowed, Export ("paddleButton3")]
+		GCControllerButtonInput PaddleButton3 { get; }
+
+		[NullAllowed, Export ("paddleButton4")]
+		GCControllerButtonInput PaddleButton4 { get; }
+	}
+
+	[Static] // I tried moving GCKey and GCKeyCode to separate files, but ran into
```
whitneyschmidt, line 1015: Is there a way for `GCKey` and `GCKeyCode` to live in their own files?

src/gamecontroller.cs
rolfbjarne: I don't think changing this has any effect on our generated code (IOW it's fine).

src/gamecontroller.cs
```diff
@@ -422,6 +476,10 @@ partial interface GCMotion {
 		[Export ("rotationRate", ArgumentSemantic.Assign)]
 		Vector3d RotationRate { get; }
 
+
```
rolfbjarne, line 479: Very minor: extra empty line here.

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+		[NullAllowed, Export ("auxiliaryButtons")]
+		GCControllerButtonInput[] AuxiliaryButtons { get; }
+	}
+
+	interface IGCDevice {}
+
+	[TV (14,0), Mac (11,0), iOS (14,0)]
+	[BaseType (typeof (NSObject))]
+	[Protocol]
+	[Model]
```
rolfbjarne, line 848:
```suggestion
	[Model (AutoGeneratedName = true)]
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+		GCControllerButtonInput PaddleButton3 { get; }
+
+		[NullAllowed, Export ("paddleButton4")]
+		GCControllerButtonInput PaddleButton4 { get; }
+	}
+
+	[Static] // I tried moving GCKey and GCKeyCode to separate files, but ran into
+	// error CS0246: The type or namespace name 'StaticAttribute' could not be found (are you missing a using directive or an assembly reference?)
+	// error CS0246: The type or namespace name 'Static' could not be found (are you missing a using directive or an assembly reference?)
+    [TV (14, 0), Mac (11, 0), iOS (14, 0)]
```
rolfbjarne, line 1018: It seems it's space-indented from here, when it should be tab-indented.

src/gamecontroller.cs
rolfbjarne: It's possible, but it's somewhat of a special case.

Basically you'd have to add it like xkit.cs is added here:

https://github.com/xamarin/xamarin-macios/blob/53dcedc0609b6f0fa18cdbb1484bb948b94159d3/src/frameworks.sources#L1799

I don't think it's worth it to have binding code in a separate file, xkit.cs is literally the only exception, and that's only because we want the code in both UIKit and AppKit at the same time.

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+        NSString International7 { get; }
+
+        [Field ("GCKeyInternational8")]
+        NSString International8 { get; }
+
+        [Field ("GCKeyInternational9")]
+        NSString International9 { get; }
+
+        [Field ("GCKeyLANG1")]
+        NSString LANG1 { get; }
```
rolfbjarne, line 1349:
```suggestion
        NSString Lang1 { get; }
```

and the same for all the other LANG ones.

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+        nint International7 { get; }
+
+        [Field ("GCKeyCodeInternational8")]
+        nint International8 { get; }
+
+        [Field ("GCKeyCodeInternational9")]
+        nint International9 { get; }
+
+        [Field ("GCKeyCodeLANG1")]
+        nint LANG1 { get; }
```
rolfbjarne, line 1732: Same:

```suggestion
        nint Lang1 { get; }
```

src/gamecontroller.cs
mandel-macaque: AFAIK it is not an issue at all. It is a get and not a set, so we do not care. @whitneyschmidt on a 'set' we do check for the value and throw a NRE depending on the value, but since we only return, we are fine.

src/gamecontroller.cs
```diff
@@ -202,7 +226,7 @@ partial interface GCGamepadSnapshot {
 
 	[iOS (7,0)]
 	[Mac (10,9)]
-	[BaseType (typeof (NSObject))]
+	[BaseType (typeof (GCPhysicalInputProfile))]
```
mandel-macaque, line 229: All this base changes are ok, but do remember that changing a base COULD be a breaking change. Not in this case since GCPhysicalInputProfile is a NSObject and it was the parent of the old class.

src/gamecontroller.cs
```diff
@@ -447,6 +505,34 @@ partial interface GCMotion {
...
+		[Export ("hasGravityAndUserAcceleration")]
+		bool HasGravityAndUserAcceleration { get; }
+
+		[TV (14, 0), Mac (11, 0), iOS (14, 0)]
+		[Export ("acceleration")]
+		GCAcceleration Acceleration { get; }
+
+		[TV (14,0), Mac (11,0), iOS (14,0)]
+		[Export ("setAcceleration:")]
+		void SetAcceleration (GCAcceleration acceleration);
```
mandel-macaque, line 535: Method and a property? That is very weird? Acceleration and SetAcceleration do look like a perfect candidate for a property.

src/gamecontroller.cs
mandel-macaque: I know this guy. It is in my feedback report, thanks for the headsup :)

whitneyschmidt: Do we want to link to the issue here for easier record keeping? :)

src/gamecontroller.cs
dalexsoto: Nope not a breaking change

src/gamecontroller.cs
dalexsoto: It will be listed as a breaking change but there is nothing we can do about it, if a base changes we have to honor that

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+		NSString LocalityRightHandle { get; }
+
+		[Field ("GCHapticsLocalityTriggers")]
+		NSString LocalityTriggers { get; }
+
+		[Field ("GCHapticsLocalityLeftTrigger")]
+		NSString LocalityLeftTrigger { get; }
+
+		[Field ("GCHapticsLocalityRightTrigger")]
+		NSString LocalityRightTrigger { get; }
```
dalexsoto, line 722: I wonder if these should be grouped on their own static class or smart enum depending on how they are used for example

```csharp
[Static]
interface GCHapticsLocality {

	[Field ("GCHapticsLocalityDefault")]
  	NSString Default { get; }
 
  	[Field ("GCHapticsLocalityAll")]
  	NSString All { get; }
 
  	[Field ("GCHapticsLocalityHandles")]
  	NSString Handles { get; }
 
  	[Field ("GCHapticsLocalityLeftHandle")]
  	NSString LeftHandle { get; }
 
  	[Field ("GCHapticsLocalityRightHandle")]
  	NSString RightHandle { get; }
 
  	[Field ("GCHapticsLocalityTriggers")]
  	NSString Triggers { get; }
 
  	[Field ("GCHapticsLocalityLeftTrigger")]
  	NSString LeftTrigger { get; }
 
  	[Field ("GCHapticsLocalityRightTrigger")]
 	NSString RightTrigger { get; }
}

File: src/gamecontroller.cs [hunk 12]
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+		[Field ("GCHapticDurationInfinite")]
+		float DurationInfinite { get; }
```
dalexsoto, line 725: I am not sure if we should remove the Haptic part, else I am unsure what infinite duration is it about

```suggestion
		[Field ("GCHapticDurationInfinite")]
		float HapticDurationInfinite { get; }
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+		[Static]
+		[NullAllowed, Export ("coalescedKeyboard", ArgumentSemantic.Strong)]
+		GCKeyboard CoalescedKeyboard { get; }
+
+		[Notification, Field ("GCKeyboardDidConnectNotification")]
+		NSString DidConnectNotification { get; }
+
+		[Notification, Field ("GCKeyboardDidDisconnectNotification")]
+		NSString DisconnectNotification { get; }
```
dalexsoto, line 766:
```suggestion
		[Notification, Field ("GCKeyboardDidDisconnectNotification")]
		NSString DidDisconnectNotification { get; }
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+	[TV (14,0), Mac (11,0), iOS (14,0)]
+	[BaseType (typeof (GCPhysicalInputProfile))]
+	interface GCKeyboardInput
+	{
+		[NullAllowed, Export ("keyChangedHandler", ArgumentSemantic.Copy)]
+		GCKeyboardValueChangedHandler KeyChangedHandler { get; set; }
+
+		[Export ("anyKeyPressed")]
+		bool AnyKeyPressed { [Bind ("isAnyKeyPressed")] get; }
```
dalexsoto, line 779:
```suggestion
		[Export ("anyKeyPressed")]
		bool IsAnyKeyPressed { [Bind ("isAnyKeyPressed")] get; }
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+		[Export ("buttonForKeyCode:")]
+		[return: NullAllowed]
+		GCControllerButtonInput Button (nint code);
```
dalexsoto, line 783:
```suggestion
		GCControllerButtonInput GetButton (nint code);
```

src/gamecontroller.cs
dalexsoto: I would remove `[Model]` and `[BaseType]` here

dalexsoto, line 848:
```suggestion
	[Protocol]
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+		[TV (13, 0), Mac (10, 15), iOS (13, 0)]
+		[Abstract]
+		[Export ("productCategory")]
+		string ProductCategory { get; }
+
+		[TV (14, 0), Mac (11, 0), iOS (14, 0)]
+		[Abstract]
+		[Export ("physicalInputProfile", ArgumentSemantic.Strong)]
+		GCPhysicalInputProfile PhysicalInputProfile { get; }
```
dalexsoto, line 869: The availability attributes do not make sense here, specially since the container protocol is `[TV (14,0), Mac (11,0), iOS (14,0)]` so someone is lying to us

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+
+		[Export ("allAxes", ArgumentSemantic.Strong)]
+		NSSet<GCControllerAxisInput> AllAxes { get; }
+
+		[Export ("allDpads", ArgumentSemantic.Strong)]
+		NSSet<GCControllerDirectionPad> AllDpads { get; }
+
+		[Export ("objectForKeyedSubscript:")]
+		[return: NullAllowed]
+		GCControllerElement ObjectForKeyedSubscript (string key);
```
dalexsoto, line 909:
```suggestion
		GCControllerElement GetObjectForKeyedSubscript (string key);
```

src/gamecontroller.cs
```diff
@@ -535,4 +621,1162 @@ interface GCEventViewController {
...
+        NSString RightShift { get; }
+
+        [Field ("GCKeyRightAlt")]
+        NSString RightAlt { get; }
+
+        [Field ("GCKeyRightGUI")]
+        NSString RightGui { get; }
+    }
+
+	[TV (14, 0), Mac (11, 0), iOS (14, 0)]
```
dalexsoto, line 1400: indentation :)

src/gamecontroller.cs
whitneyschmidt: Looks like CoreHaptics is blocked on AVAudioSession for Mac bindings :(

https://github.com/xamarin/maccore/issues/2261

src/gamecontroller.cs
whitneyschmidt: done!

src/gamecontroller.cs
whitneyschmidt: +1 😄

src/gamecontroller.cs
whitneyschmidt: Aha! So we can have just `[Protocol]` if we remove `[Model]` and `[BaseType...]`! So `[Model]` and `[BaseType...]` always go together but `[Protocol]` can be independent. Thank you @dalexsoto :)

src/gamecontroller.cs
whitneyschmidt: This is bizarre because the availability attributes in the header match these ones...but it's impossible for that to be true. I'm thinking we should just leave the container protocol's availability attribute and remove the rest of them so that everything has `[TV (14,0), Mac (11,0), iOS (14,0)]`.

src/gamecontroller.cs
mandel-macaque: This case technically should not be an issue. Others might: https://docs.microsoft.com/en-us/dotnet/core/compatibility/#types

>  REQUIRES JUDGMENT: Introducing a new base class
>
>A type can be introduced into a hierarchy between two existing types if it doesn't introduce any new abstract members or change the semantics or behavior of existing types. For example, in .NET Framework 2.0, the DbConnection class became a new base class for SqlConnection, which had previously derived directly from Component.

src/gamecontroller.cs
```diff
@@ -535,4 +615,1157 @@ interface GCEventViewController {
...
+		GCControllerButtonInput PaddleButton3 { get; }
+
+		[NullAllowed, Export ("paddleButton4")]
+		GCControllerButtonInput PaddleButton4 { get; }
+	}
+
+	[Static]
+	[TV (14, 0), Mac (11, 0), iOS (14, 0)]
+	partial interface GCKey
+    {
```
rolfbjarne, line 1009: Indentation is space here.

src/gamecontroller.cs
```diff
@@ -4,16 +4,18 @@
...
-// Copyright 2019 Microsoft Corporation
+// Copyright 2019, 2020 Microsoft Corporation
 
 using System;
 
 using CoreFoundation;
 using Foundation;
 using ObjCRuntime;
 using OpenTK;
+//#using CoreHaptics; // need CoreHaptics update for TV 14.0 inclusion
```
mandel-macaque, line 18: PR with support for CoreHaptics on tvOS landed on the xcode12 branch: https://github.com/xamarin/xamarin-macios/pull/9273

src/gamecontroller.cs
whitneyschmidt: Good catch! Thank you 😄

tests/introspection/iOS/iOSApiProtocolTest.cs
```diff
@@ -399,6 +399,14 @@ protected override bool Skip (Type type, string protocolName)
 				case "SRVisit":
 				case "SRWebUsage":
 					return true;
+				// iOS 14 beta 3 (to be reviewed)
+				case "GCDeviceBattery":
+				case "GCDeviceLight":
+				case "GCDualShockGamepad":
+				case "GCKeyboard":
+				case "GCMouse":
+				case "GCXboxGamepad":
```
dalexsoto, line 408: Is this and the one below filled as an issue somewhere? So we do not forget to review these in the future or you could check for a specific Xcode Version in the tests so they fail if not fixed in the future, also if you consider this to be an Apple mistake file a bug :)

src/gamecontroller.cs
```diff
@@ -535,4 +616,1157 @@ interface GCEventViewController {
 		[Export ("controllerUserInteractionEnabled")]
 		bool ControllerUserInteractionEnabled { get; set; }
 	}
+
+	[TV (14,0), Mac (11,0), iOS (14,0)]
+	[BaseType (typeof( NSObject))]
```
rolfbjarne, line 621:
```suggestion
	[BaseType (typeof (NSObject))]
```

PR #2273, opened 2017-06-29
[IdentityLookup] Adds Xcode 9 Beta 1 bindings

dalexsoto:

PR #9276, opened 2020-08-05
[Tests] Fix TryGetSupportedInterfaces when we have values. Fixes #2273

mandel-macaque:
Make the test check against the value returned by the native API rather
than expect the result to always be null.

Fixes https://github.com/xamarin/maccore/issues/2273

### DIFF
--- a/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
+++ b/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
@@ -14,7 +14,6 @@ using System.IO;
 using Foundation;
 using ObjCRuntime;
 using SystemConfiguration;
-using System.Runtime.InteropServices;
 #if !MONOMAC
 using UIKit;
 #endif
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
```diff
@@ -82,16 +83,27 @@ public void TryCopyCurrentNetworkInfo ()
...
 		public void TryGetSupportedInterfaces ()
 		{
 			StatusCode status;
 #if __TVOS__
 			Assert.Throws<NotSupportedException> (() => CaptiveNetwork.TryGetSupportedInterfaces (out var ifaces));
 #else
+
 			status = CaptiveNetwork.TryGetSupportedInterfaces (out var ifaces);
 			Assert.AreEqual (StatusCode.OK, status, "Status");
-			Assert.IsNull (ifaces, "Null Interfaces");
```
rolfbjarne, line 94: I think you can just remove this line (and not add the other code), because the added test code just does the same thing the `CaptiveNetwork.TryGetSupportedInterfaces` method does, so there's not much testing value.

rolfbjarne:
Test failure is unrelated and will be fixed in PR #9288.

PR #9284, opened 2020-08-05
[GameKit][Xcode12] Update for beta 4

whitneyschmidt:


whitneyschmidt:
> No xtro file changes?

Nope, looks like xtro does not pick up nullability changes for params.

PR #9290, opened 2020-08-06
[monotouch-test] Improve a few asserts to make them show better output when they fail.

rolfbjarne:


rolfbjarne:
Test failures are unrelated and will be fixed with PR #9288.

PR #9292, opened 2020-08-06
[dotnet] Use the right variable to determine whether we're in debug mode or not.

rolfbjarne:


rolfbjarne:
Test failures are unrelated and will be fixed with PR #9300 and PR #9288.

PR #9295, opened 2020-08-06
[msbuild] Pass the full path to the output file to the native linker.

rolfbjarne:
This makes it easier to copy-paste commands to debug them.


rolfbjarne:

PR #9300, opened 2020-08-06
[introspection] Some P/Invokes have moved to a QCall library in .NET 5 preview 8+.

rolfbjarne:
This fixes numerous P/Invoke test failures like this:

    Could not find the symbol 'GlobalizationNative_GetCalendars' in QCall
    Could not find the symbol 'GlobalizationNative_GetCalendarInfo' in QCall
    Could not find the symbol 'GlobalizationNative_EnumCalendarInfo' in QCall
    Could not find the symbol 'GlobalizationNative_GetLatestJapaneseEra' in QCall
    [...]

Ref: https://github.com/dotnet/runtime/commit/a56d6a8034f5cb5ad58f9e29eb61edcc792f4b86
Ref: https://github.com/dotnet/runtime/commit/fae477f34b2185135f543af14d828fcd6789fee3

PR #9304, opened 2020-08-06
[msbuild] Compute variables for where the platform assembly is, and the platform assembly name.

rolfbjarne:
This will be more useful later, because these variables will be used in more places
in the .NET code.


rolfbjarne:

PR #9305, opened 2020-08-06
[msbuild] Create a few '_*DependsOn' variables that specifies depend targets.

rolfbjarne:
The .NET logic will inject an additional dependencies in these variables.


rolfbjarne:

PR #9307, opened 2020-08-06
[CoreMedia] Update bindings to Xcode 12 Beta 4

dalexsoto:


src/CoreMedia/CoreMedia.cs
```diff
@@ -244,6 +244,8 @@ public enum CMVideoCodecType : uint
 		Mpeg4Video       = 0x6d703476,
 		Mpeg2Video       = 0x6d703276,
 		Mpeg1Video       = 0x6d703176,
+		[iOS (14,0), TV (14,0), Watch (7,0), Mac (11,0)]
```
whitneyschmidt, line 247: Correct me if I'm getting this wrong for this case! Don't we usually add the newest enums at the end of the list?

dalexsoto: You can but since these enums have an explicit value you can put them wherever you want, I like to keep the same order as their native counterpart so it is easier to compare whenever you come back to them.

PR #9345, opened 2020-08-10
[Tests] Fix fragile test. The value can be not null.

mandel-macaque:
The no null test fails as shown in
https://github.com/xamarin/maccore/issues/2276

As with https://github.com/xamarin/xamarin-macios/pull/9276 there is not
much value in testing agains null.

Fixes https://github.com/xamarin/xamarin-macios/pull/9276

### DIFF
--- a/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
+++ b/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
@@ -74,11 +74,6 @@ namespace MonoTouchFixtures.SystemConfiguration {
 				return;
 
 			Assert.AreEqual (StatusCode.OK, status, "Status");
-			// To get a non-null dictionary back, we must (https://developer.apple.com/documentation/systemconfiguration/1614126-cncopycurrentnetworkinfo)
-			// * Use core location, and request (and get) authorization to use location information
-			// * Add the 'com.apple.developer.networking.wifi-info' entitlement
-			// I tried this, and still got null back, so just assert that we get null.
-			Assert.IsNull (dict, "Dictionary");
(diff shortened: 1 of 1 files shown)


tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
```diff
@@ -74,11 +74,6 @@ public void TryCopyCurrentNetworkInfo ()
 				return;
 
 			Assert.AreEqual (StatusCode.OK, status, "Status");
-			// To get a non-null dictionary back, we must (https://developer.apple.com/documentation/systemconfiguration/1614126-cncopycurrentnetworkinfo)
-			// * Use core location, and request (and get) authorization to use location information
-			// * Add the 'com.apple.developer.networking.wifi-info' entitlement
-			// I tried this, and still got null back, so just assert that we get null.
-			Assert.IsNull (dict, "Dictionary");
```
spouliot, line 81: Since we don't know why this started I'd rather we check both conditions
* dict is null
* dict contains `en0`

That way our (manual / not-generated code) is still being called (and tested).

An alternative would be to check for `en0` only - and see if/where this fails so we might understand what's going on.

rolfbjarne: The documentation for `CaptiveNetwork.TryCopyCurrentNetworkInfo` here might give some clues: https://developer.apple.com/documentation/systemconfiguration/1614126-cncopycurrentnetworkinfo

There's a lot of requirements to get network info, and if we don't satisfy them, we get null back. This also varies by iOS version, which may be another source of confusion.

PR #9346, opened 2020-08-10
[NearbyInteraction][Xcode12] Add manual binding for NINearbyObjectDirectionNotAvailable

whitneyschmidt:
Add manual binding for `NINearbyObject NINearbyObjectDirectionNotAvailable`.

This should be updated again when we have support in for https://github.com/xamarin/maccore/issues/2274


dalexsoto:
I think we need a test for this, maybe you can inspect the native values and make sure we match those on our side

src/frameworks.sources
```diff
@@ -1202,6 +1202,9 @@ NETWORK_SOURCES = \
 NEARBYINTERACTION_API_SOURCES = \
 	NearbyInteraction/Enums.cs \
 
+NEARBYINTERACTION_SOURCE = \
```
dalexsoto, line 1205:
```suggestion
NEARBYINTERACTION_SOURCES = \
```

src/frameworks.sources
```diff
@@ -1202,6 +1202,9 @@ NETWORK_SOURCES = \
...
+	NearbyInteaction/NINearbyObject.cs \
```
whitneyschmidt, line 1206:
```suggestion
	NearbyInteraction/NINearbyObject.cs \
```

whitneyschmidt:
> I think we need a test for this, maybe you can inspect the native values and make sure we match those on our side

I did the following:

Xcode:

```
vector_float3 vectVals = NINearbyObjectDirectionNotAvailable; 
//vectValues = (1.28042596E-19, 4.59163468E-41, 0, 0)
```

```
Vector3 vect = NINearbyObject.NINearbyObjectDirectionNotAvailable;
// vect = {0, 0, 0} (w/ 0 for length scalar)
```

Do these "match" if the ObjC values are so small?

rolfbjarne:
> > I think we need a test for this, maybe you can inspect the native values and make sure we match those on our side
> 
> I did the following:
> 
> Xcode:
> 
> ```
> vector_float3 vectVals = NINearbyObjectDirectionNotAvailable; 
> //vectValues = (1.28042596E-19, 4.59163468E-41, 0, 0)
> ```
> 
> ```
> Vector3 vect = NINearbyObject.NINearbyObjectDirectionNotAvailable;
> // vect = {0, 0, 0} (w/ 0 for length scalar)
> ```
> 
> Do these "match" if the ObjC values are so small?

Hm, no, I think we'd need to be exactly the same. Maybe printing out and comparing the memory contents?

```objective-c
vector_float3* v = & NINearbyObjectDirectionNotAvailable;
uint8_t* ptr = (uint8_t*) v;
for (int i = 0; i < sizeof (vector_float3); i++) {
    printf ("%.2x ", ptr [i]);
}
printf ("\n");
```

```csharp
Vector3 vect = NINearbyObject.NINearbyObjectDirectionNotAvailable;
unsafe {
    Vector3 *v = &vect;
    byte *ptr = (byte *) v;
    for (var i = 0; i < sizeof (Vector3); i++)
        Console.WriteLine ($"{ptr [i]:X2} ");
    Console.WriteLine ();
}
```

src/NearbyInteraction/NINearbyObject.cs
```diff
@@ -0,0 +1,41 @@
...
+#if !MONOMAC
+namespace NearbyInteraction {
+    partial class NINearbyObject
+    {
+        static Vector3? _NINearbyObjectDirectionNotAvailable;
+
+        // TODO: https://github.com/xamarin/maccore/issues/2274
+        // We do not have generator support to trampoline Vector3 -> vector_float3 for Fields
+        [Field ("NINearbyObjectDirectionNotAvailable",  "NearbyInteraction")]
+        public static Vector3 NINearbyObjectDirectionNotAvailable {
```
spouliot, line 27: remove `NINearbyObject` prefix since it's the current type name -> `ObjectDirectionNotAvailable`

src/NearbyInteraction/NINearbyObject.cs
```diff
@@ -0,0 +1,41 @@
...
+            get {
+                if (_NINearbyObjectDirectionNotAvailable == null) {
+                    unsafe {
+                        Vector3 *pointer = (Vector3 *) Dlfcn.GetIndirect (Libraries.NearbyInteraction.Handle, "NINearbyObjectDirectionNotAvailable");
+                        _NINearbyObjectDirectionNotAvailable = *pointer;
+                    }
+                }
+                return (OpenTK.Vector3)_NINearbyObjectDirectionNotAvailable;
```
spouliot, line 35: minor: that could be `(Vector3)`

whitneyschmidt:
> > > I think we need a test for this, maybe you can inspect the native values and make sure we match those on our side
> > 
> > 
> > I did the following:
> > Xcode:
> > ```
> > vector_float3 vectVals = NINearbyObjectDirectionNotAvailable; 
> > //vectValues = (1.28042596E-19, 4.59163468E-41, 0, 0)
> > ```
> > 
> > 
> > ```
> > Vector3 vect = NINearbyObject.NINearbyObjectDirectionNotAvailable;
> > // vect = {0, 0, 0} (w/ 0 for length scalar)
> > ```
> > 
> > 
> > Do these "match" if the ObjC values are so small?
> 
> Hm, no, I think we'd need to be exactly the same. Maybe printing out and comparing the memory contents?
> 
> ```objc
> vector_float3* v = & NINearbyObjectDirectionNotAvailable;
> uint8_t* ptr = (uint8_t*) v;
> for (int i = 0; i < sizeof (vector_float3); i++) {
>     printf ("%.2x ", ptr [i]);
> }
> printf ("\n");
> ```
> 
> ```cs
> Vector3 vect = NINearbyObject.NINearbyObjectDirectionNotAvailable;
> unsafe {
>     Vector3 *v = &vect;
>     byte *ptr = (byte *) v;
>     for (var i = 0; i < sizeof (Vector3); i++)
>         Console.WriteLine ($"{ptr [i]:X2} ");
>     Console.WriteLine ();
> }
> ```

Thanks @rolfbjarne - I was able to test these out and the byte-by-byte values in memory are all 00, from VS as well as Xcode 👍

src/NearbyInteraction/NINearbyObject.cs
```diff
@@ -0,0 +1,41 @@
+//
+// NearbyInteraction manual bindings
+//
+// Authors:
+//	Whitney Schmidt  <whschm@microsoft.com>
+//
+// Copyright 2020 Microsoft Inc.
```
spouliot, line 7: minor: it's not `Inc.` it's `Corp.` ;-)

tests/monotouch-test/NearbyInteraction/NINearbyObjectTest.cs
```diff
@@ -0,0 +1,37 @@
...
+				Assert.Ignore ("Requires iOS 14.0");
+
+			Vector3 vect = NINearbyObject.DirectionNotAvailable;
+
+			unsafe {
+				Vector3* v = &vect;
+				byte* ptr = (byte*) v;
+				byte zero = 0;
+				for (var i = 0; i < sizeof (Vector3); i++)
+					Assert.True (ptr[i] == zero, "Expected {0} but was {1}", zero.ToString(), ptr[i].ToString());
```
spouliot, line 31: minor:
* style, missing space before `(` (twice)
* `zero.ToString()` allocates and it's a constant in the loop
* the assert error will tell you a byte is not `0` but won't tell you where it is (`I`)

```csharp
Assert.That (ptr [i], Is.EqualTo (zero), $"Position {i}");
```

The above assert will provide the expected and actual values (and turn them to string only if needed/failing) so the only missing part (to diagnose the failure) is the position `i`.

tests/monotouch-test/NearbyInteraction/NINearbyObjectTest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class NINearbyObjectTest {
+
+		[Test]
+		public void DirectionNotAvailable ()
+		{
+			if (!TestRuntime.CheckXcodeVersion (12, 0))
+				Assert.Ignore ("Requires iOS 14.0");
```
spouliot, line 22: It's a special case but, since it's all `0` and C# struct are zeroized then the test should _work_ even before iOS 14.

whitneyschmidt: I tried running on iOS 13.2 w/o the version check and received the following error:

`2020-08-11 13:55:20.518482-0400 monotouchtest[43690:16037527] 	[FAIL] NINearbyObjectTest.DirectionNotAvailable : System.NullReferenceException : Object reference not set to an instance of an object`

Since NearbyInteraction is a new framework in Xcode12, I think this is expected.

I also tried out the following to double-check:

```
NSCoder coder = new NSCoder ();
NINearbyObject n = new NINearbyObject (coder);
```

It produced this error:
`[FAIL] NINearbyObjectTest.DirectionNotAvailable : System.Exception : Could not create an native instance of the type 'NearbyInteraction.NINearbyObject': the native class hasn't been loaded.`

I believe this is correct behavior since the framework isn't available < iOS 14, lmk if I'm missing something here!

spouliot:
```
NINearbyObjectTest.DirectionNotAvailable : System.NullReferenceException : Object reference not set to an instance of an object
```

Reading a field with `dlsym` does not require the field to exists - but maybe it did not like not finding the framework itself. The stack trace would show what cause the NRE.

Anyway not a huge deal, you can leave the version check there :)


```
NSCoder coder = new NSCoder ();
NINearbyObject n = new NINearbyObject (coder);
```

> I believe this is correct behavior since the framework isn't available < iOS 14

Yes. You can't do as it will call code `initWithCoder` on a class (`NINearbyObject`) that has not been loaded into memory.

However the field is different since it's static (no instance has to be created) and return a type (`Vector3`) that already exists. Nothing in your manual code requires the framework to be loaded. In fact it should _work_ (a managed exception is acceptable, not a crash) if someone was to put `NINearbyObject.DirectionNotAvailable` in a debugger watch on an older OS version.

whitneyschmidt: That makes sense - if there's no need to instantiate an object to access the field, then I see how it would work.

Someone tried it out on an older simulator and got this stack trace: https://gist.github.com/whitneyschmidt/e5d239d4f83473d29244c7cee86adcf3

tests/monotouch-test/NearbyInteraction/NINearbyObjectTest.cs
```diff
@@ -0,0 +1,37 @@
...
```
rolfbjarne, line 22: Really minor, so no need to fix this, but for the future this is shorter:

```suggestion
			TestRuntime.AssertXcodeVersion (12, 0);
```

tests/monotouch-test/NearbyInteraction/NINearbyObjectTest.cs
spouliot: Stack trace seems incomplete. The first frame is inside nunitlite code, nothing about the test code itself.

spouliot:
Tests failures are due to the new test being executed on tvOS and watchOS, both not supporting ARKit.
Wrap your source file inside `#if __IOS__` and `#endif` to exclude other platforms.

tests/monotouch-test/NearbyInteraction/NINearbyObjectTest.cs
whitneyschmidt: I have a breakpoint set on `Vector3 vect = NINearbyObject.DirectionNotAvailable;`. When I try and step into it, I end up with the stack trace above 🤔 Is there a better way to try and see what's going on?

PR #9354, opened 2020-08-12
[AVFoundation] Update framework for Xcode 12 beta 4.

mandel-macaque:


src/avfoundation.cs
```diff
@@ -2249,6 +2337,22 @@ interface AVAudioSession {
...
+		[Export ("setPreferredInputOrientation:error:")]
+		bool SetPreferredInputOrientation (AVAudioStereoOrientation orientation, [NullAllowed] out NSError outError);
+
+		[NoWatch, NoTV, NoMac, iOS (14, 0)]
+		[Export ("inputOrientation")]
+		AVAudioStereoOrientation InputOrientation { get; }
+
+		[NoWatch, NoTV, NoMac, iOS (14, 0)]
+		[Field ("AVAudioSessionPolarPatternStereo")]
+		NSString PolarPatternStereo { get; }
```
spouliot, line 2355: There are other (older) members defined for `AVAudioSessionPolarPattern` and this one should go with them

https://developer.apple.com/documentation/avfoundation/avaudiosessionpolarpattern

src/avfoundation.cs
```diff
@@ -13601,4 +13832,68 @@ interface AVMetadataSalientObject : NSCopying {
...
+		bool IsSyncSample { get; }
+	}
+
+	[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface AVAssetSegmentTrackReport {
+		[Export ("trackID")]
+		int TrackId { get; }
+
```
mandel-macaque, line 13869: BindAs AVMediaTypes

src/avfoundation.cs
```diff
@@ -3945,6 +4067,42 @@ interface AVAssetWriter {
...
+		CMTime PreferredOutputSegmentInterval { get; set; }
+
+		[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+		[Export ("initialSegmentStartTime", ArgumentSemantic.Assign)]
+		CMTime InitialSegmentStartTime { get; set; }
+
+		[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+		[Export ("outputFileTypeProfile")]
+		[NullAllowed]
+		string OutputFileTypeProfile { get; set; }
```
spouliot, line 4084: In ObjC that's `AVFileTypeProfile` (a `NSString` subclass) which you have defined before
Better bind this as a smart enum (or an `NSString`) but not a normal `string`

src/avfoundation.cs
```diff
@@ -7908,6 +8103,10 @@ interface AVVideoCompositing {
 		[TV (13,0), NoWatch, Mac (10,15), iOS (13,0)]
 		[Export ("prerollForRenderingUsingHint:")]
 		void PrerollForRendering (AVVideoCompositionRenderHint renderHint);
+
+		[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+		[Export ("supportsHDRSourceFrames")]
+		bool SupportsHDRSourceFrames { get; }
```
spouliot, line 8109: -> `Hdr`

src/avfoundation.cs
```diff
@@ -13601,4 +13832,68 @@ interface AVMetadataSalientObject : NSCopying {
...
+
+	[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface AVAssetSegmentTrackReport {
+		[Export ("trackID")]
+		int TrackId { get; }
+
+		[Export ("mediaType")]
+		string MediaType { get; }
```
mandel-macaque, line 13871:
```suggestion
		[BindAs (typeof (AVMediaTypes))]
		[Export ("mediaType")]
		NSString MediaType { get; }
```

src/avfoundation.cs
```diff
@@ -13140,7 +13375,7 @@ interface AVContentKeySession {
 	}
 
 	[Static][Internal]
-	[NoWatch, NoTV, Mac (10,15), iOS (12,2)]
+	[Watch (2,0), NoTV, Mac (10,15), iOS (12,2)]
```
spouliot, line 13378: hmm... I don't think AVFoundation was available in 2.0

https://developer.apple.com/documentation/avfoundation?language=objc suggest it was in 3.0
In about better use the current SDK as it was likely not very useful before now

src/avfoundation.cs
mandel-macaque: You mean move it to the same location in the file?

spouliot: it should be along other `AVAudioSessionPolarPattern*` fields which, sadly, are old enough that a smart enum was not used for them :(

src/avfoundation.cs
```diff
@@ -13236,15 +13467,15 @@ interface AVContentKeyRequest {
 	}
 
 	[Category]
-	[Mac (10, 12, 4), iOS (10,3), TV (10, 2), NoWatch]
+	[Mac (10, 12, 4), iOS (10,3), TV (10, 2), Watch (7,0)]
```
whitneyschmidt, line 13470:
```suggestion
	[Mac (10, 12, 4), iOS (10,3), TV (10, 2), Watch (7, 0)]
```

src/avfoundation.cs
```diff
@@ -13236,15 +13467,15 @@ interface AVContentKeyRequest {
...
 	[BaseType (typeof(AVContentKeyRequest))]
 	interface AVContentKeyRequest_AVContentKeyRequestRenewal
 	{
 		[Export ("renewsExpiringResponseData")]
 		bool GetRenewsExpiringResponseData ();
 	}
 
-	[TV (10,2), Mac (10,12,4), iOS (10,3), NoWatch]
+	[TV (10,2), Mac (10,12,4), iOS (10,3), Watch (7, 0)]
```
whitneyschmidt, line 13478:
```suggestion
	[TV (10,2), Mac (10,12,4), iOS (10,3), Watch (7,0)]
```

src/avfoundation.cs
```diff
@@ -13254,7 +13485,7 @@ interface AVPersistableContentKeyRequest {
 
 	}
 
-	[TV (10,2), Mac (10,12,4), iOS (10,3), NoWatch]
+	[TV (10,2), Mac (10,12,4), iOS (10,3), Watch (7, 0)]
```
whitneyschmidt, line 13488:
```suggestion
	[TV (10,2), Mac (10,12,4), iOS (10,3), Watch (7,0)]
```

src/avfoundation.cs
```diff
@@ -13601,4 +13832,69 @@ interface AVMetadataSalientObject : NSCopying {
...
+
+		[Export ("duration")]
+		CMTime Duration { get; }
+
+		[NullAllowed, Export ("firstVideoSampleInformation")]
+		AVAssetSegmentReportSampleInformation FirstVideoSampleInformation { get; }
+	}
+
+	[NoWatch, NoTV, NoiOS, Mac (11,0)]
+	[BaseType (typeof(NSObject))]
```
whitneyschmidt, line 13885:
```suggestion
	[BaseType (typeof (NSObject))]
```

src/avfoundation.cs
```diff
@@ -13172,52 +13407,48 @@ interface AVContentKeySession_AVContentKeyRecipients
...
 
-		[NoWatch]
+		[Watch (7, 0)]
 		[NullAllowed, Export ("error")]
 		NSError Error { get; }
 
-		[NoWatch]
+		[Watch (7, 0)]
 		[NullAllowed, Export ("identifier")]
 		NSObject Identifier { get; }
```
whitneyschmidt, line 13428: Is it possible to do any better than an `NSObject` for `Identifier`?

Header:

```
/*
 @property      identifier
 @abstract      Container- and protocol-specific identifier for the content key.
 @discussion    In order to use a key with an HTTP Live Streaming AVURLAsset, the identifier must be an NSURL that matches a key URI in the Media Playlist.
*/
@property (readonly, nullable) id identifier;
```

src/avfoundation.cs
```diff
@@ -1827,204 +1832,241 @@ interface AVAudioSessionSecondaryAudioHintEventArgs {
 
 	delegate void AVPermissionGranted (bool granted);
 
-	[NoMac]
+	[Mac (11, 0)]
```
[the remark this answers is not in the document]
mandel-macaque: I'll make the changes,, but FYI I do this on purpose, since objc-sharpie is not consistent and I prefer to follow the mono style.

src/avfoundation.cs
mandel-macaque: Mope, identifiers are just an NSObject, we do not care about it. It is used to later retrieve it. This is 'common' in other APIs.

src/avfoundation.cs
mandel-macaque: We wont forget now: https://github.com/xamarin/xamarin-macios/issues/9356 will done once xcode12 is done.

src/avfoundation.cs
```diff
@@ -4081,7 +4082,8 @@ interface AVAssetWriter {
 		[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
 		[Export ("outputFileTypeProfile")]
 		[NullAllowed]
-		string OutputFileTypeProfile { get; set; }
+		[BindAs (typeof (AVFileTypeProfile))]
+		NSString OutputFileTypeProfile { get; set; }
```
spouliot, line 4086: missed it earlier but there's a `[NullAllowed]` here
so the smart enum should have a default member, like
```
[Field (null)]
None = 0,
```
Ideally at first position (0) when possible. Otherwise you won't be able to set the property to `null` anymore

src/avfoundation.cs
```diff
@@ -1827,204 +1832,241 @@ interface AVAudioSessionSecondaryAudioHintEventArgs {
 
 	delegate void AVPermissionGranted (bool granted);
 
-	[NoMac]
+	[Mac (11,0)]
 	[Watch (3,0)]
+#if MONOMAC
+	[Static]
```
rolfbjarne, line 1838: What does this attribute do? IIRC it only make sense for members, not types.

src/avfoundation.cs
```diff
@@ -3767,6 +3872,7 @@ interface AVAssetResourceLoadingRequest {
 		bool Finished { get; }
 
 		[Export ("finishLoadingWithResponse:data:redirect:")]
+		[Deprecated (PlatformName.MacOSX, 10, 15, message : "Use the 'Response', 'Redirect' properties and the 'AVAssetResourceLoadingDataRequest.Responds' and 'AVAssetResourceLoadingRequest.FinishLoading' methods instead.")]
```
rolfbjarne, line 3875: Is it only deprecated on macOS?

src/avfoundation.cs
```diff
@@ -3945,6 +4068,43 @@ interface AVAssetWriter {
...
+		[NullAllowed]
+		[BindAs (typeof (AVFileTypeProfile))]
+		NSString OutputFileTypeProfile { get; set; }
+
+		[Wrap ("WriterWeakDelegate")]
+		IAVAssetWriterDelegate WriterDelegate { get; set; }
+
+		[TV (14, 0), NoWatch, Mac (11, 0), iOS (14, 0)]
+		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
+		NSObject WriterWeakDelegate { get; set; }
```
rolfbjarne, line 4093: Why `Writer*Delegate` instead of `just `*Delegate`? There doesn't seem to be a Delegate property already.

src/avfoundation.cs
```diff
@@ -3844,6 +3950,18 @@ interface AVAssetResourceLoadingContentInformationRequest {
 		string[] AllowedContentTypes { get; }
 	}
 
+	interface IAVAssetWriterDelegate { }
+
+	[TV (14,0), NoWatch, Mac (11,0), iOS (14,0)]
+	[Protocol]
```
rolfbjarne, line 3956: This should be a Model too:

```suggestion
	[Protocol]
	[Model (AutoGeneratedName = true)]
	[BaseType (typeof (NSObject))]
```

src/avfoundation.cs
spouliot: it works on types too, it makes a static type - which is how we regrouped constants (before smart enums)

src/avfoundation.cs
mandel-macaque: Missing the tvOS one, ths.

src/avfoundation.cs
mandel-macaque: This comes from a category. I fear that an other category might be added with a Delegate, which could mean that we will have the Deprecated XAMCORE_4 dance. Since it is a category, it might happen. Unless you see this not being possible, I prefer to add the Writer prefix.

dalexsoto: Not that I recall, I think we can leave just the `Delegate` name here, it makes sense `AVAssetWriter` is the class name and `AVAssetWriterDelegate` is its delegate.

My vote is to just use `Delegate` and `WeakDelegate` as managed names.

PR #9357, opened 2020-08-12
[iOSurface] Remove warning from the generator.

mandel-macaque:
The following warning was being raised:

```
iosurface.cs(202,4): warning CS0657: 'return' is not a valid attribute location for this declaration. Valid attribute locations for this declaration are 'property'. All attributes in this block will be ignored.
```

Moved the attr to the correct location and fixed a missing using with
the MaybeNull attrs.

### DIFF
--- a/src/iosurface.cs
+++ b/src/iosurface.cs
@@ -8,6 +8,7 @@
 //
 
 using System;
+using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


mandel-macaque:
@spouliot I was puzzled too. Maybe due to the warning? I wonder if the compiler does more than one pass.

PR #9363, opened 2020-08-13
[mtouch/mmp] Move Application.BuildTarget, Application.IsDeviceBuild and Application.IsSimulatorBuild to shared code.

rolfbjarne:
Also add a 'None' build target for the BuildTarget enum for when we're
building for neither simulator nor device (i.e. macOS). This means the default
value will change (since 'Simulator' is no longer the first value), but as far
as I can tell we're always assigning a specific value and not relying on the
default, so this should not make any difference.

This will be needed when the .NET code starts using these classes.

### DIFF
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -21,6 +21,12 @@ using PlatformResolver = Xamarin.Bundler.MonoMacResolver;
 
 namespace Xamarin.Bundler {
 
+	public enum BuildTarget {
+		None,
+		Simulator,
+		Device,
+	}
+
 	public enum MonoNativeMode {
 		None,
 		Compat,
@@ -88,6 +94,8 @@ namespace Xamarin.Bundler {
 		public List<Application> SharedCodeApps = new List<Application> (); // List of appexes we're sharing code with.
 		public string RegistrarOutputLibrary;
 
+		public BuildTarget BuildTarget;
+
 		bool RequiresXcodeHeaders {
 			get {
 				switch (Platform) {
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Test failures are unrelated

* monotouch-test/iOS 32-bits: https://github.com/xamarin/maccore/issues/1605

tools/common/Application.cs
```diff
@@ -132,6 +140,37 @@ public partial class Application
...
+				case ApplePlatform.TVOS:
+				case ApplePlatform.WatchOS:
+					return BuildTarget == BuildTarget.Device;
+				case ApplePlatform.MacOSX:
+					return false;
+				default:
+					throw ErrorHelper.CreateError (71, Errors.MX0071, Platform, ProductName);
+				}
+			}
+		}
```
spouliot, line 159: Merging mtouch/mmp code is nice when it reduce the total amount of code.

However if every member has to switch on the platform then it's not a gain, in size or in readability. Adding a new platform also means changing code all over the place.

Why not just use some polymorphism ? `MacApplication : Application` and `iOSApplication : Application`.

It was almost the case with the partial classes and custom files, except they required different binaries. OTOH subclasses would not prevent both/all classes to be in the same binary (if that's your end goal).

tools/common/Application.cs
```diff
@@ -21,6 +21,12 @@
 
 namespace Xamarin.Bundler {
 
+	public enum BuildTarget {
+		None,
```
spouliot, line 26: If it's used for the Mac/host then let's give it that name, because `None` does not reflect the target.

PR #9364, opened 2020-08-13
[mtouch/mmp] Move a few Application.Is* properties to shared code.

rolfbjarne:
This will be needed when the .NET code starts using these classes.

### DIFF
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -9,6 +9,7 @@ using Mono.Cecil.Cil;
 using Mono.Linker;
 
 using Xamarin.Linker;
+using Xamarin.MacDev;
 using Xamarin.Utils;
 
 using ObjCRuntime;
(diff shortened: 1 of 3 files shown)


tools/common/Application.cs
```diff
@@ -209,6 +211,38 @@ public void ParseI18nAssemblies (string i18n)
...
+			get {
+				if (!IsExtension)
+					return null;
+
+				var info_plist = Path.Combine (AppDirectory, "Info.plist");
+				var plist = Driver.FromPList (info_plist);
+				var dict = plist.Get<PDictionary> ("NSExtension");
+				if (dict == null)
+					return null;
+				return dict.GetString ("NSExtensionPointIdentifier");
```
spouliot, line 242: how many times is this called ? directly or indirectly
since it's read only maybe it should be cached ?

rolfbjarne: Good idea: https://github.com/xamarin/xamarin-macios/issues/9368

PR #9367, opened 2020-08-13
[xharness] Do not mark DeviceLoaders as loaded when loading fails

premun:
1. When listing of devices would fail for some reason (timeout which was set to 30 seconds), we would mark the state as `loaded` but no devices are stored in the memory.
2. Since no devices would fit our requirements when asking for one, we would create a new device. There would probably be a duplicate one that we have created previously already but we wouldn't know.

This starts a viscous circle of:
- Loading devices now takes a little bit longer with more devices to enumerate -> increasing the chance that it would time out again.
- With increasing probability of failure (timeout), we create a new device for each failure, increasing the probability even more until some machines get cluttered with duplicate devices.
- With about 15-20 devices the `xcrun simctl list devices` command takes about 3-4 minutes to complete and mlaunch times out at every attempt from there.

### DIFF
--- a/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/HardwareDeviceLoader.cs
+++ b/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/HardwareDeviceLoader.cs
@@ -52,8 +52,6 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared.Hardware {
 				connectedDevices.Reset ();
 			}
 
-			loaded = true;
-
 			var tmpfile = Path.GetTempFileName ();
 			try {
 				using (var process = new Process ()) {
@@ -92,6 +90,8 @@ namespace Microsoft.DotNet.XHarness.iOS.Shared.Hardware {
 						}
 						connectedDevices.Add (d);
 					}
+
+					loaded = true;
 				}
 			} finally {
 				connectedDevices.SetCompleted ();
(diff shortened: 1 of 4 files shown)


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Hardware/HardwareDeviceLoader.cs
```diff
@@ -92,6 +90,8 @@ public async Task LoadDevices (ILog log, bool includeLocked = false, bool forceR
 						}
 						connectedDevices.Add (d);
 					}
+
+					loaded = true;
```
rolfbjarne, line 94: There's a different problem now, if LoadDevices is called multiple times, we'll load devices multiple times:

```
var task1 = loader.LoadDevices (...);
var task2 = loader.LoadDevices (...);
// there are now 2 mlaunch processes
```

premun: @rolfbjarne how about the semaphore?

premun: Ah, sorry, I was looking at SimulatorLoader

premun: I added the Semaphore to HW device loader as well

rolfbjarne:
There's a compilation failure:

> Hardware/HardwareDeviceLoader.cs(30,12): error CS0246: The type or namespace name 'SemaphoreSlim' could not be found (are you missing a using directive or an assembly reference?)

PR #9369, opened 2020-08-13
[xharness] Add support for getting the default AssemblyName for projects.

rolfbjarne:
In .NET projects there's a default value for most properties, which means that
there won't necessarily be an AssemblyName property in a csproj. We need to know the
AssemblyName, so calculate it from the csproj filename (which is how .NET does it).

This turned out slighly complicated, because we're pass an XmlDocument around,
and the XmlDocument doesn't know the file from where it was loaded, so we need
to keep that information separately.

### DIFF
--- a/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Utilities/PlistExtensions.cs
+++ b/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Utilities/PlistExtensions.cs
@@ -1,8 +1,17 @@
 ﻿using System.IO;
+using System.Runtime.CompilerServices;
 using System.Xml;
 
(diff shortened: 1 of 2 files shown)


tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared/Utilities/PlistExtensions.cs
```diff
@@ -1,8 +1,17 @@
 ﻿using System.IO;
+using System.Runtime.CompilerServices;
```
mandel-macaque, line 2: needed?

rolfbjarne: Yes, that's where ConditionalWeakTable is.

PR #9371, opened 2020-08-13
Partial fix for AVAudioEngine.Connect(input, sink, format) crash when format == null

whitneyschmidt:
This fixes the crash which occurs when calling `AVAudioEngine.Connect(input, sink, null)` with null as the third parameter.

The crash occurred when we tried to check for equality using `.Equals` on `null` for the null check in this generated code:

`global::ObjCRuntime.Messaging.void_objc_msgSend_IntPtr_IntPtr_nuint_IntPtr (this.Handle, Selector.GetHandle ("connect:toConnectionPoints:fromBus:format:"), sourceNode.Handle, nsa_destNodes.Handle, sourceBus, format == null ? IntPtr.Zero : format.Handle);`

### DIFF
--- a/src/AVFoundation/AVAudioFormat.cs
+++ b/src/AVFoundation/AVAudioFormat.cs
@@ -21,11 +21,19 @@ namespace AVFoundation {
 	public partial class AVAudioFormat {
 		public static bool operator == (AVAudioFormat a, AVAudioFormat b)
 		{
+			if ((object) a == (object) b)
+				return true;
+			if ((object) a == null ^ (object) b == null)
+				return false;
 			return a.Equals (b);
 		}
 		
 		public static bool operator != (AVAudioFormat a, AVAudioFormat b)
 		{
+			if ((object) a != (object) b)
+				return true;
+			if ((object) a == null ^ (object) b == null)
+				return true;
 			return !a.Equals (b);
 		}
 	}
--- a/src/AVFoundation/AVAudioFormat.cs
+++ b/src/AVFoundation/AVAudioFormat.cs
@@ -30,11 +30,7 @@ namespace AVFoundation {
 		
 		public static bool operator != (AVAudioFormat a, AVAudioFormat b)
 		{
-			if ((object) a != (object) b)
-				return true;
-			if ((object) a == null ^ (object) b == null)
-				return true;
-			return !a.Equals (b);
+			return !(a == b);
 		}
 	}
 }
--- a/tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
+++ b/tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
@@ -1,0 +1,49 @@
+﻿// Unit test for AVAudioFormat
+// Authors: 
+// 		Whitney Schmidt (whschm@microsoft.com)
+// Copyright 2020 Microsoft Corp.
+
+using System;
+using Foundation;
+using AVFoundation;
+using NUnit.Framework;
+namespace MonoTouchFixtures.AVFoundation {
+
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class AVAudioFormatTest {
+
+		AVAudioFormat nullFormat = null;
+
+		[Test]
+		public void TestEqualOperatorSameInstace ()
+		{
+			using (var format = new AVAudioFormat ())
+				Assert.IsTrue (format == format, "format == format");
+		}
+
+		[Test]
+		public void TestEqualOperatorNull ()
+		{
+			using (var format = new AVAudioFormat ())
+			{
+				Assert.IsFalse (format == null, "format == null");
+				Assert.IsFalse (null == format, "null == format");
+			}
+			Assert.IsTrue (nullFormat == null, "nullFormat == null");
+			Assert.IsTrue (null == nullFormat, "null == nullFormat");
+		}
+
+		[Test]
+		public void TestNotEqualOperatorNull ()
+		{
+			using (var format = new AVAudioFormat ())
+			{
+				Assert.IsTrue (format != null, "format != null");
+				Assert.IsTrue (null != format, "null != format");
+			}
+			Assert.IsFalse (nullFormat != null, "nullFormat != null");
+			Assert.IsFalse (null != nullFormat, "null != nullFormat");
+		}
+	}
+}
--- a/tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
+++ b/tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
@@ -13,8 +13,6 @@ namespace MonoTouchFixtures.AVFoundation {
 	[Preserve (AllMembers = true)]
 	public class AVAudioFormatTest {
 
-		AVAudioFormat nullFormat = null;
-
 		[Test]
 		public void TestEqualOperatorSameInstace ()
 		{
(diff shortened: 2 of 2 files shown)


src/AVFoundation/AVAudioFormat.cs
```diff
@@ -21,11 +21,19 @@ namespace AVFoundation {
...
 			return a.Equals (b);
 		}
 		
 		public static bool operator != (AVAudioFormat a, AVAudioFormat b)
 		{
+			if ((object) a != (object) b)
+				return true;
+			if ((object) a == null ^ (object) b == null)
+				return true;
 			return !a.Equals (b);
```
spouliot, line 37: might be smaller to `return !(a == b);`

whitneyschmidt: 🤦‍♀️ Good catch! That's much better 😄

tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
```diff
@@ -0,0 +1,49 @@
...
+	[TestFixture]
+	[Preserve (AllMembers = true)]
+	public class AVAudioFormatTest {
+
+		AVAudioFormat nullFormat = null;
+
+		[Test]
+		public void TestEqualOperatorSameInstace ()
+		{
+			using (var format = new AVAudioFormat ())
```
whitneyschmidt, line 21: Not sure whether these `using` are needed - could `format` be a field like `nullformat`?

spouliot: It's not very important (in tests) but it's good to dispose asap
if there's a disposal issue you'll know it sooner, likely with a reference to the test (stack trace)
otherwise it can happen at GC time and tracking those are much less fun :)

tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
```diff
@@ -0,0 +1,49 @@
...
+		[Test]
+		public void TestEqualOperatorNull ()
+		{
+			using (var format = new AVAudioFormat ())
+			{
+				Assert.IsFalse (format == null, "format == null");
+				Assert.IsFalse (null == format, "null == format");
+			}
+			Assert.IsTrue (nullFormat == null, "nullFormat == null");
+			Assert.IsTrue (null == nullFormat, "null == nullFormat");
```
tj-devel709, line 34: Is it a standard practice to test using a field from the class and a variable inside the method or was this specific to this case and why did you decide to do that?

spouliot: no, but if there's one place where _standard_ is not very useful is the tests ;-) Having diversity, like different code (not source) patterns, can some time help us find interesting bugs

whitneyschmidt: This test code was mostly copied/modified from here: https://github.com/whitneyschmidt/xamarin-macios/blob/d3f0916cdb5a43ad1a18d7271cbe19603e004a07/tests/monotouch-test/Foundation/UrlTest.cs

The `using` ensures that the variable inside the method is disposed as soon as possible.

From my understanding (@spouliot + @mandel-macaque please correct as needed 😆 ) if we didn't wrap the variable in `using`, it would be picked up for disposal by the garbage collector, which we have no control over.

I think we don't need to worry about `nullFormat` because there's no object created in that variable's assignment and the value is never modified so there's no potential for weirdness with memory management.

The choice to make format a variable inside of individual test methods comes from the other test, and fwict we try to stick to the policy of making sure things are disposed asap in testing code (even if it is not strictly necessary).

Maybe someone else can give an example of a bug that might occur w/o `using` to handle disposal?

mandel-macaque: @tj-devel709 @whitneyschmidt Usually, it is good to ensure that the variables in the test are disposed ASAP as in this example. There are some cases, in which the variables that we are going to be using do need certain setup before they are executed and that the setup is expensive or error prone. One example I can think of are the networking tests, in which we need to create a connection and get the available interfaces. That information can be shared for the full TestCase and is error prone to do it per test.

In https://github.com/whitneyschmidt/xamarin-macios/blob/d3f0916cdb5a43ad1a18d7271cbe19603e004a07/tests/monotouch-test/Network/NWConnectionTest.cs we do such a thing, where as you can see SetUp does a lot:

```csharp
		[SetUp]
		public void SetUp ()
		{
			// connect and once the connection is done, deal with the diff tests
			connectedEvent = new AutoResetEvent(false);
			host = NetworkResources.MicrosoftUri.Host;
			// we create a connection which we are going to use to get the availabe
			// interfaces, that way we can later test protperties of the NWParameters class.
			using (var parameters = NWParameters.CreateUdp ())
			using (var endpoint = NWEndpoint.Create (host, "80"))
			{
				using (var protocolStack = parameters.ProtocolStack) {
					var ipOptions = protocolStack.InternetProtocol;
					ipOptions.IPSetVersion (NWIPVersion.Version4);
				}
				connection = new NWConnection (endpoint, parameters);
				connection.SetQueue (DispatchQueue.DefaultGlobalQueue); // important, else we will get blocked
				connection.SetStateChangeHandler (ConnectionStateHandler);
				connection.Start (); 
				Assert.True (connectedEvent.WaitOne (20000), "Connection timed out.");
			}

		}
```

Because we use a TestCase instance variable, the life of the variable cannot be managed with a using clause, so we use the TearDown method, to call Dsipose from the IDisposable interface:

```csharp
		[TearDown]
		public void TearDown () => connection.Dispose ();
```

Ideally, when unit testing we want to start with a clean canvas, like what @whitneyschmidt does here. 

Question: @whitneyschmidt @tj-devel709 what will happen with this:

```csharp
using (var format = new AVAudioFormat ())
using (IDisposable myMemoryHungryVar = null) {
   // I test, I don't care about my hunger! 🍔 
}
```

mandel-macaque: Another nice thing to read, there is a diff between [SetUp](https://nunit.org/nunitv2/docs/2.6.4/setup.html) and [SetUpFixture](https://nunit.org/nunitv2/docs/2.6.4/setupFixture.html). 

xUnit follows a slightly diff pattern.

whitneyschmidt:
@monojenkins backport d16-8

whitneyschmidt:
@monojenkins backport main

tests/monotouch-test/AVFoundation/AVAudioFormatTest.cs
```diff
@@ -30,8 +28,11 @@ public void TestEqualOperatorNull ()
 				Assert.IsFalse (format == null, "format == null");
 				Assert.IsFalse (null == format, "null == format");
 			}
-			Assert.IsTrue (nullFormat == null, "nullFormat == null");
-			Assert.IsTrue (null == nullFormat, "null == nullFormat");
+			using (AVAudioFormat nullFormat = null)
+			{
+				Assert.IsTrue (nullFormat == null, "nullFormat == null");
+				Assert.IsTrue (null == nullFormat, "null == nullFormat");
+			}
```
spouliot, line 35: That was not really needed, but it's not harmful (keep it).

Adding the `using` is the same as 
```csharp
AVAudioFormat nullFormat;
try {
    nullFormat = null;
    ...
} finally {
   nullFormat?.Dispose ();
}
```
and we know that `Dispose` won't be called, so just more core is generated in case an exception occurs.

Not that it's possible recent C# compiler versions (or the JIT/AOT compiler) simplify this even since it's obviously always `null`.

rolfbjarne (approved):

PR #9387, opened 2020-08-17
[StoreKit] Update framework to Xcode 12 beta 4.

mandel-macaque:


src/storekit.cs
```diff
@@ -245,6 +250,15 @@ interface SKPaymentQueue {
...
 		[Export ("showPriceConsentIfNeeded")]
 		void ShowPriceConsentIfNeeded ();
+
+		[NoWatch, NoTV, NoMac, iOS (14,0)]
+		[Export ("presentCodeRedemptionSheet")]
+		void PresentCodeRedemptionSheet ();
+
+		[Watch (7, 0), TV (14, 0), Mac (11, 0), iOS (14, 0)]
+		[Export ("transactionObservers")]
+		[Protocolize]
```
spouliot, line 260: we stopped using this after XAMCORE_2_0 - that was only useful when supporting *classic*
use the interface for the protocol instead

side note: we have warnings about exposing concrete (vs interface) types in bindings but I'm not 💯% sure they all work with `[Protocolize]`

src/storekit.cs
```diff
@@ -525,34 +545,51 @@ interface SKStoreProductViewController {
 		void LoadProduct (StoreProductParameters parameters, [NullAllowed] Action<bool,NSError> callback);
 	}
 
+	[Mac (11,0), NoTV, NoWatch]
```
spouliot, line 548: missing `[iOS]` attribute

src/storekit.cs
```diff
@@ -525,34 +545,51 @@ interface SKStoreProductViewController {
...
+	[Model]
+	[Protocol]
+	interface SKStoreProductViewControllerDelegate {
+		[Export ("productViewControllerDidFinish:"), EventArgs ("SKStoreProductViewController")]
+		void Finished (SKStoreProductViewController controller);
+	}
+
 	[NoWatch]
-	[NoMac]
+	[Mac (11,0)]
```
spouliot, line 558: why ?
it looks like you added `[NoMac] on all members, so only an empty type will be generated

src/storekit.cs
```diff
@@ -525,34 +545,51 @@ interface SKStoreProductViewController {
...
 		[Export ("AdNetworkTimestamp")]
 		uint AdNetworkTimestamp { get; set; }
+
+		[NoWatch, NoMac, iOS (14, 0)]
+		[Export ("AdNetworkSourceAppStoreIdentifier")]
+		string AdNetworkSourceAppStoreIdentifier { get; }
+
+		[NoWatch, NoMac, iOS (14, 0)]
+		[Export ("AdNetworkVersion")]
+		string AdNetworkVersion { get; }
```
spouliot, line 588: those should be added in `SKStoreProductParameterKey` too

src/storekit.cs
spouliot: hmm... that was moved :(
I rather have more `#if XXX` than moving code (to preserve git history)

src/storekit.cs
spouliot: looks like they are - but code moved and the diff is much harder to read

src/storekit.cs
```diff
@@ -797,9 +834,16 @@ interface SKProductStorePromotionController {
 	[DisableDefaultCtor] // Not specified but very likely
 	interface SKStoreReviewController {
 
+		[Deprecated (PlatformName.iOS, 9, 0, message : "Use the 'RequestReview (UIWindowScene windowScene)' API instead.")]
```
spouliot, line 837: nope, because
* this API was added in iOS 10.3 - so it can't be deprecated in iOS 9.0
* `UIWindowScene` was added in iOS 13 (iirc)

src/storekit.cs
```diff
@@ -797,9 +834,16 @@ interface SKProductStorePromotionController {
...
 		[Static]
 		[Export ("requestReview")]
 		void RequestReview ();
+
+		[Introduced (PlatformName.MacCatalyst, 14, 0)]
+		[NoWatch, NoTV, iOS (14,0), Mac (11,0)]
```
spouliot, line 843: `[Mac (11,0)]` is incorrect as `UIWindowScene` does not exists in macOS
This compiled because you defined it to be `NSObject` but that won't work at runtime

src/storekit.cs
```diff
@@ -955,4 +1004,141 @@ interface SKArcadeService {
...
+
+		[Export ("position", ArgumentSemantic.Assign)]
+		SKOverlayPosition Position { get; set; }
+
+		[Export ("setAdditionalValue:forKey:")]
+		void SetAdditionalValue ([NullAllowed] NSObject value, string key);
+
+		[Export ("additionalValueForKey:")]
+		[return: NullAllowed]
+		NSObject GetAdditionalValue (string key);
```
spouliot, line 1087: I wonder if this should not be exposed as an indexer
`NSObject AdditionalValues [string key] { get; set; }`

src/storekit.cs
```diff
@@ -955,4 +1004,141 @@ interface SKArcadeService {
...
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[Advice ("This API is not available when using UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface SKOverlay {
+		[Export ("initWithConfiguration:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (SKOverlayConfiguration configuration);
+
+		[iOS (14,0)]
```
spouliot, line 1124: not needed (already on type)

src/storekit.cs
```diff
@@ -955,4 +1004,141 @@ interface SKArcadeService {
...
+		[Export ("presentInScene:")]
+		void PresentInScene (UIWindowScene scene);
+
+		[iOS (14,0)]
```
spouliot, line 1128: not needed (already on type)

src/storekit.cs
mandel-macaque: They do not, and I used because it was already in the file, doing it the right way.

src/storekit.cs
mandel-macaque: They where added or the code would not generate correctly.

src/storekit.cs
mandel-macaque: we need the type for:
```
		[Export ("loadProductWithParameters:completionBlock:")][Internal]
		[Async]
		void LoadProduct (NSDictionary parameters, [NullAllowed] Action<bool,NSError> callback);

		[Wrap ("LoadProduct (parameters.GetDictionary ()!, callback)")]
		[Async]
		void LoadProduct (StoreProductParameters parameters, [NullAllowed] Action<bool,NSError> callback);
```
It is a dictionary without those getters setters. So I think is easier to remove the type, and do not expose the Wrapped one.

mandel-macaque: Correction, the strong dictionary is missing values from an older binding, is the better approach.

src/storekit.cs
mandel-macaque: Was added in iOS 6,0.

src/storekit.cs
```diff
@@ -763,8 +744,65 @@ enum SKCloudServiceSetupMessageIdentifier {
...
 	}
+#endif
 
-	[NoWatch, iOS (11,0), TV (11,0)]
+	[Mac (11,0), Watch (7,0), iOS (9,3), TV (9,2)]
+	[BaseType (typeof (NSObject))]
+#if XAMCORE_3_0 // Avoid breaking change in iOS
+	[DisableDefaultCtor]
+#endif
+	interface SKCloudServiceController {
```
rolfbjarne, line 754: Any reason to move this class? It makes the diff harder to review :/

mandel-macaque: Moved it out of the #if over adding extra #if #else

src/storekit.cs
spouliot: yeah, larger diff (than required) due to code being moved

src/storekit.cs
spouliot: yeah, better the extra `#if` next time
or do a separate PR to move things
moving and adding makes the diff more difficult to review

src/storekit.cs
```diff
@@ -525,34 +546,59 @@ interface SKStoreProductViewController {
 		void LoadProduct (StoreProductParameters parameters, [NullAllowed] Action<bool,NSError> callback);
 	}
 
+	[Mac (11,0), NoTV, NoWatch]
+	[BaseType (typeof (NSObject))]
+	[Model (AutoGeneratedName = true)]
```
spouliot, line 551: did you add `AutoGeneratedName = true`
iirc that's a potential breaking change in existing models
c.c. @rolfbjarne

mandel-macaque: @spouliot added as per @rolfbjarne request

spouliot: hmm... is this a new type ? (then it needs it)
or does it look new because the diff is hard to read (and the code just moved) ? (then I don't think it should be there)

mandel-macaque: It was moved. So no need.
```suggestion
	[Model]
```

src/StoreKit/SKOverlayAppClipConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+	public partial class SKOverlayAppClipConfiguration
+	{
+		public NSObject this[string i] {
+			get => GetAdditionalValue (i); 
+			set => SetAdditionalValue (value, i);
+		}
+	}
+#endif
+}
+#endif
```
rolfbjarne, line 20: No new line at end of file

src/StoreKit/SKOverlayAppConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+	public partial class SKOverlayAppConfiguration 
+	{
+		public NSObject this[string i] {
+			get => GetAdditionalValue (i); 
+			set => SetAdditionalValue (value, i);
+		}
+	}
+#endif
+}
+#endif
```
rolfbjarne, line 20: No newline at end of file

src/storekit.cs
spouliot: Please don't resolve a conversation you have not started. Makes things harder to review again.

PR #9396, opened 2020-08-18
[DevOps] Just clean the xamarin-macios dir to ensure we do not delete maccore creds.

mandel-macaque:
The credentials for maccore are downloaded to a pat file (to be found).
When we call make git-clean, because we do use the -x options, all
files are deleted, including the pat file.

We move to call git clean -xdf inside xamanrin-macios, which will delete
the test result files.

Once we find the exact path pattern, we can update the make git-clean to
not remove them but this commit unblocks the failing CI builds.

### DIFF
--- a/tools/devops/device-tests/templates/device-tests.yml
+++ b/tools/devops/device-tests/templates/device-tests.yml
@@ -35,7 +35,7 @@ steps:
 - checkout: maccore
   persistCredentials: true  # hugely important, else there are some scripts that check a single file from maccore that will fail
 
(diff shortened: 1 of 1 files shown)


mandel-macaque:
@spouliot working example: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3989032&view=logs&j=8dce0938-6154-51dc-3f74-a72db5aff65a&t=80280380-dab3-5688-5071-fd8cfa4508bd

mandel-macaque:
@monojenkins backport d16-8

PR #9399, opened 2020-08-19
[monotouch-test] Fix typo in Foundation/UrlTest

whitneyschmidt:


mandel-macaque:
Good catch!

whitneyschmidt:
> Approving, but can we move to a Assert.IsNull instead, that way we wont make the mistake again.

Hmmm...I can take a closer look tomorrow. I think we want to test the '==' operator itself (need to check the source code again) so unless .isNull calls that operator under the hood we might need to leave it as-is. Will get back to you!

mandel-macaque:
@whitneyschmidt they do

mandel-macaque:
@spouliot yep, although they do use == I agree with making it more explicit, for reference:
```csharp

/// <summary>
        /// Applies the constraint to an actual value, returning a ConstraintResult.
        /// </summary>
        /// <param name="actual">The value to be tested</param>
        /// <returns>A ConstraintResult</returns>
        public override ConstraintResult ApplyTo<TActual>(TActual actual)
        {
            return new ConstraintResult(this, actual, actual == null);
        }

```

rolfbjarne (approved):

mandel-macaque:
The failing test is unrelated:
```
<message>
<![CDATA[ Status not ok Expected: Unauthorized But was: GatewayTimeout ]]>
</message>
<stack-trace>
<![CDATA[ at MonoTests.System.Net.Http.MessageHandlerTest.GHIssue8342 (System.Net.HttpStatusCode expectedStatus, System.String validUsername, System.String validPassword, System.String username, System.String password) [0x000bd] in /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/monotouch-test/System.Net.Http/MessageHandlers.cs:532 ]]>
```

I've added an issue since the parsing does not show it properly: https://github.com/xamarin/xamarin-macios/issues/9400

spouliot:
Updated labels since this is not a shipping (release) change

tj-devel709:
This may be a dumb question, but why do we even need the second check? Can null ever be something other than null?

rolfbjarne:
> This may be a dumb question, but why do we even need the second check? Can null ever be something other than null?

It's to test that this equality operator:

https://github.com/xamarin/xamarin-macios/blob/c20cbfe3012bd6d5bf0a5e1e15fcf985dc992e19/src/Foundation/NSUrl.cs#L107-L116

works when either `x` or `y` is null (so the first test verifies the operator works when `y` is null and `x` isn't, and the second test verifies the opposite).

whitneyschmidt:
> This may be a dumb question, but why do we even need the second check? Can null ever be something other than null?

Ideally, no. For this case there's an existing override to the `==` operator, and you would _expect_ that if the first check is true the second is always true. However, since the code for `==` is manually written there could be a bug where the test fails!

I think also for operators like this we usually want to verify that the operators maintain symmetry and general arithmetic correctness - I had to look this one up, explanation here: https://en.wikipedia.org/wiki/Equivalence_relation

PR #9403, opened 2020-08-19
[dotnet] Turn the linker off by default for simulator builds.

rolfbjarne:
This becomes a bit complicated because we have to wait to determine the default value
for the linker until we know if we're building for the simulator or not (which happens
in the _DetectSdkLocations target).


rolfbjarne:
build

PR #9406, opened 2020-08-19
[xcode12] Xcode 12 Beta 5 Initial bump

dalexsoto:


src/mlcompute.cs
```diff
@@ -334,19 +335,11 @@ interface MLCActivationLayer {
 		MLCActivationLayer EluLayer { get; }
 
 		[Static]
-#if MONOMAC
```
mandel-macaque, line 337: @dalexsoto as expected hahahaha

tools/common/Frameworks.cs
```diff
@@ -547,7 +547,7 @@ public static void Gather (Application app, AssemblyDefinition product_assembly,
 	// This checks if a framework is unavailable due to bugs in Xcode (such as Apple forgetting to ship a library or headers for a framework, which seems to happen at least once a year).
 	public static bool IsFrameworkBroken (Application app, string framework)
 	{
-		if (app.IsSimulatorBuild && Driver.XcodeProductVersion == "12A8179i" /* Xcode 12 beta 4 */) {
+		if (app.IsSimulatorBuild && Driver.XcodeProductVersion == "12A8189h" /* Xcode 12 beta 5 */) {
```
rolfbjarne, line 550: I wonder if Apple is going to ship these frameworks at all for the simulator... by the next beta this code should probably be made permanent, instead of checking for the exact Xcode product version.

rolfbjarne (approved):

tools/common/Frameworks.cs
spouliot: > I wonder if Apple is going to ship these frameworks at all for the simulator...

In case of MLCompute the framework exists on macOS so it should not be a technical limitation.

> by the next beta this code should probably be made permanent

It's still a good reminder that this code needs to be verified - even if we have the meta tracking issue. Still I'd wait for the GM. The PR code is ready well before the bots are updated, so it's not really on the critical path for releases.

spouliot:
Failure is a known, random issue on bots -> https://github.com/xamarin/maccore/issues/2154

PR #9426, opened 2020-08-20
[Intents] Update bindings to Xcode 12 Beta 5

dalexsoto:


mandel-macaque:
@dalexsoto why the ignores on tvOS?

dalexsoto:
Because they availability errors on tvOS those do not make sense to have.

src/intents.cs
```diff
@@ -13688,39 +13705,55 @@ interface INCar : NSCopying, NSSecureCoding {
...
 		string CarIdentifier { get; }
 
 		[NullAllowed, Export ("displayName")]
 		string DisplayName { get; }
 
-		[BindAs (typeof (nint?))]
-		[Export ("year", ArgumentSemantic.Copy)]
-		NSNumber Year { get; }
+		[NullAllowed, Export ("year")]
+		string Year { get; }
```
rolfbjarne, line 13717: I wonder what kind of year can't be described with an NSNumber/nint?

PR #9427, opened 2020-08-20
[CoreLocation] Update framework to Xcode 12 beta 5.

mandel-macaque:


src/corelocation.cs
```diff
@@ -194,30 +194,29 @@ partial interface CLLocationManager {
 		[Export ("stopUpdatingLocation")]
 		void StopUpdatingLocation ();
 
+		[NoMac]
```
spouliot, line 197: breaking change ?

mandel-macaque: Good catch.

src/corelocation.cs
```diff
@@ -434,6 +421,10 @@ partial interface CLLocationManager {
 		[Export ("requestTemporaryFullAccuracyAuthorizationWithPurposeKey:")]
 		void RequestTemporaryFullAccuracyAuthorization (string purposeKey);
 
+		[NoWatch, NoTV, Mac (11, 0), iOS (14, 0)]
+		[Export ("authorizedForWidgetUpdates")]
+		bool AuthorizedForWidgetUpdates { [Bind ("isAuthorizedForWidgetUpdates")] get; }
```
whitneyschmidt, line 426:
```suggestion
		bool IsAuthorizedForWidgetUpdates { [Bind ("isAuthorizedForWidgetUpdates")] get; }
```

PR #9451, opened 2020-08-21
[dotnet] Bump .NET and the linker.

rolfbjarne:
The old linker run into a problem that seems fixed in the new linker, so bump.

This also required a few other fixes to keep things working, since there's a
new bug in .NET + some difference in behavior we had to deal with.


rolfbjarne:
Test failures are unrelated

* link all: https://github.com/xamarin/maccore/issues/2295
* monotouch-test: https://github.com/xamarin/maccore/issues/581

PR #9460, opened 2020-08-24
[dotnet] Implement support for our different link modes.

rolfbjarne:
Tell the managed linker what to do with each input assembly depending the
selected link mode (link all, link sdk, don't link).

Rename LinkMode.All to LinkMode.Full in mmp so that we can continue to use
Enum.Parse<LinkMode> to parse 'Full' as the link mode.


rolfbjarne:
Test failure is unrelated, and will be fixed in PR #9467.

PR #9469, opened 2020-08-24
[dotnet] It's time to start signing apps.

rolfbjarne:
Also make sure that _CompileToNative never runs in .NET mode (some of the
signing targets has _CompileToNative as a dependency, but _CompileToNative
must never be run when in .NET mode).


mandel-macaque:
Same error as in https://github.com/xamarin/xamarin-macios/pull/9470 so it does not look related to this change.

mandel-macaque:
Ok, is being addressed on https://github.com/xamarin/xamarin-macios/pull/9467

rolfbjarne:
Test failure is unrelated, and will be fixed in PR #9467.

PR #9471, opened 2020-08-24
[msbuild] Share the Pack/UnpackLibraryResources targets between Xamarin.iOS and Xamarin.Mac.

rolfbjarne:
There were 2 differences that had to be reconciled:

* The Xamarin.Mac version took 'IntermediateOutputPath' as the intermediate
  output path, while the Xamarin.iOS version used
  'DeviceSpecificIntermediateOutputPath'.
  'DeviceSpecificIntermediateOutputPath' is defined to be
  'IntermediateOutputPath' for Xamarin.Mac, so make the shared version use
  'DeviceSpecificIntermediateOutputPath', since that works in both cases.

* The Xamarin.Mac version was writing the output of PackLibraryResources to
  the FileWrites item group. This was removed on purpose from the Xamarin.iOS
  version in e97d69b25cc2b85e89feb775169f46f467c66f7a, so it seemed best to
  leave it out in the shared version as well.

Also re-use the _EmbeddedResourcePrefix variable to determine the resource prefix.

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
@@ -274,30 +274,6 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		</CreateInstallerPackage>
 	</Target>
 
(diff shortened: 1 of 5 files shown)


rolfbjarne:
Test failure is unrelated, and was fixed in PR #9467.

PR #9476, opened 2020-08-24
[AppKit] Update bindings to Xcode 12 Beta 5

tj-devel709:
I have changed my deprecated warning as discussed in the meeting today (Aug. 24, 2020)


spouliot:
@tj-devel709 the previous PR was merged so you must go back on `xcode12` to do any further changes (simplest way). Right now your PR has a conflict because you're duplicating your changes (and have a different solution/code for the message).

src/appkit.cs
```diff
@@ -27918,11 +27929,14 @@ interface NSSearchToolbarItem
 		[Export ("resignsFirstResponderWithCancel")]
 		bool ResignsFirstResponderWithCancel { get; set; }
 
-		[Export ("minimumWidthForSearchFieldRepresentation")]
-		nfloat MinimumWidthForSearchFieldRepresentation { get; set; }
+		[Export ("preferredWidthForSearchField")]
```
whitneyschmidt, line 27932: Availability (here and below) if this is newer than `[Mac (10,9)]`?

I'm also a little confused re the removal of `minimumWidthForSearchFieldRepresentation` and `preferredWidthForSearchFieldRepresentation` - usually if we remove APIs it's a breaking change since customers may be using the API. Were the names of these APIs changed/removed between betas for Xcode 12? 

It's also likely that I'm missing something here so please feel free to correct me @tj-devel709 :)

tj-devel709: Yeah I think you're totally right about missing the `[Mac (11,0)]` for these, thanks! 
And yes the `MinimumWidthForSearchFieldRepresentation` and `PreferredWidthForSearchFieldRepresentation` were introduced in xcode12 beta1!

PR #9479, opened 2020-08-24
[CarPlay] [Xcode 12] Update through beta 6

whitneyschmidt:
TODO:
https://github.com/xamarin/maccore/issues/2297
https://github.com/xamarin/maccore/issues/2298

- [x] Add tests for manual code.


src/CarPlay/CPMessageListItem.cs
```diff
@@ -0,0 +1,50 @@
...
+
+    public enum  CPMessageListItemType {
+        Identifier,
+        FullName,
+    }
+
+	public partial class CPMessageListItem {
+
+		public CPMessageListItem (string identifier, string text, CPMessageListItemLeadingConfiguration leadingConfiguration,
+
```
mandel-macaque, line 27: Weird new line here, right?

src/carplay.cs
```diff
@@ -105,6 +105,61 @@ enum CPTripEstimateStyle : ulong {
...
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[Native]
+	public enum CPInformationTemplateLayout : long
+	{
+		Leading = 0,
+		TwoColumn,
+	}
+
+	[Native]
```
mandel-macaque, line 124: Missing availability attrs.

src/carplay.cs
```diff
@@ -105,6 +105,61 @@ enum CPTripEstimateStyle : ulong {
...
+	public enum CPListItemAccessoryType : long
+	{
+		None = 0,
+		DisclosureIndicator,
+		Cloud,
+	}
+
+	[Native]
```
mandel-macaque, line 132: Same

src/carplay.cs
```diff
@@ -105,6 +105,61 @@ enum CPTripEstimateStyle : ulong {
...
+	public enum CPListItemPlayingIndicatorLocation : long
+	{
+		Leading = 0,
+		Trailing,
+	}
+
+	[Native]
```
mandel-macaque, line 139: Same

src/carplay.cs
```diff
@@ -105,6 +105,61 @@ enum CPTripEstimateStyle : ulong {
...
+	public enum CPMessageLeadingItem : long
+	{
+		None = 0,
+		Pin,
+		Star,
+	}
+
+	[Native]
```
mandel-macaque, line 147: Same

src/carplay.cs
```diff
@@ -250,6 +327,49 @@ interface CPInterfaceController {
 
 		[Export ("templates", ArgumentSemantic.Strong)]
 		CPTemplate [] Templates { get; }
+
+		// Not sure about this one. ObjC header says: extern NSString * const CarPlayErrorDomain;
+		[Field ("CarPlayErrorDomain")]
+		NSString CarPlayErrorDomain { get; }
```
mandel-macaque, line 333: Is an error domain. grep for it, there are a lot of examples of this.

src/carplay.cs
```diff
@@ -383,6 +595,7 @@ interface ICPListTemplateDelegate { }
 	interface CPListTemplateDelegate {
 
 		[Abstract]
+		[Async]
```
mandel-macaque, line 598: From the docs, I don't think this is an async call: 

> When implementing this method, perform any necessary operations to prepare for completing the item selection, including updating your user interface. You must call completionHandler after handling the selected item to tell the system that it can continue.


Reference: https://developer.apple.com/documentation/carplay/cplisttemplatedelegate/2977578-listtemplate?language=objc

src/carplay.cs
```diff
@@ -395,7 +608,7 @@ interface CPManeuver : NSCopying, NSSecureCoding {
 		[Deprecated (PlatformName.iOS, 13,0, message: "Use 'CPManeuver.SymbolImage' instead.")]
 		[NullAllowed, Export ("symbolSet", ArgumentSemantic.Strong)]
 		CPImageSet SymbolSet { get; set; }
-
```
mandel-macaque, line 398: Noise.

src/carplay.cs
```diff
@@ -577,7 +825,7 @@ interface CPMapTemplateDelegate {
 	[DisableDefaultCtor]
 	interface CPNavigationAlert : NSSecureCoding {
 
-		
+
```
mandel-macaque, line 828: Noise.

src/carplay.cs
```diff
@@ -657,10 +905,12 @@ interface ICPSearchTemplateDelegate { }
 	interface CPSearchTemplateDelegate {
 
 		[Abstract]
+		[Async]
```
mandel-macaque, line 908: Same, I don't think this is async:

> After the method retrieves the search result, the method must call completionHandler for the system to display the result to the user.

Reference: https://developer.apple.com/documentation/carplay/cpsearchtemplatedelegate/2968421-searchtemplate?language=objc

src/carplay.cs
```diff
@@ -657,10 +905,12 @@ interface ICPSearchTemplateDelegate { }
...
 		[Export ("searchTemplate:updatedSearchText:completionHandler:")]
 		void UpdatedSearchText (CPSearchTemplate searchTemplate, string searchText, CPSearchTemplateDelegateUpdateHandler completionHandler);
 
 		[Abstract]
+		[Async]
```
mandel-macaque, line 913: Same.

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+	[DisableDefaultCtor]
+	interface CPBaseListItem : NSSecureCoding
+	{
+		[Deprecated (PlatformName.iOS, 14, 0)]
+		[NullAllowed, Export ("listItemHandler", ArgumentSemantic.Copy)]
+		CPListItemHandler ListItemHandler { get; set; }
+
+		[NullAllowed, Export ("userInfo", ArgumentSemantic.Strong)]
+		NSObject UserInfo { get; set; }
+	}
```
mandel-macaque, line 1308: If it was added and deprecated in iOS 14, why even add it? Our users have never seen it, we can move it to ignore AFAIK.

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+		[NullAllowed, Export ("text")]
+		string Text { get; }
+
+		[Abstract]
+		[NullAllowed, Export ("userInfo", ArgumentSemantic.Strong)]
+		NSObject UserInfo { get; set; }
+	}
+
+
+#if false
```
mandel-macaque, line 1813: On purpose?

src/CarPlay/CPMessageListItem.cs
```diff
@@ -0,0 +1,50 @@
...
+            CPMessageListItemTrailingConfiguration? trailingConfiguration, string? detailText, string? trailingText) :
+                this (identifier, text, leadingConfiguration, trailingConfiguration, detailText, trailingText, CPMessageListItemType.Identifier)
+            {
+            }
+
+		public CPMessageListItem (string identifier, string text, CPMessageListItemLeadingConfiguration leadingConfiguration, CPMessageListItemTrailingConfiguration? trailingConfiguration,
```
mandel-macaque, line 33: Manual code requires tests.

src/carplay.cs
```diff
@@ -297,13 +417,15 @@ interface CPApplicationDelegate : UIApplicationDelegate {
 	}
 
 	[NoWatch, NoTV, NoMac, iOS (12,0)]
-	[BaseType (typeof (NSObject))]
+	[BaseType (typeof (CPBaseListItem))]
```
mandel-macaque, line 420: @spouliot question about API. The class is added in iOS 14, won't this be an issue in previous versions that do have access to the class in iOS 12 and iOS 13?

tests/introspection/iOS/iOSApiProtocolTest.cs
```diff
@@ -666,6 +676,15 @@ protected override bool Skip (Type type, string protocolName)
 				case "SRVisit":
 				case "SRWebUsage":
 					return true;
+				// Xcode 12 beta 4
```
mandel-macaque, line 679: beta 4 or beta 5?

src/carplay.cs
whitneyschmidt: @mandel-macaque iirc, we saw a similar change in a recent PR.

I believe this is ok here because CPBaseListItem is a type of NSObject. Also worth noting - it looks like Apple is going to get rid of CPBaseListItem in favor of CPSelectableListItem...which isn't available and has been filed as a radar here: https://github.com/xamarin/maccore/issues/2298

```	
[Deprecated (PlatformName.iOS, 14, 0)]
[NoWatch, NoTV, NoMac, iOS (14,0)]
[DisableDefaultCtor]
interface CPBaseListItem : NSSecureCoding
```

src/carplay.cs
whitneyschmidt: Yes, see the linked issue: https://github.com/xamarin/maccore/issues/2298

Is it better to leave out for now until we know Apple is actually adding it?

src/carplay.cs
whitneyschmidt: I'm not sure about this one. Looking for CPBaseListItem in the Apple headers shows that it's still used by several API, for example:

```
API_AVAILABLE(ios(12.0)) API_UNAVAILABLE(macos, watchos, tvos)
@interface CPListItem : CPBaseListItem <CPSelectableListItem>
```

We could leave all API using CPBaseListItem out of the binding for now until Apple decides whether they're going to go with CPBaseListItem or CPSelectableListItem.

It seems like the "right" thing to do to avoid exposing an API that's only going to briefly available (and then deprecated) in the betas, but we will be leaving a not insignificant chunk of code unbound of we do this.

@spouliot and others, thoughts?

tests/introspection/iOS/iOSApiProtocolTest.cs
whitneyschmidt: good catch! beta 5 :)

src/carplay.cs
whitneyschmidt: I looked around at some of our existing examples. It appears that ErrorDomain usually goes with an error enum, like this one in CoreText:

```
[Native]
[ErrorDomain ("kCTFontManagerErrorDomain")]
 public enum CTFontManagerError : long {
                None = 0,
...
```

`CarPlayErrorDomain` shouldn't be a field inside of `CPInterfaceController` 👍 

I don't see any enums inside of CarPlay.cs or the CarPlay dir that look like error enums so I am stumped on where this should go.

spouliot: indeed, there's no enum.
```
grep "enum CP" *.g.cs
ios-14.0.g.cs:public enum CPAlertActionStyle : ulong
ios-14.0.g.cs:public enum CPBarButtonStyle : long
ios-14.0.g.cs:public enum CPBarButtonType : ulong
ios-14.0.g.cs:public enum CPTextButtonStyle : long
ios-14.0.g.cs:public enum CPInformationTemplateLayout : long
ios-14.0.g.cs:public enum CPListItemAccessoryType : long
ios-14.0.g.cs:public enum CPListItemPlayingIndicatorLocation : long
ios-14.0.g.cs:public enum CPNavigationAlertDismissalContext : ulong
ios-14.0.g.cs:public enum CPTripPauseReason : ulong
ios-14.0.g.cs:public enum CPPanDirection : long
ios-14.0.g.cs:public enum CPManeuverDisplayStyle : long
ios-14.0.g.cs:public enum CPTimeRemainingColor : ulong
ios-14.0.g.cs:public enum CPTripEstimateStyle : ulong
ios-14.0.g.cs:public enum CPMessageLeadingItem : long
ios-14.0.g.cs:public enum CPMessageTrailingItem : long
ios-14.0.g.cs:public enum CPLimitableUserInterface : ulong
ios-14.0.g.cs:public enum CPContentStyle : ulong
```

 I guess one could be added later... the error domain is rarely used in code (at least for comparison). I suggest you remove it and add this (lack of related enum) in the `.ignore` file. If an enum is added later (beta or release) then it can be added to the _right_ place.

src/carplay.cs
spouliot: @mandel-macaque it's not a problem, nor a breaking change.

We did have a bug (last year or so) where this caused an issue for 32bits code (because the new type would throw an exception as not being available on 32bits OS). This was fixed.

src/carplay.cs
```diff
@@ -348,6 +520,14 @@ interface CPListSection : NSSecureCoding {
 
 		[Export ("items", ArgumentSemantic.Copy)]
 		CPListItem [] Items { get; }
+
+		[iOS (14,0)]
+		[Export ("indexOfItem:")]
+		nuint IndexOfItem (CPBaseListItem item);
```
spouliot, line 526: no verb in the method, use `GetIndex (CPBaseListItem item);` instead

src/carplay.cs
```diff
@@ -348,6 +520,14 @@ interface CPListSection : NSSecureCoding {
...
+
+		[iOS (14,0)]
+		[Export ("itemAtIndex:")]
+		CPBaseListItem ItemAtIndex (nuint index);
```
spouliot, line 530: same, use `GetItem (nuint index);` instead

src/carplay.cs
```diff
@@ -373,6 +554,37 @@ interface CPListTemplate : CPBarButtonProviding {
...
+		nuint SectionCount { get; }
+
+		[iOS (14, 0)]
+		[Export ("itemCount")]
+		nuint ItemCount { get; }
+
+		[iOS (14,0)]
+		[Export ("indexPathForItem:")]
+		[return: NullAllowed]
+		NSIndexPath IndexPathForItem (CPBaseListItem item);
```
spouliot, line 579: same, `GetIndexPath (CPBaseListItem item);`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
 		[Export ("dashboardWindow", ArgumentSemantic.Strong)]
 		UIWindow DashboardWindow { get; }
 	}
+
+	// warning BI1115: bgen: The parameter 'completionBlock' in the delegate 'CarPlay.CPListItemHandler' does not have a [CCallback] or [BlockCallback] attribute. Defaulting to [CCallback].
+	delegate void CPListItemHandler (CPBaseListItem item, Action completionBlock);
```
spouliot, line 1294: that does not look like a C callback since it's name ends with `Block`

src/carplay.cs
spouliot: Whenever in (reasonable) doubt don't add the API
It's much easier to add it later than to remove it :)

In this case you could bind it as `@interface CPListItem : NSObject <CPSelectableListItem>`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[NullAllowed, Export ("subtitle")]
+		string Subtitle { get; set; }
+
+		[NullAllowed, Export ("informativeText")]
+		string InformativeText { get; set; }
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (CPButton))]
```
spouliot, line 1356: since `CPButton` disable `init` then this one too should
IOW add `[DisableDefaultCtor]`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+		[Export ("initWithImage:handler:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (UIImage image, [NullAllowed] Action<CPButton> handler);
+
+		[Export ("initWithHandler:")]
+		IntPtr Constructor ([NullAllowed] Action<CPButton> handler);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (CPButton))]
```
spouliot, line 1368: same, add `[DisableDefaultCtor]`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[Export ("items", ArgumentSemantic.Copy)]
+		CPInformationItem[] Items { get; set; }
+
+		[Export ("actions", ArgumentSemantic.Copy)]
+		CPTextButton[] Actions { get; set; }
+	}
+
+	// warning BI1115: bgen: The parameter 'completionHandler' in the delegate 'CarPlay.ListImageRowItemHandler' does not have a [CCallback] or [BlockCallback] attribute. Defaulting to [CCallback].
+	delegate void ListImageRowItemHandler (CPListImageRowItem item, nint index, Action completionHandler);
```
spouliot, line 1431: add `CP` as a prefix to the delegate type

double check documentation to see if it's a normal C callback or a block

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (CPBaseListItem))]
```
spouliot, line 1434: since the constructor `initWithText:images:` does not allow null `text` or `images` then it's likely `init` should not be used. Unless there's other evidence that it's available then add `[DisableDefaultCtor]` to the type.

spouliot: it's a weird one since the `Text` property allows null...
and it's not 100% clear (at least to me) if `images` == `gridImages` (looks likely)

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[NullAllowed, Export ("text")]
+		string Text { get; set; }
+
+		[NullAllowed, Export ("userInfo", ArgumentSemantic.Strong)]
+		NSObject UserInfo { get; set; }
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 1533: `[DisableDefaultCtor]` as the content can't be edited and default values likely don't make sense

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[NullAllowed, Export ("leadingImage")]
+		UIImage LeadingImage { get; }
+
+		[Export ("initWithLeadingItem:leadingImage:unread:")]
+		IntPtr Constructor (CPMessageLeadingItem leadingItem, [NullAllowed] UIImage leadingImage, bool unread);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 1550: `[DisableDefaultCtor]`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[Export ("selected")]
+		bool Selected { [Bind ("isSelected")] get; set; }
+
+		[Field ("CPNowPlayingButtonMaximumImageSize")]
+		CGSize MaximumImageSize { get; }
+	}
+
+	[BaseType (typeof (CPNowPlayingButton))]
+	interface CPNowPlayingImageButton
```
spouliot, line 1582: missing availability attributes (for us that implies it was available since iOS 7)

spouliot: `[DisableDefaultCtor]`

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+		[Export ("updateTemplates:")]
+		void UpdateTemplates (CPTemplate[] newTemplates);
+	}
+
+	interface ICPNowPlayingTemplateObserver { }
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[Protocol]
+	[Model]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 1719: not ` *Delegate`
remove `[Model]` and `[BaseType]` attributes

src/carplay.cs
```diff
@@ -1008,4 +1289,556 @@ interface CPTemplateApplicationDashboardScene {
...
+
+		[Export ("albumArtistButtonEnabled")]
+		bool IsAlbumArtistButtonEnabled { [Bind ("isAlbumArtistButtonEnabled")] get; set; }
+
+		[Export ("updateNowPlayingButtons:")]
+		void UpdateNowPlayingButtons (CPNowPlayingButton[] nowPlayingButtons);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (CPButton))]
```
spouliot, line 1761: `[DisableDefaultCtor]`

src/carplay.cs
spouliot: either way is fine
you can add a link to the issue in the source (since the original commit message takes longer to get)

src/CarPlay/CPMessageListItem.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using Foundation;
+using ObjCRuntime;
+
+#nullable enable
+
+namespace CarPlay {
+
+    public enum  CPMessageListItemType {
```
rolfbjarne, line 19: Extra space:

```suggestion
    public enum CPMessageListItemType {
```

src/CarPlay/CPMessageListItem.cs
```diff
@@ -0,0 +1,50 @@
...
+            case CPMessageListItemType.Identifier:
+                InitializeHandle (InitWithConversationIdentifier (identifier, text, leadingConfiguration, trailingConfiguration, detailText, trailingText),
+                    "initWithConversationIdentifier:text:leadingConfiguration:trailingConfiguration:detailText:trailingText:");
+                break;
+            case CPMessageListItemType.FullName:
+                InitializeHandle (InitWithFullName (identifier, text, leadingConfiguration, trailingConfiguration, detailText, trailingText),
+                    "initWithFullName:phoneOrEmailAddress:leadingConfiguration:trailingConfiguration:detailText:trailingText:");
+                break;
+            default:
+                throw new ArgumentException ("The 'CPMessageListItemType type' argument needs a value.");
```
rolfbjarne, line 46: I think this is simpler and more accurate:

```suggestion
                throw new ArgumentOutOfRangeException (nameof (type));
```

rolfbjarne (approved):

src/carplay.cs
whitneyschmidt: Removed in beta 6! 💯

src/carplay.cs
whitneyschmidt: `extern NSUInteger const CPMaximumNumberOfGridImages;` -> likely `images` == `gridImages`

Added `[DisableDefaultCtor]` 👍

src/CarPlay/CPMessageListItem.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using Foundation;
+using ObjCRuntime;
+
+#nullable enable
+
+namespace CarPlay {
+
+    public enum CPMessageListItemType {
```
mandel-macaque, line 19: Availability attributes missing.

src/CarPlay/CPMessageListItem.cs
mandel-macaque: @whitneyschmidt this has not been fixed

whitneyschmidt:
> Build success
> ✅ [Build succeeded](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15034//console)
> ✅ [API Diff (from stable)](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15034//API_20diff_20_28from_20stable_29)
> ⚠️ [API Diff (from PR only)](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15034//API_20diff_20_28PR_20only_29) (🔥 breaking changes 🔥)
> ℹ️ [Generator Diff](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15034//Generator_20Diff) (please review changes)
> ✅ [Test run succeeded](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15034//Test_20Report)

In beta 6:

```
API_AVAILABLE(ios(12.0)) API_UNAVAILABLE(macos, watchos, tvos)
@interface CPListItem : NSObject <CPSelectableListItem>
```

The api diff in beta 1 shows:

```
 API_AVAILABLE(ios(12.0)) API_UNAVAILABLE(macos, watchos, tvos)
-@interface CPListItem : NSObject <NSSecureCoding>
+@interface CPListItem : CPBaseListItem
```

It seems that the breaking change is valid if Apple meant to remove conformance to NSSecureCoding for this type. However, that seem very strange for an API that's been available since iOS 12.

dalexsoto:
I would file an apple feedback ticked and add this to our tracking issue https://github.com/xamarin/xamarin-macios/issues/8943

spouliot:
I agree with @dalexsoto, file an feedback with Apple. But in any case you re-add `INSSecureCoding` to avoid breaking changes.

whitneyschmidt:
#17461

PR #9481, opened 2020-08-24
[AppKit] Update bindings to Xcode 12 Beta 5 

tj-devel709:
I closed the last PR, made my changes, and force pushed which did not allow me to open the PR back up. I apologize for that!
Here is the link to closed PR: https://github.com/xamarin/xamarin-macios/pull/9476

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -376,7 +376,7 @@ namespace AppKit {
 		bool AllowsVibrancy { get; }
 
 		[Static, Export ("currentAppearance")]
-		[Deprecated (PlatformName.MacOSX, 10, 9, message: "Use PerformAsCurrentDrawingAppearance or CurrentDrawingAppearance instead.")]
+		[Deprecated (PlatformName.MacOSX, 10, 9, message: "Use 'CurrentDrawingAppearance' instead.")]
 		NSAppearance CurrentAppearance { get; [Bind("setCurrentAppearance:")] set; }
 
 		[Mac (11, 0)]


tj-devel709:
Hi @spouliot Manuel helped me fix the issue with the merge conflicts so hopefully it is okay now!

tj-devel709:
@whitneyschmid thanks for your comment in my last PR https://github.com/xamarin/xamarin-macios/pull/9476, but I think we were both deceived by Github. In the `Files Changed` tab, it shows arrows where you can reveal like 20 lines higher and lower than what it's showing you originally but eventually it gives up and the arrow option goes away but there is still a break in the line number. In this case it goes from around line 3XX to line 27,XXX. So the actual header above those properties and methods contained [Mac (11,0)] but its just not visible on that tab. I may not have explained it well so let me know if that makes sense. I was about to make the changes but Manuel showed me that!

whitneyschmidt:
> @whitneyschmid thanks for your comment in my last PR #9476, but I think we were both deceived by Github. In the `Files Changed` tab, it shows arrows where you can reveal like 20 lines higher and lower than what it's showing you originally but eventually it gives up and the arrow option goes away but there is still a break in the line number. In this case it goes from around line 3XX to line 27,XXX. So the actual header above those properties and methods contained [Mac (11,0)] but its just not visible on that tab. I may not have explained it well so let me know if that makes sense. I was about to make the changes but Manuel showed me that!

No worries! It totally looked fishy 🐟 Thank you for digging into it and figuring out what was going on. The same thing has happened to me too in the past 😆

spouliot:
> No worries! It totally looked fishy 🐟 

It was a good catch - even if there was no 🐟 on the line
I had to open the file myself to double check :-) Not sure why there was no up arrow to see previous lines

mandel-macaque:
The fact that GitHub decides the size of the context of a change at random is something to worry about.

rolfbjarne (approved):

Issue #9480, opened 2020-08-24
OSLog Xcode12 Bindings Postponed


tj-devel709:
Due to other higher priorities, OSLog's xcode12 bindings will be postponed until a later date.

The selectors that will not be available are the following:
**For iOS and tvOS**
OSLogEntryLogLevel
OSLogEntrySignpostType
OSLogEntryStoreCategory
OSLogEnumeratorOptions
OSLogMessageComponentArgumentCategory
OSLogStoreScope
OSLogEntryFromProcess
OSLogEntryWithPayload
OSLogStore::storeWithScope:error:
OSLogStore::storeWithURL:error: 
OSLogEntry::composedMessage
OSLogEntry::date
OSLogEntry::storeCategory
OSLogEntryActivity::parentActivityIdentifier
OSLogEntryLog::level
OSLogEntrySignpost::signpostIdentifier
OSLogEntrySignpost::signpostName
OSLogEntrySignpost::signpostType
OSLogMessageComponent::argumentCategory
OSLogMessageComponent::argumentDataValue
OSLogMessageComponent::argumentDoubleValue
OSLogMessageComponent::argumentInt64Value
OSLogMessageComponent::argumentNumberValue
OSLogMessageComponent::argumentStringValue
OSLogMessageComponent::argumentUInt64Value
OSLogMessageComponent::formatSubstring 
OSLogMessageComponent::placeholder
OSLogStore::entriesEnumeratorAndReturnError:
OSLogStore::entriesEnumeratorWithOptions:position:predicate:error:
OSLogStore::positionWithDate:
OSLogStore::positionWithTimeIntervalSinceEnd:
OSLogStore::positionWithTimeIntervalSinceLatestBoot:
OSLogEntry
OSLogEntryActivity 
OSLogEntryBoundary 
OSLogEntryLog 
OSLogEntrySignpost 
OSLogEnumerator 
OSLogMessageComponent
OSLogPosition
OSLogStore

**macOS**
The above selectors will also not be available in macOS in addition to:
OSLogStore::localStoreAndReturnError:

spouliot:
Xcode 12 GM removed iOS, tvOS and watchOS support for OSLog (back to macOS only)

spouliot:
support added in `xcode12.2` branch

PR #9482, opened 2020-08-24
[OSLog] Moving Todo File Contents to Ignore Files

tj-devel709:
Issue: https://github.com/xamarin/xamarin-macios/issues/9480

### DIFF
--- a/tests/xtro-sharpie/iOS-OSLog.ignore
+++ b/tests/xtro-sharpie/iOS-OSLog.ignore
@@ -1,3 +1,5 @@
+## Will save these bindings for later - https://github.com/xamarin/xamarin-macios/issues/9480
+
 !missing-enum! OSLogEntryLogLevel not bound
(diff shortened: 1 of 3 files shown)


rolfbjarne (approved):

tj-devel709:
> @tj-devel709 changes look good. Interestingly all the ignored enums and functions are in all platforms supported by the framework. There is an option to create a single ignore file that uses common-* as the prefix rather than the platform name. Food for though 😉

Would you recommend that I change it to common? Would it still be okay in common if it doesn't cover watchOS also?

mandel-macaque:
@tj-devel709 you can leave it as it is. But I would have moved it to common just to have to edit less. I might be lazier than the average person. Also, it was a good thing to show you ;) 

If OSLog appears on the watch, which I would be surprised, we might get those enums too. Probably will have to add yet another ignore.  I let you decide what is best, both options are valid. Since you get already several reviews, I'd land it.

PR #9490, opened 2020-08-25
[CoreHaptics] Ignore framework on MacOS X due to issues in the headers.

mandel-macaque:
Although the headers are present, the framework cannot be used since the
header references AVAudioSession and is a class that is not present in
MacOS X


tests/xtro-sharpie/macOS-CoreHaptics.ignore
```diff
@@ -111,4 +112,4 @@
 !missing-type! CHHapticEventParameter not bound
 !missing-type! CHHapticParameterCurve not bound
 !missing-type! CHHapticParameterCurveControlPoint not bound
-!missing-type! CHHapticPattern not bound
+!missing-type! CHHapticPattern not bound
```
rolfbjarne, line 115: Noise!

rolfbjarne (approved):

PR #9497, opened 2020-08-25
[dotnet] Preserve relative directory when deciding where in the app bundle files should be copied.

rolfbjarne:
There may be multiple satellite assemblies with the same name, so make sure to
put them all in the corresponding subdirectory.

This is the second of three steps to fix this test failure:

    EmbeddedResources.ResourcesTest
        [FAIL] Embedded :   en-AU
            Expected string length 5 but was 7. Strings differ at index 0.
            Expected: "G'day"
            But was:  "Welcome"
            -----------^
                at EmbeddedResources.ResourcesTest.Embedded() in [...]/xamarin-macios/tests/EmbeddedResources/ResourcesTest.cs:line 45

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -407,7 +407,7 @@
 		<ItemGroup>
 			<ResolvedFileToPublish
 				Update="@(ResolvedFileToPublish)"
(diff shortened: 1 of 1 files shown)


rolfbjarne:
build

rolfbjarne:
Test failure is unrelated

* monotouch-test/iOS 32-bits: https://github.com/xamarin/maccore/issues/1605

PR #9510, opened 2020-08-26
[xharness] Enable the .NET tests when a file containing 'dotnet' is modified.

rolfbjarne:
This also required adding support for using regex as file names when selecting
which tests to run.


tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -80,7 +81,7 @@ class TestSelector {
 		static readonly string [] dotnetPrefixes = {
 			"dotnet",
 			"msbuild",
-			"tests/dotnet",
+			".*dotnet.*",
```
spouliot, line 84: there's a `"dotnet"` entry two lines above
the variable is also misnamed/confusing now since it's not just prefixes anymore

PR #9514, opened 2020-08-26
[runtime] Give a better error message if we can't load ObjCRuntime.Runtime.Initialize ()

rolfbjarne:
Otherwise we'll get a SIGSEGV due to dereferencing NULL.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -1324,7 +1324,7 @@ xamarin_initialize ()
 	runtime_initialize = mono_class_get_method_from_name (runtime_class, "Initialize", 1);
 
 	if (runtime_initialize == NULL)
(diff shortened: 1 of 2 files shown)


runtime/runtime.m
```diff
@@ -1323,6 +1323,9 @@ -(void) xamarinSetGCHandle: (uint32_t) gc_handle;
 
 	runtime_initialize = mono_class_get_method_from_name (runtime_class, "Initialize", 1);
 
+	if (runtime_initialize == NULL)
+		xamarin_assertion_message ("Fatal error: failed to load the method '%s.%s.%s'\n.", objcruntime, "Runtime", "Initialize");
```
stephen-hawley, line 1327: Why are "Runtime" and "Initialize" used as string substitutions instead of just going into the format string?

rolfbjarne: It was a copy-paste from somewhere else. But reminds me that this could probably be simplified to use fewer new strings. I'll have a look at that.

PR #9532, opened 2020-08-27
[dotnet] Use net5.0-[ios|tvos|watchos|macos] TargetFrameworks.

rolfbjarne:
After bumping to .NET 5.0.100-rc.1.20426.3, we can use the final TargetFramework values:

    <TargetFramework>net5.0-ios</TargetFramework>

Changes required on our side:

* Set TargetPlatformSupported to 'true' if the TargetPlatformIdentifier is one
  we support (ios/tvos/watchos/macos).
* Generate a list of valid OS versions for each of our four platforms, and
  then we set SupportedTargetPlatform to that list. This is similar to how
  it's done for Windows [1].
* Set TargetPlatformVersion if it's not already set to the default OS version
  we've defined.
* Update all the project files we have to use the new TargetFramework values.

Ref: https://github.com/xamarin/xamarin-android/pull/5007
[1]: https://github.com/dotnet/sdk/blob/18ee4eac8b3abe6d554d2e0c39d8952da0f23ce5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSupportedTargetPlatforms.props

### DIFF
--- a/dotnet/generate-target-platforms.csharp
+++ b/dotnet/generate-target-platforms.csharp
@@ -25,7 +25,8 @@ doc.Load (plistPath);
 var nodes = doc.SelectNodes ($"/plist/dict/key[text()='KnownVersions']/following-sibling::dict[1]/key[text()='{platform}']/following-sibling::array[1]/string");
 
 using (TextWriter writer = new StreamWriter (outputPath)) {
-	writer.WriteLine ($"<!-- This file contains a list of the {platform} platform versions that are supported for this SDK -->");
+	writer.WriteLine ($"<!-- This file contains a generated list of the {platform} platform versions that are supported for this SDK -->");
+	writer.WriteLine ($"<!-- Generation script: https://github.com/xamarin/xamarin-macios/blob/main/dotnet/generate-target-platforms.csharp -->");
 	writer.WriteLine ("<Project>");
 	writer.WriteLine ("\t<ItemGroup>");
 
--- a/dotnet/generate-target-platforms.csharp
+++ b/dotnet/generate-target-platforms.csharp
@@ -31,11 +31,11 @@ using (TextWriter writer = new StreamWriter (outputPath)) {
 	writer.WriteLine ("\t<ItemGroup>");
 
 	foreach (XmlNode n in nodes)
-		writer.WriteLine ($"\t\t<_{platform}SupportedTargetPlatform Include=\"{n.InnerText}\" />");
+		writer.WriteLine ($"\t\t<{platform}SupportedTargetPlatform Include=\"{n.InnerText}\" />");
 
 	writer.WriteLine ("\t</ItemGroup>");
 	writer.WriteLine ("\t<ItemGroup>");
-	writer.WriteLine ($"\t\t<SupportedTargetPlatform Condition=\"'$(TargetPlatformIdentifier)' == '{platform}'\" Include=\"@(_{platform}SupportedTargetPlatform)\" />");
+	writer.WriteLine ($"\t\t<SupportedTargetPlatform Condition=\"'$(TargetPlatformIdentifier)' == '{platform}'\" Include=\"@({platform}SupportedTargetPlatform)\" />");
 	writer.WriteLine ("\t</ItemGroup>");
 	writer.WriteLine ("</Project>");
 }
--- a/dotnet/generate-target-platforms.csharp
+++ b/dotnet/generate-target-platforms.csharp
@@ -1,0 +1,42 @@
+#!/usr/bin/env /Library/Frameworks/Mono.framework/Commands/csharp
+
+// arguments are: <platform> <outputPath>
+
+using System.IO;
+using System.Xml;
+
+var args = Environment.GetCommandLineArgs ();
+var expectedArgumentCount = 2;
+if (args.Length != expectedArgumentCount + 2 /* 2 default arguments (executable + script) + 'expectedArgumentCount' arguments we're interested in */) {
+	// first arg is "/Library/Frameworks/Mono.framework/Versions/4.8.0/lib/mono/4.5/csharp.exe"
+	// second arg the script itself
+	// then comes the ones we care about
+	Console.WriteLine ($"Need {expectedArgumentCount} arguments, got {args.Length - 2}");
+	Environment.Exit (1);
+	return;
+}
+
+var platform = args [2];
+var outputPath = args [3];
+var plistPath = platform == "macOS" ? "../Versions-mac.plist.in" : "../Versions-ios.plist.in";
+
+var doc = new XmlDocument ();
+doc.Load (plistPath);
+var nodes = doc.SelectNodes ($"/plist/dict/key[text()='KnownVersions']/following-sibling::dict[1]/key[text()='{platform}']/following-sibling::array[1]/string");
+
+using (TextWriter writer = new StreamWriter (outputPath)) {
+	writer.WriteLine ($"<!-- This file contains a list of the {platform} platform versions that are supported for this SDK -->");
+	writer.WriteLine ("<Project>");
+	writer.WriteLine ("\t<ItemGroup>");
+
+	foreach (XmlNode n in nodes)
+		writer.WriteLine ($"\t\t<_{platform}SupportedTargetPlatform Include=\"{n.InnerText}\" />");
+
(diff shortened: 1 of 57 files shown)


dotnet/generate-target-platforms.csharp
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env /Library/Frameworks/Mono.framework/Commands/csharp
```
rolfbjarne, line 1: The generated files look like this https://gist.github.com/rolfbjarne/765c4e4b38b1b017380c9378d9666317

rolfbjarne:
/cc @dsplaisted @sfoslund

I think this is it for the new TargetFramework support, it seems to work fine. For our own tests I decided to not include the OS version in the TargetFramework, because that seems quite annoying to have to update in the future.

dotnet/generate-target-platforms.csharp
```diff
@@ -0,0 +1,42 @@
...
+doc.Load (plistPath);
+var nodes = doc.SelectNodes ($"/plist/dict/key[text()='KnownVersions']/following-sibling::dict[1]/key[text()='{platform}']/following-sibling::array[1]/string");
+
+using (TextWriter writer = new StreamWriter (outputPath)) {
+	writer.WriteLine ($"<!-- This file contains a list of the {platform} platform versions that are supported for this SDK -->");
+	writer.WriteLine ("<Project>");
+	writer.WriteLine ("\t<ItemGroup>");
+
+	foreach (XmlNode n in nodes)
+		writer.WriteLine ($"\t\t<_{platform}SupportedTargetPlatform Include=\"{n.InnerText}\" />");
```
jonathanpeppers, line 33: One thing I'm not sure of: should this item group have public names and drop the `_`?

Android looks like this at the moment:
```xml
<Project>
  <PropertyGroup>
    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30</TargetPlatformVersion>
  </PropertyGroup>
  <ItemGroup>
    <AndroidSupportedTargetPlatform Include="21" />
    <AndroidSupportedTargetPlatform Include="22" />
    <AndroidSupportedTargetPlatform Include="23" />
    <AndroidSupportedTargetPlatform Include="24" />
    <AndroidSupportedTargetPlatform Include="25" />
    <AndroidSupportedTargetPlatform Include="26" />
    <AndroidSupportedTargetPlatform Include="27" />
    <AndroidSupportedTargetPlatform Include="28" />
    <AndroidSupportedTargetPlatform Include="29" />
    <AndroidSupportedTargetPlatform Include="30" />
    <SupportedTargetPlatform Condition=" '$(TargetPlatformIdentifier)' == 'Android' " Include="@(AndroidSupportedTargetPlatform)" />
  </ItemGroup>
</Project>
```

jonathanpeppers: Here is what they did for `Windows`: 

https://github.com/dotnet/sdk/blob/18ee4eac8b3abe6d554d2e0c39d8952da0f23ce5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSupportedTargetPlatforms.props

dotnet/generate-target-platforms.csharp
```diff
@@ -0,0 +1,42 @@
...
+var platform = args [2];
+var outputPath = args [3];
+var plistPath = platform == "macOS" ? "../Versions-mac.plist.in" : "../Versions-ios.plist.in";
+
+var doc = new XmlDocument ();
+doc.Load (plistPath);
+var nodes = doc.SelectNodes ($"/plist/dict/key[text()='KnownVersions']/following-sibling::dict[1]/key[text()='{platform}']/following-sibling::array[1]/string");
+
+using (TextWriter writer = new StreamWriter (outputPath)) {
+	writer.WriteLine ($"<!-- This file contains a list of the {platform} platform versions that are supported for this SDK -->");
```
spouliot, line 28: minor: `a list of` -> `a generated list of` with a link back to the script ?

dsplaisted:
> /cc @dsplaisted @sfoslund
> 
> I think this is it for the new TargetFramework support, it seems to work fine. For our own tests I decided to not include the OS version in the TargetFramework, because that seems quite annoying to have to update in the future.

I would sync with @mhutch and @terrajobst on this.  When designing the `TargetFramework` changes, we had said that we would recommend being explicit about the platform versions for iOS and Android.

For a given version of .NET, we should not change the default target platform version out from under you (once we've shipped an RTM of the tools supporting that version).  So if `net6.0-ios` means version 13.6 of iOS, we can't change the SDK later on so that it means 13.7 or 14.0.  Once we add support for `net7.0-ios`, we could choose a different default platform version in that case.

So I don't know if leaving the platform version out of the `TargetFramework` would avoid needing to update the number anyway.

rolfbjarne:
> > /cc @dsplaisted @sfoslund
> > I think this is it for the new TargetFramework support, it seems to work fine. For our own tests I decided to not include the OS version in the TargetFramework, because that seems quite annoying to have to update in the future.
> 
> I would sync with @mhutch and @terrajobst on this. When designing the `TargetFramework` changes, we had said that we would recommend being explicit about the platform versions for iOS and Android.

Yes, I totally agree that the message to app developers should be that they should be explicit about the platform version (and that should be reflected in our samples and templates).

However, in this particular case, it's for our own (the Xamarin.iOS' team) tests, and things get a bit less clear.

> For a given version of .NET, we should not change the default target platform version out from under you (once we've shipped an RTM of the tools supporting that version). So if `net6.0-ios` means version 13.6 of iOS, we can't change the SDK later on so that it means 13.7 or 14.0. Once we add support for `net7.0-ios`, we could choose a different default platform version in that case.

I think one thing that should be clarified is who exactly defines the default target platform version (who is "we" in the sentence above?)

I was under the impression that it would be the SDK teams (the Xamarin.iOS team for iOS), which means it would be the Xamarin.iOS team who would be responsible for updating the default target platform version whenever that can/should be done.

And in that case, as it stands with this PR, all our tests would start using the new default once we (the Xamarin.iOS team) change it (and that seemed like the behavior I'd expect when we change the default).

> So I don't know if leaving the platform version out of the `TargetFramework` would avoid needing to update the number anyway.

This is not something we ship, so we can always change our decision later if we find out that I was wrong.

dotnet/generate-target-platforms.csharp
rolfbjarne: That's a good point, I think this looks OK to be public, so I removed the underscore.

dotnet/generate-target-platforms.csharp
rolfbjarne: > minor: `a list of` -> `a generated list of` with a link back to the script ?

Fixed!

dsplaisted:
> > For a given version of .NET, we should not change the default target platform version out from under you (once we've shipped an RTM of the tools supporting that version). So if `net6.0-ios` means version 13.6 of iOS, we can't change the SDK later on so that it means 13.7 or 14.0. Once we add support for `net7.0-ios`, we could choose a different default platform version in that case.
> 
> I think one thing that should be clarified is who exactly defines the default target platform version (who is "we" in the sentence above?)
> 
> I was under the impression that it would be the SDK teams (the Xamarin.iOS team for iOS), which means it would be the Xamarin.iOS team who would be responsible for updating the default target platform version whenever that can/should be done.
> 
> And in that case, as it stands with this PR, all our tests would start using the new default once we (the Xamarin.iOS team) change it (and that seemed like the behavior I'd expect when we change the default).

The Xamarin teams would be in charge of setting the default platform version for iOS and Android.  Once you've shipped at RTM release with it set though, you can't change it for that version of .NET.

So if you ship .NET 6 with a default iOS version of 15, then when .NET 7 comes along, the default for .NET 6 needs to stay at 15, though you can change the default for projects targeting .NET 7 to something else.

rolfbjarne:
> The Xamarin teams would be in charge of setting the default platform version for iOS and Android. Once you've shipped at RTM release with it set though, you can't change it for that version of .NET.
> 
> So if you ship .NET 6 with a default iOS version of 15, then when .NET 7 comes along, the default for .NET 6 needs to stay at 15, though you can change the default for projects targeting .NET 7 to something else.

Perfect, that's exactly as I understood it. Thanks!

Issue #9544, opened 2020-08-28
nunit2 not supported in MSBuild Tests


tj-devel709:
### Steps to Reproduce

1. Running macOS tests under MSBuild tests after calling make runner.
2. You can also reproduce it running VSfM xharness file.

### Expected Behavior
Running the macOS tests and coming back with test results.

### Actual Behavior
Main log states "Execution failed with exit code 255".
Execution log contains "Unknown result format: nunit2".
Our nunit version was 3.11.1 and the minimum required was 3.9.0 but is later classified in an nunit2 format which seems to be the issue.

### Environment

```
Branched off of main

=== Visual Studio Community 2019 for Mac ===

Version 8.7.3 (build 13)
Installation UUID: aed21816-ba65-4b20-9304-4baa776c2ad1
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000090

=== Mono Framework MDK ===

Runtime:
	Mono 6.12.0.90 (2020-02/d3daacdaa80) (64-bit)
	Package version: 612000090

=== Roslyn (Language Service) ===

3.7.0-6.20375.2+34202cc2f3e869fd70a26d8237f4552cf9e192cf

=== NuGet ===

Version: 5.7.0.6702

=== .NET Core SDK ===

SDK: /usr/local/share/dotnet/sdk/5.0.100-preview.8.20359.11/Sdks
SDK Versions:
	5.0.100-preview.8.20359.11
	3.1.401
	3.1.100
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Sdks

=== .NET Core Runtime ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	5.0.0-preview.8.20358.9
	3.1.7
	3.1.0
	2.1.21

=== Xamarin.Profiler ===

Version: 1.6.15.68
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Updater ===

Version: 11

=== Apple Developer Tools ===

Xcode 11.6 (16141)
Build 11E708

=== Xamarin.Mac ===

Version: 6.23.0.153 (Visual Studio Community)
Hash: 2848cbfb8
Branch: TJ-Translations5
Build date: 2020-08-28 14:19:45-0500

=== Xamarin.iOS ===

[...]

PR #9545, opened 2020-08-28
nunit2 format for supported in MSBuild Tests Workaround

tj-devel709:
Workaround for https://github.com/xamarin/xamarin-macios/issues/9544 where MSBuild tests are not being executed due to NUnitExecuteTask.cs requiring an nunit version of 3.9.0+ but tagging with a nunit2 format. Using this workaround, we are able to run the tests and see output in the HTML log (error 16 as of now).


mandel-macaque:
Test successfully ran, @rolfbjarne @spouliot i think the html result is good enough to unblock the translations work

rolfbjarne:
This should also work, and won't need changes to xharness: https://github.com/rolfbjarne/xamarin-macios/commit/9a746d628dd80f732573c4846b3b7923acea09e8

mandel-macaque:
@tj-devel709 please do as @spouliot and @rolfbjarne mentioned, if it works, is best. Else I'll have to update the xslt, which is not trivial

tj-devel709:
Update: still testing on my end, for some reason @rolfbjarne 's solution isn't working on my end.. Sounds like it is working for others though.

tj-devel709:
@rolfbjarne @mandel-macaque @spouliot @whitneyschmidt I changed that string back and applied Rolf's change. I needed to git clean and it worked on my end as well!

rolfbjarne (approved):

mandel-macaque (approved):

PR #9547, opened 2020-08-31
[jenkins] Clean up old simulators and Xcodes before every build.

rolfbjarne:


jenkins/clean-jenkins-bots.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# We don't care about errors in this section, we just want to clean as much as possible
+set +e
+
+# Delete all the simulator devices. These can take up a lot of space over time (I've seen 100+GB on the bots)
+/Applications/Xcode.app/Contents/Developer/usr/bin/simctl delete all
+
+# Delete old Xcodes.
+ls -lad /Applications/Xcode*.app
+sudo rm -Rf /Applications/Xcode44.app
```
stephen-hawley, line 20: Is there a reason for each of these to be their own `sudo rm` instead of
```
oldXcodes=(
   "/Applications/Xcode5.app"
   "/Applications/Xcode502.app"
   # etc
}

for oldXcode in ${oldXcodes[@]}; do
    sudo rm -Rf $oldXcode
done
```

mandel-macaque: I went that path in the yaml version: https://github.com/xamarin/xamarin-macios/blob/main/tools/devops/device-tests/scripts/System.psm1#L188

rolfbjarne: > Is there a reason for each of these to be their own `sudo rm` instead of

Not really, so I changed it.

PR #9549, opened 2020-08-31
[mtouch/mmp] Unify OptimizeGeneratedCodeSubStep into a single class.

rolfbjarne:
This makes it easier to use this code from dotnet-linker.


tools/linker/CoreOptimizeGeneratedCode.cs
```diff
@@ -935,6 +946,28 @@ int ProcessIsARM64CallingConvention (MethodDefinition caller, Instruction ins)
 			return 0;
 		}
 
+		void ProcessRuntimeArch (MethodDefinition caller, Instruction ins)
+		{
+			const string operation = "inline Runtime.Arch";
+
+			if (Optimizations.InlineRuntimeArch != true)
```
stephen-hawley, line 953: I know that this was in the previous code, but whenever a see a comparison to a constant bool I get twitchy (ask me why sometime)

rolfbjarne: That's because `InlineRuntimeArch` is a nullable bool. In this particular case, the if condition is true if `Optimizations.InlineRuntimeArch` is either `false` or `null`. It's the short version of:

```csharp
if (!Optimizations.InlineRuntimeArch.HasValue || !Optimizations.InlineRuntimeArch.Value)
    ....
```

rolfbjarne:
Test failures are unrelated

* monotouch-test/iOS 32-bit: https://github.com/xamarin/maccore/issues/2304

PR #2308, opened 2017-07-10
[opengles] Update nullability attributes up to beta 3

spouliot:
The single new API is commented until IOSurface bindings
are available [1]

[1] https://bugzilla.xamarin.com/show_bug.cgi?id=58054

PR #2299, opened 2017-07-06
[generator] Add unit test project, and port a few tests over from the makefile tests.

rolfbjarne:

PR #9557, opened 2020-09-01
[d16-7] Bump maccore to get provisioning changes

dalexsoto:
New commits in xamarin/maccore:

* xamarin/maccore@f08af6cc5c [certificates] Remove unused expired certs and renewed macOS installer cert (#2308)
* xamarin/maccore@3cda873ef0 [certificates] Use Apple Certs, One certificate to rule them all (more like two) (#2299)

Diff: https://github.com/xamarin/maccore/compare/c7b48ac49168893edbab1ce5164812b59f4be02b..f08af6cc5c27f066de65d17a62bae39ef7ae23d5

### DIFF
--- a/mk/xamarin.mk
+++ b/mk/xamarin.mk
@@ -7,7 +7,7 @@ MONO_BRANCH    := $(shell cd $(MONO_PATH) 2> /dev/null && git symbolic-ref --sho
 endif
 
 ifdef ENABLE_XAMARIN
-NEEDED_MACCORE_VERSION := c7b48ac49168893edbab1ce5164812b59f4be02b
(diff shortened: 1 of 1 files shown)


rolfbjarne (approved):

dalexsoto:
@spouliot Because we did not really just renewed the certificates, if that had been the case then yeah not bumping maccore would have applied, since I created two new Apple certs and not renewed the iOS/macOS ones, also had name changes on the files and had to change this [file](https://github.com/xamarin/maccore/commit/ab4c99c9059656110b8f5cab8c4fb07f7f2f5bb9#diff-f5e4fe0aa4ffc7b3481ed4df20a46eea) this is why this is needed. For next year only renewing the Apple certs and the installer certs should be enough.

Issue #9538, opened 2020-08-27
AVFoundation.AVAudioSourceNodeRenderHandler delegate is incorrectly bound, causing a crash when it is used as a parameter to AVAudioSourceNode constructor


KumarKodali:
### Steps to Reproduce

1. Create an instance of AVAudioSourceNode with AVAudioSourceNodeRenderHandler call back
2. Connect to AudioEngine's OutputNode using OutputNode's Bus 0 InputFormat
3. Start AudioEngine

### Expected Behavior
AVAudioSourceNodeRenderHandler delegate should be called for audio buffers.

Also, shouldn't AVAudioSourceNodeRenderHandler's isSilence parameter be a reference as AVAudioSourceNodeRenderBlock's isSilence parameter is a pointer to BOOL?

`public delegate int AVAudioSourceNodeRenderHandler (bool isSilence, AudioTimeStamp timestamp, uint frameCunt, ref AudioBuffers outputData);`

`typedef OSStatus (^AVAudioSourceNodeRenderBlock)(BOOL *isSilence, const AudioTimeStamp *timestamp, AVAudioFrameCount frameCount, AudioBufferList *outputData);`

### Actual Behavior

=================================================================
	Native Crash Reporting
=================================================================
Got a segv while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x10dd0fa45 - /Users/kumarkodali/Library/Developer/CoreSimulator/Devices/7B82221D-E0F5-4A85-8974-555DD0F30BBF/data/Containers/Bundle/Application/B3C622FE-6F02-445A-8035-D50B455F7F59/AudioTest.app/AudioTest : mono_dump_native_crash_info
	0x10dd045a5 - /Users/kumarkodali/Library/Developer/CoreSimulator/Devices/7B82221D-E0F5-4A85-8974-555DD0F30BBF/data/Containers/Bundle/Application/B3C622FE-6F02-445A-8035-D50B455F7F59/AudioTest.app/AudioTest : mono_handle_native_crash
	0x10dd1633b - /Users/kumarkodali/Library/Developer/CoreSimulator/Devices/7B82221D-E0F5-4A85-8974-555DD0F30BBF/data/Containers/Bundle/Application/B3C622FE-6F02-445A-8035- D50B455F7F59/AudioTest.app/AudioTest : mono_sigsegv_signal_handler_debug
[...]

PR #9565, opened 2020-09-01
[Bug] AVAudioSourceNodeRenderHandler delegate is incorrectly bound

whitneyschmidt:
Fixes https://github.com/xamarin/xamarin-macios/issues/9538

Header:

```
typedef OSStatus (^AVAudioSourceNodeRenderBlock)(BOOL *isSilence, const AudioTimeStamp *timestamp, AVAudioFrameCount frameCount, AudioBufferList *outputData) API_AVAILABLE(macos(10.15), ios(13.0), tvos(13.0), watchos(6.0)) ;
```

Original delegate binding:

```
delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (bool isSilence, AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
```

Verified fix locally.

Should be back ported to d16-8.


src/avfoundation.cs
```diff
@@ -13513,7 +13513,7 @@ interface AVCompositionTrackFormatDescriptionReplacement : NSSecureCoding {
 		CMFormatDescription ReplacementFormatDescription { get; }
 	}
 
-	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (bool isSilence, AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
+	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (ref bool isSilence, ref AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
```
whitneyschmidt, line 13516: In the header, timestamp is passed as a ref and also const. Is there a way to make the param both ref and const here? I tried `ref const` and `const ref` and both failed to compile.

whitneyschmidt, line 13516: Another question - do we have unit tests for delegates + should I add one for this?

rolfbjarne: Please add a test, because they're all different so there's no generic way to test all delegates.

mandel-macaque:
+1 on @spouliot we have to do the XAMCORE_4 dance etc

src/avfoundation.cs
mandel-macaque: API breaks :/

src/avfoundation.cs
```diff
@@ -13513,20 +13513,20 @@ interface AVCompositionTrackFormatDescriptionReplacement : NSSecureCoding {
 		CMFormatDescription ReplacementFormatDescription { get; }
 	}
 
-	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (bool isSilence, AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
+	    delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler2 (ref bool isSilence, ref AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
```
whitneyschmidt, line 13516: Fix weird spacing.

whitneyschmidt:
An additional note here:

Since the original API crashed, it's okay to give it a non-functional implementation because it never worked in the first place.

src/AVFoundation/AVCompat.cs
```diff
@@ -169,6 +169,22 @@ public AVAudioUnitComponentManager ()
 		}
 	}
 
+	public delegate int AVAudioSourceNodeRenderHandler (bool isSilence, AudioToolbox.AudioTimeStamp timestamp, uint frameCount, ref AudioToolbox.AudioBuffers outputData);
+
+	partial class AVAudioSourceNode {
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioSourceNodeRenderHandler2)' instead.")]
+		public IntPtr Constructor (AVAudioSourceNodeRenderHandler renderHandler)
```
spouliot, line 176: missed it earlier but that's incorrect - it's how we bind constructors
however for manual bindings you need to create _normal_ constructor

```csharp
public AVAudioSourceNode (AVAudioSourceNodeRenderHandler renderHandler)
```

API diff complains about it

src/AVFoundation/AVCompat.cs
```diff
@@ -169,6 +169,22 @@ public AVAudioUnitComponentManager ()
...
+		{
+			throw new NotImplementedException ();
+		}
+
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioFormat, AVAudioSourceNodeRenderHandler2)' instead.")]
+		public IntPtr Constructor (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler)
```
spouliot, line 182: same

spouliot (changes requested):

src/avfoundation.cs
```diff
@@ -13513,19 +13513,19 @@ interface AVCompositionTrackFormatDescriptionReplacement : NSSecureCoding {
 		CMFormatDescription ReplacementFormatDescription { get; }
 	}
 
-	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (bool isSilence, AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
+	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler2 (ref bool isSilence, ref AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
```
mandel-macaque, line 13516: Here, I'd do the following for XAMCORE_4:
```suggestion
#if !XAMCORE_4
	delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler2 (ref bool isSilence, ref AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
#else
#6564 delegate /* OSStatus */ int AVAudioSourceNodeRenderHandler (ref bool isSilence, ref AudioTimeStamp timestamp, uint frameCount, ref AudioBuffers outputData);
```

And we have a nice name when we can break the API.

src/avfoundation.cs
```diff
@@ -13513,19 +13513,19 @@ interface AVCompositionTrackFormatDescriptionReplacement : NSSecureCoding {
...
 	interface AVAudioSourceNode : AVAudioMixing {
 		[Export ("initWithRenderBlock:")]
 		[DesignatedInitializer]
-		IntPtr Constructor (AVAudioSourceNodeRenderHandler renderHandler);
+		IntPtr Constructor (AVAudioSourceNodeRenderHandler2 renderHandler);
 
 		[Export ("initWithFormat:renderBlock:")]
 		[DesignatedInitializer]
-		IntPtr Constructor (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler);
+		IntPtr Constructor (AVAudioFormat format, AVAudioSourceNodeRenderHandler2 renderHandler);
```
mandel-macaque, line 13528: If we do what I added in the previous comment, we need to do the same here, use the nicer name when we can break the API.
```csharp
#if !XAMCORE_4
IntPtr Constructor (AVAudioSourceNodeRenderHandler2 renderHandler);
#else
IntPtr Constructor (AVAudioSourceNodeRenderHandler renderHandler);
#endif
```

And the same with the other constructor.

mandel-macaque (changes requested):

src/avfoundation.cs
whitneyschmidt: A couple of questions:

1. iirc, `XAMCORE_4` is not defined yet, so we use it more for annotation than functionality. Given this, I'm not sure that the code you've recommended below would compile.

2. Please correct as needed: It seems like the AVCompat.cs file contains C# code (not bindings code) for `[Obsolete]` or broken APIs that we want to "remove" from the code w/o creating breaking changes. For the other APIs in that file, it appears that they are completely removed from the bindings file and moved into the compat file. Is using `#ifdef XAMCORE_4` preferable to using the compat file? Or should I be using both - fwict, these are two different approaches to achieve the same result.

mandel-macaque: You are correct, is a #ifdef but the code would compile since we are just defining the new delegate to be called AVAudioSourceNodeRenderHandler2 now, but not in the future, so we have:

* AVAudioSourceNodeRenderHandler2 - Now 
* AVAudioSourceNodeRenderHandler - Once XAMCORE_4 is defined.

The description of the compat files is correct, but in this case we are not create a method for compatibility reasons, but because it was wrong, so I personally would leave the handler definition were it is, with the fixed sample code I gave. In this case, code locality is important, we keep the delegate signature with the API definition.

tldr; We put inside a #ifdef XAMCORE_4 all those API fixes that we would like to do, but we can since it will break the API. One time, in the future, we will be allow to do it, and we will have nicer APIs. I've heard we had several, but I've never seen one, but XAMCORE_2 etc.. exist.

mandel-macaque: I forgot to mention, we now have to keep both versions, AVAudioSourceNodeRenderHandler and AVAudioSourceNodeRenderHandler2

src/AVFoundation/AVCompat.cs
```diff
@@ -169,6 +169,24 @@ public AVAudioUnitComponentManager ()
...
 	}
 
+	public delegate int AVAudioSourceNodeRenderHandler (bool isSilence, AudioToolbox.AudioTimeStamp timestamp, uint frameCount, ref AudioToolbox.AudioBuffers outputData);
+
+	partial class AVAudioSourceNode {
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioSourceNodeRenderHandler2)' instead.")]
+		public AVAudioSourceNode (AVAudioSourceNodeRenderHandler renderHandler)
+			: base (NSObjectFlag.Empty)
+		{
+			throw new NotImplementedException ();
```
rolfbjarne, line 179: I think we can show a better exception here:

```suggestion
			throw new InvalidOperationException ("Do not use this constructor. Use the 'AVAudioSourceNode (AVAudioSourceNodeRenderHandler2)' constructor instead.");
```

src/AVFoundation/AVCompat.cs
```diff
@@ -169,6 +169,24 @@ public AVAudioUnitComponentManager ()
...
+		}
+
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioFormat, AVAudioSourceNodeRenderHandler2)' instead.")]
+		public AVAudioSourceNode (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler)
+			: base (NSObjectFlag.Empty)
+		{
+			throw new NotImplementedException ();
```
rolfbjarne, line 186: Same

src/avfoundation.cs
rolfbjarne: The idea is to:

1. Add a new delegate (`AVAudioSourceNodeRenderHandler2`) with the correct signature.
2. To keep backwards compatibility, we have to keep the old and broken delegate, and all the methods and constructors that use it. Instead we have to add new methods and constructors using the new delegate.
3. The general rule is to move binding code to the AVCompat.cs file when the generated code isn't correct anymore. In your particular case, you want to throw an exception in the old constructors - and that's why those now are in the AVCompat.cs file. In the specific case for the broken delegate itself, it doesn't matter (because there's no generated code you'd want to replace with something else).
4. We also want to have the good delegate name in XAMCORE_4_0 (`AVAudioSourceNodeRenderHandler`), which makes the ifdefs a bit uglier:

```csharp
#if XAMCORE_4_0
delegate int AVAudioSourceNodeRenderHandler (<good signature>);
#else
delegate int AVAudioSourceNodeRenderHandler (<bad signature>);
delegate int AVAudioSourceNodeRenderHandler2 (<good signature>);
#endif
```

rolfbjarne (changes requested):

spouliot (changes requested):
Also latest API diff shows breaking changes on watchOS, likely some defines are not correct (e.g. where you added your code).
Please run it locally before your next commit.

src/AVFoundation/AVCompat.cs
```diff
@@ -10,6 +8,26 @@
 using ObjCRuntime;
 
 namespace AVFoundation {
+#if !XAMCORE_4_0
+	public delegate int AVAudioSourceNodeRenderHandler (bool isSilence, AudioToolbox.AudioTimeStamp timestamp, uint frameCount, ref AudioToolbox.AudioBuffers outputData);
```
whitneyschmidt, line 12: Is there a better way to make this delegate visible to the AVAudioSourceNode methods below? I tried `using AVFoundation.AVSourceNodeRenderHandler` but that failed compiling 😕

src/avfoundation.cs
whitneyschmidt: @rolfbjarne @mandel-macaque Thank you! I think I have made all of the changes you suggested, and all of ^ makes sense.

One more question, `In the specific case for the broken delegate itself, it doesn't matter (because there's no generated code you'd want to replace with something else).`

It seems like the delegate needs to be added to AVCompat.cs. The old delegate cannot be found by the exception throwing methods, even though it is still present in AVFoundation.cs. Why isn't the delegate visible to AVCompat.cs?

src/avfoundation.cs
```diff
@@ -13512,20 +13512,31 @@ interface AVCompositionTrackFormatDescriptionReplacement : NSSecureCoding {
...
 		IntPtr Constructor (AVAudioSourceNodeRenderHandler renderHandler);
+#else
+		IntPtr Constructor (AVAudioSourceNodeRenderHandler2 renderHandler);
+#endif
 
 		[Export ("initWithFormat:renderBlock:")]
 		[DesignatedInitializer]
-		IntPtr Constructor (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler);
+#if XAMCORE_4_0
+                IntPtr Constructor (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler);
```
rolfbjarne, line 13536: Indentation is wrong here.

src/avfoundation.cs
rolfbjarne: > It seems like the delegate needs to be added to AVCompat.cs. The old delegate cannot be found by the exception throwing methods, even though it is still present in AVFoundation.cs. Why isn't the delegate visible to AVCompat.cs?

I'm not sure, but it's hard to argue with the compiler... the current code compiles and the api diff look correct, so I wouldn't spend too much time on this unless your curious.

rolfbjarne (approved):

src/AVFoundation/AVCompat.cs
```diff
@@ -10,6 +8,26 @@
...
+
+	partial class AVAudioSourceNode {
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioSourceNodeRenderHandler2)' instead.")]
+		public AVAudioSourceNode (AVAudioSourceNodeRenderHandler renderHandler)
+			: base (NSObjectFlag.Empty)
```
spouliot, line 17: why did you add `: base (NSObjectFlag.Empty)` here ?

src/AVFoundation/AVCompat.cs
```diff
@@ -10,6 +8,26 @@
...
+		{
+			throw new InvalidOperationException ("Do not use this constructor. Use the 'AVAudioSourceNode (AVAudioSourceNodeRenderHandler2)' constructor instead.");
+		}
+
+		[Obsolete("Use 'AVAudioSourceNode (AVAudioFormat, AVAudioSourceNodeRenderHandler2)' instead.")]
+		public AVAudioSourceNode (AVAudioFormat format, AVAudioSourceNodeRenderHandler renderHandler)
+			: base (NSObjectFlag.Empty)
```
spouliot, line 24: same

src/AVFoundation/AVCompat.cs
spouliot: this allocates the instance (call `alloc` selector) for something that will throw
it seems you're just running additional code (and allocations) for nothing since you'll throwing right afterward

spouliot (changes requested):

dalexsoto (approved):
After Sebastien's comments are addressed 👍

src/avfoundation.cs
mandel-macaque: If it compiles, lets be happy with it.

mandel-macaque (approved):

src/AVFoundation/AVCompat.cs
whitneyschmidt: Without `: base (NSObjectFlag.Empty)` the compilation fails with this error:

```
➜  src git:(avaudiosourcenode_fix) ✗ make
GEN      [ios] generated_sources
CSC      [ios/64 bit] Xamarin.iOS.dll
AVFoundation/AVCompat.cs(16,10): error CS1729: 'AVAudioNode' does not contain a constructor that takes 0 arguments
make: *** [build/ios/native-64/Xamarin.iOS.dll] Error 1
```

I looked at the generated code to try and figure out what the correct full (non-binding) constructor signature should look like, which is where the `NSObject...` came from. There are also several other constructors that throw exceptions in `AVCompat.cs` that call their base class constructors (like AVCaptureDataOutputSynchronizer) so it seemed like the way to go!

Is there a hack (or non-hack!) to get around the compilation error? 😄

spouliot: That's because the base type has a `[DisableDefaultCtor]` :(

You should be able to workaround it by adding
```csharp
partial class AVAudioNode {
    internal AVAudioNode () {}
}
```
inside the `#if !XAMCORE_4_0` of `AVCompat.cs`

spouliot (approved):

src/AVFoundation/AVCompat.cs
spouliot: OTOH it's not really a big deal, if you're calling that code you're already in a code path leading to an exception (so perf/memory does not matter much). I was surprised by the addition but, now, I recall it was `Constructor` before which is why it build fine without it earlier :)

whitneyschmidt: Ah, got it! The empty constructor does not exist because of `[DisableDefaultCtor]` so we have to call one of the other ones 💡 

Update to manually add an empty constructor to `AVAudioNode` inside of `AVCompat.cs` :)

whitneyschmidt:
> Build failure
> ✅ [Provisioning succeeded](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//console)
> ✅ [Build succeeded](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//console)
> ✅ [API Diff (from stable)](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//API_20diff_20_28from_20stable_29)
> ℹ️ [API Diff (from PR only)](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//API_20diff_20_28PR_20only_29) (please review changes)
> ℹ️ [Generator Diff](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//Generator_20Diff) (please review changes)
> 🔥 [Test run failed](https://jenkins.mono-project.com/job/xamarin-macios-pr-builder/15146//Test_20Report) 🔥
> 
> # Test results
> 1 tests failed, 87 tests passed.
> ## Failed tests
> * monotouch-test/iOS Unified 32-bits - simulator/Debug: Failed

```
	[FAIL] AddEntries :   ICE 1
  Expected: <System.InvalidCastException>
  But was:  null
 :   at MonoTouchFixtures.Foundation.NSMutableDictionary2Test.AddEntries () [0x001a0] in /Users/builder/jenkins/workspace/xamarin-macios-pr-builder/tests/monotouch-test/Foundation/NSMutableDictionary2Test.cs:691 
```

^ Appears to be unrelated. @dalexsoto can you confirm that it is safe to merge?

spouliot:
@whitneyschmidt it's a known one. Search for ICE in maccore issues and you'll find it

whitneyschmidt:
@monojenkins backport d16-8

whitneyschmidt:
@monojenkins backport xcode12

PR #9571, opened 2020-09-02
[dotnet] Define and implement a 'framework' or 'sdk' assembly as an assembly that comes from the .NET BCL NuGet.

rolfbjarne:
The Assembly.IsFrameworkAssembly property is used in two places:

* In Driver.IsBoundAssembly to return early when determining if an assembly has any NSObject subclasses: https://github.com/xamarin/xamarin-macios/blob/c1c5b9aac67461122a3b66334876ed9aeb4cd63c/tools/mtouch/mtouch.cs#L1155-L1168
* In Assembly.ExtractNativeLinkInfo to return early when looking for assemblies with LinkWith attributes: https://github.com/xamarin/xamarin-macios/blob/c1c5b9aac67461122a3b66334876ed9aeb4cd63c/tools/common/Assembly.cs#L150-L154

In both cases this definition of framework assembly works today and seems likely to work in the future as well.

I also went through and looked at all the usages of Profile.IsSdkAssembly, and it's used to:

* Decide which assemblies are selected for "link sdk"
* Decide which assemblies are considered an 'sdk' assembly for creating a user framework of all the sdk assemblies
* Bail out early when deciding whether:
    * An assembly references the product assembly (Xamarin.iOS.dll, etc.)
    * An assembly can contain references to UIWebView
    * An assembly can contain user resources
    * An assembly is a binding project / has third-party native resources
    * An assembly needs the dynamic registrar
    * An assembly has FieldAttributes whose native fields must be preserved by the native linker

In all cases our .NET definition of 'SDK' seems to work both for now and in the future.

There are also a few usages which does not apply to .NET, so I've ignored them:

* When looking for a few BCL APIs that must be preserved (MobileApplyPreserveAttribute.cs): this is to be done in the upstream .NET linker now, so it doesn't apply to our own code
* When linking away parameter names (MonoTouchMarkStep.cs): this is to be done in the upstream .NET linker now, so it doesn't apply to our own code

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -125,7 +125,7 @@
 		<_AdditionalTaskAssemblyDirectory>$(_XamarinSdkRootDirectory)tools/dotnet-linker/</_AdditionalTaskAssemblyDirectory>
 		<_AdditionalTaskAssembly>$(_AdditionalTaskAssemblyDirectory)dotnet-linker.dll</_AdditionalTaskAssembly>
 	</PropertyGroup>
-	<Target Name="_ComputeLinkerArguments" DependsOnTargets="_ComputeLinkMode;_ComputeFrameworkVariables;ComputeResolvedFilesToPublishList">
+	<Target Name="_ComputeLinkerArguments" DependsOnTargets="_ComputeLinkMode;_ComputeFrameworkVariables;_ComputeFrameworkAssemblies;ComputeResolvedFilesToPublishList">
 		<!-- Validate the linker mode -->
 		<Error Text="Invalid link mode: '$(_LinkMode)'. Valid link modes are: 'None', 'SdkOnly' and 'Full'" Condition="'$(_LinkMode)' != 'None' And '$(_LinkMode)' != 'SdkOnly' And '$(_LinkMode)' != 'Full'" />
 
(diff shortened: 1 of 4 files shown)


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -144,6 +144,7 @@
 				CacheDirectory=$(_LinkerCacheDirectory)
 				Debug=$(_BundlerDebug)
 				DeploymentTarget=$(_MinimumOSVersion)
+				@(_XamarinFrameworkAssemblies -> 'FrameworkAssembly=%(Filename)')
```
dalexsoto, line 147: > @(_XamarinFrameworkAssemblies -> 'FrameworkAssembly=%(Filename)')

I think it is the first time I see this `->` what does it do? that did not google very well. I'll give it just a guess, does it give you a chance to shape the data your way when container `@()` is resolved? that is my guess

rolfbjarne: It's called an MSBuild Item Transform: https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-transforms?view=vs-2019

And yes, it's the MSBuild version of `Array.ConvertAll (_XamarinFrameworkAssemblies, (v) => $"FrameworkAssembly={v}")`

dalexsoto: Thanks 👍 I've only seen it to call functions like IndexOf and friends but not in this context.

marek-safar:
I don't think this is a step in the right direction. In .NET Core context the `_LinkMode` does not make much sense because there is already a mechanism how to control what kind of linking will be done for each assembly which can be further customized by developers or library authors. I don't think you need to be building a different mechanism to achieve that.

mandel-macaque (approved):

rolfbjarne:
@marek-safar Whatever we do both Xamarin.iOS and Xamarin.Android should do the same, so I'll start a conversation about that.

In any case not much in this PR is related to the linker mode, and many other things (unrelated to this PR) would have to change if we were to implement the linker mode differently, so there's no reason to hold up this PR.

PR #9585, opened 2020-09-05
[NetworkExtensions] Update framework to Xcode12 beta 6.

mandel-macaque:


src/networkextension.cs
```diff
@@ -2104,4 +2134,144 @@ interface NETransparentProxyNetworkSettings {
...
+		[NullAllowed]
+		INEAppPushDelegate Delegate { get; set; }
+
+		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
+		NSObject WeakDelegate { get; set; }
+
+		[Async]
+		[Static]
+		[Export ("loadAllFromPreferencesWithCompletionHandler:")]
+		void LoadAllFromPreferences (Action<NSArray<NEAppPushManager>, NSError> completionHandler);
```
dalexsoto, line 2161:
```suggestion
		void LoadAllFromPreferences (Action<NEAppPushManager [], NSError> completionHandler);
```

dalexsoto (approved):
Just a small one but everything else 👍

src/networkextension.cs
```diff
@@ -2104,4 +2134,144 @@ interface NETransparentProxyNetworkSettings {
 		[NullAllowed, Export ("excludedNetworkRules", ArgumentSemantic.Copy)]
 		NENetworkRule[] ExcludedNetworkRules { get; set; }
 	}
+
+	[NoWatch, NoTV, NoMac, iOS (14,0)]
+	[BaseType (typeof (NSObject))]
+	interface NEAppPushManager {
+		[Export ("matchSSIDs", ArgumentSemantic.Copy)]
+		string[] MatchSsIds { get; set; }
```
rolfbjarne, line 2142: I believe SSID is one big acronym, so:

```suggestion
		string[] MatchSsids { get; set; }
```

It's also what we did here: https://github.com/xamarin/xamarin-macios/blob/98a4c2bd609df1718f536b70004dd2c5dea3cf4c/src/networkextension.cs#L1990

rolfbjarne (changes requested):

src/networkextension.cs
```diff
@@ -2104,4 +2134,144 @@ interface NETransparentProxyNetworkSettings {
...
+	[NoWatch, NoTV, Mac (11,0), iOS (14,0)]
+	[BaseType (typeof (NEDnsSettings), Name = "NEDNSOverHTTPSSettings")]
+	interface NEDnsOverHttpsSettings {
+		[NullAllowed]
+		[Export ("serverURL", ArgumentSemantic.Copy)]
+		NSUrl ServerUrl { get; set; }
+
+		[Notification]
+		[Field ("NEDNSSettingsConfigurationDidChangeNotification")]
+		NSString ConfigurationDidChangeNotification { get; }
```
spouliot, line 2218: from its name it looks like it should be on `NEDnsSettings` and not on `NEDnsOverHttpsSettings`

spouliot (changes requested):

src/networkextension.cs
```diff
@@ -327,6 +327,11 @@ interface NEDnsSettings : NSSecureCoding, NSCopying
 		[NoWatch, NoTV, Mac (11, 0), iOS (14, 0)]
 		[Export ("dnsProtocol")]
 		NEDnsProtocol DnsProtocol { get; }
+
+		[Mac (11,0), iOS (14,0), NoTV]
```
spouliot, line 331: it should either be `[NoWatch]` (like above) or `[Watch (7,0)]`
unless there's defines (that I don't see in the diff) the current attributes makes it available on watchOS 2.0

spouliot (changes requested):

src/networkextension.cs
mandel-macaque: NetworkExtension is not present on the watch

spouliot (approved):

rolfbjarne (approved):

PR #9594, opened 2020-09-07
[dotnet] Make the linker always process debug symbols, and remove debugging support if we're building a release build.

rolfbjarne:
Unfortunately due to when things happen in the .NET build logic, we need to
define the DebuggerSupport property (which determines whether the app should
include debugging support or not) before importing the .NET build files. Since
we want to use the _BundlerDebug property (a.k.a. MtouchDebug/MmpDebug) to
determine if the app should include debugging support, we must figure out the
value of the _BundlerDebug property before we can define the DebuggerSupport
property. This turned out complicated, because we're currently defining
_BundlerDebug in our old-style MSBuild logic, which is imported after we
import the .NET build logic.

The end result is that we can either shuffle around a lot of MSBuild code, or
copy a few lines to set the _BundlerDebug property. Neither option makes we
very happy, but copying a few lines of code seemed the better option, so
that's what I did.

Fixes these linkall test failures in Release mode:

    LinkAll.Attributes.AttributeTest
        [FAIL] DebugAssemblyAttributes :   DebuggableAttribute
            Expected: False
            But was:  True
                at LinkAll.Attributes.AttributeTest.DebugAssemblyAttributes()

        [FAIL] DebugConstructorAttributes :   No debug attribute in release mode
            Expected: 0
            But was:  2
                at LinkAll.Attributes.AttributeTest.DebugConstructorAttributes()

        [FAIL] DebugPropertyAttributes :   DebuggerBrowsable
            Expected: False
            But was:  True
                at LinkAll.Attributes.AttributeTest.DebugPropertyAttributes()

        [FAIL] DebugTypeAttributes :   no debug attribute in release mode
            Expected: 0
            But was:  5
                at LinkAll.Attributes.AttributeTest.DebugTypeAttributes()

        [FAIL] DebuggerTypeProxy_24203 :   proxy
            Expected: null
            But was:  <System.Collections.Generic.IDictionaryDebugView`2[K,V]>
                at LinkAll.Attributes.AttributeTest.DebuggerTypeProxy_24203()

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -79,6 +79,20 @@
 		<_PlatformAssemblyName Condition=" '$(_PlatformName)' == 'tvOS' ">Xamarin.TVOS</_PlatformAssemblyName>
 		<_PlatformAssemblyName Condition=" '$(_PlatformName)' == 'watchOS' ">Xamarin.WatchOS</_PlatformAssemblyName>
 		<_PlatformAssemblyName Condition=" '$(_PlatformName)' == 'macOS' ">Xamarin.Mac</_PlatformAssemblyName>
+
+		<!-- We have to set DebuggerSupport before importing Microsoft.NET.Sdk -->
+		<!-- That means we have to set _BundlerDebug before that, because DebuggerSupport depends on it -->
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):

spouliot (approved):
> makes we very happy
Is that the royal _we_ ? ;-)

rolfbjarne:
> makes we very happy
> Is that the royal _we_ ? ;-)

🤴 😄

emaf (approved):

PR #9596, opened 2020-09-08
[tools] Refactor the Optimizations class to have no conditionally compiled code.

rolfbjarne:
Refactor the Optimizations class to have no conditionally compiled code, which makes
it re-usable from our dotnet-linker code.

Also return any errors or warnings instead of showing/throwing them, which makes
the caller able to show them using whatever means is easiest for the caller.

One test needed an update to the list of valid optimizations, because we now have
a per-platform map of valid optimizations, instead of just a iOS/tvOS/watchOS vs
macOS split ('remove-unsupported-il-for-bitcode' is only valid for watchOS, and now
we say so, while we previously said it was a valid optimization for iOS and tvOS
as well, even though we'd warn about it and do nothing if you tried to set it).

### DIFF
--- a/tests/mmptest/src/MMPTest.cs
+++ b/tests/mmptest/src/MMPTest.cs
@@ -612,7 +612,6 @@ namespace Xamarin.MMP.Tests
 		}
 
 		[Test]
-		[TestCase ("inline-runtime-arch")] // This is valid for Xamarin.iOS
 		[TestCase ("foo")]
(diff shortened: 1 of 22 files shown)


spouliot (approved):
❤️

mandel-macaque (approved):

dalexsoto (approved):

Issue #9578, opened 2020-09-03
MPMusicPlayerController NowPlayingItem throws "Can't lookup the Objective-C class when the dynamic registrar has been linked away"


adrianwaeber:
### Steps to Reproduce

1. Open attached Project in VS on a Windows PC
2. Connect to a Mac with attached Device
3. Run on Device (in my Case it was iPhone 7 and XS) in Debug Mode
4. Tab "Refresh"-Button
5. Exception in thrown in Code
6. Label in App shows the Exception Message

### Expected Behavior
Label should show the currently playing Song from your iPhone

### Actual Behavior
Exception is thrown

### Environment
1. On my Windows PC:
```
Microsoft Visual Studio Professional 2019 Preview
Version 16.8.0 Preview 2.0
VisualStudio.16.Preview/16.8.0-pre.2.0+30420.98
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Professional

Visual C++ 2019   00435-60000-00000-AA578
Microsoft Visual C++ 2019

ASP.NET and Web Tools 2019   16.8.259.4173
ASP.NET and Web Tools 2019

ASP.NET Core Razor Language Services   16.1.0.2041205+11c82f645f3693b48ceb5f9d9e1d3b863c2c470c
Provides languages services for ASP.NET Core Razor.

ASP.NET Web Frameworks and Tools 2019   16.8.259.4173
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.8.259.4173
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   16.8.259.4173
Azure Functions and Web Jobs Tools

C# Tools   3.8.0-2.20418.7+0276b9b777dd8779523b253ed30899bad4781fcc
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.6 (master@34d6af2)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Microsoft Azure Tools   2.9
Microsoft Azure Tools for Microsoft Visual Studio 2019 - v2.9.30701.1

Microsoft Continuous Delivery Tools for Visual Studio   0.4
[...]

PR #9604, opened 2020-09-09
[mediaplayer] Fix MPMusicPlayerController NowPlayingItem on recent iOS versions

spouliot:
Rolf nailed the issue in https://github.com/xamarin/xamarin-macios/issues/9578#issuecomment-688409802
> The problem is that iOS returns an instance of a private type (_MPMusicPlayerMediaItemProxy) which is an NSProxy subclass, and currently we don't support NSProxy.

https://github.com/rolfbjarne/xamarin-macios/commit/873a1e1 was on the
right track but it turns out `[ForcedType]` on properties don't need, nor
work (same generated code), with `return:`.

Inside `DynamicRegistrar.cs` the method
```csharp
public Type Lookup (IntPtr @class, bool throw_on_error)
```
did not respect (was unused) the `throw_on_error`. That made it
impossible to force the type to the pointer we got.

In `Runtime.cs` the method `LookupINativeObjectImplementation` must also
be able to work without an exception (from the `Lookup`) at least when we
want to force the type.

### DIFF
--- a/src/ObjCRuntime/DynamicRegistrar.cs
+++ b/src/ObjCRuntime/DynamicRegistrar.cs
@@ -934,7 +934,9 @@ namespace Registrar {
 					UnlockRegistrar ();
 			}
 
-			throw ErrorHelper.CreateError (4143, "The ObjectiveC class '{0}' could not be registered, it does not seem to derive from any known ObjectiveC class (including NSObject).", Marshal.PtrToStringAuto (Class.class_getName (original_class)));
+			if (throw_on_error)
+				throw ErrorHelper.CreateError (4143, "The ObjectiveC class '{0}' could not be registered, it does not seem to derive from any known ObjectiveC class (including NSObject).", Marshal.PtrToStringAuto (Class.class_getName (original_class)));
+			return null;
 		}
 
 		bool RegisterMethod (ObjCMethod method)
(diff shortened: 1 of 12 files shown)


rolfbjarne:
I wrote a test case (https://github.com/rolfbjarne/xamarin-macios/commit/e171e1e7e75dc0691a71aa71ef92f15e63cccc6b), should I push it into this PR?

mandel-macaque (approved):

mandel-macaque:
@rolfbjarne I think you should push the test case here, that way we have the fix and a test confirming it.

dalexsoto (approved):
whoa nice catch, we should backport this to xcode12 too

whitneyschmidt (approved):

spouliot:
@rolfbjarne yes, push and (if successful and no other changes are required) merge on bots are happy

rolfbjarne:
> @rolfbjarne yes, push and (if successful and no other changes are required) merge on bots are happy

The new test turned out to affect other tests in ways that were hard to resolve, so I reverted it.

rolfbjarne:
> > @rolfbjarne yes, push and (if successful and no other changes are required) merge on bots are happy
> 
> The new test turned out to affect other tests in ways that were hard to resolve, so I reverted it.

I filed #9617 to keep track of the test case, maybe we can update it and get it in later.

spouliot:

PR #9610, opened 2020-09-09
[Foundation] Fix NSUrl's implicit Uri operators. Fixes #9607.

### LINKED ISSUE #9607: InvalidOperationException while casting relative Uri instance to NSUrl
```csharp
NSUrl u = new Uri("relativeUrl", UriKind.Relative);
```

Code above produces InvalidOperationException caused by accessing **PathAndQuery** property which is valid only for absolute uri.

```
System.InvalidOperationException: This operation is not supported for a relative URI.
  at System.Uri.get_PathAndQuery () [0x00008] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/System/net/System/URI.cs:886 
  at Foundation.NSUrl.op_Implicit (System.Uri uri) [0x00014] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.20.2.2/src/Xamarin.iOS/Foundation/NSUrl.cs:61 
```

https://github.com/xamarin/xamarin-macios/blob/7a79845d86b88f036736d01aa2f58dd760ad3e84/src/Foundation/NSUrl.cs#L56-L62

rolfbjarne:
Calling Uri.PathAndQuery is not allowed on a relative Uri, which made the
previous Uri -> NSUrl implicit operator always throw if given a relative
NSUrl.

So I fixed that, added several tests, and found another issue (it turns out
that 'url.RelativePath == url.Path' is not a reliable way to detect absolute
urls, because it's true for relative urls as well) and fixed that too.

Fixes https://github.com/xamarin/xamarin-macios/issues/9607.

### DIFF
--- a/src/Foundation/NSUrl.cs
+++ b/src/Foundation/NSUrl.cs
@@ -47,10 +47,10 @@ namespace Foundation {
 		// Converts from an NSURL to a System.Uri
 		public static implicit operator Uri (NSUrl url)
 		{
-			if (url.RelativePath == url.Path)
-				return new Uri (url.AbsoluteString, UriKind.Absolute);
+			if (Uri.TryCreate (url.AbsoluteString, UriKind.Absolute, out var uri))
+				return uri;
 			else
-				return new Uri (url.RelativePath, UriKind.Relative);
+				return new Uri (url.AbsoluteString, UriKind.Relative);
 		}
 
 		public static implicit operator NSUrl (Uri uri)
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):
Nice catch, let's backport this one to xcode12

mandel-macaque (approved):

whitneyschmidt (approved):

spouliot (approved):

rolfbjarne:
@monojenkins backport xcode12

PR #9620, opened 2020-09-10
[SBApplication][Bug] SBApplication.FromBundleIdentifier<T> should return null when bundle ID is unknown

whitneyschmidt:
Fix for https://github.com/xamarin/xamarin-macios/issues/9011

- [ ] Backport to d16-8
- [ ] Backport to xcode12

Any ideas for adding tests? Right now, tests/scriptingbridge contains one test, but it seems to be broken:

```
➜  scriptingbridge git:(SBFix) ✗ make
/Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/_mac-build/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/bin/bmac --out=build/compat/finder.dll --outdir=build/compat/ -r:/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/4.5/System.Drawing.dll -baselib:../../src/build/mac/compat/XamMac.dll finder.cs
error BI0087: bgen-classic: Xamarin.Mac Classic binding projects are not supported anymore. Please upgrade the binding project to a Xamarin.Mac Unified binding project.
make: *** [build/compat/finder.dll] Error 1
```

### DIFF
--- a/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
+++ b/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
@@ -35,13 +35,13 @@ namespace Xamarin.Mac.Tests {
 #if !XAMCORE_4_0
 			using (var app1 = SBApplication.FromBundleIdentifier (knownBundle))
 			using (var app2 = SBApplication.FromBundleIdentifier<MySBApp> (knownBundle))
-			using (var app4 = SBApplication.FromBundleIdentifier<MySBApp> (unknownBundle))
 			using (var app3 = SBApplication.FromBundleIdentifier (unknownBundle))
+			using (var app4 = SBApplication.FromBundleIdentifier<MySBApp> (unknownBundle))
 #else
-			var (app1 = SBApplication.GetApplicationFromBundleIdentifier (knownBundle))
-			var (app2 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (knownBundle))
-			var (app3 = SBApplication.GetApplicationFromBundleIdentifier (unknownBundle))
-			var (app4 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (unknownBundle))
+			var (app1 = SBApplication.GetApplication (knownBundle))
+			var (app2 = SBApplication.GetApplication<MySbApp> (knownBundle))
+			var (app3 = SBApplication.GetApplication (unknownBundle))
+			var (app4 = SBApplication.GetApplication<MySbApp> (unknownBundle))
 #endif
 			{
 				Assert.IsNotNull (app1, "SBApplication from known bundle is null");
@@ -59,8 +59,8 @@ namespace Xamarin.Mac.Tests {
 			using (var app1 = SBApplication.FromURL (knownUrl))
 			using (var app2 = SBApplication.FromURL<MySBApp> (knownUrl))
 #else
-			using (var app1 = SBApplication.GetApplicationFromUrl (knownUrl))
-			using (var app2 = SBApplication.GetApplicationFromUrl<MySbApp> (knownUrl))
+			using (var app1 = SBApplication.GetApplication (knownUrl))
+			using (var app2 = SBApplication.GetApplication<MySbApp> (knownUrl))
 #endif
 			{
 				Assert.IsNotNull (app1, "SBApplication from known URL is null");
@@ -79,10 +79,10 @@ namespace Xamarin.Mac.Tests {
 			using (var app3 = SBApplication.FromProcessIdentifier (unknownPid))
 			using (var app4 = SBApplication.FromProcessIdentifier<MySBApp> (unknownPid))
 #else
-			using (var app1 = SBApplication.GetApplicationFromProcessIdentifier (knownPid))
-			using (var app2 = SBApplication.GetApplicationFromProcessIdentifier<MySbApp> (knownPid))
-			using (var app3 = SBApplication.GetApplicationFromProcessIdentifier (unknownPid))
-			using (var app4 = SBApplication.GetApplicationFromProcessIdentifier<MySbApp> (unknownPid)
+			using (var app1 = SBApplication.GetApplication (knownPid))
+			using (var app2 = SBApplication.GetApplication<MySbApp> (knownPid))
+			using (var app3 = SBApplication.GetApplication (unknownPid))
+			using (var app4 = SBApplication.GetApplication<MySbApp> (unknownPid)
 #endif
 			{
 				Assert.IsNotNull (app1, "SBApplication from known pid is null");
--- a/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
+++ b/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
@@ -8,105 +8,87 @@ using ScriptingBridge;
 
 namespace Xamarin.Mac.Tests {
 
-	public class MySbApp : SBApplication {
-		public MySbApp () : base (NSObjectFlag.Empty) { }
+	public class MySBApp : SBApplication {
+		public MySBApp () : base (NSObjectFlag.Empty) { }
 
-		public MySbApp (NSCoder coder) : base (coder) { }
+		public MySBApp (NSCoder coder) : base (coder) { }
 
-		public MySbApp (NSUrl url) : base (url) { }
+		public MySBApp (NSUrl url) : base (url) { }
 
-		public MySbApp (int pid) : base (pid) { }
+		public MySBApp (int pid) : base (pid) { }
 
-		public MySbApp (string ident) : base (ident) { }
+		public MySBApp (string ident) : base (ident) { }
 
-		protected MySbApp (NSObjectFlag t) : base (t) { }
+		protected MySBApp (NSObjectFlag t) : base (t) { }
 
-		protected internal MySbApp (IntPtr handle) : base (handle) { }
+		protected internal MySBApp (IntPtr handle) : base (handle) { }
 	}
 
 	[TestFixture]
 	public class SBApplicationTest {
 
-		[TestFixture]
-		public class GetSBApplicationTest {
-
-			[Test]
-			public void TestGetApplicationFromBundleIdentifier ()
-			{
-				const string knownBundle = "com.apple.finder";
-				const string unknownBundle = "com.unknown.bundle";
+		[Test]
+		public void TestGetApplicationFromBundleIdentifier ()
+		{
+			const string knownBundle = "com.apple.finder";
+			const string unknownBundle = "com.unknown.bundle";
 #if !XAMCORE_4_0
-				var app2 = SBApplication.FromBundleIdentifier<MySbApp> (knownBundle);
-				Assert.IsNotNull (app2);
-				var app1 = SBApplication.FromBundleIdentifier (unknownBundle);
-				Assert.IsNotNull (app1);
-				var app4 = SBApplication.FromBundleIdentifier<MySbApp> (unknownBundle);
-
-				var app3 = SBApplication.FromBundleIdentifier (unknownBundle);
-				Assert.IsNull (app3);
-				Assert.IsNull (app4);
+			using (var app1 = SBApplication.FromBundleIdentifier (knownBundle))
+			using (var app2 = SBApplication.FromBundleIdentifier<MySBApp> (knownBundle))
+			using (var app4 = SBApplication.FromBundleIdentifier<MySBApp> (unknownBundle))
+			using (var app3 = SBApplication.FromBundleIdentifier (unknownBundle))
 #else
-				var app1 = SBApplication.GetApplicationFromBundleIdentifier (knownBundle);
-				Assert.IsNotNull (app1);
-				var app2 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (knownBundle);
-				Assert.IsNotNull (app2);
-				var app3 = SBApplication.GetApplicationFromBundleIdentifier (unknownBundle);
-				var app4 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (unknownBundle);
-				Assert.IsNull (app3);
-				Assert.IsNull (app4);
+			var (app1 = SBApplication.GetApplicationFromBundleIdentifier (knownBundle))
+			var (app2 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (knownBundle))
+			var (app3 = SBApplication.GetApplicationFromBundleIdentifier (unknownBundle))
+			var (app4 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (unknownBundle))
 #endif
+			{
+				Assert.IsNotNull (app1, "SBApplication from known bundle is null");
+				Assert.IsNotNull (app2, "MySBApp from known bundle is null");
+				Assert.IsNull (app3, "SBApplication from unknown bundle is non-null");
+				Assert.IsNull (app4, "MySBApp from unknown bundle is non-null");
 			}
+		}
 
-			[Test]
-			public void TestGetApplicationFromUrl ()
-			{
-				NSUrl knownUrl = new NSUrl("http://www.xamarin.com");
-				NSUrl unknownUrl = new NSUrl ("thisisnotaurl");
+		[Test]
+		public void TestGetApplicationFromUrl ()
+		{
+			using (NSUrl knownUrl = new NSUrl ("http://www.xamarin.com"))
 #if !XAMCORE_4_0
-				var app1 = SBApplication.FromURL (knownUrl);
-				Assert.IsNotNull (app1);
-				var app2 = SBApplication.FromURL<MySbApp> (knownUrl);
-				Assert.IsNotNull (app2);
-				var app3 = SBApplication.FromURL (unknownUrl);
-				var app4 = SBApplication.FromURL<MySbApp> (unknownUrl);
-				Assert.IsNull (app3);
-				Assert.IsNull (app4);
+			using (var app1 = SBApplication.FromURL (knownUrl))
+			using (var app2 = SBApplication.FromURL<MySBApp> (knownUrl))
 #else
-				var app1 = SBApplication.GetApplicationFromUrl (knownUrl);
-				Assert.IsNotNull (app1);
-				var app2 = SBApplication.GetApplicationFromUrl<MySbApp> (knownUrl);
-				Assert.IsNotNull (app2);
-				var app3 = SBApplication.GetApplicationFromUrl (unknownUrl);
-				var app4 = SBApplication.GetApplicationFromUrl<MySbApp> (unknownUrl);
-				Assert.IsNull (app3);
-				Assert.IsNull (app4);
+			using (var app1 = SBApplication.GetApplicationFromUrl (knownUrl))
+			using (var app2 = SBApplication.GetApplicationFromUrl<MySbApp> (knownUrl))
 #endif
+			{
+				Assert.IsNotNull (app1, "SBApplication from known URL is null");
+				Assert.IsNotNull (app2, "MySBApp from known URL is null");
 			}
+		}
 
-			[Test]
-			public void TestGetApplicationFromPid ()
-			{
-				int knownPid = 1;
-				int unknownPid = -1; // valid pid is > 0?
+		[Test]
+		public void TestGetApplicationFromPid ()
+		{
+			int knownPid = System.Diagnostics.Process.GetCurrentProcess ().Id;
+			int unknownPid = -1; // valid pid is > 0
 #if !XAMCORE_4_0
-				var app1 = SBApplication.FromProcessIdentifier (knownPid);
-				Assert.IsNotNull (app1);
-				var app2 = SBApplication.FromProcessIdentifier<MySbApp> (knownPid);
-				Assert.IsNotNull (app2);
-				var app3 = SBApplication.FromProcessIdentifier (unknownPid);
-				var app4 = SBApplication.FromProcessIdentifier<MySbApp> (unknownPid);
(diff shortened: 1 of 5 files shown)


spouliot (approved):
I wonder how (and if) `(T)u` and `default(T)` code generation differs...

IIRC ScriptingBridge was not easy to test (hence the external test) except that you're not _really_ testing ScriptingBridge.

Suggestion:
1. refactor line 14, 20 and 26 into a `FromHandle (IntPtr)` method;
2. replace those lines with `return FromHandle (u.GetHandle ());`
3. write a _normal_ (apitest since it's XM-only) unit test for the new `FromHandle` method

dalexsoto (changes requested):
I am not sure I like the current `System.Activator.CreateInstance` code (not your fault @whitneyschmidt) in the 3 methods since it calls the IntPtr `ctor` and we could potentially end up with multiple managed instances with the same handle.

I think the following should be changed from

https://github.com/xamarin/xamarin-macios/blob/65237b9af78aeb58095555f6a6e29b18ccb46629/src/scriptingbridge.cs#L115-L125

to 

```csharp
 [Internal]
 [Static] 
 [Export ("applicationWithBundleIdentifier:")] 
 IntPtr _FromBundleIdentifier (string ident ); 
  
 [Internal] 
 [Static] 
 [Export ("applicationWithURL:")] 
 IntPtr _FromURL (NSUrl url ); 
  
 [Internal] 
 [Static] 
 [Export ("applicationWithProcessIdentifier:")] 
 IntPtr _FromProcessIdentifier (int /* pid_t = int */ pid ); 
```

and then add to `src/ScriptingBridge/SBApplication.cs`

```csharp
public static SBApplication FromBundleIdentifier (string ident)  => Runtime.GetNSObject<SBApplication> (_FromBundleIdentifier (ident));

public static T FromBundleIdentifier<T> (string ident) where T : SBApplication => Runtime.GetNSObject<T> (_FromBundleIdentifier (ident));

public static SBApplication FromURL (NSUrl url)  => Runtime.GetNSObject<SBApplication> (_FromURL (url));

public static T FromURL<T> (NSUrl url) where T : SBApplication => Runtime.GetNSObject<T> (_FromURL (url));

public static SBApplication FromProcessIdentifier (int pid)  => Runtime.GetNSObject<SBApplication> (_FromProcessIdentifier (pid));

public static T FromProcessIdentifier<T> (int pid) where T : SBApplication => Runtime.GetNSObject<T> (_FromProcessIdentifier (pid));
```

For me this would be the way to go about this and as a bonus maybe XAMCORE_4_0'ify those `FromURL` => `FromUrl`. Unless @rolfbjarne or @spouliot think otherwise of course because is kinda late here and I might be overthinking this 😛

rolfbjarne:
I like Alex's suggestion 👍 Except that if we're doing a bigger refactoring, we might as well fix the names too from `From...` to something that has a verb (`Create`, `Open`, `Get`, etc - `Get` seems the closest to describing what these methods do, but there's probably better names)

dalexsoto:
Yup agreed on `From` => `Create` Too

mandel-macaque (approved):
Taking into account @dalexsoto comments.

tj-devel709:
> I am not sure I like the current `System.Activator.CreateInstance` code (not your fault @whitneyschmidt) in the 3 methods since it calls the IntPtr `ctor` and we could potentially end up with multiple managed instances with the same handle.
@dalexsoto so why would this be creating multiple instances with the same handle? I don't follow this change very much.

dalexsoto:
> I am not sure I like the current `System.Activator.CreateInstance` code (not your fault @whitneyschmidt) in the 3 methods since it calls the IntPtr `ctor` and we could potentially end up with multiple managed instances with the same handle.
> @dalexsoto so why would this be creating multiple instances with the same handle? I don't follow this change very much.

Hey @tj-devel709, so with the code here

```csharp
(T)System.Activator.CreateInstance (typeof(T), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance, null, new object [] { u.Handle }, null);
```

We are using the reflection API to create an instance of the specified type (`T : SBApplication`) using the constructor that best matches the specified parameters (`.ctor (IntPtr)` in this case) so TL;DR: we are invoking the following constructor with the value of `u.Handle`.

```csharp
public T (IntPtr handle) { ... }
```

The above `.ctor` sets the `Handle` property of the instantiated object, the Handle property of every NSObject derived class holds an IntPtr that is a handle to the underlying unmanaged representation for this object. Usually we only want 1 managed object that its Handle points to 1 unmanaged object (a 1:1 relationship) this is because the .NET Runtime needs to manage the lifecycle of the managed object and if two different managed objects point to the same native object you can run into some funny scenarios like double disposing the same unmanaged handle, objects that will never be released from memory, etc. If you call several times any of the C# generic APIs listed above you could in theory be creating two or more managed objects that could potentially point to the same unmanaged object.
 
Instead of calling the IntPtr constructor I suggest using `Runtime.GetNSObject<T> (IntPtr)` which will check if we have any managed object that its Handle property matches the given pointer, if so it will return that object, if it can't find one it will create a new managed instance with the given pointer and return that instead.

I hope I managed to explain myself here heh :P if not keep posting your questions and I would also recommend having a look [here](https://docs.microsoft.com/en-us/xamarin/ios/internals/)

tj-devel709:
@dalexsoto Thanks for the reply! Definitely cleared some stuff up for me and I will check out that link!

tj-devel709 (approved):

whitneyschmidt:
After making the changes that @dalexsoto suggested, I discovered that the unit testing code fails:
```

namespace Xamarin.Mac.Tests {

	public class MySbApp : SBApplication {
		public MySbApp () : base (NSObjectFlag.Empty) { }

		public MySbApp (NSCoder coder) : base (coder) { }

		public MySbApp (NSUrl url) : base (url) { }

		public MySbApp (int pid) : base (pid) { }

		public MySbApp (string ident) : base (ident) { }

		protected MySbApp (NSObjectFlag t) : base (t) { }

		protected internal MySbApp (IntPtr handle) : base (handle) { }
	}

	[TestFixture]
	public class SBApplicationTest {

		[TestFixture]
		public class GetSBApplicationTest {

			[Test]
			public void TestGetApplicationFromBundleIdentifier ()
			{
				const string knownBundle = "com.apple.finder";
				const string unknownBundle = "com.unknown.bundle";
#if !XAMCORE_4_0
				var app2 = SBApplication.FromBundleIdentifier<MySbApp> (knownBundle);
#else
...
#endif
			} 
```

Error:
```
Xamarin.Mac.Tests.SBApplicationTest+GetSBApplicationTest
	[FAIL] TestGetApplicationFromBundleIdentifier : System.InvalidCastException : Specified cast is not valid.
		  at (wrapper castclass) System.Object.__castclass_with_cache(object,intptr,intptr)
		  at ObjCRuntime.Runtime.ConstructNSObject[T] (System.IntPtr ptr, System.Type type, ObjCRuntime.Runtime+MissingCtorResolution missingCtorResolution) [0x00045] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1121 
		  at ObjCRuntime.Runtime.GetNSObject[T] (System.IntPtr ptr) [0x00103] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1261 
		  at ScriptingBridge.SBApplication.FromBundleIdentifier[T] (System.String ident) [0x00000] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ScriptingBridge/SBApplication.cs:25 
		  at Xamarin.Mac.Tests.SBApplicationTest+GetSBApplicationTest.TestGetApplicationFromBundleIdentifier () [0x00001] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs:39 
		  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395 
	[FAIL] TestGetApplicationFromPid : System.InvalidCastException : Specified cast is not valid.
		  at (wrapper castclass) System.Object.__castclass_with_cache(object,intptr,intptr)
		  at ObjCRuntime.Runtime.ConstructNSObject[T] (System.IntPtr ptr, System.Type type, ObjCRuntime.Runtime+MissingCtorResolution missingCtorResolution) [0x00045] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1121 
		  at ObjCRuntime.Runtime.GetNSObject[T] (System.IntPtr ptr) [0x00103] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1261 
		  at ScriptingBridge.SBApplication.FromProcessIdentifier[T] (System.Int32 pid) [0x00000] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ScriptingBridge/SBApplication.cs:33 
		  at Xamarin.Mac.Tests.SBApplicationTest+GetSBApplicationTest.TestGetApplicationFromPid () [0x00013] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs:94 
		  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395 
	[FAIL] TestGetApplicationFromUrl : System.InvalidCastException : Specified cast is not valid.
		  at (wrapper castclass) System.Object.__castclass_with_cache(object,intptr,intptr)
		  at ObjCRuntime.Runtime.ConstructNSObject[T] (System.IntPtr ptr, System.Type type, ObjCRuntime.Runtime+MissingCtorResolution missingCtorResolution) [0x00045] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1121 
		  at ObjCRuntime.Runtime.GetNSObject[T] (System.IntPtr ptr) [0x00103] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ObjCRuntime/Runtime.cs:1261 
		  at ScriptingBridge.SBApplication.FromURL[T] (Foundation.NSUrl url) [0x00000] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/src/ScriptingBridge/SBApplication.cs:29 
		  at Xamarin.Mac.Tests.SBApplicationTest+GetSBApplicationTest.TestGetApplicationFromUrl () [0x00025] in /Users/whitneyschmidt/Documents/xamarin-macios/xamarin-macios/tests/apitest/src/ScriptingBridge/SBApplicationTest.cs:68 
		  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395 
Xamarin.Mac.Tests.SBApplicationTest+GetSBApplicationTest : 52 ms
```

rolfbjarne:
Try using `GetINativeObject<T>` like this instead of `Runtime.GetNSObject<T>` (in particular the `forced_type: true` part is important):

```csharp
public static T FromBundleIdentifier<T> (string ident) where T : SBApplication => Runtime.GetINativeObject<T> (_FromBundleIdentifier (ident), forced_type: true, owns: false);

**Comment by whitneyschmidt** [comment]
> Try using `GetINativeObject<T>` like this instead of `Runtime.GetNSObject<T>` (in particular the `forced_type: true` part is important):
> 
> ```cs
> public static T FromBundleIdentifier<T> (string ident) where T : SBApplication => Runtime.GetINativeObject<T> (_FromBundleIdentifier (ident), forced_type: true, owns: false);
> ```

I tried out running the tests with these updates, unfortunately the tests fail with the same error 😕

**Comment by rolfbjarne** [comment]
Where are the unit tests? They don't seem to be in the repo already.

**Comment by whitneyschmidt** [comment]
> Where are the unit tests? They don't seem to be in the repo already.



> Where are the unit tests? They don't seem to be in the repo already.

Good question! I just pushed all the changes I've made, including the (broken) tests.

File: tests/apitest/src/ScriptingBridge/SBApplicationTest.cs [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+		public class GetSBApplicationTest {
+
+			[Test]
+			public void TestGetApplicationFromBundleIdentifier ()
+			{
+				const string knownBundle = "com.apple.finder";
+				const string unknownBundle = "com.unknown.bundle";
+#if !XAMCORE_4_0
+				var app2 = SBApplication.FromBundleIdentifier<MySbApp> (knownBundle);
+				Assert.IsNotNull (app2);
```
mandel-macaque, line 40: Missing message for the assert, and we have several asserts in the test.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+				var app1 = SBApplication.FromBundleIdentifier (unknownBundle);
+				Assert.IsNotNull (app1);
```
mandel-macaque, line 42: Missing message for the assert, and we have several asserts in the test.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+				var app4 = SBApplication.FromBundleIdentifier<MySbApp> (unknownBundle);
+
+				var app3 = SBApplication.FromBundleIdentifier (unknownBundle);
+				Assert.IsNull (app3);
```
mandel-macaque, line 46: same

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+				Assert.IsNull (app4);
+#else
+				var app1 = SBApplication.GetApplicationFromBundleIdentifier (knownBundle);
+				Assert.IsNotNull (app1);
+				var app2 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (knownBundle);
+				Assert.IsNotNull (app2);
+				var app3 = SBApplication.GetApplicationFromBundleIdentifier (unknownBundle);
+				var app4 = SBApplication.GetApplicationFromBundleIdentifier<MySbApp> (unknownBundle);
+				Assert.IsNull (app3);
+				Assert.IsNull (app4);
```
mandel-macaque, line 56: Please add messages to all the asserts.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+#endif
+			}
+
+			[Test]
+			public void TestGetApplicationFromUrl ()
+			{
+				NSUrl knownUrl = new NSUrl("http://www.xamarin.com");
+				NSUrl unknownUrl = new NSUrl ("thisisnotaurl");
```
mandel-macaque, line 64: using is missing, leets clean after the test.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+				Assert.IsNull (app4);
+#else
+				var app1 = SBApplication.GetApplicationFromUrl (knownUrl);
+				Assert.IsNotNull (app1);
+				var app2 = SBApplication.GetApplicationFromUrl<MySbApp> (knownUrl);
+				Assert.IsNotNull (app2);
+				var app3 = SBApplication.GetApplicationFromUrl (unknownUrl);
+				var app4 = SBApplication.GetApplicationFromUrl<MySbApp> (unknownUrl);
+				Assert.IsNull (app3);
+				Assert.IsNull (app4);
```
mandel-macaque, line 82: Add messages to all the asserts.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
mandel-macaque: Also, all the apps are native objects, should be disposed.

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,115 @@
...
+#else
+				var app1 = SBApplication.GetApplicationFromProcessIdentifier (knownPid);
+				Assert.IsNotNull (app1);
+				var app2 = SBApplication.GetApplicationFromProcessIdentifier<MySbApp> (knownPid);
+				Assert.IsNotNull (app2);
+				var app3 = SBApplication.GetApplicationFromProcessIdentifier (unknownPid);
+				var app4 = SBApplication.GetApplicationFromProcessIdentifier<MySbApp> (unknownPid);
+				Assert.IsNull (app3);
+				Assert.IsNull (app4);
```
mandel-macaque, line 108: Add missing messages in asserts and release the native objects.

mandel-macaque (changes requested):

rolfbjarne:
> I tried out running the tests with these updates, unfortunately the tests fail with the same error 😕

Did you rebuild the src/ directory?

I tried your branch, and I don't see the problem you're seeing.

The tests actually pass with these changes: https://gist.github.com/rolfbjarne/cb2144d73186894e98adb13c3bb79d6a (the broken url ones crash (because Apple decided that was the right thing to do with broken urls), so I had to disable those).

tests/apitest/src/ScriptingBridge/SBApplicationTest.cs
```diff
@@ -0,0 +1,97 @@
...
+		[Test]
+		public void TestGetApplicationFromBundleIdentifier ()
+		{
+			const string knownBundle = "com.apple.finder";
+			const string unknownBundle = "com.unknown.bundle";
+#if !XAMCORE_4_0
+			using (var app1 = SBApplication.FromBundleIdentifier (knownBundle))
+			using (var app2 = SBApplication.FromBundleIdentifier<MySBApp> (knownBundle))
+			using (var app4 = SBApplication.FromBundleIdentifier<MySBApp> (unknownBundle))
+			using (var app3 = SBApplication.FromBundleIdentifier (unknownBundle))
```
rolfbjarne, line 39: This is quite minor, but 3 comes before 4...

rolfbjarne (changes requested):
This is quite close now! And it's great to have unit tests.

whitneyschmidt:
> This is quite close now! And it's great to have unit tests.

A couple of final questions 😆 

1. Can you explain more about the difference between these two (+ why one of them did not work for this case):

- `public static T FromBundleIdentifier<T> (string ident) where T : SBApplication => Runtime.GetINativeObject<T> (_FromBundleIdentifier (ident), forced_type: true, owns: false);`

- `public static SBApplication FromBundleIdentifier (string ident)  => Runtime.GetNSObject<SBApplication> (_FromBundleIdentifier (ident));
`

2. I removed the xtro .ignore lines for `missing-null-allowed` on the return type for the bindings I made `[Internal]`, and it's also obvious that one of them already successfully returns null. What is the value of having the explicit attribute for null-allowed return type - does the generator us it to do some magic?

rolfbjarne:
> > This is quite close now! And it's great to have unit tests.
> 
> A couple of final questions 😆
> 
> 1. Can you explain more about the difference between these two (+ why one of them did not work for this case):
> 
> * `public static T FromBundleIdentifier<T> (string ident) where T : SBApplication => Runtime.GetINativeObject<T> (_FromBundleIdentifier (ident), forced_type: true, owns: false);`
> * `public static SBApplication FromBundleIdentifier (string ident)  => Runtime.GetNSObject<SBApplication> (_FromBundleIdentifier (ident)); `

`Runtime.GetNSObject<T> (ptr)` will check the native type of `ptr`, and create the corresponding managed instance. If the created instance's type doesn't match the expected return type `T`, you get the InvalidCastException.

`Runtime.GetINativeObject<T> (ptr, forced_type: true)` will trust that the caller knows what's going on, and create a managed instance `T` for the ptr, without checking the native type. If the managed type is not compatible with the native type, you'll get all kinds of broken behavior, but sometimes this is what needs to be done.

> 1. I removed the xtro .ignore lines for `missing-null-allowed` on the return type for the bindings I made `[Internal]`, and it's also obvious that one of them already successfully returns null. What is the value of having the explicit attribute for null-allowed return type - does the generator us it to do some magic?

You had to remove the `missing-null-allowed` attribute on the return type because you changed the return type from from a class (SBApplication) to a value type (IntPtr). Value types can't be null, so xtro doesn't check for the attribute. The generated code is different when you add/remove the NullAllowed attribute, and you can see it in the generator diff.

rolfbjarne (approved):

dalexsoto (approved):

whitneyschmidt:
@monojenkins backport d16-8

dalexsoto:
@monojenkins backport d16-8-xm

mandel-macaque:
This does not need to be backported to xcode12 since we are merging with d16-8

PR #9623, opened 2020-09-11
[Network] Partial bindings of Network for Xcode 12 beta 6.

mandel-macaque:
Bound the NWConnectionGroup and all its methods and enums. Tests cannot
be added until NWGroupDescriptor is bound. All tests for this class and
NWGroupDescriptor will be added in the following PR.


mandel-macaque:
Note to reviewers, I'm using tabs, no idea why github decides to use two spaces...

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+		static nw_connection_group_send_completion_t static_SendCompletion = TrampolineSendCompletion;
+
+		[MonoPInvokeCallback (typeof (nw_connection_group_send_completion_t))]
+		static void TrampolineSendCompletion (IntPtr block, IntPtr error)
+		{
+			var del = BlockLiteral.GetTarget<Action<NWError?>> (block);
+			if (del != null) {
+				var err = error == IntPtr.Zero ? null : new NWError (error, owns: false);
+				del (err);
+				err?.Dispose ();
```
tj-devel709, line 171: To my understanding, a question mark denotes a nullable type, but since you are not declaring 'del' here, why is there a question mark?

tj-devel709 (approved):
Everything looks good to me!

src/Network/NWConnectionGroup.cs
mandel-macaque: In this case is because error can be null, the ? will mean, if not null, call the method, else don't

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+		static extern OS_nw_connection_group nw_connection_group_create (OS_nw_group_descriptor group_descriptor, OS_nw_parameters parameters);
+
+		public NWConnectionGroup (NWGroupDescriptor groupDescriptor, NWParameters parameters)
+		{
+			if (groupDescriptor == null)
+				throw new ArgumentNullException (nameof (groupDescriptor));
+			if (parameters == null)
+				throw new ArgumentNullException (nameof (parameters));
+
+			InitializeHandle (nw_connection_group_create (groupDescriptor.Handle, parameters.Handle));
```
rolfbjarne, line 48: Maybe use `GetCheckedHandle` here?

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+		static extern void nw_connection_group_reply (OS_nw_connection_group group, OS_nw_content_context inbound_message, OS_nw_content_context outbound_message, /* [NullAllowed]  DispatchData */ IntPtr content);
+
+		public void Reply (NWContentContext inboundMessage, NWContentContext outboundMessage, DispatchData content)
+		{
+			if (inboundMessage == null)
+				throw new ArgumentNullException (nameof (inboundMessage));
+			if (outboundMessage == null)
+				throw new ArgumentNullException (nameof (outboundMessage));
+
+			var contentPtr = (content == null) ? IntPtr.Zero : content.Handle;
```
rolfbjarne, line 154: `content.GetHandle ()` is a safe way to get the Handle or IntPtr.Zero.

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+		}
+
+		[BindingImpl (BindingImplOptions.Optimizable)]
+		public void Send (DispatchData? content, NWEndpoint? endpoint, NWContentContext context, Action<NWError?>? handler)
+		{
+			unsafe {
+				if (handler == null) {
+					nw_connection_group_send_message (GetCheckedHandle (),
+						content == null ? IntPtr.Zero : content.Handle,
+						endpoint == null ? IntPtr.Zero : endpoint.Handle,
```
rolfbjarne, line 182:
```suggestion
						content.GetHandle (),
						endpoint.GetHandle (),
```

This can be done in many other places as well.

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+						context.Handle,
```
rolfbjarne, line 183: GetCheckedHandle?

```suggestion
						context.GetCheckedHandle (),
```

This can be done in many other places as well.

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+		delegate void nw_connection_group_receive_handler_t (IntPtr block, IntPtr content, IntPtr context, bool isCompleted);
+		static nw_connection_group_receive_handler_t static_ReceiveHandler = TrampolineReceiveHandler;
+
+		[MonoPInvokeCallback (typeof (nw_connection_group_receive_handler_t))]
+		static void TrampolineReceiveHandler (IntPtr block, IntPtr content, IntPtr context, bool isCompleted)
+		{
+			var del = BlockLiteral.GetTarget<NWConnectionGroupReceiveDelegate> (block);
+			if (del != null) {
+				var nsContent = new DispatchData (content, owns: false);
+				var nsContext = new NWContentContext (context, owns: false);
```
rolfbjarne, line 214: This way you wouldn't need to manually Dispose later:

```suggestion
				using var nsContent = new DispatchData (content, owns: false);
				using var nsContext = new NWContentContext (context, owns: false);
```

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+				del (nsContent, nsContext, isCompleted);
+				nsContext.Dispose ();
+				nsContext.Dispose ();
```
rolfbjarne, line 217: And that would also fix a bug, because you're Disposing the same instance twice here:

```suggestion
```

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+
+		delegate void nw_connection_group_state_changed_handler_t (IntPtr block, NWConnectionGroupState state , IntPtr error);
+		static nw_connection_group_state_changed_handler_t static_StateChangedHandler = TrampolineStateChangedHandler;
+
+		[MonoPInvokeCallback (typeof (nw_connection_group_state_changed_handler_t))]
+		static void TrampolineStateChangedHandler (IntPtr block, NWConnectionGroupState state, IntPtr error)
+		{
+			var del = BlockLiteral.GetTarget<NWConnectionGroupStateChangedDelegate> (block);
+			if (del != null) {
+				var nwError = (error == IntPtr.Zero) ? null : new NWError (error, owns: false);
```
rolfbjarne, line 251: Does adding `using` work here?

```suggestion
				using var nwError = (error == IntPtr.Zero) ? null : new NWError (error, owns: false);
```

rolfbjarne (changes requested):

src/Network/NWConnectionGroup.cs
mandel-macaque: Good catch. Stupid autocomplete 🤦🏼‍♂️

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+
+	[TV (14,0), Mac (11,0), iOS (14,0), Watch (7,0)]
+	public delegate void NWConnectionGroupReceiveDelegate (DispatchData content, NWContentContext context, bool isCompleted);
+
+	[TV (14,0), Mac (11,0), iOS (14,0), Watch (7,0)]
+	public delegate void NWConnectionGroupStateChangedDelegate (NWConnectionGroupState state, NWError? error);
+
+	[TV (14,0), Mac (11,0), iOS (14,0), Watch (7,0)]
+	public class NWConnectionGroup : NativeObject {
+		public NWConnectionGroup (IntPtr handle, bool owns) : base (handle, owns) {}
```
spouliot, line 36: `protected` (or `protected internal`)

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+
+		// can return null
+		[DllImport (Constants.NetworkLibrary)]
+		static extern OS_nw_connection nw_connection_group_extract_connection_for_message (OS_nw_connection_group group, OS_nw_content_context context);
+
+		public NWConnection? GetConnection (NWContentContext context)
+		{
+			if (context == null)
+				throw new ArgumentNullException (nameof (context));
+			var ptr = nw_connection_group_extract_connection_for_message (GetCheckedHandle (), context.Handle);
```
spouliot, line 140: Use `GetCheckedHandle` so you don't end up providing a non-null managed instance that is disposed.

IOW when the instance is disposed then `Handle` is `IntPtr.Zero`. You would effectively call the native API with `nil`, which is not allowed.

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,276 @@
...
+
+		public void Reply (NWContentContext inboundMessage, NWContentContext outboundMessage, DispatchData content)
+		{
+			if (inboundMessage == null)
+				throw new ArgumentNullException (nameof (inboundMessage));
+			if (outboundMessage == null)
+				throw new ArgumentNullException (nameof (outboundMessage));
+
+			var contentPtr = (content == null) ? IntPtr.Zero : content.Handle;
+			nw_connection_group_reply (GetCheckedHandle (), inboundMessage.Handle, outboundMessage.Handle, contentPtr);
```
spouliot, line 155: same as earlier, use `GetCheckedHandle`

src/Network/NWConnectionGroup.cs
spouliot: and they would still be disposed if `del` throws :)

src/Network/NWGroupDescriptor.cs
```diff
@@ -0,0 +1,15 @@
...
+using Foundation;
+using CoreFoundation;
+
+#nullable enable
+
+namespace Network {
+
+	[TV (14,0), Mac (11,0), iOS (14,0), Watch (7,0)]
+	public class NWGroupDescriptor : NativeObject {
+		public NWGroupDescriptor (IntPtr handle, bool owns) : base (handle, owns) {}
```
spouliot, line 13: `protected internal` since it seems used outside the type

spouliot (changes requested):
Also needs tests.

Splitting the PR _is_ fine, adding tests later _can_ be fine.

OTOH merging this PR without tests is not since we could release before the tests are ready/merged (and would need extra work to obsolete API).

mandel-macaque:
Do not merge until we have the GM. Certains things have been removed from mac OS X.

tests/monotouch-test/Network/NWConnectionGroupTest.cs
```diff
@@ -0,0 +1,66 @@
...
+using ObjCRuntime;
+using Security;
+
+using NUnit.Framework;
+
+namespace MonoTouchFixtures.Network {
+
+
+	// The following class just has a subset of the methods in the class, the API allows to perform udp 
+	// communications between bojour services, which is very fragile in the CI, a sample
```
rolfbjarne, line 16:
```suggestion
	// communications between bonjour services, which is very fragile in the CI, a sample 
```

src/Network/NWConnectionGroup.cs
```diff
@@ -0,0 +1,271 @@
...
+		static extern void nw_connection_group_reply (OS_nw_connection_group group, OS_nw_content_context inbound_message, OS_nw_content_context outbound_message, /* [NullAllowed]  DispatchData */ IntPtr content);
+
+		public void Reply (NWContentContext inboundMessage, NWContentContext outboundMessage, DispatchData content)
+		{
+			if (inboundMessage == null)
+				throw new ArgumentNullException (nameof (inboundMessage));
+			if (outboundMessage == null)
+				throw new ArgumentNullException (nameof (outboundMessage));
+
+			nw_connection_group_reply (GetCheckedHandle (), inboundMessage.Handle, outboundMessage.Handle, content.GetHandle ());
```
rolfbjarne, line 154:
```suggestion
			nw_connection_group_reply (GetCheckedHandle (), inboundMessage.GetCheckedHandle  (), outboundMessage.GetCheckedHandle (), content.GetHandle ());
```

rolfbjarne (approved):

spouliot (approved):

dalexsoto (approved):

mandel-macaque:
Failure is unrelated with the PR.

PR #9630, opened 2020-09-14
[Network] Add nullability support.

mandel-macaque:
Fixed 2 issues that have been raised by the compiler.

### DIFF
--- a/src/Network/NWAdvertiseDescriptor.cs
+++ b/src/Network/NWAdvertiseDescriptor.cs
@@ -6,6 +6,8 @@
 //
 // Copyrigh 2018 Microsoft Inc
 //
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using ObjCRuntime;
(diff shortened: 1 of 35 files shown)


rolfbjarne (approved):

whitneyschmidt (approved):

tj-devel709 (approved):

spouliot (commented):
Large changes in `main` will make back porting `xcode12` harder - even if this framework has not really been updated yet

spouliot (changes requested):
see breaking changes

it's sadly easy to create a nullable type (breaking change) instead of a nullable annotation (which is not a breaking change)

Also we should figure out why xtro does not report (and does not seem to ignore) nullability for this framework

dalexsoto (approved):

mandel-macaque:
@spouliot xtro does not pick it up, but mainly because the headers do not state it :/

mandel-macaque:
@spouliot the delegate change was a mistake, is an enum and will never be null. xtro was ran with no issue.

mandel-macaque:
Failure is due to a known issue https://github.com/xamarin/maccore/issues/1605

spouliot (approved):

mandel-macaque:
@monojenkins backport xcode12

PR #9637, opened 2020-09-15
[src] Build the .NET version of our product assemblies using a .NET 5 BCL.

rolfbjarne:
We're not shipping the .NET product assemblies in any stable release, so we
can use a preview version of the .NET 5 BCL.

Also:

* Add all the nuget feeds we need to the top-level NuGet.config, even for .NET
  5/6, there shouldn't be any conflicts with stable feeds since we use exact
  version numbers.

* Generate a top-level global5.json which is copied to every directory that
  needs a .NET 5 global.json (overriding the .NET 3.1 global.json in the root
  directory).


dalexsoto (approved):

tj-devel709 (approved):

spouliot (approved):

whitneyschmidt (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated:

* monotouch-test: https://github.com/xamarin/maccore/issues/2304
* api comparison: it fails because the .NET BCL changed and it can't build the previous commit because the old BCL isn't around to use as reference assemblies.

PR #9638, opened 2020-09-15
[xcode12][intro] Fix running introspection on old macOS versions

spouliot:
Most fixes are inside Intents. Some types were not available on macOS
and marked as such, except it backfired.

* Adding `[NoMac]` on `XAMCORE_4_0` was fine
* Adding `[Obsolete]` outside `XAMCORE_4_0` was fine
* Removing the `[Mac (x,y)]` was not quite fine. It's true (since it was never on macOS) but removing it means it default to the oldest (10.9) macOS version we support. This is what the introspection tests were expecting.

Adding an `[Obsoleted (..., 10,0, ...)]` solve this.

### DIFF
--- a/src/intents.cs
+++ b/src/intents.cs
@@ -57,6 +57,7 @@ namespace Intents {
 	[NoMac]
 #elif MONOMAC
 	[Obsolete ("Unavailable on macOS, will be removed in the future.")]
+	[Obsoleted (PlatformName.MacOSX, 10,0, message: "Unavailable on macOS, will be removed in the future.")]
 #endif
 	[iOS (10, 0)]
 	[Watch (3, 2)]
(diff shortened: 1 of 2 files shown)


rolfbjarne (approved):

dalexsoto (approved):

src/intents.cs
```diff
@@ -57,6 +57,7 @@ public enum INBookRestaurantReservationIntentCode : long {
 	[NoMac]
 #elif MONOMAC
 	[Obsolete ("Unavailable on macOS, will be removed in the future.")]
+	[Obsoleted (PlatformName.MacOSX, 10,0, message: "Unavailable on macOS, will be removed in the future.")]
```
tj-devel709, line 60: Is there a difference between `Obsolete` and `Obsoleted`?

tj-devel709 (approved):

PR #9647, opened 2020-09-16
[mtouch] Refactor how we detect and avoid broken frameworks when building for the simulator.

rolfbjarne:
* Use the existing information we have in the Frameworks class to determine
  whether a particular framework works in the simulator or not.
* Show a warning (MX5223) when we run into such a framework.


spouliot (approved):
but should no hold on the release

mandel-macaque (approved):

rolfbjarne:
This warning is going to be annoying, since it'll trigger on every non-linked simulator build that doesn't use the fast sim support 😒 

I think I'll just have to downgrade it to a log message.

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2215)

Issue #9648, opened 2020-09-16
[Generator] The generator does not support the PartialAttribute in protocols


mandel-macaque:
Being able to generate protocols as partial classes/interfaces allows to solve possible API errors. Supporting the PartialAttribute will allow to add the deprecated methods in a manual code to ensure the API is not broken.

PR #9651, opened 2020-09-16
[Metal][Generator] Fix the SetBuffers methods in protocols. Fixes #9649

### LINKED ISSUE #9649: [Metal] SetBuffers calls in Metal are not correctly bound and wont work.
Following the conversation in https://github.com/xamarin/xamarin-macios/pull/9569 from comment https://github.com/xamarin/xamarin-macios/pull/9569#discussion_r486111804 the SetBuffers calls are wrong since they take a C pointer.

mandel-macaque:
Fix the generator to allow the creation of partial protocols, that
allows to create a partial protocol with the old API and define the
correct new API.

Both the Wrapper and the Protocol have to be partial. Tests have been
added for:

1. Generator changes
2. Metal calls.

Fixes https://github.com/xamarin/xamarin-macios/issues/9649
Fixes https://github.com/xamarin/xamarin-macios/issues/9648


spouliot (approved):
In theory we could generate everything as `partial` but since we already have the attribute (and I don't want to step another trap for @mandel-macaque) we might as well use it

mandel-macaque:
No traps please :)

src/metal.cs
```diff
@@ -444,9 +447,13 @@ partial interface MTLComputeCommandEncoder : MTLCommandEncoder {
...
 		void DispatchThreadgroups (IMTLBuffer indirectBuffer, nuint indirectBufferOffset, MTLSize threadsPerThreadgroup);
 
+#if XAMCORE_4_0
 		[Abstract]
+#else
 		[Export ("setBuffers:offsets:withRange:")]
-		void SetBuffers (IMTLBuffer [] buffers, IntPtr offsets, NSRange range);
+		void SetBuffers (IntPtr buffers, IntPtr offsets, NSRange range);
+#endif
+
```
rolfbjarne, line 456: This isn't quite right for XAMCORE_4_0, it'll end up as just the `[Abstract]` attribute and nothing else.

src/generator.cs
```diff
@@ -5562,7 +5562,7 @@ bool IsRequired (MemberInfo provider, Attribute [] attributes = null)
 		return false;
 	}
 
-	void GenerateProtocolTypes (Type type, string class_visibility, string TypeName, string protocol_name, ProtocolAttribute protocolAttribute)
+	void GenerateProtocolTypes (Type type, string class_visibility, string TypeName, string protocol_name, ProtocolAttribute protocolAttribute, bool isPartial=false)
```
rolfbjarne, line 5565:
```suggestion
	void GenerateProtocolTypes (Type type, string class_visibility, string TypeName, string protocol_name, ProtocolAttribute protocolAttribute, bool isPartial = false)
```

src/Metal/MTLArgumentEncoder.cs
```diff
@@ -1,11 +1,54 @@
...
+#if !XAMCORE_4_0
+		[Obsolete ("Use 'SetManagedBuffers' instead.")]
 		public unsafe static void SetBuffers (this IMTLArgumentEncoder This, IMTLBuffer [] buffers, nint [] offsets, Foundation.NSRange range)
+			=> throw new NotSupportedException ();
+#endif
+
+#if XAMCORE_4_0
+		public static void SetBuffers (this IMTLArgumentEncoder table, IMTLBuffer[] buffers, nuint[] offsets, NSRange range)
+#else
+		public static void SetManagedBuffers (this IMTLArgumentEncoder table, IMTLBuffer[] buffers, nuint[] offsets, NSRange range)
```
rolfbjarne, line 32: Why can't you just use `SetBuffers` always?

src/Metal/MTLArgumentEncoder.cs
```diff
@@ -1,11 +1,54 @@
...
+			for (var i = 0; i < buffers.Length; i++) {
+				bufferPtrArray [i] = buffers [i].Handle;
+			}
+
+			unsafe {
+				fixed (void* buffersPtr = bufferPtrArray)
+				fixed (void* offsetsPtr = offsets) { // can use fixed
+					table.SetBuffers ((IntPtr) buffersPtr, (IntPtr) offsetsPtr, range);
+				}
+			}
```
rolfbjarne, line 51: I think you need to add this at the end here, otherwise the GC can theoretically collect the array (and everything within):

```csharp
GC.KeepAlive (buffers);
```

Even that isn't bullet proof (another thread can write to the array and change the entries, causing overwritten entries to be collectable), but that seems like something we don't want to support.

src/Metal/MTLComputeCommandEncoder.cs
```diff
@@ -0,0 +1,51 @@
...
+				bufferPtrArray [i] = buffers [i].Handle;
+			}
+
+			unsafe {
+				fixed (void* buffersPtr = bufferPtrArray)
+				fixed (void* offsetsPtr = offsets) { // can use fixed
+					table.SetBuffers ((IntPtr) buffersPtr, (IntPtr) offsetsPtr, range);
+				}
+			}
+		}
```
rolfbjarne, line 49: Same, `GC.KeepAlive (buffers)` at the end.

rolfbjarne (changes requested):

tj-devel709 (approved):

src/Metal/MTLArgumentEncoder.cs
mandel-macaque: With ArgumentEncode I think we can not with the other.

dalexsoto (approved):

whitneyschmidt (approved):

src/Metal/MTLArgumentEncoder.cs
rolfbjarne: I don't see why not, because there will be only 1 method with this signature. Does it fail to compile if you try?

rolfbjarne (approved):

PR #9652, opened 2020-09-16
[tests][intro] Split attributes typos from API typos tests

spouliot:
The latter requires the spellchecker which varies by OS versions, so we
only run it on the latest OS version (and simulator, except macOS).

The former are some internal rules and can be run on every commit and
avoid finding issues late in a release cycle.

Also changed to
- process members even if a type is obsoleted
- process the properties and events on types

Note: The PR doe snot include any fixes (to typos) since it would
make the xcode12 merge harder. They will be fixed later

### DIFF
--- a/tests/introspection/ApiTypoTest.cs
+++ b/tests/introspection/ApiTypoTest.cs
@@ -738,14 +738,43 @@ namespace Introspection
 		}
 
 		[Test]
+		public virtual void AttributeTypoTest ()
+		{
+			var types = Assembly.GetTypes ();
+			int totalErrors = 0;
+			foreach (Type t in types)
+				AttributeTypo (t, ref totalErrors);
+
+			Assert.IsTrue ((totalErrors == 0), "Attributes have {0} typos!", totalErrors);
+		}
+
+		void AttributeTypo (Type t, ref int totalErrors)
+		{
+			AttributesMessageTypoRules (t, t.Name, ref totalErrors);
+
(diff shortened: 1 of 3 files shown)


tj-devel709 (approved):

tests/introspection/ApiTypoTest.cs
```diff
@@ -737,15 +737,44 @@ bool IsObsolete (MemberInfo mi)
...
 
+		[Test]
+		public virtual void AttributeTypoTest ()
+		{
+			var types = Assembly.GetTypes ();
+			int totalErrors = 0;
+			foreach (Type t in types)
+				AttributeTypo (t, ref totalErrors);
+
+			Assert.IsTrue ((totalErrors == 0), "Attributes have {0} typos!", totalErrors);
```
rolfbjarne, line 748: This seems a bit simpler, and provides the same information:

```suggestion
			Assert.AreEqual (0, totalErrors, "Attributes have typos!");
```

rolfbjarne (approved):

dalexsoto (approved):

tests/introspection/ApiTypoTest.cs
spouliot: agreed - that's a copy/paste of the original test

whitneyschmidt (approved):

rolfbjarne (approved):

PR #9654, opened 2020-09-16
[Jenkins] Zip and upload test results to azure to be later used by vsdrops.

mandel-macaque:
Test results should be uploaded to azure blob storage to later be consumed by vsdrops and moved there.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -951,6 +951,10 @@ timestamps {
                                                 } else {
                                                     echo ("${currentStage} succeeded")
                                                 }
+                                                // create a zip that contains the test run results and upload them to azure to be later served in vsdrops
+                                                echo ("Compressing and uploading test results")
+                                                sh ("7z a ${reportPrefix}.7z ${reportPrefix}")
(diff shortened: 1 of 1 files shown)


jenkins/Jenkinsfile
```diff
@@ -951,6 +951,10 @@ timestamps {
                                                 } else {
                                                     echo ("${currentStage} succeeded")
                                                 }
+                                                // create a zip that contains the test run results and upload them to azure to be later served in vsdrops
+                                                echo ("Compressing and uploading test results")
+                                                sh ("7z a ${reportPrefix}.7z ${reportPrefix}")
+                                                uploadFiles ("${reportPrefix}.7z", "wrench", virtualPath)
```
rolfbjarne, line 957: I believe this line is fatal:

> error ("Test run failed: ${reportPrefix}/tests/index.html")

which means this will never be reached if there are test failures.

Also the packaged results should be password-protected.

rolfbjarne (changes requested):

mandel-macaque:
Closing, we are not investing in Jenkins anymore.

PR #9661, opened 2020-09-17
[monotouch-test] Rework CBCentralManagerTest to handle not having permissions.

rolfbjarne:
1. Assert that we got permission to CoreBluetooth before doing anything. This
   makes it so that if permission has been denied, or if we're running on the
   bots and permission has not been granted, these tests won't run. It does
   not affect local test runs where permissions haven't been granted nor
   denied.

2. Rework the code to wait for CoreBluetooth to power on in the SetUp method.
   Mark the test as inconclusive if CoreBluetooth never powers on (this
   happens if nobody answers the bluetooth permission dialog, quite typical
   when running tests locally).

Fixes these tests when running locally on a tvOS device:

    MonoTouchFixtures.CoreBluetooth.CBCentralManagerTest
        [FAIL] Constructors
        [FAIL] ScanForPeripherals

### DIFF
--- a/tests/monotouch-test/CoreBluetooth/CentralManagerTest.cs
+++ b/tests/monotouch-test/CoreBluetooth/CentralManagerTest.cs
@@ -27,17 +27,13 @@ namespace MonoTouchFixtures.CoreBluetooth {
 	public class CBCentralManagerTest {
 
 		class ManagerDelegate : CBCentralManagerDelegate {
-			public AutoResetEvent Event { get; private set; }
-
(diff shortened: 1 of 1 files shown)


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/701)

whitneyschmidt (approved):

tj-devel709 (approved):

mandel-macaque (approved):

dalexsoto (approved):

PR #9684, opened 2020-09-23
[d16-8] Merge xcode12 into d16-8.

rolfbjarne:
There were are few more commits on xcode12 after the previous merge started,
this is those commits.


dalexsoto (approved):

spouliot (approved):

spouliot:
not related to this PR but
```
public const string Version = "13.20.2" "14.4.0";
```
in API diff means we still need to bump the URL to point to 14.0 (or the API diff won't be correct)

rolfbjarne:
@monojenkins backport d16-8-xm

PR #9686, opened 2020-09-23
[tests][cecil] Check for absence of `[NoX]` (Unavailable) in platform assemblies. Fix #4835

### LINKED ISSUE #4835: [intro] Check for absence of `[NoX]` (Unavailable) in platform assemblies
The generator process `[NoX]`, like `[NoiOS]`, as an instruction not to generate bindings for a specific platform.

However it can also be used outside the generator, which means the symbols will be part of the platform assembly. That can be a problem since that symbol might now work and could be prohibited.

We should add an introspection test to look for such symbols and report an error.

spouliot:
It's way too easy to forget that attributes like `[NoiOS]` means the code
is not generated (for bindings) on that platform but that they will be
compiled for _manual_ bindings (not run thru the generator).

This can expose types (and API) that are not usable on some platforms.
This new test checks that the `[No*]` and `[Unavailable]` attributes
are not in their respective platform assemblies.

For compatibility (existing mistakes) we ignore the check on API that
are decorated with `[Obsolete]` attributes.

Changes in the bindings are fix such mistakes - mostly adding the
`[Obsolete]` attribute.

Fix https://github.com/xamarin/xamarin-macios/issues/4835


spouliot:
Let's not merge this before `xcode12` is backported (and merged) into `main`

rolfbjarne (approved):

mandel-macaque (approved):
Agree 100% do not merge until xcode12 is there (working on it).

tj-devel709 (approved):

mandel-macaque:
@spouliot will need a rebase or a merge with main, but we have xcode12 there.

PR #9688, opened 2020-09-23
[dotnet] Mark our platform assemblies as linkable.

rolfbjarne:
Setting IsTrimmable will mark our assemblies as linkable.

This way we get closer to using the .NET way of choosing what the linker
should do.


rolfbjarne:
CC @jonathanpeppers I think this is something you might want to do as well, if you haven't already :smile:

spouliot (approved):

jonathanpeppers (approved):
We added this in https://github.com/xamarin/xamarin-android/pull/5115 👍

tj-devel709 (approved):

PR #9697, opened 2020-09-24
[ApiDiff] Bump apidiff ref to xcode12.

mandel-macaque:


rolfbjarne (approved):

spouliot (approved):
also needed in d16-8
unless I missed it ?

mandel-macaque:
@spouliot I'm stupid, wanted to do this on d16-8, looks like changing the target branch is a bad idea. I'll close and recreate.

PR #9707, opened 2020-09-25
[dotnet-linker] Add a RemoveAttributesStep that will remove some of the attributes we don't need.

rolfbjarne:
The .NET linker comes with a way to remove attributes (by passing '--link-attributes
some.xml' as a command-line argument), but this has a few problems:

* We'd need to figure out which attributes to remove before running the linker,
  but the code to figure out which optimizations have been enabled (and which attributes
  should be removed) is in our custom linker code. We'd need to refactor a big chunk
  of code to move this logic out of our custom linker code.
* We need to keep the removed attributes around, because the static registrar needs
  them. Our custom linker logic is not notified for removed attributes, which means
  we'd need to store all attributes for the attribute types we're interested in (as
  opposed to this solution, where we only store attributes that are actually removed).
* The attributes we want removed may contain references to types we don't want
  linked away. If we ask the linker to remove those attributes, then the types may
  be linked away as well, and there's no good way around this.

The end result is that a custom step is the best solution for now.

Fixes these monotouch-test tests when enabling all optimizations:

    Xamarin.BindingTests.ProtocolTest
        [FAIL] OnlyProtocol :   [Protocol] IP1
            Expected: 0
            But was:  1
                at Xamarin.BindingTests.ProtocolTest.OnlyProtocol() in /Users/rolf/work/maccore/main/xamarin-macios/tests/bindings-test/ProtocolTest.cs:line 47

        [FAIL] ProtocolWithBaseType :   [Protocol] IP2
            Expected: 0
            But was:  1
                at Xamarin.BindingTests.ProtocolTest.ProtocolWithBaseType() in /Users/rolf/work/maccore/main/xamarin-macios/tests/bindings-test/ProtocolTest.cs:line 79

        [FAIL] ProtocolWithBaseTypeAndModel :   [Protocol] IP3
            Expected: 0
            But was:  1
                at Xamarin.BindingTests.ProtocolTest.ProtocolWithBaseTypeAndModel() in /Users/rolf/work/maccore/main/xamarin-macios/tests/bindings-test/ProtocolTest.cs:line 115

### DIFF
--- a/tools/dotnet-linker/Steps/RemoveAttributesStep.cs
+++ b/tools/dotnet-linker/Steps/RemoveAttributesStep.cs
@@ -1,0 +1,61 @@
+using System;
+
+using Mono.Cecil;
+using Mono.Tuner;
+using Xamarin.Tuner;
+
+namespace Xamarin.Linker.Steps {
+	// The .NET linker comes with a way to remove attributes (by passing '--link-attributes
+	// some.xml' as a command-line argument), but this has a few problems:
+	// 
+	// * We'd need to figure out which attributes to remove before running the linker,
+	//   but the code to figure out which optimizations have been enabled (and which attributes
+	//   should be removed) is in our custom linker code. We'd need to refactor a big chunk
+	//   of code to move this logic out of our custom linker code.
+	// * We need to keep the removed attributes around, because the static registrar needs
+	//   them. Our custom linker logic is not notified for removed attributes, which means
+	//   we'd need to store all attributes for the attribute types we're interested in (as
+	//   opposed to this solution, where we only store attributes that are actually removed).
+	// * The attributes we want removed may contain references to types we don't want
+	//   linked away. If we ask the linker to remove those attributes, then the types may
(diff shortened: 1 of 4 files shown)


whitneyschmidt (approved):

dalexsoto (approved):

tools/dotnet-linker/Steps/RemoveAttributesStep.cs
```diff
@@ -0,0 +1,61 @@
...
+				goto default;
+			case Namespaces.Foundation:
+				switch (attr_type.Name) {
+				case "ProtocolAttribute":
+				case "ProtocolMemberAttribute":
+					return LinkContext.App.Optimizations.RegisterProtocols == true;
+				}
+				goto default;
+			default:
+				return false;
```
mandel-macaque, line 51: I think a switch expression could have made it more readable, but no need to change it.

mandel-macaque (approved):

PR #9719, opened 2020-09-29
[CarPlay] Update bindings to Xcode 12.2 Beta 1

dalexsoto:
CPEntity docs got removed https://developer.apple.com/documentation/carplay/cpentity
so it is very likely that Apple will keep this out from the API.

### DIFF
--- a/src/CarPlay/CPCompat.cs
+++ b/src/CarPlay/CPCompat.cs
@@ -1,0 +1,33 @@
+//
+// CPCompat.cs
+//
+// Authors:
+//	Alex Soto <alexsoto@microsoft.com>
+//
+// Copyright (c) Microsoft Corporation.
+//
+
+using System;
+using Foundation;
+using ObjCRuntime;
+using System.ComponentModel;
+
+#if !XAMCORE_4_0
+namespace CarPlay {
+	[Register (SkipRegistration = true)]
+	[Introduced (PlatformName.iOS, 14, 0)]
+	[Obsoleted (PlatformName.iOS, 14, 2, message: "API removed.")]
+	public class CPEntity : NSObject, INSSecureCoding {
+
+		public CPEntity () => throw new NotImplementedException ();
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		protected CPEntity (NSObjectFlag t) => throw new NotImplementedException ();
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
+		protected internal CPEntity (IntPtr handle) => throw new NotImplementedException ();
+
+		public virtual void EncodeTo (NSCoder coder) => throw new NotImplementedException ();
+	}
+}
+#endif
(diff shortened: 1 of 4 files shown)


src/CarPlay/CPCompat.cs
```diff
@@ -0,0 +1,35 @@
...
+#if !XAMCORE_4_0
+namespace CarPlay {
+	[Register (SkipRegistration = true)]
+	[Introduced (PlatformName.iOS, 14, 0)]
+	[Unavailable (PlatformName.iOS, PlatformArchitecture.All)]
+	[Obsolete ("'CPEntity' has been removed from iOS.")]
+	[EditorBrowsable (EditorBrowsableState.Never)]
+	public class CPEntity : NSObject, INSSecureCoding {
+
+		public CPEntity () => throw new NotImplementedException ();
```
rolfbjarne, line 24: I think we usually throw `NotSupportedException` instead of `NotImplementedException`.

rolfbjarne (approved):

src/CarPlay/CPCompat.cs
```diff
@@ -0,0 +1,35 @@
...
+using Foundation;
+using ObjCRuntime;
+using System.ComponentModel;
+
+#if !XAMCORE_4_0
+namespace CarPlay {
+	[Register (SkipRegistration = true)]
+	[Introduced (PlatformName.iOS, 14, 0)]
+	[Unavailable (PlatformName.iOS, PlatformArchitecture.All)]
+	[Obsolete ("'CPEntity' has been removed from iOS.")]
```
spouliot, line 20: we need to normalize those strings (to reduce the metadata size of our assemblies)
pick up an existing one, there's no need to mention the type name (or the OS), the IDE will put this in context

src/CarPlay/CPCompat.cs
```diff
@@ -0,0 +1,35 @@
...
+	[EditorBrowsable (EditorBrowsableState.Never)]
+	public class CPEntity : NSObject, INSSecureCoding {
+
+		public CPEntity () => throw new NotImplementedException ();
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
```
spouliot, line 26: there's already a `.Never` on the type
not sure what happens with the extra one - but `.Never` is the right one for obsolete code such as this

src/CarPlay/CPCompat.cs
```diff
@@ -0,0 +1,35 @@
...
+		protected CPEntity (NSObjectFlag t) => throw new NotImplementedException ();
+
+		[EditorBrowsable (EditorBrowsableState.Advanced)]
```
spouliot, line 29: same

spouliot (changes requested):

whitneyschmidt (approved):

spouliot (approved):

mandel-macaque (changes requested):
The compat code should maintain the API that was public.

dalexsoto:
@mandel-macaque ?? What am I missing here?

mandel-macaque (approved):

PR #9720, opened 2020-09-29
[CoreFoundation] Update bindings up to Xcode 12.2 Beta 2

dalexsoto:


src/CoreFoundation/CFBundle.cs
```diff
@@ -608,5 +608,29 @@ static string FourCCToString (uint code)
...
 			}
 		}
+
+#if MONOMAC
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool CFBundleIsExecutableLoadable (IntPtr bundle);
+
+		public static bool IsExecutableLoadable (CFBundle bundle) => CFBundleIsExecutableLoadable (bundle.Handle);
```
rolfbjarne, line 618: NRE if bundle is null (also might be worth checking for a disposed bundle using GetCheckedHandle if that's available for CFBundle).

src/CoreFoundation/CFBundle.cs
```diff
@@ -608,5 +608,29 @@ static string FourCCToString (uint code)
...
+
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool CFBundleIsExecutableLoadableForURL (IntPtr bundle);
+
+		public static bool IsExecutableLoadable (NSUrl url) => CFBundleIsExecutableLoadableForURL (url.Handle);
```
rolfbjarne, line 625: Same: potential NRE

rolfbjarne (changes requested):
Manual API needs manual tests 😄

src/CoreFoundation/CFBundle.cs
spouliot: same concern as @rolfbjarne 
also the `[Introduced]` attribute should be on the public method (to be useful to developers)

src/CoreFoundation/CFBundle.cs
spouliot: same concern as @rolfbjarne 
also the `[Introduced]` attribute should be on the public method (to be useful to developers)

src/CoreFoundation/CFBundle.cs
```diff
@@ -608,5 +608,29 @@ static string FourCCToString (uint code)
...
+
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
+		extern static bool CFBundleIsArchitectureLoadable (/*cpu_type_t => integer_t => int*/ Architecture architecture);
+
+		public static bool IsArchitectureLoadable (Architecture architecture) => CFBundleIsArchitectureLoadable (architecture);
```
spouliot, line 632: the `[Introduced]` attribute should be on the public method (to be useful to developers)

spouliot (changes requested):

whitneyschmidt (approved):

mandel-macaque (changes requested):
NRE needs to be fixed.

dalexsoto:
I'll add tests once Beta 2 PR is merged

src/CoreFoundation/CFBundle.cs
```diff
@@ -608,5 +608,44 @@ static string FourCCToString (uint code)
 				}
 			}
 		}
+
+#if MONOMAC
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		[DllImport (Constants.CoreFoundationLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
```
tj-devel709, line 615: What does this line do?

tj-devel709 (approved):

src/CoreFoundation/CFBundle.cs
spouliot: @tj-devel709 Boolean types have different size depending in the language and platform, e.g.  the C `BOOL` on Win32 was 32bits, but it's often 1 byte, like in C++.

That can be a problem if the boolean value is written as 1 byte (e.g. on native side) then read as 4 bytes (e.g. on managed side). In this case the 3 other bytes (since it's often aligned on 4 bytes, even if it takes only one) content is undefined.

So, from the above example, `0-1-1-1` would be a `(bool)0x00` on native and `(bool)0x00010101` on managed which, beside being different, make things rather interesting at runtime :)

The `[MarshalAs]` attribute tells the runtime how to interpret the value, where `I1` means 1-byte integer.

src/CoreFoundation/CFBundle.cs
```diff
@@ -615,20 +615,35 @@ static string FourCCToString (uint code)
 		[return: MarshalAs (UnmanagedType.I1)]
 		extern static bool CFBundleIsExecutableLoadable (IntPtr bundle);
 
-		public static bool IsExecutableLoadable (CFBundle bundle) => CFBundleIsExecutableLoadable (bundle.Handle);
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		public static bool IsExecutableLoadable (CFBundle bundle) {
```
spouliot, line 619: minor: style `{` does on next line for methods

src/CoreFoundation/CFBundle.cs
```diff
@@ -615,20 +615,35 @@ static string FourCCToString (uint code)
...
+		}
 
 		[Introduced (PlatformName.MacOSX, 11, 0)]
 		[DllImport (Constants.CoreFoundationLibrary)]
 		[return: MarshalAs (UnmanagedType.I1)]
 		extern static bool CFBundleIsExecutableLoadableForURL (IntPtr bundle);
 
-		public static bool IsExecutableLoadable (NSUrl url) => CFBundleIsExecutableLoadableForURL (url.Handle);
+		[Introduced (PlatformName.MacOSX, 11, 0)]
+		public static bool IsExecutableLoadable (NSUrl url) {
```
spouliot, line 634: minor: style `{` does on next line for methods

spouliot (approved):

rolfbjarne (approved):

src/CoreFoundation/CFBundle.cs
tj-devel709: Thank you very much for the explanation!

PR #9733, opened 2020-09-30
[tests] Remove duplicated code to set the requirement environment variables for building locally.

rolfbjarne:


tests/common/mac/ProjectTestHelpers.cs
```diff
@@ -244,41 +244,21 @@ public static void CleanUnifiedProject (string csprojTarget)
 
 		public static string BuildClassicProject (string csprojTarget)
```
spouliot, line 245: is that method (classic) even needed anymore ?

spouliot (approved):

tests/common/mac/ProjectTestHelpers.cs
rolfbjarne: I wondered the same thing, and I tried to test it, and the test failed because Xamarin.Mac is disabled in main... so I couldn't say for sure whether we can remove it or not. The test that calls it was disabled in the xcode12 branch (it's a test that verifies that we get a decent error message when the user tries to build a Classic project), because something in VSMac fails, so it might be that we could remove both the test and this method because VSMac is broken enough so that users would never end up in mmp anyway. I decided to punt it for now, until Xcode 12.2 is stable.

chamons (approved):
:heart:

mandel-macaque (approved):

whitneyschmidt (approved):

mandel-macaque (approved):

PR #9735, opened 2020-09-30
[msbuild] Share the _*CompileSceneKitAssets targets between Xamarin.iOS and Xamarin.Mac.

rolfbjarne:
The Xamarin.iOS version is more updated, so that's the version now for both
Xamarin.iOS and Xamarin.Mac:

* The iOS version contains Inputs/Outputs (important for incremental builds).
* It doesn't use FileWrites (see e97d69b25cc2b85e89feb775169f46f467c66f7a why this was removed for iOS)

Additionally define the _SceneKitCache variable for macOS as well, previously
it was only defined for iOS (I'm adding more variables than just
_SceneKitCache to macOS, most of the others will be used in in macOS in
upcoming pull requests, and this way turned out to be simplest).


spouliot (approved):
is there any project (samples?) that triggers the macOS code ?

chamons (approved):

rolfbjarne:
> is there any project (samples?) that triggers the macOS code ?

I ran the macOS MSBuild tests, but it doesn't look like we have any tests there that exercises the macOS build logic. I'll add a test.

mandel-macaque (approved):

whitneyschmidt (approved):

emaf (approved):

spouliot (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/1083)

Issue #9724, opened 2020-09-29
[intro] verify we have no availability attributes on enums *members* that represent error codes


spouliot:
We do not _normally_ add availability attributes on enums **members** that represent error codes. In part because it's a lot of metadata and, foremost, because it's not really helpful to write code. E.g.

```csharp
var err = Call.Api (1);
switch (err) {
case NSError.Bad:
case NSError.Wrong:
   Console.WriteLine ($"API failed: {err});
   break;
case NSError.Ok:
   break;
default:
   Console.WriteLine ($"Unknown error code {err}");
   break;
}
```

Adding version checks inside this would be complicated (source wise) and not really helpful since
* API can return undefined error code (and the error logic should work);
* Availability information is not 100% accurate;

As such we default to not add them - but we some time forgot about it. An intro rule could easily ensure we don't add them needlessly.

spouliot:
In most cases an `[ErrorDomain]` attribute is on the type itself, so we known it's an error code. 

Some other cases might need to be manually added to the test to cover cases like https://github.com/xamarin/xamarin-macios/pull/9718

PR #9737, opened 2020-09-30
[tests][cecil] Check that error-based enum values don't have availability attributes

spouliot:
and fixes the ones that have some.

From https://github.com/xamarin/xamarin-macios/issues/9724

We do not _normally_ add availability attributes on enums **members** that represent error codes. In part because it's a lot of metadata and, foremost, because it's not really helpful to write code. E.g.

```csharp
var err = Call.Api (1);
switch (err) {
case NSError.Bad:
case NSError.Wrong:
   Console.WriteLine ($"API failed: {err});
   break;
case NSError.Ok:
   break;
default:
   Console.WriteLine ($"Unknown error code {err}");
   break;
}
```

Adding version checks inside this would be complicated (source wise) and not really helpful since
* API can return undefined error code (and the error logic should work);
* Availability information is not 100% accurate;

As such we default to not add them - but we some time forgot about it. An intro rule could easily ensure we don't add them needlessly.

### DIFF
--- a/src/AVFoundation/Enums.cs
+++ b/src/AVFoundation/Enums.cs
@@ -207,30 +207,19 @@ namespace AVFoundation {
 		AirPlayControllerRequiresInternet = -11856,
 		AirPlayReceiverRequiresInternet = -11857,
 
-		[iOS (9,0), Mac (10,11)]
 		VideoCompositorFailed = -11858,
 
 #if !MONOMAC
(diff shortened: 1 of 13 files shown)


rolfbjarne (approved):

dalexsoto (approved):
❤️

spouliot:
@monojenkins backport to xcode12.2

Issue #6154, opened 2019-05-28
Add C# 8 nullability annotations to bindings


spouliot:
Xamarin.iOS|Mac bindings already support `[NullAllowed]`. This controls if `null` checks are generated inside (the body of) API bindings.

We often talked about doing more tooling but it did not yet materialize. Now the upcoming C# 8 supports nullability annotations so our attributes are even more useful - with even less work on our side (tooling wise).

**Steps**

[ ] Add nullability support to xtrospection tests - so we know where our bindings are missing (or extra) `[NullAllowed]`. The results can be ignore until we're ready for the next step;
[ ] Fix bindings (taking care not to create a merge hell with other existing/upcoming work). This will also eliminate recurrent bugs when we're missing `[NullAllowed]` (since using `null` is not possible). 
[ ] Add C#8 nullability annotations to the generator

**Example**

This binding
```csharp
[Export ("alertControllerWithTitle:message:preferredStyle:")]
UIAlertController Create ([NullAllowed] string title, [NullAllowed] string message, UIAlertControllerStyle preferredStyle);
```

currently generate
```csharp
[Export ("alertControllerWithTitle:message:preferredStyle:")]
[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
public static UIAlertController Create (string title, string message, UIAlertControllerStyle preferredStyle)
{
	global::UIKit.UIApplication.EnsureUIThread ();
	var nstitle = NSString.CreateNative (title);
	var nsmessage = NSString.CreateNative (message);
	...  
}
```

but would now generate
```csharp
[Export ("alertControllerWithTitle:message:preferredStyle:")]
[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
public static UIAlertController Create (string? title, string? message, UIAlertControllerStyle preferredStyle)
{
	global::UIKit.UIApplication.EnsureUIThread ();
	var nstitle = NSString.CreateNative (title);
	var nsmessage = NSString.CreateNative (message);
	...  
}
```

note the extra `?` on the parameter types

rolfbjarne:
Question: is adding/removing the nullability annotation from an existing binding a breaking change?

spouliot:
@rolfbjarne I don't _think_ it is (but need to confirm [1])
else there are breaking changes in the BCL

[1] *guessing* it's a custom attribute

cosminstirbu:
Hello,

[...]

PR #9740, opened 2020-09-30
[Photos] Correct nullability attributes for the Photos framework

rpendleton:
I corrected the nullability attributes of the Photos bindings based on the instructions from @spouliot in [#6154](https://github.com/xamarin/xamarin-macios/issues/6154#issuecomment-620014843).


dalexsoto:
wow thanks a bunch!!

dalexsoto:
build

src/photos.cs
```diff
@@ -400,9 +412,11 @@ interface PHAssetResourceRequestOptions : NSCopying
 	interface PHChange {
 
 		[Export ("changeDetailsForObject:")]
-		PHObjectChangeDetails GetObjectChangeDetails ([NullAllowed] PHObject obj);
```
rpendleton, line 403: Is removing this `[NullAllowed]` attribute from the first method parameter a breaking change?

If I understand things correctly, this change will result in code being generated that throws a `NullArgumentException` when given null, and that wouldn't have happened until now. I tried to test passing `null` to this method before my change and didn't get an exception from Apple, but I'm also not sure if that's just relying on undocumented behavior.

spouliot: > Is removing this [NullAllowed] attribute from the first method parameter a breaking change?

It's a functional change but not a **breaking change** (as defined by .net). Nullability information is for
* macios generator (bgen) which add null-check **inside** methods; and
* csc (roslyn) compiler (for warnings)

This means binaries will still run, source will still build (might warn or error if `/warnaserror` is used, but that's fine).

> not sure if that's just relying on undocumented behavior.

The nullability information has not always been available from Apple. The old attributes were added based on headers comments, documentation, samples, unit tests and common sense.

spouliot (approved):
Looks identical to what I would have done 😄 
Thanks a lot!

rolfbjarne (approved):
This looks great! Thanks!

whitneyschmidt (approved):

mandel-macaque (approved):

dalexsoto (approved):

spouliot:
@monojenkins backport to xcode12.2

PR #9748, opened 2020-09-30
[VideoToolbox] Update for xcode12.2 beta 1.

mandel-macaque:


dalexsoto:
not needed since it does not make much sense `NeedsRosetta`

spouliot:
there's another PR that removes availability attributes from all error codes :)

PR #9760, opened 2020-10-01
[Xcode12] Use the correct version of Xcode.

mandel-macaque:
We have a stable version, do not use the GM. We are lucky is the same
build hash.


spouliot (approved):
`if (!frozen)`

dalexsoto (approved):
With @chamons 👍

chamons (approved):

whitneyschmidt (approved):

PR #9761, opened 2020-10-01
Bump API diff to the released (stable) xcode12 version of XI

spouliot:
This will be a bit confusing (not directly usable from bots) to produce
the API diff for Xamarin.Mac.


mandel-macaque (approved):

spouliot:
unrelated, known failure -> https://github.com/xamarin/maccore/issues/2165

PR #9751, opened 2020-10-01
[tests][intro] ConstantsCheck should only be executed on latest OS

spouliot:
since the files, mapped by the constants, might not be present in earlier
versions of the OS. We can only be sure, of their presence, on the
current/supported SDK version

Fix intro running on iOS 10.3 (64 bits)

```
Introspection.iOSApiTypoTest
    [PASS] AttributeTypoTest
    [FAIL] ConstantsCheck :   NetworkLibrary
  Expected: True
  But was:  False
 :   at Introspection.ApiTypoTest.ConstantsCheck () [0x00168] in /Users/builder/jenkins/workspace/xamarin-macios/xamarin-macios/tests/introspection/ApiTypoTest.cs:1088
```

Issue #9353, opened 2020-08-12
Not quite duplicated strings, mostly in metadata, e.g. [Obsolete] attributes


spouliot:
Examples:

* `git grep Obsolete src/ | grep NotSupported`
* `git grep Obsolete src/ | grep stub`

The compiler can share the same string in several locations - but it must be the exact same (sub)string.

The linker can also remove many of them. However the linker is not always used (e.g. interpreter) and this affect the platform assembly size (e.g. when loaded in memory by the compiler, IDE, mtouch...)

Recommended solution, using string constants for the repetitive messages. This way they can be shared and will consume less space in the platform assemblies, e.g.

```
const string ObsoleteStub = "Empty stub (not a public API).";

...

[Obsolete (Constants.ObsoleteStub)]
```

We could also had an URL for the most common patterns and document the issue in more details.

ref: https://github.com/xamarin/xamarin-macios/pull/9268

PR #9766, opened 2020-10-02
[xcode12.2][tests][intro] Split attributes typos from API typos tests

spouliot:
The latter requires the spellchecker which varies by OS versions, so we
only run it on the latest OS version (and simulator, except macOS).

The former are some internal rules and can be run on every commit and
avoid finding issues late in a release cycle.

Also changed to
- process members even if a type is obsoleted
- process the properties and events on types

About the strings...

Some were fine, others were not. Copy/pasted strings are hard to
maintain. Moving them to constants will help, both maintainability and
will reduce the metadata size of the platform assemblies.

ref: #9353

backport of
https://github.com/xamarin/xamarin-macios/pull/9652
https://github.com/xamarin/xamarin-macios/pull/9751


rolfbjarne (approved):

spouliot:
timeout when [code]signing an unrelated test -> https://github.com/xamarin/maccore/issues/1929

PR #9775, opened 2020-10-05
[msbuild] Share the CodesignNativeLibraries target between Xamarin.iOS and Xamarin.Mac.

rolfbjarne:


msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -710,6 +710,56 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
+		<PropertyGroup>
+			<_CodesignDisableTimestamp>False</_CodesignDisableTimestamp>
+			<_CodesignDisableTimestamp Condition="'$(_SdkIsSimulator)' == 'true' Or '$(_BundlerDebug)' == 'true'">True</_CodesignDisableTimestamp>
+			<_LibraryCodeSigningKey>$(_CodeSigningKey)</_LibraryCodeSigningKey>
+			<_LibraryCodeSigningKey Condition="'$(_LibraryCodeSigningKey)' == ''">-</_LibraryCodeSigningKey>
+		</PropertyGroup>
+
+		<ItemGroup Condition="'$(_PlatformName)' == 'macOS'">
+			<_CodesignNativeLibrary Include="$(_AppContentsPath)\**\*.dylib" />
+			<_CodesignNativeLibrary Include="$(_AppResourcesPath)\**\*.metallib" />
```
spouliot, line 736: Can't they be on a single line, like the non-mac version, like
`"$(_AppContentsPath)\**\*.dylib;$(_AppResourcesPath)\**\*.metallib"`

Not sure how costly it is to parse, but I like having less lines of this 😄

spouliot (approved):

dalexsoto (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.targets
dalexsoto: I think if it is not too costly I'd keep the current version of it as it is easier to read

rolfbjarne: I don't have numbers, but I don't think having one line instead of two amounts to any measurable difference either way.

I put it in two lines here because that made the code easier to read (in particular because the root directory for the globs are different, which stands out with two lines, but is easy to overlook when it's all in one line). The iOS version has the same root directory for both globs.

emaf (approved):

PR #9804, opened 2020-10-07
[dotnet] Use correct relative path for install_name_tool on macOS

filipnavara:


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -381,9 +381,15 @@
 	This doesn't work after linking with these libraries, so this must be fixed to be @executable_path/. Eventually mono will ship dylibs we don't have to fix (https://github.com/dotnet/runtime/issues/34637).
 	-->
 	<Target Name="_UpdateDynamicLibraryId" DependsOnTargets="_ComputeVariables" Inputs="@(_MonoLibrary)" Outputs="@(_MonoLibrary -> '$(_IntermediateNativeLibraryDir)%(Filename)%(Extension)')">
+		<PropertyGroup>
+			<_ExecutablePathPrefix Condition="'$(_PlatformName)' != 'macOS'">@executable_path/</_ExecutablePathPrefix>
+			<_ExecutablePathPrefix Condition="'$(_PlatformName)' == 'macOS'">@executable_path/../MonoBundle/</_ExecutablePathPrefix>
```
rolfbjarne, line 386: Developers can use their own name instead of `MonoBundle`, so this is necessary:

```suggestion
			<_ExecutablePathPrefix Condition="'$(_PlatformName)' == 'macOS'">@executable_path/../$(_CustomBundleName)/</_ExecutablePathPrefix>
```

rolfbjarne (changes requested):

spouliot (approved):

dalexsoto (approved):
After Rolf's 👍

dotnet/targets/Xamarin.Shared.Sdk.targets
filipnavara: Happy to change that. There was one other place in the file where it was hard-coded which is where I copied it from.

filipnavara: Apparently `_CustomBundleName` is not set at this point. I'll look into it tomorrow.

rolfbjarne: > Apparently `_CustomBundleName` is not set at this point. I'll look into it tomorrow.

You can add a dependency on the `_ParseBundlerArguments` target to fix that.

> There was one other place in the file where it was hard-coded which is where I copied it from.

Yes, I still have to update a few existing places.

mandel-macaque (approved):
+1 after rolf commit is added.

dotnet/targets/Xamarin.Shared.Sdk.targets
filipnavara: `_ParseBundlerArguments` doesn't exist in the dotnet builds. There's only `_ComputeLinkerArguments` which doesn't parse `CustomBundleName` yet.

filipnavara: Thanks for the `_ParseBundlerArguments` tip. Apparently my VS Code search is broken and it doesn't recurse in some directories. :-/

rolfbjarne:
build

rolfbjarne (approved):

filipnavara:
I'm working on a better fix for this problem. Looks like the .NET Core system libraries now have proper identity with `@rpath` prefix so the whole `_UpdateDynamicLibraryId` step could be removed. The `LinkNativeCode` however specifies incorrect `-rpath` for macOS builds so that still needs to be fixed.

This PR is still probably worth it as a stopgap solution and because of the removal of hard-coded `MonoBundle` from one place.

PR #9815, opened 2020-10-08
[xcode12.2][qtkit] Stub removed API. Fix #7704

### LINKED ISSUE #7704: Xamarin.Mac: Remove/stub out 32-bit only frameworks (QuickTime, QTKit, QuartzComposer)
We're currently shipping bindings for a few 32-bit only macOS-specific frameworks (QUickTime, QTKit, QuartzComposer). This is redundant since we don't support 32-bit Xamarin.Mac apps anymore.

The enhancement would be to either:

* Remove these bindings (would be a breaking change, most space saved).
* Stub them out and make them throw an exception (no breaking changes, but less space saved).

Ref: #6300.

spouliot:
Xamarin.Mac only supports 64 bits - since the supported macOS and
Xcode have dropped 32bits last year.

QTKit headers were removed but we still had bindings code generated
and kept to avoid breaking changes.

This is now replaced by stubs, which are a smaller (no code and less
metadata).

```
Current Xamarin.Mac.dll 23,990,784 bytes
QTKit-stubbed XM.dll    23,865,856 bytes
Difference                 124,928 bytes
```

reference: https://github.com/xamarin/xamarin-macios/issues/7704


dalexsoto (approved):

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+
+#pragma warning disable 0067
+
+namespace QTKit {
+
+	[ObjCRuntime.Obsoleted (ObjCRuntime.PlatformName.MacOSX, 10,15, message: ObjCRuntime.Constants.MacOS32bitsUnavailable)]
+	public class QTCaptureConnection : Foundation.NSObject {
+		public Foundation.NSObject GetAttribute (System.String attributeKey)
+		{
+			return default (Foundation.NSObject);
```
rolfbjarne, line 12: Some members throw exceptions, other return default values. Is there a rule when one is chosen over the other?

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+
+		public virtual System.Boolean Enabled { get; set; }
+
+		public virtual QTKit.QTFormatDescription FormatDescription { get; }
+
+		public virtual System.String MediaType { get; }
+
+		public virtual Foundation.NSObject Owner { get; }
+
+		static public Foundation.NSString AttributeDidChangeNotification { get; }
```
rolfbjarne, line 62: Not a big deal, but most of our codebase uses `public static` and not `static public`.

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+			return default (Foundation.NSObject);
+		}
+
+		public void SetAttribute (Foundation.NSObject attribute, System.String key)
+		{
+		}
+
+		public QTCaptureConnection ()
+		{
+			throw new System.NotImplementedException (ObjCRuntime.Constants.MacOS32bitsUnavailable);
```
rolfbjarne, line 21: We've used `PlatformNotSupportedException` for other API we've stubbed out.

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+
+		static public Foundation.NSString TimeScaleAttribute { get; }
+
+		static public Foundation.NSString TypeAttribute { get; }
+
+	}
+
+	[ObjCRuntime.Obsoleted (ObjCRuntime.PlatformName.MacOSX, 10,15, message: ObjCRuntime.Constants.MacOS32bitsUnavailable)]
+	public enum QTCodecQuality : System.Int32 {
+	Lossless = 1024,
```
rolfbjarne, line 524: Indentation is off here.

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+		}
+
+		public QTKit.QTTime Time { get; set; }
+
+		public QTKit.QTTime Duration { get; set; }
+
+	}
+
+	[ObjCRuntime.Obsoleted (ObjCRuntime.PlatformName.MacOSX, 10,15, message: ObjCRuntime.Constants.MacOS32bitsUnavailable)]
+	[System.Flags]	public enum TimeFlags : System.Int32 {
```
rolfbjarne, line 1548: Missing newline after `[System.Flags]`

src/QTKit/QTCompat.cs
```diff
@@ -0,0 +1,2737 @@
...
+	StillImageTypes = 1,
+	TranslatableTypes = 2,
+	AggressiveTypes = 4,
+	DynamicTypes = 8,
+	CommonTypes = 0,
+	AllTypes = 65535,
+}
+
+	[ObjCRuntime.Obsoleted (ObjCRuntime.PlatformName.MacOSX, 10,15, message: ObjCRuntime.Constants.MacOS32bitsUnavailable)]
+	[ObjCRuntime.Native]	public enum QTCaptureDevicePlaybackMode : System.UInt64 {
```
rolfbjarne, line 1563: Missing newline after `[ObjCRuntime.Native]`

rolfbjarne (approved):

src/QTKit/QTCompat.cs
spouliot: not sure we are very consistent in our sources, but it's an easy change

src/QTKit/QTCompat.cs
spouliot: yes, it's to minimize the (IL) size. 
In many cases, like this one (and also including current manual stubs), you can't hit members throw since the type can't be instantiated (without throwing).

mandel-macaque (approved):

tj-devel709 (approved):

PR #9823, opened 2020-10-08
[dotnet] Add support for 'dotnet build -t:run'.

rolfbjarne:
* Ship mlaunch in the iOS, tvOS and watchOS NuGets. It should probably go into
  a separate NuGet (to avoid shipping the same mlaunch executable in three different
  packages), but that can be done at a later stage.

* Add a GetMlaunchArguments task that computes the mlaunch arguments to install
  or launch an app in either the simulator or on device.

* Implement the MSBuild logic to make the Run target (provided by .NET) launch
  mlaunch (for iOS, tvOS and watchOS) or the built app (for macOS). This is done
  by setting the RunCommand and RunArguments properties (which the Run target uses)
  to the correct values.

Ideally I'd would make 'dotnet run' work too, but that runs into a different problem which
I haven't figured out yet:

    A fatal error was encountered. The library 'libhostpolicy.dylib' required to execute the application was not found in '/Users/rolf/work/maccore/onedotnet/xamarin-macios/tests/dotnet/MySingleView/bin/Debug/net5.0-ios/ios-x64/'.
    Failed to run as a self-contained app.
      - The application was run as a self-contained app because '/Users/rolf/work/maccore/onedotnet/xamarin-macios/tests/dotnet/MySingleView/bin/Debug/net5.0-ios/ios-x64/MySingleView.runtimeconfig.json' did not specify a framework.
      - If this should be a framework-dependent app, specify the appropriate framework in '/Users/rolf/work/maccore/onedotnet/xamarin-macios/tests/dotnet/MySingleView/bin/Debug/net5.0-ios/ios-x64/MySingleView.runtimeconfig.json'.

That's for a different pull request though.

Ref https://github.com/xamarin/net6-samples/issues/35.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -481,14 +481,7 @@
 		<_MlaunchPath Condition="'$(_MlaunchPath)' == ''">$(_XamarinSdkRootDirectory)tools\bin\mlaunch</_MlaunchPath>
 	</PropertyGroup>
 
-	<Target Name="_PrepareRunMac" DependsOnTargets="_GenerateBundleName" Condition="'$(_PlatformName)' == 'macOS'">
-		<PropertyGroup Condition="'$(_PlatformName)' == 'macOS'">
-			<RunCommand>open</RunCommand>
-			<RunArguments>-a "$(_AppBundlePath)"</RunArguments>
-		</PropertyGroup>
-	</Target>
-
-	<Target Name="_InstallMobile" DependsOnTargets="_DetectSdkLocations;_GenerateBundleName" Condition="'$(_PlatformName)' != 'macOS' And '$(_SdkIsSimulator)' == 'false'">
+	<Target Name="_InstallMobile" DependsOnTargets="_DetectSdkLocations;_GenerateBundleName" Condition="'$(_SdkIsSimulator)' == 'false'">
 		<GetMlaunchArguments
 			SessionId="$(BuildSessionId)"
 			AppBundlePath="$(_AppBundlePath)"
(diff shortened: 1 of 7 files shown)


chamons (approved):

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -475,6 +475,73 @@
...
+
+	<PropertyGroup>
+		<_MlaunchPath Condition="'$(_MlaunchPath)' == ''">$(_XamarinSdkRootDirectory)tools\bin\mlaunch</_MlaunchPath>
+	</PropertyGroup>
+
+	<Target Name="_PrepareRunMac" DependsOnTargets="_GenerateBundleName" Condition="'$(_PlatformName)' == 'macOS'">
+		<PropertyGroup Condition="'$(_PlatformName)' == 'macOS'">
+			<RunCommand>open</RunCommand>
+			<RunArguments>-a "$(_AppBundlePath)"</RunArguments>
+		</PropertyGroup>
```
filipnavara, line 488: This is sort of redundant with https://github.com/xamarin/xamarin-macios/blob/59db0841419c9cbececea878ca74354ebf2a9bad/dotnet/Microsoft.macOS.Sdk/targets/Microsoft.macOS.Sdk.targets#L5-L8 (and doing `Condition="'$(_PlatformName)' == 'macOS'"` is likely unnecessary).

spouliot (approved):

tj-devel709 (approved):

mandel-macaque (approved):

dalexsoto (approved):

emaf (approved):

dotnet/targets/Xamarin.Shared.Sdk.targets
rolfbjarne: 🤦 (twice)

PR #9826, opened 2020-10-08
[Foundation] Update up to Xcode 12.2 Beta 2

dalexsoto:


src/foundation.cs
```diff
@@ -6264,7 +6264,7 @@ partial interface NSUrl : NSSecureCoding, NSCopying
 		[Field ("NSURLIsApplicationKey")]
 		NSString IsApplicationKey { get; }
 
-		[iOS (9,0)][NoMac]
+		[iOS (9,0), Mac(10,11)]
```
mandel-macaque, line 6267: 10,11??? Was this missed from other years, or a typo?

mandel-macaque (approved):

src/foundation.cs
dalexsoto: Headers say `macos(10.11)` but makes sense that it could be a typo from Apple

```suggestion
		[iOS (9,0), Mac(11,0)]
```

rolfbjarne (approved):

PR #9833, opened 2020-10-09
[dotnet] Default to static libxamarin[-debug].a builds on macOS

filipnavara:
This seems to match the default on classic mmp-based projects. It also avoids the relaunching of the app which can result in the UI not being displayed in front of other apps.


rolfbjarne:
build

rolfbjarne (approved):

mandel-macaque (approved):

dalexsoto (approved):

PR #9847, opened 2020-10-12
[Xcode12] Bring xharness changes to xcode12. 

mandel-macaque:
Bring all the changes that have happen in main on xharness to xcode12. In order to review this, you can focus on the diff between main and this PR using:

```bash
git diff main  -- tests/xharness
```

This will result in a much smaller diff than then one against xcode12. Diff can be found [here](https://gist.github.com/mandel-macaque/9486d72d5fce1908d8fa690152dfe83a).

Plans:

Since the history is in main, I plan to do a squash of this change, merge with xcode12.2. Once we merge xcode12.2 with main, revert the change and then merge to make the merge cleaner.


tests/bcl-test/Make.frag
```diff
@@ -0,0 +1,43 @@
+
```
rolfbjarne, line 1: This file was removed in main (39629ab93b8d723b472a52378615dc5cf03553bd), so why is it added here?

tests/bcl-test/Makefile
```diff
@@ -0,0 +1,22 @@
+TOP = ../../..
```
rolfbjarne, line 1: Same: this file does not exist in main anymore.

tests/bcl-test/Make.frag
mandel-macaque: Mistake :)

tests/bcl-test/Makefile
mandel-macaque: Same, a mistake, will remove

dalexsoto (approved):
Tests results look reasonable, I mean failing but due to issues to the tests and not xharness so 👍

tj-devel709 (approved):

rolfbjarne (commented):
The changes look good, but it's a _very_ big change to a release branch, and contains shipping changes as well, so I'm not sure if this is the best way to do this. I'd feel much more comfortable with a targeted fix that only handles the issue we're trying to fix. OTOH there might be no problems... so I'll leave it to @chamons to decide :smile:

mandel-macaque:
@rolfbjarne looking at the diff, we have the possible changes that might ship: 

* mk/xamarin.mk
* src/Makefile
* src/arkit.cs
* tools/common/.gitignore
* tools/common/Make.common
* tools/common/create-makefile-fragment.sh
* tools/mmp/mmp.csproj
* tools/mtouch/Assembly.cs
* tools/mtouch/mtouch.csproj

I'll try to revert those changes and if tests run, we will minimize the danger.

mandel-macaque:
Do not merge, we will be using this PR as an integration branch to get device tests.

mandel-macaque:
No longer required.

PR #9858, opened 2020-10-14
[msbuild] Re-use existing code to launch processes.

rolfbjarne:
There was a race condition in the code that has been removed, which may have
caused actool to randomly hang at exit (https://github.com/xamarin/maccore/issues/1124).
Fix this by re-using existing code we have to launch processes.


rolfbjarne:
Test failures are unrelated

* monotouch-test build failures: fixed with #9859.
* all other failures seem to be various types of launch failures / device problems.

spouliot (approved):

tj-devel709 (approved):

PR #9860, opened 2020-10-14
[tests] Split the iOS msbuild tests in two.

rolfbjarne:
Split the iOS msbuild tests in two:

* Xamarin.MacDev.Tasks.Tests: contains in-process unit tests for tasks.
* Xamarin.MacDev.Tasks: contains larger tests that either invoke targets or a complete
  build. These are currently in-process, but will become out-of-process soon to make
  it possible to run them with dotnet.

Also make the new projects non-iOS-specific, because the macOS msbuild tests will
be moved here as well soon.

There is some duplicated code between these two test projects now (all files
that show up as new are copies of existing files), this will be cleaned up in
later pull requests.


tests/msbuild/Xamarin.MacDev.Tests/TestHelpers/TestBase.cs
```diff
@@ -0,0 +1,561 @@
...
+				}
+				Assert.AreEqual (expectedErrorCount, Engine.Logger.ErrorEvents.Count, "#RunTask-ErrorCount" + messages);
+			}
+		}
+
+		protected string CreateTempFile (string path)
+		{
+			path = Path.Combine (TempDir, path);
+			Directory.CreateDirectory (Path.GetDirectoryName (path));
+			using (new FileStream (path, FileMode.CreateNew)) {}
```
mandel-macaque, line 358: Why not do:
```csharp
u

mandel-macaque (approved):

tj-devel709 (approved):

whitneyschmidt (approved):

spouliot (approved):

PR #9864, opened 2020-10-14
Replace IKVM.Reflection with MetadataLoadContext

filipnavara:
IKVM.Reflection is not actively maintained while MetadataLoadContext is actively maintained as part of dotnet/runtime.


src/generator-attribute-manager.cs
```diff
@@ -728,3 +728,23 @@ public static IEnumerable<T> Yield<T> (this T item)
...
+
+	static CustomAttributeDataExtensions ()
+	{
+		roCustomAttributeDataType = typeof(MetadataLoadContext).Assembly.GetType ("System.Reflection.TypeLoading.RoCustomAttributeData");
+		attributeTypeProperty = roCustomAttributeDataType.GetProperty ("AttributeType");
+	}
+
+	public static Type GetAttributeType (this CustomAttributeData data)
+	{
+		// Workaround for CustomAttributeData.AttributeType not being declared as virtual in Mono
```
rolfbjarne, line 745: You could probably do an `#if !NET` around this workaround.

src/btouch.cs
```diff
@@ -520,11 +498,10 @@ int Main3 (string [] args)
 			AttributeManager = new AttributeManager (this);
 			Frameworks = new Frameworks (CurrentPlatform);
 
-			Assembly corlib_assembly = universe.LoadFile (LocateAssembly ("mscorlib"));
 			// Explicitly load our attribute library so that IKVM doesn't try (and fail) to find it.
-			universe.LoadFile (GetAttributeLibraryPath ());
+			universe.LoadFromAssemblyPath (GetAttributeLibraryPath ());
```
rolfbjarne, line 502: Do you need this line at all, given that you do this further above:

```csharp
assemblyPaths.Add (GetAttributeLibraryPath ());
```

src/generator-attribute-manager.cs
```diff
@@ -199,16 +198,19 @@ System.Type LookupReflectionType (string fullname, ICustomAttributeProvider prov
 	}
 
 	// This method gets the System.Type for a IKVM.Reflection.Type to a System.Type.
-	System.Type ConvertType (Type type, ICustomAttributeProvider provider)
+	System.Type ConvertTypeFromMeta (Type type, ICustomAttributeProvider provider)
 	{
 		var rv = LookupReflectionType (type.FullName, provider);
 		if (rv == null)
+		{
+			Console.WriteLine (Environment.StackTrace);
```
rolfbjarne, line 206: 🙀

src/generator-attribute-manager.cs
```diff
@@ -288,11 +290,11 @@ static IEnumerable<System.Attribute> ConvertOldAttributes (CustomAttributeData a
...
 			return convertedAttributes.OfType<T> ();
 
-		var expectedType = ConvertType (typeof (T), provider);
-		if (attribute.AttributeType != expectedType && !IsSubclassOf (expectedType, attribute.AttributeType))
+		var expectedType = ConvertTypeToMeta (typeof (T), provider);
+		if (attribute.GetAttributeType () != expectedType && !IsSubclassOf (expectedType, attribute.GetAttributeType ()))
 			return Enumerable.Empty<T> ();
 
-		System.Type attribType = ConvertType (attribute.AttributeType, provider);
+		System.Type attribType = ConvertTypeFromMeta (attribute.GetAttributeType (), provider);
```
rolfbjarne, line 297: The difference between meta types and normal types is going to be more confusing now, because they're both different instances of System.Type 😒 I don't see a way around this though, except for a major refactoring of the generator where we only inspect attributes, never execute them.

src/btouch.cs
filipnavara: Likely not. I tried to keep the code changes to minimum.

src/generator-attribute-manager.cs
filipnavara: printf-debugging 👼

src/generator-attribute-manager.cs
filipnavara: Unfortunately yes. I was tempted to do some refactoring on this but in the end I decided to keep the changes to minimum for the proof of concept.

rolfbjarne:
build

src/btouch.cs
filipnavara: Removing the explicit load leads to `error BI1055: bgen: Internal error: failed to convert type 'BaseTypeAttribute, bgen, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.`. I'll investigate it but I reverted the removal for now.

rolfbjarne:
build

spouliot (commented):
looks good, waiting for the generator diff :)

filipnavara:
```
15:29:54      1>/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/xbuild/Microsoft/NuGet/Microsoft.NuGet.targets(186,5): error : Your project does not reference ".NETFramework,Version=v4.6" framework. Add a reference to ".NETFramework,Version=v4.6" in the "TargetFrameworks" property of your project file and then re-run NuGet restore. [/Users/builder/jenkins/workspace/xamarin-macios-pr-builder/src/generator.csproj]
```

Hmm. I will have to dig up where this comes from. The referenced NuGet package requires at least .NET 4.6.1 target (or .NET Standard) so I had to bump the version number in the project.

filipnavara:
I think I fixed the issue causing the last build failure.

filipnavara:
Thanks for running the pipeline. This is significantly closer to what I would expect. I'll check tomorrow the failed tests locally. API diff looks promising.

filipnavara:
Tests pass locally for me now.

src/generator.cs
```diff
@@ -8000,9 +7999,10 @@ object GetDefaultValue (MethodInfo mi)
 		if (mi.ReturnType.IsEnum) {
 			if (def is string)
 				return def;
-			var name = mi.ReturnType.GetEnumName (def);
+			var name = Enum.GetName (mi.ReturnType, def);
 			if (string.IsNullOrEmpty (name)) {
-				return "(" + mi.ReturnType.FullName + ") " + def;
+				// The value could be negative so it need to be enclosed in parenthesis
+				return "(" + mi.ReturnType.FullName + ") (" + def + ")";
```
filipnavara, line 8005: There's combination of Mono bug and BGen bug. Mono fails to lookup the name from negative integer values. BGen generates invalid C# code when negative integers are present.

`DefaultValue (UIModalPresentationStyle.None)` used to take the `else` path here and end up as `return UIModalPresentationStyle.None!;` in the generated code.

With the Mono bug it doesn't find the name in the enumeration and it was converted to `return (UIKit.UIModalPresentationStyle) -1!` which is invalid C# code because of the negative number.

The new code generates `return (UIKit.UIModalPresentationStyle) (-1)!` which is obviously not ideal but at least it's valid C#.

The bug could also be triggered with current code by using `DefaultValue` attribute with a negative value not present in the enum.

filipnavara:
Can someone copy me the generator diff to GitHub Gist please? (No rush, whenever you get to it)

rolfbjarne:
@filipnavara it looks good: https://gist.github.com/rolfbjarne/d59487762eea2628da84f2e6e6aecfbf

src/btouch.cs
```diff
@@ -682,3 +625,40 @@ static void FixRelativeReferences (string sdk_path, List<string> references)
 		}
 	}
 }
+
+class SearchPathsAssemblyResolver : MetadataAssemblyResolver
+{
+	private readonly string[] libraryPaths;
+	private readonly string[] references;
```
rolfbjarne, line 632: This is really minor, but code style - we don't use `private` when it's the default (Mono Coding Guidelines):

```suggestion
	readonly string[] libraryPaths;
	readonly string[] references;
```

src/btouch.cs
```diff
@@ -682,3 +625,40 @@ static void FixRelativeReferences (string sdk_path, List<string> references)
...
+		string? name = assemblyName.Name;
+		if (name != null) {
+			foreach (var asm in context.GetAssemblies ()) {
+				if (asm.GetName ().Name == name)
+					return asm;
+			}
+
+			string dllName = name + ".dll";
+			foreach (var libraryPath in libraryPaths) {
+				string path = Path.Combine(libraryPath, dllName);
```
rolfbjarne, line 651: Code style - space before opening parenthesis:

```suggestion
				string path = Path.Combine (libraryPath, dllName);
```

rolfbjarne (approved):

src/generator-attribute-manager.cs
```diff
@@ -225,7 +224,7 @@ Type ConvertType (System.Type type, ICustomAttributeProvider provider)
 			// Report a warning if we find the same type in multiple assemblies though.
 			var assemblies = BindingTouch.universe.GetAssemblies ();
 			foreach (var asm in assemblies) {
-				var lookup = asm.FindType (new TypeName (type.Namespace, type.Name));
+				var lookup = asm.GetType (type.Namespace + "." + type.Name);// .FindType (new TypeName (type.Namespace, type.Name));
```
spouliot, line 227: let's remove the commented part

spouliot (approved):
minor

src/generator-attribute-manager.cs
filipnavara, line 227:
```suggestion
				var lookup = asm.GetType (type.Namespace + "." + type.Name);
```

filipnavara: fixed

dalexsoto:
@filipnavara do you mind if we rebase your PR on top of main?

dalexsoto (approved):

filipnavara:
> do you mind if we rebase your PR on top of main?

Not a problem.

dalexsoto:
Thanks @filipnavara hopefully this will be the last run and we can finally merge it, thanks a lot for your patience and contribution it is truly appreciated it 👍

rolfbjarne:
Great work, thanks a lot for this!! 🎉

filipnavara:
Thanks for bearing with me on this one. Only took me half a year to push it to finish line! 😅

PR #9867, opened 2020-10-14
[Tests] Fix a test that crashes on devices due to possible bonjour services present.

mandel-macaque:


mandel-macaque:
@monojenkins backport xcode12.1

tests/monotouch-test/Network/NWBrowserTest.cs
```diff
@@ -73,6 +73,8 @@ public void TestStart ()
 		[Test]
 		public void TestStateChangesHandler ()
 		{
+			if (Runtime.Arch == Arch.DEVICE)
+				Assert.Ignore ("This test makes connection assumptions and can crash the device tests.");
```
rolfbjarne, line 77: Is there a bug report? How does it crash?

dalexsoto (approved):

tests/monotouch-test/Network/NWBrowserTest.cs
mandel-macaque: It crashes only on local devices if there is a bonjour service present. The issue is that there is an assert in the delegate, in simulator, there are no issues, on a device you might get an assert failing and will raise an exception in the delegate that crashes the app. Could be that you do have bonjour services and the assumptions are wrong.

spouliot (approved):

tests/monotouch-test/Network/NWBrowserTest.cs
rolfbjarne: Then two things should be done:

1. If an assert (exception) happens on a background thread, that exception must be marshaled to the main thread. We do this already in many other tests. This will fix the crash (but the test will still fail).
2. The failing assert should be fixed to cope with this scenario, instead of just ignoring the test.

rolfbjarne (changes requested):

mandel-macaque:
@rolfbjarne I would land this to match main, then fix the test properly in main, rather than fix it in xcode12. This is a backport from main.

rolfbjarne (approved):

tests/monotouch-test/Network/NWBrowserTest.cs
mandel-macaque: Added an issue to not forget: https://github.com/xamarin/xamarin-macios/issues/9872

PR #9871, opened 2020-10-14
[tests][intro] Detect cases where [Unavailable] is mis-used

spouliot:
For each platform the presence of `[Unavailable]` should mean there are
no other availability (introduced, deprecated or obsoleted) attributes
on the same member.

Also check if the type is unavailable. In that case no member should,
for that platform, have other availability attributes.

Also fix failures - all in WatchKit which was removed from iOS.

### DIFF
--- a/src/WatchKit/iOS/WKAccessibility.cs
+++ b/src/WatchKit/iOS/WKAccessibility.cs
@@ -5,7 +5,7 @@ using System.ComponentModel;
 using ObjCRuntime;
 
 namespace WatchKit {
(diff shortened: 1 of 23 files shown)


mandel-macaque (approved):

whitneyschmidt (approved):

tj-devel709 (approved):

spouliot:
build

dalexsoto (approved):

rolfbjarne (approved):

PR #9877, opened 2020-10-14
[xcode12.2] Initial commit for Xcode 12.2 Beta 3

dalexsoto:


tests/common/TestRuntime.cs
```diff
@@ -177,13 +177,22 @@ public static bool CheckExactXcodeVersion (int major, int minor, int beta = 0)
 			watchOS = new { Major = 7, Minor = 1, Build = "18R5561" },
 		};
 
+		var twelvedot2b3 = new {
+			Xcode = new { Major = 12, Minor = 2, Beta = 3 },
+			iOS = new { Major = 14, Minor = 2, Build = "18B5072" },
+			tvOS = new { Major = 14, Minor = 2, Build = "18K5047" },
+			macOS = new { Major = 11, Minor = 0, Build = "?" },
```
whitneyschmidt, line 184: These tuples (records?) seem to point to the version info. Is it valid to have `Build = "?"`?

whitneyschmidt (approved):

tests/common/TestRuntime.cs
dalexsoto: yep when '?' is used we throw an exception IIRC

spouliot (approved):

tests/common/TestRuntime.cs
rolfbjarne: Yes, "?" basically means something like "don't know, don't care, fix me if that changes"

rolfbjarne (approved):

dalexsoto:
Mac failures seem to be related to this PR https://github.com/xamarin/xamarin-macios/pull/9868

mandel-macaque:
Will fix it, I'm wrong

dalexsoto:
We'll get fixes for macOS in a different PR

PR #9885, opened 2020-10-15
[system-dependencies] Only use the locally installed .NET version.

rolfbjarne:
We'll soon need to install files into the dotnet directory, and we don't want
to do that to the system dotnet. So just always use a locally installed .NET.


spouliot (approved):

whitneyschmidt (approved):

mandel-macaque (approved):

tj-devel709 (approved):

spouliot (approved):
even better on 2nd commit :D

rolfbjarne:
Api comparison failed because of the diff between this commit and the previous one isn't compatible with how the api comparison is implemented (this is a known issue with the api comparison).

PR #9889, opened 2020-10-15
[msbuild] Updating LocalizationStringTest

tj-devel709:
When I updated the localization string translations from the loc team previously, I made a little test to go with it that only tested one specific error code which was helpful in making sure that this error was being fired in different languages based on someone's locale. I thought this would be even more useful if we could select which error code is firing so that if I update a localization string, I can make sure it is being translated correctly and not firing in English in other locales. I did co-write a wiki page about how to add localization strings and had a part at the bottom about how to test and thought I would replace that with instructions to using this test: https://github.com/xamarin/maccore/wiki/Localization

Closed original PR: https://github.com/xamarin/xamarin-macios/pull/9855 to not utilize any tasks.

### DIFF
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -5,41 +5,46 @@ using NUnit.Framework;
 using System.Globalization;
 using System.Threading;
 using System.Collections.Generic;
+using Xamarin.Localization.MSBuild;
 
 namespace Xamarin.iOS.Tasks {
 	[TestFixture]
 	public class LocalizationStringTest : TestBase {
-		[TestCase ("cs-CZ", "došlo k chybě: neznámý formát image")]
-		[TestCase ("de-DE", "Unbekanntes Imageformat.")]
-		[TestCase ("en-US", "Unknown image format.")]
-		[TestCase ("es-ES", "formato de imagen desconocido.")]
-		[TestCase ("fr-FR", "format d'image inconnu.")]
-		[TestCase ("it-IT", "Formato immagine sconosciuto.")]
-		[TestCase ("ja-JP", "の読み込みでエラーが発生しました: 画像の形式が不明です。")]
-		[TestCase ("ko-KR", "을(를) 로드하는 동안 오류 발생: 알 수 없는 이미지 형식입니다.")]
-		[TestCase ("pl-PL", "nieznany format obrazu.")]
-		[TestCase ("pt-BR", "formato de imagem desconhecido.")]
-		[TestCase ("ru-RU", "неизвестный формат изображения.")]
-		[TestCase ("tr-TR", "yüklenirken hata oluştu: Görüntü biçimi bilinmiyor.")]
-		[TestCase ("zh-CN", "时出错: 未知图像格式")]
-		[TestCase ("zh-TW", "時發生錯誤: 未知的映像格式。")]
-		public void AllSupportedTranslations (string culture, string errorMessage)
+		[TestCase ("cs-CZ")]
+		[TestCase ("de-DE")]
+		[TestCase ("es-ES")]
+		[TestCase ("fr-FR")]
+		[TestCase ("it-IT")]
+		[TestCase ("ja-JP")]
+		[TestCase ("ko-KR")]
+		[TestCase ("pl-PL")]
+		[TestCase ("pt-BR")]
+		[TestCase ("ru-RU")]
+		[TestCase ("tr-TR")]
+		[TestCase ("zh-CN")]
+		[TestCase ("zh-TW")]
+		public void SpecificErrorTranslation (string culture)
 		{
+			// insert which error code you'd like to test
+			string errorCode = "E0007";
 			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
-			CultureInfo newCulture;
-			try {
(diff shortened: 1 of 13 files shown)


whitneyschmidt (approved):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -5,41 +5,46 @@
...
 			} finally {
 				Thread.CurrentThread.CurrentUICulture = originalCulture;
 			}
 		}
+
+		private string TranslateError (string culture, string errorCode)
+		{
+			CultureInfo cultureInfo = new CultureInfo (culture);
+			Thread.CurrentThread.CurrentUICulture = cultureInfo;
+			PropertyInfo propertyInfo = typeof(MSBStrings).GetProperty(errorCode);
```
spouliot, line 46: same

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -5,41 +5,46 @@
...
-				var task = CreateTask<CollectITunesArtwork> ();
-				task.ITunesArtwork = new TaskItem [] { new TaskItem (Assembly.GetExecutingAssembly ().Location) };
+				string englishError = TranslateError("en-US", errorCode);
+				string newCultureError = TranslateError(culture, errorCode);
 
-				Assert.IsFalse (task.Execute (), "Execute failure");
-				Assert.AreEqual (1, Engine.Logger.ErrorEvents.Count, "ErrorCount");
-				bool isTranslated = Engine.Logger.ErrorEvents[0].Message.Contains (errorMessage);
-				Assert.IsTrue (isTranslated, culture + ": is not supported correctly. ");
+				Assert.AreNotEqual (englishError, newCultureError, $"\"{errorCode}\" is not translated in {culture}.");
```
spouliot, line 36: this will not fail if the translation string is null or empty
is there something (in the code) that ensure the English (default) is returned if no translation is available ?

spouliot (changes requested):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -5,41 +5,46 @@
...
+		[TestCase ("pl-PL")]
+		[TestCase ("pt-BR")]
+		[TestCase ("ru-RU")]
+		[TestCase ("tr-TR")]
+		[TestCase ("zh-CN")]
+		[TestCase ("zh-TW")]
+		public void SpecificErrorTranslation (string culture)
 		{
+			// insert which error code you'd like to test
+			string errorCode = "E0007";
```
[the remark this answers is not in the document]
tj-devel709: Hi Sebastien, I did not explain my motivation very well. I am very open to feedback on how to make this better here and in future tests! When I updated the localization strings translations from the loc team previously, I made a little test to go with it that only tested one specific error code which was helpful in making sure that this error was being fired in different languages based on someone's locale. I thought this would be even more useful if we could select which error code is firing so that if I update a localization string, I can make sure it is being translated correctly and not firing in English in other locales. I did co-write a wiki page about how to add localization strings and had a part at the bottom about how to test and thought I would replace that with instructions to this test unless there is a better way to do this: https://github.com/xamarin/maccore/wiki/Localization

About the user having to manually enter the code: I had thought about trying to iterate through all the error codes and see if they are translated making this more of a automatic test, but the loc team can take up to a week to do the translations after someone adds new error codes. Perhaps this could be a good way to make sure people update their translated strings, but it would slow down someone from making a PR up to a week which I did not think was a good idea.. 

Please let me know if anyone has any recommendations?

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: Very embarrassed that I didn't check for null or empty strings but that and handling non-existent error codes should be taken care of now!

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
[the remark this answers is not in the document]
spouliot: Part of your reply should be copied in the PR description (and in the commit message before merging).

>  Perhaps this could be a good way to make sure people update their translated strings,

Yes, that does sound like a good idea.

> but it would slow down someone from making a PR up to a week 

Maybe not. We can have a list of _pending_ translations and only report errors if they are not part of the list. That would not block PR, the authors would add the codes to the _pending_ list. We can also report (without error) pending members that were translated, so anyone updated the list would be able to keep it up to date (a bit like xtro force you to keep the data file clean).

That can be discussed (in a future weekly call) and implemented in a different PR.

> Also don't mind keeping the old test and adding this as another test.

I'd rather keep the existing test - it works so it's free 😄

spouliot (approved):

tj-devel709:
* accidentally pulled xamarin/main. Trying to revert merge. Added do-not-merge for now

tj-devel709:
Okay reverted the merge.

rolfbjarne (approved):

mandel-macaque (approved):

PR #9893, opened 2020-10-16
Bump to .NET 5.0.100-rtm.20509.5.

rolfbjarne:


rolfbjarne:
Api comparison failed because of the diff between this commit and the previous one isn't compatible with how the api comparison is implemented (this is a known issue with the api comparison).

spouliot (approved):
RTMFTW 🎉

PR #9901, opened 2020-10-16
[Tests][Network] Do not really check the value, just make sure that we are called.

mandel-macaque:


dalexsoto (approved):

tests/monotouch-test/Network/NWBrowserTest.cs
```diff
@@ -102,17 +99,9 @@ public void TestStateChangesHandler ()
...
 					try {
-						if (firstRun) {
-							Assert.IsNull (oldResult, "oldResult first run.");
-							Assert.IsNotNull (newResult, "newResult first run");
-							firstRun = false;
-						} else {
-							Assert.IsNotNull (oldResult, "oldResult first run.");
-							Assert.IsNull (newResult, "newResult first run");
-						}
+						Assert.IsTrue (oldResult != null || newResult != null);
```
spouliot, line 104: That's not correct* The assert will crash the test app since it fail outside the main thread.
The result should be kept in a local variable and checked after the `WaitOne`

`*` I had a similar issue on the above `SetStateChangesHandler` a while ago (need to find where my diff is hiding).

spouliot (changes requested):

tests/monotouch-test/Network/NWBrowserTest.cs
```diff
@@ -85,8 +83,7 @@ public void TestStateChangesHandler ()
 			// 5. The browser is not yet canceled, so it picks up that the service/listener is not longer then and returns it.
 			// 
 			// The test will block until the different events are set by the callbacks that are executed in a diff thread.
-
-			bool firstRun = true;
+			bool didRan = false;
```
whitneyschmidt, line 86:
```suggestion
			bool didRun = false;
```

whitneyschmidt (approved):
Tiny nit re naming, doesn't need to be fixed 🙈

tj-devel709 (approved):

rolfbjarne (approved):

spouliot (approved):

mandel-macaque:
@monojenkins backport d16-8

mandel-macaque:
@monojenkins backport xcode12.1

Issue #9902, opened 2020-10-16
NWBrowserTest.cs(76,24): error CS0103: The name 'Arch' does not exist in the current context


spouliot:
### Steps to Reproduce

1. Build a PR on `xcode12.2`

### Expected Behavior

Success

### Actual Behavior

xammac_tests fails to build because of https://github.com/xamarin/xamarin-macios/pull/9868

* xammac tests/Mac Modern/Debug: BuildFailure
* xammac tests/Mac Modern/Release: BuildFailure
* xammac tests/Mac Modern/Release: BuildFailure

### Environment

`xcode12.2`

spouliot:
The original PR was merged because it _likely_ did not trigger the execution of `xammac_tests` when the test was was updated.

spouliot:
xammac tests build failures due to https://github.com/xamarin/xamarin-macios/issues/9902

spouliot:
xammac tests build failure due to https://github.com/xamarin/xamarin-macios/issues/9902

spouliot:
https://github.com/xamarin/xamarin-macios/pull/9920 fixed the issue and it did no show up in the latest PR

PR #9918, opened 2020-10-19
[fileprovider] Update for Xcode 12.2 beta 3

spouliot:
note: this is not a breaking change (even if it looks like one, even in
API diff) since there is a `[NoMac]` in `d16-8-xm` that was removed (and
never released for XM) in `xcode12`.


dalexsoto (approved):

spouliot:
xammac tests build failure due to https://github.com/xamarin/xamarin-macios/issues/9902

PR #9935, opened 2020-10-21
[tests][xcode12.2] Fix xammac tests

dalexsoto:
xammac tests began to fail because Xcode 12.0 includes macOS 10.15 SDK so our tests needed to be adjusted for asking Xcode 12.2 when running in the Xamarin.Mac context also added an implementation for `CheckExactXcodeVersion` for macOS which is somewhat dangerous (?) because Apple docs says `OperatingSystemVersionString` is not appropriate for parsing but ¯\\(◉‿◉)/¯ this helps us to add some 🦜 tests for future betas.

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -236,7 +236,7 @@ partial class TestRuntime
 			/*
 			 * I could be parsing the string but docs says it is not suitable for parsing and this is ugly enough so
 			 * an apology in advance (I'm very sorry =]) to my future self or whoever is dealing with this if it broke
-			 * but there are no better at this time. That said this is good enough for the current use case.
+			 * but there are no better solutions at this time. That said this is good enough for the current use case.
 			 * Example: Version 10.16 (Build 20A5395g)
 			 *
 			 * The above statement also applies to 'CheckExactmacOSSystemVersion' =S
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -181,7 +181,7 @@ partial class TestRuntime
 			Xcode = new { Major = 12, Minor = 2, Beta = 3 },
 			iOS = new { Major = 14, Minor = 2, Build = "18B5072" },
 			tvOS = new { Major = 14, Minor = 2, Build = "18K5047" },
-			macOS = new { Major = 11, Minor = 0, Build = "?" },
+			macOS = new { Major = 11, Minor = 0, Build = "20A5395" },
 			watchOS = new { Major = 7, Minor = 1, Build = "18R5572" },
 		};
 
@@ -219,6 +219,20 @@ partial class TestRuntime
 			var actual = GetiOSBuildVersion ();
 			Console.WriteLine (actual);
 			return actual.StartsWith (v.tvOS.Build, StringComparison.Ordinal);
+#elif __MACOS__
+			if (!CheckExactmacOSSystemVersion (v.macOS.Major, v.macOS.Minor))
+				return false;
+			if (v.macOS.Build == "?")
+				throw new NotImplementedException ($"Build number for macOS {v.macOS.Major}.{v.macOS.Minor} beta {beta}.");
+			/*
+			 * I could be parsing the string but docs says it is not suitable for parsing and this is ugly enough so
+			 * an apology in advace (I'm very sorry =]) to my future self or whoever is dealing with this if it broke
+			 * but there are no better at this time. That said this is good enough for the current use case.
+			 * Example: Version 10.16 (Build 20A5395g)
+			 *
+			 * The above statement also applies to 'CheckExactmacOSSystemVersion' =S
+			 */
+			return NSProcessInfo.ProcessInfo.OperatingSystemVersionString.Contains (v.macOS.Build, StringComparison.Ordinal);
 #else
 			throw new NotImplementedException ();
 #endif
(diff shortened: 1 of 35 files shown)


tests/common/TestRuntime.cs
```diff
@@ -219,6 +219,20 @@ public static bool CheckExactXcodeVersion (int major, int minor, int beta = 0)
...
 			Console.WriteLine (actual);
 			return actual.StartsWith (v.tvOS.Build, StringComparison.Ordinal);
+#elif __MACOS__
+			if (!CheckExactmacOSSystemVersion (v.macOS.Major, v.macOS.Minor))
+				return false;
+			if (v.macOS.Build == "?")
+				throw new NotImplementedException ($"Build number for macOS {v.macOS.Major}.{v.macOS.Minor} beta {beta}.");
+			/*
+			 * I could be parsing the string but docs says it is not suitable for parsing and this is ugly enough so
+			 * an apology in advace (I'm very sorry =]) to my future self or whoever is dealing with this if it broke
```
rolfbjarne, line 229:
```suggestion
			 * an apology in advance (I'm very sorry =]) to my future self or whoever is dealing with this if it broke
```

tests/monotouch-test/AudioUnit/AudioUnitTest.cs
```diff
@@ -53,7 +53,11 @@ public void GetElementCount ()
 		[Test]
 		public void CopyIconTest ()
 		{ 
+#if __MACOS__
+			TestRuntime.AssertXcodeVersion (12, 2);
```
rolfbjarne, line 57: The whole point of `AssertXcodeVersion` was to avoid adding `#if`s everywhere...

and don't think this is needed though, wouldn't it be enough to change `AssertXcodeVersion` to map macOS 10.16/11.0 to Xcode 12.2?

dalexsoto: I agree, but it does not really reflect the truth specially if you compare it with CheckExactXcodeVersion or anywhere else, this is very likely the only time we’ll need to do it so my take is we can live with some extra if/defs this time in favor of consistency

tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
```diff
@@ -103,7 +104,11 @@ public void SetCaption ()
 					Assert.Null (e, "ro / set / no error"); // weird, it can't be saved back to the file metadata
 
 					var s = MAImageCaptioning.GetCaption (url, out e);
+#if __MACOS__
+					if (TestRuntime.CheckXcodeVersion (12, 2)) {
+#else
 					if (TestRuntime.CheckXcodeVersion (12, 0)) {
+#endif
```
spouliot, line 111: minor: since this happens a few time in the same file you could have a single
```
#if __MACOS__
const int MinorXcode11Mismatch = 2;
#else
const int MinorXcode11Mismatch = 0;
#endif
```
and use that constant in the checks (no more `#ifdef` for that file).

It could also be added into `TestRuntime` so it would be usable in all files.

spouliot (approved):

tests/monotouch-test/MediaAccessibility/ImageCaptioningTest.cs
dalexsoto: Oh, I like this better it keeps the intent and truth, will re-work on top of this idea 👍

rolfbjarne: Apple 😡

Adding something like `TestRuntime.CheckApiAddedInSummer2020 ()` could work (might need a better name...), and that would avoid the #ifs too.

mandel-macaque (approved):

spouliot (approved):
👍

tests/monotouch-test/GameKit/LeaderboardViewControllerTest.cs
```diff
@@ -31,6 +31,13 @@ public class LeaderboardViewControllerTest {
 		[Test]
 		public void DefaultCtor ()
 		{
+#if MONOMAC
+			// For some reason the init method is not allowed on Xcode 12.2 Beta 3 anymore
+			// but was a llowed before that said this class got deprecated in 10.10 so it may now be
```
rolfbjarne, line 36:
```suggestion
			// but was allowed before that said this class got deprecated in 10.10 so it may now be
```

tests/monotouch-test/Vision/VNRequestTests.cs
```diff
@@ -136,8 +136,12 @@ public void VNSupportedRevisionsUnsupportedTest ()
 				Assert.That (faceObservation.BoundingBox, Is.EqualTo (rect));
 
 				var recognizedObjectObservation = VNRecognizedObjectObservation.FromBoundingBox (VNRecognizedObjectObservationRequestRevision.Unspecified, rect);
+#if __MACOS__
+				if (TestRuntime.CheckXcodeVersion (11, 0) && !TestRuntime.CheckXcodeVersion (12, 2)) {
```
rolfbjarne, line 140: Doesn't `TestRuntime.MinorXcode12APIMismatch` work in this file too?

tests/monotouch-test/MLCompute/MLEnumsTest.cs
```diff
@@ -16,8 +16,10 @@ public class MLEnumsTest {
 		[SetUp]
 		public void SetUp ()
 		{
+#if __MACOS__
+			TestRuntime.AssertXcodeVersion (12, 2);
```
rolfbjarne, line 20: Doesn't `TestRuntime.MinorXcode12APIMismatch` work in this file too?

rolfbjarne (approved):

tests/monotouch-test/Vision/VNRequestTests.cs
dalexsoto: Yep, another commit is coming for the missing cases I have

whitneyschmidt (approved):

tj-devel709 (approved):

dalexsoto:
build

PR #9953, opened 2020-10-23
[tools/dotnet] Re-use the existing main generation code in our dotnet code.

rolfbjarne:
Move/refactor the code to generate the main function to make it re-usable from
our dotnet code (and then use it there as well).

This PR is probably best reviewed commit-by-commit and ignoring whitespace changes.


tools/common/Application.cs
```diff
@@ -1104,5 +1184,49 @@ public void SetObjectiveCExceptionMode ()
...
+					return true;
+				else if (opt [0] == '-' && opt.Substring (1) == assembly)
+					return false;
+			}
+
+			// There's an implicit 'all' at the start of the list.
+			return true;
+		}
+
+		public bool IsAOTCompiled (string assembly)
```
spouliot, line 1212: there's likely a context here for the call to return the correct value
IOW we support JIT (Mac / sim) which are neither interpreted nor AOTed

spouliot: it comes from `Application.mtouch.cs` and while it was already not great (sim) it's even more confusing now that it's shared with Mac.

rolfbjarne: I'm not quite sure which changes you want here: I can add comments to the code, and/or add/handle different cases (i.e. add a conditional and return `false` for Xamarin.Mac for instance). Or did you have anything else in mind?

spouliot: Not 💯% sure myself :-) I'll review the PR again once I complete today's monitoring. The main problem is that we're merging code from two tools into a single one. That's fine except when the code is specific to one product since we lose this _limitation_.

* minimally updated comments so that code does not get reused for XM at some point
* maybe a `#if MONOMAC throw NotImplementedException (); #endif` so the comment can't be missed
* ideally something more OO would be used - but that would be larger changes :(

spouliot (approved):

tools/common/Application.cs
```diff
@@ -108,6 +110,119 @@ public partial class Application
...
+
+				if (Embeddinator) {
+					return AssemblyBuildTarget.StaticObject;
+				} else if (HasFrameworks || UseMonoFramework.Value) {
+					return AssemblyBuildTarget.Framework;
+				} else if (HasDynamicLibraries) {
+					return AssemblyBuildTarget.DynamicLibrary;
+				} else {
+					return AssemblyBuildTarget.StaticObject;
+				}
```
mandel-macaque, line 150: We have this if also in LibXamarinLinkMode, I think we could share it so that we do not have mismatching results in the future.

tools/common/Target.cs
```diff
@@ -501,5 +501,352 @@ public void LoadSymbols ()
...
+
+		static string EncodeAotSymbol (string symbol)
+		{
+			var sb = new StringBuilder ();
+			/* This mimics what the aot-compiler does */
+			foreach (var b in System.Text.Encoding.UTF8.GetBytes (symbol)) {
+				char c = (char) b;
+				if ((c >= '0' && c <= '9') ||
+					(c >= 'a' && c <= 'z') ||
+					(c >= 'A' && c <= 'Z')) {
```
mandel-macaque, line 828: Would Char.IsLetter and Char.IsNumber work for this?

tools/mtouch/mtouch.cs
```diff
@@ -294,255 +294,6 @@ public static IList<string> GetAotArguments (Application app, string filename, A
...
-		{
-			var sb = new StringBuilder ();
-			/* This mimics what the aot-compiler does */
-			foreach (var b in System.Text.Encoding.UTF8.GetBytes (symbol)) {
-				char c = (char) b;
-				if ((c >= '0' && c <= '9') ||
-					(c >= 'a' && c <= 'z') ||
-					(c >= 'A' && c <= 'Z')) {
-					sb.Append (c);
-					continue;
```
mandel-macaque, line 307: Well we are moving code, maybe chaging to Char.Is* methods is not a good idea since we already had this logic.

mandel-macaque (approved):
Minor comments that can be taken or not.

tools/common/Target.cs
rolfbjarne: This has to match the corresponding implementation in the AOT compiler, which means it's not trivial to change (especially since the AOT compiler is C code).

tools/common/Application.cs
rolfbjarne: The implementations for LibMonoLinkMode and LibXamarinLinkMode aren't identical.

PR #9954, opened 2020-10-23
[dotnet-linker] Catch any exceptions from our custom steps and show them using our error reporting logic.

rolfbjarne:
* Letting the linker handle the exceptions will not result in a particularly
  good experience, because the linker will crash.

* We can also show better information, since we have more knowledge about many
  of the exceptions we raise ourselves.


tools/dotnet-linker/Steps/ConfigurationAwareStep.cs
```diff
@@ -21,5 +23,50 @@ protected void Report (List<Exception> exceptions)
...
+
+		protected sealed override void Process ()
+		{
+			try {
+				TryProcess ();
+			} catch (Exception e) {
+				Report (e);
+				throw;
+			}
+		}
```
spouliot, line 40: might want to share code/pattern with `tools/linker/ExceptionalSubStep.cs` since it faced a similar issue

tools/dotnet-linker/Steps/ConfigurationAwareSubStep.cs
```diff
@@ -32,6 +33,111 @@ protected void Report (List<Exception> exceptions)
 			// We can't just throw an exception or exit here, since there might be only warnings in the list of exceptions.
 			ErrorHelper.Show (exceptions);
 		}
+
```
spouliot, line 36: that one should subclass `ExceptionalSubStep`

spouliot (changes requested):

tj-devel709 (approved):

mandel-macaque (approved):
+1 to @spouliot comments.

spouliot (approved):

PR #9955, opened 2020-10-23
[tools] Renumber RemoveRejectedTypesStep's error code base to match other ExceptionalSubStep subclasses.

rolfbjarne:
Also document which numbers each subclass uses in Errors.resx. This makes it
much easier to figure out the next number to use.


tj-devel709:
Good idea! I was wondering how to select a number for a new code, but this answers that question!

tj-devel709 (approved):

spouliot (approved):

PR #9957, opened 2020-10-23
Exclude yaml-pipeline branch from building

mjbond-msft:
We need a branch directly in the repo and not as a fork to build & validate a new yaml build & test pipeline being developed. The branch effectively serves as the main branch for the yaml pipeline development effort with forked branches feeding into it. This change effectively prevents builds of the `yaml-pipeline` branch on Jenkins; though builds for the branch will trigger there, but they will immediately exit.


rolfbjarne (commented):
I don't think you need this in `main`, only in your `yaml-branch`.

mandel-macaque:
I agree with rl, we could add the exception there. Re-basing :)

mandel-macaque (approved):

rolfbjarne (approved):

PR #9959, opened 2020-10-26
[tools] Localize error messages from the ExceptionSubStep linker step.

rolfbjarne:
I also changed the error message slightly to hopefully make it a bit more
comprehensible when translated (since the step name won't be translated, we'll
end up with a message that mixes English with the translated string).


spouliot (approved):

rolfbjarne:
build

mandel-macaque (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2154)

PR #9971, opened 2020-10-26
[msbuild] Localization String Test for All Strings

tj-devel709:
This test will run through all available strings in MSBStrings.Designer.cs in the each of the supported languages.
If any strings should fire in other languages but are still in english, an error will be thrown unless that error code is in an ignore file.

There are two types of ignore files introduced: 
- `LocalizationIgnore/common-Translations.ignore`
    - for error codes that are not yet translated in all languages
- `LocalizationIgnore/<LOCALE>-Translations.ignore`
    - for error codes that are not yet translated in the specific locale

If an error code is in an ignore file and it turns out it is translated, an error will be thrown to remove the error code from the ignore file.

Wiki will be updated for the team to reflect this test: https://github.com/xamarin/maccore/wiki/Localization

*open to feedback on new directory and ignore file names/format

### DIFF
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
@@ -1,5 +1,3 @@
 # Insert Error codes that are waiting on translations below (one per line)
-
 W0111
-E7069 
-
+E7069
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
@@ -1,0 +1,5 @@
+# Insert Error codes that are waiting on translations below (one per line)
+
+W0111
+E7069 
+
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
@@ -1,3 +1,4 @@
 # Insert Error codes that are waiting on translations below (one per line)
 W0111
 E7069
+E7070
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -86,7 +86,7 @@ namespace Xamarin.iOS.Tasks {
 			return (string) propertyInfo.GetValue (null, null);
 		}
 
-		List<string> commonIgnoreList { get; set; } = null;
+		List<string> commonIgnoreList = null;
 
 		[SetUp]
 		public void SetUp ()
@@ -144,8 +144,8 @@ namespace Xamarin.iOS.Tasks {
 			}
 
 			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
-			Assert.IsEmpty (cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid. Please remove invalid codes");
-			Assert.IsEmpty (commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid. Please remove invalid codes");
+			Assert.IsEmpty (cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid or are repeated. Please remove.");
+			Assert.IsEmpty (commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid or are repeated. Please remove.");
 		}
 
 		IList<string> ReadFile (string path)
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Collections.Generic;
 using Xamarin.Localization.MSBuild;
 using System.ComponentModel;
+using System.Linq;
 
 namespace Xamarin.iOS.Tasks {
 	[TestFixture]
@@ -85,6 +86,9 @@ namespace Xamarin.iOS.Tasks {
 			return (string) propertyInfo.GetValue (null, null);
 		}
 
+		private List<string> commonIgnoreList { get; set; } = null;
+		private List<string> commonValidEntries { get; set; } = null;
+
 		[TestCase ("cs-CZ")]
 		[TestCase ("de-DE")]
 		[TestCase ("es-ES")]
@@ -102,28 +106,42 @@ namespace Xamarin.iOS.Tasks {
 		{
 			string errorCode = string.Empty;
 			string errorList = string.Empty;
-			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
-			string path = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore";
+			List<string> cultureIgnoreList = null;
+			List<string> cultureValidEntries = new List<string> ();
 
-			List<string> commonIgnoreList = ReadFile ($"{path}/common-Translations.ignore");
-			List<string> cultureIgnoreList = ReadFile ($"{path}/{culture}-Translations.ignore");
+			string fullCommonPath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore";
+			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
+			bool isCommonFileExistant = File.Exists (fullCommonPath);
+			bool isCultureFileExistant = File.Exists (fullCulturePath);
+			bool isFirstRun = false;
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;q
+
+			if (commonIgnoreList == null && isCommonFileExistant){
+				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
+				commonValidEntries = new List<string> ();
+				isFirstRun = true;
+			}
+
+			if (isCultureFileExistant)
+				cultureIgnoreList = new List<string> (ReadFile (fullCulturePath));
 
 			foreach (var errorCodeInfo in typeof (MSBStrings).GetProperties ()) {
 				try {
-					errorCode = errorCodeInfo.Name.ToString ();
+					errorCode = errorCodeInfo.Name;
 					if (errorCode == "ResourceManager" || errorCode == "Culture")
 						continue;
-
-					Assert.IsFalse (string.IsNullOrEmpty (errorCode), "Error code is null or empty");
 					string englishError = TranslateError ("en-US", errorCode);
 					string newCultureError = TranslateError (culture, errorCode);
 
-					if (commonIgnoreList.Contains (errorCode)) {
-						if (englishError != newCultureError)
-							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/common-Translations.ignore");
-					} else if (cultureIgnoreList.Contains (errorCode)) {
-						if (englishError != newCultureError)
-							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/{culture}-Translations.ignore");
+					if (isCommonFileExistant && commonIgnoreList.Contains (errorCode)) {
+						Assert.AreEqual (englishError, newCultureError, $"{errorCode} is translated. Remove {errorCode} from {shortCommonPath}");
+						if (isFirstRun)
+							commonValidEntries.Add (errorCode);
+					} else if (isCultureFileExistant && cultureIgnoreList.Contains (errorCode)) {
+						Assert.AreEqual (englishError, newCultureError, $"{errorCode} is translated. Remove {errorCode} from {shortCulturePath}");
+						cultureValidEntries.Add (errorCode);
 					} else if (englishError == newCultureError)
 						errorList += $"{errorCode} ";
 				} finally {
@@ -131,30 +149,14 @@ namespace Xamarin.iOS.Tasks {
 				}
 			}
 
-			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to ~/LocalizationIgnore/common-Translations.ignore or ~/LocaliztionIgnore/{culture}-Translations.ignore");
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
+			Assert.AreEqual (new List<string> (), cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid. Please remove invalid codes");
+			Assert.AreEqual (new List<string> (), commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid. Please remove invalid codes");
 		}
 
-		private List<string> ReadFile (string path)
+		private IEnumerable<string> ReadFile (string path)
 		{
-			List<string> ignoreList = new List<string> ();
-			string temp;
-			try {
-				using (var sr = new StreamReader (path)) {
-					while (!sr.EndOfStream){
-						if (sr.Peek () == -1)
-							break;
-						temp = sr.ReadLine ();
-						if (string.IsNullOrEmpty (temp) || temp.Trim ()[0] == '#') {
-							continue;
-						} else {
-							ignoreList.Add (temp.Trim ());
-						}
-					}
-				}
-			} catch (IOException e) {
-				Assert.IsTrue (false, $"Could not read file at: {path}");
-			}
-			return ignoreList;
+			return File.ReadAllLines (path).Where (line => !line.StartsWith ("#", StringComparison.Ordinal) && line != string.Empty);
 		}
 	}
 }
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -116,7 +116,7 @@ namespace Xamarin.iOS.Tasks {
 			bool isCommonFileExistant = File.Exists (fullCommonPath);
 			bool isCultureFileExistant = File.Exists (fullCulturePath);
 			bool isFirstRun = false;
-			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;q
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
 
 			if (commonIgnoreList == null && isCommonFileExistant){
 				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -119,16 +119,13 @@ namespace Xamarin.iOS.Tasks {
 					string newCultureError = TranslateError (culture, errorCode);
 
 					if (commonIgnoreList.Contains (errorCode)) {
-						if (englishError != newCultureError) {
+						if (englishError != newCultureError)
 							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/common-Translations.ignore");
-						}
 					} else if (cultureIgnoreList.Contains (errorCode)) {
-						if (englishError != newCultureError) {
+						if (englishError != newCultureError)
 							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/{culture}-Translations.ignore");
-						}
-					} else if (englishError == newCultureError) {
+					} else if (englishError == newCultureError)
 						errorList += $"{errorCode} ";
-					}
 				} finally {
 					Thread.CurrentThread.CurrentUICulture = originalCulture;
 				}
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -87,7 +87,6 @@ namespace Xamarin.iOS.Tasks {
 		}
 
 		private List<string> commonIgnoreList { get; set; } = null;
-		private List<string> commonValidEntries { get; set; } = null;
 
 		[TestCase ("cs-CZ")]
 		[TestCase ("de-DE")]
@@ -106,7 +105,9 @@ namespace Xamarin.iOS.Tasks {
 		{
 			string errorCode = string.Empty;
 			string errorList = string.Empty;
+			bool isFirstRun = false;
 			List<string> cultureIgnoreList = null;
+			List<string> commonValidEntries = new List<string> ();
 			List<string> cultureValidEntries = new List<string> ();
 
 			string fullCommonPath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore";
@@ -115,12 +116,10 @@ namespace Xamarin.iOS.Tasks {
 			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
 			bool isCommonFileExistant = File.Exists (fullCommonPath);
 			bool isCultureFileExistant = File.Exists (fullCulturePath);
-			bool isFirstRun = false;
 			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
 
 			if (commonIgnoreList == null && isCommonFileExistant){
 				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
-				commonValidEntries = new List<string> ();
 				isFirstRun = true;
 			}
 
@@ -137,8 +136,7 @@ namespace Xamarin.iOS.Tasks {
 
 					if (isCommonFileExistant && commonIgnoreList.Contains (errorCode)) {
 						Assert.AreEqual (englishError, newCultureError, $"{errorCode} is translated. Remove {errorCode} from {shortCommonPath}");
-						if (isFirstRun)
-							commonValidEntries.Add (errorCode);
+						commonValidEntries.Add (errorCode);
 					} else if (isCultureFileExistant && cultureIgnoreList.Contains (errorCode)) {
 						Assert.AreEqual (englishError, newCultureError, $"{errorCode} is translated. Remove {errorCode} from {shortCulturePath}");
 						cultureValidEntries.Add (errorCode);
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -86,12 +86,12 @@ namespace Xamarin.iOS.Tasks {
 			return (string) propertyInfo.GetValue (null, null);
 		}
 
-		List<string> commonIgnoreList = null;
+		IList<string> commonIgnoreList = null;
 
 		[SetUp]
 		public void SetUp ()
 		{
-			commonIgnoreList = ReadFile ($"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore") as List<string>;
+			commonIgnoreList = ReadFile ($"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore");
 		}
 
 		[TestCase ("cs-CZ")]
@@ -110,17 +110,16 @@ namespace Xamarin.iOS.Tasks {
 		public void AllErrorTranslation (string culture)
 		{
 			string errorList = string.Empty;
-			List<string> cultureIgnoreList = null;
+			IList<string> cultureIgnoreList = null;
 			List<string> commonValidEntries = new List<string> ();
 			List<string> cultureValidEntries = new List<string> ();
 
-			string fullCommonPath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore";
 			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
-			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
-			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
 			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
 
-			cultureIgnoreList = ReadFile (fullCulturePath) as List<string>;
+			cultureIgnoreList = ReadFile (fullCulturePath);
 
 			foreach (var errorCodeInfo in typeof (MSBStrings).GetProperties ()) {
 				try {
@@ -151,7 +150,7 @@ namespace Xamarin.iOS.Tasks {
 		IList<string> ReadFile (string path)
 		{
 			if (!File.Exists (path))
-				return new List<string> ();
+				return Array.Empty<string> ();
 			return File.ReadAllLines (path).Where (line => !line.StartsWith ("#", StringComparison.Ordinal) && line != string.Empty).ToList();
 		}
 	}
--- a/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
@@ -7,6 +7,7 @@ using System.Globalization;
 using System.Threading;
 using System.Collections.Generic;
 using Xamarin.Localization.MSBuild;
+using System.ComponentModel;
 
 namespace Xamarin.iOS.Tasks {
 	[TestFixture]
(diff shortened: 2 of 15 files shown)


tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+			string errorList = string.Empty;
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
+			string path = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore";
+
+			List<string> commonIgnoreList = ReadFile ($"{path}/common-Translations.ignore");
+			List<string> cultureIgnoreList = ReadFile ($"{path}/{culture}-Translations.ignore");
+
+			foreach (var errorCodeInfo in typeof (MSBStrings).GetProperties ()) {
+				try {
+					errorCode = errorCodeInfo.Name.ToString ();
```
spouliot, line 113: why the _extraneous_ `.ToString ()` call ?
looks like an `PropertyInfo` instance, so `Name` is already a `string` right ?

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+					if (errorCode == "ResourceManager" || errorCode == "Culture")
+						continue;
+
+					Assert.IsFalse (string.IsNullOrEmpty (errorCode), "Error code is null or empty");
```
spouliot, line 117: a `PropertyInfo.Name` cannot be `null` (or empty)

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+					string englishError = TranslateError ("en-US", errorCode);
+					string newCultureError = TranslateError (culture, errorCode);
+
+					if (commonIgnoreList.Contains (errorCode)) {
+						if (englishError != newCultureError)
+							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/common-Translations.ignore");
```
spouliot, line 123: I think what you're looking for is `Assert.Fail`

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+					} else if (cultureIgnoreList.Contains (errorCode)) {
+						if (englishError != newCultureError)
+							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/{culture}-Translations.ignore");
```
spouliot, line 126: same

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+						errorList += $"{errorCode} ";
+				} finally {
+					Thread.CurrentThread.CurrentUICulture = originalCulture;
+				}
+			}
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to ~/LocalizationIgnore/common-Translations.ignore or ~/LocaliztionIgnore/{culture}-Translations.ignore");
+		}
+
+		private List<string> ReadFile (string path)
```
spouliot, line 137: use `File.ReadAllLines` instead (in the method above)
and just skip lines that starts with `#`

spouliot (changes requested):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
rolfbjarne: Or `Assert.AreNotEqual (englishError, newCultureError, $"...")` - the advantage of this being that both `englishError` and `newCultureError` will show up in the logs.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+						if (englishError != newCultureError)
+							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/{culture}-Translations.ignore");
+					} else if (englishError == newCultureError)
+						errorList += $"{errorCode} ";
+				} finally {
+					Thread.CurrentThread.CurrentUICulture = originalCulture;
+				}
+			}
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to ~/LocalizationIgnore/common-Translations.ignore or ~/LocaliztionIgnore/{culture}-Translations.ignore");
```
rolfbjarne, line 134: This looks like I have to add these files to my home directory, which is not quite right.

rolfbjarne (changes requested):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+		[TestCase ("zh-TW")]
+		public void AllErrorTranslation (string culture)
+		{
+			string errorCode = string.Empty;
+			string errorList = string.Empty;
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
+			string path = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore";
+
+			List<string> commonIgnoreList = ReadFile ($"{path}/common-Translations.ignore");
+			List<string> cultureIgnoreList = ReadFile ($"{path}/{culture}-Translations.ignore");
```
mandel-macaque, line 109: I don't like the fact that we are adding empty files with a comment, I would prefer for the test to be written so that the file can be missing and not fail.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+						continue;
+
+					Assert.IsFalse (string.IsNullOrEmpty (errorCode), "Error code is null or empty");
+					string englishError = TranslateError ("en-US", errorCode);
+					string newCultureError = TranslateError (culture, errorCode);
+
+					if (commonIgnoreList.Contains (errorCode)) {
+						if (englishError != newCultureError)
+							Assert.IsFalse (true, $"{errorCode} is translated. Remove {errorCode} from ~/LocalizationIgnore/common-Translations.ignore");
+					} else if (cultureIgnoreList.Contains (errorCode)) {
```
mandel-macaque, line 124: We are doing the "Contains" for the same error twice. This is not needed. Rather than using two lists, use a single one that is the union of the common error and the culture specific one. That way, we remove the two contains calls and also the twi different assertions. We just need this to be done once.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+						if (string.IsNullOrEmpty (temp) || temp.Trim ()[0] == '#') {
+							continue;
+						} else {
+							ignoreList.Add (temp.Trim ());
+						}
+					}
+				}
+			} catch (IOException e) {
+				Assert.IsTrue (false, $"Could not read file at: {path}");
+			}
```
mandel-macaque, line 156: +1 on @spouliot comment. If you wanted to use a Stream, and for memory reasons go line by line, this is an overkill. You can do the following:
```csharp
using (var reader = new StreamReader (path)) {
	string line;
	while ((line = reader.ReadLine ()) != null) {
		if (line.StartsWith ("#"))
				continue;
		else
				ignoreList.Add (line.Trim ());
	}
}
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
mandel-macaque, line 109: This two lists can be a Set that is the resulting union of the ignored errors in the common file and in the current translation. What is more, there is no need to read the common ignore list for each of the cultures, we should do this once and re-use the value.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +84,77 @@ private string TranslateError (string culture, string errorCode)
...
+				} finally {
+					Thread.CurrentThread.CurrentUICulture = originalCulture;
+				}
+			}
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to ~/LocalizationIgnore/common-Translations.ignore or ~/LocaliztionIgnore/{culture}-Translations.ignore");
+		}
+
+		private List<string> ReadFile (string path)
+		{
```
mandel-macaque, line 138: What if path is not present, we will have an exception. Do a File.Exist, much nicer over doing work that is no needed.

mandel-macaque (changes requested):
I really dis-like adding empty files with not ignored errors. Both in xtro and the bcl tests we support the fact that the file is not present. The two list can be a single union Set and since we are doing that, we should be able to raise a warning of some type stating that an error is ignored in the common file and in the language specific one like we do in the xtro tests.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
rolfbjarne: I thought about this too, but the assert message is different, so I think it's worth it to keep the two checks.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
rolfbjarne: Or even a one-liner:

```csharp
return File.ReadAllLines (path).Where (line => !line.StartsWith ("#", StringComparison.Ordinal));
```

mandel-macaque: @rolfbjarne are we playing  ⛳  ?

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: I thought about this issue too and thought it would be helpful to show where the error to remove is.

tj-devel709:
@spouliot @mandel-macaque @rolfbjarne 
I believed I fixed the changes you guys requested! 
I also added two additional lists that track which error codes were found and can report back if there is anything in the ignore files that are not valid error codes.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
 			PropertyInfo propertyInfo = typeof (MSBStrings).GetProperty (errorCode);
 			return (string) propertyInfo.GetValue (null, null);
 		}
+
+		private List<string> commonIgnoreList { get; set; } = null;
```
rolfbjarne, line 89: We don't use `private` when it's the default:

```suggestion
		List<string> commonIgnoreList { get; set; } = null;
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
+			bool isCommonFileExistant = File.Exists (fullCommonPath);
+			bool isCultureFileExistant = File.Exists (fullCulturePath);
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
+
+			if (commonIgnoreList == null && isCommonFileExistant){
+				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
+				isFirstRun = true;
```
rolfbjarne, line 123: This variable is not needed, it's never read.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
rolfbjarne: You can return an empty array here if the file does not exist, which simplifies the code above:

```csharp
if (!File.Exists (path))
    return Array.Empty<string> ();
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+					Thread.CurrentThread.CurrentUICulture = originalCulture;
+				}
+			}
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
+			Assert.AreEqual (new List<string> (), cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid. Please remove invalid codes");
+			Assert.AreEqual (new List<string> (), commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid. Please remove invalid codes");
+		}
+
+		private IEnumerable<string> ReadFile (string path)
```
rolfbjarne, line 155:
```suggestion
		IEnumerable<string> ReadFile (string path)
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
+			bool isCommonFileExistant = File.Exists (fullCommonPath);
+			bool isCultureFileExistant = File.Exists (fullCulturePath);
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
+
+			if (commonIgnoreList == null && isCommonFileExistant){
+				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
+				isFirstRun = true;
+			}
```
rolfbjarne, line 124: If:

1. ReadFile is changed to handle non-existent files
2. ReadFile is changed to return `IList<string>`
3. The `cultureIgnoreList` is changed to be an `List<string>`

This can be simplified to:

```suggestion
			if (commonIgnoreList == null)
				commonIgnoreList = ReadFile (fullCommonPath);
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+
+			if (isCultureFileExistant)
+				cultureIgnoreList = new List<string> (ReadFile (fullCulturePath));
```
rolfbjarne, line 127: and:

```suggestion
			cultureIgnoreList = ReadFile (fullCulturePath);
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+
+			foreach (var errorCodeInfo in typeof (MSBStrings).GetProperties ()) {
+				try {
+					errorCode = errorCodeInfo.Name;
+					if (errorCode == "ResourceManager" || errorCode == "Culture")
+						continue;
+					string englishError = TranslateError ("en-US", errorCode);
+					string newCultureError = TranslateError (culture, errorCode);
+
+					if (isCommonFileExistant && commonIgnoreList.Contains (errorCode)) {
```
rolfbjarne, line 137: You don't need the isCommonFileExistant check here.
```suggestion
					if (commonIgnoreList.Contains (errorCode)) {
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+						Assert.AreEqual (englishError, newCultureError, $"{errorCode} is translated. Remove {errorCode} from {shortCommonPath}");
+						commonValidEntries.Add (errorCode);
+					} else if (isCultureFileExistant && cultureIgnoreList.Contains (errorCode)) {
```
rolfbjarne, line 140: and:

```suggestion
					} else if (cultureIgnoreList.Contains (errorCode)) {
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+						cultureValidEntries.Add (errorCode);
+					} else if (englishError == newCultureError)
+						errorList += $"{errorCode} ";
+				} finally {
+					Thread.CurrentThread.CurrentUICulture = originalCulture;
+				}
+			}
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
+			Assert.AreEqual (new List<string> (), cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid. Please remove invalid codes");
```
rolfbjarne, line 151: I think this looks a bit better:

```suggestion
			Assert.IsEmpty (cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid. Please remove invalid codes");
```

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			Assert.AreEqual (new List<string> (), commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid. Please remove invalid codes");
```
rolfbjarne, line 152: Same

rolfbjarne (changes requested):

spouliot (approved):
once @rolfbjarne comments are addressed

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
mandel-macaque: Do we even need a property at all? Can be an instance variable.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+		[TestCase ("ko-KR")]
+		[TestCase ("pl-PL")]
+		[TestCase ("pt-BR")]
+		[TestCase ("ru-RU")]
+		[TestCase ("tr-TR")]
+		[TestCase ("zh-CN")]
+		[TestCase ("zh-TW")]
+		public void AllErrorTranslation (string culture)
+		{
+			string errorCode = string.Empty;
```
mandel-macaque, line 106: this var is used in the loop at https://github.com/xamarin/xamarin-macios/pull/9971/files#diff-6fa1d6c0dbf0fad08398e2fc9ed0b96550f0000bd031dbe8620035340d2b39d1R131 why do we init it here? Can we just do:
```csharp
var errorCode = errorCodeInfo.Name
```
I don't see it used in other scope but the loop.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			string errorList = string.Empty;
```
mandel-macaque, line 107: Same.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			List<string> cultureIgnoreList = null;
+			List<string> commonValidEntries = new List<string> ();
+			List<string> cultureValidEntries = new List<string> ();
+
+			string fullCommonPath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore";
+			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
+			bool isCommonFileExistant = File.Exists (fullCommonPath);
+			bool isCultureFileExistant = File.Exists (fullCulturePath);
```
mandel-macaque, line 118: If we follow @rolfbjarne comment, were we return an empty list and the ReadFile supports missing files, we don't need these variables AFAIK.

mandel-macaque (changes requested):

whitneyschmidt (approved):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: I made this one a property so we could update it the first time and then use the values in for the other tests

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: errorCode is used only inside the foreach loop, but errorList is used after the loop to report which errorCodes were not translated

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: @rolfbjarne I am running into casting issues when trying to follow the above. Regarding Step 3, cultureIgnoreList is already a List<string>. Was this possibly a typo? Thanks!

tj-devel709: So I was able to return IList<string> from ReadFile by appending .ToList() to the return statement and changing Array.Empty<string> () to new List<string> (). 
I also casted the ReadFile calls to make them "ReadFile (fullCommonPath) as List<string>"
Is there a better way to do that?

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,76 @@ private string TranslateError (string culture, string errorCode)
...
+			CultureInfo originalCulture = Thread.CurrentThread.CurrentUICulture;
+
+			if (commonIgnoreList == null && isCommonFileExistant){
+				commonIgnoreList = new List<string> (ReadFile (fullCommonPath));
```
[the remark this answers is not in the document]
tj-devel709: @mandel-macaque i tried doing the common ignore in a [SetUp]. Is this how you were envisioning?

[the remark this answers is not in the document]
tj-devel709: Also can you explain why it is thread unsafe? Is it because I am trying to read two different files back to back?

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,74 @@ private string TranslateError (string culture, string errorCode)
 			PropertyInfo propertyInfo = typeof (MSBStrings).GetProperty (errorCode);
 			return (string) propertyInfo.GetValue (null, null);
 		}
+
+		List<string> commonIgnoreList = null;
```
rolfbjarne, line 89: You can declare this as IList:

```suggestion
		IList<string> commonIgnoreList = null;
```

that will remove the need for casting to `List<string>` and the `.ToList ()` call later.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,74 @@ private string TranslateError (string culture, string errorCode)
...
+
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
+			Assert.IsEmpty (cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid or are repeated. Please remove.");
+			Assert.IsEmpty (commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid or are repeated. Please remove.");
+		}
+
+		IList<string> ReadFile (string path)
+		{
+			if (!File.Exists (path))
+				return new List<string> ();
```
rolfbjarne, line 154: You can return an empty array if you do the other change to the field type at the top:

```suggestion
				return Array.Empty<string> ();
```

this is more efficient (it won't create a new instance every time this method is called).

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore
```diff
@@ -0,0 +1,3 @@
+# Insert Error codes that are waiting on translations below (one per line)
+W0111
+E7069
```
rolfbjarne, line 3: According to the test results you need to add `E7070` here as well.

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,74 @@ private string TranslateError (string culture, string errorCode)
...
+		[TestCase ("zh-CN")]
+		[TestCase ("zh-TW")]
+		public void AllErrorTranslation (string culture)
+		{
+			string errorList = string.Empty;
+			List<string> cultureIgnoreList = null;
+			List<string> commonValidEntries = new List<string> ();
+			List<string> cultureValidEntries = new List<string> ();
+
+			string fullCommonPath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/common-Translations.ignore";
```
rolfbjarne, line 117: This variable isn't used here anymore, so it should be moved to the SetUp method (and used there instead of duplicating the path creation logic).

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,74 @@ private string TranslateError (string culture, string errorCode)
...
+			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/common-Translations.ignore";
+			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Tasktests/LocalizationIgnore/{culture}-Translations.ignore";
```
rolfbjarne, line 120: Incorrect case in the path:

```suggestion
			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore";
			string shortCulturePath = $"xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
```

rolfbjarne (changes requested):
This is getting closer!

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: Should I take it out completely since I will only be using it that one time or is it worth keeping to make the code a little cleaner?

rolfbjarne: Either way works!

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: Switching the commonIgnoreList and cultureIgnoreList to IList<string> worked, but if I take off the .ToList(), I get this compiler error 
"TaskTests/LocalizationStringTest.cs(154,11): error CS0266: Cannot implicitly convert type 'System.Collections.Generic.IEnumerable<string>' to 'System.Collections.Generic.IList<string>'. An explicit conversion exists (are you missing a cast?) [/Users/tjlambert/Documents/AllStrings/xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/Xamarin.MacDev.Tasks.Tests.csproj]"

whitneyschmidt: Hm...if you look up the C# docs what is the relationship between IEnum and IList? It looks like you probably need some kind of a cast: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/casting-and-type-conversions

tj-devel709: I had also tried appending "... as IList<string>" instead of ".ToList()". It seemed that it was returning null since I was getting this NullReferenceException when trying to run the .Contains() method on commonIgnoreList in line 132. Perhaps I am doing the casting incorrectly though 

13) Error : Xamarin.iOS.Tasks.LocalizationStringTest.AllErrorTranslation("zh-TW")
System.NullReferenceException : Object reference not set to an instance of an object
  at Xamarin.iOS.Tasks.LocalizationStringTest.AllErrorTranslation (System.String culture) [0x000c8] in /Users/tjlambert/Documents/AllStrings/xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs:132 
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-06/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395

tj-devel709: Also if I try (IList<string>) as a cast, I get this:
System.InvalidCastException : Specified cast is not valid.
  at Xamarin.iOS.Tasks.LocalizationStringTest.ReadFile (System.String path) [0x00016] in /Users/tjlambert/Documents/AllStrings/xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs:154 
  at Xamarin.iOS.Tasks.LocalizationStringTest.SetUp () [0x00001] in /Users/tjlambert/Documents/AllStrings/xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs:94 
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-06/external/bockbuild/builds/mono-x64/mcs/class/corlib/System.Reflection/RuntimeMethodInfo.cs:395

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,73 @@ private string TranslateError (string culture, string errorCode)
...
+			Assert.AreEqual (string.Empty, errorList, $"The following errors were not translated. Add them to {shortCommonPath} or {shortCulturePath}");
+			Assert.IsEmpty (cultureIgnoreList.Except (cultureValidEntries), $"Not all error codes in {shortCulturePath} are valid or are repeated. Please remove.");
+			Assert.IsEmpty (commonIgnoreList.Except (commonValidEntries), $"Not all error codes in {shortCommonPath} are valid or are repeated. Please remove.");
+		}
+
+		IList<string> ReadFile (string path)
+		{
+			if (!File.Exists (path))
+				return Array.Empty<string> ();
+			return File.ReadAllLines (path).Where (line => !line.StartsWith ("#", StringComparison.Ordinal) && line != string.Empty).ToList();
```
rolfbjarne, line 154: Just leave the `ToList ()` call, that'll work just fine.

Except style:

```suggestion
			return File.ReadAllLines (path).Where (line => !line.StartsWith ("#", StringComparison.Ordinal) && line != string.Empty).ToList ();
```

rolfbjarne (approved):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
mandel-macaque: WE could use a StringBuilder rather that += "{errorcode} "

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -83,5 +85,73 @@ private string TranslateError (string culture, string errorCode)
...
+		[TestCase ("zh-TW")]
+		public void AllErrorTranslation (string culture)
+		{
+			string errorList = string.Empty;
+			IList<string> cultureIgnoreList = null;
+			List<string> commonValidEntries = new List<string> ();
+			List<string> cultureValidEntries = new List<string> ();
+
+			string fullCulturePath = $"{Directory.GetCurrentDirectory ()}/TaskTests/LocalizationIgnore/{culture}-Translations.ignore";
+			string shortCommonPath = "xamarin-macios/tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationIgnore/common-Translations.ignore";
```
mandel-macaque, line 118: Could be a static instance var.

mandel-macaque (changes requested):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
mandel-macaque: You do not need a IList, and IList and IEnumerable are diff interfaces, but an IList is a IEnumerable as per inheriting from ICollection. You should not cast since it wont work (or there is no guarantee it will). If you were to remove the IList, you will have to do a number of changes:

1. ReadFile -> Should return an IEnumerable, we remove the call toList
2. You will need to use the extension method for IEnumerable to do Contains, since IEnumerables only have GetEnumerator, that method is here: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.contains?view=netcore-3.1 which is present in Linq, means you will need the name space to be added.

@tj-devel709 @whitneyschmidt and IEnumerable is an interface that will only return an IEnumerator<out T>. That allows you to create "collections" that you cannot insert or remove from but iterate. This collections can be used in a foreach loop.

You can see here that the IL generated is a little diff https://sharplab.io/#gist:26a263aebdb501b064d4f022be8db657

IEnumerables are very useful when working with collections that you generated on the fly, with yield return. @tj-devel709  you probably have seen a lot of yield returns in binding tools for swift.

```csharp
public class Galaxies
    {

        public System.Collections.Generic.IEnumerable<Galaxy> NextGalaxy
        {
            get
            {
                yield return new Galaxy { Name = "Tadpole", MegaLightYears = 400 };
                yield return new Galaxy { Name = "Pinwheel", MegaLightYears = 25 };
                yield return new Galaxy { Name = "Milky Way", MegaLightYears = 0 };
                yield return new Galaxy { Name = "Andromeda", MegaLightYears = 3 };
            }
        }
    }
```

You can also find them in several of our tests that create test data: https://github.com/xamarin/xamarin-macios/blob/main/tests/xharness/Microsoft.DotNet.XHarness.iOS.Shared.Tests/Hardware/DeviceTest.cs#L49

For those that have used python, it is very similar to this:

```python
def csv_reader(file_name):
    for row in open(file_name, "r"):
        yield row
```
When a file is too large, you will run out of memory (in the above example). But with a yield you do not store the entire collection in memory, yet you can iterate through it.

Most of the methods from Linq do return IEnumerables because they are used to work with large collections and instead of duplicating the memory usage they use an IEnumerator with a MoveNext: https://github.com/microsoft/referencesource/blob/master/System.Core/System/Linq/Enumerable.cs#L184

yield return will do the magic for you to have a IEnumerator without the need of the class implementation.

mandel-macaque: I forgot to mention, do not remove the ToList, the contains will have to go through the entire list with no gain.

mandel-macaque (approved):

PR #9975, opened 2020-10-27
[tests] Clean up Xamarin.MacDev.Tasks.Tests.

rolfbjarne:
* Remove a lot of dead code.
* Use TestBase.CreateTask<T> to create tasks (and set required properties for all
  tasks) instead of instantiating tasks directly. This required subclassing TestBase
  in a few places, as well as making a few helper methods instance methods instead
  of static methods.
* Bump to net472.
* A few other misc simplifications.


mandel-macaque (approved):

spouliot (approved):
❤️

dalexsoto (approved):

tj-devel709 (approved):

PR #9981, opened 2020-10-27
[README] Bump downloads to our new Xamarin.iOS stable

dalexsoto:


chamons (approved):

mandel-macaque (approved):

README.md
```diff
@@ -33,7 +33,7 @@ However, we provide links to older Xamarin.iOS and Mac packages for macOS downgr
 
 | Version | Xamarin.iOS       | Xamarin.Mac       |
 |---------|-------------------|-------------------|
-| d16.7   | [14.0.0.0][15]   | [6.20.2.2][16]    |
+| d16.7   | [14.2.0.12][15]   | [6.20.2.2][16]    |
```
rolfbjarne, line 36: Shouldn't we add to the table instead of replacing the top entry?

Something like this:

| Version | Xamarin.iOS       | Xamarin.Mac       |
|---------|-------------------|-------------------|
| d16-7 (Xcode 12.1) | [14.2.0.12][15]   | [6.20.2.2][16]    |
| d16-7 (Xcode 12.0) | [14.0.0.0][15]   | [6.20.2.2][16]    |
| d16-7 | <original number, which is now lost> | <?> |

rolfbjarne (approved):

tj-devel709 (approved):

README.md
dalexsoto: Not sure, I think it might become quite confusing in the end having that xcode connotation like for example where is Xcode 12.0.1 build. So i think just keeping the latest released stable version is just fine

spouliot (approved):

PR #9983, opened 2020-10-28
[tests] Rewrite ValidateAppBundleTaskTests to not create a task in-process.

rolfbjarne:
This is a step towards making Xamarin.MacDev.Tests run tests out-of-process.

It requires adding '_GenerateBundleName' as a dependency for the
'_ValidateAppBundle' target, because we're invoking the '_ValidateAppBundle'
directly, and we can't depend on any other targets executing
'_GenerateBundleName', because there are no other executed targets.


spouliot (approved):

chamons (approved):
:heart:

whitneyschmidt (approved):

emaf (approved):

dalexsoto (approved):

PR #10008, opened 2020-10-29
[Actions] Add debug for the workflow to find why it does not work with hub

mandel-macaque:


dalexsoto (approved):

.github/workflows/label-checker.yml
```diff
@@ -9,14 +9,22 @@ jobs:
     name: Labels verification
 
     steps:
+    - run: echo "$GITHUB_CONTEXT"
+      name: 'Debug context'
+      env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+
     - run: exit 0
```
whitneyschmidt, line 17: how does the `if:` work here? asking since I'm curious about how all of these `exit 0` or `exit 1` interact w/ yaml's run conditions (on success/on failure) and the default condition, which is to only run when the prior step exited with a success code.

I could also completely off base here since I don't remember exactly how jobs work... 🙈

whitneyschmidt (approved):

.github/workflows/label-checker.yml
mandel-macaque: This syntax is specific to github actions, it wont work on vsts. On vsts it was 'condition:' in workflows is 'if:' but is the same in reality.

whitneyschmidt: Ah, got it! So all of the code from `- run:` to the next `-run:` is executed in GH actions, is not written in yaml 😆

mandel-macaque: No, run is the key in yaml. The if statement too. If you want to execute a multiline script, you need to do run: |

What I meant is that the key in GitHub is "if" while in vsts is "condition". Yaml is the format, and the schema from GitHub actions and VSTS are different.

mandel-macaque:
@monojenkins backport d16-8

mandel-macaque:

PR #10037, opened 2020-11-03
provision python packages

cadsit:


cadsit:
Devops build at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4199414&view=logs&j=5679deb7-4eb2-524c-4371-d233808e26cf

tools/devops/provision-python-packages.csx
```diff
@@ -0,0 +1,8 @@
+using Xamarin.Provisioning;
+
+ProvisioningScript.ElevatedExec (
```
mandel-macaque, line 3: If I understand correctly, we are installing this system wide. That would be a mistake. A better approach is to use a python virtualenv that inherits from the system based installed python modules.

For example:

```chsarp
pip install virtualenv
virtualenv "$VIRTUAL_ENV_PATH" --system-site-packages
source "$VIRTUAL_ENV_PATH/bin/activate"
pip install python-magic
```

The issue with using the activate in the virtual env, is that it will need to be done in the same exec process of the bash when we want to execute it. That is, we cannot do the virtual env activation in a step, then use python in a diff step. The correct possible approach is to do:

1. Setup a virtual env with dependencies.
2. Spawn a bash step that
    1. Activate the virtual env
    2. Executes python

If we are using bask steps on vsts, we wont need to call deactivate, although good citizens clean after them. It would be ideal that we have a way to provisionator to do the python installation, the virtualenv installation via pip (is not a default module in python2 AFAIK) and does call pip install on a freeze like the one we have. Then, that step should output a env variable with the virtual env. Ideally, we want the output variable in the provisionator step in case we want to setup more than one virtual env.

The issue I have with this approach, although it does work atm, is that it clutters the system and more importantly, the pipeline has side effects in the system. I do know that we already have side effects with brew.. and that is something we have to live with until we move to maybe using [chroot](https://linux.die.net/man/1/chroot) on Mac OS X + provisionator + brew (would be great to have [jails like in freebsd](https://www.freebsd.org/doc/handbook/jails.html), but Mac OS X is not really that cool).

Now that I think about it, provisionator with chroot + brew and vsts integration would be a really really amazing product. But then, I'm not PM ;)

mandel-macaque (changes requested):

whitneyschmidt (approved):

tools/devops/Makefile
```diff
@@ -8,3 +8,18 @@ device-tests-provisioning.csx: device-tests-provisioning.csx.in Makefile $(TOP)/
...
+
+build-provisioning.csx: build-provisioning.csx.in Makefile $(TOP)/Make.config
+	$(Q_GEN) sed \
+		-e 's#@XCODE_XIP_NAME@#$(notdir $(XCODE_URL))#g' \
+		-e 's#@MONO_PACKAGE@#$(MIN_MONO_URL)#g' \
+		-e 's#@VS_PACKAGE@#$(MIN_VISUAL_STUDIO_URL)#g' \
+		$< > $@;
+
+all check:
+	shellcheck *.sh
```
rolfbjarne, line 20: 🥳

tools/devops/Makefile
```diff
@@ -8,3 +8,18 @@ device-tests-provisioning.csx: device-tests-provisioning.csx.in Makefile $(TOP)/
...
+
+print-abspath-variable:
+	@echo $(VARIABLE)=$(abspath $($(VARIABLE)))
+
+provisioning: build-provisioning.csx device-tests-provisioning.csx
```
rolfbjarne, line 25: Missing eol

tools/devops/automation/build-pipeline.yml
```diff
@@ -0,0 +1,288 @@
...
+    name: VSEng-Xamarin-QA
+    demands:
+    - Agent.OS -equals Darwin
+    - Agent.OSVersion -equals 10.15
+    workspace:
+      clean: all
+
+  steps:
+  - checkout: self          # https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#checkout
+    clean: true             # Executes: git clean -ffdx && git reset --hard HEAD
```
rolfbjarne, line 65: Does it clean and reset submodules too?

tools/devops/automation/build-pipeline.yml
```diff
@@ -0,0 +1,288 @@
...
+    inputs:
+      provisioning_script: $(Build.SourcesDirectory)/xamarin-macios/tools/devops/provision-dotnetcore.csx
+      provisioning_extra_args: '-vvvv'
+      github-pat: $(GitHub.Token)
+    timeoutInMinutes: 30
+
+  - powershell: |
+      if ([IO.Directory]::Exists('/Developer/MonoTouch')) {
+        Write-Host "Deleting /Developer/MonoTouch"
+        bash -c "sudo rm -Rf /Developer/MonoTouch"
```
rolfbjarne, line 138: You're embedding bash inside powershell? Seems like this could be simplified a lot by using bash directly.

rolfbjarne (changes requested):
This seems much bigger than just "provision python packages", is this PR correct?

mandel-macaque:
@rolfbjarne  this PR is wrong, closing. Is pointing to the wrong branch AND is not the correct way.

tools/devops/automation/build-pipeline.yml
mandel-macaque: This is very dirty, i'll remove it @mcumming any reason to do it this way?

PR #10039, opened 2020-11-04
[msbuild] Share the app extensions targets between iOS and Mac.

rolfbjarne:
* Share the following targets:

    * _ResolveAppExtensionReferences
    * _SplitAppExtensionReferencesByExistent
    * _AssignAppExtensionConfiguration
    * _SeparateAppExtensionReferences
    * _CopyAppExtensionsToBundle

The first four were pretty much identical between iOS and Mac already, and
needed very few changes.

The latter, _CopyAppExtensionsToBundle, required a little bit of tweaking to
the targets to make sure it works for both iOS and Mac:

* Removed the conditions on the Ditto task to check if we have app extensions
  - this is unnecessary because Inputs/Outputs on the target itself (which
  weren't there when the Ditto task conditions were written).
* Added a '_PlaceAppExtensions' target that calculates where in the
  container's app bundle extensions should be placed (the '_AppExtensionRoot'
  property), and the name of the containing folder (the 'ContainerName'
  metadata).

### DIFF
--- a/msbuild/Xamarin.Shared/Xamarin.Shared.targets
+++ b/msbuild/Xamarin.Shared/Xamarin.Shared.targets
@@ -912,6 +912,39 @@ Copyright (C) 2018 Microsoft. All rights reserved.
 		</PrepareResourceRules>
 	</Target>
 
+	<!-- App extensions -->
+
+	<Target Name="_ResolveAppExtensionReferences" DependsOnTargets="_SplitAppExtensionReferencesByExistent">
+		<PropertyGroup>
+			<!-- When building a .sln with msbuild, the dependent projects may not be built. So, always build
+			     the referenced projects unless building from IDE. -->
+			<_BuildReferencedExtensionProjects Condition="'$(BuildingInsideVisualStudio)' != 'true'">true</_BuildReferencedExtensionProjects>
+		</PropertyGroup>
+
+		<!-- If the referenced projects have already been built, then just get the target paths -->
+		<MSBuild
+			Projects="@(_AppExtensionReferenceWithConfigurationExistent)"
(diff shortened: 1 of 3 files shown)


spouliot (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -912,6 +912,124 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
+		</AssignProjectConfiguration>
+
+		<!-- Else, just -->
+		<CreateItem Include="@(_AppExtensionReference)"
+					Condition="'$(CurrentSolutionConfigurationContents)' == ''">
+			<Output TaskParameter="Include" ItemName="_AppExtensionReferenceWithConfiguration"/>
+		</CreateItem>
+	</Target>
+
+	<Target Name="_SeparateAppExtensionReferences" BeforeTargets="AssignProjectConfiguration">
```
emaf, line 934: I know it was like this in the old target files but shouldn't the BeforeTargets be `_AssignProjectConfiguration` instead of `AssignProjectConfiguration`?

emaf (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.targets
rolfbjarne: I think that refers to the `AssignProjectConfiguration` target shipped by the msbuild: https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-targets?view=vs-2019 (we don't have a `_AssignProjectConfiguration`, and I couldn't find it anywhere else either).

PR #10046, opened 2020-11-04
[msbuild] Add support for `.xcframework`

spouliot:
This is done early so we can resolve the inner framework, inside the
xcframework, and let the existing framework support do most of the
work.

The resolving code has unit tests. Custom projects for "NoEmbedding"
exists for all supported platforms and executed by xharness.

A sample `xcframework` with tests projects is also available [here](https://github.com/spouliot/xcframework).

**Work in progress**
- [x] add XM support and tests
- [ ] add/confirm correct .dSYM support


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/BundlerToolTaskBase.cs
```diff
@@ -156,5 +157,64 @@ protected string CreateResponseFile (CommandLineArgumentBuilder arguments, IList
...
+				var supported_platform_variant = (PString) item ["SupportedPlatformVariant"];
+				if (supported_platform_variant?.Value != variant)
+					continue;
+				var supported_architectures = (PArray) item ["SupportedArchitectures"];
+				// each architecture we request must be present in the xcframework
+				// but extra architectures in the xcframework are perfectly fine
+				foreach (var arch in archs) {
+					bool found = false;
+					foreach (PString xarch in supported_architectures) {
+						found = String.Compare (arch, xarch.Value, StringComparison.OrdinalIgnoreCase) == 0;
```
rolfbjarne, line 206: I think you can use String.Equals:

```suggestion
						found = String.Equals (arch, xarch.Value, StringComparison.OrdinalIgnoreCase);
```

makes it a tiny bit simpler.

msbuild/Xamarin.iOS.Tasks.Core/Tasks/MTouchTaskBase.cs
```diff
@@ -119,9 +119,14 @@ void BuildNativeReferenceFlags (GccOptions gcc)
 					continue;
 				}
 
-				if (kind == NativeReferenceKind.Static) {
-					var libName = Path.GetFileName (item.ItemSpec);
+				var libName = Path.GetFileName (item.ItemSpec);
 
+				// existing (as of this commit) IDE versions will add a `.xcframework` directorty as a `Static` native reference
```
rolfbjarne, line 124:
```suggestion
				// existing (as of this commit) IDE versions will add a `.xcframework` directory as a `Static` native reference
```

msbuild/Xamarin.iOS.Tasks.Core/Tasks/MTouchTaskBase.cs
```diff
@@ -119,9 +119,14 @@ void BuildNativeReferenceFlags (GccOptions gcc)
...
+				if ((kind == NativeReferenceKind.Static) && (Path.GetExtension (libName) == ".xcframework"))
```
rolfbjarne, line 125: Thinking about dotnet support, I think it would be easier if the xcframework -> framework resolution was put into a new MSBuild target that ran before the _CompileToNative target, and rewrote the NativeReference item groups to be Frameworks pointing to the right path instead. This new target could then be re-used as-is in the dotnet code.

The end result is that there would be no changes to MTouchTaskBase (nor MmpTaskBase) classes, they would work as-is.

dalexsoto: What about dSYM files included in xcframeworks should we take this into account to put them in the output folder? here is an example of the structure of one xcframework

unknown

spouliot: @dalexsoto bad example as it is not the _normal_ structure. That's how Xcode do them: https://github.com/spouliot/xcframework#tree

and PSPDFKit tweaks them https://pspdfkit.com/blog/2020/supporting-xcframeworks/#bcsymbolmaps-and-dsyms

The idea is to make them them (xcframework) transparent to the down-level tooling. It might be easy to deal with custom dSYM or not, but it's not a top priority.

dalexsoto: Interesting, I guess you need to enable it on demand since [Xcode 12.0 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes) state the following

> XCFrameworks can now include .dSYM and .bcsymbolmap debugging symbol files in your library bundle with the -debug-symbols flag. Run the command xcodebuild -create-xcframework -help for additional usage information. (64910707)

spouliot: interesting, I'll update my repo-sample to add the flag and see how it affects the files inside the xcframework

spouliot: added https://github.com/spouliot/xcframework/commit/9fe205228e52163df563d9d208a38f6bd3a2a8d1
it does make it clear / well-defined (inside the `.plist`) where the dSYM can be found

tj-devel709 (approved):

spouliot:
build

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -132,11 +133,20 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
 	-->
-	<Target Name="_ExpandNativeReferences" Condition="'$(DesignTimeBuild)' != 'true'">
+	<Target Name="_ExpandNativeReferences" Condition="'$(DesignTimeBuild)' != 'true'" DependsOnTargets="_DetectSdkLocations;_ComputeTargetArchitectures">
 		<ItemGroup>
-			<_FrameworkNativeReference Include="@(NativeReference -> '%(Identity)/%(Filename)')" Condition="'%(Extension)' == '.framework'" />
-			<_FileNativeReference Include="@(NativeReference)" Condition="'%(Extension)' != '.framework'" />
+			<_FrameworkNativeReference Include="@(NativeReference -> '%(Identity)/%(Filename)')" Condition="'%(Extension)' == '.framework' Or '%(Extension)' == '.xcframework'" />
+			<_FileNativeReference Include="@(NativeReference)" Condition="'%(Extension)' != '.framework' And '%(Extension)' != '.xcframework'" />
 		</ItemGroup>
+		<ResolveNativeReferences
```
emaf, line 141: We need the SessionId prop to be set (used from Windows to run the task remotely) and I think we should conditionally skip the execution of this task if `IsMacEnabled` != true

emaf (changes requested):

tests/test-libraries/Makefile
```diff
@@ -191,6 +191,39 @@ $(eval $(call FatFrameworkTemplate,ios-fat,iphoneos,iphonesimulator))
 $(eval $(call FatFrameworkTemplate,tvos-fat,tvos,tvsimulator))
 $(eval $(call FatFrameworkTemplate,watchos-fat,watchos,watchsimulator))
 
+define XCTemplate
+.libs/by-platform/$(1)/XTest.framework/XTest: $(foreach arch,$(3),.libs/$(2)/libtest$(4).$(arch).dylib) | .libs/by-platform/$(1)/XTest.framework
```
rolfbjarne, line 195: Any reason you can't use the existing `XTest.framework`s?

```
$ ls -lad tests/test-libraries/.libs/*/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/ios-fat/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/iphoneos/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/iphonesimulator/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/macos/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/tvos-fat/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/tvos/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/tvsimulator/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/watchos-fat/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/watchos/XTest.framework
drwxr-xr-x  4 rolf  wheel  128 Nov 24 08:20 tests/test-libraries/.libs/watchsimulator/XTest.framework
```

msbuild/Xamarin.Shared/Xamarin.Shared.props
```diff
@@ -133,6 +133,7 @@ Copyright (C) 2020 Microsoft. All rights reserved.
 		<!-- The default architecture for Xamarin.Mac is x86_64 -->
 		<TargetArchitectures Condition="'$(TargetArchitectures)' == '' And '$(_PlatformName)' == 'macOS'">x86_64</TargetArchitectures>
 		<!-- There should always be an MtouchArch value in newer projects, but for older projects default to old values -->
+		<TargetArchitectures Condition="'$(TargetArchitectures)' == '' And '$(_PlatformName)' == 'tvOS' And '$(ComputedPlatform)' == 'iPhoneSimulator'">x86_64</TargetArchitectures>
```
rolfbjarne, line 136: This shouldn't be necessary, all tvOS projects should have an MtouchArch value.

rolfbjarne (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.props
spouliot: I'll try again but I'm pretty sure it did not work. However it might just be the xharness created/cloned projects that are missing the value.

tests/test-libraries/Makefile
spouliot: That should work, it came from https://github.com/rolfbjarne/xamarin-macios/commit/xcframework#diff-c0db661233e8bd837a40c8de3bdbb802b63d5edb88fa262a82d9799ee39e50ebR155 , but then I renamed it XTest since XCTest is the Apple framework for Xcode unit tests and something (linking or codesigning) did not appreciate the renaming.

rolfbjarne: Oh I see what happened here. My point was that this code is more complicated than it needs to be; in any case I fixed it and pushed the fix.

spouliot: thanks, even simpler of the changes I made locally

spouliot:
> make[1]: *** [downloads/dotnet-sdk-6.0.100-alpha.1.20562.2-osx-x64.tar.gz] Error 56

emaf (approved):

dalexsoto (approved):

nesevis:
There may have been some sort of regression here, as even @spouliot's test project referenced in the PR fails to build with the newest version of Xamarin.

PR #10049, opened 2020-11-05
[msbuild] Add Metal files to our items included by default.

rolfbjarne:


dalexsoto (approved):

tests/common/DotNet.cs
```diff
@@ -100,8 +100,6 @@ public static void CompareApps (string old_app, string new_app)
 						return false;
 
 					// There's a lot of TODOs here, those correspond with missing features in .NET and will be removed as those features are implemented
```
spouliot, line 102: comment still needed ?

spouliot (approved):

tests/common/DotNet.cs
rolfbjarne: Yes, there are more TODOs further down, and in any case I have pending PRs that will eventually remove this line.

PR #10065, opened 2020-11-06
[tools] Add all our product constants to SdkVersions.cs and use them in dotnet-linker.

rolfbjarne:
Also, the correct constants to use is now determined by current platform in the Application
instance, instead of a constant value.


spouliot (approved):

tj-devel709 (approved):

whitneyschmidt (approved):

mandel-macaque (approved):

tools/common/cache.cs
```diff
@@ -186,12 +186,12 @@ public unsafe static bool CompareStreams (Stream astream, Stream bstream, bool i
 		} while (true);
 	}
 	
-	string GetArgumentsForCacheData ()
+	string GetArgumentsForCacheData (Application app)
```
stephen-hawley, line 189: Are you likely to use any other properties of `app`? If not, why not pass in `ProductConstants`?

rolfbjarne: > Are you likely to use any other properties of `app`?

Logic says no, experience says yes... somehow most places end up using many `app` properties.

PR #10076, opened 2020-11-09
[Xcode12.2] Bring xcode12.2 changes to d16-8-xcode12.2

mandel-macaque:
Create a branch to be ready for the release with the changes from xcode12.2 on top of those in xcode12.1.1


.gitmodules
```diff
@@ -1,7 +1,3 @@
-[submodule "external/guiunit"]
-    path = external/guiunit
-    url = ../../mono/guiunit.git
-    branch = main
```
mandel-macaque, line 4: Not sure about this.

Make.config
```diff
@@ -322,7 +322,6 @@ MONO_PATH=$(TOP)/external/mono
 TOUCH_UNIT_PATH=$(TOP)/external/Touch.Unit
 OPENTK_PATH=$(TOP)/external/opentk
 XAMARIN_MACDEV_PATH=$(TOP)/external/Xamarin.MacDev
-GUI_UNIT_PATH=$(TOP)/external/guiunit
```
mandel-macaque, line 325: Not sure if we need this.

.gitmodules
spouliot: should be identical to `d16-8` since we have not touched this in `xcode12.x`

spouliot (approved):

dalexsoto (approved):

rolfbjarne (approved):

tj-devel709 (approved):

PR #10079, opened 2020-11-10
[dotnet] Bump to .NET 6.0.100-alpha.1.20556.2 and net6.0

rolfbjarne:
This involves a few changes:

* Change everything to reference net6.0 instead of net5.0
* Update various variables to be NET6* instead of NET5*
* Reorder build logic to account for that our targets are imported earlier in
  the build process:

   In the latest .NET 6, our Workloads.targets is imported earlier in the
   build. This requires a few changes, because we still need to run most of
   our logic later in the process, which we do by adding targets files we
   want imported later to the AfterMicrosoftNETSdkTargets property.

   What we're loading as soon as possible:

   * Our version information (Xamarin.Shared.Sdk.Versions.targets)
   * The supported OS versions
     (Microsoft.<platform>.Sdk.SupportedTargetPlatforms.targets)
   * The default OS version
     (Xamarin.Shared.Sdk.TargetFrameworkInference.targets).

   This is all information that the .NET build require early on.

   Changes:

   * Rename all files that are loaded early to *.props.
   * Updated documentations to reflect these changes.
   * Remove Microsoft.<platform>.TargetFrameworkInference.targets, these
     files aren't used and don't contain anything useful.
   * Move the logic to calculate _ComputedTargetFrameworkMoniker has been
     delayed to later, because it needs TargetFrameworkMoniker set.

* Add a StoreAttributesStep to store attributes that are removed by the
  linker, but that the static registrar needs.

   In particular, in .NET 6 the linker removes the
   System.Runtime.CompilerServices.ExtensionAttribute, which the static
   registrar needs to handle category methods properly.

   This involved copying and slightly modifying the RemoveAttributesBase
   code.


spouliot (commented):
> Add a StoreAttributesStep to store attributes that are removed by the
linker, but that the static registrar needs.

is duplicated in the commit message

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -126,7 +120,15 @@
...
-	<Import Project="Xamarin.Shared.Sdk.TargetFrameworkInference.targets" />
+
+  <!-- This is a hack until we get the TFM values for real from the .NET build logic -->
+  <PropertyGroup>
+    <_TargetFrameworkPlatform Condition="'$(_PlatformName)' == 'iOS'">ios</_TargetFrameworkPlatform>
+    <_TargetFrameworkPlatform Condition="'$(_PlatformName)' == 'tvOS'">tvos</_TargetFrameworkPlatform>
+    <_TargetFrameworkPlatform Condition="'$(_PlatformName)' == 'watchOS'">watchos</_TargetFrameworkPlatform>
+    <_TargetFrameworkPlatform Condition="'$(_PlatformName)' == 'macOS'">macos</_TargetFrameworkPlatform>
+    <_ComputedTargetFrameworkMoniker Condition=" '$(_TargetFrameworkPlatform)' != '' ">$(TargetFrameworkMoniker),Profile=$(_TargetFrameworkPlatform)</_ComputedTargetFrameworkMoniker>
+  </PropertyGroup>
```
spouliot, line 131: indent seems off

spouliot (approved):

Make.config
```diff
@@ -440,17 +440,17 @@ DOTNET_NUPKG_DIR ?= $(DOTNET_DESTDIR)/nupkgs
...
+DOTNET6_VERSION=6.0.100-alpha.1.20559.4
+DOTNET6_VERSION_BAND=$(firstword $(subst -, ,$(DOTNET6_VERSION)))
+DOTNET6_URL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20559.4/dotnet-sdk-6.0.100-alpha.1.20559.4-osx-x64.pkg
+DOTNET6_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20559.4/dotnet-sdk-6.0.100-alpha.1.20559.4-osx-x64.tar.gz
+DOTNET6_TARBALL_NAME=$(notdir $(DOTNET6_TARBALL))
+DOTNET6_DIR=$(abspath $(TOP)/builds/downloads/$(basename $(basename $(DOTNET6_TARBALL_NAME))))
+DOTNET6=$(DOTNET6_DIR)/dotnet
 DOTNET5_BCL_REF_URL=https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/7d9f5c21-0d79-403f-bfe3-9a4506529760/nuget/v3/flat2/microsoft.netcore.app.ref/5.0.0-rc.1.20404.16/microsoft.netcore.app.ref.5.0.0-rc.1.20404.16.nupkg
 DOTNET5_BCL_REF_NAME=microsoft.netcore.app.ref.5.0.0-rc.1.20404.16.nupkg
 DOTNET5_BCL_DIR:=$(abspath $(TOP)/builds/downloads/$(basename $(DOTNET5_BCL_REF_NAME)))/ref/net5.0
```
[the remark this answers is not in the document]
rolfbjarne: Yes, we can continue to use the .NET 5 BCL as reference assemblies when building our own platform assemblies for now. We'll probably have to bump at some point, but that's independent of this PR.

tj-devel709 (approved):

mandel-macaque (approved):

Make.config
```diff
@@ -440,17 +440,17 @@ DOTNET_NUPKG_DIR ?= $(DOTNET_DESTDIR)/nupkgs
...
 # avoid consuming a lot of disk space (since they're never automatically deleted). The system-dependencies.sh script will install locally as long
 # as there's a TARBALL url.
-DOTNET5_VERSION=5.0.100-rtm.20509.5
-DOTNET5_VERSION_BAND=$(firstword $(subst -, ,$(DOTNET5_VERSION)))
-DOTNET5_URL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/5.0.100-rtm.20509.5/dotnet-sdk-5.0.100-rtm.20509.5-osx-x64.pkg
-DOTNET5_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/5.0.100-rtm.20509.5/dotnet-sdk-5.0.100-rtm.20509.5-osx-x64.tar.gz
-DOTNET5_TARBALL_NAME=$(notdir $(DOTNET5_TARBALL))
-DOTNET5_DIR=$(abspath $(TOP)/builds/downloads/$(basename $(basename $(DOTNET5_TARBALL_NAME))))
-DOTNET5=$(DOTNET5_DIR)/dotnet
+DOTNET6_VERSION=6.0.100-alpha.1.20559.4
```
jonathanpeppers, line 447: We ended up landing with `6.0.100-alpha.1.20562.2`:

https://github.com/xamarin/xamarin-android/commit/6f761ea854f043f88dd41bb25bac608c7a680dfe

It was the first build that fixed the MSBuild bug with spaces in properties.

rolfbjarne: Ok, trying that!

Make.config
```diff
@@ -444,10 +444,10 @@ DOTNET_FEED_DIR ?= $(DOTNET_DESTDIR)/nuget-feed
 # We're using preview versions, and there will probably be many of them, so install locally (into builds/downloads) if there's no system version to
 # avoid consuming a lot of disk space (since they're never automatically deleted). The system-dependencies.sh script will install locally as long
 # as there's a TARBALL url.
-DOTNET6_VERSION=6.0.100-alpha.1.20559.4
+DOTNET6_VERSION=6.0.100-alpha.1.20562.2
 DOTNET6_VERSION_BAND=$(firstword $(subst -, ,$(DOTNET6_VERSION)))
-DOTNET6_URL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20559.4/dotnet-sdk-6.0.100-alpha.1.20559.4-osx-x64.pkg
-DOTNET6_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20559.4/dotnet-sdk-6.0.100-alpha.1.20559.4-osx-x64.tar.gz
+DOTNET6_URL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20562.2/dotnet-sdk-6.0.100-alpha.1.20562.2-osx-x64.pkg
+DOTNET6_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-alpha.1.20562.2/dotnet-sdk-6.0.100-alpha.1.20562.2-osx-x64.tar.gz
```
spouliot, line 450: could (or should) `DOTNET6_VERSION` be used in both URL ?

rolfbjarne: I tried that once, and it breaks in `system-dependencies.sh`, because that file greps Make.config for the url, and ends up with a url with DOTNET6_VERSION in it (instead of the actual version). It's possible to fix it, but I think the fix would be worse than a little bit of copy/paste.

spouliot: k, they are close together so it should not be too much of a trouble :)

stephen-hawley (approved):

PR #10099, opened 2020-11-16
[fileprovider] Fix breaking change

mandel-macaque:


rolfbjarne (approved):

dalexsoto (approved):

spouliot (approved):

src/FileProvider/Compat.cs
```diff
@@ -5,7 +5,16 @@
 
 namespace FileProvider {
 
-#if !MONOMAC
+#if MONOMAC
+	public static partial class NSFileProviderItem_Extensions {
+
+		[Obsolete ("This API was removed from macOS in Xcode 12.2 beta 2")]
```
stephen-hawley, line 11: Do we have a policy about using the `error` variant?
I don't know that a runtime `NotSupportedException` is better than a compile error.

mandel-macaque:
Failures are unrelated.

PR #10118, opened 2020-11-19
[VSTS] Add support to get the info of a PR. 

mandel-macaque:
Use scripts under automation, device-tests will go away.

### DIFF
--- a/tools/devops/automation/scripts/GitHub.Tests.ps1
+++ b/tools/devops/automation/scripts/GitHub.Tests.ps1
@@ -489,3 +489,82 @@ Describe 'Test-JobSuccess' {
         Test-JobSuccess -Status "Random value" | Should -Be $False
     }
 }
+
+Describe 'Get-GitHubPRInfo' {
+    Context 'with all env variables present' {
+
+        BeforeAll {
+            $Script:envVariables = @{
+                "GITHUB_TOKEN" = "GITHUB_TOKEN";
+            }
+
+            $Script:envVariables.GetEnumerator() | ForEach-Object { 
+                $key = $_.Key
+                Set-Item -Path "Env:$key" -Value $_.Value
+            }
+        }
+
+        It 'calls the method succesfully' {
+            Mock Invoke-RestMethod {
+                return @{"status"=200;}
+            }
+            $changeId = "ChangeId"
+
+            Get-GitHubPRInfo -ChangeId $changeId
+
+            # assert the call and compare the expected parameters to the received ones
+            Assert-MockCalled -CommandName Invoke-RestMethod -Times 1 -Scope It -ParameterFilter {
+                # validate each of the params and the payload
+                if ($Uri -ne "https://api.github.com/repos/xamarin/xamarin-macios/pulls/$changeId") {
+                    return $False
+                }
+                if ($Headers.Authorization -ne ("token {0}" -f $envVariables["GITHUB_TOKEN"])) {
+                    return $False
+                }
+                if ($Method -ne "POST") {
+                    return $False
+                }
+                if ($ContentType -ne "application/json") {
+                    return $False
+                }
+
+                return $True
+            }
+
+        }
+
+        It 'calls the method with an error and throws' {
+            Mock Invoke-RestMethod {
+                throw [System.Exception]::("Test")
+            }
+
+            $changeId = "ChangeId"
+
+            { Get-GitHubPRInfo -ChangeId $changeId } | Should -Throw
+        }
+
+    }
+    Context 'without an env variable' {
+        BeforeAll {
+            # clear the env vars
+            $envVariables = @{
+                "GITHUB_TOKEN" = "GITHUB_TOKEN";
+            }
+            $envVariables.GetEnumerator() | ForEach-Object { 
+                $key = $_.Key
+                Remove-Item -Path "Env:$key"
+            }
+        }
+        It 'throws and error' {
+
+            Mock Invoke-RestMethod {
+                return @{"status"=200;}
+            }
+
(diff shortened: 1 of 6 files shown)


cadsit:
Closing for now -- implementing Alex's feedbacks and will re-introduce a new PR when ready.

mandel-macaque:
We got the wrong PRs @cadsit dirty pastebin lol

tj-devel709 (approved):

tools/devops/automation/templates/configure-build.yml
```diff
@@ -0,0 +1,30 @@
+# This job will parse all the labels present in a PR, will set
+# the tags for the build AND will set a number of cofiguration
```
rolfbjarne, line 2:
```suggestion
# the tags for the build AND will set a number of configuration
```

tools/devops/automation/templates/configure-build.yml
```diff
@@ -0,0 +1,30 @@
...
+    $buildSourceBranchName = "$(Build.SourceBranchName)"
+    $buildSourceBranch = "$(Build.SourceBranch)"
+
+    Write-Host "buildReason: ${buildReason}"
+    Write-Host "buildSourceBranchName: ${buildSourceBranchName}"
+    Write-Host "buildSourceBranch: $buildSourceBranch"
+
+    if ($buildReason -eq 'PullRequest') {
+      Write-Host "Configuring build from PR."
+      $prInfo = Get-GitHubPRInfo -ChangeId 10115
```
rolfbjarne, line 22: Is the number supposed to be hard coded?

mandel-macaque: Testing, it is not :)

mandel-macaque: As in landing hardcoded, will remove later

rolfbjarne (approved):

tools/devops/automation/scripts/GitHub.Tests.ps1
```diff
@@ -488,4 +488,83 @@ Describe 'Test-JobSuccess' {
...
-}
\ No newline at end of file
+}
+
+Describe 'Get-GitHubPRInfo' {
+    Context 'with all env variables present' {
+
+        BeforeAll {
+            $Script:envVariables = @{
+                "GITHUB_TOKEN" = "GITHUB_TOKEN";
```
mjbond-msft, line 498: Should the value `GITHUB_TOKEN` (right side of the assignment) include some sort of value resolution such as $(GITHUB_TOKEN) or $GITHUB_TOKEN)? otherwise, it seems that the token value will then get set to the constant `GITHUB_TOKEN` rather than an actual token

tools/devops/automation/scripts/GitHub.Tests.ps1
```diff
@@ -488,4 +488,83 @@ Describe 'Test-JobSuccess' {
...
+
+            { Get-GitHubPRInfo -ChangeId $changeId } | Should -Throw
+        }
+
+    }
+    Context 'without an env variable' {
+        BeforeAll {
+            # clear the env vars
+            $envVariables = @{
+                "GITHUB_TOKEN" = "GITHUB_TOKEN";
```
mjbond-msft, line 551: Same here. Should the right hand of the assignment be $Env:GITHUB_TOKEN?

tools/devops/automation/templates/configure-build.yml
```diff
@@ -0,0 +1,47 @@
...
+    Write-Host "buildReason: ${buildReason}"
+    Write-Host "buildSourceBranchName: ${buildSourceBranchName}"
+    Write-Host "buildSourceBranch: $buildSourceBranch"
+
+    if ($buildReason -eq 'PullRequest') {
+      Write-Host "Configuring build from PR."
+      # This is an interesting step, we do know we are dealing with a PR, but we need the PR id to
+      # be able to get the labels, the buildSourceBranch follows the pattern: refs/pull/{ChangeId}/merge
+      # we could use a regexp but then we would have two problems instead of one
+      $changeId = $buildSourceBranch.Replace("refs/pull/", "").Replace("/merge", "")
```
mjbond-msft, line 25: There's actually a variable that will give you this commit id without having to parse it out:  $(System.PullRequest.SourceCommitId)

tools/devops/automation/scripts/GitHub.Tests.ps1
mandel-macaque: is a test. can be any value, we just use the value to check if it is added to the auth header. So the tests sets the env var, used by the cmdlet, we later make sure that the Invoce-RestMethod set the auth header with the token. If we add something that looks like a cred, cred scan will complain and we are just testing that it was something :)

tools/devops/automation/scripts/GitHub.Tests.ps1
mandel-macaque: same

tools/devops/automation/templates/configure-build.yml
mandel-macaque: awesome.

mjbond-msft: Also note, if you need the source branch, in the context of a PR trigger, you can also use `$(System.PullRequest.SourceBranch)`, which i believe gives the short name of the topic branch and not the `merge` branch

mjbond-msft: If `$(Build.Reason)` is `manual` (as with all other trigger types in the `else` case) then use `$(Build.SourceVersion)`

PR #10138, opened 2020-11-23
[VSTS] Perform configuration pkgs in the pipeline

mandel-macaque:
Get the vars, parse them, ensure we set a output var for the pipeline. We are not yet deciding what steps to run atm.


tools/devops/provision-brew-packages.csx
```diff
@@ -1,9 +1,11 @@
...
+    "cmake", 
     "autoconf",
     "automake",
     "libtool",
     "p7zip",
     "python",
     "libmagic", 
-    "msitools",
-    "wget");
+    "msitools"
```
spouliot, line 9: no `,` ?

mandel-macaque: good catch, it is working because I disabled the step

mjbond-msft (approved):
👍 Looks good to me

tj-devel709 (approved):

dalexsoto (approved):

PR #10143, opened 2020-11-24
[VSTS] Use the rest API to workaround the fact that we build from forks.

mandel-macaque:


dalexsoto (approved):
Give Manuel a computer and an endpoint and he can bend the CIs to his will

mandel-macaque:
not to be landed since we are missing a configuration from the devdiv side.

mandel-macaque:
Works using the auth as base, I guess that my understanding of a PAT and a token is diff that those in VSTS ;) 
Screenshot 2020-11-30 at 18 42 16

PR #10147, opened 2020-11-24
[msbuild] Default to verbose logs for mmp/mtouch if creating a binary log.

rolfbjarne:
This way we get the full output from mtouch/mmp when the user creates a binary log.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/VerbosityUtils.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/VerbosityUtils.cs
@@ -41,50 +41,58 @@ namespace Xamarin.MacDev.Tasks {
 		//
 		static public string[] GetVerbosityLevel (string commandLine)
 		{
-			const string shortForm = "v:";
-			const string longForm = "verbosity:";
-
 			if (!StringUtils.TryParseArguments (commandLine, out var args, out _))
 				return GetVerbosityLevel (LoggerVerbosity.Normal);
 
+			var hasBinaryLog = false;
 			foreach (var arg in args) {
-				// the minimum length we're looking for is `/v:q`
-				if (arg.Length < 4)
+				// the minimum length we're looking for is `/bl`
+				if (arg.Length < 2)
 					continue;
 				// msbuild accepts two types of argument separator
 				if (arg [0] != '/' && arg [0] != '-')
 					continue;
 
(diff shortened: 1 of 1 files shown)


emaf (approved):

mandel-macaque (approved):

msbuild/Xamarin.MacDev.Tasks.Core/VerbosityUtils.cs
```diff
@@ -41,50 +41,58 @@ static public string[] Merge (string extraArguments, LoggerVerbosity taskVerbosi
...
+					case "d":
+					case "detailed":
+						return GetVerbosityLevel (LoggerVerbosity.Detailed);
+					case "diag":
+					case "diagnostic":
+						return GetVerbosityLevel (LoggerVerbosity.Diagnostic);
+					}
+				case "bl":
+				case "binarylogger":
+					hasBinaryLog = true;
```
spouliot, line 87: Unlike the verbosity options you're not returning when you first see `-bl`, so if you provide `-bl -verbosity:quiet` then you end up with `quiet`.

Not sure what would be the best interpretation but the commit message implies otherwise ("show verbose").

spouliot (changes requested):
maybe it's just changing the commit message to "Have mmp/mtouch default to diagnostic logs if creating a binary log" ?

spouliot:
label update: It's a behaviour change that should be part of the release notes.

msbuild/Xamarin.MacDev.Tasks.Core/VerbosityUtils.cs
rolfbjarne: I changed the commit message as you suggested.

The intention was to default to diagnostic mtouch/mmp unless `-verbosity:...` was passed, but I'm not sure what's the best behavior is, I figured we could just try this and see how it works out. Showing verbose output when asked to be quiet doesn't seem right, but that would also mean the binary log won't have the verbose output either.

spouliot: Parsing conflicting (or less logical) arguments is not easy. The early return here does not match how `mtouch` (or `mmp`) handle multiple arguments - normally the last, not the first, one wins. Still that's the current behaviour...

spouliot (approved):

PR #10151, opened 2020-11-25
[VSTS] Move the build pkgs to a template, since we will need to add extra jobs for the html publishing.

mandel-macaque:


dalexsoto (approved):

tj-devel709 (approved):

mandel-macaque:
Merging, no need to see it compile, we just needed vsts to trigger: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4253913&view=results

PR #10154, opened 2020-11-25
[VSTS] Fix code that add the summary to the vsts UI.

mandel-macaque:


dalexsoto (approved):
Too much C#? :P that `$` always gets us

mandel-macaque:
@dalexsoto is not the first time, won't be the last time

PR #10158, opened 2020-11-25
[ARKit] Update bindings to Xcode 12.3 Beta 1

dalexsoto:


tests/introspection/ApiProtocolTest.cs
```diff
@@ -334,6 +334,8 @@ protected virtual bool Skip (Type type, string protocolName)
 				case "PKShareablePassMetadata":
 				// Xcode 12.2
 				case "VSAccountApplicationProvider": // Conformance not in headers
+				// Xcode 12.2
```
spouliot, line 337: 12.3 ?

tests/introspection/ApiSelectorTest.cs
```diff
@@ -193,6 +193,13 @@ protected virtual bool Skip (Type type, string selectorName)
 				if (Runtime.Arch == Arch.SIMULATOR)
 					return true;
 				break;
+			case "ARBodyTrackingConfiguration":
+			case "ARGeoTrackingConfiguration":
```
spouliot, line 197: works on `ARWorldTrackingConfiguration` ?

spouliot (approved):

tests/introspection/ApiSelectorTest.cs
dalexsoto: The test doesn't complain there, yeah that was a bit weird but device is fine so 🤷‍♂️

tests/introspection/ApiProtocolTest.cs
dalexsoto: indeed

```suggestion
				// Xcode 12.3
```

tj-devel709 (approved):

PR #10171, opened 2020-11-27
[src] Generate a define for each framework for every platform.

rolfbjarne:
Take our list of frameworks per platform, and generate a list of conditional
compiler constants that states that a framework exists for a particular
platform.

Basically generate a list of `-d:HAS_<FRAMEWORK>` for each platform.

This makes it possible to do this:

```csharp
#if HAS_COREMEDIA
using CoreMedia;
#endif
```

instead of this:

```csharp
#if (PLATFORM1 || PLATFORM2) && !PLATFORM3 || PLATFORM4 || BUT_MAYBE_IF_ITS_FRIDAY
using CoreMedia;
#endif
```

Which makes it much easier when adding new platforms (say Mac Catalyst), or
when a framework is added to an already existing platform.

I've included a few examples of how it would end up looking. Changing all the
code to use these new defines is not the goal (nor is it that these defines
should always be used: sometimes they make most sense, sometimes they don't).


spouliot (approved):

dalexsoto (approved):
#if OVERTHERAINBOW

mandel-macaque (approved):

PR #10175, opened 2020-11-27
[bgen] Use ErrorHelper.Verbosity to store the verbosity to avoid storing it in multiple places with different values

rolfbjarne:
This avoids some code using one value, and other code using another value.,
and things don't quite work as expected.

In particular it makes it so that passing -v to bgen makes errors show the
stack trace of the corresponding exception.


spouliot (approved):

rolfbjarne:
build

mandel-macaque (approved):

rolfbjarne:
build

Issue #8901, opened 2020-06-19
[.NET 6] External dotnet issues


rolfbjarne:
This issue lists all issues from any dotnet/* repository that affect our .NET 6 effort (so far [dotnet/runtime](https://github.com/dotnet/runtime), [dotnet/sdk](https://github.com/dotnet/sdk) and [dotnet/roslyn](https://github.com/dotnet/roslyn)).

Issues in the [mono/linker](https://github.com/mono/linker) repository are listed in #12657.

The checkbox should be checked once the external issue is fixed, and the issue
moved down to "Fixed in &lt;external repo&gt;". When we've removed any workarounds or
implemented other related behavior to the issue in question (or verified that nothing else needs to be done)
the issue should be moved down to "Fixed in our repo".

Equivalent issue for .NET 7: #14548.

Known breaking changes in .NET 6+ should be listed here: #8900.
Known breaking changes in CoreCLR should be listed here: #10795.

* P0: something that's blocking other work for us
* P1: something we can't go to stable with
* P2: neither P0 nor P1

### Moved to .NET 7 (#14548).

* [x] [Marshal.AllocHGlobal/FreeHGlobal is ~150x slower in .NET than legacy mono on device (tvOS)](https://github.com/dotnet/runtime/issues/58939): moved to .NET 7

### Fixed in the dotnet repo

* [x] [Mono: the runtime packages for .NET 6+ do not contain the runtime as a .framework for iOS/tvOS/watchOS](https://github.com/dotnet/runtime/issues/42846)
    * Ref: [Ship a .framework version of Mono](https://github.com/dotnet/runtime/pull/53370)

### Fixed in our repo

* [x] ~[XmlSerializer tries to use Reflection.Emit even if RuntimeFeature.IsDynamicCodeSupported is false]~(https://github.com/dotnet/runtime/issues/59167)
    * [x] https://github.com/xamarin/xamarin-macios/pull/14255
* [x] ~[Console.Write prints a new line on iOS](https://github.com/dotnet/runtime/issues/36440)~
* [x] ~[[dotnet] Missing mono-cil-strip (or alternative)](https://github.com/xamarin/xamarin-macios/issues/11445)~
* [x] ~[DOTNETSDK_WORKLOAD_MANIFEST_ROOTS pointing to workloads that are not lowercased fails](https://github.com/dotnet/sdk/issues/19994)~ Worked around.
[...]

PR #10182, opened 2020-11-30
[dotnet] Prevent linking out code referenced by P/Invoke

filipnavara:
Fixes running the following code:

```csharp
	var ciFilter = CIFilter.FromName ("CIColorMatrix");
	var result = (CIImage)ciFilter.ValueForKey (new NSString ("outputImage"));
```

which resulted into call to `xamarin_IntPtr_objc_msgSend_IntPtr` which was linked out from the native code.

This adds a step to dotnet-linker that mimics what mmp/mtouch was doing. Depending on symbol mode setting it either generates: a) references.m file that contains code referencing all the P/Invoked symbols to prevent the native linker from linking them out, or b) `-u<symbol>` linker flags that mark the symbols to be preserved.

### DIFF
--- a/tools/dotnet-linker/SetupStep.cs
+++ b/tools/dotnet-linker/SetupStep.cs
@@ -89,6 +89,7 @@ namespace Xamarin {
 			Steps.Add (new ExtractBindingLibrariesStep ());
 			Steps.Add (new RegistrarStep ());
 			Steps.Add (new GenerateMainStep ());
+			Steps.Add (new GenerateReferencesStep ());
 			Steps.Add (new GatherFrameworksStep ());
 
 			Configuration.Write ();
--- a/tools/dotnet-linker/SetupStep.cs
+++ b/tools/dotnet-linker/SetupStep.cs
@@ -85,6 +85,7 @@ namespace Xamarin {
 				post_sweep_substeps.Add (new RemoveAttributesStep ());
 			}
 
+			Steps.Add (new ListExportedSymbols (null));
 			Steps.Add (new LoadNonSkippedAssembliesStep ());
 			Steps.Add (new ExtractBindingLibrariesStep ());
 			Steps.Add (new RegistrarStep ());
--- a/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
+++ b/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
@@ -38,7 +38,7 @@ namespace Xamarin {
 			var hasSymbols = false;
 			if (assembly.MainModule.HasModuleReferences) {
 				hasSymbols = true;
-			} else if (assembly.MainModule.HasTypeReference (Namespaces.Foundation + ".FieldAttribute")) {
+			} else if (assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute")) {
 				hasSymbols = true;
 			}
 			if (!hasSymbols)
--- a/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
+++ b/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
@@ -10,8 +10,6 @@ using Xamarin.Linker;
 namespace Xamarin {
 
 	public class GenerateReferencesStep : ConfigurationAwareStep {
-		private Symbols required_symbols = new Symbols ();
-		
 		protected override string Name { get; } = "Generate References";
 		protected override int ErrorCode { get; } = 2320;
 
@@ -19,50 +17,32 @@ namespace Xamarin {
 		{
 			base.TryEndProcess ();
 
+			var app = Configuration.Application;
+			var required_symbols = Configuration.DerivedLinkContext.RequiredSymbols;
 			var items = new List<MSBuildItem> ();
-			var file = Path.Combine (Configuration.CacheDirectory, $"references.mm");
-			if (Configuration.Target.GenerateReferencingSource (file, required_symbols) != null) {
-				var item = new MSBuildItem { Include = file };
-				items.Add (item);
-			}
-			Configuration.WriteOutputForMSBuild ("_ReferencesFile", items);
-		}
-
-		protected override void TryProcessAssembly (AssemblyDefinition assembly)
-		{
-			base.TryProcessAssembly (assembly);
-
-			if (!assembly.MainModule.HasTypes)
-				return;
-
-			var hasSymbols = assembly.MainModule.HasModuleReferences || assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute");
-			if (!hasSymbols)
-				return;
-
-			foreach (var type in assembly.MainModule.Types)
-				ProcessType (type);
-		}
-
-		void ProcessType (TypeDefinition type)
-		{
-			if (type.HasNestedTypes) {
-				foreach (var nested in type.NestedTypes)
-					ProcessType (nested);
-			}
-
-			if (type.HasMethods) {
-				foreach (var method in type.Methods)
-					ProcessMethod (method);
-			}
-		}
-
-		void ProcessMethod (MethodDefinition method)
-		{
-			if (method.IsPInvokeImpl && method.HasPInvokeInfo && method.PInvokeInfo != null) {
-				var pinfo = method.PInvokeInfo;
-				if (pinfo.Module.Name == "__Internal") {
-					required_symbols.AddFunction (pinfo.EntryPoint).AddMember (method);
+				
+			switch (app.SymbolMode) {
+			case SymbolMode.Ignore:
+				break;
+			case SymbolMode.Code:
+				string reference_m = Path.Combine (Configuration.CacheDirectory, "reference.m");
+				reference_m = Configuration.Target.GenerateReferencingSource (reference_m, required_symbols);
+				if (!string.IsNullOrEmpty (reference_m)) {
+					var item = new MSBuildItem { Include = reference_m };
+					items.Add (item);
 				}
+				Configuration.WriteOutputForMSBuild ("_ReferencesFile", items);
+				break;
+			case SymbolMode.Linker:
+			case SymbolMode.Default:
+				foreach (var symbol in required_symbols) {
+					var item = new MSBuildItem { Include = "-u" + symbol.Prefix + symbol.Name };
+					items.Add (item);
+				}
+				Configuration.WriteOutputForMSBuild ("_ReferencesLinkerFlags", items);
+				break;
+			default:
+				throw ErrorHelper.CreateError (99, Errors.MX0099, $"invalid symbol mode: {app.SymbolMode}");
 			}
 		}
 	}
--- a/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
+++ b/tools/dotnet-linker/Steps/GenerateReferencesStep.cs
@@ -35,12 +35,7 @@ namespace Xamarin {
 			if (!assembly.MainModule.HasTypes)
 				return;
 
-			var hasSymbols = false;
-			if (assembly.MainModule.HasModuleReferences) {
-				hasSymbols = true;
-			} else if (assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute")) {
-				hasSymbols = true;
-			}
+			var hasSymbols = assembly.MainModule.HasModuleReferences || assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute");
 			if (!hasSymbols)
 				return;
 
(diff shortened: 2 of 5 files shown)


tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,74 @@
...
+		{
+			base.TryProcessAssembly (assembly);
+
+			if (!assembly.MainModule.HasTypes)
+				return;
+
+			var hasSymbols = false;
+			if (assembly.MainModule.HasModuleReferences) {
+				hasSymbols = true;
+			} else if (assembly.MainModule.HasTypeReference (Namespaces.Foundation + ".FieldAttribute")) {
```
spouliot, line 41: `"Foundation.FieldAttribute"` can be used here

The `Namespaces` code was required when we supported classic (`MonoTouch.Foundation` and `MonoMac.Foundation`)

filipnavara:
Hmm, apparently I was running the local test with `--nosymbolstrip` flag and it may affect whether this solution works or not. A unit test like this - https://github.com/xamarin/xamarin-macios/blob/72c7b1ffcc5b6c2a24be53d60a97043db54dc8b2/tests/mtouch/MTouch.cs#L92-L120 - should be adapted.

tools/dotnet-linker/SetupStep.cs
```diff
@@ -89,6 +89,7 @@ protected override void TryProcess ()
 			Steps.Add (new ExtractBindingLibrariesStep ());
 			Steps.Add (new RegistrarStep ());
 			Steps.Add (new GenerateMainStep ());
+			Steps.Add (new GenerateReferencesStep ());
```
spouliot, line 92: Not clear in PR but it seems to be after sweeping/cleaning.
Do check that enabling the linker result in less generated symbols (to keep around)

spouliot:
build

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,74 @@
...
+
+			if (!assembly.MainModule.HasTypes)
+				return;
+
+			var hasSymbols = false;
+			if (assembly.MainModule.HasModuleReferences) {
+				hasSymbols = true;
+			} else if (assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute")) {
+				hasSymbols = true;
+			}
```
mandel-macaque, line 43: could we just do this with an or and do it in the above line? We remove the code branches and we just do later a return according to the value.
```csharp
hasSymbols = assembly.MainModule.HasModuleReferences || assembly.MainModule.HasTypeReference ("Foundation.FieldAttribute");
```

filipnavara: Sure, I copy & pasted this part from the old mmp code.

spouliot:
build

rolfbjarne:
build

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,69 @@
...
+			var items = new List<MSBuildItem> ();
+			var file = Path.Combine (Configuration.CacheDirectory, $"references.mm");
+			if (Configuration.Target.GenerateReferencingSource (file, required_symbols) != null) {
+				var item = new MSBuildItem { Include = file };
+				items.Add (item);
+			}
+			Configuration.WriteOutputForMSBuild ("_ReferencesFile", items);
+		}
+
+		protected override void TryProcessAssembly (AssemblyDefinition assembly)
```
rolfbjarne, line 31: I think we should re-use the existing ListExportedSymbols step instead of duplicating the logic here, because sooner or later we'll also need the rest of the logic in that step.

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,69 @@
...
+		protected override string Name { get; } = "Generate References";
+		protected override int ErrorCode { get; } = 2320;
+
+		protected override void TryEndProcess ()
+		{
+			base.TryEndProcess ();
+
+			var items = new List<MSBuildItem> ();
+			var file = Path.Combine (Configuration.CacheDirectory, $"references.mm");
+			if (Configuration.Target.GenerateReferencingSource (file, required_symbols) != null) {
```
rolfbjarne, line 24: The existing code for iOS has three modes: ignore, generate code (what you're doing), and add linker flags (`-u symbol1 ...`), based on the App.SymbolMode value. This code should do the same (for iOS we want to use linker flags if possible, because that doesn't increase the app size with otherwise unused code).

rolfbjarne (changes requested):

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
filipnavara: That's what I initially planned to do. Then I discovered that not all of the logic is applicable anymore in the .NET 6 context so I went with an approach that kept the code to absolute minimum. I am happy to rewrite it to use `ListExportedSymbols` but I need to study the `PInvokeWrapperGenerator` and related code first. Also, all the code for `System.*.Native.*` libraries is not necessary in the macOS context since the libraries are linked dynamically.

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
filipnavara: Agreed. This was just proof of concept which tried to do minimal amount of work to see if it fixes my problem.

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,49 @@
...
+				if (!string.IsNullOrEmpty (reference_m)) {
+					var item = new MSBuildItem { Include = reference_m };
+					items.Add (item);
+				}
+				Configuration.WriteOutputForMSBuild ("_ReferencesFile", items);
+				break;
+			case SymbolMode.Linker:
+			case SymbolMode.Default:
+				foreach (var symbol in required_symbols) {
+					var item = new MSBuildItem { Include = "-u" + symbol.Prefix + symbol.Name };
```
rolfbjarne, line 39: I think you need a space here:

```suggestion
					var item = new MSBuildItem { Include = "-u " + symbol.Prefix + symbol.Name };
```

although that might have to be treated as two different MSBuildItems.

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
```diff
@@ -0,0 +1,49 @@
...
+				string reference_m = Path.Combine (Configuration.CacheDirectory, "reference.m");
+				reference_m = Configuration.Target.GenerateReferencingSource (reference_m, required_symbols);
+				if (!string.IsNullOrEmpty (reference_m)) {
+					var item = new MSBuildItem { Include = reference_m };
+					items.Add (item);
+				}
+				Configuration.WriteOutputForMSBuild ("_ReferencesFile", items);
+				break;
+			case SymbolMode.Linker:
+			case SymbolMode.Default:
```
rolfbjarne, line 37: The default mode should already be changed to one of the other modes:

https://github.com/xamarin/xamarin-macios/blob/725190243995c63d26c89c2a200f1cea07fb82b6/tools/common/Application.cs#L673-L679

so the `Default` case can be removed here

tools/linker/MonoTouch.Tuner/ListExportedSymbols.cs
```diff
@@ -10,18 +10,26 @@
...
 
-namespace MonoTouch.Tuner
+namespace Xamarin.Linker.Steps
 {
 	public class ListExportedSymbols : BaseStep
 	{
 		PInvokeWrapperGenerator state;
 		bool skip_sdk_assemblies;
 
+#if NET
```
rolfbjarne, line 20: You don't need the #if around the entire property, only around the implementation:

```csharp
#if NET
    return LinkerConfiguration.GetInstance (Context).DerivedLinkContext;
#else
    return (DerivedLinkContext) Context;
#endif
```

rolfbjarne (changes requested):
Just a few minor things left, this is looking very good!

tools/dotnet-linker/Steps/GenerateReferencesStep.cs
filipnavara: I actually checked it and the space is not necessary. In fact the space would create a problem because it would pass the whole thing quoted to the linker and the linker would expect the space to be part of the symbol name.

rolfbjarne:
build

filipnavara:
The failed test is related:
```
20.5795130                      Undefined symbols for architecture x86_64:
10:53:20.5795270                        "_NetSecurityNative_AcceptSecContext", referenced from:
10:53:20.5795360                           -u command line option
10:53:20.5795440                        "_NetSecurityNative_DeleteSecContext", referenced from:
10:53:20.5795520                           -u command line option
10:53:20.5795590                        "_NetSecurityNative_DisplayMajorStatus", referenced from:
10:53:20.5795670                           -u command line option
10:53:20.5795750                        "_NetSecurityNative_DisplayMinorStatus", referenced from:
10:53:20.5795830                           -u command line option
10:53:20.5795910                        "_NetSecurityNative_GetUser", referenced from:
10:53:20.5795990                           -u command line option
10:53:20.5796060                        "_NetSecurityNative_ImportPrincipalName", referenced from:
10:53:20.5796140                           -u command line option
10:53:20.5796230                        "_NetSecurityNative_ImportUserName", referenced from:
10:53:20.5796320                           -u command line option
10:53:20.5796590                        "_NetSecurityNative_InitSecContext", referenced from:
10:53:20.5796700                           -u command line option
10:53:20.5796790                        "_NetSecurityNative_InitSecContextEx", referenced from:
10:53:20.5796870                           -u command line option
10:53:20.5796940                        "_NetSecurityNative_InitiateCredSpNego", referenced from:
10:53:20.5797020                           -u command line option
10:53:20.5797100                        "_NetSecurityNative_InitiateCredWithPassword", referenced from:
10:53:20.5797180                           -u command line option
10:53:20.5797290                        "_NetSecurityNative_IsNtlmInstalled", referenced from:
10:53:20.5797370                           -u command line option
10:53:20.5797450                        "_NetSecurityNative_ReleaseCred", referenced from:
10:53:20.5797520                           -u command line option
10:53:20.5797600                        "_NetSecurityNative_ReleaseGssBuffer", referenced from:
10:53:20.5797680                           -u command line option
10:53:20.5797760                        "_NetSecurityNative_ReleaseName", referenced from:
10:53:20.5797840                           -u command line option
10:53:20.5797910                        "_NetSecurityNative_Unwrap", referenced from:
10:53:20.5798120                           -u command line option
10:53:20.5798220                        "_NetSecurityNative_Wrap", referenced from:
10:53:20.5798310                           -u command line option
```

filipnavara:
Not sure how to solve it. Seems like System.Net.Security.Native is missing from tvOS builds (https://github.com/dotnet/runtime/blob/3c39a5d3b8310d9adcb906718cf985d3e40122da/src/libraries/Native/Unix/CMakeLists.txt#L224) so the failure is somewhat warranted. I can place `#if !NET` around the code in ListExportedSymbols.cs but it will only hide the problem.

rolfbjarne:
> Not sure how to solve it. Seems like System.Net.Security.Native is missing from tvOS builds (https://github.com/dotnet/runtime/blob/3c39a5d3b8310d9adcb906718cf985d3e40122da/src/libraries/Native/Unix/CMakeLists.txt#L224) so the failure is somewhat warranted. I can place `#if !NET` around the code in ListExportedSymbols.cs but it will only hide the problem.

Something like this:

```csharp
case "System.Net.Security.Native":
#if NET
    if (app.Platform == ApplePlatform.tvOS)
        break; // tvOS does not ship with System.Net.Security.Native due to https://github.com/dotnet/runtime/issues/####
#endif
```

basically file an issue and hide the problem (and add the issue to the list here: https://github.com/xamarin/xamarin-macios/issues/8901)

filipnavara:
Done. You'll have to link https://github.com/dotnet/runtime/issues/45535 in #8901 yourself since I cannot other people's issues :-)

rolfbjarne:
build

rolfbjarne (approved):

dpozimski:
can someone from the xamarin team look at the [https://github.com/xamarin/xamarin-macios/issues/10884](url) issue?
it's possible that this PR created issue with build error in ios release mode

PR #10185, opened 2020-11-30
[dotnet] Avoid double registration of static registrar

filipnavara:
The method `xamarin_create_classes_Xamarin_Mac` was called twice which resulted in corruption of the type map. The first call was made from the generated `xammac_setup` method. The second call was made through registration method added in the RegistrarStep of dotnet-linker. This PR removes the first call for consistency with iOS and other platforms.

### DIFF
--- a/tools/common/Target.cs
+++ b/tools/common/Target.cs
@@ -558,8 +558,10 @@ namespace Xamarin.Bundler {
 		{
 			sw.WriteLine ("#define MONOMAC 1");
 			sw.WriteLine ("#include <xamarin/xamarin.h>");
+#if !NET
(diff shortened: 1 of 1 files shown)


spouliot:
build

mandel-macaque (approved):

mandel-macaque:
@filipnavara you can ignore that failure :)

I wonder who wrote that test:

```
GHIssue8342(OK,"mandel","12345678","mandel","12345678") : Status not ok
```

Is testing a network request, so probably is a network issue and not related to the PR since it would raise in all others.

rolfbjarne (approved):

PR #10204, opened 2020-12-03
[VSTS] Ensure that we have a reasonable timeout. Will unblock device tests.

mandel-macaque:
Some bots are not configure correctly and block, that means we wait for ever. Just wait for 30 mins max.


mandel-macaque:
@monojenkins backport d16-8
@monojenkins backport xcode12.3

dalexsoto (approved):

tj-devel709 (approved):

PR #10207, opened 2020-12-03
[VSTS] Machines can give problems and the default image does have enough HDD.

mandel-macaque:


mandel-macaque:
Sample build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4271290&view=results

mjbond-msft (approved):
Looks good

PR #10208, opened 2020-12-04
[VSTS] Add branchname in the nugets.

mandel-macaque:
If the env is not set, we have names such as
'Microsoft.watchOS.Runtime.watchos-x86.7.1.100-ci.HEAD.165+sha.14522b296.nupkg'


mjbond-msft (approved):
👍

PR #10218, opened 2020-12-04
Replace manually maintained code about frameworks with generated code.

rolfbjarne:
* [src] Generate the string constants for each framework from our list of
  frameworks.

  This makes it less error prone to add new frameworks (one less place to add 
  them), and easier to add new platforms (only one place to make sure the
  list of frameworks is correct).

  This also revelead a few issues in our list of frameworks (missing
  frameworks, frameworks that have been moved, etc.), which have been fixed.

* [mmp] Generate the library path -> namespace map for the linker to optimize away dlopen statements.


src/Makefile
```diff
@@ -725,6 +728,7 @@ WATCHOS_WARNINGS_TO_FIX = -nowarn:219,618,114,414,1635,3021,$(IOS_WARNINGS_THAT_
 
 WATCHOS_EXTRA_CORE_SOURCES = \
     $(WATCH_BUILD_DIR)/Constants.cs    \
+	$(BUILD_DIR)/Constants.watchos.generated.cs \
```
spouliot, line 731: indent ?

src/generate-frameworks-constants/Program.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+using Xamarin.Utils;
+
+namespace GenerateTypeForwarders {
```
spouliot, line 8: ^ weird name ;-)

src/generate-frameworks-constants/generate-frameworks-constants.csproj
```diff
@@ -0,0 +1,48 @@
...
+    </Compile>
+    <Compile Include="..\..\tools\common\ApplePlatform.cs">
+      <Link>ApplePlatform.cs</Link>
+    </Compile>
+    <Compile Include="..\..\tools\common\StringUtils.cs">
+      <Link>StringUtils.cs</Link>
+    </Compile>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Mono.Cecil" Version="0.11.3" />
```
spouliot, line 45: did not see Cecil used

spouliot (commented):
❤️  it

maybe `src/generate-frameworks-constants` should be under `tools` ? that's where I would look for it instinctively

rolfbjarne:
> ❤️ it
> 
> maybe `src/generate-frameworks-constants` should be under `tools` ? that's where I would look for it instinctively

The problem is that tools/ is built after src/:

https://github.com/xamarin/xamarin-macios/blob/14ad1a122bd1bacae402dc4a2337eed438cd2a71/Makefile#L2

So yes, I could move it to tools/, but I would still have to build it from src/ (something like `make -C $(TOP)/tools/generate-frameworks-constants`), and that's somewhat ugly too, so I decided to just put this in src/ (but either way would work fine for me).

src/generate-frameworks-constants/Program.cs
rolfbjarne: c&p...

rolfbjarne:
The breaking API changes are expected:

- public const string CFNetworkLibrary = ~"/System/Library/Frameworks/CoreServices.framework/Frameworks/CFNetwork.framework/CFNetwork"~ "/System/Library/Frameworks/CFNetwork.framework/CFNetwork";

Unsure if this was a mistake or if CFNetwork was a part of CoreServices at some point, but in any case CFNetwork is not a subframework of CoreServices anymore (in 10.15).

- public const string CallKitLibrary = ~"/System/Library/Frameworks/CallKitLibrary.framework/CallKitLibrary"~ "/System/Library/Frameworks/CallKit.framework/CallKit";

The previous value was just wrong.

- public const string CoreBluetoothLibrary = ~"/System/Library/Frameworks/IOBluetooth.framework/Versions/A/Frameworks/CoreBluetooth.framework/CoreBluetooth"~ "/System/Library/Frameworks/CoreBluetooth.framework/CoreBluetooth";

Moved from inside an umbrella framework (IOBluetooth) into its own framework in 10.10

- public const string CoreImageLibrary = ~"/System/Library/Frameworks/QuartzCore.framework/Versions/A/Frameworks/CoreImage.framework/CoreImage"~ "/System/Library/Frameworks/CoreImage.framework/CoreImage";

Moved from inside an umbrella framework (QuartzCore) into its own framework in 10.11

- public const string CoreTextLibrary = ~"/System/Library/Frameworks/ApplicationServices.framework/Frameworks/CoreText.framework/CoreText"~ "/System/Library/Frameworks/CoreText.framework/CoreText";

It's own framework since at least 10.9

- public const string ImageIOLibrary = ~"/System/Library/Frameworks/ApplicationServices.framework/Frameworks/ImageIO.framework/ImageIO"~ "/System/Library/Frameworks/ImageIO.framework/ImageIO";

It's own framework since at least 10.9

- public const string ImageKitLibrary = ~"/System/Library/Frameworks/Quartz.framework/Frameworks/ImageKit.framework/ImageKit"~ "/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/ImageKit.framework/ImageKit";

The same path, because of symlinks (now follows the same pattern as other paths).

- public const string PdfKitLibrary = ~"/System/Library/Frameworks/Quartz.framework/Frameworks/PDFKit.framework/PDFKit"~ "/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/PDFKit.framework/PDFKit";

The same path, because of symlinks (now follows the same pattern as other paths).

- public const string QuartzComposerLibrary = ~"/System/Library/Frameworks/Quartz.framework/Frameworks/QuartzComposer.framework/QuartzComposer"~ "/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/QuartzComposer.framework/QuartzComposer";

The same path, because of symlinks (now follows the same pattern as other paths).

- public const string QuickLookUILibrary = ~"/System/Library/Frameworks/Quartz.framework/Frameworks/QuickLookUI.framework/QuickLookUI"~ "/System/Library/Frameworks/Quartz.framework/Versions/A/Frameworks/QuickLookUI.framework/QuickLookUI";

The same path, because of symlinks (now follows the same pattern as other paths).

- public const string SearchKitLibrary = ~"/System/Library/Frameworks/CoreServices.framework/Frameworks/SearchKit.framework/SearchKit"~ "/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/SearchKit.framework/SearchKit";

The same path, because of symlinks (now follows the same pattern as other paths).

rolfbjarne:
build

src/generate-frameworks-constants/generate-frameworks-constants.csproj
rolfbjarne: Fixed.

spouliot (approved):

dalexsoto (approved):
Nice!

PR #10243, opened 2020-12-09
[VSTS] Add a postfix to the artifacts and download accordingly

mandel-macaque:
Since we are moving all the stages to the same pipeline, we need to make sure that the different devices and simulator tests do not step on each other when they upload the results to be later uploaded to vsdrops.


spouliot (approved):

dalexsoto (approved):

tools/devops/automation/templates/download-artifacts.yml
```diff
@@ -9,27 +14,29 @@ steps:
...
   inputs:
-    patterns: '**/TestSummary.md'
+    patterns: '**/TestSummary-${{ parameters.devicePrefix }}/TestSummary.md'
     allowFailedBuilds: true
     path: $(System.DefaultWorkingDirectory)\Reports
 
 - powershell: |
-    echo "##vso[task.setvariable variable=TEST_SUMMARY_PATH]$Env:SYSTEM_DEFAULTWORKINGDIRECTORY\Reports\TestSummary\TestSummary.md"
-    echo "##vso[task.setvariable variable=HTML_REPORT_PATH]$Env:SYSTEM_DEFAULTWORKINGDIRECTORY\HtmlReport"
+    Get-ChildItem -Recurse $Env:SYSTEM_DEFAULTWORKINGDIRECTORY
```
mjbond-msft, line 37: I assume this is here for diagnostics. Just noting it in case you did not intend to leave it in. However, if generally useful for providing insights into the directory contents then certainly keep it.

mjbond-msft (approved):
Just left a comment regarding diagnostic code in case you did not intend to keep it; otherwise, looks good

tools/devops/automation/templates/download-artifacts.yml
mandel-macaque: it is useful at the moment since we need to add some extra steps that extend this templates... so I want to be avle to see what is downloaded. Should delete later.

PR #10244, opened 2020-12-09
[CallKit] Add to macOS.

rolfbjarne:


dalexsoto (approved):

tools/common/StaticRegistrar.cs
```diff
@@ -2123,6 +2123,20 @@ void CheckNamespace (string ns, List<Exception> exceptions)
...
 			string h;
 			switch (ns) {
+			case "CallKit":
+				if (App.Platform == ApplePlatform.MacOSX) {
+					// AVFoundation can't be imported before CallKit on macOS
+					// Ref: https://github.com/xamarin/maccore/issues/2301
+					// Ref: https://github.com/xamarin/maccore/issues/2257
+					// The fun part is that other frameworks can import AVFoundation, so we can't check for AVFoundation specifically.
+					// Instead add CallKit before any other imports.
+					var firstImport = header.StringBuilder.ToString ().IndexOf ("#import <");
```
spouliot, line 2133: it's a shame .net does not have an `IndexOf` method for `StringBuilder` :(

spouliot (approved):

mandel-macaque (approved):

tools/common/StaticRegistrar.cs
Therzok: You could technically use GetChunks if targeting NS2.1.

PR #10280, opened 2020-12-14
[VSTS] Remove all the device pipelines.

mandel-macaque:
With the migration away from Jenkins, we do not longer ping them.


mandel-macaque:
You will not be forgotten :(

spouliot (approved):
👋 🤖

dalexsoto (approved):
You will be remembered as part of our git history, farewell 👋

PR #10281, opened 2020-12-15
[macos] Handle symlinks in user frameworks

spouliot:
User frameworks for macOS often uses symlinks (as Xcode creates them
this way).

This cause problem cause the symlink is on the binary and we expected
the `_CodeSignature` directory to by side-by-side with the binary. This
was missing and cause exceptions when codesigning such frameworks.

A second problem happened because `mmp` use `lipo -thin` to remove
non-required architectures. However when a framework has symlinks, like:

```
├── Frameworks
│   └── Universal.framework
│       ├── Resources -> Versions/Current/Resources
│       ├── Universal -> Versions/Current/Universal
│       └── Versions
│           ├── A
│           │   ├── Resources
│           │   │   └── Info.plist
│           │   ├── Universal
│           │   └── _CodeSignature
│           │       └── CodeResources
│           └── Current -> A
```

then this actually replaced the (very small) symlink with a thin version
of the framework. Which means the original one was still _fat_ and the
whole app was now larger than the original version.

Sample used: https://github.com/spouliot/xcframework/tree/main/xamarin/xcf-mac

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
@@ -354,6 +354,7 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 			ProjectGuid="$(ProjectGuid)"
 			ProjectTypeGuids="$(ProjectTypeGuids)"
 			OutputPath="$(OutputPath)"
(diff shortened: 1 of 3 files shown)


rolfbjarne (approved):

dalexsoto (approved):
symlinks everywhere!

stephen-hawley (approved):

tj-devel709 (approved):

emaf (approved):

PR #10301, opened 2020-12-15
Yaml pipeline triggers: Enable CI trigger for all topic branches

mjbond-msft:


mandel-macaque (commented):
Works! https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4305514&view=results

PR #10264, opened 2020-12-11
[compare-commits] Rework logic to build xamarin-macios again from a separate clone.

rolfbjarne:
Rework the code to generate an API comparison and a generator comparison: we
now clone xamarin-macios into a different directory, checks out the hash we
want to compare against, build the whole thing, and then compare the results.

This is somewhat slower than the previous implementation (which would checkout
the previous hash and build that on top of the existing build), but since we
don't have to build mono, it shouldn't make too much of a difference (the
biggest chunk of our build is building src/, and we had to do that anyway with
the previous implementation as well).

The upside is that it should be much more reliable.

Fixes https://github.com/xamarin/maccore/issues/2034.
Fixes https://github.com/xamarin/maccore/issues/1467.
Fixes https://github.com/xamarin/maccore/issues/1059.
Maybe https://github.com/xamarin/maccore/issues/1173 too.

PR #10306, opened 2020-12-16
[main] Merge xcode12.3 into main

dalexsoto:


Make.versions
```diff
@@ -66,8 +66,7 @@ MAC_PACKAGE_VERSION=7.3.0.$(MAC_COMMIT_DISTANCE)
 # WARNING: Do **not** use versions higher than the available Xcode SDK or else we will have issues with mtouch (See https://github.com/xamarin/xamarin-macios/issues/7705)
 # When bumping the major macOS version in MACOS_NUGET_VERSION also update the macOS version where we execute on bots in jenkins/Jenkinsfile (in the 'node' element)
 
-IOS_NUGET_VERSION=14.2.100
-TVOS_NUGET_VERSION=14.2.100
-WATCHOS_NUGET_VERSION=7.1.100
-MACOS_NUGET_VERSION=11.0.100
-MACCATALYST_NUGET_VERSION=14.2.100
```
rolfbjarne, line 73: MACCATALYST_NUGET_VERSION disappeared!

src/frameworks.sources
```diff
@@ -1952,6 +1953,7 @@ IOS_FRAMEWORKS =         \
 	$(COMMON_FRAMEWORKS) \
 	AVKit \
 	Accounts \
+	AdServices \
```
rolfbjarne, line 1956: Does `AdServices` exist for Mac Catalyst as well?

rolfbjarne (changes requested):

Make.versions
dalexsoto: whoops, nice catch

src/frameworks.sources
dalexsoto: Yes according to docs [Mac Catalyst 14.3+](https://developer.apple.com/documentation/AdServices?language=objc)

Make.config
```diff
@@ -36,7 +36,7 @@ $(TOP)/Make.config.inc: $(TOP)/Make.config $(TOP)/mk/mono.mk
 
 include $(TOP)/Make.versions
 
-APIDIFF_REFERENCES=https://bosstoragemirror.blob.core.windows.net/wrench/jenkins/d16-8/e30c41de32f64eca655953b17c78c1d9d70249e3/39/package/bundle.zip
+APIDIFF_REFERENCES=https://bosstoragemirror.blob.core.windows.net/wrench/jenkins/d16-8/87a1b18d8890ab76c4c658abd331dfde68d630ab/49/package/bundle.zip
```
spouliot, line 39: we might want to compare with the inserted package for xcode12.3
or at least make a note, somewhere, to do it later - once published

dalexsoto:
@rolfbjarne Fixed, thanks!

spouliot (approved):

src/frameworks.sources
rolfbjarne: Then you need an entry in the list of Mac Catalyst frameworks in this file (`MACCATALYST_FRAMEWORKS`).

dalexsoto: @rolfbjarne done, thank you!

rolfbjarne (approved):

tj-devel709 (approved):

rolfbjarne:
build

dalexsoto:
Merging since everything else is green, will make another PR to verify the API diff, it looks like we hit an unforeseen case for the API diff https://github.com/xamarin/xamarin-macios/pull/10264/files#diff-23232b8bb6d3c19948ebd0a6a98bb7278ff2b7db4f145460e8a1e65ff5f3d602R171 triggered by the change of Xcode, @rolfbjarne can you have a look? Maybe we need a way to ignore the check or account for it.

rolfbjarne:
> @rolfbjarne can you have a look? Maybe we need a way to ignore the check or account for it.

It's a foreseen case, just not a supported one :smile: It shouldn't happen all that often, and if it turns out to be a problem we can fix it. Although it probably shouldn't be an error, it could be just a warning instead. I'll see if I can change that at some point.

PR #10318, opened 2020-12-18
[mtouch] Allow referencing Xamarin.iOS.dll in a Mac Catalyst app.

rolfbjarne:


dalexsoto (approved):
Interesting codepath but makes sense 👍

PR #10321, opened 2020-12-18
[Security] Fix the underlying integral type for SslCipherSuite for Mac Catalyst.

rolfbjarne:
This difference manifested as a memory corruption (because we were passing an array of the enum to native code, and the array was half the size it should be, so native code write into random memory).

### DIFF
--- a/src/Security/SecureTransport.cs
+++ b/src/Security/SecureTransport.cs
@@ -275,7 +275,7 @@ namespace Security {
 	[Deprecated (PlatformName.iOS, 13,0, message: "Use 'TlsCipherSuite' instead.")]
 	[Deprecated (PlatformName.TvOS, 13,0, message: "Use 'TlsCipherSuite' instead.")]
 	[Deprecated (PlatformName.WatchOS, 6,0, message: "Use 'TlsCipherSuite' instead.")]
-#if MONOMAC
+#if MONOMAC || __MACCATALYST__
 	public enum SslCipherSuite : uint {
 #else
 	public enum SslCipherSuite : ushort {


dalexsoto (approved):

tj-devel709 (approved):

src/Security/SecureTransport.cs
```diff
@@ -275,7 +275,7 @@ public enum SslClientCertificateState {
 	[Deprecated (PlatformName.iOS, 13,0, message: "Use 'TlsCipherSuite' instead.")]
 	[Deprecated (PlatformName.TvOS, 13,0, message: "Use 'TlsCipherSuite' instead.")]
 	[Deprecated (PlatformName.WatchOS, 6,0, message: "Use 'TlsCipherSuite' instead.")]
-#if MONOMAC
+#if MONOMAC || __MACCATALYST__
```
spouliot, line 278: that actually hurts binary compatibility :(
considering it's already deprecated it might be better to add an `[Advice]` and fix the call sites
but I guess it all depends on just how many cases we have to deal with...

mandel-macaque (approved):
But I believe @spouliot to be right

src/Security/SecureTransport.cs
rolfbjarne: Yeah, there are no good choices. There are a few P/Invokes that would have to be special-cased for Mac Catalyst:

```
$ git grep extern.*SslCipherSuite[^G]
Security/SecProtocolMetadata.cs:                extern static SslCipherSuite sec_protocol_metadata_get_negotiated_ciphersuite (IntPtr handle);
Security/SecProtocolOptions.cs:         static extern void sec_protocol_options_add_tls_ciphersuite (sec_protocol_options_t handle, SslCipherSuite cipherSuite);
Security/SslContext.cs:         extern unsafe static /* OSStatus */ SslStatus SSLGetSupportedCiphers (/* SSLContextRef */ IntPtr context, SslCipherSuite *ciphers, /* size_t* */ ref nint numCiphers);
Security/SslContext.cs:         extern unsafe static /* OSStatus */ SslStatus SSLGetEnabledCiphers (/* SSLContextRef */ IntPtr context, SslCipherSuite *ciphers, /* size_t* */ ref nint numCiphers);
Security/SslContext.cs:         extern unsafe static /* OSStatus */ SslStatus SSLSetEnabledCiphers (/* SSLContextRef */ IntPtr context, SslCipherSuite *ciphers, /* size_t */ nint numCiphers);
Security/SslContext.cs:         extern unsafe static /* OSStatus */ SslStatus SSLGetNegotiatedCipher (/* SSLContextRef */ IntPtr context, /* SslCipherSuite* */ out SslCipherSuite cipherSuite);
```

And we'd have to hope that no user code have such P/Invokes.

There's also a BindAs attribute in Foundation.

```
foundation.cs:          [BindAs (typeof (SslCipherSuite?))]
foundation.cs-          NSNumber NegotiatedTlsCipherSuite { get; }
```

 That said, we're not going to be 100% binary compatible with Xamarin.iOS assemblies (this is one of the reasons we're probably going with a new TFM for Mac Catalyst), it will be best-effort and cross-your-fingers approach.

spouliot: Let's create an issue to build a list of those incompatibilities (including PR links if applicable).

We can later decide to use it for documentation (only) purpose;
or see if it's worth the effort to kill/close the gap between the public API of both platforms.

rolfbjarne: > Let's create an issue to build a list of those incompatibilities (including PR links if applicable).
> 
> We can later decide to use it for documentation (only) purpose;
> or see if it's worth the effort to kill/close the gap between the public API of both platforms.

https://github.com/xamarin/xamarin-macios/issues/10377

PR #10340, opened 2021-01-05
[CI] Point to blob storage now that we cannot access xamarin-storage.

mandel-macaque:


mandel-macaque:
No needed now that we have VSTS 'running'

PR #10349, opened 2021-01-06
[CI][VSTS] Add trigger for xcode branches.

mandel-macaque:
fixes: https://github.com/xamarin/maccore/issues/2355

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -71,6 +71,7 @@ pr:
     include:
     - main
     - d16-*
+    - xcode*
 
 stages:
 


dalexsoto (approved):

dalexsoto:
is there a way to build any PR if it comes from project maintainers and have external contributors PR's built only if we approve? If not, is there an issue filled for that?

tj-devel709 (approved):

mandel-macaque:
This is something I have talked with mbond, we need to set the triggers better, but might be that VSTS is not considered for this scenario 😞

dalexsoto:
@mandel-macaque can you make an issue so we do not forget? we really need a way to make this work

PR #10350, opened 2021-01-06
[CI][VSTS] Add comments to PR not only to commits.

mandel-macaque:
The GitHup url to be used to create comments in PRs is diff to the one
for comments. Use the build reason AND the changeID to identify if we
are building due to a PR and use the correct url.

fixes: https://github.com/xamarin/maccore/issues/2356

### DIFF
--- a/tools/devops/automation/scripts/GitHub.psm1
+++ b/tools/devops/automation/scripts/GitHub.psm1
@@ -170,6 +170,8 @@ function New-GitHubComment {
         "SYSTEM_TEAMPROJECT" = $Env:SYSTEM_TEAMPROJECT;
         "BUILD_DEFINITIONNAME" = $Env:BUILD_DEFINITIONNAME;
         "BUILD_REVISION" = $Env:BUILD_REVISION;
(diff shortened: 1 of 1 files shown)


tools/devops/automation/scripts/GitHub.psm1
```diff
@@ -193,7 +195,16 @@ function New-GitHubComment {
     $msg.AppendLine()
     $msg.AppendLine("[Pipeline]($targetUrl) on Agent $Env:TESTS_BOT") # Env:TESTS_BOT is added by the pipeline as a variable coming from the execute tests job
 
-    $url = "https://api.github.com/repos/xamarin/xamarin-macios/commits/$Env:BUILD_REVISION/comments"
+    # if the build was due to PR, we want to write the comment in the PR rather than in teh comment
```
dalexsoto, line 198:
```suggestion
    # if the build was due to PR, we want to write the comment in the PR rather than in the comment
```

😛

dalexsoto (approved):

mandel-macaque:
This exact PR is a perfect example that it works :)

PR #10376, opened 2021-01-11
[msbuild] Simplify resolving xcframeworks

spouliot:
TD&LR: This PR simplifies how we refer to user frameworks and fixes both
warnings and non-optimal (app) output.

Much longer story:

Additional testing on macOS showed some build-time warnings and an
[extra (dupe) file](https://github.com/spouliot/xcframework/commit/a20f8aba419e32e2338cb7c5e4b45d0f38862893#diff-54fd7d9cd5deae57f30195be0a43133eace03c1132401741a317e0ae8d5e13fdR34).

Logs shows that we referred to the xcframework several times, where once
should have been enough.

```
/native-reference:/Users/poupou/git/spouliot/xcframework/Universal.xcframework
/native-reference:/Users/poupou/git/spouliot/xcframework/Universal.xcframework/macos-arm64_x86_64/Universal.framework
/native-reference:/Users/poupou/git/spouliot/xcframework/Universal.xcframework/macos-arm64_x86_64/Universal.framework/Universal
```

The first `/native-reference` line produced a warning like:

```
MMP warning MM2006: Native library 'Universal.xcframework' was referenced but could not be found.
```

which makes sense as the tools (both `mmp` and `mtouch`) are not, by
design, aware of (unresolved) xcframeworks.

Removing `{NativeReference}` from `Xamarin.Mac.Common.targets` (and
`Xamarin.iOS.Common.targets`) as it has already been processed by
`_ExpandNativeReferences` solves this.

The other part of the issue (next two lines) is because `msbuild` does
not track changes to directories like it does for files - and the
workaround (in `_ExpandNativeReferences`) had to be copied in other
places (both XI and XM `_CompileToNative`) and that was not enough (and
would eventually need to be duplicated again and again).

This could lead to duplicate entries (i msbuild logs) like

```
NativeReferences
../../Universal.xcframework/macos-arm64_x86_64/Universal.framework
../../Universal.xcframework/macos-arm64_x86_64/Universal.framework/Univeral
```
which maps to our extra entries.

In order to simplify things we make the `_ExpandNativeReferences` resolve
the full path to the library name (not the `.framework` directory) which
simplifies both `_CompileToNative` and ensure a single way (at least for
`msbuild`) to provide this data to the tools (`mmp` and `mtouch`).

Using a file, instead of a directory, is also more consistent for the
existing `-framework` option, e.g. we provide the names like:

```
--framework=CoreLocation
--framework=ModelIO
```

So adding a full path that include the name is more appropriate, e.g.

```
--framework=/Users/poupou/git/master/xamarin-macios/tests/xharness/tmp-test-dir/xcframework-test760/bin/AnyCPU/Debug/bindings-xcframework-test.resources/XTest.xcframework/ios-i386_x86_64-simulator/XTest.framework/XTest
```

Finally for macOS applications it turns out we were embedding yet another
copy of the framework's library inside the `MonoBundle`, which is clearly
wrong, because of the last entry.

```
$ l bin/Release/xcf-mac.app/Contents/MonoBundle/Universal
-rwxr-xr-x  1 poupou  staff  167152  2 Dec 16:16 bin/Release/xcf-mac.app/Contents/MonoBundle/Universal
```

The tool now checks if a provided library is inside a framework (or not)
which is a good validation to have anyway when it gets called directly,
i.e. not thru `msbuild`.

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
@@ -205,10 +205,6 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		Inputs="@(_CompileToNativeInput);@(_FileNativeReference);@(BundleDependentFiles)"
 		Outputs="$(_AppBundlePath)Contents\MacOS\$(_AppBundleName);$(DeviceSpecificOutputPath)bundler.stamp">
 
-		<ItemGroup>
-			<_FrameworkNativeReference Include="@(_FrameworkNativeReference -> '%(Identity)/%(Filename)')" Condition="'%(Extension)' == '.framework' Or '%(Extension)' == '.xcframework'" />
-		</ItemGroup>
-
 		<Mmp
 			Condition="'$(IsMacEnabled)' == 'true'"
 			SessionId="$(BuildSessionId)"
@@ -227,7 +223,7 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 			I18n="$(I18n)"
 			Profiling="$(Profiling)"
 			ExtraArgs="$(_BundlerArguments)"
-			NativeReferences="@(NativeReference);@(_FrameworkNativeReference);@(_FileNativeReference)"
+			NativeReferences="@(_FrameworkNativeReference);@(_FileNativeReference)"
 			References="@(ReferencePath);@(_BundlerReferencePath)"
 			ResponseFilePath="$(IntermediateOutputPath)response-file.rsp"
 			SdkRoot="$(_SdkDevPath)"
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ResolveFrameworksBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ResolveFrameworksBase.cs
@@ -20,11 +20,11 @@ namespace Xamarin.MacDev.Tasks {
 		#region Inputs
 
 		[Required]
-		public string Architectures { get; set; }
+		public string? Architectures { get; set; }
 
-		public ITaskItem [] NativeReferences { get; set; }
+		public ITaskItem []? NativeReferences { get; set; }
 
-		public ITaskItem[] References { get; set; }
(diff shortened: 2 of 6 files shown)


msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
```diff
@@ -205,10 +205,6 @@ Copyright (C) 2014 Xamarin. All rights reserved.
 		Inputs="@(_CompileToNativeInput);@(_FileNativeReference);@(BundleDependentFiles)"
 		Outputs="$(_AppBundlePath)Contents\MacOS\$(_AppBundleName);$(DeviceSpecificOutputPath)bundler.stamp">
 
-		<ItemGroup>
-			<_FrameworkNativeReference Include="@(_FrameworkNativeReference -> '%(Identity)/%(Filename)')" Condition="'%(Extension)' == '.framework' Or '%(Extension)' == '.xcframework'" />
```
rolfbjarne, line 209: Should the identical code in Xamarin.iOS.Common.targets be removed too?

rolfbjarne (approved):

dalexsoto (approved):

mandel-macaque:
Test failures are unrelated: https://github.com/xamarin/maccore/issues/2357

mandel-macaque (approved):

msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets
spouliot: definitively!

rolfbjarne (approved):

tj-devel709 (approved):

spouliot:
Original failures (first build) are due to

* https://github.com/xamarin/maccore/issues/2154 (known networking issue)
* https://github.com/xamarin/xamarin-macios/issues/10382 (new but not related to this PR)

Additional failures (2nd build) are due to
```
error : No valid macOS code signing keys found in keychain. You need to request a codesigning certificate from https://developer.apple.com. 
```

emaf (approved):

PR #10399, opened 2021-01-12
[CI][VSTS] Simplify notarization to get it back.

mandel-macaque:
This change perse does not do other thing than removing the use of
virtuanenv and remove the installation of python-magic.

We have taken this approach because the images that will be used to provision the CI
bots will have python-magic installed by default. We want to do this
because:

1. CI should not be modifying the machine, we do not install any pkg.
2. python-magic should be present, if not, we will get an import error
   and we will know how to fix it (re-image).


dalexsoto (approved):

dalexsoto:
> CI should not be modifying the machine, we do not install any pkg.

What about when we need to bump Xcode or ObjectiveSharpie? what will the process be then?

mandel-macaque:
@dalexsoto my sentence is a philosophical stand, like many of the ones I have and do not follow

PR #10408, opened 2021-01-13
[msbuild] Fix code signing logic for Mac Catalyst.

rolfbjarne:
Code signing for Mac Catalyst is very much like code signing for macOS, except that:

* If a provisioning profile is required, then we must find a code signing certificate.
* Provide a code signing key even if we don't need a code signing certificate.

This makes the tests in monotouch-test that require a correctly signed app pass.


spouliot (approved):

dalexsoto (approved):

mandel-macaque (approved):

rolfbjarne:
Test failures are known issues: https://github.com/xamarin/maccore/issues/2154

tj-devel709 (approved):

emaf (approved):

PR #10409, opened 2021-01-13
[macos] Add correct support for producing/archiving `dSYM`

spouliot:
TL&DR: This PR

1. Removes the creation of the `.dSYM` based on `Debug Information` [1]

2. Adds dSYM support to XM msbuild (now shared with XI implementation)

3. Archive the `.dSYM` directories (plural) properly, e.g.

```
msbuild -p:Configuration=Release -p:ArchiveOnBuild=true
```

Why ? The long story...

Historically `.dSYM` for Xamarin.Mac have not been very useful, largely
because (most of) the code is JITed so not much is known before runtime.
So they were simply not generated during the builds...

However AOT options were added to Xamarin.Mac, making them potentially
more useful. Also symbols from `libmono` and other native libraries /
frameworks can prove useful when diagnosing application crashes.

Unsurprisingly developers looking to get symbols eventually found _a way_
[1] to get a `.dSYM` for their applications - but it was not quite
correct because:

* setting the debug information option meant that `mmp` would be supplied with `-debug`. This disables several optimizations that are, by default, enabled for release builds. IOW generating symbols should have no effect on the executing code (but it had);

* it was produced when compiling the native launcher, so the symbols coverage was incomplete. How much depends if mono was statically or dynamically linked. However this would not cover any AOTed code nor bundled libraries or user frameworks.

* the .dSYM was produced inside the `x.app/Contents/MacOS/`, side-by-side with the native executable, which makes it part of the **signed** `.app` and also part of the created (and signed) `.pkg`. This had a large impact on the application's, disk and download, size(s). Manually (re)moving the `.dSYM` means re-signing the app and re-creating (and signing) the `.pkg` is not a good solution.

[1] https://forums.xamarin.com/discussion/139705/how-to-symbolicate-a-xam-mac-crash-log


emaf (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.props
```diff
@@ -144,9 +144,8 @@ Copyright (C) 2020 Microsoft. All rights reserved.
 		<TargetArchitectures Condition="'$(TargetArchitectures)' == '' And '$(_PlatformName)' != 'macOS' And '$(ComputedPlatform)' == 'iPhone'">ARMv7</TargetArchitectures>
 
 		<!-- Debug -->
-		<!-- Xamarin.Mac: use MmpDebug, but if that isn't set, use DebugSymbols -->
+		<!-- Xamarin.Mac: use MmpDebug -->
 		<_BundlerDebug Condition="'$(_BundlerDebug)' == '' And '$(_PlatformName)' == 'macOS'">$(MmpDebug)</_BundlerDebug>
-		<_BundlerDebug Condition="'$(_BundlerDebug)' == '' And '$(_PlatformName)' == 'macOS'">$(DebugSymbols)</_BundlerDebug>
```
rolfbjarne, line 149: Is there a way to set MmpDebug from the project properties in the IDE?

tools/mmp/aot.cs
```diff
@@ -235,9 +235,10 @@ public void Compile (IFileEnumerator files)
 
 			if (IsRelease) {
 				// mono --aot creates .dll.dylib.dSYM directories for each assembly AOTed
-				// There isn't an easy was to disable this behavior, so clean up under release
+				// There isn't an easy was to disable this behavior
+				// We move them (cheap) so they can be archived for release builds
 				Parallel.ForEach (filesToAOT, ParallelOptions, file => {
-					if (RunCommand (DeleteDebugSymbolCommand, new [] { "-r", file + ".dylib.dSYM/" }, monoEnv) != 0)
+					if (RunCommand (MoveDebugSymbolCommand, new [] { file + ".dylib.dSYM/", Path.Combine (Path.GetDirectoryName (file), "..", "..", ".."), },  monoEnv) != 0)
```
rolfbjarne, line 241: Why not use `Directory.Move`?

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -1183,6 +1183,203 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
+		</PropertyGroup>
+	</Target>
+
+	<Target Name="_ReadAppExtensionDebugSymbolProperties">
+		<ReadItemsFromFile File="%(_ResolvedAppExtensionReferences.Identity)\..\dsym.items" Condition="Exists('%(_ResolvedAppExtensionReferences.Identity)\..\dsym.items')">
+			<Output TaskParameter="Items" ItemName="_AppExtensionDebugSymbolProperties" />
+		</ReadItemsFromFile>
+	</Target>
+
+	<Target Name="_GenerateAppExtensionDebugSymbols" Condition="'$(ComputedPlatform)' == 'iPhone' And '$(IsAppExtension)' == 'false'"
```
rolfbjarne, line 1207: The condition here will never include Xamarin.Mac - is that on purpose?

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
```diff
@@ -24,10 +28,22 @@ public override bool Execute ()
 				var appDestDir = Path.Combine (productsDir, "Applications", Path.GetFileName (AppBundleDir.ItemSpec));
 				Ditto (AppBundleDir.ItemSpec, appDestDir);
 
-				// Archive the dSYMs...
-				if (Directory.Exists (DSYMDir)) {
-					var destDir = Path.Combine (archiveDir, "dSYMs", Path.GetFileName (DSYMDir));
-					Ditto (DSYMDir, destDir);
+				// Archive the main dSYM...
```
rolfbjarne, line 31: Would it be complicated to merge the iOS and Mac versions of ArchiveTaskBased?

We'll most likely need this logic for Mac Catalyst, but we're using the iOS version of the task assemblies...

msbuild/Xamarin.Shared/Xamarin.Shared.props
spouliot: I though it was but I'm not 100% sure.
It did work like I expected it - but I'm not the author of that line ;-)

tools/mmp/aot.cs
spouliot: It should work. I've re-used the existing logic, which did not use `Directory.Delete`

msbuild/Xamarin.Shared/Xamarin.Shared.targets
spouliot: Yes, app extensions differs between the platforms and this PR (originally for [xc]frameworks support) does not cover it.
I'll file an issue so it's not forgotten.

msbuild/Xamarin.Mac.Tasks.Core/Tasks/ArchiveTaskBase.cs
spouliot: > Would it be complicated...

hmmm... a definite nyes :) 
It would not really be complicated but it would (should) not be a merge like it could be done for most other tasks and best done in a different PR, since the current code is not easy to share.

From memory (likely incomplete) this means
* deal with different paths for `Frameworks`, e.g. extra `Contents` on macOS
* deal with different paths for extensions, e.g. `PlugIns` for macOS. Not sure for Catalyst - is it even possible to have iOS extensions in them ?!?
* AOT differences, e.g. bunch of `.dSYM` for macOS
* Native libraries, e.g. deal with `MonoBundle` on macOS, but not on other platforms
* Bitcode handling - not applicable on macOS

I think the simplest approach would be to have the (shared) base class include a bunch of helper methods (I added `ArchiveDSym` in this PR) and make very simple (and likely not shared) platform specific tasks.

msbuild/Xamarin.Shared/Xamarin.Shared.targets
spouliot: https://github.com/xamarin/xamarin-macios/issues/10412

rolfbjarne (approved):

tools/mmp/aot.cs
spouliot: fixed
it's faster than spawning processes - even when done in parallel

spouliot:
> MTouch tests/NUnit: Failed (Execution failed with exit code 5)
known, unrelated issue -> https://github.com/xamarin/maccore/issues/2154

PR #10414, opened 2021-01-13
[dotnet-linker] Link with GSS when building for iOS/Mac Catalyst.

rolfbjarne:
Add a ComputeNativeBuildFlagsStep, which computes the flags to pass to the
native compiler + native linker. This is currently a very simple
implementation, but it will become more complex as support for missing
features are added.

This also required adding support for writing to the same MSBuild output items
multiple times:

* Split parts of LinkerConfiguration.WriteOutputForMSBuild into a
  FlushOutputForMSBuild method (the part that does the actual writing).
* Make WriteOutputForMSBuild just store the items in a dictionary.
* Add a DoneStep that runs at the very end and that writes out the MSBuild
  output items.

### DIFF
--- a/tools/dotnet-linker/LinkerConfiguration.cs
+++ b/tools/dotnet-linker/LinkerConfiguration.cs
@@ -46,6 +46,8 @@ namespace Xamarin.Linker {
 		
 		string user_optimize_flags;
 
+		Dictionary<string, List<MSBuildItem>> msbuild_items = new Dictionary<string, List<MSBuildItem>> ();
+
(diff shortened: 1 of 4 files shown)


spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne:
Test failure is a known issue: https://github.com/xamarin/maccore/issues/2154

PR #10418, opened 2021-01-14
[dotnet] Don't AOT compile resource assemblies.

rolfbjarne:
Xamarin.iOS don't AOT compile resource assemblies, so we shouldn't do so in
.NET either (and in any case they end up causing native linking failures due
to duplicate native symbols).


spouliot (approved):

rolfbjarne:
Test failures are unrelated: https://github.com/xamarin/maccore/issues/2154

PR #10428, opened 2021-01-14
[msbuild] Fix semi-conflicting options to set codesign timestamp

spouliot:
TL&DR:

This effectively change nothing - but prevents (warn) if both options
contradict themselves.

Long Story....

So we have two ways to control the codesign's `--timetamp` option but
they both ignore each other so we can end up with the option being
set more than once at build time.

`DisableTimestamp` was the original one. It was meant for iOS (and
derivative OS) and disable the option (which requires the network)
for simulator or debug builds. IOW we _wanted_ timestamps when doing
release builds for devices.

```
DisableTimestamp="$(_CodesignDisableTimestamp)"
```

```
<_CodesignDisableTimestamp>False</_CodesignDisableTimestamp>
<_CodesignDisableTimestamp Condition="'$(_SdkIsSimulator)' == 'true' Or '$(_BundlerDebug)' == 'true'">True</_CodesignDisableTimestamp>
```

Now disabling the timestamp did not mean it was enabled. We did not ask
for a timestamp, leaving it to the default which from `man codesign`
means:

> If this option is not given at all, a system-specific default behavior is invoked.
> This may result in some but not all code signatures being timestamped.

Then `UseSecureTimestamp` was added for macOS builds. If hardening is
enabled then a secure timestamp is required.

`msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.targets` `_CodesignAppBundle`

```
UseSecureTimestamp="$(UseHardenedRuntime)"
```

However it's also exposed for iOS (shared target) in
`msbuild/Xamarin.Shared/Xamarin.Shared.targets` `__CodesignNativeLibraries`
but it would always be `false` in that case.

Adding this option means there's now always a `--timestamp` option given,
either to enable it (no URL so it means using Apple's server) or to
disable it (`=none`) but since it's controlled by `UseHardenedRuntime`,
which is macOS only, then iOS device builds are never timestamped.

An alternative would be to have `UseSecureTimestamp` as a macOS-only
option - but that would change how we currently sign the iOS applications
and I'd rather not change things that are known to work.

### DIFF
--- a/msbuild/Xamarin.Localization.MSBuild/MSBStrings.Designer.cs
+++ b/msbuild/Xamarin.Localization.MSBuild/MSBStrings.Designer.cs
@@ -952,7 +952,13 @@ namespace Xamarin.Localization.MSBuild {
                 return ResourceManager.GetString("E0175", resourceCulture);
             }
         }
-        
+
+        public static string W0176 {
+            get {
+                return ResourceManager.GetString("W0176", resourceCulture);
+            }
+        }
+		
         public static string E7001 {
             get {
                 return ResourceManager.GetString("E7001", resourceCulture);
--- a/msbuild/Xamarin.Localization.MSBuild/MSBStrings.resx
+++ b/msbuild/Xamarin.Localization.MSBuild/MSBStrings.resx
@@ -834,6 +834,11 @@
         </value>
     </data>
 
+    <data name="W0176" xml:space="preserve">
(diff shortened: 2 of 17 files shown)


rolfbjarne (approved):

dalexsoto (approved):

spouliot:
> MTouch tests/NUnit: Failed (Execution failed with exit code 5)

known, unrelated issue -> xamarin/maccore#2154

emaf (approved):

PR #10432, opened 2021-01-14
[CI][VSTS] The pkg of the tests fail because there is a path not longer found.

mandel-macaque:
fixes: https://github.com/xamarin/maccore/issues/2363


mandel-macaque:
@rolfbjarne I'm not sure about this fix. I used to run the make in the yaml pipeline and worked then it stopped. Did we change a path in the test generation? Maybe with the catalist work?

rolfbjarne (approved):
This regressed when apitest was merged into xammac_tests, so this is the correct fix.

Issue #10146, opened 2020-11-24
Generated API bindings aren't GC safe


rolfbjarne:
Example generated code:

```csharp
protected virtual NSEvent NextEvent (nuint mask, NSDate expiration, string mode, bool deqFlag)
{
	return NextEvent ((NSEventMask) (ulong) mask, expiration, (NSString) mode, deqFlag);
}
```

Ref: https://discord.com/channels/732297728826277939/732297808148824115/780770773144043521

The problem is that the `NSString` version of `mode` parameter is collectible by the GC as soon as its pointer is passed to native code, which means it can be collected while native code is still using it.

The general solution is that every time we get the pointer of a NSObject instance, we make sure that the GC doesn't collect it until after we're done with the pointer.

spouliot:
`NSString.CreateNative` (and `FreeNative`) should be used, it's cheaper since there won't be a managed allocation - which also means the GC can't do anything.

spouliot:
Looks like it's not _really_ generated code.
```
 [Wrap ("NextEvent ((NSEventMask) (ulong) mask, expiration, (NSString) mode, deqFlag)", IsVirtual = true), Protected]
```

rolfbjarne:
@spouliot yes, for this particular case both are true, but it's a more general problem in our bindings.

spouliot:
@rolfbjarne found other generated cases ?

IIRC we dispose of all locally created `NSObject` instances which should make then unavailable to collection during the native call. The ones on the stack should also not be a problem.

rolfbjarne:
It doesn't have to be us who creates the object in question, this also has the problem:

```csharp
[Export ("convertPoint:fromWindow:")]
[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
public virtual CGPoint ConvertPointFromWindow (CGPoint point, UIWindow? window)
{
	global::UIKit.UIApplication.EnsureUIThread ();
	var window__handle__ = window.GetHandle ();
	CGPoint ret;
	if (IsDirectBinding) {
		if (Runtime.Arch == Arch.DEVICE) {
			if (IntPtr.Size == 8) {
				ret = global::ObjCRuntime.Messaging.CGPoint_objc_msgSend_CGPoint_IntPtr (this.Handle, Selector.GetHandle ("convertPoint:fromWindow:"), point, window__handle__);
			} else {
[...]

PR #10439, opened 2021-01-15
[monotouch-test] Make it work with Mac Catalyst.

rolfbjarne:
* Fix system version checks to work properly on Mac Catalyst (which uses the macOS
  version as its system version).

* Add the framework-specific defines to the build for monotouch-test.csproj (using
  the generated response files), this way we can use them in the tests.

* Sprinkle conditionals in numerous places - I tried using either framework-specific
  or XAMCORE_3_0 whenever that made since (instead of using Mac Catalyst as a condition).

* Updated a few tests to use non-deprecated API (because the deprecated API often
  doesn't exist in Mac Catalyst).

* Also a few minor API fixes to make any corresponding tests compile.

monotouch-test isn't quite green yet - the remaining tests will be fixed in
later pull requests.


spouliot:
So the version returned is the macOS one, like `11.0`
but all API are decorated with versions like `[MacCatalyst (14,0)]` ?
😞

spouliot (approved):

dalexsoto (approved):

rolfbjarne:
> So the version returned is the macOS one, like `11.0`
> but all API are decorated with versions like `[MacCatalyst (14,0)]` ?
> 😞

I had a deeper look, and it seems the `@available` block in Objective-C reads the `iOSSupportVersion` from `/System/Library/CoreServices/SystemVersion.plist`: 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>ProductBuildVersion</key>
	<string>19H114</string>
	<key>ProductCopyright</key>
	<string>1983-2020 Apple Inc.</string>
	<key>ProductName</key>
	<string>Mac OS X</string>
	<key>ProductUserVisibleVersion</key>
	<string>10.15.7</string>
	<key>ProductVersion</key>
	<string>10.15.7</string>
	<key>iOSSupportVersion</key>
	<string>13.6</string>
</dict>
</plist>
```

So I guess this is something we might want to discuss.

tj-devel709 (approved):

src/ObjCRuntime/Runtime.cs
```diff
@@ -1806,6 +1806,21 @@ internal static IntPtr AllocGCHandle (object value)
...
 		}
+
+#if __MACCATALYST__
+		static string _iOSSupportVersion;
+		internal static string iOSSupportVersion {
+			get {
+				if (_iOSSupportVersion == null) {
+					// This is how Apple does it: https://github.com/llvm/llvm-project/blob/62ec4ac90738a5f2d209ed28c822223e58aaaeb7/lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm#L100-L105
+					using (var dict = NSMutableDictionary.FromFile ("/System/Library/CoreServices/SystemVersion.plist")) {
+						_iOSSupportVersion = dict.ObjectForKey ((NSString) "iOSSupportVersion").ToString ();
```
spouliot, line 1817: that `NSString` could be collected before being used
https://github.com/xamarin/xamarin-macios/issues/10146

spouliot (approved):

mandel-macaque (approved):

src/ObjCRuntime/Runtime.cs
rolfbjarne: Fixed.

tests/monotouch-test/AVFoundation/CaptureDeviceTest.cs
```diff
@@ -34,7 +34,12 @@ public void CompareConstantEnum ()
 				Compare (AVMediaType.MetadataObject, AVMediaTypes.MetadataObject);
 
 			// obsoleted in iOS 6, removed in iOS12
-			if (TestRuntime.CheckSystemVersion (PlatformName.iOS, 12,0))
+			var removed = TestRuntime.CheckSystemVersion (PlatformName.iOS, 12, 0);
+#if __MACCATALYST__
+			removed = true;
```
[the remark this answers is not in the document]
rolfbjarne: Fixed.

spouliot (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2378)

PR #10441, opened 2021-01-15
[mtouch][mmp] Exclude extraneous framework files from being copied into apps

spouliot:
Today both `mtouch` and `mmp` are copying the entire `.framework`
directories inside the `.app\[Contents\]Frameworks\` directory.

However not everything in a framework is required at runtime. The most
common unrequired files would be headers (`Headers/*.h`) and modules
(`Modules/*`).

Looking at Xcode build output we can see something like:

```
builtin-copy -exclude .DS_Store -exclude CVS -exclude .svn -exclude .git -exclude .hg -exclude Headers -exclude PrivateHeaders -exclude Modules -exclude \*.tbd -bitcode-strip replace-with-marker -bitcode-strip-tool
```

which excludes a few more, less common, files.

This _builtin_ command is not available externally (for us to re-use)
but it hints that Xcode is likely using `rsync` to avoid copying part of
the files.

Note: the builtin command also _likely_ calls `bitcode_strip` too (or has
similar code embedded) and `mtouch` already does so too

There's a cost to spawning an external process, like `rsync`, which we
avoid by having our own file copier, which clones files (almost zero
cost). That does not support excluding files, but deleting files is also
very cheap. Testing shows copying a framework to be less than 1 ms, even
with with extra deletion step.

### DIFF
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -697,19 +697,16 @@ namespace Xamarin.Bundler {
 			FileCopier.UpdateDirectory (source, target);
 		}
 
+		static string[] NonEssentialDirectoriesInsideFrameworks = { "CVS", ".svn", ".git", ".hg", "Headers", "PrivateHeaders", "Modules" };
+
 		// Duplicate xcode's `builtin-copy` exclusions
 		public static void ExcludeNonEssentialFrameworkFiles (string framework)
 		{
 			// builtin-copy -exclude .DS_Store -exclude CVS -exclude .svn -exclude .git -exclude .hg -exclude Headers -exclude PrivateHeaders -exclude Modules -exclude \*.tbd
 			File.Delete (Path.Combine (framework, ".DS_Store"));
-			DeleteDir (Path.Combine (framework, "CVS"));
-			DeleteDir (Path.Combine (framework, ".svn"));
-			DeleteDir (Path.Combine (framework, ".git"));
-			DeleteDir (Path.Combine (framework, ".hg"));
-			DeleteDir (Path.Combine (framework, "Headers"));
-			DeleteDir (Path.Combine (framework, "PrivateHeaders"));
-			DeleteDir (Path.Combine (framework, "Modules"));
 			File.Delete (Path.Combine (framework, "*.tbd"));
+			foreach (var dir in NonEssentialDirectoriesInsideFrameworks)
+				DeleteDir (Path.Combine (framework, dir));
 		}
 
 		static void DeleteDir (string dir)
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -697,6 +697,33 @@ namespace Xamarin.Bundler {
 			FileCopier.UpdateDirectory (source, target);
 		}
 
+		// Duplicate xcode's `builtin-copy` exclusions
+		public static void ExcludeNonEssentialFrameworkFiles (string framework)
+		{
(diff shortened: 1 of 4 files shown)


tools/common/Application.cs
```diff
@@ -697,6 +697,33 @@ public static void UpdateDirectory (string source, string target)
...
+		{
+			// builtin-copy -exclude .DS_Store -exclude CVS -exclude .svn -exclude .git -exclude .hg -exclude Headers -exclude PrivateHeaders -exclude Modules -exclude \*.tbd
+			File.Delete (Path.Combine (framework, ".DS_Store"));
+			DeleteDir (Path.Combine (framework, "CVS"));
+			DeleteDir (Path.Combine (framework, ".svn"));
+			DeleteDir (Path.Combine (framework, ".git"));
+			DeleteDir (Path.Combine (framework, ".hg"));
+			DeleteDir (Path.Combine (framework, "Headers"));
+			DeleteDir (Path.Combine (framework, "PrivateHeaders"));
+			DeleteDir (Path.Combine (framework, "Modules"));
```
mandel-macaque, line 711: Question, what about having a list with the paths we are not interested and loop over it? 
```suggestion
			foreach (var dir in new string [] {"CVS", ".svn", ".git", ".hg", "Headers", "PrivateHeaders", "Modules"}) {
						DeleteDir (Path.Combine (framework, dir));
			}
```
We can even put that in a var and do a unit test.

mandel-macaque (approved):

dalexsoto (approved):

tools/common/Application.cs
```diff
@@ -697,6 +697,33 @@ public static void UpdateDirectory (string source, string target)
...
+		{
+			// Xcode generates symlinks inside macOS frameworks
+			var realdir = Target.GetRealPath (dir);
+			// unlike File.Delete this would throw if the directory does not exists
+			if (Directory.Exists (realdir)) {
+				Directory.Delete (realdir, true);
+				if (realdir != dir)
+					File.Delete (dir); // because a symlink is a file :)
+			}
+		}
```
mandel-macaque, line 725: Second question, what if we move DeleteDir to tools/common/FileCopier.cs ince we have all the file stuff there?

tools/common/Application.cs
spouliot: a loop makes sense :)
not sure a unit test is needed - unless it become something more general (not private)
the end goal is to have add the test repo to sample tester and check the tree/file structure

tools/common/Application.cs
spouliot: > ince we have all the file stuff 

well I wish it was, but `Target.GetRealPath` hints that's not quite the case...
using `FileCopier.DeleteDir` is just weird, and
it's not meant to be used everywhere (at least not before reviewing all `Directory.Delete` uses).

I might move it to another (new?) file in another PR since I have a few other (file ops) things to fix.

dalexsoto (commented):
@spouliot Nit that just occurred to me, any chance we can also include sh files? IIRC some frameworks also include shellscripts to do arch stripping but we already handle that

rolfbjarne (approved):

spouliot:
@dalexsoto we could but that's a slippery road. While a `.sh` would not be usable on iOS (and derivatives) it could be something a macOS framework wish to embed (not a real good practice IMO but still possible).

AFAICT Xcode only removes it's own files and the one from different source control.

PR #10449, opened 2021-01-18
[msbuild] Remove duplicate trailing slash from _XamarinBclPath. Fixes #10446.

### LINKED ISSUE #10446: Xamarin.Mac 7.3.0.27 Preview breaks our hack to remove facade assemblies
We ship our own System.Drawing[.Common].dll implementation in our application. Hence we have to avoid the Mono/Xamarin facade creeping into the build process. We used the following code to accomplish that:

```
  <Target Name="FixImplicitlyExpandDesignTimeFacades" AfterTargets="ImplicitlyExpandDesignTimeFacades" Condition="'$(TargetFramework)' == 'xamarinmac20'">
    <ItemGroup>
      <_SystemDrawingCommonDesignTimeFacadeAssemblies Include="%(DesignTimeFacadeDirectories.Identity)System.Drawing.Common.dll" />
      <ReferencePath Remove="@(_SystemDrawingCommonDesignTimeFacadeAssemblies)" />
    </ItemGroup>
  </Target>
```

That no longer works due to double-slash introduced into the `DesignTimeFacadeDirectories` paths at  https://github.com/xamarin/xamarin-macios/blob/2b6eb8438a2d812085d780f68b17b0ae89bd9d18/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.TargetFrameworkFix.targets#L36. `_XamarinBclPath` variable already has slash at the end so there ends up being double slash in the path list. This doesn't affect MSBuild `Include` wildcards since they match by path but it affect the `Remove` we use since it matches by string equality.

We currently workaround it by adding the following line, but removing the double-slash would be preferred option:
```
      <_SystemDrawingCommonDesignTimeFacadeAssemblies Include="$(_XamarinBclPath)Facades/System.Drawing.Common.dll" />
```

rolfbjarne:
_XamarinBclPath always has a trailing slash, so no need to add another one
when adding subdirectories to it.

Fixes https://github.com/xamarin/xamarin-macios/issues/10446.

### DIFF
--- a/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.TargetFrameworkFix.targets
+++ b/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.TargetFrameworkFix.targets
@@ -33,7 +33,7 @@ Copyright (c) 2018 Microsoft Corp. (www.microsoft.com)
 	<Target Name="FixDesignTimeFacades" AfterTargets="GetReferenceAssemblyPaths" Condition="('$(OS)' != 'Windows_NT')">
 		<ItemGroup>
 			<DesignTimeFacadeDirectories Remove="@(DesignTimeFacadeDirectories)" />
-			<DesignTimeFacadeDirectories Include="$(_XamarinBclPath)/Facades/" />
+			<DesignTimeFacadeDirectories Include="$(_XamarinBclPath)Facades/" />
 		</ItemGroup>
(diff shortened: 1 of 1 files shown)


spouliot (approved):

rolfbjarne:
Test failures were fixed in #10451.

emaf (approved):

PR #10450, opened 2021-01-18
[dotnet] Bump ILLink to get an updated linker reference assembly.

rolfbjarne:
This allows us to undo a workaround we made for a missing API in the linker
reference assembly (an AnnotationStore.AddPreservedMethod overload).

This also requires a change to use MessageContainer.CreateCustomErrorMessage
instead of MessageContainer.CreateErrorMessage, because apparently having a
reference assembly doesn't mean there can't be incompatible changes in it 😒.


spouliot (approved):

rolfbjarne:
Test failures were fixed in #10451.

PR #10459, opened 2021-01-18
[msbuild] Fix codesign `StampPath`. Fix #10445

### LINKED ISSUE #10445: Xamarin.Mac 7.3.0.27 Preview breaks signing native libraries
Not sure if it's something that is already fixed but the latest Xamarin.Mac previews (non-NET6, version 7.3.0.27) broke something with signing native libraries (possibly related to the CodesignNativeLibraries unification). Now the first pass correctly recognizes the types of the files (ie. Mach-O, etc.) but in the resulting bundle they end up being truncated to zero bytes. Every successive rebuild fails. Even the first build produces unusable bundle with zero-sized .dylib files though.

On the first build the files are still non-empty at this step:
Screenshot 2021-01-16 at 12 22 06

spouliot:
Using `Path.Combine` with a full qualified path for the 2nd argument will
return that 2nd argument (ignoring the first one).

That meant the `StampPath` was pointing to the actual files that were
just signed - overwriting them with a 0-length data (empty).

The solution is to make the 2nd argument relative, starting after
`.app/` so `Path.Combine` works as expected (being relative) while
allowing signing multiple files that have the same name (in different
directories).

ref: https://github.com/xamarin/xamarin-macios/issues/10445

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
@@ -77,7 +77,9 @@ namespace Xamarin.MacDev.Tasks
 
 		string GetOutputPath (ITaskItem item)
 		{
-			return Path.Combine (StampPath, item.ItemSpec);
+			var path = item.ItemSpec;
+			var app = path.LastIndexOf (".app/");
+			return Path.Combine (StampPath, path.Substring (app + ".app/".Length));
 		}
 
 		bool NeedsCodesign (ITaskItem item)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
```diff
@@ -77,7 +77,9 @@ string GetFullPathToTool ()
 
 		string GetOutputPath (ITaskItem item)
 		{
-			return Path.Combine (StampPath, item.ItemSpec);
+			var path = item.ItemSpec;
+			var app = path.LastIndexOf (".app/");
```
rolfbjarne, line 81: I'm not sure if it can actually happen, but this won't work with app extensions.

The original code used `AppBundleDir` to compute the relative path:

https://github.com/xamarin/xamarin-macios/blob/a5d06f0d1e9eeea7c34738677d5d78eca1582916/msbuild/Xamarin.iOS.Tasks.Core/Tasks/CodesignNativeLibrariesTaskBase.cs#L147

rolfbjarne (changes requested):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
spouliot: My thoughts were that extensions, `.appex`, would be signed inside the `.app` of the container - giving them a unique , longer stamp name.

Otherwise you run into the risk of having the same native name (and path / stamp) inside the `.app` and `.appex` and since they could be different (like arch for watch) then the stamp would not a correct signature witness.

spouliot: > but this won't work with app extensions.

I double checked and what happens, right now, with extensions is:

```
    StampPath obj/Debug/codesign/
    outputPath obj/Debug/codesign/Debug/FinderSyncExtension.appex/Contents/MonoBundle/libmono-native.dylib
```

so the path is unique for each extension, so this should work. I'll compare with your suggestion...

spouliot: 😞 
```
Tasks/CodesignTaskBase.cs(83,41): error CS0103: The name 'AppBundleDir' does not exist in the current context [/Users/poupou/git/xamarin/xamarin-macios/msbuild/Xamarin.MacDev.Tasks.Core/Xamarin.MacDev.Tasks.Core.csproj]
```

spouliot: if I use
```csharp
string GetOutputPath (ITaskItem item)
{
	return Path.Combine (StampPath, PathUtils.AbsoluteToRelative (AppBundleDir, item.ItemSpec));
}
```

I get 
```
    /Users/poupou/Downloads/Xamarin_Mac___ExtensionSamples/FinderSyncExtension/bin/Debug/FinderSyncExtension.appex/Contents/MonoBundle/libmono-native.dylib: signed Mach-O thin (x86_64) [libmono-native-555549441d6f6137edfa3546a3fd036157175dae]
    
    StampPath obj/Debug/codesign/
    outputPath obj/Debug/codesign/Contents/MonoBundle/libmono-native.dylib
```

which is incorrect since it will overwrite the original (main app) and other (many extensions in this project) signature witness files

rolfbjarne (approved):

dalexsoto (approved):

emaf (approved):

PR #10475, opened 2021-01-20
[msbuild] Unify how CFBundleName is calculated.

rolfbjarne:
Previously, CFBundleName would default to:

* iOS: CFBundleDisplayName (if set), otherwise the app bundle name.
* all other platforms: the app bundle name.

Now unify the logic so that we have the same behavior on all platforms.

This is a breaking change under the following conditions:

* Building for iOS
* CFBundleName is not set in the Info.plist
* CFBundleDisplayName is set in the Info.plist
* CFBundleDisplayName from the Info.plist is different from AssemblyName in
  the csproj (which is the value used to calculate the app bundle name).

The fix would be to:

* Set CFBundleName in the Info.plist to the desired value.

This change works in previous versions of Xamarin.iOS as well.


dalexsoto (approved):
I wonder if we should document this on the docs site, but definitely a remark on the release notes

spouliot (approved):

emaf (approved):

stephen-hawley (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2365)

PR #10492, opened 2021-01-21
[monotouch-test] Rework big parts of KeyChainTest. Fixes #xamarin/maccore@2365.

rolfbjarne:
'GenericPassword' keychain items are unique by their Service+Account
properties [1]. This means that changing the Label property will not create a
different 'GenericPassword', which has a few consequences:

* It's possible to filter (and try to delete) using the Label property.
* It's possible to try to delete a 'GenericPassword' item, and have that
  deletion attempt fail with 'no item found', and then subsequently trying to
  add the same item will fail with a DuplicateItem, because the deletion
  was filtered using the Label property.

The change I've made is to:

* Make the Label property much more descriptive, and unique per process. This
  makes it easier to figure out where things come from in the Keychain Access
  app.
* Make the Service property unique per process. This way these tests are
  parallel safe and they won't stomp on eachother.
* Keep the Account property the same (a constant value), so that it's easy to
  filter to just these items in the Keychain Access app.
* Remove the Label property from all queries, it doesn't matter anyway. The
  Label property is still set when adding items to the keychain.

Finally try to clean up after ourselves as good as possible. This way we don't
fill the keychain with test stuff. This involves removing certificates and
passwords we add to the keychain at the end of tests.

Fixes https://github.com/xamarin/maccore/issues/2365.

[1]: https://stackoverflow.com/a/11672200/183422


spouliot (approved):

dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

 * link sdk/iOS Unified 64-bits - simulator/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2379
 * link sdk/iOS Unified 64-bits - simulator/Release [dotnet]: https://github.com/xamarin/maccore/issues/2379
 * monotouch-test/iOS Unified 64-bits - simulator/Release (all optimizations) [dotnet]: https://github.com/xamarin/maccore/issues/2380
 * monotouch-test/iOS Unified 64-bits - simulator/Debug (all optimizations) [dotnet]: https://github.com/xamarin/maccore/issues/2380
 * monotouch-test/watchOS 32-bits - simulator/Release (all optimizations): https://github.com/xamarin/maccore/issues/1605
 * introspection/watchOS 32-bits - simulator/Debug (watchOS 3.2): https://github.com/xamarin/maccore/issues/2378

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2378)

PR #10535, opened 2021-01-27
[CI][VSTS] Use the creds store in git for https.

mandel-macaque:
VSTS is using https and not ssh to clone the repos, this means than if
we have any bash that uses git, it will fail. Cred store fixes this by
removing the need to pass the token in all bash scripts.

We remove allways the cred store once we are done.

Sample error that is fixed by this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4412813&view=logs&j=600d49c6-e96b-5e8b-2c3a-9f351e2e5848&t=3ef36ff7-7add-52a4-c578-528e39bd10ed&l=206


dalexsoto (approved):

spouliot (approved):

rolfbjarne (approved):

mandel-macaque:
Unrelated failures and mac tests are passing!

Issue #10236, opened 2020-12-08
introspection/watchOS 32-bits - simulator/Debug (watchOS 3.2): Crashed


spouliot:
watchOS was re-enabled in `main` and `xcode12.3`

However xharness needs to provide a fat or i386 binary when running introspection for watchOS 3.2 (oldest simulator supported)

Only Xcode 12.0 supported either i386 or x86_64.
Before 12.0 (watchOS 7.0) only i386 was supported.
After 12.0 (watchOS 7.1) only x86_64 is supported.

rolfbjarne:
This was fixed here: https://github.com/xamarin/xamarin-macios/pull/10648

PR #10583, opened 2021-02-05
[dotnet] Don't use install-location in pkgs.

rolfbjarne:
Don't use install-location to specify where files should be installed, because
our notarization process replaces this value in the pkg manifest with the root
directory (without updating anything else), with the result that notarized
packages tries to install files in the wrong location.

Instead specify the full path for each individual file, so that the
install-location can be the root directory (and thus the notarization process
won't corrupt the package).

### DIFF
--- a/dotnet/.gitignore
+++ b/dotnet/.gitignore
@@ -1,3 +1,4 @@
+tmpdir
 Microsoft.NET.Workload.*/LICENSE
 WorkloadManifest.json
 nupkgs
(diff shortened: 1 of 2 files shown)


spouliot (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated:

* introspection/watchOS 32-bits - simulator/Debug (watchOS 3.2):: https://github.com/xamarin/xamarin-macios/issues/10236

Issue #10569, opened 2021-02-03
[.NET 6] Add support for CoreCLR


rolfbjarne:
This issue will list what we need to do to support CoreCLR as the runtime.

---

Time estimate: 6 weeks

srxqds:
this means the MonoVM will drop at least?

dalexsoto:
No, MonoVM will still be an option along with CoreCLR for Xamarin.Mac apps. For the rest iOS and company MonoVM will still be used.

rolfbjarne:
As of #11757 CoreCLR is feature-complete with the MonoVM runtime, so this can be closed.

trympet:
How does one opt in to using coreclr (in reference to dotnet/runtime#57877)?

rolfbjarne:
@trympet 

Set `UseMonoRuntime=false` in your project file: https://github.com/xamarin/xamarin-macios/wiki/.NET-6#choosing-the-runtime

PR #10597, opened 2021-02-09
Use pointer-sized GCHandles everywhere.

rolfbjarne:
This involves:

* Stop using some bits of the GCHandle to store extra flags in our native
  code, and instead add an extra field to store those flags.
* Define a INVALID_GCHANDLE constant and use it instead of 0/NULL. This is not
  strictly required, but it makes the code more self-documenting.
* Define a GCHandle type (typedef'ed to void*) and change all variables and
  parameters to use it instead of guint32.
* Use our own xamarin_gchandle_* methods (with pointer-sized types) that wraps
  the mono_gchandle_* embedding API (which uses 32-bit types) everywhere.
* Update managed code (registrars, runtime code, etc) accordingly.

This is required for CoreCLR support, and contributing to #10569.


dalexsoto (approved):
Looks like you really `handled` everything!

src/ObjCRuntime/Method.cs
```diff
@@ -112,6 +112,14 @@ internal static class Method {
 		internal unsafe static IntPtr SetGCHandleTrampoline {
 			get { return Runtime.options->Trampolines->set_gchandle_tramp; }
 		}
+
+		internal unsafe static IntPtr GetFlagsTrampoline {
+			get { return Runtime.options->Trampolines->get_flags_tramp; }
+		}
+
+		internal unsafe static IntPtr SetFlagsTrampoline {
+			get { return Runtime.options->Trampolines->get_flags_tramp; }
```
spouliot, line 121: should not it be `set_flags_tramp` ?

spouliot (changes requested):

runtime/trampolines.m
```diff
@@ -721,57 +723,119 @@
...
-	uint32_t gc_handle = 0;
+	GCHandle gc_handle = INVALID_GCHANDLE;
 	pthread_mutex_lock (&gchandle_hash_lock);
-	if (gchandle_hash != NULL)
-		gc_handle = GPOINTER_TO_UINT (CFDictionaryGetValue (gchandle_hash, self));
+	if (gchandle_hash != NULL) {
+		struct gchandle_dictionary_entry *entry;
+		entry = (struct gchandle_dictionary_entry *) CFDictionaryGetValue (gchandle_hash, self);
+		if (entry != NULL)
+			gc_handle = entry->gc_handle;
```
mandel-macaque, line 793: Questions, and I know that you are going to tell me we don't care, but I want to be sure, does the value gchandle_hash can contain NULL, if that is the case, should we use [CFDictionaryGetValueIfPresent](https://developer.apple.com/documentation/corefoundation/1516739-cfdictionarygetvalueifpresent?language=objc) instead?

runtime/trampolines.m
```diff
@@ -721,57 +723,119 @@
...
+	// COOP: Called by ObjC (when the getFlags selector is called on an object).
+	// COOP: Safe mode upon entry, and doesn't access managed memory, so no need to switch.
+	MONO_ASSERT_GC_SAFE;
+
+	/* This is for types registered using the dynamic registrar */
+	enum XamarinGCHandleFlags flags = XamarinGCHandleFlags_None;
+	pthread_mutex_lock (&gchandle_hash_lock);
+	if (gchandle_hash != NULL) {
+		struct gchandle_dictionary_entry *entry;
+		entry = (struct gchandle_dictionary_entry *) CFDictionaryGetValue (gchandle_hash, self);
```
mandel-macaque, line 811: same question.

mandel-macaque (approved):
Small question about the dict value, but I expect that the answer is we do not care.

runtime/trampolines.m
spouliot:
```
	if (gchandle_hash != NULL) {
```
a few lines up

tj-devel709 (approved):

runtime/trampolines.m
rolfbjarne: The dictionary shouldn't contain NULL values, but even if it does, this is the right way to handle them, so yeah, we don't care 😄

src/ObjCRuntime/Method.cs
rolfbjarne: fixed

spouliot (approved):

rolfbjarne:
Test failures are unrelated

* [xUnit] Mono Mac OS X BCL tests group 4/Mac Modern/Debug: https://github.com/xamarin/maccore/issues/2370
* introspection/watchOS 32-bits - simulator/Debug (watchOS 3.2): https://github.com/xamarin/xamarin-macios/issues/10236

PR #10660, opened 2021-02-18
[dotnet-linker] Use dlsym for System.Net.Security and System.Net.Quic.

rolfbjarne:


tools/common/Application.cs
```diff
@@ -1600,6 +1600,17 @@ public bool UseDlsym (string assembly)
 			if (UseInterpreter)
 				return true;
 
+#if NET
+			asm = Path.GetFileNameWithoutExtension (assembly);
+			switch (asm) {
+			case "System.Net.Security":
+			case "System.Net.Quic":
+				// Some .NET assemblies have P/Invokes to native functions they don't ship. We need to use dlsym for this assemblies.
+				// https://github.com/dotnet/runtime/issues/47533
```
spouliot, line 1609: should this override `DlsymOptions.None` ?
right now this would fail _silently_ (more like an hard to diagnose/ignore error) right ?

spouliot (approved):

tj-devel709 (approved):

mandel-macaque (approved):
Perfect CI run!

tools/common/Application.cs
rolfbjarne: `DlsymOptions.None` is handled earlier, so this will not override `DlsymOptions.None`.

Without this fix the native linker will fail (so it's not silent at all, but not really obvious unless you really know what's going on).

PR #10667, opened 2021-02-18
[dotnet] Force the GC mode to be preemptive.

rolfbjarne:
The default was changed to be hybrid suspend, and due to bugs in Mono that
crashes:

https://github.com/dotnet/runtime/issues/47121

In the meantime we can go back to the older GC mode until Mono has fixed any
issues with the hybrid suspend mode.


spouliot (approved):

dalexsoto (approved):
Looks like this will be short lived! https://github.com/dotnet/runtime/pull/48465

mandel-macaque (approved):

lambdageek:
Hey all, please keep pinging me if you see more of these kinds of glaring issues - we're testing hybrid suspend in dotnet/runtime on arm64, but XI uncovers cases we don't see in the runtime sometimes.

It would be great if we could leave hybrid suspend turned for a couple of previews (maybe not the ones the are important milestones for you), but of course if it doesn't meet the quality bar we can switch the default in dotnet/runtime back to preemptive for XI as we get closer to release.

PR #10668, opened 2021-02-18
[generator] Reduce the number of duplicate availability attributes in generated code

spouliot:
This does not have a huge impact (e.g. size) by itself. However it helps
to find duplicates in non-generated (manual) bindings and that will (in
future PR) remove them, which means less (manual) work to update the
attributes for net6.

Before (iOS)
> [FAIL] Duplicates: 555 API with members duplicating type-level attributes

After (iOS)
> [FAIL] Duplicates : 64 API with members duplicating type-level attributes

Notes

* The above test will be added in a PR, one that has 0 failure :)

* Ideally the duplicated entries would be removed from binding files (as
  they are not needed, just noise) but that more work and a common source
  of merge conflicts

* Some dupes exists due to protocol members inlining (and we might want
  to fix this later)

* Bots are not produce the generator diff right now. Here's a manually
  generated link to that diff
  https://gist.githubusercontent.com/spouliot/4b3f277e0c6d58ccbd09293bbb16bebf/raw/26915a5ad160a1bfe917436917700afeae489bf4/gistfile1.txt

### DIFF
--- a/src/generator-filters.cs
+++ b/src/generator-filters.cs
@@ -146,14 +146,14 @@ public partial class Generator {
 
 			print ("");
 			print ($"// {pname} protocol members ");
-			GenerateProperties (i, fromProtocol: true);
+			GenerateProperties (i, type);
 
 			// also include base interfaces/protocols
 			GenerateProtocolProperties (i, processed);
 		}
 	}
 
-	void GenerateProperties (Type type, bool fromProtocol = false)
+	void GenerateProperties (Type type, Type originalType = null)
 	{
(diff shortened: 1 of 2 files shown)


rolfbjarne (approved):

dalexsoto (approved):

mandel-macaque (approved):

spouliot:
That has an unexpected effect on core filter generation that I need to check... -> **do-not-merge**

tj-devel709 (approved):

PR #10580, opened 2021-02-04
[generator][dotnet] Add support for `[Uns|S]upportedOSPlatformAttribute`

spouliot:
This moves our current/legacy attributes to the ones added in dotnet 5 [1].

Short Forms (only in bindings)

| Old                                   | New                                 |
|---------------------------------------|-------------------------------------|
| [iOS (7,0)]                           | [SupportedOSPlatform ("ios7.0")]    |
| [NoIOS]                               | [UnsupportedOSPlatform ("ios")]     |

Long Forms

| Old                                   | New                                 |
|---------------------------------------|-------------------------------------|
| [Introduced (PlatformName.iOS, 7,0)]  | [SupportedOSPlatform ("ios7.0")]    |
| [Obsoleted (PlatformName.iOS, 12,1)]  | [Obsolete (...)]                    |
| [Deprecated (PlatformName.iOS, 14,3)] | [UnsupportedOSPlatform ("ios14.3")] |
| [Unavailable (PlatformName.iOS)]      | [UnsupportedOSPlatform ("ios")]     |

Other changes

* `[SupportedOSPlatform]` and `[UnsupportedOSPlatform]` are not allowed on `interface` [2] which means they cannot be used for protocols. This is currently handled by inlining the existing attributes on all members.
* `[ObsoletedInOSPlatform]` was removed in net5 RC. This PR is now mapping the existing attributes to `[Obsolete]`, however multiple ones cannot be added so they need to be platform specific.

Tasks

* [x] Generator deduplication of type/members attributes for interfaces (due to [2])
* [ ] ~~Manual bindings~~ -> https://github.com/xamarin/xamarin-macios/issues/11055
* [x] ~~Introspection tests updates~~

References

* [1] https://github.com/xamarin/xamarin-macios/issues/10170
* [2] https://github.com/dotnet/runtime/issues/47599
* [3] https://github.com/dotnet/runtime/issues/47601

PR #10683, opened 2021-02-22
[generator] Do not inline duplicate attributes when processing protocols

spouliot:
- this requires small (but tricky) generator changes
- duplicate availability attributes detection is tested by introspection
- it also requires removing dupes in manual bindings (which was the
  current, if not the _named_, goal for this PR)

Long story:

It's always been hard to detect and enforce removal of extraneous
availability attributes since the generator inlined them _freely_
when processing ObjC protocols.

That meant introspection could not report dupes and led to many
extra attributes. This was **not** a problem in binding files since we
could ignore them. IOW their presence (in the input) does not mean they
are _all_ dupes in generated code (output).

However dupes in manual bindings were also (forced to be) _ignored_ and
those were part of the assemblies we ship. Again it was not a big deal,
nor a source of much extra metadata/size, so it was ignored.

Forward to today, manual bindings needs to be updated for net6 [1] so any
extra we have requires more (manual) work. Cleaning this problem up in
the generator code reduce the (manual) work we need to do.

It also means introspection can detect dupes (in generated and manual
code) so we don't end up adding more (which would also require more
manual work to support both set of attributes).

[1] https://github.com/xamarin/xamarin-macios/pull/10580


spouliot:
Made it a draft PR at the last moment... since I realized it's missing a case to detect dupes on properties (even if the getter and setters are fine).

src/VideoToolbox/VTCompressionSession.cs
```diff
@@ -17,7 +18,13 @@
 using CoreVideo;
 
 namespace VideoToolbox {
+
+#if NET
+	[SupportedOSPlatform ("ios8.0")]
+	[SupportedOSPlatform ("tvos10.2")]
```
rolfbjarne, line 24: I wonder if we should make these strings constants?

rolfbjarne (approved):

src/VideoToolbox/VTCompressionSession.cs
spouliot: hmm... that change was not meant to be included in _this_ PR but yeah I came to the same ~~conclusion~~ question

If we use strings we'll end up with mistakes - even if I added introspection tests against _most_ typos
OTOH using string constants requires a bit more typing
* we can have sharpie generate them...
* but I'd rather only have the old ones inside binding files (and limit the dual set to manual bindings)

Maybe some internal short hands like `_.ios_8_0`...

dalexsoto (approved):
Looking good! Wish we had the generator diff available

spouliot:
@dalexsoto I'm merging `main` and doing the diff locally. I'll gist the diff of the PR before getting this out of draft :)

spouliot:
generator diff (huge)
https://gist.githubusercontent.com/spouliot/a5d7809e9d8f23a0c1ba0a1d020c442c/raw/8935de7fe803eb6c52852c592a4e41445b3c81ee/gistfile1.txt

tj-devel709 (approved):

PR #10701, opened 2021-02-23
[bgen] Remove trailing newlines from warning/error messages.

rolfbjarne:
This fixes an issue where a composed error message would show up in two lines:

    BTOUCH : error BI1079: bgen: Do not know how to make a signature for System.Boolean*
    		 in parameter `inventoryAbortResultCodeFound' from iOSBinding.Easy2ReadProtocolMessage.FromRawData

Leaving customers confused (https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1280923),
because they wouldn't see the second part of the error.

After the fix:

    BTOUCH : error BI1079: bgen: Do not know how to make a signature for System.Boolean* in parameter `inventoryAbortResultCodeFound' from iOSBinding.Easy2ReadProtocolMessage.FromRawData

---

Before:

Screen Shot 2021-02-23 at 09 16 27

After:

Screen Shot 2021-02-23 at 09 20 54

On another note, we should probably do this for all our other warning/error messages as well.

### DIFF
--- a/src/Resources.resx
+++ b/src/Resources.resx
@@ -140,43 +140,35 @@
 	</data>
 
 	<data name="BI0026" xml:space="preserve">
-		<value>Could not parse the command line argument '--warnaserror': {0}
-		</value>
+		<value>Could not parse the command line argument '--warnaserror': {0}</value>
 	</data>
 
 	<data name="BI0068" xml:space="preserve">
-		<value>Invalid value for target framework: {0}.
-		</value>
+		<value>Invalid value for target framework: {0}.</value>
 	</data>
 
 	<data name="BI0070" xml:space="preserve">
-		<value>Invalid target framework: {0}. Valid target frameworks are: {1}.
-		</value>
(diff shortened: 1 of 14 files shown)


spouliot (approved):
Are we sure the translations won't have the extra line ?

If that's the case then the correct fix would be different (either in the IDE, the `ToString()` of the error/warning or both)

dalexsoto (approved):

mandel-macaque (approved):

tj-devel709 (approved):

rolfbjarne:
@spouliot 

> Are we sure the translations won't have the extra line ?
> 
> If that's the case then the correct fix would be different (either in the IDE, the `ToString()` of the error/warning or both)

Good question, and yes, the translations had the extra lines too, so I fixed the translations as well (nothing has actually been translated for these strings yet).

rolfbjarne:
Test failures are unrelated:

* monotouch-test/iOS Unified 64-bits: https://github.com/xamarin/maccore/issues/2304

PR #10723, opened 2021-02-25
[runtime] Build a .NET-specific libxamarin-dotnet[-debug].[a|dylib].

rolfbjarne:
This also means linking with the runtime packs from .NET instead of the mono archive
(thus we have one less reliance on the mono archive).

We're also using the Xamarin.iOS code for our macOS launch sequence now, since
(at least at first) we're only going to support self-contained .NET macOS apps
(so no need to support a system-installed runtime, which simplifies things a
bit).

### DIFF
--- a/tools/common/Target.cs
+++ b/tools/common/Target.cs
@@ -559,8 +559,9 @@ namespace Xamarin.Bundler {
 					case ApplePlatform.MacOSX:
 #if NET
 						GenerateIOSMain (sw, abi);
-#endif
+#else
 						GenerateMacMain (sw);
+#endif
 						break;
 					default:
 						throw ErrorHelper.CreateError (71, Errors.MX0071, platform, App.ProductName);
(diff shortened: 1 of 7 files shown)


tools/common/Target.cs
```diff
@@ -557,6 +557,9 @@ public void GenerateMain (StringBuilder sb, ApplePlatform platform, Abi abi, str
 						GenerateIOSMain (sw, abi);
 						break;
 					case ApplePlatform.MacOSX:
+#if NET
+						GenerateIOSMain (sw, abi);
+#endif
```
spouliot, line 562: no `#else` ?

spouliot (approved):

tools/common/Target.cs
rolfbjarne: 🤦

dalexsoto (approved):

tj-devel709 (approved):

tools/common/Target.cs
filipnavara: I looked at that too and after pondering for few minutes I came to the conclusion that it was intentional and the generated `xammac_setup` method is called somewhere... Well, perhaps a wrong conclusion on my part :-)

PR #10724, opened 2021-02-25
[dotnet] Fix the Run target for Mac Catalyst to run the app directly instead of using mlaunch.

rolfbjarne:


dotnet/Microsoft.MacCatalyst.Sdk/targets/Microsoft.MacCatalyst.Sdk.targets
```diff
@@ -1,4 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="Xamarin.Shared.Sdk.targets" />
+
+  <PropertyGroup>
+    <RunCommand>open</RunCommand>
```
spouliot, line 6: is this not an easy case to map this to `dotnet run` since there's no sim/device to consider ? (not quite sure about the mapping itself)

spouliot (approved):

dotnet/Microsoft.MacCatalyst.Sdk/targets/Microsoft.MacCatalyst.Sdk.targets
rolfbjarne: This is how we make `dotnet run` work (by setting `RunCommand`)

dalexsoto (approved):

tj-devel709 (approved):

PR #10725, opened 2021-02-25
[dotnet-linker] Disable Objective-C exception marshalling by default for macOS and Mac Catalyst.

rolfbjarne:


spouliot (approved):

tools/common/Application.cs
```diff
@@ -1357,7 +1357,14 @@ public void SetObjectiveCExceptionMode ()
 						break;
 					case ApplePlatform.MacOSX:
 					case ApplePlatform.MacCatalyst:
+#if NET
+						// Mono doesn't support dllmaps for Mac Catalyst / macOS in .NET for now:
+						// macOS: https://github.com/dotnet/runtime/issues/43204
+						// Mac Catalyst: https://github.com/dotnet/runtime/issue
```
dalexsoto, line 1363: Did you mean to link an issue here?

dalexsoto (approved):

tools/common/Application.cs
filipnavara, line 1363:
```suggestion
						// Mac Catalyst: https://github.com/dotnet/runtime/issues/48110
```

tj-devel709 (approved):

tools/common/Application.cs
rolfbjarne: Yes, it was an incomplete copy-paste.

PR #10729, opened 2021-02-25
[Speech] Update bindings for Xcode 12.5 Beta 1

rachelkang:


src/speech.cs
```diff
@@ -262,6 +290,7 @@ interface SFTranscriptionSegment : NSCopying, NSSecureCoding {
 
 		[iOS (13, 0)]
 		[NullAllowed, Export ("voiceAnalytics")]
+		[Advice ("Use 'SFVoiceAnalytics' from 'SFSpeechRecognitionMetadata' instead.")]
```
dalexsoto, line 293: Small typo here

```suggestion
		[Advice ("Use 'VoiceAnalytics' from 'SFSpeechRecognitionMetadata' instead.")]
```

dalexsoto (changes requested):

tj-devel709 (approved):
Congrats on first PR! Good for me after Alex's suggestion! :)

spouliot (approved):

spouliot:
removed `note-deprecation` since we do not use it inside release notes for API (there's too much of them in general)

src/speech.cs
rachelkang: oops! thanks for catching

dalexsoto (approved):
👍

PR #10731, opened 2021-02-25
[catalyst] Remove "unknown" API inside existing frameworks

spouliot:
xtro tests are based on Apple's header files and report as _unknown_
bindings to API that are not found in headers - Catalyst in this case.

Removing them is required so the applications can be submitted to the
AppStore.

### DIFF
--- a/src/foundation.cs
+++ b/src/foundation.cs
@@ -11643,11 +11643,13 @@ namespace Foundation
 	[Model, BaseType (typeof (NSObject))]
 	[Protocol]
 	interface NSPortDelegate {
+		[NoMacCatalyst]
 		[Export ("handlePortMessage:")]
 		void MessageReceived (NSPortMessage message);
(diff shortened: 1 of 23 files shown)


dalexsoto (approved):

src/foundation.cs
```diff
@@ -11643,11 +11643,13 @@ interface NSPort : NSCoding, NSCopying {
 	[Model, BaseType (typeof (NSObject))]
 	[Protocol]
 	interface NSPortDelegate {
```
dalexsoto, line 11645: Any reason why `[NoMacCatalyst]` is instead on the parent delegate?

spouliot: it was noted in the `.ignore` file - that I forgot to commit :|

rolfbjarne (approved):

PR #10748, opened 2021-03-01
[CallKit] Update bindings for Xcode12.5 Beta 1

rachelkang:


src/callkit.cs
```diff
@@ -77,6 +78,17 @@ public enum CXErrorCodeCallDirectoryManagerError : long {
...
 
+	[iOS (14,5), NoWatch, NoTV, NoMac]
+	[Introduced (PlatformName.MacCatalyst, 14, 5)]
+	[ErrorDomain ("CXErrorDomainNotificationServiceExtension")]
+	[Native]
+	public enum CXErrorCodeNotificationServiceExtensionError : long
+	{
+		Unknown = 0,
+		InvalidClientProcess = 1,
+		MissingNotificationFilteringEntitlement = 2
```
dalexsoto, line 89: Add a `,` to the end of every last enum member to reduce future diff noise

```suggestion
		MissingNotificationFilteringEntitlement = 2
```

src/callkit.cs
```diff
@@ -32,6 +32,7 @@ public enum CXErrorCode : long {
 		Unknown = 0,
 		Unentitled = 1,
 		InvalidArgument = 2,
+		MissingVoIPBackgroundMode = 3
```
spouliot, line 35: minor: end new enum members with `,`
it's valid C# and it means the next update won't create a larger diff (since your line won't have to be modified)

src/callkit.cs
spouliot: same (end with `,`)

dalexsoto (approved):

spouliot (approved):

rolfbjarne (approved):

mandel-macaque (approved):

tj-devel709 (approved):

PR #10751, opened 2021-03-02
[runtime] Always enable debugging in .NET/macOS and .NET/Mac Catalyst in debug versions of libxamarin.

rolfbjarne:
This makes it so that it's possible to attach to the debugger by setting the
\_\_XAMARIN_DEBUG_HOSTS\_\_/\_\_XAMARIN_DEBUG_PORT\_\_ environment variables, without
the need for enabling a setting somewhere else (like in
Settings.bundle/Root.plist on mobile devices).

### DIFF
--- a/runtime/monotouch-debug.m
+++ b/runtime/monotouch-debug.m
@@ -519,7 +519,7 @@ void monotouch_configure_debugging ()
 	NSDictionary *settings = [NSDictionary dictionaryWithContentsOfFile: root_plist];
 	NSArray *preferences = [settings objectForKey:@"PreferenceSpecifiers"];
 	NSMutableArray *hosts = [NSMutableArray array];
-	bool debug_enabled;
+	bool debug_enabled = true;
 	NSString *monodevelop_host;
 	NSString *monotouch_debug_enabled;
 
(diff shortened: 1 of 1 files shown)


runtime/monotouch-debug.m
```diff
@@ -537,13 +537,15 @@ void monotouch_configure_debugging ()
 		return;
 	}
  
-	// If debugging is enabled
+ #if !(TARGET_OS_MACCATALYST || TARGET_OS_OSX)
+	// If debugging is enabled (only check for mobile builds - for macOS / Mac Catalyst debugging is always enabled in debug versions of libxamarin)
```
spouliot, line 541: title say it's not .net/Mac[catalyst] only
but it seems to apply to current/legacy too
That should be fixed (title) and documented (in the release notes at least)

spouliot (approved):

runtime/monotouch-debug.m
rolfbjarne: This file (monotouch-debug.m) isn't included in legacy Xamarin.Mac, so this will not have any effect on legacy code.

PR #10761, opened 2021-03-02
[Xharness] Bump to the new nuget in the xharness main branch.

mandel-macaque:
@premun please take a look. AFAIK it should not make a diff on this side.


dalexsoto (approved):
await review 🔥

rolfbjarne (approved):

premun:
@mandel-macaque yes, the await was happening on the first line of the method that used to accept the `Task` before. So should be all good. I will have a look at the failures here.

@mandel-macaque are you rushing with this bump? I have a change here that could use a bump too:
https://github.com/dotnet/xharness/pull/491
(also if you could have a look at that one)

mandel-macaque:
I can look at it and bump here too. This is not blocking main

tj-devel709 (approved):

mandel-macaque:
Was successful, but I used the VSTS UI to re-trigger and that resulted in the result not being posted: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4562238&view=logs&s=e6f8fb51-ebc6-57c3-05b7-6decad883f29&j=600d49c6-e96b-5e8b-2c3a-9f351e2e5848

PR #10778, opened 2021-03-03
[CI] Remove login keychain references

cadsit:
Cleans up existing usage of login keychain. Used in conjunction with https://github.com/xamarin/maccore/pull/2404


rolfbjarne (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2409).

rolfbjarne:
@cadsit is there anything else that needs to be done for this PR, or can we merge it? You marked it as **do-not-merge**.

dalexsoto (approved):

tools/devops/automation/build-pipeline.yml
```diff
@@ -98,6 +98,8 @@ stages:
         runTests: ${{ parameters.runTests }}
         runDeviceTests: ${{ parameters.runDeviceTests }}
         keyringPass: $(xma-password)
+        gitHubToken: ${{ variables['GitHub.Token'] }}
```
mandel-macaque, line 101: Why using ${{ }} over $() for a variable?
```suggestion
        gitHubToken: $(GitHub.Token)
```
Is the '.' right?

tools/devops/automation/templates/devices/build.yml
```diff
@@ -55,14 +55,6 @@ steps:
...
   continueOnError: true
 
-- bash: |
-    security set-key-partition-list -S apple-tool:,apple: -s -k $KEYCHAIN_PASS login.keychain
-  env:
-    KEYCHAIN_PASS: ${{ parameters.keyringPass }}
-  displayName: 'Remove security UI-prompt (http://stackoverflow.com/a/40039594/183422)'
-  condition: succeededOrFailed() # we do not care about the previous process cleanup
-  continueOnError: true
-
```
mandel-macaque, line 65: We don't longer need it?

tools/devops/automation/build-pipeline.yml
cadsit: They should be semantically equivalent in this case.

I chose to use ${{ }} here because the value of GitHub.Token is known at compile-time, since ${{ }} gets replaced before the pipeline runs (where as $() gets replaced as the pipeline is running). It's become habit for me to use ${{ }} since working on GitHub actions because a lot of substitutions there can only be made during compile time.

tools/devops/automation/templates/devices/build.yml
cadsit: correct. you will still want to set it on xamarin-macios.keychain, but the login.keychain should no longer be used in CI.

PR #10780, opened 2021-03-04
[mediaplayer] Update for xcode12.5 beta 3

spouliot:


src/MediaPlayer/MPMediaItem.cs
```diff
@@ -282,6 +290,13 @@ int Int32ForProperty (NSString property)
 				return (ValueForProperty (PlaybackStoreIDProperty) as NSString);
 			}
 		}
+
+		[Watch (7,3), TV (14,5), Mac (11,3), iOS (14,5)]
+		public bool IsPreorder {
+			get {
+				return BoolForProperty (IsPreorderProperty);
+			}
+		}
```
dalexsoto, line 299: 😇

```suggestion
		[Watch (7,3), TV (14,5), Mac (11,3), iOS (14,5)]
		public bool IsPreorder => BoolForProperty (IsPreorderProperty);
```

I know I know I'll set myself out 😅

dalexsoto (approved):

src/MediaPlayer/MPMediaItem.cs
spouliot: yeah, that would be fine too :) 

My biggest problem with `=>` is that people forget the width limit of the screen and you have to scroll in two dimensions ;-)

Otherwise I dream in IL and that ends up identical 😸

dalexsoto: width limit of the screen? I do not know what you are talking about 😅 (Says the guy with a res of 5120 x 1440)

rolfbjarne (approved):

src/MediaPlayer/MPMediaItem.cs
mandel-macaque: Alex! I use vim, 80 chars max, that is the way

mandel-macaque (approved):

src/MediaPlayer/MPMediaItem.cs
dalexsoto: @mandel-macaque `public bool IsPreorder => BoolForProperty (IsPreorderProperty);` is only 63 :P plus 2 tabs (if your tab size is 4 spaces) so 71 it perfectly fits your window 😁

PR #10782, opened 2021-03-04
Add a MarshalAs attribute to all boolean return types and parameters in P/Invokes.

rolfbjarne:
There is a mismatch between the size of booleans for managed and native code:
in most cases native bools are 1 byte, but the default marshalling behavior is
to marshal 4 bytes. This difference hasn't come up often with Mono, but with
CoreCLR numerous tests fail due to this (and presumably many more are randomly
succeeding).

So add a MarshalAs attribute to all boolean return types and parameters for
P/Invokes (including delegates passed to P/Invokes), and also add a test to
make sure we'll never forget to add the attribute when needed.

Ref: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1414?view=vs-2019

### DIFF
--- a/src/AddressBook/ABAddressBook.cs
+++ b/src/AddressBook/ABAddressBook.cs
@@ -218,6 +218,7 @@ namespace AddressBook {
 		}
 
 		[DllImport (Constants.AddressBookLibrary)]
+		[return: MarshalAs (UnmanagedType.I1)]
 		extern static bool ABAddressBookHasUnsavedChanges (IntPtr addressBook);
 		public bool HasUnsavedChanges {
 			get {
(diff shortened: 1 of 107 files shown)


filipnavara (approved):

mandel-macaque:
🤯

mandel-macaque (approved):

dalexsoto (approved):
So the day has finally come

spouliot (approved):
@dalexsoto this should be added to sharpie since this will affect 3rd party bindings too

dalexsoto:
@spouliot issue created so we do not forget.

tj-devel709 (approved):

PR #10785, opened 2021-03-04
Turns out we need MarshalAs for characters as well.

rolfbjarne:
The default marshalling size for characters in .NET is a single byte, and that
is potentially lossy (and causes tests to fail for non-ascii characters when using CoreCLR).

### DIFF
--- a/src/generator.cs
+++ b/src/generator.cs
@@ -1127,6 +1127,8 @@ public partial class Generator : IMemberGatherer {
 			return "float";
 		if (t == TypeManager.System_Boolean)
 			return "bool";
+		if (t == TypeManager.System_Char)
+			return "char";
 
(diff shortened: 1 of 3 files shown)


src/generator.cs
```diff
@@ -3454,8 +3460,8 @@ public string FormatTypeUsedIn (string usedInNamespace, Type type, bool protocol
 			return "nint";
 		if (type == TypeManager.System_nuint)
 			return "nuint";
-		if (type == TypeManager.System_Boolean)
```
rolfbjarne, line 3457: It looks like I'm removing a case for boolean, but it's already handled above (the code was duplicated before).

spouliot (approved):

dalexsoto (approved):

PR #10798, opened 2021-03-05
[runtime] Build a new version of libxamarin specifically for CoreCLR.

rolfbjarne:
We're currently only building for macOS, but we'll eventually build for Mac
Catalyst as well.


runtime/Makefile
```diff
@@ -440,6 +440,10 @@ MAC_LIBS =                   \
 #    This is used linked into the native executable when using the system Mono (i.e. not)
 #    embedding the Mono runtime).
 #
+# libxamarin-coreclr.a:
```
filipnavara, line 443: `libxamarin-dotnet-coreclr.a` ?

spouliot (approved):

dalexsoto (approved):

runtime/Makefile
rolfbjarne: Yes, that's better :smile: I'll fix it in a different PR.

Issue #10774, opened 2021-03-03
Binding project using xcframework can't be used in Xamarin.iOS app


leonluc-dev:
### Steps to Reproduce

1.  Create a Binding project with a xcframework as the native framework reference
2.  Build binding project and reference it in a Xamarin.iOS app.
3.  Build the Xamarin.iOS app

### Expected Behavior
The build process includes the native reference (or derived framework in case of xcframework) used in the referenced binding project.

### Actual Behavior
If a xcframework was used as a native reference in the binding project the expected derived framework isn't included in the Xamarin.iOS build process, which causes builds to fail with a MT5211 error. 

I've added an example project showcasing the issue below (using the Universal.xcframework that was supplied for xcframework testing).

### Environment

```
=== Visual Studio Professional 2019 for Mac ===

Version 8.9 (build 1651)
Installation UUID: 57691245-242a-4da8-98b8-14c6159c8d17
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000122

=== Mono Framework MDK ===

Runtime:
	Mono 6.12.0.122 (2020-02/c621c35ffa0) (64-bit)
	Package version: 612000122

=== Roslyn (Language Service) ===

3.9.0-5.21112.8+f3ff04378c972d435826e6181de364b3c0db8d14

=== NuGet ===

Version: 5.8.0.6860

=== .NET Core SDK ===

SDK: /usr/local/share/dotnet/sdk/5.0.103/Sdks
SDK Versions:
	5.0.103
	5.0.102
	5.0.101
	5.0.100
	3.1.406
	3.1.405
	3.1.404
	3.1.403
	3.1.402
	3.1.401
	3.1.302
	3.1.301
MSBuild SDKs: /Applications/Visual Studio.app/Contents/Resources/lib/monodevelop/bin/MSBuild/Current/bin/Sdks

=== .NET Core Runtime ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	5.0.3
	5.0.2
	5.0.1
	5.0.0
	3.1.12
	3.1.11
	3.1.10
	3.1.9
	3.1.8
	3.1.7
	3.1.6
	3.1.5
	2.1.23
	2.1.22
	2.1.21
	2.1.20
	2.1.19

=== .NET Core 3.1 SDK ===

SDK: 3.1.406

=== Xamarin.Profiler ===

[...]

PR #10799, opened 2021-03-05
[msbuild] Fix localization when the error code is shared with mtouch/mmp

spouliot:
Code shared between msbuild and mtouch/mmp needs to share the
localization strings, otherwise the exception being thrown is of no help
to solve the issue.

THis happened if a wrong path was given for an `.xcframework` and
`FileCopier` reported an `MT1022` error.

Reference (not a fix for) https://github.com/xamarin/xamarin-macios/issues/10774

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Xamarin.MacDev.Tasks.Core.csproj
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Xamarin.MacDev.Tasks.Core.csproj
@@ -47,8 +47,13 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <None Include="..\..\tools\mtouch\errors.resx">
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):

rolfbjarne (approved):
This won't pick up the translations though, will it?

I think we'll have to use a single resx eventually, because we keep reusing error numbers/messages everywhere.

spouliot:
> This won't pick up the translations though, will it?

It does pick the original (English) one instead of throwing.

> I think we'll have to use a single resx eventually, because we keep reusing error numbers/messages everywhere.

We're moving to a new translation process, maybe things will be better...

My main concern is a single file is that we use some errors inside the platform assemblies and, for to avoid bloating them needlessly, they should not get all the error strings only used for the tools.

PR #10814, opened 2021-03-09
[runtime] Move some code from native to managed for xamarin_release_managed_ref.

rolfbjarne:
* Avoids one usage of xamarin_set_nsobject_flags (which pokes into managed
  memory from native code, which won't be possible with CoreCLR).

* Makes it possible to move more code from native to managed for
  xamarin_release_managed_ref in the future.

* Since the code order is exactly the same, it shouldn't have any other side
  effects.


filipnavara:
> Since the code order is exactly the same, it shouldn't have any other side effects.

There's a side effect that the managed code is in GC safe mode and the unmanaged code is in GC unsafe mode. The transition happens at different point in this PR which may have a side effect. Now the garbage collector can run between the two statements while it previously could not.

That said, in this particular case I don't see any issue with it.

Issue #10793, opened 2021-03-05
Should we add `[assembly:Introduced (...)]` to the platform assemblies ?


spouliot:
Right now our minimum supported OS are **not** defined inside the assembly metadata (but inside our tools / build scrips).  However we'll [need this for dotnet](https://github.com/xamarin/xamarin-macios/issues/10728) so it make sense to consider doing the same for _legacy_ assemblies.

Beside the slight change to the attribute itself

```diff
diff --git a/src/ObjCRuntime/PlatformAvailability2.cs b/src/ObjCRuntime/PlatformAvailability2.cs
index 04477a67d..c5c5390a9 100644
--- a/src/ObjCRuntime/PlatformAvailability2.cs
+++ b/src/ObjCRuntime/PlatformAvailability2.cs
@@ -51,6 +51,7 @@ namespace ObjCRuntime
        }
 
        [AttributeUsage (
+               AttributeTargets.Assembly |
                AttributeTargets.Class |
                AttributeTargets.Constructor |
                AttributeTargets.Delegate |
```

We would need to mark API that are not available differently. 

E.g. All `ARKit.*` types would need to mention they are not available outside iOS, otherwise the current (legacy) tooling would presume it's available in the minimum version (since the lack of attribute on the type would fallback to the assembly, which would mention something with `[Watch (2,0)]`).

This has indirect benefits since we often inline (for protocols) several `[Introduced]` attributes in members - but if we had an `[Unavailable]` on the type then we could (actually would) skip them.

Issue #10728, opened 2021-02-25
Map TargetPlatformVersion to OSMinimumVersion


rolfbjarne:
See: https://github.com/dotnet/designs/blob/main/accepted/2020/minimum-os-version/minimum-os-version.md#msbuild-properties

spouliot:
hmm... that spec mention things, like attributes, that are not part of (the final) .net5. I suspect it's (partially) outdated... 

@terrajobst anything more recent ?

rolfbjarne:
This is the same as #12336, so closing this one.

Issue #10659, opened 2021-02-18
Use .NET-specific min OS versions to build native code for .NET


rolfbjarne:
It looks like we're going to drop support for a few older OS versions.

In particular the new min OS versions are going to be:

* iOS: 10 (https://github.com/dotnet/runtime/issues/44654)
* tvOS 10 (https://github.com/dotnet/runtime/issues/44654)
* macOS: 10.13 (https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0-supported-os.md#macos)

We have to update our code accordingly to handle this (instead of using the min OS versions defined for legacy Xamarin)

spouliot:
That would be 10.14+ for net6

https://github.com/dotnet/core/blob/master/release-notes/6.0/6.0-supported-os.md

spouliot:
we also need to define/confirm the minimum for Mac Catalyst

it started shipping in 10.15 (so before our minimum Mac) so the (rock bottom) minimum would be 13.0 (since it match the iOS version)

rolfbjarne:
> so the (rock bottom) minimum would be 13.0

Xcode 12.4 doesn't let you select anything lower than 10.15.1/13.2.

IMHO we shouldn't support anything lower than what the current stable Xcode does when our Mac Catalyst support goes to stable.

rolfbjarne:
This will be easier to implement after we've dropped support for legacy Xamarin, which will happen for Xcode 15, so I'm postponing this until .NET 9.

PR #10820, opened 2021-03-09
Add the minimum OS supported inside platform assemblies metadata

spouliot:
The minimum OS supported for .net assemblies are specified in
https://github.com/dotnet/core/blob/master/release-notes/6.0/6.0-supported-os.md

The minimum for legacy assemblies are unchanged. However this information
will now be present in the assemblies (and not just in documentation and
code).

This PR partially address https://github.com/xamarin/xamarin-macios/issues/10659 and possibly https://github.com/xamarin/xamarin-macios/issues/10728

Fix https://github.com/xamarin/xamarin-macios/issues/10793


Make.config
```diff
@@ -189,6 +189,12 @@ MIN_WATCH_OS_VERSION=2.0
 MIN_TVOS_SDK_VERSION=9.0
 MIN_MACCATALYST_SDK_VERSION=13.1
 
+# Minimum OS versions for dotnet6 support
```
dalexsoto, line 192: I wonder if we should have these inside the (Make.versions)[https://github.com/xamarin/xamarin-macios/blob/main/Make.versions] file. Just a nit really and possibly out of the scope of this PR

```shell
# Minimum OS versions for running XI/XM apps.
MIN_IOS_SDK_VERSION=7.0
MIN_OSX_SDK_VERSION=10.9
MIN_WATCHOS_SDK_VERSION=2.0
MIN_WATCHOS64_32_SDK_VERSION=5.1
MIN_WATCH_OS_VERSION=2.0
MIN_TVOS_SDK_VERSION=9.0
MIN_MACCATALYST_SDK_VERSION=13.1

# Minimum OS versions for dotnet6 support
DOTNET_MIN_IOS_SDK_VERSION=10.0
DOTNET_MIN_TVOS_SDK_VERSION=10.0
DOTNET_MIN_MACCATALYST_SDK_VERSION=13.1
DOTNET_MIN_OSX_SDK_VERSION=10.14
```

dalexsoto (approved):

tools/common/Make.common
```diff
@@ -2,8 +2,18 @@
...
 		-e 's/@IOS_SDK_VERSION@/$(IOS_SDK_VERSION)/g' -e 's/@WATCHOS_SDK_VERSION@/$(WATCH_SDK_VERSION)/' -e 's/@TVOS_SDK_VERSION@/$(TVOS_SDK_VERSION)/' -e 's/@OSX_SDK_VERSION@/$(OSX_SDK_VERSION)/' \
 		-e 's/@MACCATALYST_SDK_VERSION@/$(MACCATALYST_SDK_VERSION)/g' \
-		-e 's/@MIN_IOS_SDK_VERSION@/$(MIN_IOS_SDK_VERSION)/g' -e 's/@MIN_WATCHOS_SDK_VERSION@/$(MIN_WATCHOS_SDK_VERSION)/' -e 's/@MIN_TVOS_SDK_VERSION@/$(MIN_TVOS_SDK_VERSION)/' -e 's/@MIN_OSX_SDK_VERSION@/$(MIN_OSX_SDK_VERSION)/' \
+		\
+		-e 's/@MIN_IOS_SDK_VERSION@/$(MIN_IOS_SDK_VERSION)/g' \
+		-e 's/@MIN_WATCHOS_SDK_VERSION@/$(MIN_WATCHOS_SDK_VERSION)/' \
+		-e 's/@MIN_TVOS_SDK_VERSION@/$(MIN_TVOS_SDK_VERSION)/' \
+		-e 's/@MIN_OSX_SDK_VERSION@/$(MIN_OSX_SDK_VERSION)/' \
 		-e 's/@MIN_MACCATALYST_SDK_VERSION@/$(MIN_MACCATALYST_SDK_VERSION)/g' \
+		\
```
rolfbjarne, line 11: Did you forget to push any changes to tools/common/SdkVersions.cs.in? Because these changes will only have any effect if that file is modified as well.

rolfbjarne (changes requested):

tools/common/Make.common
spouliot: yes (good catch) and no :-)

I have local changes for the file, to expose the `DOTNET*` versions when `NET` is defined. I did not include it inside _this_PR since it was not used (and I could have split the above change across diff PR too).

One of the next PR will include it as it enables introspection to do more accurate testing (and detect attributes that are not required, when older than the minimum). Those will come with a lot of binding changes (to make the test pass).

rolfbjarne (approved):

PR #10826, opened 2021-03-10
[runtime] Simplify calls to xamarin_create_managed_ref.

rolfbjarne:
I don't see why we should avoid calling xamarin_create_managed_ref from
NSObject's managed code, and then immediately call xamarin_create_managed_ref
upon return from NSObject's managed code.

This code is old ([1]), and from my reading of it, there's no specific reason
it was done this way.

So simplify the logic to call xamarin_create_managed_ref from a single place
(NSObject's managed code).

[1]: https://github.com/xamarin/maccore/commit/e59c45d3f9408d2b61f47ae90d126612fbb6351b


dalexsoto (approved):

spouliot (approved):
agreeing with the logic :-)

tj-devel709 (approved):

PR #10846, opened 2021-03-11
[builds] Remove the code to optionally inject an x64 slice into our binaries.

rolfbjarne:
It's been disabled for a while, which indicates that Apple has fixed the issue
on their side, and we don't need this code anymore.


dalexsoto (approved):

spouliot (approved):
cut!

PR #10848, opened 2021-03-11
[runtime] Speed up make

rolfbjarne:
Make variables can have dashes, which means we don't have to convert dashes in
rids to underscores to be able to compose variable names.

This speeds up make because now we don't have to execute hundreds of
subprocesses when parsing the makefile.

Before:

    $ make -j16 > /dev/null && /usr/bin/time make && /usr/bin/time make && /usr/bin/time make
        2.11 real         0.57 user         1.26 sys
        2.15 real         0.57 user         1.28 sys
        2.17 real         0.58 user         1.30 sys

After:

    $ make -j16 > /dev/null && /usr/bin/time make && /usr/bin/time make && /usr/bin/time make
        0.52 real         0.18 user         0.25 sys
        0.52 real         0.18 user         0.25 sys
        0.52 real         0.18 user         0.26 sys

So now it's ~4x faster (1.6s) for make to figure out there's nothing to do.

### DIFF
--- a/runtime/Makefile
+++ b/runtime/Makefile
@@ -602,27 +602,27 @@ $(foreach platform,$(DOTNET_CORECLR_PLATFORMS),$(foreach rid,$(DOTNET_$(platform
 
 # a few lookup tables, because the data we have is not always in the format we need it
 
-DOTNET_ios_x64_ARCHITECTURES=x86_64
-DOTNET_ios_x86_ARCHITECTURES=x86
-DOTNET_ios_arm_ARCHITECTURES=armv7 armv7s
-DOTNET_ios_arm64_ARCHITECTURES=arm64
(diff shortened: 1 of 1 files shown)


spouliot (approved):
not quite sure if I'm hyphened or dashed by this ?!?

dalexsoto (approved):
Oh I wonder how you found about this 😅

rolfbjarne:
> Oh I wonder how you found about this 😅

Oh, that's quite simple. After running `make`, running `make` again without any changes should be quite fast. It wasn't (and that's very noticeable 😄)

rolfbjarne:
> not quite sure if I'm hyphened or dashed by this ?!?

The only thing I know is that this is certainly a very dashing PR 😄

PR #10851, opened 2021-03-12
[UIKit] Update bindings for Xcode 12.5 Beta 1 to Beta 3

dalexsoto:


src/UIKit/UIListSeparatorConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Runtime.InteropServices;
+using Foundation;
+using ObjCRuntime;
+
+namespace UIKit {
+	public partial class UIListSeparatorConfiguration {
+
+		public static NSDirectionalEdgeInsets AutomaticInsets { get; private set; }
+
+		static UIListSeparatorConfiguration ()
```
spouliot, line 21: hmm... that makes the property "un-link'able" since the .cctor will always be preserved if the type is used

Maybe just lazy fetch (and cache) it on `get` ?

spouliot (approved):

src/UIKit/UIListSeparatorConfiguration.cs
dalexsoto: Who can think of the linker!!! Thanks good catch!

tests/monotouch-test/UIKit/UIListSeparatorConfigurationTest.cs
```diff
@@ -0,0 +1,36 @@
+//
+// UIListSeparatorConfiguration.cs
+//
+// Authors:
+//	Alex Soto <alexsoto@microsoft.com>
+//
+// Copyright (c) Microsoft Corporation.
+//
+
+#if __WATCHOS__ || __IOS__
```
rolfbjarne, line 10: I think you need to fix this condition too.

rolfbjarne (approved):

tests/monotouch-test/UIKit/UIListSeparatorConfigurationTest.cs
dalexsoto:
```suggestion
#if __IOS__
```
Good catch

spouliot (approved):

PR #10860, opened 2021-03-12
[coremedia] Update for Xcode 12.5 beta 3

spouliot:


dalexsoto (approved):

mandel-macaque (approved):

mandel-macaque:
Did xhrness behave when you were working in your machine?

rolfbjarne (approved):

spouliot:
> Did xhrness behave when you were working in your machine?

@mandel-macaque Kind of... beside things we already discussed, like:
* test for watchOS 5.0 does not work
* many tests for iOS 64bits hangs the UI (but it runs on the background)

IOW I can run most tests locally - but what happens on bots is a small (macOS only) subset of them.

PR #10870, opened 2021-03-15
[tests] Preserve all test fixtures.

rolfbjarne:
This fixes the wildly diffetent number of tests when running xammac tests with and without linking.

Without linking:

> Tests run: 2446 Passed: 2321 Inconclusive: 9 Failed: 0 Ignored: 125

vs with linking:

> Tests run: 1885 Passed: 1802 Inconclusive: 4 Failed: 0 Ignored: 83

Now we run the same tests either with or without linking.

One test was updated to not fail when linking is enabled.

Also add a test to ensure all future test fixtures are preserved.

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -1150,4 +1150,3 @@ partial class TestRuntime
 		}
 	}
 }
-
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -1150,3 +1150,4 @@ partial class TestRuntime
 		}
 	}
 }
+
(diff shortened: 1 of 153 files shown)


mandel-macaque:
devops is down :/

mandel-macaque (approved):

dalexsoto (approved):

tests/common/TestRuntime.cs
```diff
@@ -1150,3 +1150,4 @@ class LinkerSentinel { }
 		}
 	}
 }
+
```
spouliot, line 1153: that file should not be changed

spouliot (approved):

tj-devel709 (approved):

PR #10877, opened 2021-03-15
[CI] Fix configuration step. Keys are unique in yaml.

mandel-macaque:


tj-devel709 (approved):

dalexsoto (approved):

mandel-macaque:
The issue meant that the CI will not start, the fact that it starts shows the fix is correct.

PR #10888, opened 2021-03-16
[NSObject] Rework the SuperHandle property to not return a pointer into managed memory.

rolfbjarne:
Previously, the SuperHandle property returned a pointer into the middle of the
NSObject instance. This is not safe, because we can't make pin the NSObject
instance past the return of the SuperHandle property getter.

So rework the logic, to allocate a chunk of native memory that SuperHandle can
return, and use that instead. This means a little bit of book-keeping to
create and destroy this native memory.

Unfortunately it will also increase the memory usage for each NSObject
instance, but I don't see a way around that, while still fixing the problem
with SuperHandle, given that we can't change the signature of SuperHandle.

### DIFF
--- a/src/Foundation/NSObject2.cs
+++ b/src/Foundation/NSObject2.cs
@@ -25,6 +25,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Threading;
 #if !NO_SYSTEM_DRAWING
 using System.Drawing;
 #endif
@@ -65,9 +66,8 @@ namespace Foundation {
 		// replace older Mono[Touch|Mac]Assembly field (ease code sharing across platforms)
 		public static readonly Assembly PlatformAssembly = typeof (NSObject).Assembly;
 
-		// The order of 'handle' and 'class_handle' is important: do not re-order unless SuperHandle is modified accordingly.
 		IntPtr handle;
-		IntPtr class_handle;
+		IntPtr super; /* objc_super* */
 		Flags flags;
 
 		// This enum has a native counterpart in runtime.h
@@ -81,6 +81,12 @@ namespace Foundation {
 			HasManagedRef = 32,
 			// 64, // Used by SoM
 			IsCustomType = 128,
+		}
+
+		[StructLayout (LayoutKind.Sequential)]
+		struct objc_super {
+			public IntPtr Handle;
+			public IntPtr ClassHandle;
 		}
 
 		bool disposed { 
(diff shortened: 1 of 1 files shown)


src/Foundation/NSObject2.cs
```diff
@@ -739,6 +768,16 @@ internal void ClearHandle ()
...
 				}
+			} else {
+				FreeData ();
+			}
+		}
+
+		unsafe void FreeData ()
+		{
+			if (super != IntPtr.Zero) {
+				Marshal.FreeHGlobal (super);
```
mandel-macaque, line 779: From the docs:
```
You can use FreeHGlobal to free any memory from the global heap allocated by AllocHGlobal, ReAllocHGlobal, or any equivalent unmanaged API method. If the hglobal parameter is IntPtr.Zero the method does nothing.
```
I think we can remove the need of the if at all, just do the call and set super to be IntPtrZero.

src/Foundation/NSObject2.cs
```diff
@@ -155,6 +161,34 @@ internal static IntPtr CreateNSObject (IntPtr type_gchandle, IntPtr handle, Flag
...
+					ptr = Marshal.AllocHGlobal (sizeof (objc_super));
+					*(objc_super*) ptr = default (objc_super); // zero fill
+				}
+
+				var previousValue = Interlocked.CompareExchange (ref super, ptr, IntPtr.Zero);
+				if (previousValue != IntPtr.Zero) {
+					// somebody beat us to the assignment.
+					Marshal.FreeHGlobal (ptr);
+					ptr = IntPtr.Zero;
+				}
```
mandel-macaque, line 179: AFAIK, same, not really need to check against IntPtr.Zeo

mandel-macaque (approved):
I don't think we need the ifs, does it hurt, probably not. Yet again, with the if, it does read better if you do not know the docs of FreeHGlobal. Your call.

src/Foundation/NSObject2.cs
rolfbjarne: Yes, here we do, because we're not comparing and freeing the same pointer.

src/Foundation/NSObject2.cs
```diff
@@ -155,6 +161,34 @@ internal static IntPtr CreateNSObject (IntPtr type_gchandle, IntPtr handle, Flag
...
+				unsafe {
+					ptr = Marshal.AllocHGlobal (sizeof (objc_super));
+					*(objc_super*) ptr = default (objc_super); // zero fill
+				}
+
+				var previousValue = Interlocked.CompareExchange (ref super, ptr, IntPtr.Zero);
+				if (previousValue != IntPtr.Zero) {
+					// somebody beat us to the assignment.
+					Marshal.FreeHGlobal (ptr);
+					ptr = IntPtr.Zero;
```
spouliot, line 178: minor: the local `ptr` is not used after `Marshal.FreeHGlobal` so it does not have to be set back to `IntPtr.Zero`

spouliot (approved):

dalexsoto (approved):

src/Foundation/NSObject2.cs
```diff
@@ -83,6 +83,12 @@ internal enum Flags : byte {
 			IsCustomType = 128,
 		}
 
+		[StructLayout (LayoutKind.Sequential)]
+		struct objc_super {
+			public IntPtr Handle;
+			public IntPtr ClassHandle;
+		}
```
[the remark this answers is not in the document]
rolfbjarne: Not really, but it turns out I went pretty much full circle: https://github.com/xamarin/maccore/commit/5625f84f6c80e9b7e9a8a61477cb6158d7bc810c

src/Foundation/NSObject2.cs
rolfbjarne: I like it this way: the JIT/AOT compiler should be able to optimize the assignment away, and if the code is refactored on day it's easier to not forget that this pointer has been freed.

filipnavara:
Are there any plans to backport this to branches? Alternatively, what is the expected release with this change?

rolfbjarne:
@filipnavara there are no plans to backport this, since we haven't any reported bugs about it (it's also a dangerous change in that it touches some very basic code, which makes it less likely to be back ported). The expected release would be the next major release that branches off `main`, which will be `d16-10`.

PR #10889, opened 2021-03-16
[runtime] Resolve symlinks using API provided by NSString instead of Mono.

rolfbjarne:
Mono's way doesn't work with CoreCLR anyway.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -1444,7 +1444,6 @@ xamarin_get_bundle_path ()
 
 	NSBundle *main_bundle = [NSBundle mainBundle];
 	NSString *bundle_path;
-	char *result;
 
 	if (main_bundle == NULL)
 		xamarin_assertion_message ("Could not find the main bundle in the app ([NSBundle mainBundle] returned nil)");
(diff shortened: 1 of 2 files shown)


runtime/runtime.m
```diff
@@ -1460,10 +1459,7 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
 	bundle_path = [main_bundle bundlePath];
 #endif
 
-	result = mono_path_resolve_symlinks ([bundle_path UTF8String]);
-
-	x_bundle_path = strdup (result);
-	mono_free (result);
+	x_bundle_path = strdup ([[bundle_path stringByStandardizingPath] UTF8String]);
```
mandel-macaque, line 1462: That call does a lot of path checks! Had not idea.

mandel-macaque (approved):

runtime/runtime.m
spouliot: docs mention
> Note that the path returned by this method may still have symbolic link components in it.

which seems to be a problem only when relative paths are used. Not sure if the mono version handled that (or not) but the call to `bundlePath` (earlier) returns a full path (so it's fine).

spouliot (approved):
less code == 🎉

dalexsoto (approved):

runtime/runtime.m
Therzok: This looks neat and something we may want to use in VSMac for some path operations.

PR #10772, opened 2021-03-03
Bump to .NET 6.0.100-preview.3.21161.23

rolfbjarne:

PR #10501, opened 2021-01-22
Bump .NET to 6.0.100-alpha.1.21064.27.

rolfbjarne:

PR #10890, opened 2021-03-16
[build] Use arcade dependency management tooling

pjcollins:
Context: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md

The dotnet core engineering group has some dependency management tooling
(known as `darc`) that we'd like to adopt.  Integrating this tooling in
to the build system will make it easier to stay up to date with the
latest .NET 6 SDK changes.

Many dotnet repos use a [publishing workflow][0] that will push build
artifact data to a central location known as the "Build Asset Registry".
This data includes a "[Channel][1]" association, which is the key to
dependency updating.  Local updates and automatic update "Subscriptions"
compare the version files in a given repo against the product versions
available in the channel that you are interested in.

We hope to be able to publish Xamarin SDK build information to this
central registry in the near future, so that other products can take a
dependency on us as needed (MAUI for instance).

The `darc` tool looks for four different files in a repo when adding a
dependency or when checking for an update:

 * eng/Version.Details.xml
 * eng/Versions.props
 * nuget.config
 * global.json

Both of the `Version` files present in the `eng` folder are updated when
a new dependency is available.

To work with `darc` locally you'll need to install the global tool, join
the `arcade-contrib` GitHub team, and configure your [auth settings][2].

To add a new dependency, use the [`darc add-dependency`][3] command:

    darc add-dependency -n Microsoft.NetCore.App.Runtime.ios-arm64 -t product -v 6.0.0-preview.2.21154.6 -r https://github.com/dotnet/runtime

To update all dependencies, use the [`darc update-dependencies`][4]
command:

    darc update-dependencies --channel ".NET 6"

To configure automatic updates, use the [`darc add-subscription`][5]
command to enroll a target repo/branch into updates from a particular
channel:

     darc add-subscription --channel ".NET 6" --source-repo https://github.com/dotnet/installer --target-repo https://github.com/xamarin/xamarin-macios --target-branch main --update-frequency everyWeek --standard-automerge

Once a subscription is configured, [a pull request][6] will be created
automatically by the dotnet Maestro bot when dependency updates are
available.

[0]: https://github.com/dotnet/arcade/blob/681511f2f63a3563494f1f27904b2842abef6b35/Documentation/CorePackages/Publishing.md#what-is-v3-publishing-how-is-it-different-from-v2
[1]: https://github.com/dotnet/arcade/blob/main/Documentation/BranchesChannelsAndSubscriptions.md#branches-channels-and-subscriptions-explained
[2]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#setting-up-your-darc-client
[3]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#add-dependency
[4]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#update-dependencies
[5]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#add-subscription
[6]: https://github.com/xamarin/xamarin-android/pull/5744


Make.config
```diff
@@ -508,12 +508,12 @@ DOTNET_FEED_DIR ?= $(DOTNET_DESTDIR)/nuget-feed
 # avoid consuming a lot of disk space (since they're never automatically deleted). The system-dependencies.sh script will install locally as long
 # as there's a TARBALL url.
 
+-include $(TOP)/net6.config
```
rolfbjarne, line 511: Running msbuild to generate this file seems a bit heavy, when a simple make target can do it (in the top-level Makefile):

```make
net6.config: env/Versions.props
    grep MicrosoftDotnetSdkInternalPackageVersion env/Versions.props | sed "some expression here" > $@.tmp
    mv $@.tmp $@
```

NuGet.config
```diff
@@ -1,18 +1,17 @@
...
-		<add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" protocolVersion="3" />
-		<add key="Dotnet arcade" value ="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
-		<add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
-		<add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
-		<add key="xamarin-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/xamarin-impl/nuget/v3/index.json" />
-	</packageSources>
-	<disabledPackageSources>
-		<clear />
-	</disabledPackageSources>
+  <config>
```
rolfbjarne, line 3: The changes in this file looks like whitespace only changes?

eng/Versions.props
```diff
@@ -0,0 +1,8 @@
+<Project>
```
rolfbjarne, line 1: Is this file modified with the automatic bump logic too?

pjcollins: Yeah, `darc update-dependencies` will update both `Version.Details.xml` and `Versions.props`. Here's an example of what that looks like: https://github.com/xamarin/xamarin-android/pull/5744/files

NuGet.config
pjcollins: The `darc` tool formatted this file as part of dependency updating.  I figure it's best to include those formatting changes in this initial PR so future auto PRs will only change version information.

@rolfbjarne one other related issue that I wanted your opinion on is that the update tooling currently requires that a `global.json` file exist in the root of the directory. I am going to test this later today, but I think we may need to change the way the `global.json` is generated as part of this PR as well. Do you have any initial thoughts on a preferred way to include a default `global.json`? It can be empty, it just needs to exist for the tool to run.

Make.config
```diff
@@ -508,12 +508,12 @@ DOTNET_FEED_DIR ?= $(DOTNET_DESTDIR)/nuget-feed
...
+
 BUNDLED_NETCORE_PLATFORMS_PACKAGE_VERSION:=$(shell if test -f $(TOP)/builds/BundledNETCorePlatformsPackageVersion.txt; then cat $(TOP)/builds/BundledNETCorePlatformsPackageVersion.txt; else echo "run-make-in-builds-directory-first"; fi)
 
-DOTNET6_VERSION=6.0.100-preview.2.21155.3
 DOTNET6_VERSION_BAND=$(firstword $(subst -, ,$(DOTNET6_VERSION)))
-DOTNET6_URL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-preview.2.21155.3/dotnet-sdk-6.0.100-preview.2.21155.3-osx-x64.pkg
-DOTNET6_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/6.0.100-preview.2.21155.3/dotnet-sdk-6.0.100-preview.2.21155.3-osx-x64.tar.gz
+DOTNET6_TARBALL=https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$(DOTNET6_VERSION)/dotnet-sdk-$(DOTNET6_VERSION)-osx-x64.tar.gz
```
spouliot, line 516: @rolfbjarne that was requested (thrice iirc) and you mentioned using a variable would break other stuff (not that I can remember what exactly)

mandel-macaque: @spouliot  the reason was: https://github.com/xamarin/xamarin-macios/pull/10501#discussion_r562704517

mandel-macaque (changes requested):
As mentioned in the comment: https://github.com/xamarin/xamarin-macios/pull/10501#discussion_r562704517 we should not use the variable, or fix the dependencies.sh script.

Make.config
pjcollins: I think the changes in https://github.com/xamarin/xamarin-macios/pull/9885/files#diff-9e58bdf3f9a2fcd07e4cc0d4647a855b75995005a790f785d4ba6333d1fa7872L1021 removed the dependency that `system-dependencies.sh` had on the `DOTNET6*` variables. I removed `DOTNET6_URL` in this PR because I think that was the only place referencing it previously (and nothing seems to reference it now). We can generate the URL variable declaration as well if needed though.

rolfbjarne: This is an example of my left hand not knowing what my right hand does 🤦‍♂️ 

@pjcollins is right, we only look for `DOTNET_VERSION` now in system-dependencies.sh, the `DOTNET6_*` variables don't matter anymore.

Makefile
```diff
@@ -11,6 +11,10 @@ endif
 # Common
 #
 
+net6.config: eng/Versions.props
+	grep MicrosoftDotnetSdkInternalPackageVersion eng/Versions.props | sed 's/<*\/*MicrosoftDotnetSdkInternalPackageVersion>//g' | sed 's/ */DOTNET6_VERSION=/' > $@.tmp
+	mv $@.tmp $@
```
rolfbjarne, line 16: This makes it a bit more silent unless a verbose build is requested:

```suggestion
	$(Q) grep MicrosoftDotnetSdkInternalPackageVersion eng/Versions.props | sed 's/<*\/*MicrosoftDotnetSdkInternalPackageVersion>//g' | sed 's/ */DOTNET6_VERSION=/' > $@.tmp
	$(Q) mv $@.tmp $@
```

mk/versions.mk
```diff
@@ -42,7 +42,7 @@ test-$(1)::
 	@echo "   $(2)_VERSION=$($(2)_VERSION)"
 	@echo "   $(2)_PATH=$($(2)_PATH) => $(abspath $($(2)_PATH))"
 
-reset-$(1)::
+reset-$(1):: net6.config
```
rolfbjarne, line 45: This isn't needed, Make will automatically make included files, so this line from above has the same effect:

```make
-include $(TOP)/net6.config
```

```suggestion
reset-$(1)::
```

NuGet.config
rolfbjarne: If need a `global.json` in the repo, we have to stop generating it (or modifying it during the build), otherwise things will become annoying quite fast (with the file always showing up as modified).

I think the best option would be to:

* Stop generating it
* Commit the current generated result
* Compute DOTNET_VERSION from it
* Modify system-dependencies.sh to grep global.json instead of Make.config for the .NET version.

rolfbjarne (changes requested):

NuGet.config
mandel-macaque: Question, an we just do the trick I do with make + sed to stop needed to parse? Call make system-dependecies.sh and take the values from make.config.. I know, it add an extra step. Or maybe, is there a way to import the env values from the makefile?

mandel-macaque (approved):

NuGet.config
pjcollins: A root global.json file has been added, and the new `dotnet.config` make file now includes both `DOTNET_VERSION` and `DOTNET6_VERSION`. This keeps other `$(DOTNET_VERSION)` references working. `DOTNET_URL` has been moved from `Make.config` to `system-dependencies.sh`, and made "evergreen" by using a URL that stays constant (aside from the version).

rolfbjarne (approved):

rolfbjarne:
I think it would be nice to have the explanation of this whole process documented in the repo itself (i.e. the PR description), because almost nobody think think to look at the PR after it's been merged.

Maybe add it as `docs/DARC.md`?

rolfbjarne:
At least some of these failures are due to the bump itself, and will be fixed with #10772.

emaf (approved):

rolfbjarne:
Threading changes broke waits: https://github.com/dotnet/runtime/issues/50245

This is blocked until that bug is fixed (and an updated .NET 6 version is available), then this PR will have to be updated with the new .NET 6 version.

PR #10900, opened 2021-03-17
[dotnet] Add an IsCoreCLR initialization flag.

rolfbjarne:
So that we can detect in managed code whether we're running with CoreCLR or
MonoVM.

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -127,6 +127,9 @@ namespace ObjCRuntime {
 			DynamicRegistrar		= 0x04,
 			/* unused				= 0x08,*/
 			IsSimulator				= 0x10,
+#if NET
+			IsCoreCLR				= 0x20,
+#endif
 		}
 
 #if MONOMAC
@@ -158,6 +161,15 @@ namespace ObjCRuntime {
 					return (Flags & InitializationFlags.IsSimulator) == InitializationFlags.IsSimulator;
 				}
 			}
+
+#if NET
+			// Future optimization potential: make the linker change this into a constant.
+			internal bool IsCoreCLR {
+				get {
+					return (Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
+				}
+			}
+#endif
 		}
 
 		internal static unsafe InitializationOptions* options;
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -132,6 +132,7 @@ enum InitializationFlags : int {
 	InitializationFlagsDynamicRegistrar			= 0x04,
 	/* unused									= 0x08,*/
 	InitializationFlagsIsSimulator				= 0x10,
+	InitializationFlagsIsCoreCLR                = 0x20,
 };
 
 struct InitializationOptions {
@@ -1393,6 +1394,10 @@ xamarin_initialize ()
 	options.size = sizeof (options);
 #if MONOTOUCH && (defined(__i386__) || defined (__x86_64__))
 	options.flags = (enum InitializationFlags) (options.flags | InitializationFlagsIsSimulator);
+#endif
+
+#if defined (CORECLR_RUNTIME)
+	options.flags = (enum InitializationFlags) (options.flags | InitializationFlagsIsCoreCLR);
 #endif
 
 	options.Delegates = &delegates;


dalexsoto (approved):

src/ObjCRuntime/Runtime.cs
```diff
@@ -127,6 +127,9 @@ internal enum InitializationFlags : int {
 			DynamicRegistrar		= 0x04,
 			/* unused				= 0x08,*/
 			IsSimulator				= 0x10,
+#if NET
+			IsCoreCLR				= 0x20,
```
spouliot, line 131: TLA -> `Clr` but it's not public so it's not a big deal

src/ObjCRuntime/Runtime.cs
```diff
@@ -158,6 +161,15 @@ internal unsafe struct InitializationOptions {
 					return (Flags & InitializationFlags.IsSimulator) == InitializationFlags.IsSimulator;
 				}
 			}
+
+#if NET
+			// Future optimization potential: make the linker change this into a constant.
```
spouliot, line 166: better file an issue :)

spouliot (approved):

tj-devel709 (approved):

src/ObjCRuntime/Runtime.cs
rolfbjarne: > better file an issue :)

Here it is: https://github.com/xamarin/xamarin-macios/issues/10903

src/ObjCRuntime/Runtime.cs
```diff
@@ -158,6 +161,15 @@ internal unsafe struct InitializationOptions {
...
+			internal bool IsCoreCLR {
+				get {
+					return (Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
```
mandel-macaque, line 169: nit: I like to use https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?view=net-5.0 easier for new developers and does the same.

mandel-macaque (approved):

mandel-macaque:
Probably we want to rebase and execute all tests with what we have in main

src/ObjCRuntime/Runtime.cs
filipnavara: CoreCLR would be able to optimize readonly static with initilizer btw.

rolfbjarne: > CoreCLR would be able to optimize readonly static with initilizer btw.

That would require a refactoring to do the initialization from a static constructor (which is not impossible, just a bit more work).

Also that does not make the linker able to remove additional dependencies that results from constant propagation:

```csharp
if (IsCoreCLR) {
    DoThisThing ();
} else {
    DoThisOtherThing ();
}
```

using the linker to replace IsCoreCLR with a constant value, also makes the linker able to remove the call to either `DoThisThing` or `DoThisOtherThing`, along with the corresponding implementation, and so on, and this can get quite big.

src/ObjCRuntime/Runtime.cs
rolfbjarne: Not that it really matters (this property isn't going to be called a billion times), but it's a lot slower...

```
$ dotnet run
TestHasFlag: 20686
InLine: 2635
```

Test case: https://gist.github.com/rolfbjarne/d75f107521a0d7a3e452df7ad7ab8904

For curiosity I tried mono as well, and mono is much faster than .NET in this case:

```
$ csc *.cs /nologo && mono *.exe
TestHasFlag: 1808
InLine: 973
```

src/ObjCRuntime/Runtime.cs
filipnavara: Sure, it makes sense to link it out as a constant for the final deployment. The `static readonly` would however work even for the development scenario without linker. The JIT would be able to eliminate the other branch just like linker can.

src/ObjCRuntime/Runtime.cs
mandel-macaque: What are they doing to make a flag check slow??????

filipnavara: That's really weird. The call should be "intrinsic" in the runtime, for .NET 6 with both MonoVM and CoreCLR and even in interpreter (iirc). You are likely hitting the JIT tiering. The microbenchmarks with BenchmarkDotNet easily fix that measurement mistake (and there's a `dotnet new benchmark` template so there's no excuse to not use it; https://github.com/dotnet/BenchmarkDotNet/blob/10abc4f4daa5142f0257c9d512a38afc2ab28be1/docs/articles/guides/dotnet-new-templates.md).

filipnavara: To settle the debate on (value & E.A) == E.A vs value.HasFlag(E.A) I run the benchmark with BDN:
```
|      Method |   Mean |  Error |
|------------ |-------:|-------:|
| TestHasFlag | 0.0 ns | 0.0 ns |
|  TestInLine | 0.0 ns | 0.0 ns |
```
(netcoreapp3.1, running on M1 under Rosetta)

That said, on the first run it measured some difference (but likely a fluke). CoreCLR should generate exactly the same code for both so you will likely observe memory access patterns, tiering and other stuff. Mono also implements special codegen for `HasFlag`. IL is slightly bigger for `HasFlag`, machine code should be identical.

filipnavara: And obligatory Sharplab.io comparison: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKBoDtt8ZcAHbMGAAgAs8Y6ArvgD6wfmG40A3jU5zOxAMwLSC8gHZZ8mdQCQ8zgG0AYgBtsAc1wBdGvvn8hnAKKcdB3boCCnALydyNDsPACE/TlIg6gMDAF8aLRjOFigASwA3bAwuV0zTAS5/ZwA6LwBuBOikhWVgCAhTTgAVZgwACTwzSwAKAEo/AD5OPILijtwui26Sr16K2j1dJU46hubWgEk6ABlUuhg+wc5uka4AMhdS/t8i0vm4mnjqIA==

Looks like there may be small difference in addressing but that's quite specific to the actual usage and may disappear on inlining. (/cc @EgorBo)

EgorBo: @filipnavara that redundant move should gone once jit gets "forward substitution" :) 

@rolfbjarne regarding your benchmark, it seems inliner doesn't want to inline `TestHasFlag` - if you help it by adding `[MethodImpl(MethodImplOptions.AggressiveInlining)]` it should be the same (there will be a small difference - but if you re-order your stopwatches - `TestHasFlag` will be faster)

PR #10905, opened 2021-03-18
[mlcompute] Update for xcode12.5 beta 3

spouliot:


src/mlcompute.cs
```diff
@@ -2121,4 +2207,73 @@ interface MLCInferenceGraph {
...
+		Less = 2,
+		Greater = 3,
+		LessOrEqual = 4,
+		GreaterOrEqual = 5,
+		LogicalAND = 6,
+		LogicalOR = 7,
+		LogicalNOT = 8,
+		LogicalNAND = 9,
+		LogicalNOR = 10,
+		LogicalXOR = 11,
```
dalexsoto, line 2224: `AND` or `And`? TBH my head says uppercase but guidelines says otherwise :P

dalexsoto (approved):

rolfbjarne (approved):

src/mlcompute.cs
mandel-macaque: AND is not and acronym, NADN and NOR.. maybe

mandel-macaque (approved):

Issue #10504, opened 2021-01-22
Xamarin iOS should call monovm_initialize for .NET 6


CoffeeFlux:
`monovm_initialize` is a cleaned-up equivalent to `coreclr_initialize`, which is one of the .NET Core hosting functions in `CoreClrHost.h`.

Signature: https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/mono/mono/mini/mono-private-unstable.h#L30

It needs to be called in net6 embedding projects before the runtime is started. Specifically, what it's doing is:

* Passing in various [runtime properties](https://docs.microsoft.com/en-us/dotnet/core/tutorials/netcore-hosting#step-3---prepare-runtime-properties), a subset of which we support. The important ones are:
  * `TRUSTED_PLATFORM_ASSEMBLIES`: Hard-coded list of assemblies, intended to be used for the BCL and fast-path'd in the loading algorithm. They also have some ALC loading implications (default context only).
  * `APP_PATHS`: Essentially the same as `MONO_PATH`. Prefer `TRUSTED_PLATFORM_ASSEMBLIES` where possible.
  * `NATIVE_DLL_SEARCH_DIRECTORIES`: Extra directories to check when loading native libraries for pinvokes.
  * `System.Globalization.Invariant`: Alternative way to set the `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` env var.
  * `PINVOKE_OVERRIDE`: First-chance native callback to control P/Invoke resolution, see https://github.com/dotnet/runtime/pull/44505
* Adding netcore-specific loader hooks, some of which use those properties
* Setting the runtime to use strict assembly name checking, as we still default to the historical behavior otherwise

The runtime properties are also set in files like `.runtimeconfig.json`, which is parsed in by the default host and passed to `coreclr_initialize`. Mono embedders not using the default host will have to handle this themselves, probably at build time in the case of Xamarin, if they wish to support those files.

Applications not calling `monovm_initialize` may result in a runtime error at some undetermined point in the future.

---

* [x] PINVOKE_OVERRIDE: #10932.
* [x] TRUSTED_PLATFORM_ASSEMBLIES: #12272
* [x] APP_PATHS
* [ ] NATIVE_DLL_SEARCH_DIRECTORIES

---

Time estimate: 2 weeks

rolfbjarne:
OK, I'll have a look at this.

[...]

PR #10909, opened 2021-03-18
[runtime/dotnet] Call coreclr_initialize/monovm_initialize at startup.

rolfbjarne:
We need to call coreclr_initialize/monovm_initialize at startup, so do that.
This is a partial implementation, in that we're not setting all the properties
that we should, and also the PINVOKE_OVERRIDE callback is not doing everything
it should either yet.

Ref: #10504.

### DIFF
--- a/runtime/xamarin/coreclr-bridge.h
+++ b/runtime/xamarin/coreclr-bridge.h
@@ -6,6 +6,9 @@
 #ifndef __CORECLR_BRIDGE__
 #define __CORECLR_BRIDGE__
 
+//#define LOG_CORECLR(...)
+#define LOG_CORECLR(...) fprintf (__VA_ARGS__)
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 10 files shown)


spouliot (approved):

dalexsoto (approved):

runtime/xamarin/coreclr-bridge.h
```diff
@@ -6,6 +6,9 @@
 #ifndef __CORECLR_BRIDGE__
 #define __CORECLR_BRIDGE__
 
+//#define LOG_CORECLR(...)
+#define LOG_CORECLR(...) fprintf (__VA_ARGS__)
```
mandel-macaque, line 10: What's the diff between LOG_CORECLR and LOG_MONOVM ?

mandel-macaque (approved):

runtime/xamarin/coreclr-bridge.h
rolfbjarne: There's no difference really, except that one is used in corelcr-bridge.m and the other in monovm-bridge.m.

PR #10910, opened 2021-03-18
Fixes on iOS.Windows targets

mauroa:
- Fixed iOS Binding projects build in Windows 
- Fixing CoreiOSSdkDirectory definition for legacy project system

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.ObjCBinding.CSharp.After.props
+++ b/msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.ObjCBinding.CSharp.After.props
@@ -1,20 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-	<Import Project="$(MSBuildThisFileDirectory)Xamarin.iOS.Windows.props" Condition="Exists('$(MSBuildThisFileDirectory)Xamarin.iOS.Windows.props')" />
-	
-	<UsingTask TaskName="Xamarin.iOS.Tasks.PrepareObjCBindingNativeFrameworks" AssemblyFile="$(CoreiOSSdkDirectory)Xamarin.iOS.Tasks.dll" />
-	
-	<Import Project="$(MSBuildThisFileDirectory)Xamarin.Messaging.Build.targets" Condition="Exists('$(MSBuildThisFileDirectory)Xamarin.Messaging.Build.targets') And '$(MessagingBuildTargetsImported)' != 'true'" />
-	
-	<Target Name="CopyCompressedNativeFrameworkResources" Condition="'@(_NativeFrameworkResource)' != ''" AfterTargets="_CompressNativeFrameworkResources">
-		<CopyFileFromBuildServer SessionId="$(BuildSessionId)" File="$(IntermediateOutputPath)%(_NativeFramework.Filename)%(_NativeFramework.Extension)" />
-	</Target>
-	
-	<!-- This target copies the ObjCBindingNativeFrameworks to the Mac -->
-	<Target Name="PrepareObjCBindingNativeFrameworks" BeforeTargets="_CompressObjCBindingNativeFrameworkResources">
-		<PrepareObjCBindingNativeFrameworks SessionId="$(BuildSessionId)" ObjCBindingNativeFrameworks="@(ObjCBindingNativeFramework)" />
-	</Target>
-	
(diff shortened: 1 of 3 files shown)


emaf (approved):

rolfbjarne (approved):

mandel-macaque (approved):

msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.ObjCBinding.CSharp.After.props
```diff
@@ -1,20 +1,5 @@
...
-	<Target Name="CopyCompressedNativeFrameworkResources" Condition="'@(_NativeFrameworkResource)' != ''" AfterTargets="_CompressNativeFrameworkResources">
-		<CopyFileFromBuildServer SessionId="$(BuildSessionId)" File="$(IntermediateOutputPath)%(_NativeFramework.Filename)%(_NativeFramework.Extension)" />
-	</Target>
-	
-	<!-- This target copies the ObjCBindingNativeFrameworks to the Mac -->
-	<Target Name="PrepareObjCBindingNativeFrameworks" BeforeTargets="_CompressObjCBindingNativeFrameworkResources">
-		<PrepareObjCBindingNativeFrameworks SessionId="$(BuildSessionId)" ObjCBindingNativeFrameworks="@(ObjCBindingNativeFramework)" />
-	</Target>
-	
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
spouliot, line 2: minor: formatting change (space at end of line)

msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.Windows.props
```diff
@@ -13,8 +13,9 @@ Copyright (C) 2011-2013 Xamarin. All rights reserved.
 -->
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<PropertyGroup>
-		<!-- Naming convention that relies on the fact that the Microsoft.iOS.Sdk pack exists with the same version that the Windows SDK -->
-		<CoreiOSSdkDirectory>$(MSBuildThisFileDirectory.Replace('Microsoft.iOS.Windows.Sdk', 'Microsoft.iOS.Sdk'))</CoreiOSSdkDirectory>
+		<CoreiOSSdkDirectory>$(MSBuildThisFileDirectory)</CoreiOSSdkDirectory>
+		<!-- Naming convention that relies on the fact that the Microsoft.iOS.Sdk pack exists with the same version that the Windows SDK (.net6 only) -->
+		<CoreiOSSdkDirectory Condition="$(MSBuildThisFileDirectory.Contains('Microsoft.iOS.Windows.Sdk'))">$(MSBuildThisFileDirectory.Replace('Microsoft.iOS.Windows.Sdk', 'Microsoft.iOS.Sdk'))</CoreiOSSdkDirectory>
```
spouliot, line 18: not sure why you need to check with `Contains` before doing a `Replace` ?
If the original string does not contain the text then replace simply return the original string, so you're effectively doubling the `Contains` checks (as `Replace` does one)

spouliot (changes requested):

msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.Windows.props
mauroa: @spouliot this is a trick to avoid using `.Replace` on old project systems (non .net SDK). If you are in a non .net SDK installation, the iOS MSBuild content in Windows will be under `Program Files (x86)`. Because the `.Replace` function is getting and encoded value, when you do it being in a sub folder of `Program Files (x86)`, it returns `Program Files %28x86%29` instead, which makes the `UsingTask` statement to fail so the build also fails.

This doesn't happen for .net SDK because the packs are under `Program Files`. This is an issue that we need to discuss with the MSBuild team, but for now we are avoiding the replace function call in old project systems doing the above.
See the current error:

```The "Xamarin.Messaging.Build.Client.Tasks.GenerateBuildSessionId" task could not be loaded from the assembly C:\Program Files %28x86%29\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\iOS\Xamarin.iOS.Tasks.dll. Could not load file or assembly 'file:///C:\Program Files %28x86%29\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\iOS\Xamarin.iOS.Tasks.dll' or one of its dependencies.```

spouliot (approved):

dalexsoto (approved):

PR #10935, opened 2021-03-23
[FileProvider] Update Bindings to Xcode 12.5 Beta 1 to Beta 3

dalexsoto:


src/fileprovider.cs
```diff
@@ -1,4 +1,4 @@
-﻿﻿﻿//
+﻿﻿//
 // FileProvider C# bindings
```
dalexsoto, line 2: Intended change, UTF-8 replacing UTF-8 BOM

rolfbjarne (approved):

src/fileprovider.cs
```diff
@@ -316,6 +323,12 @@ interface NSFileProviderDomain {
 		[Export ("hidden")]
 		bool Hidden { [Bind ("isHidden")] get; set; }
 
+		[Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 326: should not be needed anymore (and it's not on other members)

src/fileprovider.cs
```diff
@@ -673,6 +703,43 @@ interface NSFileProviderManager {
...
+		[NoWatch, NoTV, NoiOS, Mac (11,3)]
+		[Export ("runTestingOperations:error:")]
+		[return: NullAllowed]
+		NSDictionary<INSFileProviderTestingOperation, NSError> GetRunTestingOperations (INSFileProviderTestingOperation[] operations, [NullAllowed] out NSError error);
+
+	}
+
+	interface INSFileProviderPendingSetEnumerator { }
+
+	[Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 730: same

src/fileprovider.cs
```diff
@@ -793,6 +861,12 @@ interface NSFileProviderRequest {
 		[NullAllowed]
 		[Export ("requestingExecutable", ArgumentSemantic.Copy)]
 		NSUrl RequestingExecutable { get; }
+
+		[Unavailable (PlatformName.MacCatalyst)]
+		[NoWatch, NoTV, NoiOS, Mac (11,3)]
+		[Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 867: same

src/fileprovider.cs
```diff
@@ -910,5 +984,290 @@ interface NSFileProviderReplicatedExtension : NSFileProviderEnumerating {
...
+		Lookup = 1,
+		Creation = 2,
+		Modification = 3,
+		Deletion = 4,
+		ContentFetch = 5,
+		ChildrenEnumeration = 6,
+		CollisionResolution = 7,
+	}
+
+	interface INSFileProviderTestingOperation : global::ObjCRuntime.INativeObject { }
```
spouliot, line 1050: why was `: global::ObjCRuntime.INativeObject { }` needed ?

spouliot (approved):
minor

src/fileprovider.cs
dalexsoto: Because it is used inside `NSDictionary<INSFileProviderTestingOperation, NSError>` and the info available just yet.

src/fileprovider.cs
dalexsoto: @spouliot does it hurt to keep it as is (the rest of the old code has it)? or do you want me to go and remove them?

I may also need to remove this from Objective-Sharpie since it is coming from there.

spouliot: they can all be removed together, later
as for sharpie there's other stuff like using the short forms `[NoMacCatalyst]` and `[MacCatalyst (x,y)]` but that might be best to wait to see if other (like dotnet related) changes are also needed

PR #10938, opened 2021-03-23
[runtime] Move the declaration of the MonoObject struct to MonoVM-specific code.

rolfbjarne:
The fields of the MonoObject struct is specific to MonoVM, so this makes sure
we don't accidentally poke into random memory on CoreCLR.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -17,6 +17,10 @@
 #include "runtime-internal.h"
 #include "xamarin/xamarin.h"
 
+#if !defined (CORECLR_RUNTIME)
+#include "xamarin/monovm-bridge.h"
+#endif
+
 #if defined (DEBUG)
 //extern BOOL NSZombieEnabled;
 #endif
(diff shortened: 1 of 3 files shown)


spouliot (approved):

mandel-macaque (approved):

runtime/runtime.m
```diff
@@ -623,21 +632,18 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
...
-} PublicMonoReflectionMethod;
-
 MonoMethod *
 xamarin_get_reflection_method_method (MonoReflectionMethod *method)
 {
 	// COOP: Reads managed memory, needs to be in UNSAFE mode
 	MONO_ASSERT_GC_UNSAFE;
 	
+#if defined (CORECLR_RUNTIME)
+	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
```
tj-devel709, line 642:
```suggestion
	xamarin_assertion_message ("The method %s is not implemented yet for CoreCLR", __func__);
```
Maybe typo?

tj-devel709 (approved):

runtime/runtime.m
dalexsoto: ah nice catch! I had to blink several times (on email) before I could spot the typo haha

dalexsoto (approved):
After the typos are gone 👍

stephen-hawley (approved):

PR #10939, opened 2021-03-23
[runtime] Make the exception_gchandle argument optional for callbacks to managed code in CoreCLR.

rolfbjarne:
If no exception handling is provided when calling a managed delegate from native
code, and the managed code throws, then we'll abort.

It's not entirely clear how we'll handle managed exceptions that go through native
code yet, so this makes the initial implementation easier. By making the exception
handling optional, it'll be easy to find all cases where we need to fix it later,
by making it non-optional. The alternative is to add exception handling code all
over the place that would potentially have to be updated when we figure out exactly
what needs to be done.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -47,4 +47,17 @@ xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, cons
 	return rv == 0;
 }
 
+void
+xamarin_handle_bridge_exception (GCHandle gchandle, const char *method)
+{
+	if (gchandle == INVALID_GCHANDLE)
+		return;
+
+	if (method == NULL)
+		method = "<unknown method";
+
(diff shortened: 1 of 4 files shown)


spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

runtime/coreclr-bridge.m
```diff
@@ -47,4 +47,17 @@
...
 }
 
+void
+xamarin_handle_bridge_exception (GCHandle gchandle, const char *method)
+{
+	if (gchandle == INVALID_GCHANDLE)
+		return;
+
+	if (method == NULL)
+		method = "<unknown method";
```
stephen-hawley, line 57: Did you mean this to be `"<unknown method>"`?

rolfbjarne: Yes, I did.

PR #10987, opened 2021-03-26
[ObjCRuntime] Add a NSObject.HasManagedRef property

rolfbjarne:


src/Foundation/NSObject2.cs
```diff
@@ -264,15 +269,15 @@ internal void MarkDirty (bool allowCustomTypes)
 
 		void CreateManagedRef (bool retain)
 		{
-			flags |= Flags.HasManagedRef;
+			HasManagedRef = false;
```
filipnavara, line 272: should this be `= true`?

rolfbjarne: 🤦

rolfbjarne: fixed

dalexsoto (approved):

spouliot (approved):

mandel-macaque (approved):

tj-devel709 (approved):

PR #11014, opened 2021-03-29
[dotnet] Call monovm_initialize before mono_jit_init

spouliot:
Those are called respectively inside `xamarin_vm_initialize` and
`xamarin_bridge_initialize` functions.

This fix the **AppContext.GetData always return null for iOS** issue
https://github.com/dotnet/runtime/issues/50290

Thanks for @filipnavara for diagnosing this quicker than anyone else!

Added unit tests to ensure `AppContext.GetData` can read back the values
we're providing at startup.


dalexsoto (approved):

spouliot:
Failure is unrelated to PR. Known random issue -> https://github.com/xamarin/maccore/issues/2405

PR #11021, opened 2021-03-29
[Localization] Adding Merge Conditional

tj-devel709:
Adding in a conditional to trigger OneLocBuild task only on merges


mandel-macaque:
Step did not ran as expected. Merging.

PR #11031, opened 2021-03-30
[CI] 50gb is a lot. Reduce to 20gb since our device bots are old.

mandel-macaque:


spouliot (commented):
Last time I ran all tests, including device ones, on my machine it consumed a lot more than 20GB for logs

mandel-macaque:
@spouliot we run less on device

PR #11032, opened 2021-03-30
[CI] Fix pool selector

mandel-macaque:
Use the untrusted pool only when is a PR, use the trusted one in all other cases (we push to origin to build pkgs in some cases and we want to use the trusted pool).

### DIFF
--- a/tools/devops/automation/templates/agent-pool-selector.yml
+++ b/tools/devops/automation/templates/agent-pool-selector.yml
@@ -30,7 +30,6 @@ steps:
 
       [bool] $isTopicBranch = $False
       [bool] $isPullRequest = $False
-      [bool] $isManualBuild = $False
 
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):

tools/devops/automation/templates/agent-pool-selector.yml
```diff
@@ -49,7 +49,7 @@ steps:
       Write-Host "  isTopicBranch: ${isTopicBranch}"
       Write-Host "  isPullRequest: ${isPullRequest}"
 
-      if ($isTopicBranch -or $isPullRequest) {
+      if ($isPullRequest) {
```
rolfbjarne, line 52: Looks like the `isTopicBranch` variable isn't used anymore, so it can be removed (and the comments just below this should be updated to reflect the actual behavior as well).

PR #11036, opened 2021-03-30
[dotnet] Fix device builds

spouliot:
With P3 addition on ICU we must now link the native executable as C++.

Remove an old workaround, in many tests, referencing old (5.0/previews)
packages that caused native link time failures.

ref: https://github.com/mono/linker/issues/1139


spouliot:
This fix the build itself.

There's a crash happening at startup -> https://github.com/dotnet/runtime/issues/50290#issuecomment-809903027

dalexsoto (approved):

mandel-macaque:
I 👀 U

mandel-macaque (approved):

emaf (approved):

Issue #10923, opened 2021-03-19
Xharness auto-selection test is broken


mandel-macaque:
We are running ALL tests on PRs that should just be running monotouch and friends. MTouch tests and friends add several hours that can be avoided.

PR #11035, opened 2021-03-30
[CI][XHarness] Ensure we filter the tests to run.

mandel-macaque:
VSTS does not longer have a file with the pat yet it does allow to use
an env variable with the pat provided by the keyvault. Before this
change we have been running all the tests which results in several extra
hours when we do not need to. For example, if nothing was changed in
msbuild, do not run its tests which are 45 mins long.

Changes are:

* provide pat in the env.
* update xharness to use an env var, do not longer read from a file.

fixes: https://github.com/xamarin/xamarin-macios/issues/10923

PR #11047, opened 2021-03-31
[CI][XHarness] Ensure we filter the tests to run. (#11035)

mandel-macaque:
VSTS does not longer have a file with the pat yet it does allow to use
an env variable with the pat provided by the keyvault. Before this
change we have been running all the tests which results in several extra
hours when we do not need to. For example, if nothing was changed in
msbuild, do not run its tests which are 45 mins long.

Changes are:

* provide pat in the env.
* update xharness to use an env var, do not longer read from a file.

fixes: https://github.com/xamarin/xamarin-macios/issues/10923

### DIFF
--- a/tests/xharness/GitHub.cs
+++ b/tests/xharness/GitHub.cs
@@ -41,9 +41,9 @@ namespace Xharness {
 		{
 			var client = new WebClient ();
 			client.Headers.Add (HttpRequestHeader.UserAgent, "xamarin");
-			var xharness_github_token_file = Environment.GetEnvironmentVariable ("XHARNESS_GITHUB_TOKEN_FILE");
-			if (!string.IsNullOrEmpty (xharness_github_token_file) && File.Exists (xharness_github_token_file))
-				client.Headers.Add (HttpRequestHeader.Authorization, File.ReadAllText (xharness_github_token_file));
+			var xharness_github_token = Environment.GetEnvironmentVariable ("GITHUB_TOKEN");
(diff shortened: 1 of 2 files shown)


mandel-macaque:
Manual backport of https://github.com/xamarin/xamarin-macios/commit/b686ab69c816e319caa44057fb43c2f629905038

dalexsoto (approved):

tests/xharness/GitHub.cs
```diff
@@ -41,9 +41,9 @@ static WebClient CreateClient ()
 		{
 			var client = new WebClient ();
 			client.Headers.Add (HttpRequestHeader.UserAgent, "xamarin");
-			var xharness_github_token_file = Environment.GetEnvironmentVariable ("XHARNESS_GITHUB_TOKEN_FILE");
-			if (!string.IsNullOrEmpty (xharness_github_token_file) && File.Exists (xharness_github_token_file))
-				client.Headers.Add (HttpRequestHeader.Authorization, File.ReadAllText (xharness_github_token_file));
+			var xharness_github_token = Environment.GetEnvironmentVariable ("GITHUB_TOKEN");
+			if (!string.IsNullOrEmpty (xharness_github_token))
```
stephen-hawley, line 45: How important is it that `GITHUB_TOKEN` is present or not?
If not will this fail silently, and is that OK?

stephen-hawley (approved):

PR #11052, opened 2021-03-31
[Localization] Changing the Build.SourceBranchName Variable

tj-devel709:


mandel-macaque (changes requested):
Variable is not expanding correctly.

dalexsoto (approved):

mandel-macaque (approved):
Approving my own changes :(

PR #11060, opened 2021-03-31
[CI] Run the dotnet tests by default ONLY on main.

mandel-macaque:


spouliot (approved):

dalexsoto (approved):

tj-devel709 (approved):

tools/devops/automation/templates/build/build.yml
```diff
@@ -594,6 +594,8 @@ steps:
   timeoutInMinutes: 600
   enabled: ${{ parameters.runTests }}
   env:
+    ${{ if eq(variables['Build.SourceBranchName'], 'main') }}: # only works if you have a main branch
```
rolfbjarne, line 597: This means that we won't run the .NET tests after branching for .NET releases. Is this what we want?

PR #11092, opened 2021-04-02
[CI] Move sign and notarize out.

mandel-macaque:


dalexsoto (approved):

mandel-macaque:
This is an interesting PR, we remove the sign and notarize steps depending on how the build was triggered. We do this via a ${{ if }} expresion and a template. The steps executed are diff:

* From CI: We include the template
* From PR: We skip pluing and steps.

PR #11101, opened 2021-04-02
[CI] Black magic to make life better.

mandel-macaque:
We are doing some black magic. We have several templates that
are executed with different parameters.

The problem with that is that templates cannot be used with the
matrix strategy, so we are doing a little trick based on the following:

1. We can create a template that expands.
2. We can use the each keyword, which allow a loop
3. yaml is a super set of json, therefore, it was jsond dictionaties

We use the parameters as yaml objs (simple json ones) and we pass them.
Now we can:

1. Add more tests with less typing.
2. Modify the tests when we manually trigger a build.

Yes! number 2 is very cool, do you want to remove a stage from a manual
build, just delete it. Do you want to change the device test labels,
just edit the object.

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -14,6 +14,76 @@ parameters:
   type: boolean
   default: true
 
+# We are doing some black magic. We have several templates that 
+# are executed with different parameters. 
+#
+# The problem with that is that templates cannot be used with the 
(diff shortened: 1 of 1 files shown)


mandel-macaque:
Example on how to edit the tests:

Screenshot 2021-04-02 at 18 19 24

dalexsoto (approved):

PR #11105, opened 2021-04-03
Merge xcode12.5 into main

dalexsoto:


dalexsoto:
@spouliot I'll check in the morning, it crashed for me but just `ab.GetSource` tho.

I'll check on the intel one since I was using the M1. Also the bot is running 11.2 so that definitely is > MacCat 14 and the test SetUp already has 

```
// The API here was introduced to Mac Catalyst later than for the other frameworks, so we have this additional check
TestRuntime.AssertSystemVersion (PlatformName.MacCatalyst, 14, 0, throwIfOtherPlatform: false);
```

dalexsoto:
@mandel-macaque looks like the bots are not respecting the `run-all-tests` label :( look at the last run above

spouliot:
> Also the bot is running 11.2 so that definitely is > MacCat 14 and the test SetUp already has

@dalexsoto k, if running on macOS 11.x then it's probably a permission issue (our sim sql hack won't work here). There might be a fix inside the catalyst PR (for intro)

tests/common/TestProjects/SystemMemoryReference/Program.cs
```diff
@@ -1,13 +1,13 @@
...
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine (typeof(ReadOnlySpan<char>).Assembly.FullName);
+            System.Console.WriteLine (typeof (UIKit.UIWindow));
+        }
+    }
+}
```
spouliot, line 13: not clear why this file is changing for xcode12.5

tests/common/TestProjects/SystemMemoryReference/SystemMemoryReference.csproj
```diff
@@ -1,76 +1,76 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="Info.plist" />
+    <None Include="Entitlements.plist" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="4.5.4" />
+  </ItemGroup>
+  <Import Project="$(MSBuildExtensionsPath)\Xamarin\iOS\Xamarin.iOS.CSharp.targets" />
+</Project>
```
spouliot, line 76: not clear why this file is changing for xcode12.5

spouliot (approved):

src/fileprovider.cs
```diff
@@ -1,4 +1,4 @@
-﻿﻿﻿//
```
tj-devel709, line 1: Is this intentional?

tj-devel709 (approved):

tests/common/TestProjects/SystemMemoryReference/Program.cs
dalexsoto: Yes looks like git got 🍌s here since there is really no actual change, same for SystemMemoryReference.csproj :/

src/fileprovider.cs
mandel-macaque: Is an editor either adding or removing a BOM (https://en.wikipedia.org/wiki/Byte_order_mark)

mandel-macaque (approved):

mandel-macaque:
> @mandel-macaque looks like the bots are not respecting the `run-all-tests` label :( look at the last run above

Fixed here: https://github.com/xamarin/xamarin-macios/pull/11144

dalexsoto:
Great, everything else looks good (reason why I removed `run-all-tests`) but monotouch-tests the watch ones are expected, xharness needs a fix but the MacCat is green for me locally :/

PR #11108, opened 2021-04-03
[Make.config] Allow '.' on CURRENT_BRANCH_ALPHANUMERIC to match android package names

dalexsoto:


spouliot (approved):

Make.config
```diff
@@ -57,7 +57,7 @@ CURRENT_BRANCH_SED_ESCAPED:=$(subst |,\|,$(subst &,\&,$(subst $$,\$$,$(subst /,\
 # The branch name in the nuget version has to be alphanumeric, and we follow the semantic versioning spec,
 # which defines "alphanumeric" as the letters, numbers and the dash character (and nothing else).
 # So here we replace all non-alphanumeric characters in the branch name with a dash.
-CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-]' '-')
+CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-.]' '-')
```
rolfbjarne, line 60: @dalexsoto why was this necessary? It makes it possible to create package versions that don't conform to the semantic versioning spec (and contradicts the comment just above).

dalexsoto: It was to match the naming of android packages but IIRC it was revealed later

PR #11115, opened 2021-04-05
Revert "[Make.config] Allow '.' on CURRENT_BRANCH_ALPHANUMERIC to match android package names"

dalexsoto:
Reverts xamarin/xamarin-macios#11108

Not needed since we will only do this for previews and we will be hardcoding the branch name at branching time.

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -57,7 +57,7 @@ CURRENT_BRANCH_SED_ESCAPED:=$(subst |,\|,$(subst &,\&,$(subst $$,\$$,$(subst /,\
 # The branch name in the nuget version has to be alphanumeric, and we follow the semantic versioning spec,
 # which defines "alphanumeric" as the letters, numbers and the dash character (and nothing else).
 # So here we replace all non-alphanumeric characters in the branch name with a dash.
-CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-.]' '-')
+CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-]' '-')
 
 # Get the current hash
 CURRENT_HASH:=$(shell git log -1 --pretty=%h)


Make.config
```diff
@@ -57,7 +57,7 @@ CURRENT_BRANCH_SED_ESCAPED:=$(subst |,\|,$(subst &,\&,$(subst $$,\$$,$(subst /,\
 # The branch name in the nuget version has to be alphanumeric, and we follow the semantic versioning spec,
 # which defines "alphanumeric" as the letters, numbers and the dash character (and nothing else).
 # So here we replace all non-alphanumeric characters in the branch name with a dash.
-CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-.]' '-')
+CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9-]' '-')
```
tj-devel709, line 60:
```suggestion
CURRENT_BRANCH_ALPHANUMERIC:=$(shell export LANG=C; printf "%s" "$(CURRENT_BRANCH)" | tr -c '[a-zA-Z0-9]' '-')
```
A little confused about this. Does the "-." represent a character range to the wildcard for any character? If so, should the '-' also be removed?

mandel-macaque (approved):

Make.config
mandel-macaque: I am approving since it is a revert. But since we are using tr, I would have opted for the new syntax [:lower:] [:upper:] and [:alnum:]. The reason for that would be to be able to ensure that the correct use of the locale's collation order, please read the compatibility section in tr:
```
COMPATIBILITY
     Previous FreeBSD implementations of tr did not order characters in range expressions according to the current locale's collation order, making it possible to convert unaccented Latin characters (esp. as found in English text) from upper to lower case using the traditional UNIX idiom of ``tr A-Z a-z''.  Since tr now obeys
     the locale's collation order, this idiom may not produce correct results when there is not a 1:1 mapping between lower and upper case, or when the order of characters within the two cases differs.  As noted in the EXAMPLES section above, the character class expressions ``[:lower:]'' and ``[:upper:]'' should be used
     instead of explicit character ranges like ``a-z'' and ``A-Z''.
```

@tj-devel709 regarding your question, we are including the char - since it is not in a rage, that is, we are not doing a-z so it will be considered to be a collection with a-z A-Z 0-9 and - that is why it is added. When declaring a range in a regex you need to be careful since

1. - can be decorated as a character (I prefer the notation with - as the first char to avoid confusion)
2. The characters connected by a - have to be in order. 9-0 is not a valid range.
3. Funny things like [A-z] can happen an introduce bugs.

We now have two problems.

PR #11129, opened 2021-04-06
[d16-9][CI] Black magic to make life better. (#11101)

mandel-macaque:
We are doing some black magic. We have several templates that
are executed with different parameters.

The problem with that is that templates cannot be used with the
matrix strategy, so we are doing a little trick based on the following:

1. We can create a template that expands.
2. We can use the each keyword, which allow a loop
3. yaml is a super set of json, therefore, it was jsond dictionaties

We use the parameters as yaml objs (simple json ones) and we pass them.
Now we can:

1. Add more tests with less typing.
2. Modify the tests when we manually trigger a build.

Yes! number 2 is very cool, do you want to remove a stage from a manual
build, just delete it. Do you want to change the device test labels,
just edit the object.


mandel-macaque:
CI is failing because the github token is used in a diff way, will have to fix that properly :/

tj-devel709 (approved):

mandel-macaque:
Failures are unrelated with the changes.

PR #11131, opened 2021-04-06
[CI] Simplify if for stage selection.

mandel-macaque:


mandel-macaque:
CI triggered (all stages): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4626657&view=results
PR triggered (just sim tests): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4626658&view=results

tools/devops/automation/build-pipeline.yml
```diff
@@ -131,8 +131,6 @@ variables:
   value: true
 - name: TeamName
   value: 'xamarin-macios'
-- name: isPR
-  value: $[ or(eq(variables['Build.Reason'], 'PullRequest'), and(eq(variables['Build.SourceBranchName'], 'merge'), or(eq(variables['Build.Reason'], 'Manual'), eq(variables['Build.Reason'], 'IndividualCI')))) ]
```
dalexsoto, line 135: Not sure about this one, this has a more explicit intent for me, I mean the `isPR` variable reads better than the condition. I kinda prefer readability over one liners

```yaml
- ${{ if eq(variables['isPR'], 'False') }}:
- ${{ if or(eq(variables['Build.Reason'], 'IndividualCI'), eq(variables['Build.Reason'], 'Manual')) }}:
```

mandel-macaque: The main reason is that variables for templates just can use default variables. It is a better idea to keep it in the new way so that we know that the correct default variables are used. Else we will always get a false, reference: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script#can-parameters-be-set-based-on-variables

dalexsoto (approved):

PR #11136, opened 2021-04-06
[Localization] Changes to use Resx files from OneLocBuild instead of xlf files

tj-devel709:
These are the changes I believe are necessary to transition from using the xliff nuget and xlf files to use the new resx files from the OneLocBuild task in our localization build. **This currently only contains changes to the msbuild localization but I will soon add the changes needed for mtouch localization as well.**

# Checked the Packages and inside Packages/msbuild/iOS the translation dlls are in the same spot.

## Old method
Created translations for: 
- cs
- de
- es
- fr
- it
- ja
- ko
- pl
- pt-BR
- ru
- tr
- zh-Hans
- zh-Hant

## New Method (OneLocBuild)
Creates translation for:
- cs
- de
- es
- fr
- **hu (Hungarian)**
- it
- ja
- ko
- **nl (Dutch)**
- pl
- pt-BR
- ru
- **sv (Swedish)**
- tr
- zh-Hans
- zh-Hant

So moving to use OneLocBuild task will actually include 3 more locales in our localization (Hungarian, Dutch, and Swedish)

Since the current resx files from OneLocBuild are not translated yet, I altered error E0007 in all the resx files and some of the test conditions to make sure that they are using these resx files. They are working as expected.


mandel-macaque:
@tj-devel709 something is very wrong with this PR, you I'm sure you have not changed 400+ files for this, have you?

PR #11144, opened 2021-04-07
[CI] Fix wrong fwd labels.

mandel-macaque:
We are trying to join the json rather than the collection of strings.
$tag is a collection we can join.


dalexsoto (approved):
join all the things!

mandel-macaque:
152 tests thanks to the bcl, so it is a success :)

mandel-macaque:
Do not backport to xcode12.5 since I did a commit in the merge PR with this change.

Issue #11145, opened 2021-04-07
[net6] iOS release binaries are dynamically linking to .dylib


spouliot:
see `dotnet/targets/Xamarin.Shared.Sdk.targets`

```xml
<ResolvedFileToPublish
   Update="@(ResolvedFileToPublish)"
  RelativePath="$([MSBuild]::MakeRelative($(MSBuildProjectDirectory)$(PublishDir),$(_DylibPublishDir)))%(Filename)%(Extension)"
  Condition="'%(Extension)' == '.dylib'" />
```

This copies all `.dylib` into the final app bundle (.app).

This greatly increase the size of the `.app` since there are duplicates and all (or at least part) of the code is also statically linked into the main executable binary. This is visible from the reports included in https://github.com/xamarin/xamarin-macios/issues/10249

This is **not** allowed by App Store rules (dynamic linking is only possible to user frameworks, not `dylib`).

Note: Removal of the `.dylib` leads to crash at runtime so there **is** dynamic linking being done.

```
MySingleView -> /Users/poupou/git/xamarin/xamarin-macios/tests/dotnet/size-comparison/MySingleView/dotnet/bin/Debug/net6.0-ios/ios-arm64/MySingleView.dll
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  Installing application bundle 'com.xamarin.mycomparableapp' on 'Mercure'
  Application bundle 'com.xamarin.mycomparableapp' installed on 'Mercure'
  Launched application 'com.xamarin.mycomparableapp' on 'Mercure' with pid 3113
  xamarin_vm_initialize (2, 0x16d643130, 0x16d643120): rv: 0
  2021-04-06 20:37:50.830 MySingleView[3113:3659312] Xamarin.iOS: An exception occurred when calling Runtime.Initialize:
  libSystem.Native (System.DllNotFoundException)
     at System.Threading.LowLevelMonitor.Initialize()
     at System.Threading.WaitSubsystem.ThreadWaitInfo..ctor(Thread thread)
     at System.Threading.Thread.EnsureWaitInfo()
     at System.Threading.Thread.InitializeCurrentThread()
     at System.Threading.Thread.get_CurrentThread()
     at System.Threading.SynchronizationContext.SetSynchronizationContext(SynchronizationContext syncContext)
     at UIKit.UIApplication.Initialize()
[...]

PR #11153, opened 2021-04-07
[dotnet][msbuild] Do not copy `.pdb` files inside `.app` for release builds

spouliot:
The original goal was to remove the `.dylib` from the app bundle.
However some are required, even in release, so that will need to be
fixed first [1]

[1] https://github.com/xamarin/xamarin-macios/issues/11145


mandel-macaque:
@spouliot notes or no notes, that is the question

dalexsoto (approved):

mandel-macaque (approved):

tj-devel709 (approved):

PR #11158, opened 2021-04-07
[dotnet] Handle `lib` prefix for P/Invokes for runtime native libraries

filipnavara:
Contributes to #11145

### DIFF
--- a/tools/common/Target.cs
+++ b/tools/common/Target.cs
@@ -786,9 +786,15 @@ namespace Xamarin.Bundler {
 					else
 						mono_native_lib = app.GetLibNativeName () + ".dylib";
 					sw.WriteLine ();
+#if NET
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Native\", NULL, \"{mono_native_lib}\", NULL);");
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Security.Cryptography.Native.Apple\", NULL, \"{mono_native_lib}\", NULL);");
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Net.Security.Native\", NULL, \"{mono_native_lib}\", NULL);");
+#else
 					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"System.Native\", NULL, \"{mono_native_lib}\", NULL);");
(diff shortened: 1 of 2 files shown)


tools/common/Target.cs
```diff
@@ -786,9 +786,15 @@ void GenerateIOSMain (StringWriter sw, Abi abi)
 					else
 						mono_native_lib = app.GetLibNativeName () + ".dylib";
 					sw.WriteLine ();
+#if NET
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Native\", NULL, \"{mono_native_lib}\", NULL);");
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Security.Cryptography.Native.Apple\", NULL, \"{mono_native_lib}\", NULL);");
+					sw.WriteLine ($"\tmono_dllmap_insert (NULL, \"libSystem.Net.Security.Native\", NULL, \"{mono_native_lib}\", NULL);");
```
spouliot, line 792: `libSystem.IO.Compression.Native.dylib` might be needed too.
It's shipped so it's likely used somewhere in the BCL (have not checked).

That can wait for the removal of the .dylib from the app bundle since it's the only way to test it properly.

spouliot (approved):

tools/common/Target.cs
filipnavara: It's not needed for "Hello World" but it's needed in general. I hesitate to do more changes here since it's likely going to be replaced by `PINVOKE_OVERRIDE` callback anyway (unless there are some AOT implications from using dllmap).

filipnavara:
The tests fail because of https://github.com/dotnet/runtime/issues/47910. This is expected and I can add a workaround.

mandel-macaque (approved):

dalexsoto (approved):

mandel-macaque:
Something really bad is going on in certain bots.

filipnavara:
JFYI I re-run some of the timed out tests locally and they worked.

filipnavara:
The last remaining failure (`introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): LaunchFailure`) happened on other PRs too (even the empty test one from few days ago) so I consider it unrelated.

spouliot:
Yes, that's an unrelated CI/xharness issue that we're tracking internally - https://github.com/xamarin/maccore/issues/2411

spouliot (approved):

PR #11165, opened 2021-04-08
Bump maccore to fix mlaunch NRE when starting Simulator bundles

premun:
New commits in xamarin/maccore:
 
* [xamarin/maccore@c89fd6a](https://github.com/xamarin/maccore/commit/c89fd6a) Fix `--launchsimbundleid` action (#2412)

Diff: https://github.com/xamarin/maccore/compare/b9aaee72545d046614dd5fa77e947a9372381013..c89fd6a694110794099592488bf7f15915d99e74


spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

mandel-macaque:
@premun  looks like we have failures and the are related with the sim.

premun:
@mandel-macaque I might have actually created the commit on an old commit from main. I checked the logs and it doesn't seem related. Also the command that I was fixing is not used in this repository but hopefully I didn't introduce some side effect. 

I rebased so let's try again.

spouliot:
@premun this is because of the P4 bump that was auto merged, even if not tested (and incorrect).

You can just merge `main` back into your PR and it should be fine (since it was reverted).

mandel-macaque:
Ah! much better!

dalexsoto (approved):

PR #11180, opened 2021-04-09
[Build] Add a rule to execute the first run of Xcode. Why copy paste a command?

mandel-macaque:


Makefile
```diff
@@ -147,6 +147,9 @@ fix-install-permissions:
 fix-xcode-select:
 	sudo xcode-select -s $(XCODE_DEVELOPER_ROOT)
 
+fix-xcode-first-run:
+	$(XCODE_DEVELOPER_ROOT)/usr/bin/xcodebuild -runFirstLaunch
+
```
spouliot, line 152: > Why copy paste a command?

because no one will know about that make target ? ;-)

mandel-macaque: I can add a little string :)

spouliot (approved):

dalexsoto (approved):

tj-devel709 (approved):

PR #11194, opened 2021-04-12
[CI] If the device tests could not be provisioned do communicate it.

mandel-macaque:
Rather than let the pipeline continue doing steps that do not make
sense, report the issue to be reviewed and reported to the bot admins.

The need of that template and extra env variable is due to the lack of
proper scripting in yaml so I have to get creative.


mandel-macaque:
Do not merge until we have device test results.

tj-devel709 (approved):

dalexsoto (approved):

Issue #11151, opened 2021-04-07
dontlink.app crash with `EntryPointNotFoundException: xamarin_release_managed_ref`


spouliot:
### Steps to Reproduce

1.  Build dontlink.csproj for device (arm64)
2. Run on device

### Expected Behavior

Should run the tests

### Actual Behavior

```
2021-04-07 10:40:24.537 dont link[3286:3879143] 
Unhandled Exception:
System.EntryPointNotFoundException: xamarin_release_managed_ref
  at Foundation.NSObject.ReleaseManagedRef () [0x0001f] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:285 
  at Foundation.NSObject.Dispose (System.Boolean disposing) [0x00025] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:772 
  at Foundation.NSObject.Dispose () [0x00000] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:155 
  at ObjCRuntime.Runtime.ThreadPoolDispatcher (System.Func`1[TResult] callback) [0x0000f] in /Users/poupou/git/xamarin/xamarin-macios/src/ObjCRuntime/Runtime.cs:303 
  at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback () [0x00009] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/threading/threadpool.cs:1258 
2021-04-07 10:40:24.539 dont link[3286:3879143] Unhandled managed exception: xamarin_release_managed_ref (System.EntryPointNotFoundException)
  at Foundation.NSObject.ReleaseManagedRef () [0x0001f] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:285 
  at Foundation.NSObject.Dispose (System.Boolean disposing) [0x00025] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:772 
  at Foundation.NSObject.Dispose () [0x00000] in /Users/poupou/git/xamarin/xamarin-macios/src/Foundation/NSObject2.cs:155 
[...]

PR #11196, opened 2021-04-13
[runtime] Fix EntryPointNotFoundException: xamarin_release_managed_ref

spouliot:
Revert https://github.com/xamarin/xamarin-macios/commit/a5f970069aa541a0a27d52ba0fcb74ad427ab057

Fix issue https://github.com/xamarin/xamarin-macios/issues/11151

PR #11205, opened 2021-04-14
[d16-10][msbuild] Fix copy/paste error from #11196

spouliot:
Not sure how I could mess up a copy/paste between file - but I succeeded
to create a subtle bug (would have expected it more crash happy)

Fix issue #11151

Revert the previous (revert) fix (but not the build fixes)


dalexsoto (approved):

mandel-macaque:
You should use vim :P yy

mandel-macaque (approved):

PR #11211, opened 2021-04-14
[static-registrar] Re-order some code in ToObjCParameterType

spouliot:
This avoid some computation that might not be needed.

It also avoids a `NullReferenceException` when updating to net6 preview 4
builds (where `ResolveType` can return `null` but the type reference was
enough to complete the job).


dalexsoto (approved):

mandel-macaque (approved):

spouliot:
failure is a unrelated, known issue -> https://github.com/xamarin/maccore/issues/2411

PR #11242, opened 2021-04-20
[CI] Reduce the number of statuses when we build nugets.

mandel-macaque:
We move from having a status per nugets to two statues:

1. xamarin-macios (Nugets built) -  to let us know that we did build
   nugets.
2. xamarin-macios (Nuvets published) - to let us know tat we did pubish
   the nugets.

### DIFF
--- a/tools/devops/automation/templates/build/build-nugets.yml
+++ b/tools/devops/automation/templates/build/build-nugets.yml
@@ -1,0 +1,39 @@
+# all steps that are required to build the nugets and publish them
+
+steps:
+
+- bash:  $(Build.SourcesDirectory)/xamarin-macios/tools/devops/automation/scripts/bash/build-nugets.sh
+  displayName: 'Build Nugets'
+  condition: and(succeeded(), contains(variables['configuration.BuildNugets'], 'True'))
+  continueOnError: true # should not stop the build since is not official just yet.
+  timeoutInMinutes: 180
+
+- script: >-
+    dotnet build -t:GenerateBuildAssetRegistryManifest /v:n
+    $(Build.SourcesDirectory)/xamarin-macios/dotnet/package/Microsoft.iOS.Ref/package.csproj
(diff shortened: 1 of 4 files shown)


tj-devel709 (approved):

tools/devops/automation/templates/build/build-nugets.yml
```diff
@@ -0,0 +1,39 @@
+# all steps that are required to build the nugets and publish them
+
+steps:
+
+- bash:  $(Build.SourcesDirectory)/xamarin-macios/tools/devops/automation/scripts/bash/build-nugets.sh
```
rolfbjarne, line 5: What happens if this produces at least one nuget, but still fails?

rolfbjarne (approved):

tools/devops/automation/templates/build/build-nugets.yml
mandel-macaque: It will continue to the next steps due to the continueOnError: true

rolfbjarne: And then it looks like it will set the "Nugets built" status to "Success", or am I reading this wrong?

mandel-macaque: We can remove the continueOnError and do an on failed() but the we need to add branching for the other code. I think that is better to do it in a diff PR since the code for building is just a move from the main build.yml to build-nugget.yml

But you are definitely right this is at least fishy and should have been fixed when we landed the 'Generate and Upload Build Asset Registry Manifest' step, because it is also going to return an error on a continueOnError :/

mandel-macaque: Issue https://github.com/xamarin/maccore/issues/2419 I'll fix it in the morning.

mandel-macaque:
Known failure: https://github.com/xamarin/maccore/issues/1605

PR #11249, opened 2021-04-21
[CI] Fix import issue in device test templates.

mandel-macaque:
We only have one checkout which changes the path. This is a "feature" from VSTS.


spouliot (approved):

dalexsoto (approved):

mandel-macaque:
Works <3

PR #11250, opened 2021-04-21
[CI][VSTS] Ensure statuses are set in the correct commit and not the merge commit.

mandel-macaque:
On a pull request, Azure DevOps does not build the exact version of the code that you pushed,
but a version merged with your target branch. This means that the commit
we see in the Buid.Revision is not the one of the last commit of the PR,
which makes some statuses get "lost" and checks be missed.

We make a distintion on which variable to used base on the build reason.

PS: Change is clearer this way that add logic in the yml.


mandel-macaque:
The governance check already shows it is a correct fix :)

spouliot (approved):

dalexsoto (approved):

rolfbjarne (approved):

filipnavara (approved):
🎉

PR #11251, opened 2021-04-21
[build] Create Microsoft.iOS.Windows.Sdk workload pack

pjcollins:
Converts the Microsoft.iOS.Windows.Sdk NuGet package into a proper
[workload SDK pack][0].  The entry point for this pack has been changed,
and it is now imported through the `WorkloadManifest.targets` file
included in `Microsoft.NET.Workload.iOS`, rather than being imported
directly from `Microsoft.iOS.Sdk`.

Import ordering has otherwise changed slightly.  The following files are
now imported before the majority of the `Microsoft.iOS.Sdk` (and the 
majority of the .NET SDK targets):

 * Xamarin.iOS.Common.Before.props
 * Xamarin.iOS.Common.Before.targets

After this the majority of the .NET SDK targets will load, followed by
the `Microsoft.iOS.Sdk` targets. Finally, everything declared in the
`<AfterMicrosoftNETSdkTargets/>` hook loads, which consists of:

 * Microsoft.iOS.Windows.Sdk.targets
 * tools/msbuild/*

[0]: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-manifest.md#sdk-packs


dotnet/Makefile
```diff
@@ -108,13 +106,13 @@ endef
 $(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call SupportedTargetPlatforms,$(platform))))
 
 define WorkloadTargets
-Microsoft.NET.Workload.$(1)/WorkloadManifest.json: targets/WorkloadManifest.template.json Makefile $(TOP)/Make.config.inc $(TOP)/.git/HEAD $(TOP)/.git/index
+Microsoft.NET.Workload.$(1)/WorkloadManifest.json: Makefile $(TOP)/Make.config.inc $(TOP)/.git/HEAD $(TOP)/.git/index
```
rolfbjarne, line 109: This way the file will be updated when the source changes (and also you won't have to change the line below):

```suggestion
Microsoft.NET.Workload.$(1)/WorkloadManifest.json: $4 Makefile $(TOP)/Make.config.inc $(TOP)/.git/HEAD $(TOP)/.git/index
```

dotnet/targets/Xamarin.Shared.Sdk.props
```diff
@@ -9,6 +9,8 @@
 	<Import Project="Xamarin.Shared.Sdk.TargetFrameworkInference.props" />
 
 	<PropertyGroup>
+		<!-- Set to true when using the Microsoft.<platform>.Sdk NuGet. This is used by pre-existing/shared targets to tweak behavior depending on build system -->
+		<UsingAppleNETSdk>true</UsingAppleNETSdk>
```
rolfbjarne, line 13: Renaming `_UsingXamarinSdk` to `UsingAppleNETSdk` can be split out into a separate PR.

dotnet/Makefile
```diff
@@ -123,7 +121,8 @@ Microsoft.NET.Workload.$(1)/LICENSE: $(TOP)/LICENSE
 WORKLOAD_TARGETS += Microsoft.NET.Workload.$(1)/WorkloadManifest.json
 WORKLOAD_TARGETS += Microsoft.NET.Workload.$(1)/LICENSE
 endef
-$(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call WorkloadTargets,$(platform),$(shell echo $(platform) | tr A-Z a-z),$($(platform)_NUGET_VERSION_NO_METADATA))))
+$(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call WorkloadTargets,$(platform),$(shell echo $(platform) | tr A-Z a-z),$($(platform)_NUGET_VERSION_NO_METADATA), \
+	$(if $(findstring $(platform),$(DOTNET_WINDOWS_PLATFORMS)),targets/WorkloadManifest.windows.template.json,targets/WorkloadManifest.template.json))))
```
rolfbjarne, line 125: I think this will stop generating workloads for platforms included in DOTNET_WINDOWS_PLATFORMS, are you sure you don't want to add a line instead:

```suggestion
$(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call WorkloadTargets,$(platform),$(shell echo $(platform) | tr A-Z a-z),$($(platform)_NUGET_VERSION_NO_METADATA,targets/WorkloadManifest.template.json))))
$(foreach platform,$(DOTNET_WINDOWS_PLATFORMS),$(eval $(call WorkloadTargets,$(platform),$(shell echo $(platform) | tr A-Z a-z),$($(platform)_NUGET_VERSION_NO_METADATA,targets/WorkloadManifest.windows.template.json))))
```

dotnet/targets/Xamarin.Shared.Sdk.props
pjcollins: Do we want one of these per platform/SDK type, or is `UsingAppleNETSdk` preferred? Android currently sets `UsingAndroidNETSdk`.  I suppose something more granular could also be added in the future if needed.

pjcollins: Moved to https://github.com/xamarin/xamarin-macios/pull/11270.

dotnet/Makefile
pjcollins: I want to use the new "windows" template file for all workloads that include a windows SDK.  These changes _should_ result in the `Microsoft.NET.Workload.iOS` workload declaring this new Windows only pack, however since it's conditionally imported in `WorkloadManifest.targets` it will never need to be installed on macOS.  This all seemed to work locally but it looks like CI didn't run for this yesterday so I'm still waiting to check an official build.  I don't want to introduce new Windows specific workloads (just one new workload pack).  We should be able to continue to ship the same `Microsoft.NET.Workload.iOS` workload on both macOS and Windows.

rolfbjarne (approved):

dotnet/targets/WorkloadManifest.windows.template.json
```diff
@@ -0,0 +1,32 @@
+{
+	"version": 5,
+	"workloads": {
```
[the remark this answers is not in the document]
pjcollins: The reasoning behind the two files is that `WorkloadManifest.template.json` is used for all other platforms (mac, watch, tv), and these workloads don't (and some will never?) have a `*.Windows.Sdk` entry.

Rather than adding more complex string manipulation to `dotnet/Makefile` to conditionally add this windows entry to _only_ the iOS manifest, I added a new template for workloads that should include a "Windows" Sdk.

We may want add the `"platforms": [ "win-x86", "win-x64" ]` declaration here, though the conditional import in the iOS `WorkloadManifest.targets` file allows the new manifest which declares a windows specific pack to work on macOS when the pack is not installed.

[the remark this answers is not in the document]
jonathanpeppers: Ok makes sense 👍 

I think we need to put `platforms` so that the future `dotnet workload install` commands won't install these Windows packs on Mac.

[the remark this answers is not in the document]
pjcollins: It looks like this feature may not be working yet, I'm going to revert this portion for now https://github.com/dotnet/sdk/issues/17151

jonathanpeppers (approved):

rolfbjarne (approved):

rolfbjarne:
Waiting on review from @emaf to merge.

emaf (approved):

Issue #11256, opened 2021-04-21
CoreCLR: NSAutoreleasePool for every thread


rolfbjarne:
CoreCLR has support for an NSAutoreleasePool for thread pool work items (https://github.com/dotnet/runtime/pull/47592), but it's missing support for having an NSAutoreleasePool for every thread.

This issue is to track the required work in CoreCLR and in our own code.

Ref: https://github.com/dotnet/designs/pull/197#issuecomment-818561306
Ref: https://github.com/dotnet/designs/pull/197#issuecomment-823907238
Ref: https://github.com/dotnet/runtime/pull/52023

AaronRobinsonMSFT:
@jkoritzinsky and @kouvel Do either of you have thoughts on how to extend the work in https://github.com/dotnet/runtime/pull/47592 to enable support for all threads?

akoeplinger:
@rolfbjarne with the support added in https://github.com/dotnet/runtime/pull/52023 we can use the `System.Threading.Thread.EnableAutoreleasePool` switch for Mono too and replace the custom code that uses the mono profiler for hooking thread start events in .NET 6 right?

rolfbjarne:
@akoeplinger yes, that would be ideal.

PR #11254, opened 2021-04-21
[CoreCLR] Implement mono_assembly_get_object.

rolfbjarne:
Next failure is now:

> monotouchtest[27142:4675921] Xamarin.Mac: The method mono_profiler_install has not been implemented for CoreCLR.

PR #11257, opened 2021-04-21
[runtime] Move the code to wrap every managed thread with an NSAutoreleasePool to the MonoVM bridge.

rolfbjarne:
The current implementation is MonoVM-specific, and won't work with CoreCLR
(something else will have to be implemented for CoreCLR, which is tracked
here: https://github.com/xamarin/xamarin-macios/issues/11256).

Next failure is now (after #11254 is merged as well):

> monotouchtest[50744:5001174] Xamarin.Mac: The method mono_jit_exec has not been implemented for CoreCLR.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -49,6 +49,13 @@ xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, cons
 }
 
 void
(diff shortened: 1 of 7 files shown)


spouliot (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2420)

PR #11270, opened 2021-04-21
[msbuild] Rename $(_UsingXamarinSdk) to $(UsingAppleNETSdk)

pjcollins:
The `$(_UsingXamarinSdk)` property has been renamed to help improve
external usability.  This change increases parity with the Android SDK,
which currently defines `$(UsingAndroidNETSdk)` for internal and
external use.  The property has also been moved so that it will be
set as early as possible, before the .NET Sdk targets are imported.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.props
+++ b/dotnet/targets/Xamarin.Shared.Sdk.props
@@ -9,6 +9,8 @@
 	<Import Project="Xamarin.Shared.Sdk.TargetFrameworkInference.props" />
 
 	<PropertyGroup>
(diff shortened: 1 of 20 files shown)


spouliot (approved):

dalexsoto (approved):

jonathanpeppers (approved):
👍 I wanted to change this, too, but never got around to it.

emaf (approved):

rolfbjarne (approved):

mandel-macaque:
Test can be ignored. Is known.

PR #11271, opened 2021-04-21
[runtime] Move xamarin_create_managed_ref internal call to managed code.

rolfbjarne:
Move the xamarin_create_managed_ref internal call to managed code, to ease things
with CoreCLR.

In order to preserve performance, this wasn't a straight forward port.

* monotouch_create_managed_ref used to detect if there already was a GCHandle for
  a native object. To avoid a managed->native transition, this logic has now been
  moved into the code that sets the GCHandle (the `xamarinSetGCHandle:flags:` / `xamarin_set_gchandle_trampoline`
  code), and these methods return a value saying whether the GCHandle was set or
  not.

* xamarin_create_gchandle will check the retain count to determine whether to create
  a weak or a strong GCHandle for the managed object. In this particular case we
  should never need to create a strong GCHandle, which means that we don't need to
  check the retain count (saving a managed->native transition).

~Performance measurements: slightly faster in all test cases I could come up with.~

* ~'new NSObject ()': 3% faster~
* ~'new CustomObject ()': 6% faster~
* ~Creating an NSObject in native code, and getting the managed wrapper: 0.6% faster~
* ~Creating a CustomObject in native code (the type being defined in managed code): 2% faster~

~All measurements done in an iOS simulator on an iMac Pro.~

Using the new perftest (#11298), I get very similar numbers for both old code and new code: https://gist.github.com/rolfbjarne/e0fc2ae0f21da15062b4f051138679af (multiple runs). Sometimes the old code is faster, sometimes the new code is faster (although the old code tends to be the one who wins).

In any case there aren't any significant performance hits due to this change, so it should be good to go.

---

The next failure is now:

> monotouchtest[91755:5770368] Xamarin.Mac: The method mono_jit_thread_attach has not been implemented for CoreCLR.

```
[...]
3   com.xamarin.monotouch-test    	0x000000010a2795d8 xamarin_assertion_message + 408 (runtime.m:1501)
4   com.xamarin.monotouch-test    	0x000000010a249601 mono_jit_thread_attach + 33 (mono-runtime.m:2424)
5   com.xamarin.monotouch-test    	0x000000010a27eb4e xamarin_switch_gchandle + 142 (runtime.m:1830)
6   com.xamarin.monotouch-test    	0x000000010a281cca xamarin_retain_trampoline + 42 (trampolines.m:701)
7   com.xamarin.monotouch-test    	0x000000010a2d53f0 -[Foundation_NSDispatcher retain] + 32 (Microsoft.macOS.registrar.x86_64.m:37192)
8   com.apple.Foundation          	0x00007fff2143b896 -[NSObject(NSThreadPerformAdditions) performSelector:onThread:withObject:waitUntilDone:modes:] + 148
9   com.apple.Foundation          	0x00007fff2143b6a1 -[NSObject(NSThreadPerformAdditions) performSelectorOnMainThread:withObject:waitUntilDone:] + 124
10  com.xamarin.monotouch-test    	0x000000010a290ad9 xamarin_dyn_objc_msgSend + 217 (trampolines-x86_64-objc_msgSend.s:15)
```


spouliot (approved):

tj-devel709 (approved):

dalexsoto (approved):

runtime/runtime.m
```diff
@@ -1882,11 +1900,18 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
...
 {
 	// COOP: no managed memory access: any mode
-	set_gchandle (self, gchandle, flags);
+	return set_gchandle (self, gchandle, flags);
+}
+
+bool
+xamarin_set_gchandle_with_flags_safe (id self, GCHandle gchandle, enum XamarinGCHandleFlags flags)
+{
+	// COOP: no managed memory access: any mode
```
filipnavara, line 1913: This comment sounds outdated.

runtime/xamarin/trampolines.h
```diff
@@ -13,10 +13,12 @@
 extern "C" {
 #endif
 
+// Must be kept in sync with the same enum in NSObject2.cs
 enum XamarinGCHandleFlags : uint32_t {
 	XamarinGCHandleFlags_None = 0,
 	XamarinGCHandleFlags_WeakGCHandle = 1,
 	XamarinGCHandleFlags_HasManagedRef = 2,
+	XamarinGCHandleFlags_InitialSet = 4,
```
mandel-macaque, line 21: What happened with 3?

mandel-macaque (approved):
Small stupid question.

runtime/xamarin/trampolines.h
rolfbjarne: These are flags, so 1 + 2 = 3.

PR #11272, opened 2021-04-21
[CI] Remove Mac OS devices that will not happen.

mandel-macaque:
Remove those devices of Mac OS X that will not happen because we cannot
run the agent in such and old device. This is due to the fact that the
agent runs using c# and we do not support such and old OS.

Also add a new stage to run tests on catalina. ATM older OS as not
working, but Catalina should work with our trusted pool.

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -70,6 +70,12 @@ parameters:
   type: object
   default: [
     {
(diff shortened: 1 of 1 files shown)


spouliot (approved):
😞

dalexsoto (approved):

rolfbjarne (approved):

mandel-macaque:
Known issue: https://github.com/xamarin/maccore/issues/2420

PR #11288, opened 2021-04-22
[mmp] Use the resolver (and configure it correctly) when using --runregistrar.

rolfbjarne:
* Extract the code we use to configure the assembly resolver during a normal
  mmp run to make it usable for --runregistrar.
* Configure the assembly resolver we use for --runregistrar.
* Pass the assembly resolver to the registrar so that it's actually used.
* Adjust the System.Void lookup to look everywhere even if we find a corlib,
  since behavior changes a bit now that we're using an assembly resolver:
	* Previous behavior:
		1. In .NET mode, look for a corlib named System.Private.CoreLib, and fail to find it.
		2. Look in all the loaded assemblies for System.Void (and find it in System.Runtime.dll).
	* Broken behavior as a result of the resolver changes:
		1. Find corlib as System.Private.CoreLib.dll
		2. Fail to find System.Void in System.Private.CoreLib.dll, since we'd only look in corlib.
	* New behavior
		1. Find corlib as System.Private.CoreLib.dll
		2. Fail to find System.Void in System.Private.CoreLib.dll, but find it in System.Runtime.dll,
		   since we're now looking in all the loaded assemblies.

This is required to make VSMac's usage of --runregistrar

### DIFF
--- a/tools/common/Application.cs
+++ b/tools/common/Application.cs
@@ -912,6 +912,7 @@ namespace Xamarin.Bundler {
 				CommandLineAssemblies = RootAssemblies,
 #endif
 			};
+			resolver.Configure ();
 
 			if (Platform == ApplePlatform.iOS && !Driver.IsDotNet) {
 				if (Is32Build) {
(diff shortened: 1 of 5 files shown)


dalexsoto (approved):

Therzok (commented):
Thanks!

PR #11300, opened 2021-04-22
[catalyst] remove splash screens

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/807

When implementing a new `@(MauiSplashScreen)` feature for .NET MAUI, I
found out that Catalyst doesn't have support for splash screens.

I even tested this in Xcode:

* Create a new Objective-C iOS app
* Make `LaunchScreen.storyboard` red and `Main.storyboard` green.
* Check the new `macOS` checkbox in project settings to enable
  Catalyst and be a multiplatform app.

I see red, then green on iOS but only green on macOS. I think Catalyst
completely ignores `UILaunchStoryboardName`.

This removes splash screens from the `dotnet new maccatalyst` template
as well as the `MyCatalystApp` sample project.


spouliot (approved):
falls under the "even if wrong it's easier to add, than to remove, at a later date"

rolfbjarne (approved):

PR #11302, opened 2021-04-22
[msbuild] Removes explicit Mqtt dependency

emaf:
There were 2 "problems":
1- Messaging depends on `System.Net.Mqtt` and not in `System.Net.Mqtt.Server`. It still worked because the latter depends on the former package, but we were restoring unnecessary dependencies.
2- We don't need an excplicit reference, since `Xamarin.Messaging.Build.Client` already depends on it, and by removing it we avoid having another dependency to keep up to date.


spouliot (approved):

rolfbjarne (approved):

mauroa (approved):

rolfbjarne:
Test failures are unrelated

* Xtro/Mac: https://github.com/xamarin/xamarin-macios/issues/11308
* monotouch-test/tvOS - simulator/Debug [dotnet]: https://github.com/xamarin/maccore/issues/1605
* introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): https://github.com/xamarin/maccore/issues/2420

PR #11303, opened 2021-04-22
[CI] Add triggers for new release branches.

mandel-macaque:


spouliot (approved):

mandel-macaque:
No need to wait for the merge, build triggered which is all we need.

PR #11320, opened 2021-04-24
[dotnet] Allow choosing the ICU data file to be used for globalization

spouliot:
Defaults to include all globalization data. Depending on the target
audience it might be possible to reduce the app size.

Draft doc: https://github.com/xamarin/xamarin-macios/wiki/Globalization


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -143,6 +143,10 @@
 		<UseSystemResourceKeys Condition="'$(UseSystemResourceKeys)' == ''">true</UseSystemResourceKeys>
 	</PropertyGroup>
 
+	<PropertyGroup>
```
rolfbjarne, line 146: I don't think we should have the documentation + logic to select the ICU data. The runtime provides the ICU, so that's also where the other logic should live. That way the logic (and variable name) is shared with Android (and any other products that use the ICU data). We really only need the name of the file to embed in the .app.

spouliot: @steveisok any other platform, beside Blazor, using an embedded (not system) ICU like we do ?

steveisok: No, only blazor and iOS right now use an embedded icu.  I wanted android to also have it, but requiring libc++ wasn't going to work.  

I would keep this for now.  We can provide a task, but unsure exactly when that will come.

spouliot: @lewing are you exposing this in any way for blazor ?

If not then I might just stick an underscore to the name and move on... :)

spouliot:
Superceded by https://github.com/xamarin/xamarin-macios/pull/11983

PR #11323, opened 2021-04-24
[CI] Workaorund misconfig pools.

mandel-macaque:
For some reason, some of the internal hosted pools have been set without
the OS version (WRONG WRONG). The simples way is to allow to skip this
check in those pools over re-configuring all bots.

Example of tests starting: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4684353&view=logs&j=ac718a0d-597a-528b-5222-a379ea9341a1

The commit does not fix the tests, just makes sure they run.

'We can lick gravity, but sometimes the paperwork is overwhelming.'


rolfbjarne (approved):

spouliot (approved):

mandel-macaque:
I might need to do the back ports by had.. :(

PR #11334, opened 2021-04-26
Make NSWindow appearanceSource null allowed

Therzok:
Native headers show it's null_resettable.

```objc
/// If set, the receiver will inherit the appearance of that object, as well as use KVO to observe its effectiveAppearance for changes. Typically this is used for child windows that are shown from a parent window or specific view. Defaults to NSApp.
@property (weak, null_resettable) NSObject<NSAppearanceCustomization> *appearanceSource API_AVAILABLE(macos(10.14));
```

### DIFF
--- a/src/appkit.cs
+++ b/src/appkit.cs
@@ -20963,6 +20963,7 @@ namespace AppKit {
 		CGPoint ConvertPointFromBacking (CGPoint point);
 
 		[Mac (10, 14)]
+		[NullAllowed]
 		[Export ("appearanceSource", ArgumentSemantic.Weak)]
 		INSAppearanceCustomization AppearanceSource { get; set; }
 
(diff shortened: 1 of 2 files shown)


spouliot (changes requested):
you'll need to update xtro data files (because the build will, or should, fail)

mandel-macaque:
@spouliot not yet, I have not landed the fix. Added the tag, @Therzok update xtro, with the label it will work.

Therzok:
Should the null-allowed be applied on the setter only?

spouliot:
> Should the null-allowed be applied on the setter only?

No, `[NullAllowed]` does not affect the code generated for the getter, only the setter.

Also we generate C# nullability attributes based on the property.

spouliot (approved):

dalexsoto (approved):

spouliot:
```
12:56:16.7859620 ?fixed-todo? Entry '!missing-protocol-member! MTLDevice::supports32BitFloatFiltering not found' in 'MacCatalyst-Metal.todo' is not found in corresponding 'MacCatalyst-Metal.raw' file
12:56:16.7860210 ?fixed-todo? Entry '!missing-protocol-member! MTLDevice::supports32BitMSAA not found' in 'MacCatalyst-Metal.todo' is not found in corresponding 'MacCatalyst-Metal.raw' file
12:56:16.7860700 ?fixed-todo? Entry '!missing-protocol-member! MTLDevice::supportsQueryTextureLOD not found' in 'MacCatalyst-Metal.todo' is not found in corresponding 'MacCatalyst-Metal.raw' file
```

failures are unrelated (because xtro was not enabled correctly)

PR #11335, opened 2021-04-26
[CI] Fix nuget messaging in case of failure.

mandel-macaque:
-gt instead of >


rolfbjarne (approved):

mandel-macaque:
Failure is unrelated and will be fixed in a diff PR.

PR #11345, opened 2021-04-27
[dotnet] Build the partial static registrar for CoreCLR.

rolfbjarne:
* The generated static registration code will eventually be different.
* The generated code has to be compiled with different compiler flags.

This also required adding a new overload of xamarin_mono_object_release for the generated
code to compile.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -164,6 +164,12 @@ xamarin_mono_object_release (MonoReflectionType **mobj)
 	xamarin_mono_object_release ((MonoObject **) mobj);
 }
 
+void
+xamarin_mono_object_release (MonoString **mobj)
+{
+	xamarin_mono_object_release ((MonoObject **) mobj);
+}
+
 /* Implementation of the Mono Embedding API */
 
 // returns a retained MonoAssembly *
--- a/runtime/xamarin/runtime.h
+++ b/runtime/xamarin/runtime.h
@@ -291,10 +291,11 @@ MonoObject *	xamarin_gchandle_unwrap (GCHandle handle); // Will get the target a
  */
 #if defined(CORECLR_RUNTIME)
 void			xamarin_mono_object_retain (MonoObject *mobj);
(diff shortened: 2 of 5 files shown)


runtime/coreclr-bridge.m
```diff
@@ -164,6 +164,12 @@
 	xamarin_mono_object_release ((MonoObject **) mobj);
 }
 
+void
+xamarin_mono_object_release (MonoString **mobj)
+{
+	xamarin_mono_object_release ((MonoObject **) mobj);
```
mandel-macaque, line 170: Are this things inlined?

mandel-macaque (approved):

runtime/coreclr-bridge.m
rolfbjarne: For non-.NET, they're completely removed (#define'd to empty strings). For .NET, they might be inlined. That said, inlining these methods is unlikely to have any measurable benefit, we're spending a lot of time doing other stuff.

spouliot (approved):

runtime/xamarin/runtime.h
```diff
@@ -291,10 +291,11 @@ MonoObject *	xamarin_gchandle_unwrap (GCHandle handle); // Will get the target a
  */
 #if defined(CORECLR_RUNTIME)
 void			xamarin_mono_object_retain (MonoObject *mobj);
-// Use C++ linking to be able to use method overloading.
+// Use C++ linking to be able to use method overloading, so that callers don't have to cast their variables to 'MonoObject**' (which improves type safety a lot).
```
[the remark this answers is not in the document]
rolfbjarne: This is the only method that will use overloads, and this is the final list:

https://github.com/rolfbjarne/xamarin-macios/blob/2c45133a217796f8609094f94b56f53115d851d5/runtime/xamarin/runtime.h#L318-L322

Also, we'll be able to use `MonoObject` everywhere when we remove support for legacy Xamarin (whenever we can do that).

PR #11352, opened 2021-04-27
[build] Add installer dependency to linker packs

pjcollins:
Adds a coherent parent dependency to `Microsoft.NET.ILLink.Tasks`, which
ensures that it will not be updated past the version included in
`Microsoft.Dotnet.Sdk.Internal`.

These changes allow us to remove our mono/linker darc subscriptions, as
`Microsoft.Dotnet.Sdk.Internal` updates will also bring in the latest
`Microsoft.NET.ILLink.Tasks` that the SDK references.  This will reduce
the number of dependency update PRs created by maestro.

Since the `Microsoft.NET.ILLink.Tasks` and `Microsoft.NET.ILLink` NuGet
packages are created by the same build, we only need to track one of
these package IDs in eng/Version* files.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -4,11 +4,7 @@
       <Uri>https://github.com/dotnet/installer</Uri>
       <Sha>2455d34cebebfa62772f57be93b2cf99d14dbad2</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.ILLink" Version="6.0.100-preview.2.21212.1">
-      <Uri>https://github.com/mono/linker</Uri>
-      <Sha>6bfa2c0657d2dcfd6dce684ab34b470ce567631a</Sha>
-    </Dependency>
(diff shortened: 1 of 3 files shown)


pjcollins:
This will cause issues with open maestro PRs #11235 and #11327 so we'll want to coordinate merge efforts. The linker PR can potentially be closed, as the next update to the SDK after this PR lands will bring in a new linker package version.

rolfbjarne (approved):

spouliot (approved):

rolfbjarne:
@spouliot I think we can merge this, close the other to PRs (#11235 and #11327), and then just wait for / trigger the next bump?

spouliot:
@pjcollins anything else you wanted to add/change before we merge this ? (and close the two other ones)

pjcollins:
There's nothing else that I was planning to change here, feel free to resolve and merge whenever you're ready.

PR #11354, opened 2021-04-27
[runtime] Remove dead code.

rolfbjarne:
This seems to be a leftover from when we were implementing support for third-party crash reporters many, many moons ago.


spouliot (approved):

dalexsoto (approved):
Crash the lytics!

PR #11356, opened 2021-04-27
Use yaml-templates agent pool selector

mjbond-msft:
Related work item: VS #[1322817](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1322817)
Related PR: https://github.com/xamarin/yaml-templates/pull/115

The yaml-templates solution is a migration of the template that used to be in this repo. A key difference is that a change was made so that manually triggered builds target the PR pool and not the CI pool.


mandel-macaque:
Is this a draft?

mandel-macaque (approved):

PR #11373, opened 2021-04-28
[CI] Set provi channel to pr branch

cadsit:
Sets `PROVISIONATOR_CHANNEL` to https://github.com/xamarin/provisionator/pull/402

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -138,6 +138,8 @@ variables:
   value: true
 - name: TeamName
   value: 'xamarin-macios'
+- name: PROVISIONATOR_CHANNEL
+  value: 'pr/cadsit/402'
 
 trigger:
   branches:


mandel-macaque:
We already use the var since I did a fix adding it 3 hours before this one, we will need to merge and backport.

cadsit:
Oddly enough -- I think the revert didn't actually succeed in unblocking the build. I still see failures in d16-9, e.g.

This has landed in provisionator. Using the latest channel should work for CI, but I think it makes sense to keep the variable around, possibly as a parameter, to easily swap between provisionator releases.

cadsit:
For reference: validation build with the latest provisionator change can be found [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4702180&view=logs&j=600d49c6-e96b-5e8b-2c3a-9f351e2e5848&t=12837421-857d-5f7a-7e12-9c60f4caaf40)

PR #11374, opened 2021-04-28
Update to new linker custom steps API

sbomer:
@spouliot @rolfbjarne here's a start on the update to the new custom steps API. It's not complete yet, but I'd be happy to get early feedback if you have a chance to look.

The changes so far are intended to turn steps into `IMarkHandler`s when applicable so that they are triggered by marking:
- PreserveBlockCodeSubStep is triggered for marked types
- ApplyPreserveAttribute is triggered for marked assemblies (and scans for attributes/members in them)
  - This means it won't find preserved things in assemblies that are unused. This is the same compromise made in https://github.com/xamarin/xamarin-android/pull/5748.
  - For now I removed the functionality where an assembly-level attribute may point to another attribute type to indicate that it should be treated like "Preserve", since I don't see a good way to make this work without a pre-scan, but we'll need to figure out what to do here.
- OptimizeGeneratedCodeSubStep is triggered for marked methods
  - I added a cache for checking whether an assembly has optimizable code, since the method processing can be interleaved for methods from different assemblies.
  - It looks like the type hierarchy checks for `IsNsObject` aren't cached because CoreTypeMapStep isn't part of the pipeline. I will look into changing things to use the type resolution cache that @marek-safar added in https://github.com/mono/linker/pull/1979.
- MarkNSObjects is triggered for marked assemblies (and scans for members in them)
  - Like Preserve, this won't discover members in unused assemblies.
  - I believe the check `IsOverriddenInUserCode` will miss overrides that are discovered by the linker later during marking. I think it could be fixed by inverting the logic to check whether a user method is an override of an exported method, and preserving it if it is.
- PreserveSmartEnumConversions is triggered for marked methods, and just marks things (instead of conditionally preserving them like before)


tools/dotnet-linker/SetupStep.cs
```diff
@@ -27,6 +27,17 @@ public class SetupStep : ConfigurationAwareStep {
 			}
 		}
 
+		List<IMarkHandler> _markHandlers;
+		List<IMarkHandler> MarkHandlers {
+			get {
+				if (_markHandlers == null) {
+					var pipeline = typeof (LinkContext).GetProperty ("Pipeline").GetGetMethod ().Invoke (Context, null);
+					_markHandlers = (List<IMarkHandler>) pipeline.GetType ().GetProperty ("MarkHandlers").GetValue (pipeline);
```
spouliot, line 35: is there any plan to make those API public ?

tools/dotnet-linker/Steps/ExceptionalMarkHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+    // that were registered for handling by the subclass.
+	public abstract class ExceptionalMarkHandler : IMarkHandler
+	{
+		public abstract void Initialize (LinkContext context, MarkContext markContext);
+
+		protected DerivedLinkContext LinkContext => Configuration.DerivedLinkContext;
+
+		protected LinkContext context { get; set; }
+
+        protected AnnotationStore Annotations => context.Annotations;
```
spouliot, line 25: minor: alignment

tools/dotnet-linker/Steps/ExceptionalMarkHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+
+using Xamarin.Tuner;
+
+using Mono.Linker;
+using Mono.Linker.Steps;
+
+namespace Xamarin.Linker {
+
+    // Similar to ExceptionalSubStep, but this only runs for marked members
+    // that were registered for handling by the subclass.
```
spouliot, line 16: minor: alignment

tools/dotnet-linker/Steps/PreserveBlockCodeHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+			if (field.Name != "Handler")
+				return;
+
+			// The class has a parameterless 'Invoke' method with a 'MonoPInvokeCallback' attribute
+			if (!type.HasMethods)
+				return;
+			var method = type.Methods.SingleOrDefault (v => {
+				if (v.Name != "Invoke")
+					return false;
+				if (v.Parameters.Count == 0)
```
spouliot, line 68: should be `v.HasParameters` to avoid allocating an always empty collection

tools/dotnet-linker/Steps/PreserveBlockCodeHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			We need to make sure the linker doesn't remove the Handler field
+			and the Invoke method. 
+			*/
+
+			// First make sure we got the right class
+			// The type for the field we're looking for is abstract, sealed and nested and contains exactly 1 field.
+			if (!type.IsAbstract || !type.IsSealed || !type.IsNested)
+				return;
```
spouliot, line 46: `HasFields` should be checked first

tools/dotnet-linker/Steps/PreserveBlockCodeHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+			if (type.Fields.Count != 1)
+				return;
+
+			// The type is also nested inside ObjCRuntime.Trampolines class)
+			var nestingType = type.DeclaringType;
+			if (!nestingType.Is ("ObjCRuntime", "Trampolines"))
+				return;
+
+			// The class has a readonly field named 'Handler'
+			var field = type.Fields.Single ();
```
spouliot, line 56: `[0]` since there can be only one

tools/linker/MonoTouch.Tuner/PreserveSmartEnumConversions.cs
```diff
@@ -142,20 +174,48 @@ void ProcessAttributeProvider (ICustomAttributeProvider provider, MethodDefiniti
...
 			}
 		}
 
 		protected override void Process (MethodDefinition method)
 		{
+#if NET
+			static bool IsPropertyMethod (MethodDefinition method)
+			{
+				return (method.SemanticsAttributes & MethodSemanticsAttributes.Getter) != 0 ||
+					(method.SemanticsAttributes & MethodSemanticsAttributes.Setter) != 0;
```
spouliot, line 191: simpler (more readable) to call `IsGetter` or `IsSetter` on the `method`

https://github.com/jbevain/cecil/blob/634a37d3f2be7b8c72a612fce7b4907a12533d4c/Mono.Cecil/MethodDefinition.cs#L452

spouliot (commented):
Looking good! 
Few comments (some might even be fromcopy/paste code)

tools/dotnet-linker/Steps/ExceptionalMarkHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+using Xamarin.Bundler;
+
+using Xamarin.Tuner;
+
+using Mono.Linker;
+using Mono.Linker.Steps;
+
+namespace Xamarin.Linker {
+
+    // Similar to ExceptionalSubStep, but this only runs for marked members
```
rolfbjarne, line 15: We use tabs for indentation

tools/linker/CoreOptimizeGeneratedCode.cs
```diff
@@ -84,18 +105,42 @@ public override void Initialize (LinkContext context)
 					is_arm64_calling_convention = false;
 				}
 			}
+#if NET
+			this.context = context;
```
rolfbjarne, line 109: Maybe add an `ExceptionalMarkHandler.Initialize (LinkContext)` method that stores the context? That would make the .NET-specific code smaller.

rolfbjarne (commented):
Looks good so far!

tools/dotnet-linker/SetupStep.cs
marek-safar: @sbomer why is it needed?

tools/linker/CoreOptimizeGeneratedCode.cs
```diff
@@ -18,16 +19,31 @@
 
 namespace Xamarin.Linker {
 
+#if NET
```
marek-safar, line 22: I don't want to block this PR but wouldn't be better to adopt this method to net6? Looking at the logic it seems to be a simpler version of dead blocks elimination which linker does automatically. At first look, all the code looks like it can be replaced with few substitutions.

rolfbjarne: Yes, it might be possible to port it to use substitutions. Our list of things to do is pretty big though, and the current code isn't blocking anything.

tools/dotnet-linker/SetupStep.cs
sbomer: This was just to get something working - I am looking into passing all steps individually as --custom-step args.

tools/linker/CoreOptimizeGeneratedCode.cs
sbomer: This should also fix https://github.com/xamarin/xamarin-macios/pull/11374#issuecomment-829557041.

tools/dotnet-linker/SetupStep.cs
sbomer: SetupStep currently logs the pipeline steps which is one case where making this API public could be useful:
https://github.com/xamarin/xamarin-macios/blob/993c7443476b78bbee2a0c7fe7dcebb73778850f/tools/dotnet-linker/SetupStep.cs#L116-L123

tools/dotnet-linker/SetupStep.cs
```diff
@@ -27,78 +27,9 @@ public class SetupStep : ConfigurationAwareStep {
...
-			}
-			DumpSteps ();
-			throw new InvalidOperationException ($"Could not insert {step} after {stepName} because {stepName} wasn't found.");
-		}
 
 		protected override void TryProcess ()
 		{
-			// Don't use --custom-step to load each step, because this assembly
-			// is loaded into the current process once per --custom-step,
-			// which makes it very difficult to share state between steps.
```
rolfbjarne, line 58: It doesn't look like https://github.com/mono/linker/issues/1314 has been fixed yet?

tools/dotnet-linker/SetupStep.cs
marek-safar: I'd like to see that in linker together with timing details under verbose mode

tools/dotnet-linker/SetupStep.cs
sbomer: That's true. @marek-safar has been encouraging the steps to move to the official API despite this limitation. For xamarin-android we avoided shared state by moving a cache to be built-in to the linker (more context [here](https://github.com/xamarin/xamarin-android/pull/5748#discussion_r606634212)). I didn't notice a case (yet) where this would cause problems for xamarin-macios - are you aware of any?

marek-safar: Yeah, we have several mechanisms how to share a state between steps more effectively. The issue can be still fixed but it's probably a low priority now if we don't have valid use case for it

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -321,11 +321,45 @@
...
+			<!-- TODO: these steps should probably run after mark. -->
+			<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" BeforeStep="MarkStep" Condition="'$(_LinkMode)' != 'None'" Type="Xamarin.Linker.Steps.StoreMembersAndAttributesDispatcher" />
+			<!-- IMarkHandlers which run during Mark -->
+			<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" Condition="'$(_LinkMode)' != 'None'" Type="Xamarin.Linker.Steps.PreserveBlockCodeHandler" />
+			<!--
+				We need to run the ApplyPreserveAttribute step even we're only linking sdk assemblies, because even
+				though we know that sdk assemblies will never have Preserve attributes, user assemblies may have
+				[assembly: LinkSafe] attributes, which means we treat them as sdk assemblies and those may have
+				Preserve attributes.
+				TODO: LinkSafeAttribute doesn't appear to be handled anywhere. Is this attribute still supported?
```
spouliot, line 338: yes, but that code might just be (for now) in legacy
not everything was ported to dotnet-linker

spouliot: > builds//mono-ios-sdk-destdir/ios-sources/external/linker/src/tuner/Mono.Tuner/CustomizeActions.cs

The code that handle `[LinkerSafe]` is inside mono (repo) and the legacy SDK uses the source code (for a few files) from the mono package (2020-02) when building `mtouch`.

tools/dotnet-linker/SetupStep.cs
spouliot: > I didn't notice a case (yet) where this would cause problems for xamarin-macios - are you aware of any?

hmm... there's state being shared. Maybe the confusion is that not all _legacy_ steps are present in the dotnet version ?

A, quite important, missing one is `[Core]TypeMapStep`.

It might be better to pause moving customization code to the target file until the port (all features) are complete.

tools/linker/CoreOptimizeGeneratedCode.cs
spouliot: Also much easier to debug / compare by doing this step-by-step.

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -321,11 +321,45 @@
...
+				ApplyPreserveAttribute no longer runs on all assemblies. [assembly: Preserve (typeof (SomeAttribute))] will
+				no longer give SomeAttribute "Preserve" semantics.
+			-->
+			<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" Condition="'$(_LinkMode)' != 'None'" Type="Xamarin.Linker.Steps.ApplyPreserveDispatcher" />
+			<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" Condition="'$(_LinkMode)' != 'None'" Type="Xamarin.Linker.OptimizeGeneratedCodeHandler" />
+			<!-- MarkNSObjects will run for all marked assemblies. -->
+			<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" Condition="'$(_LinkMode)' != 'None'" Type="Xamarin.Linker.Steps.MarkNSobjectsDispatcher" />
```
spouliot, line 345: `MarkNSobjectsDispatcher` should be `NSObject` since it's how the type in question is named

tools/dotnet-linker/SetupStep.cs
rolfbjarne: > I didn't notice a case (yet) where this would cause problems for xamarin-macios - are you aware of any?

Yes, the problem is that you can't share data using types defined in the custom step assembly. The custom step assembly is loaded once for each --custom-step argument, and they're considered as different assemblies, which means the types in the assemblies have different identifies for each time the assembly is loaded. So using this API:

https://github.com/mono/linker/blob/795106705fedfceb6d1de60999673dfe39dc7acf/src/linker/ref/Linker/Annotations.cs#L34-L35

won't work, because the second --custom-step argument will get back an object whose type is from the first --custom-step argument's assembly load.

sbomer: Thanks!
> Maybe the confusion is that not all legacy steps are present in the dotnet version ?

Yes, I meant to say that the already-ported steps seemed to work without shared state. However I think that is also not true for the reason @rolfbjarne pointed out. I had thought that the static `ConditionalWeakTable<LinkContext, LinkerConfiguration>` on `LinkerConfiguration` was somehow avoiding the problem, but now I realize it is the same since `LinkerConfiguration` is loaded once per assembly so its static members exist on different type identities. I will revert those changes until we fix the linker issue.

tools/dotnet-linker/SetupStep.cs
sbomer: Per discussion with @spouliot and @rolfbjarne the plan is to remove the reflection once https://github.com/mono/linker/issues/1314 is fixed, since there's no good workaround for sharing state without it. I also filed https://github.com/mono/linker/issues/2013 about the suggestion to output diagnostic info.

marek-safar: I had quick look mono/linker#1314 won't be probably fixed anytime soon as it needs to introduce complex ALC handling. We could just add generic shared state holder to LinkContext and use that instead.

rolfbjarne: > We could just add generic shared state holder to LinkContext and use that instead.

That won't solve the problem where the custom step assembly has a different identity every time it's loaded (so you can't use custom types when storing state).

tools/linker/CoreOptimizeGeneratedCode.cs
spouliot: I'm not sure substitutions, as available, today would work here. At least not for `IsDirectBinding` since the decision to always make it `true` is based on each individual type (computed in an earlier step) and not on some global state (known before build) given to the linker.

There are other places where substitutions would work (e.g. what remains of `RemoveCode`). We can start there and compare the results (since it won't be identical).

spouliot:
> 4 tests failed, 103 tests passed.

@sbomer those are unrelated to your PR and already fixed in HEAD (just rebase to get the fix)

spouliot (changes requested):
It does seems to break the, just merged, code that enable removing the dynamic registrar.
That makes comparing the other changes much harder than it needs to be.
Investigating...

spouliot:
Looking at the linked code we can see inside `ObjCRuntime.Runtime.Initialize` that

```
if (DynamicRegistrationSupported)
	Registrar = new DynamicRegistrar ();
```
is turned into 
```
Registrar = new DynamicRegistrar ();
```
instead of being removed. That, of course, will mark the `DynamicRegistar` and all code that it use.

That might be a `true` instead of `false` evaluation caused the a change when the optimization was applied, e.g. if applied before we know `DynamicRegistrationSupported` could be `false`.

marek-safar:
> That might be a true instead of false evaluation caused the a change when the optimization was applied,

How do you set it to false nowadays?

spouliot:
> How do you set it to false nowadays?

The decision part was just splitoff from CoreMapType to ease sharing the code with legacy and dotnet.
https://github.com/xamarin/xamarin-macios/pull/11474
The actual change is made by `CoreOptimizeGeneratedCode` which is changed by this PR.


The problem is the updated `CoreOptimizeGeneratedCode` is not run in the same time/order as before.

The linker runs an optimization [1] before our code is executed. Since it see a `true` constant (from the property) then it rewrite the code to remove the condition.

[1] https://github.com/mono/linker/blob/a5f4931bc577ca94d05cd6b64e810c5cc4f035bb/src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs#L244

However we now know that this value can be changed to `false` and try to change it... but it's too late - the call to the property is still there (so we're called to process it) but the branch (for false) is not, it was replaced with a `pop`, so the instance of the `DynamicRegistrar` is marked.

spouliot (approved):
back on track :)

spouliot:
>> 4 tests failed, 103 tests passed.

> @sbomer those are unrelated to your PR and already fixed in HEAD (just rebase to get the fix)

I was wrong (same suites but different reasons or I opened the wrong log file). 

The failures are related to the PR and are still present in the last build (along new ones).

```
[FAIL] PreserveTypeWithCustomAttribute : type 
[FAIL] IntPtrSizeTest : empty body 
Tests run: 83 Passed: 81 Inconclusive: 0 Failed: 2 Ignored: 13
```

Those might have to be _adapted_ since the dotnet linker behave differently. Lack of custom preserve and how IntPtr.Size is removed.

The two "new" ones happens because the dynamic registrar is removed - but some tests are not happy (meaning something else is missing to make this work properly).

sbomer:
Thanks, I'm looking into it.

sbomer:
Ah, even better. Thanks!

spouliot:
> Do we still want to optimize the above into

Yes, the code exists and has never been an issue (only applies to our generated binding code). Even if it has a minimal impact (we can measure later) it's has zero cost, so infinite value ;-)

> interpreter

That I did not expect, nice!

spouliot:
Right now the net effect of the PR is (dotnet numbers only)

# Application Comparer

* **A** `./main-dotnet-MySingleView.app/`
* **B** `./gh11374-dotnet-MySingleView.app/`

| Directories / Files |  A  |  B  | diff |  %  |
| ------------------- | --: | --: | ---: | --: |
| ./_CodeSignature | | | | ||     CodeResources | 5,233 | 5,233 | 0 | 0.0% |
| ./ | | | | ||     archived-expanded-entitlements.xcent | 392 | 392 | 0 | 0.0% |
|     embedded.mobileprovision | 12,391 | 12,391 | 0 | 0.0% |
|     icudt.dat | 1,713,152 | 1,713,152 | 0 | 0.0% |
|     Info.plist | 1,053 | 1,067 | 14 | 1.3% |
|     MySingleView | 7,446,304 | 7,446,304 | 0 | 0.0% |
|     MySingleView.aotdata.arm64 | 792 | 792 | 0 | 0.0% |
|     MySingleView.dll | 4,608 | 4,608 | 0 | 0.0% |
|     PkgInfo | 8 | 8 | 0 | 0.0% |
|     System.Private.CoreLib.aotdata.arm64 | 670,416 | 670,416 | 0 | 0.0% |
|     System.Private.CoreLib.dll | 782,848 | 782,848 | 0 | 0.0% |
|     System.Runtime.aotdata.arm64 | 504 | 504 | 0 | 0.0% |
|     System.Runtime.dll | 5,120 | 5,120 | 0 | 0.0% |
|     Xamarin.iOS.aotdata.arm64 | 53,536 | 53,584 | 48 | 0.1% |
|     Xamarin.iOS.dll | 91,136 | 91,136 | 0 | 0.0% |
| | | | | |
| **Statistics** | | | | |
| | | | | |
| Native subtotal | 8,171,552 | 8,171,600 | 48 | 0.0% |
|     Executable | 7,446,304 | 7,446,304 | 0 | 0.0% |
|     AOT data | 725,248 | 725,296 | 48 | 0.0% |
| | | | | |
| Managed *.dll/exe | 883,712 | 883,712 | 0 | 0.0% |
| | | | | |
| **TOTAL** | 10,787,493 | 10,787,555 | 62 | 0.0% |


The order of changes (to the IL) differs so empty branches (and properties) are left around (see decompiled diff). Not enough to be seen (within padding limits) in the `.dll` [1] but visible (48 bytes) in the .`aotdata.arm64` file [2].

[1] which is not a matter since we'll (at some point) do an IL strip like it's done for legacy.
[2] if only related to `DynamicRegistrationSupported` property (and not the IL) then it should not scale with the size of the linked Xamarin.iOS.dll (91,136 versus 19,704,832 bytes)

```diff
--- dotnet-Xamarin.iOS.cs	2021-05-10 22:39:30.000000000 -0400
+++ sbomer-Xamarin.iOS.cs	2021-05-11 09:23:58.000000000 -0400
@@ -483,6 +483,8 @@
 
 		public static Arch Arch;
 
+		public static bool DynamicRegistrationSupported => false;
+
 		[MonoPInvokeCallback(typeof(gc_collect_delegate))]
 		private static void gc_collect(out IntPtr exception_gchandle)
 		{
@@ -1319,6 +1321,9 @@
 			options->Delegates->bridge_is_valuetype = GetFunctionPointer(new bridge_is_valuetype_delegate(bridge_is_valuetype));
 			options->Delegates->bridge_is_delegate = GetFunctionPointer(new bridge_is_delegate_delegate(bridge_is_delegate));
 			options->Delegates->bridge_is_class_of_type = GetFunctionPointer(new bridge_is_class_of_type_delegate(bridge_is_class_of_type));
+			if (!DynamicRegistrationSupported)
+			{
+			}
 		}
 
 		private unsafe static void Initialize(InitializationOptions* options)
@@ -1331,6 +1336,8 @@
 			}
 			if (8 != sizeof(nint))
 			{
+				_ = 8;
+				_ = 8;
 				string text2 = string.Format("Native type size mismatch between Xamarin.iOS.dll and the executing architecture. Xamarin.iOS.dll was built for {0}-bit, while the current process is {1}-bit.", "32", "64");
 				NSLog(text2);
 				throw ErrorHelper.CreateError(8010, text2);
@@ -1344,6 +1351,9 @@
 			intptr_bool_ctor_cache = new Dictionary<Type, ConstructorInfo>(TypeEqualityComparer);
 			lock_obj = new object();
 			NSObjectClass = NSObject.Initialize();
+			if (DynamicRegistrationSupported)
+			{
+			}
 			RegisterDelegates(options);
 			Class.Initialize(options);
 			InitializePlatform(options);
@@ -2911,7 +2921,7 @@
 		[DllImport("__Internal")]
 		private static extern IntPtr xamarin_get_block_descriptor();
 
-		public unsafe void SetupBlockImpl(Delegate trampoline, Delegate userDelegate, bool safe, string signature)
+		private unsafe void SetupBlockImpl(Delegate trampoline, Delegate userDelegate, bool safe, string signature)
 		{
 			isa = NSConcreteStackBlock;
 			invoke = Marshal.GetFunctionPointerForDelegate(trampoline);
@@ -3025,6 +3035,9 @@
 			BlockLiteral block = default(BlockLiteral);
 			if (signature == null)
 			{
+				if (Runtime.DynamicRegistrationSupported)
+				{
+				}
 				throw ErrorHelper.CreateError(8026, "BlockLiteral.GetBlockForDelegate with a null signature is not supported when the dynamic registrar has been linked away (delegate type: " + @delegate.GetType().FullName + ").");
 			}
 			block.SetupBlockImpl((Delegate)value, (Delegate)@delegate, safe: true, signature);
@@ -3070,6 +3083,7 @@
 			if (registrationMap != null)
 			{
 				class_to_type = new Type[registrationMap->map_count];
+				_ = Runtime.DynamicRegistrationSupported;
 			}
 		}
 
@@ -3119,7 +3133,6 @@
 
 		private static IntPtr GetClassHandle(Type type, bool throw_if_failure, out bool is_custom_type)
 		{
-			//Discarded unreachable code: IL_00ea
 			IntPtr value = IntPtr.Zero;
 			if (type.IsByRef || type.IsPointer || type.IsArray)
 			{
@@ -3149,6 +3162,7 @@
 			}
 			if (value == IntPtr.Zero)
 			{
+				_ = Runtime.DynamicRegistrationSupported;
 				if (throw_if_failure)
 				{
 					throw ErrorHelper.CreateError(8026, "Can't register the class " + type.FullName + " when the dynamic registrar has been linked away.");
@@ -3200,9 +3214,13 @@
 				{
 					return type;
 				}
+				_ = Runtime.DynamicRegistrationSupported;
 				intPtr = class_getSuperclass(intPtr);
 			}
 			while (intPtr != IntPtr.Zero);
+			if (Runtime.DynamicRegistrationSupported)
+			{
+			}
 			if (throw_on_error)
 			{
 				throw ErrorHelper.CreateError(8026, "Can't lookup the Objective-C class 0x" + klass.ToString("x") + " (" + Marshal.PtrToStringAuto(class_getName(klass)) + ") when the dynamic registrar has been linked away.");
@@ -3858,15 +3876,16 @@
 		[Export("setTitle:forState:")]
 		public virtual void SetTitle(string title, UIControlState forState)
 		{
-			//Discarded unreachable code: IL_0030
 			UIApplication.EnsureUIThread();
 			IntPtr arg = NSString.CreateNative(title);
 			if (base.IsDirectBinding)
 			{
+				_ = 8;
 				Messaging.void_objc_msgSend_IntPtr_UInt64(base.Handle, Selector.GetHandle("setTitle:forState:"), arg, (ulong)forState);
 			}
 			else
 			{
+				_ = 8;
 				Messaging.void_objc_msgSendSuper_IntPtr_UInt64(base.SuperHandle, Selector.GetHandle("setTitle:forState:"), arg, (ulong)forState);
 			}
 			NSString.ReleaseNative(arg);
@@ -3929,12 +3948,14 @@
 			[Export("bounds")]
 			get
 			{
-				//Discarded unreachable code: IL_0030
+				//Discarded unreachable code: IL_002f
 				UIApplication.EnsureUIThread();
 				if (base.IsDirectBinding)
 				{
+					_ = 8;
 					return Messaging.CGRect_objc_msgSend(base.Handle, Selector.GetHandle("bounds"));
 				}
+				_ = 8;
 				return Messaging.CGRect_objc_msgSendSuper(base.SuperHandle, Selector.GetHandle("bounds"));
 			}
 		}
@@ -3980,11 +4001,13 @@
 			[Export("bounds")]
 			get
 			{
-				//Discarded unreachable code: IL_002b
+				//Discarded unreachable code: IL_002a
 				if (base.IsDirectBinding)
 				{
+					_ = 8;
 					return Messaging.CGRect_objc_msgSend(base.Handle, Selector.GetHandle("bounds"));
 				}
+				_ = 8;
 				return Messaging.CGRect_objc_msgSendSuper(base.SuperHandle, Selector.GetHandle("bounds"));
 			}
 		}
@@ -4626,12 +4649,13 @@
 		[Export("compare:")]
 		public virtual NSComparisonResult Compare(NSString aString)
 		{
-			//Discarded unreachable code: IL_0030
 			IntPtr nonNullHandle = NativeObjectExtensions.GetNonNullHandle(aString, "aString");
 			if (base.IsDirectBinding)
 			{
+				_ = 8;
 				return (NSComparisonResult)Messaging.Int64_objc_msgSend_IntPtr(base.Handle, Selector.GetHandle("compare:"), nonNullHandle);
 			}
+			_ = 8;
 			return (NSComparisonResult)Messaging.Int64_objc_msgSendSuper_IntPtr(base.SuperHandle, Selector.GetHandle("compare:"), nonNullHandle);
 		}
 
@@ -5201,6 +5225,7 @@
 			{
 				return true;
 			}
+			_ = Runtime.DynamicRegistrationSupported;
 			return false;
 		}
 
```

spouliot:
> monotouch-test/iOS Unified 32-bits - simulator/Debug: Failed

This is new...

```
[FAIL] TestCallbackPlayAlert : TestCallbackPlayAlert 
[FAIL] TestCallbackPlaySystem : TestCallbackPlaySystem 
[FAIL] EnumValues_22351 : AppleMoveTime could not be created: System.ArgumentNullException: Value cannot be null.
[FAIL] ReturnReleaseTest : System.NullReferenceException : Object reference not set to an instance of an object 
[FAIL] TryCopyCurrentNetworkInfo : Dictionary 
Tests run: 2460 Passed: 2410 Inconclusive: 45 Failed: 5 Ignored: 342
```

Unlikely to come from this PR (but not impossible) since that project's configuration is not linked. It might be coming from merging `main` back into the PR. @rolfbjarne any failure ring a bell ? ^ (`ReturnReleaseTest` in particular)

PR #11382, opened 2021-04-29
[runtime] Return the exception from wrapper methods for exception marshalling.

rolfbjarne:
There's no general way to set a pending managed exception in CoreCLR (the
current plan is to support setting a pending managed exception for the
objc_msgSend family of functions). This means that the way we've implemented
custom wrappers that can handle Objective-C exceptions won't work, because
those wrappers currently tries to set a pending managed exception (which Mono
throws upon returning from the corresponding native wrapper function).

So rewrite this a bit: these custom wrappers now return a GCHandle with the
managed exception as an out parameter, and the calling managed code throws
that exception instead.

This also required adjusting a few API definitions to match how their wrapper
functions are defined.


spouliot (approved):
❤️ it!

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated:

* link all/Mac Catalyst/Debug: https://github.com/xamarin/maccore/issues/1083
* link sdk/Mac Catalyst/Debug: https://github.com/xamarin/maccore/issues/1083
* monotouch-test/tvOS - simulator/Debug: https://github.com/xamarin/maccore/issues/1605

PR #11400, opened 2021-04-30
[tests] Make the mtouch test project reference the actual mtouch project.

rolfbjarne:
This makes it easier to test localized strings used in mtouch, since we don't have
to replicate the build for all the resources.

This required a few changes to avoid including code in the mtouch tests that already
exists in the mtouch executable.

Also rename the mtouch test project to mtouchtests.csproj.

This way the test project can reference the actual mtouch.csproj without
causing conflicts due to having two projects with the same name.


tj-devel709 (approved):

spouliot (approved):

mandel-macaque:
The link tests failure is due to a misconfig from vseng in the DNS for the redirect of xamarin.com. It is fixed in main, so merging to get it green, yet i'm sure it will be ok.

mandel-macaque (approved):

PR #11436, opened 2021-05-04
[dotnet] rename workload to Microsoft.NET.Sdk._platform_.Manifest-6.0.100

jonathanpeppers:
Context: https://github.com/dotnet/designs/pull/188/files#diff-8fcaa29d8e6f00b34b3cb1830d93f33e75f04424780a66a3c658c7021048e74fR125
Context: https://github.com/xamarin/xamarin-android/pull/5898

The `$(PackageId)` of our workload `.nupkg` needs to be:

    Microsoft.NET.Sdk._platform_.Manifest-6.0.100

While the `$(PackageVersion)` remains the same as before.

The layout on disk will change to:

    dotnet\sdk-manifests\6.0.100\Microsoft.NET.Sdk._platform_\
        WorkloadManifest.json
        WorkloadManifest.targets

Note that `.Manifest` and `-6.0.100` are not in the folder name on disk.

At the same time, let's also update the `version` in
`WorkloadManifest.json` so it contains the proper version for our
workload. This used to not be possible because `version` was a `long`,
but it now is a `string` where we can put our version.


dotnet/Makefile
```diff
@@ -169,7 +169,7 @@ $(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call CreateNuGetTemplate,Microso
 $(foreach platform,$(DOTNET_WINDOWS_PLATFORMS),$(eval $(call CreateWindowsNuGetTemplate,Microsoft.$(platform).Windows.Sdk,$(IOS_WINDOWS_NUGET_VERSION_NO_METADATA),$($(platform)_WINDOWS_NUGET_TARGETS))))
 $(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call CreateNuGetTemplate,Microsoft.$(platform).Ref,$($(platform)_NUGET_VERSION_NO_METADATA))))
 $(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call CreateNuGetTemplate,Microsoft.$(platform).Templates,$($(platform)_NUGET_VERSION_NO_METADATA))))
-$(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call CreateNuGetTemplate,Microsoft.NET.Workload.$(platform),$($(platform)_NUGET_VERSION_NO_METADATA))))
+$(foreach platform,$(DOTNET_PLATFORMS),$(eval $(call CreateNuGetTemplate,Microsoft.NET.Sdk.$(platform),$($(platform)_NUGET_VERSION_NO_METADATA),,.Manifest-6.0.100)))
```
spouliot, line 172: I don't like spreading that `6.0.100` string in multiple places

jonathanpeppers: Is there a way to it in one place, so both MSBuild and make can use it? If not, I can put it in one place in this `Makefile`.

spouliot: Both ? Not sure... it might need to be computed twice from other variables.

Is `100` the band ? If so then it's already available in the makefiles

```
./Make.config:DOTNET6_VERSION_BAND=$(firstword $(subst -, ,$(DOTNET6_VERSION)))
```

jonathanpeppers: Yeah `$(DOTNET6_VERSION_BAND)` works, thanks. The `6.0.100` number is in one other place in an MSBuild file, but that is probably fine.

dotnet/package/microsoft.workloads.csproj
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
-    <PackageId>Microsoft.NET.Workload.$(_PlatformName)</PackageId>
+    <PackageId>Microsoft.NET.Sdk.$(_PlatformName).Manifest-6.0.100</PackageId>
```
rolfbjarne, line 3: You could do `export DOTNET6_VERSION_BAND` somewhere in `dotnet/Makefile`, and then you can use that variable here:

```suggestion
    <PackageId>Microsoft.NET.Sdk.$(_PlatformName).Manifest-$(DOTNET6_VERSION_BAND)</PackageId>
```

jonathanpeppers: I just passed it with `-p:VersionBand`, this works, thanks.

jonathanpeppers:
I manually tested the build artifacts here, and the files make it to the place I expect:

image

spouliot (approved):

rolfbjarne (approved):

mandel-macaque (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/xamarin-macios/issues/11456)

PR #11451, opened 2021-05-05
[dotnet] Fix _DetectSdkLocations from Windows

emaf:
This task ends up setting as env variable the Xamarin Sdk root directory on the Mac, but when building from Windows it was setting the Windows path, so instead we need to override it with the proper value on macOS.

This should not change the original behavior when building from macOS.


spouliot (approved):

mandel-macaque (approved):

rolfbjarne (approved):

rolfbjarne:
Test failures are unrelated:

* link all/*: https://github.com/xamarin/xamarin-macios/issues/11456
* introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): https://github.com/xamarin/maccore/issues/2420

PR #11452, opened 2021-05-05
[CI] Remove apitest from the tests executed in older macs.

mandel-macaque:
The test was merge with the xammac_tests in commit
https://github.com/xamarin/xamarin-macios/commit/93bbfe7a86abaaef369d884ad397f590ecd409af
but we did not have the tests running to know.

This should fix some of the failures we have in older macs.

### DIFF
--- a/tools/devops/automation/templates/mac/build.yml
+++ b/tools/devops/automation/templates/mac/build.yml
@@ -147,7 +147,7 @@ steps:
 
     [System.Collections.Generic.List[string]]$failures = @()
 
-    $macTest = @("dontlink", "apitest", "introspection", "linksdk", "linkall", "xammac_tests")
+    $macTest = @("dontlink", "introspection", "linksdk", "linkall", "xammac_tests")
     foreach ($t in $macTest) {
       $testName = "exec-mac-$t"
       Write-Host "Execution test $testName"


tools/devops/automation/templates/mac/build.yml
```diff
@@ -147,7 +147,7 @@ steps:
 
     [System.Collections.Generic.List[string]]$failures = @()
 
-    $macTest = @("dontlink", "apitest", "introspection", "linksdk", "linkall", "xammac_tests")
+    $macTest = @("dontlink", "introspection", "linksdk", "linkall", "xammac_tests")
```
spouliot, line 150: but `monotouch-test` was added for `Mac [dotnet]`
and we should start to run tests on Catalyst (in 10.15+)

spouliot (approved):

tools/devops/automation/templates/mac/build.yml
mandel-macaque: This are only ran Catalina and later, should we add those? We might just want to target catalina, meaning is a much bigger change in the pipeline definition.

spouliot: > https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md

10.14+ so Mojave should run `Mac [dotnet]` too... ideally

mandel-macaque: We can try #YOLO

PR #11453, opened 2021-05-05
[CI] Dump the hardware of the bot to see if we are virtual.

mandel-macaque:
Are we virtual or are we dancers?


mandel-macaque:
My quotes are getting better: https://www.youtube.com/watch?v=RIZdjT1472Y

spouliot (approved):

PR #11455, opened 2021-05-05
Changing OneLocBuild to Manual instead of Scheduled

tj-devel709:
I am changing this so I can manually push the button to get OneLocBuild set up sooner. 
Later we can revert this change!

### DIFF
--- a/tools/devops/automation/templates/build/configure.yml
+++ b/tools/devops/automation/templates/build/configure.yml
@@ -128,7 +128,7 @@ steps:
   continueOnError: true
   workingDirectory: $(Build.SourcesDirectory)\tools\devops 
 
-- ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+- ${{ if eq(variables['Build.Reason'], 'Manual') }}:
   - task: OneLocBuild@2
     condition: and(succeeded(), eq(variables.isMain, 'True'))
     continueOnError: true


mandel-macaque (changes requested):
Do not merge it, push to a branch in the project, use it for testing by manually triggering it. Other teams are using manual triggers for testing and this will result in issues. Landing this will affect @emaf and @mauroa branch net6-hr

tj-devel709:
@mandel-macaque I have a question about doing it in another branch.
Since the condition being set [here](https://github.com/xamarin/xamarin-macios/pull/11455/files#diff-f9dc94ad49e8d4951a0b26b28f042632ae835f6d324d331a245d4138fc57232dR133) on line 133 is checking the 'isMain' variable which is equal to:
```isMain: $[eq(variables['Build.SourceBranch'], 'refs/heads/main')]```

if I run this in a different branch, won't 'isMain' equate to false?

Then I was thinking I guess I could change that line that sets the value of isMain as well but
I remembered that I need this run on main for the Loc team to consume it anyways. 
So perhaps this won't be a viable option.

File: tools/devops/automation/templates/build/configure.yml [hunk 1]
```diff
@@ -128,7 +128,7 @@ steps:
   continueOnError: true
   workingDirectory: $(Build.SourcesDirectory)\tools\devops 
 
-- ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+- ${{ if eq(variables['Build.Reason'], 'Manual') }}:
   - task: OneLocBuild@2
     condition: and(succeeded(), eq(variables.isMain, 'True'))
```
mandel-macaque, line 133: Remove this condition too and will run in your own branch.

mandel-macaque:
@tj-devel709 you are correct, that condition will stop it. Remove and use a branch name of your choice. We can't block others

rolfbjarne (approved):

tj-devel709:
Closing this PR since we need to stay on main for the Loc team, but we also need to make sure we are always running the task. Moving this task to this PR: https://github.com/xamarin/xamarin-macios/pull/11472

PR #11460, opened 2021-05-06
[CI] Store the config in a json to later be shared with other pipelines.

mandel-macaque:
We are using cascade pipelines, those are triggered automatically and
get all the artifacts. We create a json file that will be used by one of
the cascaded pipelines to make educated choices.

### DIFF
--- a/tools/devops/automation/templates/build/configure.yml
+++ b/tools/devops/automation/templates/build/configure.yml
@@ -16,18 +16,24 @@ steps:
     Import-Module $Env:SYSTEM_DEFAULTWORKINGDIRECTORY/tools/devops/automation/scripts/GitHub.psm1
     Import-Module $Env:SYSTEM_DEFAULTWORKINGDIRECTORY/tools/devops/automation/scripts/VSTS.psm1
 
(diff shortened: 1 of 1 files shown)


tools/devops/automation/templates/build/configure.yml
```diff
@@ -105,13 +111,18 @@ steps:
...
-      $tags.Add("$buildSourceBranchName")
+      $tags.Add("$($configuration.BuildSourceBranchName)")
       Write-Host "##vso[task.setvariable variable=prBuild;isOutput=true]False"
     }
     # set the tags using the cmdlet
     $tagCount = $tags.Count
     Write-Host "Found '$tagsCount' tags"
     Set-BuildTags -Tags $tags.ToArray()
+    # add the tags and serialize to json
+    $configuration | Add-Member -NotePropertyName Tags -NotePropertyValue $tags
```
mandel-macaque, line 122: Dynamically adds a new member to the object.

rolfbjarne (approved):

spouliot (approved):

PR #11469, opened 2021-05-06
[release/6.0.1xx-preview4] [dotnet] Fix _DetectSdkLocations from Windows.

rolfbjarne:
This task ends up setting as env variable the Xamarin Sdk root directory on the Mac, but when building from Windows it was setting the Windows path, so instead we need to override it with the proper value on macOS.

This should not change the original behavior when building from macOS.

Backport of #11451.


emaf (approved):

spouliot:
c.c. @dalexsoto not sure if that ship has sailed yet ?

dalexsoto (approved):
We will depend on maui-check so I think it is ok to merge

PR #11471, opened 2021-05-06
[CI] Detect if we are running a vm and set the vendor.

mandel-macaque:
We are using vms for older machines, this results in certain tests that required hardware to fail, how can we skip those? If we find that there is a manufacturer that matches a known VM vendor, we set it in an env var.

Once this is landed, we can change tests to be skipped in vms. I just have 2 vendors I have tested this with, if anyone has another, we can add it.


rolfbjarne (approved):

spouliot (approved):

dalexsoto (approved):
👍 🤯

PR #11472, opened 2021-05-06
[Localization] Enabling OneLocBuild for every build

tj-devel709:
I found out that we need to run OneLocBuild in every build since we need to publish the loc artifact every build once we are linked up with the Loc pipeline. 

Right now, we will only run the OneLocBuild task on Sundays and that is the only time we will be publishing the artifacts. 
(It will try to publish the artifacts every build, but since there is no OneLocBuild task running, there are no artifacts to publish).

This just moves the condition of checking for the schedule build reason from running the task to setting inputs.

*with help from @mandel-macaque

### DIFF
--- a/tools/devops/automation/templates/build/configure.yml
+++ b/tools/devops/automation/templates/build/configure.yml
@@ -158,11 +158,10 @@ steps:
     isCreatePrSelected: eq(variables['Build.Reason'], 'Schedule')
     packageSourceAuth: patAuth
     patVariable: '$(OneLocBuild--PAT)'
-    ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
-      isAutoCompletePrSelected: false
-      prSourceBranchPrefix: 'locfiles'
-      repoType: gitHub
-      gitHubPatVariable: '$(GitHub.Token)'
+    isAutoCompletePrSelected: false
+    prSourceBranchPrefix: 'locfiles'
+    repoType: gitHub
+    gitHubPatVariable: '$(GitHub.Token)'
 
 - task: PublishBuildArtifacts@1
   condition: and(succeeded(), eq(variables.isMain, 'True'))
--- a/tools/devops/automation/templates/build/configure.yml
+++ b/tools/devops/automation/templates/build/configure.yml
@@ -156,12 +156,13 @@ steps:
     locProj: '$(Build.SourcesDirectory)\Localize\LocProject.json'
     outDir: '$(Build.ArtifactStagingDirectory)'
     isCreatePrSelected: eq(variables['Build.Reason'], 'Schedule')
-    isAutoCompletePrSelected: false
-    prSourceBranchPrefix: 'locfiles'
-    repoType: gitHub
-    gitHubPatVariable: '$(GitHub.Token)'
-    packageSourceAuth: patAuth
-    patVariable: '$(OneLocBuild--PAT)'
+    ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+      isAutoCompletePrSelected: false
+      prSourceBranchPrefix: 'locfiles'
+      repoType: gitHub
+      gitHubPatVariable: '$(GitHub.Token)'
+      packageSourceAuth: patAuth
+      patVariable: '$(OneLocBuild--PAT)'
 
 - task: PublishBuildArtifacts@1
   condition: and(succeeded(), eq(variables.isTJ, 'True'))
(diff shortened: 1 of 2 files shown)


tools/devops/automation/templates/build/configure.yml
```diff
@@ -147,16 +147,16 @@ steps:
...
+  env:
+    SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+  inputs:
+    locProj: '$(Build.SourcesDirectory)\Localize\LocProject.json'
+    outDir: '$(Build.ArtifactStagingDirectory)'
+    isCreatePrSelected: eq(variables['Build.Reason'], 'Schedule')
+    ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
       isAutoCompletePrSelected: false
       prSourceBranchPrefix: 'locfiles'
       repoType: gitHub
```
mandel-macaque, line 162: This means that the defaults are used when we are not on a 'Schedule' builds, can you make this more explicit for those of us that do not know them?
```suggestion
    ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
      isAutoCompletePrSelected: false
      prSourceBranchPrefix: 'locfiles'
      repoType: gitHub
    ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
       ## other settings go here.
```

mandel-macaque (changes requested):

tools/devops/automation/templates/build/configure.yml
tj-devel709: Good point! I looked at the sample and when we do not have isCreatePrSelected = true, we only need to define packageSourceAuth and patVariable with regards to github.

```
steps:
- task: cesve.one-loc-build.one-loc-build.OneLocBuild@2
  displayName: 'Localization Build: Localize/LocProject.json'
  inputs:
    packageSourceAuth: patAuth
    patVariable: '$(OneLocBuildPat) ```

mandel-macaque: There are always default values:

* locProj:  "Localize/LocProject.json"
* name: "outDir: "$(Build.ArtifactStagingDirectory)"
* lclSource: lclFilesInRepo"
* lclPackageId: ""
* lclPackageVersion: "highestLclPackageVersion"
* lclPackageVersionVariable: : ""
* isEnablePseudoLanguagesSelected: "false",
* isCreatePrSelected: "false",
* repoType: "azureReposGit",
* prSourceBranchPrefix:  "locfiles",
* azureReposGitPatVariable:: "",
* gitHubPatVariable:  "",
* isAutoCompletePrSelected: "true",
* isDeletePrSourceBranchSelected:  "true",
* isUseLfLineEndingsSelected: "false",
* gitHubPrMergeMethod:"merge",
* packageSourceAuth: "credentialProviderAuth",
* patVariable: "",
* lsBuildPackageId: "lsBuildXLoc",
* lsBuildXLocPackageVersion: "",
* lsBuildLegacyXLocPackageVersion:  "",
* lsBuildWarningLevel: "0",
* xLocConsoleLoggingLevel: "information",
* nugetCliVersion: "",
  
Looking at it, we are interested in 

**isCreatePrSelected** : 'Create a pull request to check-in localized files to the repo"', defaults to false, so should be false except when we are in 'Schedule' (correct in the PR
**isAutoCompletePrSelected**: "Automatically complete a pull request for localized file check-in skipping review." DEFAULT TRUE MUST BE FALSE This is where you are wrong since you added under the Schedule if it means that is true, should always be false no matter what.

Those are the two I have checked, please review all of them with the docs/src I fwd you.

tj-devel709: Thanks, I looked at this and will just assign the other inputs without the conditional since I think it makes more sense to have those values instead of the defaults! The only one that **needs** to be different is that  isCreatePrSelected I'll keep that one with the conditional!

tj-devel709:
I noticed the OneLocBuild task has failing due to the PAT in a different branch while testing this.
Connor was able to fix it!
I also added another variable to use in place of isMain in this other branch so it runs OneLocBuild similar to how this PR should run on main [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4733060&view=logs&j=d4d86769-bae0-5955-544a-0740a81efd3c&t=b076b8fd-1a89-5fdb-2a15-f02f9055d0fd&l=211).

spouliot (approved):

dalexsoto (approved):
LGTM 👍

PR #11478, opened 2021-05-07
[Localization] Using old translations for our MSBuild localization

tj-devel709:
This PR takes the old translations we used in MSBuild and adds them to our new localization flow.
This will allow us to not drop translations for any customers until we get the localization strings back from the Loc team in the new OneLocBuild localization pipeline.

This also re-enables 2 of the MSBuild localization tests that I ignored in a previous PR. 
I still cannot remove the ignores for any mtouch localization tests since we did not have those translations.
There are also 4 new languages added from OneLocBuild that do not have existing translations so I needed to keep one of the msbuild tests ignored.


dalexsoto (approved):
Nice

spouliot (approved):

mandel-macaque:
Now we are talking :)

mandel-macaque (approved):

PR #11487, opened 2021-05-10
[runtime] Implement a managed implementation of a mono_reference_queue for CoreCLR.

rolfbjarne:
Also remove the declaration of mono_gc_reference_queue_free, it's never called.

### DIFF
--- a/src/ObjCRuntime/Runtime.CoreCLR.cs
+++ b/src/ObjCRuntime/Runtime.CoreCLR.cs
@@ -12,6 +12,7 @@ using System;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
 
 using Foundation;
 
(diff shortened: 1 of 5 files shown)


spouliot (approved):

src/ObjCRuntime/Runtime.CoreCLR.cs
```diff
@@ -450,6 +451,49 @@ static object PtrToStructure (IntPtr ptr, Type type)
...
+			public ConditionalWeakTable<object, object> Table = new ConditionalWeakTable<object, object> ();
+		}
+
+		class ReferenceQueueEntry {
+			public ReferenceQueue Queue;
+			public IntPtr UserData;
+
+			~ReferenceQueueEntry ()
+			{
+				Queue.Callback (UserData);
```
Therzok, line 475: Note: The ReferenceQueue is tracking the lifetime of our entry, not the actual targeted object. So this happens _later_ than the original object.

Also, this looks like it is just one callback that could be completely done in managed, avoiding native/managed transitions.

Therzok: Also, this looks really chose to PhantomReference. Looking at this, it looks like one such queue is NSObject_Disposer, so maybe that can be reused in a fully managed impl?

rolfbjarne: > Note: The ReferenceQueue is tracking the lifetime of the entry, not the actual targeted object. So this happens _later_ than the original object.

Yes, calling the callback later for CoreCLR than for MonoVM is fine (the GC is supposed to be unpredictable anyway), as long as it's guaranteed that the callback will be called (and not too soon).



> Also, this looks like it is just one callback that could be completely done in managed, avoiding native/managed transitions (except release pinvoke)

Yeah, the callback could be ported to managed code, but it would still have to do P/Invokes:

https://github.com/xamarin/xamarin-macios/blob/a3932e8765519bb2e7ccbddf01e18bf7dc104fc9/runtime/runtime.m#L1991-L1999

so I don't see much value there (at the very least as long as the existing code is still in use for MonoVM).

Therzok: Apparently, there's already something like the managed code:
https://github.com/xamarin/xamarin-macios/blob/2b7dc07d20fc06344d7f164af8934f5e755572c2/src/ObjCRuntime/TrampolineBlockBase.cs#L26

That ends up calling into the native `xamarin_release_block_on_main_thread`.

The advantage of CoreCLR's GC is that the reference can go away when the object is out of scope and GC.Collect() is called. This will make profiling for memory issues a bit more unpredictable than is expected.

Therzok: Wouldn't registering for a GC notification then checking liveness be better?

PR #11500, opened 2021-05-11
[runtime] Implement mono_class_get_element_class for CoreCLR.

rolfbjarne:
This also meant reviewing calling code to make sure that the return value is
released when it should be.


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2304).

PR #11508, opened 2021-05-11
[Tests] Allow to assert on device and not vm on macOS.

mandel-macaque:


mandel-macaque:
reference https://github.com/xamarin/xamarin-macios/pull/11471

rolfbjarne:
> It make sense to have a `AssertNotVirtualMachine` (for the code you added) and have it do nothing on iOS (and friends).

You can have different results in the simulator depending on whether the simulator runs in a VM or not.

spouliot (approved):

mandel-macaque:
@rolfbjarne I really hope we don't get to that, but I will keep an eye. Is a very good point.

Issue #11497, opened 2021-05-11
BI1055 error is thrown for iOS Binding project after migrating to .NET6


APopatanasov:
### Steps to Reproduce

I am currently migrating our iOS Binding project to .NET6. One of the APIs I am binding is a protocol that inherits from the native UIScrollViewDelegate protocol. When I try to build the project the an error is thrown.
1.  Open the attached project.
2.  Try to build it: dotnet build Net6Binding.csproj

### Expected Behavior
The binding should be built.

### Actual Behavior
An error is thrown.

**BTOUCH : error BI1055: bgen: Internal error: failed to convert type 'System.Runtime.Versioning.SupportedOSPlatformAttribute, System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Please file a bug report (https://github.com/xamarin/xamarin-macios/issues/new) with a test case.**

### Environment
MacOS Big Sur 11.3.1
dotner version - **6.0.100-preview.4.21255.9**

### Build Logs
[log.txt](https://github.com/xamarin/xamarin-macios/files/6459207/log.txt)
### Example Project (If Possible)
[Net6Binding.zip](https://github.com/xamarin/xamarin-macios/files/6459211/Net6Binding.zip)

Issue #10210, opened 2020-12-04
[.NET 6] Implement API diff for our .NET assemblies


rolfbjarne:
Implement API diff for our .NET assemblies, like we have it for our old-style assemblies.

spouliot:
Xamarin.iOS.cs for MacCatalyst does not generate a perfect copy of what we ship with Xamarin.iOS. The only way to fix (and avoid future) breakage is to do / publish an API diff between them -> I removed the "Possibly" from above.

PR #11522, opened 2021-05-12
[dotnet][generator] Fix creating 3rd party bindings

spouliot:
Internally the generator uses `AvailabilityBaseAttribute` to make its
decisions. For `dotnet` we generated the newer `[SupportedOSPlatform]`
and `[UnsupportedOSPlatform]`.

A 3rd-party (dotnet) binding might refer to members decorated with the
newer attributes and fail the build with an error [1]. Avoiding the error
is easy (first block of the diff) but it does not make the _right_
decisions (e.g. if a member is unavailable for the platform) when
generating the code.

To fix this we need to be able to convert the new attributes into the
well know `AvailabilityBaseAttribute` subclasses. We have not spotted
such cases yet because

* the bindings pass our tests (good but extra code would not fail)
* API diff across legacy and dotnet is not yet done [2]

[1] https://github.com/xamarin/xamarin-macios/issues/11497
[2] https://github.com/xamarin/xamarin-macios/issues/10210

### DIFF
--- a/src/generator-attribute-manager.cs
+++ b/src/generator-attribute-manager.cs
@@ -182,6 +182,12 @@ public class AttributeManager
 			return typeof (Visibility);
 		case "WrapAttribute":
 			return typeof (WrapAttribute);
+#if NET
+		case "System.Runtime.Versioning.SupportedOSPlatformAttribute":
+			return typeof (System.Runtime.Versioning.SupportedOSPlatformAttribute);
+		case "System.Runtime.Versioning.UnsupportedOSPlatformAttribute":
+			return typeof (System.Runtime.Versioning.UnsupportedOSPlatformAttribute);
+#endif
 		}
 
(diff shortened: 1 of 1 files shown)


tj-devel709 (approved):

src/generator-attribute-manager.cs
```diff
@@ -281,11 +290,64 @@ static IEnumerable<System.Attribute> ConvertOldAttributes (CustomAttributeData a
...
+			else
+				return Enumerable.Empty<System.Attribute> ();
+#endif
 		default:
 			return Enumerable.Empty<System.Attribute> ();
 		}
 	}
 
+#if NET
+	static (PlatformName, Version) ParseOSPlatformAttribute (string arg)
```
dalexsoto, line 313: Nice!

dalexsoto (approved):
❤️

PR #11524, opened 2021-05-12
[Tests] Ignore tests that fail on VMs

mandel-macaque:
There are a number of tests that do not work on VMs yet our older
machines are using virtualization. Ignore those tests since we cannot
assert if they work or not.

fixes https://github.com/xamarin/maccore/issues/2438

### DIFF
--- a/tests/introspection/Mac/MacApiSelectorTest.cs
+++ b/tests/introspection/Mac/MacApiSelectorTest.cs
@@ -381,7 +381,12 @@ namespace Introspection {
 						return true;
 					}
 					break;
-
+				case "NSMenuView":
+					switch (selectorName) {
+					case "menuBarHeight":
+						return TestRuntime.IsVM; // skip on vms due to hadware problems
+					}
+					break;
 #if !XAMCORE_3_0		// These should be not be marked [Abstract] but can't fix w/o breaking change...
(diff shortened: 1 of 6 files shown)


tests/introspection/Mac/MacApiSelectorTest.cs
```diff
@@ -381,7 +381,12 @@ protected override bool Skip (Type type, string selectorName)
 						return true;
 					}
 					break;
-
+				case "NSMenuView":
+					switch (selectorName) {
```
tj-devel709, line 385: Why switch here and not just an if?

tj-devel709 (approved):

tests/introspection/Mac/MacApiSelectorTest.cs
spouliot: easier to add more items later, with a minimal diff

spouliot (approved):

dalexsoto (approved):

rolfbjarne (approved):
Have there been bugs filed for these failing tests? Or can you link to test results in the PR description?

mandel-macaque:
@rolfbjarne no, I'll add a big with them and will add it in the description.

PR #11525, opened 2021-05-12
[CI] If the api & generator is not built do not fail the build.

mandel-macaque:
If can happen that when we bump xcode we do not get an api & generator
diff. This results in a set of cascading events that make the build
fail.

We know track when the result is not 0, store it in a var and pass that
variable to all the tempaltes that need it. The pwsh script already has
an if to check if the file if present and if not, it shows a warning.

### DIFF
--- a/tools/devops/automation/templates/build/stage.yml
+++ b/tools/devops/automation/templates/build/stage.yml
@@ -143,12 +143,13 @@ jobs:
     parameters:
       devicePrefix: sim
       runTests: ${{ parameters.runTests }}
+      apiGeneratorDiffBuilt: $[ dependencies.build.outputs['apiGeneratorDiff.API_GENERATOR_BUILT'] ]
 
 - job: upload_vsts_tests
   displayName: 'Upload xml to vsts'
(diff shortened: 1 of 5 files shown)


dalexsoto (approved):

tools/devops/automation/templates/build/stage.yml
```diff
@@ -143,12 +143,13 @@ jobs:
     parameters:
       devicePrefix: sim
       runTests: ${{ parameters.runTests }}
+      apiGeneratorDiffBuilt: $[ dependencies.build.outputs['apiGeneratorDiff.API_GENERATOR_BUILT'] ]
```
mandel-macaque, line 146: YAML is a mess, but at this stage.. dependencies is not know if we use ${{ }} os we have to use the runtime $[ ] and $[ ] is never a boolean is always a string, so we cannot do a contains() because the expression is a string.. until runtime!!!

WELL DONE 

/me slow clap

rolfbjarne (approved):

PR #11529, opened 2021-05-13
Update the README with download links for d16-8 and d16-9 packages.

rolfbjarne:


mandel-macaque (approved):

mandel-macaque:
@rolfbjarne you are not going to get builds since we are not triggering when we update the README to safe bots: https://github.com/xamarin/xamarin-macios/blob/main/tools/devops/automation/build-pipeline.yml#L198

spouliot (approved):

PR #11530, opened 2021-05-13
[runtime] Refactor xamarin_get_nsnumber_converter to use string comparisons instead of mono_type_get_type.

rolfbjarne:
This makes it easier for CoreCLR. Also, at least for CoreCLR, it's unlikely to
be slower, since we'd have to compute the MONO_TYPE_* value in any
compatibility function.

### DIFF
--- a/runtime/exports.t4
+++ b/runtime/exports.t4
@@ -450,10 +450,6 @@
 			HasCoreCLRBridgeFunction = true,
 		},
 
-		new Export ("int", "mono_type_get_type",
-			"MonoType *", "type"
-		),
-
 		#endregion
 
 		#region metadata/mono-debug.h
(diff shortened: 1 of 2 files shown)


filipnavara:
On CoreCLR you can call `Type.GetTypeCode` and get value from this enum: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/CorElementType.cs
(Mono for cross-reference: https://github.com/mono/mono/blob/bdd772531d379b4e78593587d15113c37edd4a64/mono/metadata/blob.h#L12-L51). The values even come from the same specification so they should match perhaps with some exception for enum types (needs to be checked).

That would likely be faster and much more similar to what the Mono code did.

filipnavara:
Eh, my bad, the public API returns the value converted to `TypeCode` so it's no longer 1-to-1 with the `CorElementType`/`MONO_TYPE_`. It's still pretty close match though.

rolfbjarne:
> That would likely be faster and much more similar to what the Mono code did.

It would still be a transition from native to managed code.

The only way to know for sure is to test :smile: but this code is not really a hot path, so we can optimize later if we need to.

filipnavara:
You could always include the type code when creating the `MonoObject` in `GetMonoObjectImpl` in the managed code. That way you would avoid the transition. Then again, if it's not a hot path it might not matter too much.

spouliot (approved):

Issue #10940, opened 2021-03-23
CoreCLR: figure out how to handle managed exceptions through native code


rolfbjarne:
CoreCLR doesn't support unwinding through native frames, and will by default failfast/abort the process. We're getting support for catching these types of unhandled exceptions, so that we can convert them into Objective-C exceptions.

So the plan is to:

* [x]  Disallow any managed exception marshaling except "throw Objective-C exception" and "abort". The default should be "throw Objective-C exception": #11535.
* [ ] All our native code that calls into managed, should return managed exceptions as GCHandle (using the current pattern we have of having the last parameter be a `GCHandle*` parameter.
* [x] Add support for setting the pending managed exception upon return from any of our objc_msgSend override functions.
* [x] Revert https://github.com/xamarin/xamarin-macios/commit/ee5e2f0b0934be9b02a2feae4cb146aa313ca501 when complete

Ref: https://github.com/xamarin/xamarin-macios/pull/10939

rolfbjarne:
The remaining work is less important, so I'm lowering the priority.

rolfbjarne:
The remaining part is basically applying the following patch and fix all the compiler errors: https://gist.github.com/rolfbjarne/89aa4df414734652d77b46ccef33d445

rolfbjarne:
> The remaining part is basically applying the following patch and fix all the compiler errors: [gist.github.com/rolfbjarne/89aa4df414734652d77b46ccef33d445](https://gist.github.com/rolfbjarne/89aa4df414734652d77b46ccef33d445)

I've never seen any issues with not doing this, so I don't think it's necessary to do so.

PR #11535, opened 2021-05-13
[tools] Make exception marshalling the default for CoreCLR (and make it an error to choose otherwise).

rolfbjarne:
* Make 'throw Objective-C exception' the default for managed exception marshalling.
* Make 'throw managed exception' the default for Objective-C exception marshalling.
* Disallow the 'unwind through native frames' option: CoreCLR won't do it.
* Disallow the 'unwind through managed frames' option: it's the safeset
  option by far, and also matches the reverse case.
* Disallow the 'disable' option: this is also not safe, let's try to go the
  safe route with CoreCLR.

Partial fix for #10940.


spouliot (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2438).

PR #11537, opened 2021-05-13
[msbuild] Add support for smelting metal for Mac Catalyst.

rolfbjarne:
This makes monotouch-test's Metal test file compile for Mac Catalyst.


spouliot (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2438).

dalexsoto (approved):

spouliot:
Note that `[FAIL] Could not load AppKit.NSMenuView` is a different failure here in the PR (which is not run in a VM)
The "VM" PR had the issue too and @mandel-macaque is fixing this
Basically `NSMenuView` is only available on 32bits OS and should have been skipped (but the VM might have affected that)

emaf (approved):

PR #11552, opened 2021-05-13
[runtime] Implement xamarin_get_[nsnumber|nsvalue]_type for CoreCLR.

rolfbjarne:
* Remove a few unused xamarin_get_*_class functions.
* Make the remaining two (xamarin_get_[nsnumber|nsvalue]_type) return a
  MonoType* instead of MonoClass* - that makes things slightly simpler for
  CoreCLR (the MonoClass* return values from the previous functions were
  always converted to MonoType*s anyway).
* Implement the xamarin_get_[nsnumber|nsvalue]_type functions.
* Make the existing mono_get_string_class use the new (and more generic)
  xamarin_bridge_lookup_class method instead of the specific
  xamarin_bridge_get_string_class (which can now be removed).

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -108,40 +108,20 @@ xamarin_bridge_get_mono_method (MonoReflectionMethod *method)
 	return method;
 }
 
-MonoClass *
-xamarin_get_nsnumber_class ()
+MonoType *
+xamarin_get_nsnumber_type ()
 {
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
+	MonoClass *rv = xamarin_bridge_lookup_class (XamarinLookupTypes_Foundation_NSNumber);
+	LOG_CORECLR (stderr, "%s () => %p\n", __func__, rv);
+	return rv;
 }
 
-MonoClass *
-xamarin_get_nsvalue_class ()
+MonoType *
+xamarin_get_nsvalue_type ()
 {
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
-}
-
-MonoClass *
-xamarin_get_inativeobject_class ()
-{
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
-}
-
-MonoClass *
-xamarin_get_nsobject_class ()
-{
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
-}
-
-MonoClass *
-xamarin_get_nsstring_class ()
-{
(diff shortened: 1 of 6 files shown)


runtime/coreclr-bridge.m
```diff
@@ -108,40 +108,20 @@
...
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
-}
-
-MonoClass *
-xamarin_get_nsstring_class ()
+MonoType *
+xamarin_get_nsnumber_type ()
 {
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
+	MonoClass *rv = xamarin_bridge_lookup_class (XamarinLookupTypes_Foundation_NSNumber);
```
spouliot, line 114: should not the local variable be a `MonoType` ?

runtime/coreclr-bridge.m
```diff
@@ -108,40 +108,20 @@
...
+	return rv;
 }
 
-MonoClass *
-xamarin_get_runtime_class ()
+MonoType *
+xamarin_get_nsvalue_type ()
 {
-	xamarin_assertion_message ("The method %s it not implemented yet for CoreCLR", __func__);
+	MonoClass *rv = xamarin_bridge_lookup_class (XamarinLookupTypes_Foundation_NSValue);
```
spouliot, line 122: should not the local variable be a `MonoType` ?

runtime/coreclr-bridge.m
```diff
@@ -492,7 +472,7 @@
 MonoClass *
 mono_get_string_class ()
```
spouliot, line 473: inconsistent (MonoClass* vs MonoType*) with the above ones

spouliot (changes requested):
I understand the type does not matter on the CoreCLR side and the commit message explains it. However this will be confusing while reading the source in the future (without looking at the commit message). Maybe add a comment ?

dalexsoto (approved):
After Sebastien's comments are addressed 👍

runtime/coreclr-bridge.m
rolfbjarne: I've added a comment to explain this.

runtime/coreclr-bridge.m
rolfbjarne: That's because `mono_get_string_class` is used differently (changing the call sites to take a MonoType is more complex, and doesn't really make things better, which it did for the other two functions).

spouliot (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2440).

PR #11560, opened 2021-05-14
[CoreGraphics] Adjust CGPDFOperatorTable to not use a generic Action delegate.

rolfbjarne:
* CoreCLR doesn't support generic Action delegates in reverse (P/Invokes), so
  we need to find a different solution.
* The native CGPDFOperatorTable callback API is not very friendly to us,
  because we can't pass it callback-specific data, which means that the
  managed caller must conform to the FullAOT requirement of the managed
  callback (must be a static function; must have a MonoPInvokeCallback
  attribute).
* We can leverage the new function pointer syntax in C# 9 to make these
  requirements enforced by the C# compiler (unmanaged function pointer +
  UnmanagedCallersOnly attribute). The resulting API is still clunky to use,
  but I don't see any way around that.

Fixes this monotouch-test failure with CoreCLR:

    [FAIL] Tamarin : System.Runtime.InteropServices.MarshalDirectiveException : Cannot marshal 'parameter #3': Non-blittable generic types cannot be marshaled.
               at CoreGraphics.CGPDFOperatorTable.CGPDFOperatorTableSetCallback(IntPtr table, String name, Action`2 callback)
               at CoreGraphics.CGPDFOperatorTable.SetCallback(String name, Action`2 callback)
               at MonoTouchFixtures.CoreGraphics.PDFScannerTest.Tamarin() in xamarin-macios/tests/monotouch-test/CoreGraphics/PDFScannerTest.cs:line 102

Ref: https://github.com/dotnet/runtime/issues/32963

### DIFF
--- a/src/CoreGraphics/CGPDFOperatorTable.cs
+++ b/src/CoreGraphics/CGPDFOperatorTable.cs
@@ -71,10 +71,25 @@ namespace CoreGraphics {
 
 		public IntPtr Handle { get; private set; }
 
+		// We need this P/Invoke for legacy AOT scenarios (since we have public API taking a 'Action<IntPtr, IntPtr>', and with this particular native function we can't wrap the delegate)
+		// Unfortunately CoreCLR doesn't support generic Action delegates in P/Invokes: https://github.com/dotnet/runtime/issues/32963
 		[DllImport (Constants.CoreGraphicsLibrary)]
 		extern static void CGPDFOperatorTableSetCallback (/* CGPDFOperatorTableRef */ IntPtr table, /* const char */ string name, /* CGPDFOperatorCallback */ Action<IntPtr,IntPtr> callback);
 
+#if NET
+		// This signature requires C# 9 (so .NET only).
+		// The good part about this signature is that it enforces at compile time that 'callback' is callable from native code in a FullAOT scenario.
+		// The bad part is that it's unsafe code (and callers must be in unsafe mode as well).
+		[DllImport (Constants.CoreGraphicsLibrary)]
+		unsafe extern static void CGPDFOperatorTableSetCallback (/* CGPDFOperatorTableRef */ IntPtr table, /* const char */ string name, /* CGPDFOperatorCallback */ delegate* unmanaged<IntPtr, IntPtr, void> callback);
+#endif
+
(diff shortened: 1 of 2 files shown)


spouliot (approved):
👍

src/CoreGraphics/CGPDFOperatorTable.cs
```diff
@@ -71,20 +71,35 @@ protected virtual void Dispose (bool disposing)
...
+
 		// this won't work with AOT since the callback must be decorated with [MonoPInvokeCallback]
 		public void SetCallback (string name, Action<CGPDFScanner,object> callback)
 		{
 			if (name == null)
 				throw new ArgumentNullException ("name");
 
 			if (callback == null)
-				CGPDFOperatorTableSetCallback (Handle, name, null);
+				CGPDFOperatorTableSetCallback (Handle, name, (CGPDFOperatorCallback) null);
```
tj-devel709, line 100: Woah I don't recall seeing this before haha. A type cast for null! What does this do?

mandel-macaque: @tj-devel709 imaging you are a compiler and you see:
```
public void CGPDFOperatorTableSetCallback (IntPtr handel, String name, CGPDFOperatorCallback callback) {...}
public void CGPDFOperatorTableSetCallback (IntPtr handel, String name, OtherObject obj) {..}
```

Later in the code I have:
```csharp
CGPDFOperatorTableSetCallback (Handle, name, null);
```

Which of the above overloads do you call??

And if you see:
```csharp
CGPDFOperatorTableSetCallback (Handle, name, (CGPDFOperatorCallback) null);
```

AFAIK you can also do:
```csharp
CGPDFOperatorTableSetCallback (Handle, name, default(CGPDFOperatorCallback));
```

which helps the compiler too without a cast.

mandel-macaque (approved):

src/CoreGraphics/CGPDFOperatorTable.cs
tj-devel709: Ah thank you for explaining!

dalexsoto (approved):

PR #11577, opened 2021-05-17
[dotnet][linker] Eliminate code that is only required for CoreCLR on iOS and tvOS

spouliot:
CoreCLR is being added for macOS only. We can eliminate a lot of recent
code additions (about half a megabyte in the past week) by eliminate code
branches under `if (IsCoreCLR)`.

Unlike handling (non required at runtime) custom attributes the
`ILLink.Substitutions.xml` will vary quite a bit across platforms so we
use a unique file for each platform.

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -171,7 +171,7 @@ namespace ObjCRuntime {
 		internal unsafe static bool IsCoreCLR {
 			get {
 				// The linker may turn calls to this property into a constant
-				return (options->Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
+				return (options->Flags.HasFlag (InitializationFlags.IsCoreCLR));
 			}
 		}
 #endif
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -162,15 +162,6 @@ namespace ObjCRuntime {
 					return (Flags & InitializationFlags.IsSimulator) == InitializationFlags.IsSimulator;
 				}
 			}
-
-#if NET
-			// Future optimization potential: make the linker change this into a constant.
-			internal bool IsCoreCLR {
-				get {
-					return (Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
-				}
-			}
-#endif
 		}
 
 		internal static unsafe InitializationOptions* options;
@@ -178,7 +169,10 @@ namespace ObjCRuntime {
 #if NET
 		[BindingImpl (BindingImplOptions.Optimizable)]
 		internal unsafe static bool IsCoreCLR {
-			get { return options->IsCoreCLR; }
+			get {
+				// The linker may turn calls to this property into a constant
+				return (options->Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
+			}
 		}
 #endif
 
--- a/runtime/Delegates.cs.t4
+++ b/runtime/Delegates.cs.t4
@@ -98,8 +98,10 @@ namespace ObjCRuntime {
 <# foreach (var d in delegates) {
 	if (d.OnlyDynamicUsage) continue; #>
 <#
-	if (d.OnlyCoreCLR)
+	if (d.OnlyCoreCLR) {
 		Write ("#if NET\n");
+		Write ("\t\t\tif (IsCoreCLR)\n\t");
+	}
 #>
 			options->Delegates-><#= d.SimpleEntryPoint #> = GetFunctionPointer (new <#= d.SimpleEntryPoint #>_delegate (<#= d.SimpleEntryPoint #>));
 <#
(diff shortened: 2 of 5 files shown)


src/ObjCRuntime/Runtime.cs
```diff
@@ -162,23 +162,17 @@ internal unsafe struct InitializationOptions {
...
 
 		internal static unsafe InitializationOptions* options;
 
 #if NET
 		[BindingImpl (BindingImplOptions.Optimizable)]
 		internal unsafe static bool IsCoreCLR {
-			get { return options->IsCoreCLR; }
+			get {
+				// The linker may turn calls to this property into a constant
+				return (options->Flags & InitializationFlags.IsCoreCLR) == InitializationFlags.IsCoreCLR;
```
mandel-macaque, line 174: The line is evil and will make people read it wrong, I'd add a ()

mandel-macaque (approved):

src/ObjCRuntime/Runtime.cs
spouliot: it's (mostly) a copy/paste of the existing code, removing a call indirection
I can replace it with `HasFlag` [1] since it's easier/faster to read

[1] https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?view=net-5.0

rolfbjarne (commented):
I'm wondering if a different approach would be better: to conditionally pass a substitution file to the linker at build time, conditioned on the "UseMonoRuntime" property.

So, basically make the _ComputeLinkerArguments target add `--substitutions path/to/xml` where the the xml sets the IsCoreCLR property to a constant (either true or false). It might be easier to generate the xml than to have 2 different files for every platform we support.

Advantages:
* Would work the reverse way (not include MonoVM-specific code when targeting CoreCLR).
* More future-proof: it'll automatically work if we add new platforms in the future, or when we add CoreCLR support to existing platforms.

spouliot:
> I'm wondering if a different approach would be better: to conditionally pass a substitution file to the linker at build time, conditioned on the "UseMonoRuntime" property.

That sounds like a different file per feature. IOW potentially many, many files...

or generating a file for each build ? doable, not sure it's worth the complexity and perf hit (writing/reading).

>  It might be easier to generate the xml than to have 2 different files for every platform we support.

Not sure I follow the 2 files per platforms. You mean in case MonoVM and CoreCLR would both be available ? 

In such case instead of using a constant, like `false`, we would use a _feature_ in the syntax. IOW I have no plan to have more than one such xml file for each platform.

Example of what I have locally, most of the decisions are conditional - but we keep having a single file (per platform).

```xml
<linker>
  <assembly fullname="Xamarin.iOS">
    <type fullname="ObjCRuntime.Runtime">
      <method signature="System.Void RegisterEntryAssembly(System.Reflection.Assembly)" body="stub" feature="ObjCRuntime.Runtime.RunningOnDevice" featurevalue="false" />
      <method signature="System.Boolean get_IsCoreCLR()" body="stub" value="false" />
    </type>
    <type fullname="UIKit.UIApplication">
      <method signature="System.Void EnsureEventAndDelegateAreNotMismatched()" body="stub" feature="System.Diagnostics.Debugger.IsSupported" featurevalue="false" />
      <method signature="System.Void EnsureDelegateAssignIsNotOverwritingInternalDelegate()" body="stub" feature="System.Diagnostics.Debugger.IsSupported" featurevalue="false" />
    </type>
    <type fullname="UIKit.UIButton">
      <method signature="System.Void VerifyIsUIButton()" body="stub" feature="System.Diagnostics.Debugger.IsSupported" featurevalue="false" />
    </type>
  </assembly>
</linker>
```

In theory we could have a single file _for all platforms_ but that would make the file larger, both to parse and process, reducing its readability. In any case we can choose to merge them later, just like we could chose to split the one for attributes (if needed, I just did not see much platform specific attributes).

rolfbjarne:
> In such case instead of using a constant, like `false`, we would use a _feature_ in the syntax. IOW I have no plan to have more than one such xml file for each platform.

Yes, the `feature` support in the linker seems much more appropriate than what I was thinking, it looks like something like this would be possible:

```xml
<linker>
  <assembly fullname="Xamarin.Mac">
    <type fullname="ObjCRuntime.Runtime" feature="IsMonoRuntime" featurevalue="true">
      <method signature="System.Boolean get_IsCoreCLR()" body="stub" value="false" />
    </type>
    <type fullname="ObjCRuntime.Runtime" feature="IsMonoRuntime" featurevalue="false">
      <method signature="System.Boolean get_IsCoreCLR()" body="stub" value="true" />
    </type>
  </assembly>
</linker>
```

spouliot:
@rolfbjarne yup, did not try a case like this yet - but that how I thought about it.

Any other comment(s) ?

rolfbjarne (approved):

dalexsoto (approved):
Nice 👍

spouliot:
> monotouch-test/iOS Unified 64-bits - simulator/Debug (static registrar): Failed

Failure is unrelated, known test issue -> https://github.com/xamarin/maccore/issues/2040

PR #11588, opened 2021-05-17
[Runtime] Don't skip dynamic user type checking as long as dynamic registration is supported.

rolfbjarne:
When checking whether a type is a user type or not, we might have to do a
dynamic check if the dynamic registrar is available. Otherwise we may run into
a situation where the static registrar ran during the app build, but then the
app loaded additional assemblies at runtime, and those assemblies contained
user types that were registered using the dynamic registrar, so the static
registrar doesn't know about those custom types (in other words: we need to
check at runtime).

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -1606,8 +1606,8 @@ namespace ObjCRuntime {
 					if (idx >= 0)
 						return (map [idx].flags & MTTypeFlags.UserType) == MTTypeFlags.UserType;
 					// If using the partial static registrar, we need to continue
-					// If full static registrar, we can return false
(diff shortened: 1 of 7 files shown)


Therzok (approved):

spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/1605)

rolfbjarne:
Manual backport to d16-10 is in progress here: https://github.com/xamarin/xamarin-macios/commit/b01db91ddcb686ba139dc7c241ce861d90a7dcb7

PR #11594, opened 2021-05-18
[Runtime/CoreCLR] Only read/write a single byte for booleans when marshalling parameters in dynamic method calls for CoreCLR.

rolfbjarne:
Fixes this unit test:

    TestINativeObject: #b1
        Expected: 0
        But was: 140187942967200

### DIFF
--- a/src/ObjCRuntime/Runtime.CoreCLR.cs
+++ b/src/ObjCRuntime/Runtime.CoreCLR.cs
@@ -286,7 +286,12 @@ namespace ObjCRuntime {
 			if (obj == null)
 				return;
 
-			Marshal.StructureToPtr (obj, ptr, false);
+			if (obj is bool b) {
+				// Only write a single byte for bools
+				Marshal.WriteByte (ptr, b ? (byte) 1 : (byte) 0);
+			} else {
+				Marshal.StructureToPtr (obj, ptr, false);
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


src/ObjCRuntime/Runtime.CoreCLR.cs
```diff
@@ -286,7 +286,12 @@ static void StructureToPtr (object obj, IntPtr ptr)
 			if (obj == null)
 				return;
 
-			Marshal.StructureToPtr (obj, ptr, false);
+			if (obj is bool b) {
+				// Only write a single byte for bools
+				Marshal.WriteByte (ptr, b ? (byte) 1 : (byte) 0);
```
Therzok, line 291: Can't this be specialized for bool, as in, provide an overload that takes bool?

rolfbjarne: You mean add a `StructureToPtr (bool value, IntPtr ptr)` overload (in this class)?

Therzok: just checked, it won't matter: https://github.com/rolfbjarne/xamarin-macios/blob/6438efd805beb8a9f8c92d23b93e1040d14ebf36/src/ObjCRuntime/Runtime.CoreCLR.cs#L252

spouliot (approved):

PR #11599, opened 2021-05-18
[runtime] Throw a runtime exception instead of execution engine exception.

rolfbjarne:
* One less Mono Embedding method used: good for CoreCLR.
* More consistent with the rest of our code / behavior.

### DIFF
--- a/tools/mtouch/Errors.resx
+++ b/tools/mtouch/Errors.resx
@@ -2129,4 +2129,16 @@
 {2} - The name of a method.
 		</comment>
 	</data>
+
+	<data name="MX8039" xml:space="preserve">
+		<value>Unable to marshal from {0} to an Objective-C object. The managed class must either inherit from NSObject or implement INativeObject.</value>
+		<comment>
+{0} - The name of a type.
+		</comment>
+	</data>
+
+	<data name="MX8040" xml:space="preserve">
+		<value>Invalid type encoding for parameter.</value>
+	</data>
+
 </root>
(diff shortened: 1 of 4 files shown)


mandel-macaque (approved):

tools/mtouch/Errors.resx
```diff
@@ -2129,4 +2129,16 @@
...
+		<value>Unable to marshal from {0} to an Objective-C object. The managed class must either inherit from NSObject or implement INativeObject.</value>
+		<comment>
+{0} - The name of a type.
+		</comment>
+	</data>
+
+	<data name="MX8040" xml:space="preserve">
+		<value>Invalid type encoding for parameter.</value>
+	</data>
+
```
spouliot, line 2143: Is it normal with the new translation service that the original `.resx` for other language are not present ? c.c. @tj-devel709

spouliot (approved):

tools/mtouch/Errors.resx
tj-devel709: Do you mean the files like Errors.cs.resx?
If so, then these files are in tools/mtouch/TranslatedAssemblies/Errors.cs.resx as well as the other locales' resx files.
Currently they appear to be there for me, but these ones for mtouch are not translated yet!

tj-devel709: Or if you were asking if those other resx files should have these new MX8039 and MX8040 right now, 
they should not get those right away and not translated like we used to with the xlf files.
They will just be included and translated in the other Errors.__.resx files when they are updated!

dalexsoto (approved):

PR #11600, opened 2021-05-18
[d16-10] [Runtime] Don't skip dynamic user type checking as long as dynamic registration is supported.

rolfbjarne:
When checking whether a type is a user type or not, we might have to do a
dynamic check if the dynamic registrar is available. Otherwise we may run into
a situation where the static registrar ran during the app build, but then the
app loaded additional assemblies at runtime, and those assemblies contained
user types that were registered using the dynamic registrar, so the static
registrar doesn't know about those custom types (in other words: we need to
check at runtime).

Backport of #11588.

### DIFF
--- a/tests/monotouch-test/ObjCRuntime/RegistrarTest.cs
+++ b/tests/monotouch-test/ObjCRuntime/RegistrarTest.cs
@@ -2145,6 +2145,55 @@ namespace MonoTouchFixtures.ObjCRuntime {
 			}
 		}
 
+#if __MACOS__
+		[Test]
+		public void CustomUserTypeWithDynamicallyLoadedAssembly ()
+		{
+			var customTypeAssemblyPath = global::System.IO.Path.Combine (global::Xamarin.Tests.Configuration.RootPath, "tests", "test-libraries", "custom-type-assembly", ".libs", "macos", "custom-type-assembly.dll");
+			Assert.That (customTypeAssemblyPath, Does.Exist, "existence");
+
+			var size = 10;
+			var handles = new GCHandle [size];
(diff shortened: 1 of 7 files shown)


dalexsoto (approved):

mandel-macaque (approved):

Therzok (approved):

tests/monotouch-test/ObjCRuntime/RegistrarTest.cs
```diff
@@ -2145,6 +2145,55 @@ public void TestCtors ()
...
+			var thread = new Thread (() => {
+				var customTypeAssembly = global::System.Reflection.Assembly.LoadFrom (customTypeAssemblyPath);
+				var customType = customTypeAssembly.GetType ("MyCustomType");
+				for (var i = 0; i < size; i++) {
+					var obj = (NSObject) global::System.Activator.CreateInstance (customType);
+					array.Add (obj);
+					handles [i] = GCHandle.Alloc (obj, GCHandleType.Weak);
+				}
+				// Run the GC a couple of times.
+				GC.Collect ();
```
tj-devel709, line 2170: Just wondering why you run this twice?

rolfbjarne: In certain scenarios multiple GC collections have to occur for something to be collected.

And this test is testing that something _isn't_ collected.

So just run the GC more than once to be absolutely sure that the test isn't unintentionally passing because we're not running the GC enough times to run into failure scenarios.

sevoku (approved):

rolfbjarne:

PR #11601, opened 2021-05-18
[runtime] Be a bit defensive after handling Objective-C exceptions and return a consistent value.

rolfbjarne:
The value should never be used if everything is working fine, but if something
is wrong, then this value will at least get consistent behavior.

The visible result for us right now is that monotouch-test will fail with a
test failure instead of crashing.


spouliot (approved):

mandel-macaque (approved):
<3

tj-devel709 (approved):

dalexsoto (approved):

PR #11604, opened 2021-05-18
[CI] Allow to ignore tests if dotnet is enabled.

mandel-macaque:
Allow to uses constants to ignore dotnet tests without creating a new project.

### DIFF
--- a/tests/msbuild/Xamarin.MacDev.Tests/DotnetTest.cs
+++ b/tests/msbuild/Xamarin.MacDev.Tests/DotnetTest.cs
@@ -1,4 +1,3 @@
-#if ENABLE_DOTNET
 using System;
 using System.IO;
 
@@ -52,6 +51,8 @@ namespace Xamarin.iOS.Tasks {
 		[TestCase ("MyXamarinFormsApp")]
 		public void CompareBuilds (string project, int expectedErrorCount = 0)
 		{
+			Configuration.AssertDotNetAvailable ();
+
 			tfi = "Xamarin.iOS";
 			switch (project) {
 			case "MyMetalGame":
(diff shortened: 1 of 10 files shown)


dalexsoto (approved):

rolfbjarne (changes requested):
We already have code that does something similar (for other variables), it just has to get support for ENABLE_DOTNET: https://gist.github.com/rolfbjarne/8a40df637c8f09abcc0b9a5f20108111

tests/msbuild/Xamarin.MacDev.Tests/DotnetTest.cs
```diff
@@ -1,3 +1,4 @@
+#if ENABLE_DOTNET
```
rolfbjarne, line 1: Instead of conditionally compiling the file, you can do something like this in each test, which is how we usually disable tests according to build configuration:

```csharp
if (!Configuration.IncludeDotNet) {
    Assert.Ignore (".NET tests not enabled");
}
```

and also add IncludeDotNet variable to the Configuration class and load it. See `include_device` for an example: https://github.com/xamarin/xamarin-macios/blob/7c9cf9c3783d6c2f5a32cb41171e6a0f3c1f80c8/tests/common/Configuration.cs#L272

spouliot (approved):

tests/msbuild/Xamarin.MacDev.Tests/DotnetTest.cs
mandel-macaque: let me be lazy! I'll do it ;)

rolfbjarne (approved):

PR #11609, opened 2021-05-19
[CI] Fix API diff uploads since they broke when we wanted to continue on failure.

mandel-macaque:
AZP yaml variables vs parameter is hard. The template ignores the settings of the run time variables resulting in skipping the upload and we later have urls that point nowhere. Latests comment in the commit shows we fixed this.

### DIFF
--- a/tools/devops/automation/templates/build/download-artifacts.yml
+++ b/tools/devops/automation/templates/build/download-artifacts.yml
@@ -3,10 +3,6 @@ parameters:
   type: boolean
   default: true
 
-- name: apiGeneratorDiffBuilt
-  type: boolean
-  default: true
-
 steps:
 
 - checkout: self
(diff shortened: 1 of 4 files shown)


spouliot (approved):
I'm beginning to think that there should be absolutely no logic inside the yaml files.
IOW only variables and constants (no condition)
The rest should be calling scripts (shell, csx) that can be tested independently and returns `0` or fail

mandel-macaque:
@spouliot problem.. here is the logic to execute yaml steps, there is no way around that :/

spouliot:
> @spouliot problem.. here is the logic to execute yaml steps, there is no way around that :/

@mandel-macaque then the human race is doomed :( maybe AI will be able to figure it out once we're out of the way

dalexsoto (approved):

mandel-macaque:
@spouliot is so bad that we started sharpliner ;)

PR #11612, opened 2021-05-19
[runtime] Re-implement code that fetches the assembly name to use existing API implemented by CoreCLR.

rolfbjarne:
This simplifies porting to CoreCLR a bit, because the code that's being rewritten used to call `mono_assembly_name_get_name` and `mono_assembly_get_name`, and those methods have not been implemented for CoreCLR, and with this change they won't have to be implemented either (because it's not needed).

I could make this change limited to CoreCLR only, but this is just a log statement, only printed if the app was built with verbose logging, so there's more value in keeping a single code base between MonoVM and CoreCLR than keeping the MonoVM code unchanged.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -29,6 +29,12 @@ xamarin_bridge_initialize ()
 {
 }
 
+void
+xamarin_enable_new_refcount ()
+{
+	// Nothing to do here.
+}
+
 bool
 xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, const char **propertyValues)
 {
@@ -395,6 +401,14 @@ MonoException *
 xamarin_create_system_entry_point_not_found_exception (const char *entrypoint)
 {
 	MonoException *rv = xamarin_bridge_create_exception (XamarinExceptionTypes_System_EntryPointNotFoundException, entrypoint);
+	LOG_CORECLR (stderr, "%s (%p) => %p\n", __func__, entrypoint, rv);
+	return rv;
+}
+
+MonoException *
+mono_get_exception_out_of_memory ()
+{
+	MonoException *rv = xamarin_bridge_create_exception (XamarinExceptionTypes_System_OutOfMemoryException, NULL);
 	LOG_CORECLR (stderr, "%s (%p) => %p\n", __func__, entrypoint, rv);
 	return rv;
 }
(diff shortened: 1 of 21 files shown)


spouliot (commented):
what's the win here ?

It seems it just add (new and move) more code into the MonoVM build.

rolfbjarne:
> It seems it just add (new and move) more code into the MonoVM build.

The win is to not having to implement `mono_assembly_name_get_name` and `mono_assembly_get_name` for CoreCLR.

spouliot:
> The win is to not having to implement mono_assembly_name_get_name and mono_assembly_get_name for CoreCLR.

Sorry but it's not clear how adding code to both VM (MonoVM and CoreCLR), instead of having it limited to CoreCLR, avoids that :)

Most of your PR avoid new native code (for CoreCLR) by using similar managed code - but those case are specific to CoreCLR (`Runtime.CoreCLR.cs`) and not general (in `Runtime.cs`).

Can you expand your commit message with additional information ? Curious minds (at least mean) _haz_ to know :)

stephen-hawley (approved):

rolfbjarne:
> Sorry but it's not clear how adding code to both VM (MonoVM and CoreCLR), instead of having it limited to CoreCLR, avoids that :)

Because this line removes the call to those methods: https://github.com/xamarin/xamarin-macios/pull/11612/files#diff-8bbfcca4e49fcf01536b6f7c128f4dc9c9bdf33914bea0841a8875c81944f63aL977

I've updated the PR description (and I'll use that as the commit message when I merge the PR).

spouliot:
> Because this line removes the call to those methods:

Yep, I get that one :) but I still fail to see the added simplicity. 

Both would be more identical - but that come as a cost (considering it is, at least in this PR, limited to some log function).

Why not do

```c
#if !defined (CORECLR_RUNTIME)
		// the code that already exists for MonoVM
		LOG (PRODUCT ": Skipping assembly registration for %s since it's not needed (dynamic registration is not supported)", mono_assembly_name_get_name (mono_assembly_get_name (assembly)));
#else
		// the code you added
#endif
```

and keep the extra native code and the new managed method entirely on the CoreCLR side ?

rolfbjarne:
> Both would be more identical - but that come as a cost (considering it is, at least in this PR, limited to some log function).

My thinking was that that cost was worth it to keep the implementations identical, but it's not a big deal, so I changed it as you've suggested.

rolfbjarne:

PR #11614, opened 2021-05-19
[runtime] Fix declaration of xamarin_bridge_get_method_declaring_type.

rolfbjarne:
The native xamarin_bridge_get_method_declaring_type method and the
corresponding managed method (GetMethodDeclaringType) takes and returns a
MonoObject*, not a GCHandle.

Due to the wonders of void pointers, this worked just fine before - there's no
actual change to the compiled code - but the code is now more consistent and
less confusing.

### DIFF
--- a/runtime/delegates.t4
+++ b/runtime/delegates.t4
@@ -471,8 +471,8 @@
 			OnlyCoreCLR = true,
 		},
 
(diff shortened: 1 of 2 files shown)


spouliot (approved):
void all the things!

stephen-hawley (approved):

PR #11616, opened 2021-05-19
[runtime] Refactor the toggle ref code to separate MonoVM-specific code and generic logic.

rolfbjarne:
This makes it simpler to implement the toggle ref logic for CoreCLR once we can
do that.

### DIFF
--- a/runtime/monovm-bridge.m
+++ b/runtime/monovm-bridge.m
@@ -419,4 +419,64 @@ xamarin_mono_object_retain (MonoObject *mobj)
 
 #endif // DOTNET
 
+/*
+ * ToggleRef support
+ */
+// #define DEBUG_TOGGLEREF 1
+
+static void
+gc_register_toggleref (MonoObject *obj, id self, bool isCustomType)
+{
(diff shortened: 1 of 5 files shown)


runtime/monovm-bridge.m
```diff
@@ -419,4 +419,64 @@
...
+
+	PRINT ("**Registering object %p handle %p RC %d flags: %i isCustomType: %i",
+		obj,
+		handle,
+		(int) (handle ? [handle retainCount] : 0),
+		xamarin_get_nsobject_flags (obj),
+		isCustomType
+		);
+#endif
+	mono_gc_toggleref_add (obj, TRUE);
```
stephen-hawley, line 444: What's the difference between `TRUE` here and `true` a few lines further down?

stephen-hawley (approved):

runtime/monovm-bridge.m
rolfbjarne: None really, it was probably written by different people (this is code that's been moved).

mandel-macaque (approved):

spouliot (approved):

PR #11640, opened 2021-05-20
[CI] If we modified xharness, run all tests.

mandel-macaque:


tests/xharness/Jenkins/TestSelector.cs
```diff
@@ -180,6 +184,7 @@ void SelectTestsByModifiedFiles (int pullRequest)
 			SetEnabled (files, cecilPrefixes, "cecil", ref jenkins.IncludeCecil);
 			SetEnabled (files, dotnetFilenames, "dotnet", ref jenkins.IncludeDotNet);
 			SetEnabled (files, msbuildFilenames, "msbuild", ref jenkins.IncludeMSBuild);
+			SetEnabled (files, xharnessPrefix, "all", ref jenkins.IncludeAll);
```
rolfbjarne, line 187: Won't this (try to) run the device tests too?

mandel-macaque: AFAIk it won't, but we shall see in this run :)

mandel-macaque: shows it does not Screenshot 2021-05-20 at 17 32 15

The main reason is that the all is overriden by the skip-device-tests from the make. And auto selected tests are override by labels.

rolfbjarne (approved):

tj-devel709 (approved):

PR #11655, opened 2021-05-21
[Localization] Adding localize-update workflow

tj-devel709:
This PR should allow us to receive PRs in main whenever the Loc team adds new translation to our lcl files in the 'Localization' branch.

Reasoning: The Loc team needs to be able to auto-merge these lcl file changes so they will do so in the 'Localization' branch.
We will create the PRs to main from the branch they are using to push into 'Localization' which has the prefix 'lego'. 
Therefore, we will look for 'lego/*'.

### DIFF
--- a/.github/workflows/localization-update.yml
+++ b/.github/workflows/localization-update.yml
@@ -1,0 +1,24 @@
+name: Get Localization Translations
+on:
+  push:
+    branches:
+    - "lego/*"
+
+jobs:
+  pull-request:
+    name: [Open PR to main]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      name: checkout
+
+    - uses: repo-sync/pull-request@v2
+      name: pull-request
+      with:
+        destination_branch: "main"
+        pr_title: "[Localization] Pulling New Localization Translations"
+        pr_body: "Automated PR. Bring new translated changes in the lcl files for OneLocBuild to create translated resx files."
+        pr_label: "not-notes-worthy"
+        pr_milestone: "Future"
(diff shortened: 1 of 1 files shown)


rolfbjarne (approved):
I loved legos when I was a kid! And I still do 😁 

```
    🧱 
   🧱🧱  
  🧱🧱🧱   
 🧱🧱🧱🧱 
🧱🧱🧱🧱🧱 
```

.github/workflows/localization-update.yml
```diff
@@ -0,0 +1,24 @@
+name: Get Localization Translations
+on:
+  push:
+    branches:
+    - "lego/*"
+
+jobs:
+  pull-request:
+    name: [Open PR to main]
```
mandel-macaque, line 9: We might use use another action like this in the future. Can we rename the action to something more localisation specific?

mandel-macaque (approved):
Small fix, we should use a better name.

.github/workflows/localization-update.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - uses: repo-sync/pull-request@v2
+      name: pull-request
+      with:
+        destination_branch: "main"
+        pr_title: "[Localization] Pulling New Localization Translations"
+        pr_body: "Automated PR. Bring new translated changes in the lcl files for OneLocBuild to create translated resx files."
+        pr_label: "not-notes-worthy"
+        pr_milestone: "Future"
+        pr_allow_empty: false
```
dalexsoto, line 23: ❤️

dalexsoto (approved):

PR #11662, opened 2021-05-24
[Actions] [] are special in yaml we need the ''

mandel-macaque:


tj-devel709 (approved):

mandel-macaque:
No need to run the tests since the actions does not touch any.

PR #11680, opened 2021-05-26
[Localization] Test to make sure the new resx files are compiled to Resources

tj-devel709:
I randomly reflected today on [this comment](https://github.com/xamarin/xamarin-macios/pull/11599/files#r634590317) left by @spouliot last week on Rolf's PR. I looked at my reply and think it was a half truth (or half lie) by accident of course. 

As new strings are added to the english resx files (either mtouch/Errors.resx or MSBStrings.resx), I believe the corresponding translated resx files should not be updated to contain english versions of these errors. When we call these new errors, the resource manager should see that they are not available in the other languages and just return us the english string. 

However, looking back at it, I thought it was suspicious that nothing was added to the MSBStrings.designer.cs file. I **do not** think it is true that if something is added inside the designer file, that it will be included in the resources, but I **do** think that if a string is not in the designer then it is probably not included in the resources.

I created a test to make sure that all the strings in the resx files are showing up in the resources and every string in the resources are showing up in the resx files. I ran the test on a branch straight from main and it in fact turns out that these strings (and a few others) weren't in the resources as shown in the picture below.

image

I made sure to recompile both the mtouch project and the test project and the strings were added to both the designer and resources although the designer was a little finicky/inconsistent.

This does not make a huge difference right now as the mtouch strings are not yet translated, but perhaps this can help us in the future as calling these error strings would have produced an error (I believe).

### DIFF
--- a/msbuild/Xamarin.Localization.MSBuild/Xamarin.Localization.MSBuild.csproj
+++ b/msbuild/Xamarin.Localization.MSBuild/Xamarin.Localization.MSBuild.csproj
@@ -7,19 +7,17 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
   </PropertyGroup>
   <ItemGroup>
-    <EmbeddedResource Update="MSBStrings.resx">
-      <Generator>PublicResXFileCodeGenerator</Generator>
-      <LastGenOutput>MSBStrings.Designer.cs</LastGenOutput>
-    </EmbeddedResource>
-  </ItemGroup>
-  <ItemGroup>
     <Compile Update="MSBStrings.Designer.cs">
       <DependentUpon>MSBStrings.resx</DependentUpon>
     </Compile>
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="MSBStrings.resx">
+      <Generator>PublicResXFileCodeGenerator</Generator>
+      <LastGenOutput>MSBStrings.Designer.cs</LastGenOutput>
     </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
     <EmbeddedResource Include="TranslatedAssemblies/MSBStrings.cs.resx">
       <Link>MSBStrings.cs.resx</Link>
     </EmbeddedResource>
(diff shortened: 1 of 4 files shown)


rolfbjarne (approved):

msbuild/Xamarin.Localization.MSBuild/Xamarin.Localization.MSBuild.csproj
```diff
@@ -6,20 +6,18 @@
...
-    </EmbeddedResource>
-  </ItemGroup>
   <ItemGroup>
     <Compile Update="MSBStrings.Designer.cs">
       <DependentUpon>MSBStrings.resx</DependentUpon>
     </Compile>
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="MSBStrings.resx">
+      <Generator>PublicResXFileCodeGenerator</Generator>
```
rolfbjarne, line 16: I have this from a local experiment I was doing: https://gist.github.com/rolfbjarne/7987f95c6751823eb4f7edf25c88cf94 - the hope is to make it so that the  *designer.cs file is re-generated when the resx file is changed in a text editor and we rebuild from the command line. That might be nice to get for both resx files we have (mtouch.csproj and Xamarin.Localization.MSBuild.csproj).

spouliot (approved):

tools/mtouch/Errors.designer.cs
```diff
@@ -30,7 +30,7 @@ internal class Errors {
         internal static System.Resources.ResourceManager ResourceManager {
             get {
                 if (object.Equals(null, resourceMan)) {
-                    System.Resources.ResourceManager temp = new System.Resources.ResourceManager("Errors.mtouch", typeof(Errors).Assembly);
+                    System.Resources.ResourceManager temp = new System.Resources.ResourceManager("mtouch.Errors", typeof(Errors).Assembly);
```
tj-devel709, line 33: I need to figure out why this change occurred. I believe this is causing a name mismatch since mtouch, mmp, and dotnet-linker are looking for the resources named "Errors.mtouch" and not "mtouch.Errors"

PR #11692, opened 2021-05-26
[runtime] Intercept the objc_msgSend family of functions using CoreCLR's supported mechanisms.

rolfbjarne:
This allows us to re-introduce a few tests.

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -146,6 +146,12 @@ struct InitializationOptions {
 	const char *EntryAssemblyPath;
 #endif
 	struct AssemblyLocations* AssemblyLocations;
+#if DOTNET
+	void *xamarin_objc_msgsend;
+	void *xamarin_objc_msgsend_super;
+	void *xamarin_objc_msgsend_stret;
+	void *xamarin_objc_msgsend_super_stret;
+#endif
 };
 
 static struct Trampolines trampolines = {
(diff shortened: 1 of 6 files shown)


runtime/runtime.m
```diff
@@ -146,6 +146,12 @@
 	const char *EntryAssemblyPath;
 #endif
 	struct AssemblyLocations* AssemblyLocations;
+#if DOTNET
```
spouliot, line 149: why not use the same condition as below ?
> #if defined (CORECLR_RUNTIME)
> #if !defined(__aarch64__)

spouliot: maybe not the 2nd one as the struct would not be identical in managed land (at least it would be harder than it needs to be)

runtime/runtime.m
```diff
@@ -1253,6 +1259,15 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
 	options.EntryAssemblyPath = xamarin_entry_assembly_path;
 #endif
 
+#if defined (CORECLR_RUNTIME)
+	options.xamarin_objc_msgsend = (void *) xamarin_dyn_objc_msgSend;
+	options.xamarin_objc_msgsend_super = (void *) xamarin_dyn_objc_msgSendSuper;
+#if !defined(__aarch64__)
+	options.xamarin_objc_msgsend_stret = (void *) xamarin_dyn_objc_msgSend_stret;
+	options.xamarin_objc_msgsend_super_stret = (void *) xamarin_dyn_objc_msgSendSuper_stret;
```
spouliot, line 1267: _might_ want to set them to null on arm64, not needed but cleaner

spouliot (approved):
minor

runtime/runtime.m
rolfbjarne: Yes, that was exactly my thinking: this struct must match the corresponding struct in managed land, and there we use the same managed code for MonoVM and CoreCLR.

runtime/runtime.m
rolfbjarne: The options struct is zero-initialized (https://github.com/xamarin/xamarin-macios/blob/4d6ab1d6c8ae555ec819efc94fa44d2c3aa0fa5d/runtime/runtime.m#L191), so setting to NULL explicitly is redundant.

runtime/runtime.m
rolfbjarne: > Maybe add a comment ? about same struct on both VM (for dotnet)

Will do.

> I assume there's already one (not seen in the diff) about matching struct in the managed land

Correct:

https://github.com/xamarin/xamarin-macios/blob/4d6ab1d6c8ae555ec819efc94fa44d2c3aa0fa5d/runtime/runtime.m#L99

mandel-macaque (approved):

rolfbjarne:

PR #11693, opened 2021-05-26
[dotnet] Tell .NET to not generate files we don't need. Fixes #9687.

### LINKED ISSUE #9687: [.NET] Set a few MSBuild variables to make .NET stop generating unnecessary files
Variables:

* `GenerateRuntimeConfigurationFiles=false`
* `GenerateDependencyFile=false`
* `ProduceReferenceAssembly=false`

Ref:

@rolfbjarne I don't know if you are seeing lots of extra files for iOS projects, but you might consider `GenerateDependencyFile=false` by default and `ProduceReferenceAssembly=false` by default for iOS application projects.

_Originally posted by @jonathanpeppers in https://github.com/xamarin/xamarin-android/pull/5142#discussion_r492863726_

rolfbjarne:
Tell .NET to not generate files we don't need:

* Runtime configuration file (*.runtimeconfig.json).
* Dependency file (*.deps.json).
* Reference assemblies for executable projects.

Fixes https://github.com/xamarin/xamarin-macios/issues/9687.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.props
+++ b/dotnet/targets/Xamarin.Shared.Sdk.props
@@ -23,9 +23,6 @@
 		<!-- _XamarinSdkRootOnMac this should be passed to tasks that need to access the Xamarin Sdk dir on the Mac, this value will be overriden from Windows -->
 		<_XamarinSdkRootOnMac>$(_XamarinSdkRoot)</_XamarinSdkRootOnMac>
 
-		<!-- We don't need any runtime configuration files -->
-		<GenerateRuntimeConfigurationFiles Condition="'$(GenerateRuntimeConfigurationFiles)' ==''">false</GenerateRuntimeConfigurationFiles>
-
 		<!-- We don't need any dependency files -->
 		<GenerateDependencyFile Condition="'$(GenerateDependencyFile)' == ''">false</GenerateDependencyFile>
 	</PropertyGroup>
--- a/dotnet/targets/Xamarin.Shared.Sdk.props
+++ b/dotnet/targets/Xamarin.Shared.Sdk.props
@@ -22,5 +22,11 @@
 		<_XamarinSdkRoot Condition="'$(_XamarinSdkRoot)' == ''">$(_XamarinSdkRootDirectory)</_XamarinSdkRoot>
 		<!-- _XamarinSdkRootOnMac this should be passed to tasks that need to access the Xamarin Sdk dir on the Mac, this value will be overriden from Windows -->
 		<_XamarinSdkRootOnMac>$(_XamarinSdkRoot)</_XamarinSdkRootOnMac>
+
+		<!-- We don't need any runtime configuration files -->
+		<GenerateRuntimeConfigurationFiles Condition="'$(GenerateRuntimeConfigurationFiles)' ==''">false</GenerateRuntimeConfigurationFiles>
+
+		<!-- We don't need any dependency files -->
+		<GenerateDependencyFile Condition="'$(GenerateDependencyFile)' == ''">false</GenerateDependencyFile>
 	</PropertyGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -132,6 +132,9 @@
 		<StartupHookSupport Condition="'$(StartupHookSupport)' == ''">false</StartupHookSupport>
 		<UseSystemResourceKeys Condition="'$(UseSystemResourceKeys)' == ''">true</UseSystemResourceKeys>
 		<UseNativeHttpHandler Condition="'$(_PlatformName)' != 'macOS' And '$(UseNativeHttpHandler)' == ''">true</UseNativeHttpHandler>
+
+		<!-- We don't need to generate reference assemblies for apps or app extensions -->
+		<ProduceReferenceAssembly Condition="'$(ProduceReferenceAssembly)' == '' And ('$(OutputType)' == 'Exe' Or '$(IsAppExtension)' == 'true')">false</ProduceReferenceAssembly>
```
spouliot, line 137: hmm... iirc we're using `.dll` for the main assembly for net6 so that would never work (except for app extensions)

spouliot (changes requested):

dotnet/targets/Xamarin.Shared.Sdk.targets
rolfbjarne: We use `OutputType=Exe`: 

https://github.com/xamarin/xamarin-macios/blob/9d97d43f74c28fa4e0c1cc9fa70a0983f24b7ad9/tests/dotnet/MySingleView/MySingleView.csproj#L6

We also depend on this elsewhere in this exact file: 

https://github.com/xamarin/xamarin-macios/blob/2aaa7a31b2313dc7bcd50af04c85b4db6d2c4ab0/dotnet/targets/Xamarin.Shared.Sdk.targets#L15-L35

spouliot: hmm... is that recent ?

because size comparison shows `MySingleView.dll` for .net

or maybe the issue is elsewhere... `<OutputType>` not doing what seems logical ?

rolfbjarne: > hmm... is that recent ?

No, it's not recent, it's always been this way.

> 
> because size comparison shows `MySingleView.dll` for .net
> 
> or maybe the issue is elsewhere... `<OutputType>` not doing what seems logical ?

Yes, OutputType is doing something different now. I'm not quite sure what, but I believe it doesn't affect the extension for the output assembly anymore.

spouliot (approved):

mandel-macaque (approved):

dotnet/targets/Xamarin.Shared.Sdk.props
```diff
@@ -22,5 +22,11 @@
 		<_XamarinSdkRoot Condition="'$(_XamarinSdkRoot)' == ''">$(_XamarinSdkRootDirectory)</_XamarinSdkRoot>
 		<!-- _XamarinSdkRootOnMac this should be passed to tasks that need to access the Xamarin Sdk dir on the Mac, this value will be overriden from Windows -->
 		<_XamarinSdkRootOnMac>$(_XamarinSdkRoot)</_XamarinSdkRootOnMac>
+
+		<!-- We don't need any runtime configuration files -->
+		<GenerateRuntimeConfigurationFiles Condition="'$(GenerateRuntimeConfigurationFiles)' ==''">false</GenerateRuntimeConfigurationFiles>
```
filipnavara, line 27: You will actually need it at some point. There's a Mono task that converts the JSON into binary blob and that can be passed to the Mono runtime. I recently implemented it in AppleAppBuilder on the dotnet/runtime side.

filipnavara: PR (for example usage): https://github.com/dotnet/runtime/pull/53172
Tracking issue: https://github.com/dotnet/runtime/issues/49237

It's necessary for features like `EnableUnsafeUTF7Encoding` and `EnableUnsafeBinaryFormatterSerialization`.

rolfbjarne: So I guess we'll have to:

* Keep generating *.runtimeconfig.json file
* Use Mono's task to convert the json to a binary blob.
* Bundle the blob in the .app
* Don't bundle the json in the app.

Is that correct?

rolfbjarne: I've filed #11745 to track this.

filipnavara: Sounds correct to me.

rolfbjarne:
Test failure is unrelated and will be fixed with #11732.

PR #11701, opened 2021-05-26
[CI] Expressions should be between $()

mandel-macaque:


tj-devel709 (approved):

dalexsoto (approved):

spouliot (approved):
$$$$$$$$$$
^ use any and as much you need next time

Issue #11686, opened 2021-05-26
Xamarin.iOS.Common.targets: warning MSB4011, targets cannot be imported again. It was already imported


tranb3r:
New warning when building xamarin ios app with VS 2019 16.10.0:

```
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.targets(27,2): warning MSB4011: "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.Before.targets" cannot be imported again. It was already imported at "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.targets (25,2)". This is most likely a build authoring error. This subsequent import will be ignored. [D:\Dev\Zitch\secure-httpclient\SecureHttpClient.TestRunner.iOS\SecureHttpClient.TestRunner.iOS.csproj]
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.targets(1168,2): warning MSB4011: "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.After.targets" cannot be imported again. It was already imported at "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\iOS\Xamarin.iOS.Common.targets (1166,2)". This is most likely a build authoring error. This subsequent import will be ignored. [D:\Dev\Zitch\secure-httpclient\SecureHttpClient.TestRunner.iOS\SecureHttpClient.TestRunner.iOS.csproj]
```

Was not happening in 16.9.5.
Could you please take a look ?

spouliot:
There's some difference between `main` and `d16-10` but even the later is not a direct match to your line numbers.

Please provide the exact versions of all software being used

* Visual Studio: Help > About Microsoft Visual Studio > Copy Info [button]
* Visual Studio for Mac: Visual Studio > About Visual Studio > Show Details > Copy Information [button]

and paste this into the issue. Thanks!

c.c. @emaf there was code added for Windows (in `d16-10`) that seems removed in `main`. Know of anything that should be back ported ? thanks!

tranb3r:
[...]

PR #11702, opened 2021-05-27
[d16-10][msbuild] Backport part of b88c3bb031e4164ba2c35466d0d3b2856dfe8e61

spouliot:
Those are the non-net6 changes that also fixed a warning about
target files being imported twice.

Fixes https://github.com/xamarin/xamarin-macios/issues/11686

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
+++ b/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets
@@ -17,15 +17,8 @@ Copyright (C) 2013-2016 Xamarin. All rights reserved.
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<Import Project="$(MSBuildThisFileDirectory)Xamarin.iOS.Common.props" />
 
-	<PropertyGroup>
-		<!-- Naming convention that relies on the fact that there will be a Microsoft.Windows.iOS.Sdk pack with the same version that the core SDK, on Windows installations only -->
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):

mandel-macaque (approved):

rolfbjarne (approved):

emaf:
@spouliot @rolfbjarne are we shipping net6 bits from d16-10? Because these changes will break net6 support from Windows on packs shipped from this branch. On b88c3bb031e4164ba2c35466d0d3b2856dfe8e61 the imports for net6 were moved to different files (Microsoft.iOS.Windows.Sdk.props and Microsoft.iOS.Windows.Sdk.targets). 

To fix this I'd just remove the imports that are **not** using the `WindowsiOSSdkDirectory` prop (https://github.com/xamarin/xamarin-macios/blob/d16-10/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets#L25-L26 and https://github.com/xamarin/xamarin-macios/blob/d16-10/msbuild/Xamarin.iOS.Tasks.Core/Xamarin.iOS.Common.targets#L1166-L1167), because for the legacy support the replacement done over `MSBuildThisFileDirectory` won't make any difference (and that's why we end up importing the targets twice) so both props will have the same value. In that way the imports will still work on net6 too.

spouliot:
>  are we shipping net6 bits from d16-10? 

@emaf no :)

emaf (approved):
Approving since we're not shipping net6 packs from this branch :)

PR #11709, opened 2021-05-27
[Runtime] Complete the CoreCLR implementation of setting the pending managed exeption.

rolfbjarne:
* This made it possible to unignore some tests.
* Also update the TestConstrainedGenericType to cope with slightly different
  exceptions in CoreCLR.

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -1174,13 +1174,4 @@ partial class TestRuntime
 #endif
 		}
 	}
-
-#if NET
-	// If we're not Mono, then we're CoreCLR.
-	public static bool IsCoreCLR {
-		get {
-			return Type.GetType ("Mono.Runtime") is null;
-		}
-	}
-#endif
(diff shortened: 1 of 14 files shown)


tests/common/TestRuntime.cs
```diff
@@ -1174,4 +1174,13 @@ class LinkerSentinel { }
 #endif
 		}
 	}
+
+#if NET
+	// If we're not Mono, then we're CoreCLR.
+	public static bool IsCoreCLR {
+		get {
+			return Type.GetType ("Mono.Runtime") == null;
```
dalexsoto, line 1182: 😜
```suggestion
			return Type.GetType ("Mono.Runtime") is null;
```

dalexsoto (approved):
🎉🎉🎉

spouliot (approved):

mandel-macaque (approved):

tj-devel709 (approved):

PR #11737, opened 2021-05-28
[Localization] Test to make sure the new resx files are compiled to Resources (Edited)

tj-devel709:
I closed this PR https://github.com/xamarin/xamarin-macios/pull/11680 so that I could add the changes that Rolf suggested from his [diff](https://gist.github.com/rolfbjarne/7987f95c6751823eb4f7edf25c88cf94).

Rolf's changes should allow us to edit `xamarin-macios/tools/mtouch/Errors.resx` or `xamarin-macios/msbuild/Xamarin.Localization.MSBuild/MSBStrings.resx` in a code editor, run make in the respective directory, and then those changes we added should appear inside the respective designer.cs file without us needing to use the IDE.

Similar to the closed PR, this PR adds tests to make sure that everything we add to the resx files are compiled in our resources so that we don't try calling an error that is not actually there. These tests along with Rolf's changes should help solve this issue.


tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
```diff
@@ -160,5 +161,27 @@ IList<string> ReadFile (string path)
...
+
+		readonly string [] ignoreList = {
+			"ResourceManager",
+			"Culture",
+		};
+
+		[Test]
+		public void UpdatedResources ()
+		{
+			var resxPath = Path.Combine (Directory.GetCurrentDirectory (), "../../../msbuild/Xamarin.Localization.MSBuild/MSBStrings.resx");
```
rolfbjarne, line 173: This depends on the current directory, which the test shouldn't really do.

You can use this code to get somewhere relative to the root path of the repository:

```suggestion
			var resxPath = Path.Combine (Configuration.RootPath, "msbuild", "Xamarin.Localization.MSBuild", "MSBStrings.resx");
```

tests/mtouch/LocalizationTests.cs
```diff
@@ -106,6 +107,24 @@ public void AllErrorTranslation (string culture)
 			}
 			Assert.IsEmpty (errorList.ToString (), $"The following errors were not translated:");
 		}
+
+		[Test]
+		public void UpdatedResources ()
+		{
+			var resxPath = Path.Combine (Directory.GetCurrentDirectory (), "../../tools/mtouch/Errors.resx");
```
rolfbjarne, line 114: Same:

```suggestion
			var resxPath = Path.Combine (Configuration.RootPath, "tools", "mtouch", "Errors.resx");
```

rolfbjarne (approved):

tests/msbuild/Xamarin.MacDev.Tasks.Tests/TaskTests/LocalizationStringTest.cs
tj-devel709: OOh nice I like that haha thanks!

PR #11738, opened 2021-05-28
[Localization] Update Localization READMEs

tj-devel709:
The README.md file inside msbuild is now outdated since we switched Localization processes.
I also removed information about the source and mtouch localization from this README 
and added them to src and mtouch READMEs. I think this makes more sense, but I can change it back if preferred.


rolfbjarne (approved):

tj-devel709:
Failed 2 unrelated tests due to https://github.com/xamarin/maccore/issues/2447.

PR #11739, opened 2021-05-29
Fixes for .NET 6 linker

sbomer:
- Pass steps as separate --custom-step arguments
- Use LinkContext as IMetadataResolver
- Port CoreTypeMapStep to work with .NET6 linker
  - Track assemblies to work around non-public linker APIs
  - Only run for assemblies which transitively reference the product assembly
  
I was having some trouble running tests locally, so triggering CI to see what happens.


tools/dotnet-linker/Steps/CollectAssembliesStep.cs
```diff
@@ -19,6 +19,10 @@ protected override void TryProcess ()
 			var getReferencedAssemblies = Configuration.Context.GetType ().GetMethod ("GetReferencedAssemblies");
 			var assemblies = (IEnumerable<AssemblyDefinition>) getReferencedAssemblies.Invoke (Configuration.Context, new object [0]);
 			Configuration.Assemblies.AddRange (assemblies);
+			foreach (var assembly in assemblies) {
+				if (!Configuration.AssembliesByName.TryAdd (assembly.Name.Name, assembly))
+					throw new InvalidOperationException ($"Multiple assemblies with the same name: {assembly.Name.Name}");
```
spouliot, line 24: Is this not something handled by the linker itself, i.e. before it reach this code

tools/dotnet-linker/Steps/RemoveAttributesDispatcher.cs
```diff
@@ -0,0 +1,11 @@
+using Mono.Linker.Steps;
+using Xamarin.Linker;
+
+namespace Xamarin.Linker.Steps {
+	class RemoveAttributesDispatcher : SubStepsDispatcher {
```
spouliot, line 5: It might be better to name them depending on when they are called, not about what they do.

E.g. `PostSweepDispatcher`

That way the name remains meaningful (and less changes are required) if we add a new substep at that time.

The alternative is to have an incorrect/incomplete name or to make a lot more changes to add that new substep.

tools/linker/CoreTypeMapStep.cs
```diff
@@ -20,27 +20,101 @@
...
+			if (_transitivelyReferencesProduct.TryGetValue (assembly, out bool result))
+				return result;
+
+			if (Profile.IsProductAssembly (assembly)) {
+				_transitivelyReferencesProduct.Add (assembly, true);
+				return true;
+			}
+
+			foreach (var reference in assembly.MainModule.AssemblyReferences) {
+				if (!Configuration.AssembliesByName.TryGetValue (reference.Name, out AssemblyDefinition resolvedReference))
```
spouliot, line 49: That works because you enforced the (assembly) _weak_ name to be unique earlier.
Not sure it will work, unless everything was resolved (to latest version) before the linker is given the list (but then that check would not be needed)

spouliot (commented):
Looks good

tools/linker/CoreTypeMapStep.cs
```diff
@@ -97,10 +177,18 @@ void SetIsDirectBindingValue (TypeDefinition type)
 			// * https://bugzilla.xamarin.com/show_bug.cgi?id=15465
 			if (IsCIFilter (type)) {
 				isdirectbinding_value [type] = null;
+#if NET
+				var base_type = Context.Resolve (type.BaseType);
+#else
 				var base_type = type.BaseType.Resolve ();
```
rolfbjarne, line 183: Creating an extension method on LinkContext for legacy Xamarin would make it possible to avoid all this conditional code:

```csharp
public static TypeDefinition Resolve (this LinkContext context, TypeReference type)
{
    return type.Resolve ();
}

// and now the following code works for both legacy Xamarin and .NET:

var base_type = Context.Resolve (type.BaseType);
```

tools/dotnet-linker/LinkerConfiguration.cs
```diff
@@ -47,7 +47,8 @@ public class LinkerConfiguration {
 
 		// The list of assemblies is populated in CollectAssembliesStep.
 		public List<AssemblyDefinition> Assemblies = new List<AssemblyDefinition> ();
-		
+		public Dictionary<string, AssemblyDefinition> AssembliesByName = new Dictionary<string, AssemblyDefinition> ();
```
rolfbjarne, line 50: Are assembly names case-sensitive?

tools/dotnet-linker/Steps/CollectAssembliesStep.cs
sbomer: You're right, somehow I missed that we can just use GetLoadedAssembly

tools/linker/CoreTypeMapStep.cs
sbomer: The linker also expects weak assembly names to be unique. You're right that the check was unnecessary - I forgot that GetLoadedAssembly is available.

This still relies on the workaround to load assemblies up-front: https://github.com/xamarin/xamarin-macios/blob/882dde132a0b893c1e929969d7472e93ec7745c8/tools/dotnet-linker/Steps/CollectAssembliesStep.cs#L19

I thought there was a way to get rid of this by resolving assemblies ourselves, but that doesn't appear to be the case - I filed https://github.com/mono/linker/issues/2073 about it.

tools/dotnet-linker/LinkerConfiguration.cs
sbomer: They're not, good catch (removing this logic anyway)

sbomer:
Remaining work depends on dependency updates:
- https://github.com/xamarin/xamarin-macios/pull/11852 will bring in a linker update that fixes the custom step order. If that dependency update goes in before this change I can fix this TODO: https://github.com/xamarin/xamarin-macios/pull/11739/files#diff-2e24f1b6f44df7580244d00f106fa1a9e2257536fbca9d7bbdc74571549324a9R323
- Once https://github.com/mono/linker/pull/2074 flows I can get rid of the last reflection workaround at https://github.com/xamarin/xamarin-macios/blob/main/tools/dotnet-linker/Steps/CollectAssembliesStep.cs#L19 - I would prefer to do this as a separate change since I don't know how long it'll be until the next dependency update.

tools/linker/CoreTypeMapStep.cs
```diff
@@ -20,37 +21,162 @@
...
 			// bonus: we cache, for every type, whether or not it inherits from NSObject (very useful later)
 			if (!IsNSObject (type))
 				return;
-			
+
 			// if not, it's a user type, the IsDirectBinding check is required by all ancestors
 			SetIsDirectBindingValue (type);
 		}
-		
+
```
spouliot, line 179: white noise

spouliot (approved):

tools/linker/CoreTypeMapStep.cs
```diff
@@ -20,37 +21,162 @@
...
+#if NET
+		ConfigurationAwareStep
+#else
+		TypeMapStep
+#endif
+	{
+
+#if NET
+		protected override string Name { get; } = "CoreTypeMap";
+		protected override int ErrorCode { get; } = 2381;
```
rolfbjarne, line 34: This should be a multiple of 10, because the base class may increase the actual error number to identify each specific error case.

```suggestion
		protected override int ErrorCode { get; } = 2390;
```

rolfbjarne (changes requested):

spouliot:
I'm testing this locally. 
At first look it seems `IsDirectBinding` is doing the right thing :-)

sbomer:
Thanks a lot!

rolfbjarne (approved):

spouliot:
Looking at `Xamarin.iOS.dll` content there's two major differences between legacy and dotnet
* dotnet does not mark protocol correctly (something is not backported) which means several types/members are not included in the app [example](https://gist.github.com/spouliot/f932a53ba1fdfe5b9927ce73144c501e#file-pr11739-diff-L23)
* The synchronization context code is marking different methods (needs investigation) [example](https://gist.github.com/spouliot/f932a53ba1fdfe5b9927ce73144c501e#file-pr11739-diff-L393)

None are blockers (or regressions from current) so the PR can be merged (once bots complete the current build).

spouliot:

unrelated, known failure -> https://github.com/xamarin/maccore/issues/2454

PR #11749, opened 2021-05-31
[coreclr] Support for NSAutoreleasePools has now been implemented for background threads for both MonoVM and CoreCLR. Fixes #11256.

### LINKED ISSUE #11256: CoreCLR: NSAutoreleasePool for every thread

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/issues/11256.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -244,8 +244,6 @@ xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, cons
 void
 xamarin_install_nsautoreleasepool_hooks ()
 {
-	// https://github.com/xamarin/xamarin-macios/issues/11256
-	fprintf (stderr, "TODO: add support for wrapping all threads with NSAutoreleasePools.\n");
 }
 
 void
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -244,6 +244,7 @@ xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, cons
 void
 xamarin_install_nsautoreleasepool_hooks ()
 {
+	// No need to do anything here for CoreCLR.
 }
 
 void
--- a/tests/monotouch-test/ObjCRuntime/RuntimeTest.cs
+++ b/tests/monotouch-test/ObjCRuntime/RuntimeTest.cs
@@ -512,8 +512,8 @@ namespace MonoTouchFixtures.ObjCRuntime {
 
 			// Strangely enough there seems to be a race condition here, not all threads will necessarily
 			// have completed the autorelease by this point. Some should have though, so assert that the object
-			// was released on at least one thread.
-			Assert.That ((int) obj.RetainCount, Is.LessThan (10), "RC");
+			// was released on at least half the threads.
+			Assert.That ((int) obj.RetainCount, Is.LessThan (count / 2), "RC");
 
 			obj.Dispose ();
 		}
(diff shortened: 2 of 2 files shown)


tests/monotouch-test/ObjCRuntime/RuntimeTest.cs
```diff
@@ -489,6 +489,34 @@ public void NSAutoreleasePoolInThreadPool ()
...
+			}
+
+			for (var i = 0; i < count; i++) {
+				Assert.IsTrue (threads [i].Join (TimeSpan.FromSeconds (1)), $"Thread #{i}");
+			}
+
+			// Strangely enough there seems to be a race condition here, not all threads will necessarily
+			// have completed the autorelease by this point. Some should have though, so assert that the object
+			// was released on at least one thread.
+			Assert.That ((int) obj.RetainCount, Is.LessThan (10), "RC");
```
spouliot, line 516: Use `count`, which could be made a `const`, instead of `10`.

> so assert that the object was released on at least one thread.

Just by reading the `NSAutoreleasePoolInThread` I'm not convinced it shows the feature to be working.

E.g. If only done by a single thread (out of 10) then the issue might not be a race. It could be that it somehow get called just  once. 

If you use `count - 1` then we know (but reading the test, not the code) this works on multiple (not just the first) threads.

spouliot (changes requested):

runtime/coreclr-bridge.m
```diff
@@ -244,8 +244,6 @@
 void
 xamarin_install_nsautoreleasepool_hooks ()
```
dalexsoto, line 245: Why is this empty?

dalexsoto (approved):

runtime/coreclr-bridge.m
rolfbjarne: @dalexsoto Because nothing needs to happen where - there's another version of this method for MonoVM:

https://github.com/xamarin/xamarin-macios/blob/5a0600491c05e34cdc40c765ea4a18c5debab484/runtime/monovm-bridge.m#L283-L290

I've added a comment to explain this.

tests/monotouch-test/ObjCRuntime/RuntimeTest.cs
rolfbjarne: @spouliot 

> Use `count`, which could be made a `const`, instead of `10`.

Done.

> If you use `count - 1` then we know (but reading the test, not the code) this works on multiple (not just the first) threads.

I decided to be braver and used `count / 2` instead :smile:

spouliot (approved):

runtime/coreclr-bridge.m
```diff
@@ -244,8 +244,7 @@
...
 {
-	// https://github.com/xamarin/xamarin-macios/issues/11256
-	fprintf (stderr, "TODO: add support for wrapping all threads with NSAutoreleasePools.\n");
+	// No need to do anything here for CoreCLR.
```
akoeplinger, line 247: it'd be good to do the same for Mono too in .NET6 so we're consistent (and not rely on the mono profiler hook to install this)

rolfbjarne: @akoeplinger would that require any changes in Mono that we have to wait for? Or is that already implemented?

akoeplinger: @rolfbjarne it should already be implemented

rolfbjarne: > @rolfbjarne it should already be implemented

Ok, filed: https://github.com/xamarin/xamarin-macios/issues/11788

PR #11732, opened 2021-05-28
[tests] Adjust MX8029_b and MX8033 tests to cope with slightly different exception message output for CoreCLR.

rolfbjarne:
Fixes these unit tests:

    [FAIL] MX8029_b :   Message
      Expected string length 217 but was 238. Strings differ at index 149.
      Expected: "...lector: setIntArray:\n\tMethod: MonoTouchFixtures.ObjCRuntim..."
      But was:  "...lector: setIntArray:\n\tMethod: System.Void MonoTouchFixture..."
      ----------------------------------------------^
      at MonoTouchFixtures.ObjCRuntime.RuntimeTest.MX8029_b() in /Users/rolf/work/maccore/onedotnet/xamarin-macios/tests/monotouch-test/ObjCRuntime/RuntimeTest.cs:line 625

    [FAIL] MX8033 :   Message
      Expected string length 192 but was 207. Strings differ at index 129.
      Expected: "...\n\tSelector: intArray\n\tMethod: MonoTouchFixtures.ObjCRuntim..."
      But was:  "...\n\tSelector: intArray\n\tMethod: System.Int32[] MonoTouchFixt..."
      ------------------------------------------------^
      at MonoTouchFixtures.ObjCRuntime.RuntimeTest.MX8033() in /Users/rolf/work/maccore/onedotnet/xamarin-macios/tests/monotouch-test/ObjCRuntime/RuntimeTest.cs:line 644

PR #11751, opened 2021-05-31
[runtime] Release the block_wrapper_queue and xamarin_wrapper_hash dictionaries upon process exit.

rolfbjarne:
While not strictly necessary to not leak (because the process is exiting
anyway), it makes it easier to read leak reports, because these dictionaries
won't show up as leaked memory anymore.

Before:

    There were 258096 MonoObjects created, 258015 MonoObjects freed, so 81 were not freed. (dynamic registrar)
    There were 205834 MonoObjects created, 205833 MonoObjects freed, so 1 were not freed. (static registrar)

After:

    There were 258104 MonoObjects created, 258025 MonoObjects freed, so 79 were not freed. (dynamic registrar)
    There were 205834 MonoObjects created, 205834 MonoObjects freed, so no leaked MonoObjects. (static registrar)

### DIFF
--- a/runtime/monotouch-main.m
+++ b/runtime/monotouch-main.m
@@ -495,6 +495,8 @@ xamarin_main (int argc, char *argv[], enum XamarinLaunchMode launch_mode)
 
 	xamarin_mono_object_release (&assembly);
 	
+	xamarin_release_static_dictionaries ();
+
 	xamarin_bridge_shutdown ();
 
 	return rv;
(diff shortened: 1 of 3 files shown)


spouliot (approved):

dalexsoto (approved):

rolfbjarne:
Test failure is unrelated and will be fixed by #11732.

PR #11754, opened 2021-06-01
[generator] Use `is` instead of `==` when doing null checks

spouliot:
In most cases this will generate the exact same IL [1].

However it will perform _better_ (like you expect) if the `==` operator
is overloaded (even more if the operator implementation is buggy).

A last, small benefit is that avoiding `==` means the linker is more
likely to be able to remove it from the final applications.

This PR only change the generated code. IOW not the generator code,
nor the manual bindings. Those should also be updated but requires
more work.

References:
[1] https://gist.github.com/spouliot/fb05e733e1f9f403de69ab3ab6bdac2b
[2] https://www.thomasclaudiushuber.com/2020/03/19/c-why-you-should-prefer-the-is-keyword-over-the-operator/


dalexsoto (approved):
Nice!

rolfbjarne (approved):

PR #11761, opened 2021-06-01
[release/6.0.1xx-preview5] [dotnet] Fix the AOT compiler on iOS and tvOS.

rolfbjarne:
Mono changed the location of the AOT compiler, so we have to update our code accordingly.

They're now in the net6 feed with the following format:

    Microsoft.NETCore.App.Runtime.AOT.<host-rid>.Cross.<target-rid>

* Make the iOS and tvOS workloads load the AOT compiler packages from Mono.
* Install the runtime workloads for iOS and tvOS so that we have the AOT
  compilers downloaded for our local tests. It seems they're not automatically
  downloaded by 'dotnet build/dotnet restore' :/
* Adjust the .NET unit tests to build for device to catch these kinds of
  problems sooner in the future.

Ref: https://github.com/dotnet/runtime/pull/51782


spouliot (approved):

dalexsoto (approved):

dotnet/Microsoft.NET.Sdk.iOS/WorkloadManifest.targets
```diff
@@ -1,4 +1,7 @@
 <Project>
 	<Import Project="Sdk.props" Sdk="Microsoft.iOS.Sdk" Condition="'$(TargetPlatformIdentifier)' == 'iOS'" />
 	<Import Project="Sdk.props" Sdk="Microsoft.iOS.Windows.Sdk" Condition=" '$(TargetPlatformIdentifier)' == 'iOS' and $([MSBuild]::IsOSPlatform('windows'))" />
+
+	<Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.ios-arm" Condition="'$(TargetPlatformIdentifier)' == 'iOS' And '$(RuntimeIdentifier)' == 'ios-arm'" />
```
steveisok, line 5: You should be able to extend the workloads defined in the Mono and not need to have these in your targets.  

`microsoft-net-runtime-ios` and `microsoft-net-runtime-tvos` are defined in https://github.com/dotnet/runtime/blob/main/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.json.in#L33-L63

dotnet/targets/WorkloadManifest.tvOS.template.json
```diff
@@ -0,0 +1,27 @@
+{
+	"version": "@VERSION@",
+	"workloads": {
+		"microsoft-@PLATFORM_LOWERCASE@-sdk-full": {
+			"description": "Microsoft @PLATFORM@ SDK",
```
steveisok, line 5: If you add extends `microsoft-net-runtime-tvos`, your workload will pull along the tvOS runtime packs and aot compilers as part of the install

dotnet/Microsoft.NET.Sdk.iOS/WorkloadManifest.targets
rolfbjarne: Those only work if `RunAOTCompilation` is set to true:

https://github.com/dotnet/runtime/blob/7e651850f7eeb69fa5e9a291197f22a9721a48a1/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets#L13-L18

and we can't set it to true in our WorkloadManifest.targets, because we might be loaded after you, which means that users would have to set it in their csproj, which we don't really want.

rolfbjarne: In any case, I do believe the end goal would be to extend your workloads, but for P5 this solution seems to work. It's not necessarily what we'd end up doing in `main`.

Also, your workload is missing an AOT compiler for Mac Catalyst.

rolfbjarne: > Also, your workload is missing an AOT compiler for Mac Catalyst.

I filed https://github.com/dotnet/runtime/issues/53579 for this.

Issue #11504, opened 2021-05-11
monotouch-test / iOS Unified 32-bits - simulator / Debug -> multiple failures


spouliot:
### Steps to Reproduce

1. Start xharness 
2. Run "monotouch-test / iOS Unified 32-bits - simulator / Debug"

Sadly it seems the commit does not trigger this test suite to run (even if all of them should run, unless impossible - which is not the case here)

### Expected Behavior

No errors

### Actual Behavior

```
	[FAIL] TestCallbackPlayAlert :   TestCallbackPlayAlert
	[FAIL] TestCallbackPlaySystem :   TestCallbackPlaySystem
	[FAIL] EnumValues_22351 : AppleMoveTime could not be created: System.ArgumentNullException: Value cannot be null.
	[FAIL] ReturnReleaseTest : System.NullReferenceException : Object reference not set to an instance of an object
	[FAIL] TryCopyCurrentNetworkInfo :   Dictionary
Tests run: 2460 Passed: 2410 Inconclusive: 45 Failed: 5 Ignored: 342
```

### Environment

`main`

### Example

https://github.com/xamarin/xamarin-macios/pull/11374#issuecomment-837937044

PR with `run-all-tests` and `skip-device-tests`

rolfbjarne:
@mandel-macaque is there a way to determine if the test run above occurred on a VM or if it was on real hardware?

spouliot:
@rolfbjarne I could duplicate them locally (not on a VM)

rolfbjarne:
* TestCallbackPlayAlert :  Fix in #11763.
* TestCallbackPlaySystem :  Fix in #11763.
* EnumValues_22351 : Fix in #11765.
* ReturnReleaseTest : Fixed in #11742.
* TryCopyCurrentNetworkInfo :   Fix in #11764.

rolfbjarne:
This is fixed in `main` now.

PR #11764, opened 2021-06-01
[monotouch-test] It doesn't look like the TryCopyCurrentNetworkInfo test works in the iOS 11 simulator either.

rolfbjarne:
Partial fix for #11504.

### DIFF
--- a/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
+++ b/tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
@@ -79,13 +79,8 @@ namespace MonoTouchFixtures.SystemConfiguration {
 			// * Add the 'com.apple.developer.networking.wifi-info' entitlement
 			// We're not using custom entitlements when building for device, which means that we can't make this work at the moment.
 			// So just assert that we get null if running on iOS 12+.
-			if (TestRuntime.CheckXcodeVersion (10, 0)) {
-				Assert.IsNull (dict, "Dictionary");
-			} else if (TestRuntime.CheckXcodeVersion (9, 0)) {
-				Assert.IsNull (dict, "Dictionary"); // It doesn't seem to work in iOS 11 either, I don't know why.
-			} else {
-				Assert.IsNotNull (dict, "Dictionary");
-			}
+			// Also assert that we get null in all other cases as well, since we link with the iOS 13+ SDK.
+			Assert.IsNull (dict, "Dictionary");
 #endif
 		}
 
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
```diff
@@ -81,6 +81,8 @@ public void TryCopyCurrentNetworkInfo ()
 			// So just assert that we get null if running on iOS 12+.
 			if (TestRuntime.CheckXcodeVersion (10, 0)) {
 				Assert.IsNull (dict, "Dictionary");
+			} else if (TestRuntime.CheckXcodeVersion (9, 0)) {
+				Assert.IsNull (dict, "Dictionary"); // It doesn't seem to work in iOS 11 either, I don't know why.
```
spouliot, line 85: > An app linked against iOS 13 or later receives NULL.
https://developer.apple.com/documentation/systemconfiguration/1614126-cncopycurrentnetworkinfo

Since we're linking with the latest SDK then it's likely always null.

spouliot (changes requested):

tests/monotouch-test/SystemConfiguration/CaptiveNetworkTest.cs
rolfbjarne: I understood that differently: "apps linked against iOS 13+ receives NULL (and this is kind of implied:) _when executing on iOS 13+_".

I think the previous sentence explains it: "An app linked against iOS 12 or earlier receives a dictionary with pseudo-values." - the point is that old apps would get the new behavior if re-built with the iOS 13+ SDK (and executing on an iOS 13+ OS), not that apps build with iOS 13+ SDK will get the new behavior on existing OSes (which would require OS changes).

spouliot: Yes. It's not common but it's not the first setting I've seen that behave differently depending on which SDK version it's linked with :-)

* If you rebuild (then you re-link) then it's up to you to test your code for changes (and adapt for the null value).

* If you don't rebuild (pre-iOS 13 SDK) and run on iOS 13+ (like updating your phone) you'll get the pseudo-values so the app doesn't crash while accessing a null dictionary.

Since our test always relink then a single
```CSharp
Assert.IsNull (dict, "Dictionary");
```
is the only thing that should be required.

rolfbjarne: @spouliot fixed

spouliot (approved):

PR #11766, opened 2021-06-01
[monotouch-test] Exclude any *.cs file in any obj subdirectory, not only the immediate subdirectory.

rolfbjarne:
This means that we won't pick up generated *.cs files from the
monotouch-test/dotnet/macOS/monotouch-test.csproj project when building the
monotouch-test/monotouch-test.csproj project, which results in the C# compiler
complaining about duplicate types, or unknown types (because the generated
code is for .NET 6, while we might be compiling for legacy Xamarin).


spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2450).

PR #11772, opened 2021-06-01
[Actions] Need to debug where is the action adding the static page.

mandel-macaque:


.github/workflows/github-page-clean-up.yml
```diff
@@ -20,11 +20,20 @@ jobs:
...
+        ls -R $WORKSPACE
+      name: 'Debug checkouts'
+      env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+        WORKSPACE: ${{ github.workspace }}
+
     - uses: actions/checkout@master
       with:
         name: xamarin/macios.ci
-        github_token: ${{ secrets.GITHUB_TOKEN }}
```
mandel-macaque, line 26: We were getting a warning, i read docs that were to old :/

dalexsoto (approved):

PR #11787, opened 2021-06-02
[release/6.0.1xx-preview5] [dotnet] Rework how we handle workloads.

rolfbjarne:
* Stop installing workloads from builds/, and install from dotnet/ instead.

* Set DOTNETSDK_WORKLOAD_MANIFEST_ROOTS to where our local workloads are to
  use them when installing our locally built workloads into .NET (the local
  one).

* Move the workloads into a common subdirectory (this is necessary for
  DOTNETSDK_WORKLOAD_MANIFEST_ROOTS to work correctly - also there can't be
  anything other than workloads in this directory for
  DOTNETSDK_WORKLOAD_MANIFEST_ROOTS to work properly).

* Delete the watchOS workload, we don't need it for the moment.

Hopefully this fixes problems installing the right workload locally.


spouliot (approved):

rolfbjarne:

PR #11793, opened 2021-06-03
[msbuild] Fixes remote tasks execution

emaf:
When building from Windows we stopped including the `SessionId` value when executing tasks remotely, because we started using that value to decide whether we should execute a task locally or remotely. The problem is the SessionId is also used from some of the base tasks (like ACToolTaskBase) to execute different code paths depending if the build is being executed from Windows or macOS (regardless where the task is currently being executed). Most of those code paths are related to calculating relative paths, and we started getting reports of weird behaviors like missing icons when deploying apps or even Apple Store submission errors.

To fix this problem we're back propagating the SessionId value when executing tasks remotely, and to decide if a task should be executed remotely or not we also check the Environment.OSVersion.Platform value (because SessionId will have a value on macOS too when executing the task remotely), so we only try to execute tasks remotely if the task is being executed from Windows and SessionId is not empty.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XamarinTask.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XamarinTask.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 
 using Microsoft.Build.Framework;
+using Microsoft.Build.Tasks;
 using Microsoft.Build.Utilities;
 
 using Xamarin.Localization.MSBuild;
@@ -128,7 +129,7 @@ namespace Xamarin.MacDev.Tasks {
 			return rv;
 		}
 
-		public bool ShouldExecuteRemotely () => Environment.OSVersion.Platform == PlatformID.Win32NT && !string.IsNullOrEmpty(SessionId);
+		public bool ShouldExecuteRemotely () => this.ShouldExecuteRemotely (SessionId);
 	}
 }
 
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XamarinTask.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XamarinTask.cs
@@ -127,6 +127,8 @@ namespace Xamarin.MacDev.Tasks {
 
 			return rv;
 		}
+
+		public bool ShouldExecuteRemotely () => Environment.OSVersion.Platform == PlatformID.Win32NT && !string.IsNullOrEmpty(SessionId);
 	}
 }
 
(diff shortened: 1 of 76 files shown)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/XamarinTask.cs
```diff
@@ -127,6 +127,8 @@ protected async System.Threading.Tasks.Task<Execution> ExecuteAsync (string file
 
 			return rv;
 		}
+
+		public bool ShouldExecuteRemotely () => Environment.OSVersion.Platform == PlatformID.Win32NT && !string.IsNullOrEmpty(SessionId);
```
rolfbjarne, line 131: Minor code style:

```suggestion
		public bool ShouldExecuteRemotely () => Environment.OSVersion.Platform == PlatformID.Win32NT && !string.IsNullOrEmpty (SessionId);
```

msbuild/Xamarin.MacDev.Tasks/MsBuildTasks/Copy.cs
```diff
@@ -1,12 +1,13 @@
...
+using Xamarin.Messaging.Build.Client;
 
 namespace Microsoft.Build.Tasks
 {
 	public class Copy : CopyBase
 	{
 		public override bool Execute ()
 		{
-			if (string.IsNullOrEmpty (SessionId))
+			if (Environment.OSVersion.Platform != PlatformID.Win32NT || string.IsNullOrEmpty (SessionId))
```
rolfbjarne, line 10: I liked the `ShouldExecuteRemotely` method, but this looks like code duplication in _a lot_ of places.

Maybe add an extension method on `Task`:

```csharp
public static bool ShouldExecuteRemotely (this Task task, string sessionId)
{
    return Environment.OSVersion.Platform != PlatformID.Win32NT || string.IsNullOrEmpty (sessionId);
}
```

and then this could become:

```suggestion
			if (ShouldExecuteRemotely (SessionId))
```

which looks much better.

rolfbjarne (changes requested):

dalexsoto (approved):
For 16-10 I guess we can let it slide since we are running against the clock but for main I would love to have Rolf’s comments addressed and we can backport them later to d16-10

msbuild/Xamarin.MacDev.Tasks/MsBuildTasks/Copy.cs
emaf: Done :)

emaf:
I'll send the changes to **d16-10** too, because I think we still have some time for the next SR

rolfbjarne (approved):

rolfbjarne:
I believe we have multiple issues about this, those should all be referenced and closed when this is merged (CC @spouliot).

spouliot (approved):

PR #11795, opened 2021-06-03
[editorconfig] Set the charset to utf-8.

rolfbjarne:
An additional point here is that the 'utf-8' charset is specifically defined
to not include the byte-order-mark.

Ref: https://editorconfig.org


spouliot (approved):

mandel-macaque (approved):
I hate BOM

PR #11809, opened 2021-06-04
Use `CFString` instead of `NSString` to get a (.net) `System.String` from an handle

spouliot:
This is possible because both types are toll-free bridged [0].

How ?

* Renamed `CFString.FetchString` to `FromHandle` so it match `NSString` API
* Make it `public` so it can be used for generated/3rd party bindings
* Update generator to use the new API (instead of the `NSString` version)
* Update manual bindings to use the new API

Why ?

It's no secret that p/invoke (C) are faster than calling a selector
(ObjC). In most cases we do not have a choice what to call... but in a
few, but commonly used, cases we can pick the fastest call.

```
// * Summary *

BenchmarkDotNet=v0.12.1.1528-nightly, OS=macOS Big Sur 11.3.1 (20E241) [Darwin 20.4.0]
Intel Core i7-4790K CPU 4.00GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
  [Host] : Mono 6.12.0 (2020-02/c633fe92383), X64

Job=InProcess  Toolchain=InProcessEmitToolchain  InvocationCount=1
IterationCount=3  LaunchCount=1  WarmupCount=3

|                Type |                              Method | UnrollFactor | ArraySize |            Mean |            Error |          StdDev |          Median |
|-------------------- |------------------------------------ |------------- |---------- |----------------:|-----------------:|----------------:|----------------:|
|      TollFreeString |               ReturnString_NSString |           16 |         ? |      2,522.0 ns |        878.81 ns |        48.17 ns |      2,541.7 ns |
|      TollFreeString |               ReturnString_CFString |           16 |         ? |        808.1 ns |          7.48 ns |         0.41 ns |        808.2 ns |
```

Next ?

The next step is to reduce other `NSString` selectors usage,
then `NSArray`...

[0] https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Toll-FreeBridgin/Toll-FreeBridgin.html

### DIFF
--- a/src/AppKit/NSAccessibility.cs
+++ b/src/AppKit/NSAccessibility.cs
@@ -11,6 +11,7 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 
+using CoreFoundation;
 using CoreGraphics;
 using Foundation;
 using ObjCRuntime;
@@ -143,7 +144,7 @@ namespace AppKit
 				subroleHandle = subrole.Handle;
 
 			IntPtr handle = NSAccessibilityRoleDescription (role.Handle, subroleHandle);
-			return NSString.FromHandle (handle);
+			return CFString.FromHandle (handle);
 		}
 
 		[DllImport (Constants.AppKitLibrary)]
@@ -155,7 +156,7 @@ namespace AppKit
 				throw new ArgumentNullException ("element");
 
 			IntPtr handle = NSAccessibilityRoleDescriptionForUIElement (element.Handle);
-			return NSString.FromHandle (handle);
+			return CFString.FromHandle (handle);
 		}
 
 		[DllImport (Constants.AppKitLibrary)]
@@ -167,7 +168,7 @@ namespace AppKit
 				throw new ArgumentNullException ("action");
 
 			IntPtr handle = NSAccessibilityActionDescription (action.Handle);
-			return NSString.FromHandle (handle);
+			return CFString.FromHandle (handle);
 		}
 
 		[DllImport (Constants.AppKitLibrary)]
--- a/tests/perftest/TollFreeBridge.cs
+++ b/tests/perftest/TollFreeBridge.cs
@@ -1,0 +1,63 @@
+using System;
+
+using CoreFoundation;
+using Foundation;
+using ObjCRuntime;
+
+using BenchmarkDotNet.Attributes;
+
+using Bindings.Test;
+
+namespace PerfTest {
+
+	[Register ("StringClass")]
+	class StringClass : NSObject {
+			public override string Description => "constant";
+	}
+
+	public class TollFreeString {
+		IntPtr StringClassClassHandle = Class.GetHandle (typeof (StringClass));
+
+		string s;
+
+		StringClass sc;
+		IntPtr description;
+
+		[GlobalSetup]
+		public void ReturnString_Setup ()
+		{
+			sc = new StringClass ();
+			description = Messaging.IntPtr_objc_msgSend (sc.Handle, Selector.GetHandle ("description"));
+			Messaging.void_objc_msgSend (description, Selector.GetHandle ("retain"));
+		}
+
+		[GlobalCleanup]
+		public void ReturnString_Cleanup ()
+		{
+			Messaging.void_objc_msgSend (description, Selector.GetHandle ("release"));
+			sc.Dispose ();
+		}
+
+		/*
+		 * Measure time required to get a string - using NSString (ObjC) selector-based API
+		 */
+
+		[Benchmark]
+		public string ReturnString_NSString ()
+		{
+			var d = NSString.FromHandle (description);
+			return s = d;
+		}
+
+		/*
+		 * Measure time required to get a string - using CFString (C) p/invoke-based API
+		 */
+
+		[Benchmark]
+		public string ReturnString_CFString ()
+		{
+			var d = CFString.FromHandle (description);
+			return s = d;
+		}
+    }
+}
--- a/tests/perftest/TollFreeBridge.cs
+++ b/tests/perftest/TollFreeBridge.cs
@@ -12,7 +12,7 @@ namespace PerfTest {
 
 	[Register ("StringClass")]
 	class StringClass : NSObject {
-			public override string Description => "constant";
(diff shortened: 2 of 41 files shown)


dalexsoto (approved):
Ahh love it ❤️

src/AppKit/NSAccessibility.cs
```diff
@@ -143,7 +144,7 @@ public static string GetRoleDescription (NSString role, NSString subrole)
 				subroleHandle = subrole.Handle;
 
 			IntPtr handle = NSAccessibilityRoleDescription (role.Handle, subroleHandle);
-			return NSString.FromHandle (handle);
+			return CFString.FromHandle (handle);
```
rolfbjarne, line 147: Wild idea: if NSString and CFString are toll-free bridged, then can't we just change the implementation of `NSString.FromHandle` to call `CFString.FromHandle`?

tests/perftest/TollFreeBridge.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using BenchmarkDotNet.Attributes;
+
+using Bindings.Test;
+
+namespace PerfTest {
+
+	[Register ("StringClass")]
+	class StringClass : NSObject {
+			public override string Description => "constant";
```
rolfbjarne, line 15: Indentation is off.

tests/perftest/TollFreeBridge.cs
```diff
@@ -0,0 +1,63 @@
...
+		string s;
+
+		StringClass sc;
+		IntPtr description;
+
+		[GlobalSetup]
+		public void ReturnString_Setup ()
+		{
+			sc = new StringClass ();
+			description = Messaging.IntPtr_objc_msgSend (sc.Handle, Selector.GetHandle ("description"));
```
rolfbjarne, line 30: This seems a bit more complicated than necessary, can't you just create an NSString?

```csharp
public class TollFreeString {
    NSString text;
    IntPtr text_handle;

    [GlobalSetup]
    public void ReturnString_Setup ()
    {
        text = new NSString ("constant");
        text_handle = text.Handle;
    }

    [Benchmark]
    public string ReturnString_NSString ()
    {
        var d = NSString.FromHandle (text_handle);
        return s = d;
    }

    ...
```

tests/perftest/TollFreeBridge.cs
rolfbjarne, line 15: Also it might be interesting to test other strings than "constant", especially ones containing non-ascii characters.

rolfbjarne (approved):
❤️

src/AppKit/NSAccessibility.cs
spouliot: Yeah, I thought about that one as it would have required less code changes.
Somehow it feels like cheating... because it would not do what you would expect (even if the result is the same).
Maybe we could add an `[Obsolete]` on `NSString.FromHandle` ?

tests/perftest/TollFreeBridge.cs
spouliot: > can't you just create an NSString?

Yes, that would have achieved the same in this case.
But this is a subset of the tests I have, the rest is just not ready yet, and it should make more sense in a few additional commits.

tests/perftest/TollFreeBridge.cs
spouliot: Good point. Maybe I should re-write the test, like you suggested, to make that easier...

src/AppKit/NSAccessibility.cs
dalexsoto: Yeah, I agree with Sebastien, having `CFString.FromHandle` is cleaner and clear on the intention, not sure about the obsolete, maybe better adding an Advice?

rolfbjarne: > Somehow it feels like cheating...

That's why I called it a wild idea, it feels wrong to me too :smile:

> Maybe we could add an `[Obsolete]` on `NSString.FromHandle` ?
> maybe better adding an Advice?

I don't lean particularly in any way, I think either would work just fine.

spouliot: @dalexsoto `[Advice]` might not show up (without additional work in both IDE) on .net6

However `[Obsolete]` has been enhanced so you can now give them error codes (to ignore warnings or make them errors) and you can supply an URL for more information. That seems to cover (and more) what we wanted to `[Advice]`

spouliot: done (adding `[Obsolete]`)

tests/perftest/TollFreeBridge.cs
spouliot: Due to the high probably of conflicts (with next week WWDC) I'll merge this without additional tests. I'll cover that in the next (related) PR

mandel-macaque (approved):

spouliot:
failures all seems to be unrelated -> https://github.com/xamarin/maccore/issues/2450

spouliot:
failure is unrelated to PR -> https://github.com/xamarin/maccore/issues/2451

src/AppKit/NSAccessibility.cs
Therzok: > [Advice] might not show up (without additional work in both IDE) on .net6

The analyzers can be made into non-IDE analyzers

Therzok:
CFString doesn't go through the Runtime NSObject cache, which might be a big part of this optimization.

src/AppKit/NSAccessibility.cs
spouliot: > The analyzers can be made

Yes, but key words are: _without additional work_. We already have a bunch of analyzers that requires more work for net6.

> nto non-IDE analyzers

The goal of `[Advice]` was to give IDE-only hints to developers.

spouliot:
> CFString doesn't go through the Runtime NSObject cache, 

True, but...

> which might be a big part of this optimization.

False. Please read the `NSString.FromHandle` code (where the `[Obsolete]` was added). You'll see that no temporary managed `NSString` instance is created, which means `Runtime.GetNSObject` is never called... IOW it does not play any part of this optimization.

Issue #11827, opened 2021-06-04
iOS app crashes with System.NullArgumentException when calling MPNowPlayingInfoCenter.DefaultCenter.NowPlaying - iOS 10


adams-family:
### Steps to Reproduce

After upgrading Xamarin.iOS from 14.8.0.3 to 14.20.0.1 the following code started to cause an iOS app to crash on iOS 10.

Verified on iOS 10.3.2 (physical device) and BrowserStack iOS 10 cloud device.

1. Create a universal project
2. Run the following code **shortly after the app starts**:

```
var nowPlayingInfo = MPNowPlayingInfoCenter.DefaultCenter?.NowPlaying;
```

### Expected Behavior

The app should not crash.

### Actual Behavior

The app crashes with:

```
NativeObjectExtensions.GetNonNullHandle (ObjCRuntime.INativeObject self, System.String argumentName) /Users/builder/azdo/_work/1/s/xamarin-macios/src/ObjCRuntime/INativeObject.cs:27
NSDictionary.ObjectForKey (Foundation.NSObject key) /Users/builder/azdo/_work/1/s/xamarin-macios/src/build/ios/native/Foundation/NSDictionary.g.cs:380
NSDictionary.TryGetValue (Foundation.NSObject key, Foundation.NSObject& value) /Users/builder/azdo/_work/1/s/xamarin-macios/src/Foundation/NSDictionary.cs:335
/Users/builder/azdo/_work/1/s/xamarin-macios/src/MediaPlayer/MPNowPlayingInfoCenter.cs:181
MPNowPlayingInfoCenter.get_NowPlaying () /Users/builder/azdo/_work/1/s/xamarin-macios/src/MediaPlayer/MPNowPlayingInfoCenter.cs:216
```

### Environment

```
=== Visual Studio Community 2019 for Mac ===

Version 8.9.1 (build 34)
Installation UUID: 894112d3-ddf3-45dc-a475-2ee8e682f7a3
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000140

=== Mono Framework MDK ===

Runtime:
	Mono 6.12.0.140 (2020-02/51d876a041e) (64-bit)
	Package version: 612000140

=== Roslyn (Language Service) ===

3.9.0-6.21152.10+c10f884b30737542ddd84ca889a4aad9281ce210

=== NuGet ===

Version: 5.8.0.6860

[...]

PR #11847, opened 2021-06-05
[mediaplayer] Fix MPNowPlayingInfoCenter.DefaultCenter.NowPlaying < iOS 11.1

spouliot:
Ironically my nullability changes caused this... because they are not
yet complete and the lack of `ArgumentNullException` in `TryGetValue`
was misinterpreted [0] as null-aware (instead of not correctly caught/
handled inside `NSDictionary`).

Fix https://github.com/xamarin/xamarin-macios/issues/11827

[0] https://github.com/xamarin/xamarin-macios/pull/10790

### DIFF
--- a/src/MediaPlayer/MPNowPlayingInfoCenter.cs
+++ b/src/MediaPlayer/MPNowPlayingInfoCenter.cs
@@ -7,6 +7,8 @@
 // Copyright 2011, Xamarin Inc
 //
 
+using System.Diagnostics.CodeAnalysis;
+
 using Foundation;
 using ObjCRuntime;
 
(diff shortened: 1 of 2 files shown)


src/MediaPlayer/MPNowPlayingInfoCenter.cs
```diff
@@ -138,73 +140,81 @@ void Add (NSMutableDictionary dictionary, NSObject key, NSObject value)
 				dictionary.Add (key, value);
 		}
 
+		bool TryGetValue (NSDictionary source, NSObject? key, [NotNullWhen (true)] out NSObject? result)
+		{
+			if (key != null)
```
Therzok, line 145: I don't understand the null check here, it's already done in MPNowPlayingInfo?

rolfbjarne (approved):

src/MediaPlayer/MPNowPlayingInfoCenter.cs
spouliot: > it's already done in MPNowPlayingInfo?

where is `key` checked for `null` ?

mandel-macaque (approved):

src/MediaPlayer/MPNowPlayingInfoCenter.cs
```diff
@@ -138,73 +140,81 @@ void Add (NSMutableDictionary dictionary, NSObject key, NSObject value)
...
+				return source.TryGetValue (key, out result);
+			result = null;
+			return false;
+		}
+
 		internal MPNowPlayingInfo (NSDictionary? source)
 		{
 			if (source == null)
```
Therzok, line 153: Key is null checked here @spouliot

Therzok: Ah, nvm...

PR #11846, opened 2021-06-04
[Localization] Adding New Github Action

tj-devel709:
- When new localization translations are created, the Loc team will put them into branches prefixed with "lego/*".
- Then a github action will take those translations and create a PR into main with those changes.
- After we merge that PR into main, we will need to sync the Localization branch with the changes we just made in main.
- This github action "Sync Localization Branch" will look for when we merge a PR in main coming from the 'github-actions[bot]' and that has the label 'localization_bot' and sync the Localization branch by deleting it and remaking it out of main!

Solves this issue: https://github.com/xamarin/xamarin-macios/issues/11791

Issue #11791, opened 2021-06-02
[Localization] Replace Localization branch


tj-devel709:
We will need to make sure that we replace the Localization branch every time that we land a PR containing the lcl translation files coming from our Get Localization Translations github action! 
Perhaps this can be done from inside the same github action or another will be necessary?

PR #11848, opened 2021-06-07
[Localization] Adding New Github Action (In new branch)

tj-devel709:
Trying [this PR](https://github.com/xamarin/xamarin-macios/pull/11846/) again here in a different branch 
since Azure Devops was giving this complaint "Azure Pipelines could not run because the pipeline triggers exclude this branch/path."

This could be referring to a change in the yaml and not the branch I am making the PR in, but I thought I could check just in case!

Details from[ other PR](https://github.com/xamarin/xamarin-macios/pull/11846/)
- When new localization translations are created, the Loc team will put them into branches prefixed with "lego/*".
- Then a github action will take those translations and create a PR into main with those changes.
- After we merge that PR into main, we will need to sync the Localization branch with the changes we just made in main.
- This github action "Sync Localization Branch" will look for when we merge a PR in main coming from the 'github-actions[bot]' and that

 has the label 'localization_bot' and sync the Localization branch by deleting it and remaking it out of main!
Solves this issue: #11791


spouliot (approved):

mandel-macaque:
@tj-devel709 this will not run either as with https://github.com/xamarin/xamarin-macios/pull/11846 . You are changing the .githubactions folder only and we ignore those: https://github.com/xamarin/xamarin-macios/blob/main/tools/devops/automation/build-pipeline.yml#L174

Please, when something does not work as you expect and you don't know why, don't create two branches, it will just make things more confusing. If you want to execute the build on a branch that contains changes only in the github actions you should push to origin and not to a fork, then manually trigger the build. AFAIK the actions were ran by github here: https://github.com/xamarin/xamarin-macios/runs/2749920577?check_suite_focus=true

.github/workflows/localization_branch_sync.yml
```diff
@@ -0,0 +1,32 @@
+name: Sync Localization Branch
+on:
+  pull_request:
+    branches:
+      - 'main'
+    types: [closed]
+
+jobs:
+  delete_branch:
```
mandel-macaque, line 9: Do we really need two jobs? I don't see what are we gaining in creating to instances one that depends on the other one, both run-in a single step and both using ubuntu. I would understand if we were executing steps that need a diff OS. This is not needed.

.github/workflows/localization_branch_sync.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: 'Delete Localization branch'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dawidd6/action-delete-branch@v3
+      name: 'delete'
+      if: ${{ github.event.pull_request.merged == true && github.event.pull_request.user.login == 'github-actions[bot]' && contains(github.event.pull_request.labels.*.names, 'localization_bot') }}
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
```
mandel-macaque, line 17: Secrets will not be available in PRs that come from forks, have you considered this?

mandel-macaque (changes requested):

.github/workflows/localization_branch_sync.yml
tj-devel709: Oh I was not aware of that..
Although this is only for the Bot created in our other github action which I believe does not come from a fork so we should be okay here?

.github/workflows/localization_branch_sync.yml
tj-devel709: Ah okay I can make it one job

.github/workflows/localization_branch_sync.yml
mandel-macaque: yes, if the PR comes form a fork we will have an error. it will say that the token is empty, which is not very intuitive. Can you add a check and if secrets.GITHUB_TOKEN == "" print some message, like 'Action cannot be executed on forks" Or something of the kind?

.github/workflows/localization_branch_sync.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches:
+      - 'main'
+    types: [closed]
+
+concurrency: 
+  group: 'Localization-Sync'
+  cancel-in-progress: false
+
+jobs:
+  Replace Localization Branch:
```
mandel-macaque, line 13: I would remove the spaces and will just lowercase everything. You already have a human readable name in line 14.

.github/workflows/localization_branch_sync.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - uses: peterjgrainger/action-create-branch@v2.0.1
+      name: 'Create Localization Branch'
+      if: ${{ github.event.pull_request.merged == true && github.event.pull_request.user.login == 'github-actions[bot]' && contains(github.event.pull_request.labels.*.names, 'localization_bot') }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        branch: 'Localization'
+
+    - name: 'Check for Fork'
```
mandel-macaque, line 33: Move this to be the first step, but i really really like the trick :)

mandel-macaque (changes requested):
Small fixes.

.github/workflows/localization_branch_sync.yml
```diff
@@ -0,0 +1,35 @@
...
+  group: 'Localization-Sync'
+  cancel-in-progress: false
+
+jobs:
+  replaceLocalizationBranch:
+    name: 'Replace Localization Branch'
+    runs-on: ubuntu-latest
+    steps:
+    - name: 'Check for Fork'
+      run: echo "Action cannot be executed on forks"
```
mandel-macaque, line 18: Sorry I forgot to mention, do an exit 1 to stop the execution of the other steps.

mandel-macaque (changes requested):

mandel-macaque (approved):

PR #11851, opened 2021-06-07
[devops] Publish mlaunch as .nupkg during main build

premun:
Packages the mlaunch that was built during the pipeline previously as a .nupkg and publishes it into the [dotnet-eng feed](https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-eng):
- Version contains the maccore commit SHA
- If this SHA is already published in the feed, packaging and publishing is skipped

Target package: https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-eng&package=Microsoft.DotNet.Mlaunch&protocolType=NuGet (will show up after main build completes)

https://github.com/dotnet/core-eng/issues/13150


tools/devops/automation/scripts/bash/build-nugets.sh
```diff
@@ -18,3 +23,42 @@ DOTNET_PKG_DIR=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET_PKG
 make -C dotnet package -j
 cp -c "$DOTNET_PKG_DIR"/*.pkg ../package/
 cp -c "$DOTNET_PKG_DIR"/*.msi ../package/
+
+# Down from here we publish mlaunch as .nupkg
```
rolfbjarne, line 27: I think this logic could go into its own shell script (say `build-mlaunch-nuget.sh`)

tools/devops/automation/scripts/bash/build-nugets.sh
```diff
@@ -18,3 +23,42 @@ DOTNET_PKG_DIR=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET_PKG
...
+mkdir -p "$MLAUNCH_WORK_DIR/mlaunch/bin"
+mkdir -p "$MLAUNCH_WORK_DIR/mlaunch/lib/mlaunch"
+
+DOTNET6=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET6 | grep "^DOTNET6=" | sed -e 's/^DOTNET6=//')
+IOS_DESTDIR=$(make -C tools/devops print-abspath-variable VARIABLE=IOS_DESTDIR | grep "^IOS_DESTDIR=" | sed -e 's/^IOS_DESTDIR=//')
+MONOTOUCH_PREFIX=$(make -C tools/devops print-abspath-variable VARIABLE=MONOTOUCH_PREFIX | grep "^MONOTOUCH_PREFIX=" | sed -e 's/^MONOTOUCH_PREFIX=//')
+
+# Copy mlaunch to staging area
+cp -r "$MACCORE_TOP/tools/mlaunch/Xamarin.Hosting/Xamarin.Launcher/bin/Debug/mlaunch.app" "$MLAUNCH_WORK_DIR/mlaunch/lib/mlaunch"
+cp "$IOS_DESTDIR$MONOTOUCH_PREFIX/bin/mlaunch" "$MLAUNCH_WORK_DIR/mlaunch/bin/"
```
rolfbjarne, line 49: Cloning is faster and uses less disk space:

```suggestion
cp -c -r "$MACCORE_TOP/tools/mlaunch/Xamarin.Hosting/Xamarin.Launcher/bin/Debug/mlaunch.app" "$MLAUNCH_WORK_DIR/mlaunch/lib/mlaunch"
cp -c "$IOS_DESTDIR$MONOTOUCH_PREFIX/bin/mlaunch" "$MLAUNCH_WORK_DIR/mlaunch/bin/"
```

tools/devops/automation/scripts/bash/build-nugets.sh
```diff
@@ -18,3 +23,42 @@ DOTNET_PKG_DIR=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET_PKG
...
+
+# Add the .csproj we will use to create the .nupkg
+cp "$XAM_TOP/tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj" "$MLAUNCH_WORK_DIR"
```
rolfbjarne, line 52:
```suggestion
cp -c "$XAM_TOP/tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj" "$MLAUNCH_WORK_DIR"
```

tools/devops/automation/scripts/bash/build-nugets.sh
```diff
@@ -18,3 +23,42 @@ DOTNET_PKG_DIR=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET_PKG
...
+
+# We need to override global.json to use .NET 6.0
+cp "$XAM_TOP/global6.json" "$MLAUNCH_WORK_DIR/global.json"
```
rolfbjarne, line 55:
```suggestion
cp -c "$XAM_TOP/global6.json" "$MLAUNCH_WORK_DIR/global.json"
```

tools/devops/automation/scripts/bash/build-nugets.sh
```diff
@@ -18,3 +23,42 @@ DOTNET_PKG_DIR=$(make -C tools/devops print-abspath-variable VARIABLE=DOTNET_PKG
...
+
+# We have to build from within the dir to respect the global.json
+cd "$MLAUNCH_WORK_DIR"
+"$DOTNET6" pack --version-suffix "$MACCORE_HASH"
+
+# We store mlaunch NuGet in [build work root]/mlaunch
+cd "$XAM_TOP"
+mkdir ../mlaunch
+cp "$MLAUNCH_WORK_DIR"/*.nupkg ../mlaunch/
```
rolfbjarne, line 64:
```suggestion
cp -c "$MLAUNCH_WORK_DIR"/*.nupkg ../mlaunch/
```

tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<!-- This file is used to pack the already-built mlaunch as a .nupkg -->
+
+	<PropertyGroup>
+		<TargetFramework>net6.0</TargetFramework>
+		<!-- Suppresses the warnings about the package not having assemblies in lib/*/.dll.-->
+		<NoPackageAnalysis>true</NoPackageAnalysis>
+		<VersionPrefix>1.0.0</VersionPrefix>
```
rolfbjarne, line 9: I think we should come up with a better version than 1.0.0.

We could use the existing mlaunch version:

```shell
$ /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/bin/mlaunch --version 
Xamarin.Launcher 14.21.0.330 (xamarin/main: cddbd1915d)
```

or we could come up with something else (Xcode version? Something completely different that's calculated in maccore and doesn't take xamarin-macios into account? Go with "0.0.1" and decide something later?)

@spouliot / @dalexsoto what do you think? I lean towards the mlaunch version (although the downside here is that there can be multiple mlaunch versions for the same maccore hash).

rolfbjarne (changes requested):

tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj
spouliot: IMO it would be useful to use the Xcode version for which `mlaunch` was built (and known to work [1]) for. We should also print that version (at least for the nuget version, maybe not the one we ship bundled in the legacy SDK)

[1] and _should_ work on earlier Xcode too...

Using legacy version is confusing as 14.21 means almost nothing, except iOS 14.x -> Xcode 12.x
That already requires mapping (-2) but only works for the major version.

Using 1.0.0 would require a new totally mapping, ending up in more confusion.

tools/devops/automation/scripts/bash/build-mlaunch-nuget.sh
```diff
@@ -0,0 +1,56 @@
...
+cp -c -r "$MACCORE_TOP/tools/mlaunch/Xamarin.Hosting/Xamarin.Launcher/bin/Debug/mlaunch.app" "$MLAUNCH_WORK_DIR/mlaunch/lib/mlaunch"
+cp -c "$IOS_DESTDIR$MONOTOUCH_PREFIX/bin/mlaunch" "$MLAUNCH_WORK_DIR/mlaunch/bin/"
+
+# Add the .csproj we will use to create the .nupkg
+cp -c "$XAM_TOP/tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj" "$MLAUNCH_WORK_DIR"
+
+# We need to override global.json to use .NET 6.0
+cp -c "$XAM_TOP/global6.json" "$MLAUNCH_WORK_DIR/global.json"
+
+version=$("$MLAUNCH_WORK_DIR/mlaunch/bin/mlaunch" --version | cut -d " " -f 2)
```
rolfbjarne, line 47: Let's do what Sebastien said, and use the Xcode version (the XCODE_VERSION variable in Make.config).

premun: Done!

premun:
I have one more change I want to add - I want to add some NuGet hooks so that I will be able to specify something like this:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <MlaunchDestinationDir>$(IntermediateOutputPath)\mlaunch</MlaunchDestinationDir>
  </PropertyGroup>
  
  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.Mlaunch" Version="12.5-cdbefac" />
  </ItemGroup>

</Project>
```

And the NuGet will copy the `mlaunch` folder there before Build.

@rolfbjarne should I add this as part of this PR or separate?

rolfbjarne (approved):

rolfbjarne:
@premun that can be a separate PR once this is in

premun:
The current package name is `Microsoft.DotNet.Mlaunch.12.5.0-cddbd1915d.nupkg`

If the build goes green, we can merge?

tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj
spouliot: I still see 1.0.0 here ?

tools/devops/automation/scripts/bash/build-mlaunch-nuget.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Add the .csproj we will use to create the .nupkg
+cp -c "$XAM_TOP/tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj" "$MLAUNCH_WORK_DIR"
+
+# We need to override global.json to use .NET 6.0
+cp -c "$XAM_TOP/global6.json" "$MLAUNCH_WORK_DIR/global.json"
+
+# We have to build from within the dir to respect the global.json
+cd "$MLAUNCH_WORK_DIR"
+"$DOTNET6" pack --version-suffix "$MACCORE_HASH" /p:VersionPrefix=$XCODE_VERSION
```
spouliot, line 50: Sadly we might need more than one mlaunch release for each version of Xcode, bug happens :(

> The current package name is Microsoft.DotNet.Mlaunch.12.5.0-cddbd1915d.nupkg

If you do an update (for Xcode 12.5.0) then only the digest (cddbd1915d) will change. 
It will still be a unique name but no one will be able to figure out which is the newest version.

Something more (like the commit distance or however net6 compute `21276.13`) is needed to distinguish the published packages.

spouliot (changes requested):

tools/mlaunch/Microsoft.DotNet.Mlaunch.csproj
premun: @spouliot there's a default that needs to be in (for example if you build it locally). We override that in the CI:
```bash
"$DOTNET6" pack --version-suffix "$MACCORE_HASH" /p:VersionPrefix=$XCODE_VERSION
```

tools/devops/automation/scripts/bash/build-mlaunch-nuget.sh
premun: I searched Arcade and we use this formula in .NET 6:
```
(yy * 1000 + mm * 50 + dd).(BuildRevision)
```

So today's number is:
```
21 * 1000 + 06 * 50 + 10 = 21310.x
```

source: https://github.com/dotnet/arcade/blob/e81d8c9bdc1de22623657afb23fab210dd04ca82/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L57

So I can do `12.5.0-21310.1-cddbd1915d`?
Or we rather want `21310.1-12.5.0-cddbd1915d`?

premun: Or I have seen in Xamarin previously `+cddbd1915d` or `+sha_cddbd1915d` I think

premun:
@rolfbjarne I added the feature with the extraction because I wasn't able to merge, had the changes in my branch and wouldn't be able to use the mlaunch without it anyway. It's not a lot of code and I added a README explaining the usage to the NuGet:
https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/

Hope it's fine

tools/devops/automation/scripts/bash/build-mlaunch-nuget.sh
rolfbjarne: I think something like this could work:

> 12.5.0-21310.1.cddbd1915d

rolfbjarne (approved):

tools/devops/automation/scripts/bash/build-mlaunch-nuget.sh
premun: So, we now have `Microsoft.DotNet.Mlaunch.12.5.0-21310.9.cddbd1915d.nupkg`

Built here:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4857230&view=logs&j=600d49c6-e96b-5e8b-2c3a-9f351e2e5848&t=c9117ceb-ee46-564f-0d31-c57e59243b74&l=181

spouliot (approved):

PR #11854, opened 2021-06-07
[CI] Unify the password across templates.

mandel-macaque:
Needs to be only merge when all device bots and mac bots have at least used provisionator to install the deps. If they fail, the PR is wrong.


spouliot (approved):

mandel-macaque:
Catalina ran tests, intro failed but it is unrelated. \o/

PR #11887, opened 2021-06-10
[dotnet] Add support for generating a binary version of runtimeconfig.json. Fixes #11745.

### LINKED ISSUE #11745: .NET: Generate a binary version of runtimeconfig.json for mobile targets
We need to generate a binary version of runtimeconfig.json for certain features to work.

Steps:

* Use Mono's task to convert *.runtimeconfig.json to a binary blob.
* Bundle the blob in the .app
* Don't bundle the json in the app.

Ref: https://github.com/xamarin/xamarin-macios/pull/11693#discussion_r640092943
Ref: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/docs/design/mono/mobile-runtimeconfig-json.md
Ref: https://github.com/dotnet/runtime/issues/49237
Ref: https://github.com/xamarin/xamarin-android/pull/6009

rolfbjarne:
Use Mono's RuntimeConfigParserTask to parse the *.runtimeconfig.json file and
produce a binary version of it. This way Mono doesn't have to do json parsing
at runtime, which makes it possible to make the runtimer smaller and execution
faster.

This also means implementing support for finding the on-disk location of the
file at runtime, and passing it to mono.

Ref: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/docs/design/mono/mobile-runtimeconfig-json.md

Fixes https://github.com/xamarin/xamarin-macios/issues/11745.


rolfbjarne:
CC @jonathanpeppers

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -507,6 +510,26 @@
 	</Target>
 
 	<!-- App bundle creation tasks -->
+
+	<Target Name="_CreateRuntimeConfiguration"
+		Condition="'$(UseMonoRuntime)' != 'false'"
```
jonathanpeppers, line 515: You might also check here if `$(GenerateRuntimeConfigurationFiles)` is `true`?

If a user sets it to `false` in their project, I think the `*.runtimeconfig.json` file would not be created:

https://github.com/dotnet/sdk/blob/415f1b90e14ddbbcfd6e2cb2d91b27e72db54db0/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L238

rolfbjarne: Hm yeah, good point, maybe we could even warn that some features won't work in that case.

mandel-macaque (approved):

runtime/monovm-bridge.m
```diff
@@ -392,11 +394,48 @@
...
+}
+
 bool
 xamarin_bridge_vm_initialize (int propertyCount, const char **propertyKeys, const char **propertyValues)
 {
 	int rv;
 
+	xamarin_initialize_runtime_config ();
+
 	rv = monovm_initialize (propertyCount, propertyKeys, propertyValues);
```
spouliot, line 439: does not this call reset the values of the config file ?

spouliot (commented):
> which makes it possible to make the runtimer 

typo `runtime[r]`

> smaller and execution faster.

Any numbers ?

runtime/monovm-bridge.m
rolfbjarne: No, the values we pass to the task here:

https://github.com/xamarin/xamarin-macios/pull/11887/files#diff-2e24f1b6f44df7580244d00f106fa1a9e2257536fbca9d7bbdc74571549324a9R522-R524

list the values we will later pass to `monovm_initialize`, so that the task can make sure there are no duplicates.

This is documented here: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/docs/design/mono/mobile-runtimeconfig-json.md#design-overview

rolfbjarne:
> > smaller and execution faster.
> 
> Any numbers ?

No, I didn't measure this, it's partially mentioned here: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/docs/design/mono/mobile-runtimeconfig-json.md#motivation, but I've also seen it elsewhere (although I don't remember exactly where anymore).

rolfbjarne:
Test failures are unrelated

* monotouch-test/Mac [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2442
* introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): https://github.com/xamarin/maccore/issues/2454
* Generator tests/NUnit: https://github.com/xamarin/xamarin-macios/issues/11947

tipa:
@rolfbjarne I get this warning (which was introduced by this PR) on a iOS extension:
`
/usr/local/share/dotnet/packs/Microsoft.iOS.Sdk/16.2.1024/targets/Xamarin.Shared.Sdk.targets(979,3): warning : Some features may not work correctly, because the generation of the runtime configure file (*.runtimeconfig.json) has been disabled. [/Users/.../ShareExtension/ShareExtension.csproj]
`

I have not set `GenerateRuntimeConfigurationFiles` to `false`, but when I set it to `true` explicitly, the warning disappears. On the main project, the warning does not show up, even without having to set `GenerateRuntimeConfigurationFiles=false`.
Is it safe to ignore the warning? Should I set `GenerateRuntimeConfigurationFiles=false` or does this have any unwanted side-effects?

PR #11891, opened 2021-06-10
[release/6.0.1xx-preview5] [dotnet] Don't allow any other branches to pretend to be release branches.

rolfbjarne:
This prevents a problem where:

a) A branch is created in origin based on this release branch.
b) The new branch is automatically built in CI, and produces packages which
   are uploaded to NuGet.
c) Some other PR is merged into the same release branch.
d) The merged commit ends up with the same version number as the branch from
   b), because we hardcoded branch name in the version number.
e) The CI for the merged commit tries to upload packages to NuGet, which
   fails, because there already are packages on NuGet.
e) Hair pulling ensues.


mandel-macaque (approved):

dalexsoto (approved):
ahh nice catch

spouliot (approved):
💇

PR #11892, opened 2021-06-10
[msbuild] Fix P/Invoke to stat.

rolfbjarne:
The native 'stat' function is not identical between x64 and arm64:

* The x64 version is the 32-bit version of the function, and takes a 32-bit
  version of the stat structure.
* The arm64 version is the 64-bit version of the function, and takes a 64-bit
  version of the stat structure.

We used the 32-bit version of the stat structure everywhere, which went
horribly wrong on arm64, because the layout and size of the struct is quite
different (120 bytes vs 144 bytes). The actual result on arm64 was that the
call to stat would null out one of the parameters to the calling function, and
the C# code would throw an ArgumentNullException when the code tried to use
that argument, and you'd end up with this exception which didn't make any
sense at all when looking at the code:

    error : System.ArgumentNullException: Parameter "itemSpec" cannot be null. [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]
    error :    at Microsoft.Build.Shared.ErrorUtilities.VerifyThrowArgumentNull(Object parameter, String parameterName, String resourceName) in Microsoft.Build.Utilities.Core.dll:token 0x6000121+0x20 [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]
    error :    at Microsoft.Build.Shared.ErrorUtilities.VerifyThrowArgumentNull(Object parameter, String parameterName) in Microsoft.Build.Utilities.Core.dll:token 0x6000120+0x0 [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]
    error :    at Microsoft.Build.Utilities.TaskItem..ctor(String itemSpec) in Microsoft.Build.Utilities.Core.dll:token 0x60003ba+0x6 [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]
    error :    at Xamarin.MacDev.Tasks.SmartCopyTaskBase.CopyFile(String source, String target, String targetItemSpec) in /Users/builder/azdo/_work/1/s/xamarin-macios/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/SmartCopyTaskBase.cs:line 74 [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]
    error :    at Xamarin.MacDev.Tasks.SmartCopyTaskBase.Execute() in /Users/builder/azdo/_work/1/s/xamarin-macios/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/SmartCopyTaskBase.cs:line 99 [/Users/rolf/work/net6-mobile-samples/HelloiOS/HelloiOS.csproj]

So fix the P/Invoke to stat:

* Use the 64-bit version of the stat struct.
* For x64, call the 64-bit version of stat (stat$INODE64), since we're now
  using the 64-bit version of the stat struct.
* For arm64, call the standard version of stat (which is the 64-bit version).

And this means that we can now use the ARM64 version of .NET to build .NET
apps.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/SmartCopyTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/SmartCopyTaskBase.cs
@@ -123,34 +123,50 @@ namespace Xamarin.MacDev.Tasks
 			return !Log.HasLoggedErrors;
 		}
 
-		[DllImport ("libc")]
-		static extern int stat (string file_name, out Stat buf);
+		[DllImport ("/usr/lib/libc.dylib", EntryPoint = "stat$INODE64")]
+		static extern int stat_x64 (string file_name, out Stat buf);
+
+		[DllImport ("/usr/lib/libc.dylib", EntryPoint = "stat")]
+		static extern int stat_arm64 (string file_name, out Stat buf);
+
+		static int stat (string path, out Stat buf)
+		{
+			if (RuntimeInformation.ProcessArchitecture == Architecture.Arm64) {
+				return stat_arm64 (path, out buf);
+			} else {
+				return stat_x64 (path, out buf);
+			}
+		}
 
 		[DllImport ("libc")]
 		static extern int chmod (string path, FilePermissions mode);
 
+		/* when _DARWIN_FEATURE_64_BIT_INODE is defined */
 		struct Stat {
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):
wow nice! 👍

rolfbjarne:

PR #11893, opened 2021-06-10
Try to prevent identical version numbers across branches by matching hardcoded prerelease identifiers with hardcoded branch name.

rolfbjarne:
If we hardcode a prelease identifier for a .NET release, we can easily end up
with multiple branches using the same version number, because the hardcoded
prerelease identifier ends up being the same for any other branches that are
created later from the release branch (one typical example would be backport
branches).

This is bad.

So try to prevent this, by hardcoding two values instead!

We'll hardcode the branch name where the hardcoded prerelease identifier
should be applied, and this way any other branches will not use the hardcoded
prerelease identifier.

Ref: https://github.com/xamarin/xamarin-macios/pull/11891.

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -73,9 +73,22 @@ else ifneq ($(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER),)
 PULL_REQUEST_ID=$(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER)
 endif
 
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne:

PR #11908, opened 2021-06-11
[ARKit] Provide updates for Xcode13 beta1.

mandel-macaque:
Simple diff, all other enums are the same, just got moved to a diff header.

### DIFF
--- a/src/arkit.cs
+++ b/src/arkit.cs
@@ -190,6 +190,7 @@ namespace ARKit {
 		HorizontalPlane,
 		VerticalPlane,
 		AnyPlane,
+		GeoTracking,
 	}
 
 	[iOS (13,0)]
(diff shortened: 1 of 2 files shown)


spouliot (approved):

src/arkit.cs
```diff
@@ -190,6 +190,7 @@ public enum ARCoachingGoal : long {
 		HorizontalPlane,
 		VerticalPlane,
 		AnyPlane,
```
spouliot, line 192: missing `[iOS (15,0)]`

spouliot (changes requested):

src/arkit.cs
mandel-macaque: I noticed just when I did it :/ I'm rusty

rolfbjarne (approved):

spouliot (approved):

PR #11919, opened 2021-06-14
[CI] Fix BRANCH_NAME to be the entire branch name.

rolfbjarne:
The Build.SourceBranchName variable is documented to be the last segment of
the git ref. If the git ref is 'refs/heads/main', then Build.SourceBranchName
is 'main' (as expected). However, if the git ref is
'refs/heads/release/6.0.1xx-preview5', then Build.SourceName is
'6.0.1xx-preview5', and that's not what we want.

Instead use the Build.SourceBranch variable, which is the entire git ref, and
then we strip out the 'refs/heads/' part from the beginning. This way we get
the correct branch name.


rolfbjarne:
Test failures are unrelated:

* monotouch-test/iOS Unified 32-bits - simulator/Debug: https://github.com/xamarin/xamarin-macios/issues/11504
* introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): https://github.com/xamarin/maccore/issues/2433

spouliot (approved):

tj-devel709 (approved):

PR #11928, opened 2021-06-14
[CarPlay] Add support for Xcode13 beta1.

mandel-macaque:


src/carplay.cs
```diff
@@ -164,6 +164,97 @@ public enum CPTextButtonStyle : long
...
+		RoundaboutExit10 = 37,
+		RoundaboutExit11 = 38,
+		RoundaboutExit12 = 39,
+		RoundaboutExit13 = 40,
+		RoundaboutExit14 = 41,
+		RoundaboutExit15 = 42,
+		RoundaboutExit16 = 43,
+		RoundaboutExit17 = 44,
+		RoundaboutExit18 = 45,
+		RoundaboutExit19 = 46,
```
rolfbjarne, line 248: I wonder where this roundabout is?

rolfbjarne (approved):

spouliot (approved):

src/carplay.cs
dalexsoto: Una rotonda with 19 exits!

dalexsoto (approved):

src/carplay.cs
```diff
@@ -164,6 +164,97 @@ public enum CPTextButtonStyle : long
...
+
+	[NoWatch, NoTV, NoMac, iOS (15,0)]
+	public enum CPTrafficSide : byte
+	{
+		Right = 0,
+		Left = 1,
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (15,0)]
+	public enum CPManeuverType : byte
```
tj-devel709, line 200: So the "uint8_t" type should be "byte" in our bindings?

tj-devel709 (approved):

src/carplay.cs
spouliot: yes, `u` for unsigned and `8` for 8-bits -> `System.Byte`

src/carplay.cs
mandel-macaque: Place Charles de Gaulle is a good contender 


Screenshot 2021-06-14 at 17 47 29

PR #11932, opened 2021-06-15
[corefoundation] Fix `CFString.ToString` to cache the result

spouliot:
`CFString` cache the native immutable string into it's `str` field and
use it to avoid p/invokes when possible.

`ToString` retrieved the string without setting `str` (if `null`) so it
missed opportunities to avoid the native calls.


dalexsoto (approved):

rolfbjarne (approved):

spouliot:

PR #11940, opened 2021-06-15
[linker] Ignore `[AssemblyMetadata]` before net6

spouliot:
This was allowed to ease existing (3rd party) code migration into net6.

However the reverse can also happen: trying to use net6 code with the
legacy SDK. In such case it's possible other (newer) attributes are being
used to preserve members.

ref: https://github.com/dotnet/aspnetcore/issues/33269

### DIFF
--- a/src/Foundation/LinkerSafeAttribute.cs
+++ b/src/Foundation/LinkerSafeAttribute.cs
@@ -25,8 +25,10 @@
 using System;
 
 namespace Foundation {
-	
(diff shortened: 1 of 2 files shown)


rolfbjarne (approved):

mandel-macaque (approved):

dalexsoto (approved):

spouliot:

PR #11943, opened 2021-06-15
[LinkPresentation] Xcode 13 Beta 1 Changes

tj-devel709:
I believe this one should have the Mac Catalyst header:

1. The method documentation says supported in mac catalyst https://developer.apple.com/documentation/linkpresentation/lpmetadataprovider/3752081-startfetchingmetadataforrequest?language=objc
2. The interface also says it supports mac catalyst https://developer.apple.com/documentation/linkpresentation/lpmetadataprovider?language=objc
3. This method was in the mac catalyst todo file


rolfbjarne (approved):

tj-devel709:
Question: The wiki says this is also available in tvOS but there was not todo file for tvOS, the online docs do not mention tv, sharpie did not have any tv availability, and the tests passed locally. I can assume that the wiki was lying, right?

spouliot:
> Question: The wiki says this is also available in tvOS but there was not todo file for tvOS, the online docs do not mention tv, sharpie did not have any tv availability, and the tests passed locally. I can assume that the wiki was lying, right?

What the wiki (Xcode-diff) tells you is that the headers for tvOS has the symbol - but it does not mean the symbol is available on tvOS. 

This often happens because the same header files are shared (good thing) across all platforms.

If you look at https://developer.apple.com/documentation/linkpresentation/lpmetadataprovider?language=objc you'll see that the tvOS is **not** mentioned for the type. Now web docs are often incorrect so a better way to check this is to open the header file (path included in the wiki's diff).

```
LP_EXTERN API_AVAILABLE(macos(10.15), ios(13.0)) API_UNAVAILABLE(watchos, tvos)
@interface LPMetadataProvider : NSObject
```

As you'll have confirmation that, in this case, the type is not available on tvOS - so it make sense that none of it's member will be available (unless there's a mistake somewhere... that can happen)

spouliot (approved):

tj-devel709:
There are 2 unrelated test failures:

- monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2443
- introspection/watchOS 32-bits - simulator/Debug (watchOS 5.0): https://github.com/xamarin/maccore/issues/2454

dalexsoto (approved):

PR #11945, opened 2021-06-15
[PhotosUI] Xcode13 PhotosUI beta1

tj-devel709:
I wasn't quite sure if I should create an enums file for this one enum introduced to the file?


spouliot:
> I wasn't quite sure if I should create an enums file for this one enum introduced to the file?

Add new enums into existing binding files. That way the generator will process them. That's useful in many cases, including automatic conversion of availability attributes for net6.

src/photosui.cs
```diff
@@ -580,5 +592,17 @@ interface PHPhotoLibrary_PhotosUISupport
...
+		[Export ("presentLimitedLibraryPickerFromViewController:completionHandler:")]
+		void PresentLimitedLibraryPicker (UIViewController controller, Action<NSArray<NSString>> completionHandler);
+	}
+
+	[NoWatch, NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+	[Native]
+	public enum PHPickerConfigurationSelection : long
+	{
+		Default = 0,
+		Ordered = 1
```
spouliot, line 606: Add a `,` after the last element, because:
* It's valid C# :-)
* It avoid changing this line the next time a member is added (smaller diff to review, better git history)

src/photosui.cs
```diff
@@ -580,5 +592,17 @@ interface PHPhotoLibrary_PhotosUISupport
 	{
 		[Export ("presentLimitedLibraryPickerFromViewController:")]
 		void PresentLimitedLibraryPicker (UIViewController controller);
+
+		[NoWatch, NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+		[Export ("presentLimitedLibraryPickerFromViewController:completionHandler:")]
+		void PresentLimitedLibraryPicker (UIViewController controller, Action<NSArray<NSString>> completionHandler);
```
spouliot, line 598: The API clash with `string[] PreselectedAssetIdentifiers { get; set; }`

In one case you force the use of `NSArray<NSString>` as the parameter of the `completionHandler`.
But you expose the pre-selection assets as a `string[]`. 
That can force (another) conversion of the data inside user code.

Using `Action<string[]>` should work fine.

tests/xtro-sharpie/MacCatalyst-PhotosUI.todo
```diff
@@ -1,10 +0,0 @@
-!missing-protocol! PHContentEditingController not bound
-!missing-type! PHEditingExtensionContext not bound
-## appended from unclassified file
```
spouliot, line 3: ^ hint ^ everything above this line is part of the backlog and needs a bit more validation

src/photosui.cs
```diff
@@ -24,8 +24,8 @@ namespace PhotosUI {
 	[NoTV]
 	[iOS (8, 0)]
 	[Mac (10, 13)]
+	[MacCatalyst (15,0)]
```
spouliot, line 27: According to web doc this was in Catalyst 14+

When working on the backlog you need to check the version where this was added (since it's not new in Xcode 13)

`/Applications/Xcode_13.0.0-beta.app//Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/PhotosUI.framework/Headers/PHContentEditingController.h`
shows this

```
API_AVAILABLE_BEGIN(macos(10.11), ios(8), macCatalyst(14))
```

since it's not new to Xcode 13 (iOS15) then `[MacCatalyst (14,0)]` makes sense (unless it fails running intro on BS)

src/photosui.cs
```diff
@@ -472,9 +476,9 @@ interface PHProjectTypeDescriptionInvalidator
 	}
 
 	[iOS (8,0)]
+	[MacCatalyst (15,0)]
```
spouliot, line 479: web doc shows it's deprecated since iOS 13
Double check headers - it might not be decorated for catalyst (and since it's deprecated has little value to include)

spouliot (changes requested):

src/photosui.cs
tj-devel709: Oops, thanks!

src/photosui.cs
tj-devel709: Ah okay I see! Thank you!

src/photosui.cs
tj-devel709: @spouliot  Can you explain this a little more?
I think I get that this action is taking an NSArray<NSString> and PreselectedAssetIdentifiers returns a string[]
but how are these two related? Is this a general thing within files to not sure both string[] and NSArray<NSString> and just one of them?

src/photosui.cs
```diff
@@ -126,6 +126,10 @@ interface PHLivePhotoView {
 	[Protocol, Model]
 	[BaseType (typeof (NSObject))]
 	interface PHLivePhotoViewDelegate {
+		[TV (15,0), iOS (15,0), MacCatalyst (15,0), Mac (12,0)]
+		[Export ("livePhotoView:canBeginPlaybackWithStyle:")]
+		bool LivePhotoView (PHLivePhotoView livePhotoView, PHLivePhotoViewPlaybackStyle playbackStyle);
```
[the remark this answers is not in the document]
tj-devel709: Good catch! That makes sense, thank you!

src/photosui.cs
spouliot: > but how are these two related? I

I was suspicious reading the diff... so I looked at the docs/headers and they seems related.

Also we generally do not use `NSArray<NSString>` inside bindings (grep src) but prefer `string[]`

`NSString` is more useful (even required in some cases) when dealing with constants (since they can be compared by pointers, inside native code, and not by value).

src/photosui.cs
```diff
@@ -525,6 +529,10 @@ interface PHPickerConfiguration : NSCopying
 		[Export ("preferredAssetRepresentationMode", ArgumentSemantic.Assign)]
 		PHPickerConfigurationAssetRepresentationMode PreferredAssetRepresentationMode { get; set; }
 
+		[NoWatch, NoTV, NoMac, iOS (15, 0), MacCatalyst (15,0)]
```
mandel-macaque, line 532: Lets me consistent with the spaces, either (15, 0) or (15,0)

src/photosui.cs
```diff
@@ -533,6 +541,10 @@ interface PHPickerConfiguration : NSCopying
 
 		[Export ("initWithPhotoLibrary:")]
 		IntPtr Constructor (PHPhotoLibrary photoLibrary);
+
+		[NoWatch, NoTV, NoMac, iOS (15, 0), MacCatalyst (15,0)]
```
mandel-macaque, line 545: Same

mandel-macaque (changes requested):

src/photosui.cs
tj-devel709: I changed that and also removed the NoWatch, NoTV and NoMac since the interface header already had those labels

tj-devel709: Oops sorry, actually forgot it haha

src/photosui.cs
```diff
@@ -24,8 +24,8 @@ namespace PhotosUI {
 	[NoTV]
 	[iOS (8, 0)]
 	[Mac (10, 13)]
+	[MacCatalyst (14,0)]
 	[Protocol]
-	[Unavailable (PlatformName.MacCatalyst)]
 	[Advice ("This API is not available when using UIKit on macOS.")]
```
rolfbjarne, line 29: You need to remove this advice too if you remove `[Unavailable (PlatformName.MacCatalyst)]`

```suggestion
```

src/photosui.cs
```diff
@@ -474,7 +478,6 @@ interface PHProjectTypeDescriptionInvalidator
 	[iOS (8,0)]
 	[NoMac][NoTV]
 	[DisableDefaultCtor]
-	[Unavailable (PlatformName.MacCatalyst)]
```
rolfbjarne, line 477: This doesn't look correct, the `PHEditingExtensionContext` type doesn't show up for Mac Catalyst: https://developer.apple.com/documentation/photokit/pheditingextensioncontext?language=objc

If it turns out it actually is available (the documentation is wrong), then you need to remove the advice on the next line too.

rolfbjarne (changes requested):

src/photosui.cs
tj-devel709: I'm sorry, you are right! I had added the MacCatalyst there and since removed it, but forgot to add that Unavailable back in!

spouliot (approved):

rolfbjarne (approved):

src/photosui.cs
```diff
@@ -580,5 +591,17 @@ interface PHPhotoLibrary_PhotosUISupport
 	{
 		[Export ("presentLimitedLibraryPickerFromViewController:")]
 		void PresentLimitedLibraryPicker (UIViewController controller);
+
+		[iOS (15,0), MacCatalyst (15,0)]
+		[Export ("presentLimitedLibraryPickerFromViewController:completionHandler:")]
+		void PresentLimitedLibraryPicker (UIViewController controller, Action<string[]> completionHandler);
```
mandel-macaque, line 597: Looking at the docs, this looks like an Async candidate: https://developer.apple.com/documentation/photokit/phphotolibrary/3752108-presentlimitedlibrarypickerfromv

mandel-macaque (changes requested):

src/photosui.cs
mandel-macaque: The new swift async/await makes it easy to know, from the docs:

```
You can call this method from synchronous code using a completion handler, as shown on this page, or you can call it as an asynchronous method that has the following declaration:
```

tj-devel709: Ah good call, are most completionHandlers a good indicator of an async method?

mandel-macaque: Most of the time, the new docs for swift make it easier to know. Intro will complain sometimes if it has the info.

tj-devel709: Updated, but forgot to run the tests, I'll do that now

tj-devel709: Okay they look like they are passing!

mandel-macaque (approved):

PR #11946, opened 2021-06-16
Use `CFString` instead of `NSString` to get a handle from a (.net) `System.String`

spouliot:
This is possible because both types are toll-free bridged [0].

How ?

* Renamed `CFString.LowLevelCreate` to `CreateNative` so it match `NSString` API
* Make it public so it can be used for generated/3rd party bindings
* Added a _safe_ `CFString.ReleaseNative` matching `NSString` API
* Update generator to use the new API (instead of the NSString version)
* Update manual bindings (using older API) to use the new API (few in PrintCore)

Why ?

It's no secret that p/invoke (C) are faster than calling a selector
(ObjC). In most cases we do not have a choice what to call... but in a
few, but commonly used, cases we can pick the fastest call.

In this case the difference is larger than the previous case [1] since there's
two call (create and release) involved.

|                 Method |          name |                value |        Mean |       Error |    StdDev |
|----------------------- |-------------- |--------------------- |------------:|------------:|----------:|
| NSString_CreateRelease |         empty |                      |  4,936.1 ns |   349.08 ns |  19.13 ns |
| CFString_CreateRelease |         empty |                      |    243.1 ns |    10.30 ns |   0.56 ns |
| NSString_CreateRelease |   long_string | ????(...)???? [4096] |  8,270.3 ns |   837.09 ns |  45.88 ns |
| CFString_CreateRelease |   long_string | ????(...)???? [4096] |  3,212.4 ns |   213.73 ns |  11.72 ns |
| NSString_CreateRelease |   short_7bits |              Bonjour |  4,858.5 ns | 1,671.67 ns |  91.63 ns |
| CFString_CreateRelease |   short_7bits |              Bonjour |    242.5 ns |    18.36 ns |   1.01 ns |
| NSString_CreateRelease |  short_accent |               Québec |  4,990.1 ns |   343.96 ns |  18.85 ns |
| CFString_CreateRelease |  short_accent |               Québec |    377.6 ns |    15.96 ns |   0.87 ns |
| NSString_CreateRelease |   short_emoji | I'm f(...)ight. [23] |  5,080.2 ns |   794.29 ns |  43.54 ns |
| CFString_CreateRelease |   short_emoji | I'm f(...)ight. [23] |    391.2 ns |     8.60 ns |   0.47 ns |
| NSString_CreateRelease | short_unicode |                汉语 漢語 |  4,980.4 ns |   272.41 ns |  14.93 ns |
| CFString_CreateRelease | short_unicode |                汉语 漢語 |    376.4 ns |    15.98 ns |   0.88 ns |

Also simplified/updated previous tests as suggested in [1].

|                 Method |          name |                value |        Mean |       Error |    StdDev |
|----------------------- |-------------- |--------------------- |------------:|------------:|----------:|
|    NSString_FromString |         empty |      140735350713040 |  2,252.2 ns |   591.63 ns |  32.43 ns |
|    CFString_FromString |         empty |      140735350713040 |    408.4 ns |    47.56 ns |   2.61 ns |
|    NSString_FromString |   long_string |      140295885995872 | 14,887.2 ns |   183.67 ns |  10.07 ns |
|    CFString_FromString |   long_string |      140295885995872 |  7,418.7 ns | 1,844.42 ns | 101.10 ns |
|    NSString_FromString |   short_7bits |  1292713981587835779 |  2,259.8 ns |   210.17 ns |  11.52 ns |
|    CFString_FromString |   short_7bits |  1292713981587835779 |    453.0 ns |   122.77 ns |   6.73 ns |
|    NSString_FromString |  short_accent |      140295885995776 |  2,448.2 ns |    97.43 ns |   5.34 ns |
|    CFString_FromString |  short_accent |      140295885995776 |    256.8 ns |    75.81 ns |   4.16 ns |
|    NSString_FromString |   short_emoji |      140296423669120 |  2,531.6 ns |   199.02 ns |  10.91 ns |
|    CFString_FromString |   short_emoji |      140296423669120 |    259.0 ns |    21.32 ns |   1.17 ns |
|    NSString_FromString | short_unicode |      140295885992304 |  2,426.2 ns |   339.06 ns |  18.58 ns |
|    CFString_FromString | short_unicode |      140295885992304 |    260.6 ns |    97.03 ns |   5.32 ns |

[1] https://github.com/xamarin/xamarin-macios/pull/11809

### DIFF
--- a/src/CoreFoundation/CFString.cs
+++ b/src/CoreFoundation/CFString.cs
@@ -115,14 +115,14 @@ namespace CoreFoundation {
 		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
 		extern static IntPtr CFStringGetCharacters (IntPtr handle, CFRange range, IntPtr buffer);
 
-		public static IntPtr CreateNative (string str)
+		public static IntPtr CreateNative (string value)
 		{
-			if (str == null)
+			if (value is null)
 				return IntPtr.Zero;
 			
-			return CFStringCreateWithCharacters (IntPtr.Zero, str, str.Length);
+			return CFStringCreateWithCharacters (IntPtr.Zero, value, value.Length);
 		}
-		
+
 		public static void ReleaseNative (IntPtr handle)
 		{
 			if (handle != IntPtr.Zero)
--- a/src/CoreFoundation/CFString.cs
+++ b/src/CoreFoundation/CFString.cs
@@ -115,7 +115,7 @@ namespace CoreFoundation {
 		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
 		extern static IntPtr CFStringGetCharacters (IntPtr handle, CFRange range, IntPtr buffer);
 
-		internal static IntPtr LowLevelCreate (string str)
+		public static IntPtr CreateNative (string str)
 		{
 			if (str == null)
 				return IntPtr.Zero;
@@ -123,6 +123,12 @@ namespace CoreFoundation {
 			return CFStringCreateWithCharacters (IntPtr.Zero, str, str.Length);
 		}
 		
+		public static void ReleaseNative (IntPtr handle)
+		{
+			if (handle != IntPtr.Zero)
+				CFObject.CFRelease (handle);
+		}
+
 		public CFString (string str)
 		{
 			if (str == null)
--- a/src/Foundation/NSString.cs
+++ b/src/Foundation/NSString.cs
@@ -82,6 +82,7 @@ namespace Foundation {
 			}
 		}
 
+		[Obsolete ("Use of 'CFString.CreateNative' offers better performance.")]
 		public static IntPtr CreateNative (string str)
 		{
 			return CreateNative (str, false);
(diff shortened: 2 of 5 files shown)


src/CoreFoundation/CFString.cs
```diff
@@ -115,14 +115,20 @@ public class CFString
 		[DllImport (Constants.CoreFoundationLibrary, CharSet=CharSet.Unicode)]
 		extern static IntPtr CFStringGetCharacters (IntPtr handle, CFRange range, IntPtr buffer);
 
-		internal static IntPtr LowLevelCreate (string str)
+		public static IntPtr CreateNative (string str)
```
rolfbjarne, line 118: Not a big deal, but now that the API is public I think we should also make the parameter name nicer:

```suggestion
		public static IntPtr CreateNative (string value)
```

rolfbjarne (approved):

src/CoreFoundation/CFString.cs
spouliot: hmm... I copied the exact signature from `NSString`
```csharp
public static IntPtr CreateNative (string str)
```
but I also find it nicer :)

tj-devel709 (approved):

spouliot:
> xcframework-test/Mac Catalyst/Debug: TimedOut (Execution timed out after 1200 seconds. No test log file was produced)

unrelated, known issue -> https://github.com/xamarin/maccore/issues/2414

> Generator tests/NUnit: Failed (Execution failed with exit code 1)

Fixed by https://github.com/xamarin/xamarin-macios/pull/11953

PR #11948, opened 2021-06-16
[monotouch-tests] Use a LinkDescription item to specify extra linker description files

rolfbjarne:
* It's easier to fix up the path to the linker description in xharness when
  cloning project files. This way maybe we'll be able to remove the [hardcoded
  logic in xharness][1] to handle ${ProjectDir}.
* .NET doesn't understand the ${ProjectDir} syntax, so this makes it possible
  to build these projects from the command line.

[1]: https://github.com/dotnet/xharness/blob/b2297d610df1ae15fc7ba8bd8c9bc0a7192aaefa/src/Microsoft.DotNet.XHarness.iOS.Shared/Utilities/ProjectFileExtensions.cs#L1268


spouliot (approved):

dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne:

PR #11949, opened 2021-06-16
[tools] Fix quoting quotes in StringUtils.QuoteForProcess.

rolfbjarne:
Previously quoting a string with a quote:

    a"b

would result in:

    "a\"""b"

There are way too many quotes there.

We'll now get:

    "a\"b"

which is correct.

### DIFF
--- a/tools/common/StringUtils.cs
+++ b/tools/common/StringUtils.cs
@@ -84,7 +84,6 @@ namespace Xamarin.Utils {
 			foreach (var c in f) {
 				if (c == '"') {
 					s.Append ('\\');
-					s.Append (c).Append (c);
 				} else if (c == '\\') {
 					s.Append (c);
 				}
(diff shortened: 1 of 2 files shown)


spouliot (changes requested):
`tests/mtouch/StringUtilsTest.cs` is begging for attention ;-)

dalexsoto (approved):
After attention is given 😄

tools/common/StringUtils.cs
```diff
@@ -84,7 +84,6 @@ public static string QuoteForProcess (string f)
 			foreach (var c in f) {
 				if (c == '"') {
 					s.Append ('\\');
```
tj-devel709, line 86: In your example, you used single backslash `\` but it looks like here you are using double backslash. Is this desired?

tj-devel709 (approved):

mandel-macaque (approved):

spouliot (approved):

rolfbjarne:

Issue #11947, opened 2021-06-16
GeneratorTests.BGenTests.StackOverflow20696157 fails


rolfbjarne:
The `GeneratorTests.BGenTests.StackOverflow20696157` fails in `main`:

```
GeneratorTests.BGenTests.StackOverflow20696157:
build
Expected: 0
But was: 1
```

    => GeneratorTests.BGenTests.StackOverflow20696157
    GeneratorTests.BGenTests.StackOverflow20696157
    	sof20696157.cs
[...]

PR #11953, opened 2021-06-16
[tests][generator] Remove references to iAd.framework types

spouliot:
TL&DR: They were removed from headers so we can't use them for our tests.

Long Story:

To avoid app being rejected the selectors must be removed from the app.
This means the removal of the `[Export]` and `[Bind]` attributes inside
the platform assemblies. That, in turn, makes the binding generator
unable to create bindings for those types.

If some 3rd party bindings extended those types then they also need to
remove them (they will not be able to submit with the symbols anyway).
Stubs can be used, whether or not binary compatibility is provided for
the 3rd party bindings consumers.

Fix https://github.com/xamarin/xamarin-macios/issues/11947

### DIFF
--- a/tests/generator/sof20696157.cs
+++ b/tests/generator/sof20696157.cs
@@ -88,7 +88,7 @@ using VideoToolbox;
 using WatchConnectivity;
 using WatchKit;
 using WebKit;
-using iAd;
+
 namespace Test {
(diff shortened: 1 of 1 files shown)


rolfbjarne (approved):

dalexsoto (approved):

spouliot:

rolfbjarne:
@sudo backport d16-10

PR #11963, opened 2021-06-16
[CoreHaptics] Add support for Xcode13 beta1.

mandel-macaque:


mandel-macaque:
@spouliot  locally I have the following when running xtro:

```
!missing-field! CHHapticAudioResourceKeyUseVolumeEnvelope not bound
!missing-field! CHHapticPatternKeyEventWaveformUseVolumeEnvelope not bound
```

Yet we have:

* https://github.com/xamarin/xamarin-macios/pull/11963/files#diff-72997dd733774832d35cfe84972b74f4957eb7c5a566ee86c63f83dac2531b4fR417 
* https://github.com/xamarin/xamarin-macios/pull/11963/files#diff-72997dd733774832d35cfe84972b74f4957eb7c5a566ee86c63f83dac2531b4fR353

Files are added in mac (else all the other will fail) and the other platforms pick the fields correctly from the keys of the strong dict. Any ideas?

src/corehaptics.cs
```diff
@@ -373,6 +377,8 @@ partial interface CHHapticPatternDefinition {
 		NSObject WeakParameterCurve { get; set; }
 		[Export ("ParameterCurveControlPointsKey")]
 		NSObject WeakParameterCurveControlPoints { get; set; }
+		[Advice ("The default value is true.")]
+		bool EventWaveformUseVolumeEnvelope { get; set; }
```
spouliot, line 381: hmm... is this only used as an output ?
because if it's an output then, when created, the value will be `false`
This can be bound as `bool?` if the default can (or should) be ignore

spouliot: You need to copy the availability `[TV (15,0), NoWatch, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]`
since this is the public API developers will be using (they keys being `internal`)

src/corehaptics.cs
```diff
@@ -403,4 +409,20 @@ interface CHHapticPattern {
...
+	partial interface CHHapticAudioResourceKeys {
+		[Field ("CHHapticAudioResourceKeyUseVolumeEnvelope")]
+		NSString UseVolumeEnvelopeKey { get; }
+	}
+
+	[Mac (12,0), iOS (15,0), TV (15,0), MacCatalyst (15,0), NoWatch]
+	[StrongDictionary ("CHHapticAudioResourceKeys")]
+	partial interface CHHapticAudioResourceDefinition {
+		[Advice ("The default value is true.")]
+		bool UseVolumeEnvelope { get; set; }
```
spouliot, line 425: same as above, better use `bool?` to let developers not change the default

spouliot (changes requested):

src/corehaptics.cs
mandel-macaque: So this is backed by a NSDictionary, AFAIK if they do nothing, the key should not be added. But We can do a bool?

mandel-macaque: error BI1033: bgen: Limitation: can not automatically create strongly typed dictionary for (System.Nullable`1[System.Boolean]) the value type of the CoreHaptics.CHHapticPatternDefinition.EventWaveformUseVolumeEnvelope property

rolfbjarne (approved):

src/corehaptics.cs
spouliot: hmm... is that not a `NSNumber` backed bool ?

mandel-macaque: @dalexsoto why are we not able to do nullable? Do we need to fix that?

dalexsoto: Please leave your message after the tone. beep... Yeah [docs](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#strong-dictionary-types) says nothing about nullable types so probably are not supported.

mandel-macaque: Already supported as per the API diff:
```csharp
public class CHHapticAudioResourceDefinition : Foundation.DictionaryContainer {
	// constructors
	public CHHapticAudioResourceDefinition ();
	public CHHapticAudioResourceDefinition (Foundation.NSDictionary dictionary);
	// properties
	public bool? UseVolumeEnvelope { get; set; }
}
```

spouliot (approved):

PR #11965, opened 2021-06-17
[CoreLocation] Add xcode13 beta 1 support.

mandel-macaque:


src/corelocation.cs
```diff
@@ -871,4 +890,19 @@ interface CLBeaconIdentityConstraint : NSCopying, NSSecureCoding {
 		[BindAs (typeof (short?))]
 		NSNumber Minor { get; }
 	}
+
+	[Model]
```
rolfbjarne, line 894: It's unclear how this protocol is supposed to be used, if it's a *Delegate type somewhere or not (it doesn't seem to show up anywhere else in the API).

In the meantime, I think we can remove the `[Model]` attribute, just in case it's not a *Delegate type. It's easier to add it later than to remove it.

rolfbjarne (changes requested):

src/corelocation.cs
```diff
@@ -431,6 +441,15 @@ partial interface CLLocationManager {
 		[Export ("authorizedForWidgetUpdates")]
 		bool IsAuthorizedForWidgetUpdates { [Bind ("isAuthorizedForWidgetUpdates")] get; }
 
+		[Async]
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15, 0)]
```
spouliot, line 445: Missing Catalyst. It's not excluded on type-level so it must be annotated for new API.

src/corelocation.cs
```diff
@@ -431,6 +441,15 @@ partial interface CLLocationManager {
...
+		[Export ("startMonitoringLocationPushesWithCompletion:")]
+		void StartMonitoringLocationPushes ([NullAllowed] Action<NSData, NSError> completion);
+
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15, 0)]
```
spouliot, line 449: Missing Catalyst. It's not excluded on type-level so it must be annotated for new API.

src/corelocation.cs
```diff
@@ -871,4 +890,19 @@ interface CLBeaconIdentityConstraint : NSCopying, NSSecureCoding {
 		[BindAs (typeof (short?))]
 		NSNumber Minor { get; }
 	}
+
+	[Model]
+	[iOS (15,0), TV (15,0), MacCatalyst (15,0), Mac (12,0), Watch (8,0)]
+	[Protocol]
+	[BaseType (typeof(NSObject))]
```
spouliot, line 897: remember to remove the `[BaseType]` when removing the `[Model]`

spouliot (changes requested):

src/corelocation.cs
mandel-macaque: I added model due to the documentation from the method that the protocol has:

```
+// Called just before this extension is terminated by the system.
```

But I'll remove, since it can be re-added.

spouliot (approved):

PR #11966, opened 2021-06-17
[CFNetwork] Add Xcode13 beta 1 support.

mandel-macaque:
CFHost and all its callers have been deprecated. The other deprecated APIs are not in used by Xamarin.


rolfbjarne (changes requested):
We use `CFStreamCreatePairWithSocketToCFHost`

https://github.com/xamarin/xamarin-macios/blob/2b7dc07d20fc06344d7f164af8934f5e755572c2/src/CoreFoundation/CFStream.cs#L200

which is deprecated too: https://developer.apple.com/documentation/cfnetwork/1426831-cfstreamcreatepairwithsockettocf?language=objc

This shows up in xtro as well:

https://github.com/xamarin/xamarin-macios/blob/0921ae4e26fcceab13b4dc084390eda5219eaf88/tests/xtro-sharpie/iOS-CoreFoundation.todo#L1-L4

src/CFNetwork/CFHost.cs
```diff
@@ -21,7 +21,11 @@ namespace CFNetwork {
 namespace CoreServices {
 #endif
 
-	// used by CFStream.cs (only?)
+	[Deprecated (PlatformName.WatchOS, 8, 0, message: "Use the new Network framework instead.")]
+	[Deprecated (PlatformName.TvOS, 15, 0, message: "Use the new Network framework instead.")]
+ 	[Deprecated (PlatformName.iOS, 15, 0, message: "Use the new Network framework instead.")]
+ 	[Deprecated (PlatformName.MacCatalyst, 15, 0, message: "Use the new Network framework instead.")]
+ 	[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use the new Network framework instead.")]
```
spouliot, line 28: you can remove "new" since it's not new anymore (and that won't surface the test of time)

src/CoreFoundation/CFStream.cs
```diff
@@ -203,6 +203,11 @@ public abstract class CFStream : CFType, INativeObject, IDisposable {
 			/* CFReadStreamRef __nullable * __nullable */ out IntPtr readStream,
 			/* CFWriteStreamRef __nullable * __nullable */ out IntPtr writeStream);
 
+		[Deprecated (PlatformName.WatchOS, 8, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.TvOS, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.iOS, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.MacCatalyst, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use the new Network framework instead.")]
```
spouliot, line 210: same

src/CoreFoundation/CFStream.cs
```diff
@@ -222,6 +227,11 @@ public abstract class CFStream : CFType, INativeObject, IDisposable {
 			/* CFStringRef */ IntPtr host, /* UInt32 */ int port,
 			/* CFReadStreamRef* */ out IntPtr readStream, /* CFWriteStreamRef* */ out IntPtr writeStream);
 
+		[Deprecated (PlatformName.WatchOS, 8, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.TvOS, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.iOS, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.MacCatalyst, 15, 0, message: "Use the new Network framework instead.")]
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use the new Network framework instead.")]
```
spouliot, line 234: same.

We already use `"Use 'Network.framework' instead."` in a lot of other places. It would be best to use the same string to reduce the metadata size of the assembly.

You could also add it to `src/ObjCRuntime/ObsoleteConstants.cs` so it's easier to avoid different string ending up inside the platform assemblies.

That would require updating (now or later) `src/Security/SecureTransport.cs` and `src/Security/SslConnection.cs`

spouliot (changes requested):

src/CoreFoundation/CFStream.cs
mandel-macaque: Its a trap

spouliot (approved):

rolfbjarne (approved):

dalexsoto (approved):

PR #11971, opened 2021-06-17
Remove the option of disabling the windows-specific part of the .NET build.

rolfbjarne:
* Having .NET enabled with the windows-specific parts disabled is not a
  scenario tested on CI (where we've always enabled both). This means it's
  prone to bitrotting.
* It's another configuration to keep track of and make work locally.
* It doesn't work right now anyway.

So just always enable the windows-specific parts of the .NET build, if the
.NET build is enabled.


spouliot (approved):

dalexsoto (approved):

mandel-macaque (approved):

emaf (approved):

rolfbjarne:

PR #11977, opened 2021-06-17
[VideoSubscriberAccount] Updates for Xcode13 Beta 1

tj-devel709:
There were a lot of items inside the MacCatalyst todo file.
After running introspection tests, checking the web docs for all of them, and the xcode headers for most of them, 
it does not look like they are valid to be marked for MacCatalyst.


mandel-macaque (changes requested):
If there is an API that is not present in the platforms, it does not go to the ignore file, but we use a No*Attr, in all those, you should have used [NoMacCatalyst] like here: https://github.com/xamarin/xamarin-macios/blob/main/src/uikit.cs#L1519

tj-devel709:
> If there is an API that is not present in the platforms, it does not go to the ignore file, but we use a No*Attr, in all those, you should have used [NoMacCatalyst] like here: https://github.com/xamarin/xamarin-macios/blob/main/src/uikit.cs#L1519

Ah so we add the NoMacCatalyst and remove it from the ignore?

spouliot:
Part of the issue is that 

```
	VideoSubscriberAccount \
```

is part of of `MACCATALYST_FRAMEWORKS =` inside `src/frameworks.sources`

So the bindings are included in Catalyst - and that means you must tell the generator not to process stuff that is not available on that platform.

Now looking back at the framework headers it seems that everything is not available at Catalyst, so you can remove its inclusion.

However to help "future you", the day Apple starts to include it, you better add the same `[NoMacCatalyst]` to new API when it's documented that way in the header files.

tj-devel709:
@spouliot So after I removed VideoSubscriberAccount from the frameworks.sources and removed the ignore file, xtro still complains saying the items are not bound in MacCatalyst. So I still need the NoMacCatalyst right?

spouliot:
> xtro still complains saying the items are not bound in MacCatalyst

I'm not sure why xtro pick the API as missing :(
There might be a bad macro somewhere since the headers are clear

```
/// A VSAccountManager instance coordinates access to a subscriber's account.
API_AVAILABLE(ios(10.0), tvos(10.0)) API_UNAVAILABLE(macCatalyst)
@protocol VSAccountManagerDelegate <NSObject>
```

> So I still need the NoMacCatalyst right?

You should add it (like my previous comment about "future you") as it might become useful later. However that won't solve the problem.

Open an issue (for xtro) and add a link to it inside the `macCatalyst-VideoSubscriberAccount.ignore`. I'll have a look why later (but that might be after my vacations).

tj-devel709:
@spouliot @mandel-macaque 
I removed this API from the frameworks.sources under MacCatalyst
Then VSSubscriptionAccessLevel and VSSubscriptionAccess started complaining about not being in MacCatalyst so I added the NoMacCatalyst to those as well and added all the selectors xtro was complaining about to the ignore file as well as linked this issue here https://github.com/xamarin/maccore/issues/2462

src/videosubscriberaccount.cs
```diff
@@ -103,10 +103,12 @@ interface IVSAccountManagerDelegate { }
 	interface VSAccountManagerDelegate {
 
 		[Abstract]
+		[NoMac]
```
rolfbjarne, line 106: This is a breaking change, it shows up as such in the `API & Generator diff`

Note that it seems you'll have to download the diff from Azure DevOps and hunt through the files to find it, because the links here don't work: https://xamarin.github.io/macios.ci/pr/PR11977/4890030/apicomparison/mac-api-diff.html

I think the best you can do is to add an `[Obsoleted (...)]` attribute, and hardcode an exception in introspection if there are any errors.

CC @mandel-macaque ^

rolfbjarne (changes requested):

src/videosubscriberaccount.cs
tj-devel709: @rolfbjarne So I made these changes https://gist.github.com/tj-devel709/24bf5f179de2cb09184b6946f229637b based on other examples I found. Introspection does not seem to complain but Xtro is saying:
```
!extra-protocol-member! unexpected selector VSAccountManagerDelegate::accountManager:dismissViewController: found
!extra-protocol-member! unexpected selector VSAccountManagerDelegate::accountManager:presentViewController: found
```
Would I also add this to the ignore with a comment about marking it obsoleted?

mandel-macaque: @rolfbjarne those links should work, is there a diff between API diff and Generator diff? I might have been cleaning too many files..

rolfbjarne: @mandel-macaque open https://xamarin.github.io/macios.ci/pr/PR11977/4890030/apicomparison/mac-api-diff.html and try to click on any of the links there. Nothing happens.

rolfbjarne: @tj-devel709 yes, this warrants an ignore addition to xtro.

tj-devel709: @rolfbjarne Okay, updated!

tj-devel709: I'm also not really sure how to read the Generator Diff.. Looks like the last iteration may have been complaining about me adding the NoMacCatalyst at certain places?

rolfbjarne (approved):

spouliot (approved):

mandel-macaque (approved):

PR #11979, opened 2021-06-18
[CoreLocationUI] Add new framework for Xcode13 beta1.

mandel-macaque:


tools/common/Frameworks.cs
```diff
@@ -484,6 +486,8 @@ public static Frameworks GetwatchOSFrameworks (bool is_simulator_build)
 
 				{ "Accessibility", "Accessibility", 7,0 },
 				{ "UniformTypeIdentifiers", "UniformTypeIdentifiers", 7,0 },
+
+				{ "CoreLocationUI", "CoreLocationUI", 8,0 },
```
spouliot, line 490: why ? it does not look like it's available for watchOS

src/corelocationui.cs
```diff
@@ -0,0 +1,49 @@
...
+	public enum CLLocationButtonLabel : long {
+		None = 0,
+		CurrentLocation,
+		SendCurrentLocation,
+		SendMyCurrentLocation,
+		ShareCurrentLocation,
+		ShareMyCurrentLocation,
+	}
+
+	[NoTV, NoMac, NoWatch, iOS (15,0), MacCatalyst (15,0)]
```
spouliot, line 29: `API_AVAILABLE(ios(15.0),watchos(8.0))`
I guess it's not in the watchOS header ?

spouliot (changes requested):
watchOS is inconsistent 
if the API were not added (like headers suggest) then it should not be defined for watchOS frameworks

rolfbjarne (approved):

tools/common/Frameworks.cs
mandel-macaque: I think we can remove it and add the ignore

spouliot (approved):

tools/mtouch/Makefile
```diff
@@ -169,6 +169,8 @@ SIMLAUNCHER_FRAMEWORKS =  \
 	-weak_framework NearbyInteraction \
 						\
 	-weak_framework AdServices	\
+						\
+	-weak_framework CoreLocationUI \
```
tj-devel709, line 173: What is this for?

tools/common/Frameworks.cs
```diff
@@ -418,6 +418,8 @@ public static Frameworks CreateiOSFrameworks (bool is_simulator_build)
 
 				{ "AdServices", "AdServices", 14,3 },
 
+				{ "CoreLocationUI", "CoreLocationUI", 15,0 },
```
tj-devel709, line 421: When do we do this?

mandel-macaque: When a new framework is added.

mandel-macaque:
Known issues: https://github.com/xamarin/maccore/issues/2443

PR #11980, opened 2021-06-18
[corefoundation] Enable nullability on `CFString` and `NativeObject`

spouliot:
then fix the impact this had on existing nullable-aware code,
including the generator

Also
* Hide `CFObject` empty (of public members) for future profiles
* Fix `CFString.ToString ()` to never return `null`


rolfbjarne (approved):

spouliot:

Issue #10861, opened 2021-03-12
Default RuntimeIdentifier


rolfbjarne:
Currently we require the RuntimeIdentifier to be set in order to build on all the Apple platforms, but having a default RuntimeIdentifier might make sense, at least in certain cases (and we don't have to do the same thing on all platforms).

### Desktop platforms (macOS + Mac Catalyst)

For desktop platforms (macOS + Mac Catalyst), we have the choice between x64 and arm64, and at least for these platforms I think we should set a default. Options:

1. Require the developer to specify (what we do now).
2. Set the default to x64. Most compatible (runs on both arm64 and x64).
3. Set the default to arm64. Most future-proof (probably), but will be non-intuitive when building on an x64 machine (since the app won't launch).
4. Set the default to the CPU of the machine. Probably the best user experience, but it's undeterministic, since the results depend on the machine you're using to build (CI can become frustrating).

IMHO the second option is the one that makes the most sense.

At least for now I think it makes most sense to have the same default for both platforms.

### Mobile platforms (iOS + tvOS)

We have more options here, because the app can target either a simulator or a device - this is really what it comes down to, if we should default to simulator or device, or if we should require the user to specify.

1. Require the developer to specify (what we do now).
2. Default to build for simulator.
    Pro: should always be able to build something that can be launched.
    Con: ?
3. Default to build for device.
4. Detect if there are any devices attached, and if so build for device, if not build for the simulator
    Cons: undeterministic. Someone trips over a network cable that disconnects your Apple TV, and suddenly your build behaves differently?
5. Build for both the simulator and device just in case.
    Cons: Slow. Will always spend time building something that is not needed.

IMHO we should default to the simulator (if we decide to default to anything).

We should also have in mind that we might want to change the default later.

Any other options/opinions?

We should also validate that the RuntimeIdentifier is one of the values we support.

jonathanpeppers:
My thinking for making the iOS simulator default is to make the getting started experience smoother. A first-time user probably won't have a device connected yet, but the simulator would be able to launch fine. An experienced user will more likely understand how to change a setting to deploy to a device.

Hopefully, the IDE would be able to select the right `$(RuntimeIdentifier)` when you select a device in the dropdown?

spouliot:
> Desktop platforms (macOS + Mac Catalyst)

Agreed with option 2.
We can print something like `Building for x86_64 (default). Override with XXX` on the console (useful in logs).

> Mobile platforms (iOS + tvOS)

[...]

Issue #10294, opened 2020-12-15
.NET 6: add support for builds with multiple architectures (fat apps)


rolfbjarne:
Android does an inner build per RID: https://github.com/xamarin/xamarin-android/blob/95a9c28d10827c61bdf04a050b9020c11d78966d/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets#L66-L71

See also: [chat link](https://teams.microsoft.com/l/message/19:63068bcab1d34aa199acbe9dfcf4d24d@thread.skype/1620195574318?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=42121604-a67e-4f9a-9f3f-57c3c6d7ccd3&parentMessageId=1620195574318&teamName=Xamarin%20Platform%20Engineering&channelName=NET6&createdTime=1620195574318)

Ref: https://github.com/dotnet/sdk/issues/14359

---

Time estimate: 4 weeks.

PR #11983, opened 2021-06-18
[dotnet] Add support for universal apps.

rolfbjarne:
Add support for universal / fat apps for iOS, macOS and Mac Catalyst.

We detect if we need to build a universal app by checking if `RuntimeIdentifiers` (plural) is set, and in that case we do a complete inner build for every `RuntimeIdentifier`, and then once those inner builds are done, we merge the resulting .app bundles together (using a new MSBuild task called `MergeAppBundles`).

When merging app bundles together, we'll run into files that exist in both apps. Depending on the file type, we do different things:

* MachO flies: lipo'ed together.
* Managed assemblies: we do a binary comparison, if the assemblies are different, we put them in a RID-specific subdirectory. At runtime we know to look for assemblies in this directory.
* runtimeconfig.bin, icudt.dat: put in a RID-specific subdirectory.
* Info.plist: computed in the outer (fat) build, the one from the inner build is ignored.
* Other files: for identical files we just copy one, otherwise we show an error.

If we run into files that are different between apps, but we should handle somehow, then we'll have to decide on a case-to-case basis what to do.

Some code shuffling was required to increase code sharing between the tools/ code, the msbuild/ code, and tests.

I've also added support for a default `RuntimeIdentifier`.

@spouliot FYI this PR incorporates most of #11320 (it was needed to resolve the location of `icudt.dat` at runtime).

Fixes https://github.com/xamarin/xamarin-macios/issues/10294.
Fixes https://github.com/xamarin/xamarin-macios/issues/10861.


runtime/runtime.m
```diff
@@ -2384,6 +2387,15 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
 xamarin_vm_initialize ()
 {
 	char *pinvokeOverride = xamarin_strdup_printf ("%p", &xamarin_pinvoke_override);
+	char *icu_dat_file_path = NULL;
+
+	char path [1024];
+	if (!xamarin_locate_app_resource (xamarin_icu_dat_file_name, path, sizeof (path))) {
```
spouliot, line 2393: That looks like `ios-arm` and `ios-arm64` each as a copy of the icu data file ?
I don't recall if those are identical (or not) between architectures... but if they are then it's a big opportunity for sharing it

spouliot (approved):
❤️

runtime/runtime.m
rolfbjarne: Yes, the icu data file is different between architectures (the arm one is bigger):

```
 ls -la [...]/xamarin-macios/tests/dotnet/packages/microsoft.netcore.app.runtime.mono.ios-arm*/6.0.0-preview.6.21276.13/runtimes/ios-arm*/native/icudt.dat
-rwxr--r--  1 rolf  wheel  1713152 May 18 21:54 [...]/xamarin-macios/tests/dotnet/packages/microsoft.netcore.app.runtime.mono.ios-arm/6.0.0-preview.6.21276.13/runtimes/ios-arm/native/icudt.dat
-rwxr--r--  1 rolf  wheel  1636880 May 18 21:54 [...]/xamarin-macios/tests/dotnet/packages/microsoft.netcore.app.runtime.mono.ios-arm64/6.0.0-preview.6.21276.13/runtimes/ios-arm64/native/icudt.dat
```

tools/common/MachO.cs
```diff
@@ -495,6 +495,27 @@ public static bool IsDynamicFramework (string filename)
...
 		}
+
+		public static bool IsMachOFile (string filename)
+		{
+			using (var fs = new FileStream (filename, FileMode.Open, FileAccess.Read, FileShare.Read)) {
+				if (fs.Length < 4)
+					return false;
+				using (var reader = new BinaryReader (fs)) {
+					var magic = reader.ReadUInt32 ();
+					reader.BaseStream.Position = 0;
```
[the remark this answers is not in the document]
rolfbjarne: This was copied from somewhere else - I'll remove this line.

tools/common/PathUtils.cs
```diff
@@ -130,6 +131,25 @@ public static void CreateSymlink (string symlink, string target)
...
+			do {
+				buffer = new byte [(buffer?.Length ?? 0) + 1024];
+				rv = readlink (path, buffer, (IntPtr) (buffer.Length - 1));
+			} while (rv == buffer.Length - 1);
+
+			if (rv == -1)
+				throw new Exception (string.Format ("Could not readlink '{0}': {1}", path, Marshal.GetLastWin32Error ()));
+
+			return Encoding.UTF8.GetString (buffer, 0, rv);
+		}
```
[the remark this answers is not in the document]
rolfbjarne: Are you referring to https://github.com/xamarin/xamarin-macios/pull/11983/commits/7aec86fa0498def7a3dd02f98d2ea38f52a0a03d?

If so, this is new code (to get the target of a symlink), the code that moved was to check if a file is a symlink or not.

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -209,14 +243,101 @@
...
+		<Warning Text="$(_RuntimeIdentifiersClashMessage)" />
+	</Target>
+
+	<!-- If we have multiple runtime identifiers, then we need to run an inner build for each of them -->
+	<Target Name="_BuildRidSpecificAppBundle"
+		DependsOnTargets="Build"
+		Returns="@(_AssemblyPublishDirectory)"
+		>
+		<ItemGroup>
+			<_AssemblyPublishDirectory Include="$(MSBuildProjectDirectory)/$(_AppBundlePath)">
```
emaf, line 260: I'm not sure if using full path here will work from Windows, because we'll end up passing a Windows full path to the task that will be executed on macOS.

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -209,14 +243,101 @@
...
+			<_ArchitectureSpecificFiles Include="$(_GlobalizationDataFile)" Condition="'$(_GlobalizationDataFile)' != ''" />
+		</ItemGroup>
+
+		<MergeAppBundles
+			SessionId="$(BuildSessionId)"
+			Condition="'$(IsMacEnabled)' == 'true'"
+			ArchitectureSpecificFiles="@(_ArchitectureSpecificFiles)"
+			IgnoreFiles="Info.plist"
+			InputAppBundles="@(_AssemblyPublishDirectories)"
+			OutputAppBundle="$(MSBuildProjectDirectory)/$(_AppBundlePath)"
```
emaf, line 319: And we could potentially have another problem here.

dotnet/targets/Xamarin.Shared.Sdk.targets
rolfbjarne: @emaf I've changed this to be a relative path now.

dotnet/targets/Xamarin.Shared.Sdk.targets
rolfbjarne: @emaf same, I've changed this to be a relative path now.

emaf (approved):

dalexsoto (approved):

runtime/runtime.m
filipnavara: The difference is unintentional. I don't think there's a tracking issue for it but I already mentioned it few times in some comments. The ICU data are generated from same input using the same tools. My theory is that there is some compression involved and the zlib library consumed by different builds is slightly different version with different defaults.

akoeplinger: Yeah that's unintended, they should be identical. Filed https://github.com/dotnet/runtime/issues/55637

filipnavara: Thanks, I was about to file an issue and look into it.

EgorBo: @filipnavara I wonder if ICU data for 64bit is a bit bigger due to some 64bit paddings etc for performance sake 🙂

filipnavara: @EgorBo Unlikely. The arm64 ones on Andorid and iOS are different size (https://github.com/dotnet/designs/pull/225#discussion_r647159143). I don't think it's even memory mapped so the micro-optimizations would not make much difference.

PR #11986, opened 2021-06-18
[CoreSpotlight] Add Xcode13 beta1 support.

mandel-macaque:


src/corespotlight.cs
```diff
@@ -1075,5 +1092,14 @@ interface CSSearchQuery {
 		[Export ("cancel")]
 		void Cancel ();
 	}
+
+	[NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
+	interface CSImportExtension : NSExtensionRequestHandling
```
spouliot, line 1098: doc suggest this is to be subclassed inside an extension
IOW it's not usable as-is inside apps
might be worth making it `abstract`

src/corespotlight.cs
```diff
@@ -1075,5 +1092,14 @@ interface CSSearchQuery {
...
+	{
+		[Export ("updateAttributes:forFileAtURL:error:")]
+		bool UpdateAttributes (CSSearchableItemAttributeSet attributes, NSUrl contentUrl, [NullAllowed] out NSError error);
```
spouliot, line 1101: -> `Update`

`attributes` is the first parameter, not the action

In Swift the name is `update`

src/corespotlight.cs
```diff
@@ -985,6 +985,19 @@ interface CSSearchableItemAttributeSet : NSCopying, NSSecureCoding {
 		[NullAllowed, Export ("supportsNavigation", ArgumentSemantic.Strong)]
 		NSNumber SupportsNavigation { get; set; }
 
+		[NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+		[Field ("CSActionIdentifier")]
+		NSString ActionIdentifier { get; }
+
+		[NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+		[Export ("actionIdentifiers", ArgumentSemantic.Copy)]
+		string[] ActionIdentifiers { get; set; }
```
spouliot, line 994: you expose `ActionIdentifier` as a `NSString`
but `ActionIdentifiers` as `string[]`
if it's to be used with constant then it's better to keep them all as `NSString`, it avoid conversions and remove the risk that the constants are compared using pointers

spouliot (changes requested):

src/corespotlight.cs
tj-devel709: We want the abstract keyword because CSImportExtension will be subclassed and we want the user to implement it's properties and methods? (just Update in this case)

tj-devel709 (approved):

src/corespotlight.cs
rolfbjarne: Adding `[Abstract]` has its own set of problems (https://github.com/xamarin/xamarin-macios/issues/4969), although those might not necessarily apply to this case.

rolfbjarne (approved):

src/corespotlight.cs
mandel-macaque: ActionIdentifier is a Field that is used for the plist as per the docs:
```
Info dictionary of the specified NSUserActivity includes the corresponding Info.plist entry using the key CSActionIdentifier.
```
I think is safe to leave it as it is, maybe if anything move the plist to a string, but all other keys are NSStrings.

spouliot (approved):

mandel-macaque:
network issue :(

PR #11987, opened 2021-06-19
[PDFKit] Updating Xcode13 Beta 1

tj-devel709:
Wasn't sure if the PdfAccessPermissions type should be ulong.
Sharpie gave it the ulong type and used the `(1uL << #)` but the original header used NSUInteger which according to [this](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/objective-c-libraries?context=xamarin%2Fios&tabs=macos#simple-types) should be `nuint`

### DIFF
--- a/src/pdfkit.cs
+++ b/src/pdfkit.cs
@@ -317,6 +317,20 @@ namespace PdfKit {
 		Push,
 	}
 
+	[Native]
+	[iOS (15,0), Mac (12,0), MacCatalyst (15,0)]
+	public enum PdfAccessPermissions : ulong
+	{
+		LowQualityPrinting = (1uL << 0),
+		HighQualityPrinting = (1uL << 1),
+		DocumentChanges = (1uL << 2),
+		DocumentAssembly = (1uL << 3),
+		ContentCopying = (1uL << 4),
+		ContentAccessibility = (1uL << 5),
+		Commenting = (1uL << 6),
+		FormFieldEntry = (1uL << 7),
+	}
+
 	[Mac (10,13)]
 	[iOS (11,0)]
 	[Static]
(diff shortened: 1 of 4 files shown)


rolfbjarne (approved):
> Wasn't sure if the PdfAccessPermissions type should be ulong.
Sharpie gave it the ulong type and used the (1uL << #) but the original header used NSUInteger which according to this should be nuint

`ulong` is correct. We'd like to use `nuint`, but we can't do that for enums, C# doesn't allow it (only [s]byte, [u]short, [u]int and [u]long are allowed as base types for enums), so we go with the second best option: the 64-bit type. It's the correct type for 64-bit platforms (which will one day be all platforms), and then we add the `[Native]` attribute to tell the generator and our runtime that it should be treated as a 32-bit value on 32-bit platforms.

src/pdfkit.cs
```diff
@@ -1157,6 +1171,18 @@ interface PdfDocument : NSCopying {
...
+		[Field ("PDFDocumentFoundSelectionKey")]
+		NSString FoundSelectionKey { get; }
+
+		[iOS (15,0), Mac (12,0), MacCatalyst (15,0)]
+		[Field ("PDFDocumentPageIndexKey")]
+		NSString PageIndexKey { get; }
+
+		[iOS (15,0), Mac (12,0), MacCatalyst (15,0)]
+		[Field ("PDFDocumentAccessPermissionsOption")]
+		NSString AccessPermissionsOption { get; }
```
spouliot, line 1184: that last field should with inside `PdfDocumentWriteOptionKeys`

the _hint_ is that the previous fields (in the headers) are similarly named and (in binding files) already grouped into their own type

spouliot: and also update the `PdfDocumentWriteOptions` string dictionary :)

spouliot (changes requested):

tj-devel709:
@spouliot I added `PDFDocumentAccessPermissionsOption` to `PdfDocumentWriteOptionKeys` and `PdfDocumentWriteOptions`. I followed along with the pattern of the other elements in there, but not sure why the "+PDFKit" is there or if I needed to add that.

src/pdfkit.cs
tj-devel709: @spouliot I added `PDFDocumentAccessPermissionsOption` to `PdfDocumentWriteOptionKeys` and `PdfDocumentWriteOptions`. I followed along with the pattern of the other elements in there, but not sure why the "+PDFKit" is there or if I needed to add that.

tj-devel709: Oops forgot to run tests, I'll do that now

src/pdfkit.cs
```diff
@@ -434,6 +438,7 @@ interface PdfDocumentWriteOptions {
 
 		string OwnerPassword { get; set; }
 		string UserPassword { get; set; }
```
mandel-macaque, line 440: Is missing the availability attires which should be the same as those of the key.

mandel-macaque (changes requested):

src/pdfkit.cs
tj-devel709: Okay the tests look good!

src/pdfkit.cs
spouliot: without them the developer won't know they are not available (since the keys are not public)

and they (internal) keys also needs them otherwise introspection would not be able to do its job :)

tj-devel709: Updated :)

spouliot (approved):

mandel-macaque (approved):

PR #11991, opened 2021-06-21
[dotnet] Only limit the default inclusion items with a condition on 'EnableDefault<Platform>Items'. Fixes #11990.

### LINKED ISSUE #11990: Enable reuse of default items SDK script from Xamarin.Legacy.Sdk
[Xamarin.Legacy.Sdk](https://github.com/xamarin/Xamarin.Legacy.Sdk) is supposed to allow builds using legacy Xamarin SDKs for multi-targeting. One of the features is to use the same default items auto-import as on new SDKs. On MonoAndroid that's implemented using [including AutoImport.props](https://github.com/xamarin/Xamarin.Legacy.Sdk/blob/467d661ae9a11c7f9af8981f90be94492104f87d/src/Xamarin.Legacy.Sdk/Sdk/Sdk.props#L7) from the non-legacy SDK. The same approach doesn't work on iOS/macOS at the moment due to [extra 
`TargetPlatformIdentifier` check](https://github.com/xamarin/xamarin-macios/blob/6b78a868498e7e112c48d4d4a58b57324c79eadc/dotnet/targets/Microsoft.Sdk.DefaultItems.template.props#L21) in `Microsoft.Sdk.DefaultItems.template.props`. On Android the [check is not present](https://github.com/xamarin/xamarin-android/blob/15b40af7d62e0e2003d2a009576834a71967dbb2/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props#L23).

It would be nice to enable reuse of the AutoImport.props from the SDK. One possible solution suggested by @rolfbjarne was to change to conditions to `Condition="'$(EnableDefault@PLATFORM@Items)' == 'true'"` in the template.

cc @jonathanpeppers

rolfbjarne:
* This is the way Android does it:
  https://github.com/xamarin/xamarin-android/blob/15b40af7d62e0e2003d2a009576834a71967dbb2/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props#L23
* Makes it easier for the default inclusion logic to be used by
  Xamarin.Legacy.Sdk (as explained in the fixed issue).

Fixes https://github.com/xamarin/xamarin-macios/issues/11990.

### DIFF
--- a/dotnet/targets/Microsoft.Sdk.DefaultItems.template.props
+++ b/dotnet/targets/Microsoft.Sdk.DefaultItems.template.props
@@ -32,7 +32,7 @@
 	</ItemGroup>
 
 	<!-- Default Asset Catalog file inclusion -->
-	<ItemGroup Condition="'$(TargetPlatformIdentifier)' == '@PLATFORM@' And '$(_EnableDefaultXamarinItems)' == 'true' ">
+	<ItemGroup Condition="'$(EnableDefault@PLATFORM@Items)' == 'true'">
 		<ImageAsset Include="**\*.xcassets\**\*.png;**\*.xcassets\*\*.jpg;**\*.xcassets\**\*.pdf;**\*.xcassets\**\*.json" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)">
 			<Link>$([MSBuild]::MakeRelative ('$(MSBuildProjectDirectory)', '%(Identity)'))</Link>
 			<Visible>false</Visible>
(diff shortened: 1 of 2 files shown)


spouliot (approved):

filipnavara:
Can you change the `Default Asset Catalog file inclusion` too or is there a reason you skipped over it?

filipnavara (approved):

rolfbjarne:
> Can you change the `Default Asset Catalog file inclusion` too or is there a reason you skipped over it?
> 
There was no reason, except that my Search & Replace didn't catch it, and then I didn't verify there results 🤦

PR #11993, opened 2021-06-21
[CoreText] Add support for Xcode13 beta1.

mandel-macaque:
Added the invokes to be ignored as with macOS.


spouliot (commented):
why not implement them ?

if not, then don't label as `highlight` since it will go out in release notes without anything new in the release

mandel-macaque:
@spouliot We ignored them in mac OS for a reason, no one has requested them. I am just following what we did in macOS. (I've removed the label.)

spouliot:
> We ignored them in mac OS for a reason,

Yes, the API were not available cross-platform, so of limited use (unless someone provided a use case).

mandel-macaque:
@spouliot issues is, I have not found any sample from apple or any example on how to use the API. I an add it if you want, should not be too much work.

dalexsoto (approved):
I am fine ignoring them for now and if someone needs it we can always add it later

PR #11997, opened 2021-06-21
[CoreVideo] Add support for Xcode13 beta 1.

mandel-macaque:
Added new APIs from Xcode13, add missing ones that have been added in
catalyst and enabled nullable in the manual code.


src/CoreVideo/CVPixelBuffer.cs
```diff
@@ -84,6 +92,32 @@ public NSDictionary GetAttributes (NSDictionary [] attributes)
...
 		}
+
+#if NET
+		[SupportedOSPlatform ("ios15.0")]
+		[SupportedOSPlatform ("tvos15.0")]
+		[SupportedOSPlatform ("macos12.0")]
+		[SupportedOSPlatform ("maccatalyst15.0")] 
+		[UnsupportedOSPlatform ("watchos")]
+#else
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
```
rolfbjarne, line 103: This is missing `[MacCatalyst (15, 0]`

src/CoreVideo/CVTime.cs
```diff
@@ -43,15 +45,15 @@ public struct CVTime {
 		public int Flags { get { return (int) TimeFlags; } set { TimeFlags = (CVTimeFlags) value; }}
 
 #if !COREBUILD
-		public static CVTime ZeroTime {
+		public static CVTime? ZeroTime {
```
rolfbjarne, line 48: This is a breaking change, because `CVTime` is a struct, so it becomes `Nullable<CVTime>` if you make it nullable.

src/CoreVideo/CVPixelFormatType.cs
```diff
@@ -121,4 +127,33 @@ public enum CVPixelFormatType : uint {
...
+	public static class CVPixelFormatTypeExtensions {
+
+#if NET
+		[SupportedOSPlatform ("ios15.0")]
+		[SupportedOSPlatform ("tvos15.0")]
+		[SupportedOSPlatform ("macos12.0")]
+		[SupportedOSPlatform ("maccatalyst15.0")]
+		[UnsupportedOSPlatform ("watchos")]
+#else
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
```
rolfbjarne, line 141: Missing attribute for MacCatalyst.

src/CoreVideo/CVPixelFormatType.cs
```diff
@@ -121,4 +127,33 @@ public enum CVPixelFormatType : uint {
...
+		static extern bool CVIsCompressedPixelFormatAvailable (uint pixelFormatType);
+
+#if NET
+		[SupportedOSPlatform ("ios15.0")]
+		[SupportedOSPlatform ("tvos15.0")]
+		[SupportedOSPlatform ("macos12.0")]
+		[SupportedOSPlatform ("maccatalyst15.0")]
+		[UnsupportedOSPlatform ("watchos")]
+#else
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
```
rolfbjarne, line 153: Missing attribute for MacCatalyst.

src/CoreVideo/CVPixelBuffer.cs
```diff
@@ -147,7 +181,7 @@ public static CVPixelBuffer Create (nint width, nint height, CVPixelFormatType p
 
 		class PlaneData
 		{
-			public GCHandle[] dataHandles;
+			public GCHandle[]? dataHandles;
```
rolfbjarne, line 184: The `dataHandles` field is set to a non-null value as soon as a `PlaneData` instance is created, so logically it should always be non-null:

https://github.com/xamarin/xamarin-macios/blob/ae6577ed3c2e353b42ba67347cd6eb4b709cb73e/src/CoreVideo/CVPixelBuffer.cs#L263-L264

Reworking this so that the `dataHandles` field is set in a constructor would make it possible to make the field non-nullable (and avoid the other related changes below).

rolfbjarne (changes requested):

src/CoreVideo/CVTime.cs
mandel-macaque: This was not needed after the addition of '!' good catch.

tj-devel709 (approved):

rolfbjarne (approved):

src/CoreVideo/CVPixelBuffer.cs
```diff
@@ -84,6 +92,32 @@ public NSDictionary GetAttributes (NSDictionary [] attributes)
 				throw new ArgumentException (ret.ToString ());
 			return Runtime.GetNSObject<NSDictionary> (resolvedDictionaryOut);
 		}
+
+#if NET
+		[SupportedOSPlatform ("ios15.0")]
+		[SupportedOSPlatform ("tvos15.0")]
+		[SupportedOSPlatform ("macos12.0")]
+		[UnsupportedOSPlatform ("maccatalyst")] 
+		[UnsupportedOSPlatform ("watchos")]
```
spouliot, line 101: ^ for `NET` it's done at the assembly-level

src/CoreVideo/CVPixelBufferPool.cs
```diff
@@ -159,7 +161,7 @@ public CVPixelBufferPool (NSDictionary poolAttributes, NSDictionary pixelBufferA
 		}
 
 		public CVPixelBufferPool (CVPixelBufferPoolSettings settings, CVPixelBufferAttributes pixelBufferAttributes)
-			: this (settings.GetDictionary (), pixelBufferAttributes.GetDictionary ())
+			: this (settings.GetDictionary () ?? new NSDictionary (), pixelBufferAttributes.GetDictionary () ?? new NSDictionary ())
```
spouliot, line 164: why ?

`x.GetHandle` is an extension method that will return `IntPtr.Zero` if `x` is null

Also you're creating two instances of the same, immutable instance of `NSDictionary`.

if `CVPixelBufferPoolCreate` does not accept `null` then it's incorrect and should be using `GetNonNullHandle` extension method

src/CoreVideo/CVPixelFormatType.cs
```diff
@@ -121,4 +127,33 @@ public enum CVPixelFormatType : uint {
...
+#if NET
+		[SupportedOSPlatform ("ios15.0")]
+		[SupportedOSPlatform ("tvos15.0")]
+		[SupportedOSPlatform ("macos12.0")]
+		[UnsupportedOSPlatform ("maccatalyst")]
+		[UnsupportedOSPlatform ("watchos")]
+#else
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0), NoMacCatalyst]
+#endif
+		[DllImport (Constants.CoreVideoLibrary)]
```
spouliot, line 143: double check the native return type
you might need a `[return: MarshalAs (UnmanagedType.I1)]` to correctly convert the native `bool`

spouliot (changes requested):

src/CoreVideo/CVPixelBuffer.cs
mandel-macaque, line 101: Like this?

```suggestion
		[assembly: SupportedOSPlatform ("ios15.0")]
		[assembly: SupportedOSPlatform ("tvos15.0")]
		[assembly: SupportedOSPlatform ("macos12.0")]
		[assembly: UnsupportedOSPlatform ("maccatalyst")] 
```

spouliot: no, my comment was that `watchos` is already handled with an assembly-level attribute (no need to repeat it)

spouliot: IOW those were fine
```
 		[SupportedOSPlatform ("ios15.0")]
 		[SupportedOSPlatform ("tvos15.0")]
 		[SupportedOSPlatform ("macos12.0")]
 		[UnsupportedOSPlatform ("maccatalyst")] 
```

mandel-macaque: ahhh see, I misunderstood, got it

dalexsoto (approved):

spouliot (approved):

PR #19, opened 2016-05-03
Correct delegate signature on CNContactStoreEnumerateContactsHandler

sushihangover:
The signature on  `CNContactStoreEnumerateContactsHandler` should include a `ref`on the `bool stop` parameter in order to have the ability to early exit the enumeration.

Ref: http://stackoverflow.com/questions/36992982/early-exit-a-cncontactstoreenumeratecontactshandler-enumeration

PR #11998, opened 2021-06-21
[Accessibility] Update bindings for Xcode 13.0 beta 1

rachelkang:


src/Accessibility/AXHearingUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Runtime.InteropServices;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+
+namespace Accessibility {
+    public static partial class AXHearingUtilities {
+
+        [iOS (15,0), NoMac, NoTV, Watch (8,0)]
```
spouliot, line 21: since the bindings already have this (partial) type decorated with the attributes then they are not needed on the members (since they are identical and are inherited from the generated type-level declaration).

To ease readability you can add 
```csharp
// accessibility.cs already provide the following attributes on the type
// [iOS (15,0), NoMac, NoTV, Watch (8,0)]
```
above line #19.

src/accessibility.cs
```diff
@@ -1,9 +1,96 @@
...
+		IntPtr Constructor (string title, string[] categoryOrder);
+
+		[Export ("initWithAttributedTitle:categoryOrder:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSAttributedString attributedTitle, string[] categoryOrder);
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[Protocol]
+	[BaseType (typeof(NSObject))]
```
spouliot, line 27: no `[BaseType]` unless there's a `[Model]`

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+	{
+		[Export ("number", ArgumentSemantic.Assign)]
+		double Number { get; set; }
+
+		[Export ("category")]
+		string Category { get; set; }
+
+		[Static]
+		[Export ("valueWithNumber:")]
+		AXDataPointValue ValueWithNumber (double number);
```
spouliot, line 202: Use a name with a verb, like `Create`

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+
+		[Static]
+		[Export ("valueWithCategory:")]
+		AXDataPointValue ValueWithCategory (string category);
```
spouliot, line 206: same

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		Right = 1 << 2,
+		Both = Left | Right,
+	}
+
+	[Watch (8,0), NoTV, NoMac, iOS (15,0)]
+	[Static]
+	[Partial]
+	partial interface AXHearingUtilities
+	{
+		[Field ("AXMFiHearingDeviceStreamingEarDidChangeNotification")]
```
spouliot, line 251: missing `[Notification]` attribute

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		NSString StreamingEarDidChangeNotification { get; }
+
+		[Field ("AXMFiHearingDevicePairedUUIDsDidChangeNotification")]
```
spouliot, line 254: missing `[Notification]` attribute

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+	[BaseType (typeof(NSObject))]
+	interface AXLiveAudioGraph
+	{
+		[Static]
+		[Export ("start")]
+		void Start ();
+
+		[Static]
+		[Export ("updateValue:")]
+		void UpdateValue (double value);
```
spouliot, line 268: -> `Update` since `Value:` is the parameter (double check with Swift name when unsure)

_update: swift uses `updateValue` but I still think `Update` is better :)_

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+	{
+		[Field ("AXMFiHearingDeviceStreamingEarDidChangeNotification")]
+		NSString StreamingEarDidChangeNotification { get; }
+
+		[Field ("AXMFiHearingDevicePairedUUIDsDidChangeNotification")]
+		NSString PairedUUIDsDidChangeNotification { get; }
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[BaseType (typeof(NSObject))]
```
spouliot, line 259: since it's only `static` members you should add `[DisableDefaultCtor]` since creating an instance of this type is not helpful (and can be confusing)

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		void Stop ();
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[Native]
+	public enum AXNumericDataAxisDescriptorScale : long
+	{
+		linear = 0,
+		log10,
+		ln,
```
spouliot, line 281: Caps on the first letters!

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+
+		[Export ("valueDescriptionProvider", ArgumentSemantic.Copy)]
+		Func<double, NSString> ValueDescriptionProvider { get; set; }
+
+		[Export ("gridlinePositions", ArgumentSemantic.Copy)]
+		NSNumber[] GridlinePositions { get; set; }
+
+		[Export ("initWithTitle:lowerBound:upperBound:gridlinePositions:valueDescriptionProvider:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string title, double lowerbound, double upperBound, [NullAllowed] NSNumber[] gridlinePositions, Func<double, NSString> valueDescriptionProvider);
```
spouliot, line 306: camel-case `lowerBound` to match your `upperBound`

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+
+		[Export ("initWithAttributedTitle:lowerBound:upperBound:gridlinePositions:valueDescriptionProvider:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (NSAttributedString attributedTitle, double lowerbound, double upperBound, [NullAllowed] NSNumber[] gridlinePositions, Func<double, NSString> valueDescriptionProvider);
```
spouliot, line 310: same

spouliot (changes requested):
needs a few changes, mostly minor :)

src/Accessibility/AXHearingUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+        [DllImport (Constants.AccessibilityLibrary)]
+        static extern /* AXHearingDeviceEar */ nint AXMFiHearingDeviceStreamingEar ();
+
+        [iOS (15,0), NoMac, NoTV, Watch (8,0)]
+        [DllImport (Constants.AccessibilityLibrary, EntryPoint = "AXSupportsBidirectionalAXMFiHearingDeviceStreaming")]
+        [return: MarshalAs (UnmanagedType.I1)]
+        static extern bool SupportsBidirectionalStreaming ();
```
rolfbjarne, line 28: If you don't make it public, nobody will be able to use it!

```suggestion
        public static extern bool SupportsBidirectionalStreaming ();
```

src/Accessibility/AXHearingUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            return (AXHearingDeviceEar) (long) AXMFiHearingDeviceStreamingEar ();
+        }
+
+        [iOS (15,0), NoMac, NoTV, Watch (8,0)]
+        public static NSUuid[] GetMFiHearingDevicePairedUuids ()
+        {
+            return NSArray.ArrayFromHandle<NSUuid> (AXMFiHearingDevicePairedUUIDs ());
+        }
+    }
```
rolfbjarne, line 45: It looks like you've used spaces for indentation - we use tabs.

rolfbjarne (changes requested):
This looks good!

Just a few minor comments, and also the manual API you added (in AXHearingUtilities.cs) needs manual tests (in monotouch-test).

src/accessibility.cs
rachelkang: By "unless there's a [Model]", do you mean unless the class inherits from another class, or unless the class literally has the attribute `[Model]`? And how do we know when the tag is needed?

src/accessibility.cs
rachelkang: haha, sounds good!!! :D

src/accessibility.cs
rachelkang: good catch, thank you!

src/Accessibility/AXHearingUtilities.cs
rachelkang: oops, thank you!

src/accessibility.cs
mandel-macaque: He means the attribute, since Model will create an instance.

src/accessibility.cs
```diff
@@ -1,9 +1,96 @@
...
+		LeftToRight = 0,
+		RightToLeft,
+		TopToBottom,
+		BottomToTop,
+		RadialClockwise,
+		RadialCounterClockwise,
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[BaseType (typeof(NSObject))]
```
mandel-macaque, line 48: Same as what @spouliot mentioned.

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		[Export ("title")]
+		string Title { get; set; }
+
+		[Abstract]
+		[Export ("attributedTitle", ArgumentSemantic.Copy)]
+		NSAttributedString AttributedTitle { get; set; }
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[BaseType (typeof(NSObject))]
```
mandel-macaque, line 159: Same.

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		[Export ("initWithX:y:additionalValues:")]
+		IntPtr Constructor (AXDataPointValue xValue, [NullAllowed] AXDataPointValue yValue, [NullAllowed] AXDataPointValue[] additionalValues);
+
+		[Export ("initWithX:y:additionalValues:label:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (AXDataPointValue xValue, [NullAllowed] AXDataPointValue yValue, [NullAllowed] AXDataPointValue[] additionalValues, [NullAllowed] string label);
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[BaseType (typeof(NSObject))]
```
mandel-macaque, line 190: Same

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		[Export ("valueWithNumber:")]
+		AXDataPointValue ValueWithNumber (double number);
+
+		[Static]
+		[Export ("valueWithCategory:")]
+		AXDataPointValue ValueWithCategory (string category);
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+	[BaseType (typeof(NSObject))]
```
mandel-macaque, line 210: Same

src/accessibility.cs
dalexsoto: The `[Model]` attribute, you only add `[Model]`, `[BaseType]` and `[Protocol]` to types with a suffix `Delegate` or `DataSource` so `FooDelegate` and `FooDataSource` will need the 3 of them any other `@protocol` just needs the `[Protocol]` attribute.

src/accessibility.cs
mandel-macaque: Do we need BaseType?

src/accessibility.cs
```diff
@@ -53,4 +140,173 @@ interface AXCustomContentProvider
...
+		AXNumericDataAxisDescriptorScale ScaleType { get; set; }
+
+		[Export ("lowerBound")]
+		double LowerBound { get; set; }
+
+		[Export ("upperBound")]
+		double UpperBound { get; set; }
+
+		[Export ("valueDescriptionProvider", ArgumentSemantic.Copy)]
+		Func<double, NSString> ValueDescriptionProvider { get; set; }
```
mandel-macaque, line 299: Maybe a delegate will be more helpful for the developers to understand what is going on.

mandel-macaque (changes requested):

src/accessibility.cs
rachelkang: gotcha, thanks for clarifying!

rachelkang: Should I also remove the attribute from existing types in the file, or just let them be?

tj-devel709: I may be wrong, but I think we try to leave things already there alone unless it causes an issue

dalexsoto: Do not remove existing attributes since it will be a breaking change

src/accessibility.cs
dalexsoto: Nope, this is an actual object `@interface AXChartDescriptor : NSObject` not a delegate so BaseType is fine here

src/Accessibility/AXHearingUtilities.cs
spouliot: Avoid marking comments as "resolved". It makes it hard for the reviewer to see if/how the comment was resolved.

It's better to ket the commenter mark it as "resolved" after it's been validated.

src/accessibility.cs
rachelkang: If I'm understanding correctly, would that be something like this? 

```cs
delegate double AXDataValue;
delegate NSString AXDataValueDescription;

[Export ("valueDescriptionProvider", ArgumentSemantic.Copy)]
Func<AXDataValue, AXDataValueDescription> ValueDescriptionProvider { get; set; }
```

if not, what were you thinking?

src/Accessibility/AXHearingUtilities.cs
rachelkang: sounds good, sorry about that!

src/accessibility.cs
spouliot: No, more replacing `Func` with a `delegate` type. The later are self-documenting. 

The more arguments the more important it becomes because `Func<byte,short,int,long>` has to meaning (without reading the docs). However a delegate type like `delegate long Hash (byte b, short s, int i);` describe what the function is expected to do.

spouliot (approved):
pending requested tests and green / no related failures from bots

src/accessibility.cs
mandel-macaque: Something like:
```shaper
delegate AXDataValueDescription ValueDescriptionProviderHandler (AXDataValue dataValue);
```
To be honest, the headers do not help much.

mandel-macaque (approved):

tests/monotouch-test/Accessibility/AXHearingUtilitiesTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+#if !__TVOS__ && !MONOMAC
+
+using System;
+using ObjCRuntime;
+
+using Foundation;
+using Accessibility;
+using NUnit.Framework;
+using MonoTests.System.Net.Http;
```
rolfbjarne, line 16: Why do you need this `using`?

src/Accessibility/AXHearingUtilities.cs
rolfbjarne: This is still the case :smile:

rolfbjarne (changes requested):

src/Accessibility/AXHearingUtilities.cs
rachelkang: :( so strange... I'll check through everything again!

tests/monotouch-test/Accessibility/AXHearingUtilitiesTests.cs
rachelkang: I don't :p poor copy/paste! removing

rolfbjarne (approved):

dalexsoto (changes requested):
Ouch looks like you will need to fix your PR :(

rachelkang:
@dalexsoto yes :/ there was a bad merge - @mandel-macaque is helping me fix it though 😅

rachelkang:
intro and xtro are all greean and happy :)

tj-devel709:
The one failing test seems to be unrelated: https://github.com/xamarin/maccore/issues/2434

PR #12001, opened 2021-06-22
[dotnet][linker] Remove unused backing fields

spouliot:
Problems:

* `Dispose` set the generated backing fields to `null` which means
  the linker will mark every backing fields, even if not used
  elsewhere (generally properties) inside the class

* Backing fields increase the memory footprint of the managed peer
  instance (for the type and all it's subclasses)

* Backing fields also increase the app size. Not a huge problem as
  they are all declared _weakly_ as `NSObject` but still...

Solution:

* When the linker process a `Dispose` method of an `NSObject`
  subclass with the _optimizable_ attribute then we remove the
  method body. This way the linker cannot mark the fields.

* Before saving back the assemblies we replace the cached method
  body and NOP every field that were not marked by something else
  than the `Dispose` method.

```diff
--- a.cs	2021-06-22 16:56:57.000000000 -0400
+++ b.cs	2021-06-22 16:57:00.000000000 -0400
@@ -3107,8 +3107,6 @@

 		private static readonly IntPtr class_ptr = Class.GetHandle("UIApplication");

-		private object __mt_WeakDelegate_var;
-
 		public override IntPtr ClassHandle => class_ptr;

 		[DllImport("__Internal")]
@@ -3141,9 +3139,8 @@
 		protected override void Dispose(bool P_0)
 		{
 			base.Dispose(P_0);
-			if (base.Handle == IntPtr.Zero)
+			if (!(base.Handle == IntPtr.Zero))
 			{
-				__mt_WeakDelegate_var = null;
 			}
 		}
 	}
@@ -3209,10 +3206,6 @@
 	{
 		private static readonly IntPtr class_ptr = Class.GetHandle("UIScreen");

-		private object __mt_FocusedItem_var;
-
-		private object __mt_FocusedView_var;
-
 		public override IntPtr ClassHandle => class_ptr;

 		public virtual CGRect Bounds
@@ -3242,10 +3235,8 @@
 		protected override void Dispose(bool P_0)
 		{
 			base.Dispose(P_0);
-			if (base.Handle == IntPtr.Zero)
+			if (!(base.Handle == IntPtr.Zero))
 			{
-				__mt_FocusedItem_var = null;
-				__mt_FocusedView_var = null;
 			}
 		}
 	}
@@ -3254,10 +3245,6 @@
 	{
 		private static readonly IntPtr class_ptr = Class.GetHandle("UIView");

-		private object __mt_ParentFocusEnvironment_var;
-
-		private object __mt_PreferredFocusedView_var;
-
 		public override IntPtr ClassHandle => class_ptr;

 		public virtual CGRect Bounds
@@ -3303,10 +3290,8 @@
 		protected override void Dispose(bool P_0)
 		{
 			base.Dispose(P_0);
-			if (base.Handle == IntPtr.Zero)
+			if (!(base.Handle == IntPtr.Zero))
 			{
-				__mt_ParentFocusEnvironment_var = null;
-				__mt_PreferredFocusedView_var = null;
 			}
 		}
 	}
@@ -3315,12 +3300,6 @@
 	{
 		private static readonly IntPtr class_ptr = Class.GetHandle("UIViewController");

-		private object __mt_ParentFocusEnvironment_var;
-
-		private object __mt_PreferredFocusedView_var;
-
-		private object __mt_WeakTransitioningDelegate_var;
-
 		public override IntPtr ClassHandle => class_ptr;

 		public virtual UIView View
@@ -3363,11 +3342,8 @@
 		protected override void Dispose(bool P_0)
 		{
 			base.Dispose(P_0);
-			if (base.Handle == IntPtr.Zero)
+			if (!(base.Handle == IntPtr.Zero))
 			{
-				__mt_ParentFocusEnvironment_var = null;
-				__mt_PreferredFocusedView_var = null;
-				__mt_WeakTransitioningDelegate_var = null;
 			}
 		}
 	}
@@ -3376,8 +3352,6 @@
 	{
 		private static readonly IntPtr class_ptr = Class.GetHandle("UIWindow");

-		private object __mt_WindowScene_var;
-
 		public override IntPtr ClassHandle => class_ptr;

 		public virtual UIViewController RootViewController
@@ -3411,9 +3385,8 @@
 		protected override void Dispose(bool P_0)
 		{
 			base.Dispose(P_0);
-			if (base.Handle == IntPtr.Zero)
+			if (!(base.Handle == IntPtr.Zero))
 			{
-				__mt_WindowScene_var = null;
 			}
 		}
 	}
```


spouliot:
> Workload installation failed: One or more errors occurred. (Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.)

mandel-macaque:
> Workload installation failed: One or more errors occurred. (Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.)

Ein???

tools/dotnet-linker/BackingFieldDelayHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+			il.Emit (OpCodes.Ret);
+			method.Body = body;
+		}
+
+		public static void ReapplyDisposedFields (DerivedLinkContext context)
+		{
+			foreach (var kvp in dispose) {
+				var method = kvp.Key;
+				// note: all methods in the dictionary are marked
+				var body = kvp.Value;
```
mandel-macaque, line 71: Lazy Manuel likes: 
```
foreach ((int key, string value) in dictionary)
{
    // things here 
}
```

mandel-macaque (approved):
Use the deconstructed if you like :)

dalexsoto (approved):

tools/dotnet-linker/BackingFieldDelayHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+	// * When the linker process a `Dispose` method of an `NSObject` 
+	//   subclass with the _optimizable_ attribute then we remove the
+	//   method body. This way the linker cannot mark the fields.
+	// * Before saving back the assemblies we replace the cached method
+	//   body and NOP every field that were not marked by something else
+	//   than the `Dispose` method.
+	public class BackingFieldDelayHandler : ConfigurationAwareMarkHandler {
+
+		protected override string Name { get; } = "Backing Fields Optimizer";
+		protected override int ErrorCode { get; } = 2200;
```
rolfbjarne, line 30: 2200 is taken: https://github.com/xamarin/xamarin-macios/blob/d11852e02c3d51be71b6989e0eacb07e945b0e82/tools/linker/MonoTouch.Tuner/PreserveSmartEnumConversions.cs#L28

tools/dotnet-linker/BackingFieldDelayHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+				var body = kvp.Value;
+				foreach (var ins in body.Instructions) {
+					switch (ins.OpCode.OperandType) {
+					// case OperandType.InlineTok:
+					case OperandType.InlineField:
+						var field = (ins.Operand as FieldReference)?.Resolve ();
+						if (!context.Annotations.IsMarked (field)) {
+							ins.OpCode = OpCodes.Nop; // stfld
+							ins.Previous.OpCode = OpCodes.Nop; // ldnull
+							ins.Previous.Previous.OpCode = OpCodes.Nop; // ldarg.0
```
rolfbjarne, line 80: You're assuming that the instructions are `ldarg.0 + ldnull + stdfld` here (which sounds like a reasonable assumption today), but that might not be true in the future, where a new C# compiler might end up with a different sequence. Maybe verify this assumption, and if not show an error/warning?

```csharp
if (ins == stdfld && ins.Previous == ldnull && ins.Previous.Previous == ldarg.0) {
    // yay!
} else {
    // nay?
}
```

rolfbjarne (changes requested):

tools/dotnet-linker/BackingFieldDelayHandler.cs
spouliot: oops, forgot to update it

rolfbjarne (approved):

PR #12003, opened 2021-06-23
[corefoundation] Update `CFArray` and make it public

spouliot:
* Make `CFArray` public so it can be used for bindings, including 3rd party bindings
* Subclass `NativeObject` to remove some boilerplate code
* Most members are `internal` until needed
* Add nullability annotations

The next step is to make the same kind of optimization done on `NSString`
vs `CFString` by preferring faster p/invokes over calling selectors while
keeping generated code identical.


mandel-macaque (approved):

src/CoreFoundation/CFArray.cs
```diff
@@ -119,13 +91,13 @@ public static CFArray FromNativeObjects (params INativeObject[] values)
 
 		public IntPtr GetValue (nint index)
 		{
-			return CFArrayGetValueAtIndex (handle, index);
+			return CFArrayGetValueAtIndex (GetCheckedHandle (), index);
 		}
 
-		public static unsafe IntPtr Create (params IntPtr[] values)
+		internal static unsafe IntPtr Create (params IntPtr[] values)
```
rolfbjarne, line 97: Maybe rename methods, so that the public functions are:

```csharp
public static CFArray Create (params IntPtr[] values);
public static CFArray Create (params INativeObject[] values);
```

and then for the internal ones that return IntPtr, something like this (which doesn't matter as much):

```csharp
static IntPtr CreateHandle (params IntPtr[] values);
static IntPtr CreateHandle (params INativeObject[] values);
```

that way we follow our naming guidelines.

src/CoreFoundation/CFArray.cs
```diff
@@ -119,13 +91,13 @@ public static CFArray FromNativeObjects (params INativeObject[] values)
...
 		{
-			if (values == null)
-				throw new ArgumentNullException ("values");
+			if (values is null)
+				ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (values));
 			fixed (IntPtr* pv = values) {
```
rolfbjarne, line 101: Not a big deal at all, but you can change the P/Invoke to take an `IntPtr[]`, and avoid the fixed/unsafe logic.

src/CoreFoundation/CFArray.cs
```diff
@@ -38,54 +38,26 @@
...
-			if (!owns)
-				CFObject.CFRetain (handle);
 		}
 		
-		public IntPtr Handle {
-			get {return handle;}
-		}
-
 		[DllImport (Constants.CoreFoundationLibrary, EntryPoint="CFArrayGetTypeID")]
 		public extern static /* CFTypeID */ nint GetTypeID ();
```
rolfbjarne, line 59: Do we want to expose this publicly?

rolfbjarne (changes requested):

src/CoreFoundation/CFArray.cs
spouliot: We expose them for other types... but they are not very useful.
I'll change it to `internal` like most others since it's how it effectively is today (without causing any harm)

src/CoreFoundation/CFArray.cs
spouliot: That's should be measured first [1] (even more since we have some tooling now) and out-of-scope of the PR.

[1] I don't think it matters but, for performance, what I think does not matter 😄

src/CoreFoundation/CFArray.cs
spouliot: Yes, but only if/when they are made `public`. Right now they remains `internal` (at member instead of type-level)

rolfbjarne: I'm sorry I wasn't clear, I was referring to renaming these methods to `Create`:

https://github.com/xamarin/xamarin-macios/blob/f26f6c37849b4a8f843f979aac217c8d4f02d6f0/src/CoreFoundation/CFArray.cs#L72-L80

spouliot: hmm... I should have made them `internal` for this first step. In part because I did not want, nor need, to modify too many other files.

The other part is that `NSArray` already expose both `FromIntPtrs` and `FromNativeObjects`... but I that does not mean I like their names :) We can certainly decide to renamed when obsoleting the existing `NSArray` versions.

rolfbjarne (approved):

PR #12014, opened 2021-06-24
[GameController] Updates for Xcode13 beta1

tj-devel709:


src/gamecontroller.cs
```diff
@@ -1922,4 +1996,50 @@ enum GCInputDirectional {
...
+		void Connect ([NullAllowed] Action<NSError> reply);
+
+		[Export ("disconnect")]
+		void Disconnect ();
+
+		[NullAllowed, Export ("controller", ArgumentSemantic.Weak)]
+		GCController Controller { get; }
+
+		[Export ("changeElement:configuration:")]
+		void Change (string element, GCVirtualControllerElementUpdateBlock config);
```
rolfbjarne, line 2021: This is minor, but we try to not use abbreviations for parameter names:

```suggestion
		void Change (string element, GCVirtualControllerElementUpdateBlock configuration);
```

src/gamecontroller.cs
```diff
@@ -1922,4 +1996,50 @@ enum GCInputDirectional {
 		[Field ("GCInputDirectionalCardinalDpad")]
 		CardinalDpad,
 	}
+
+	delegate GCVirtualControllerElementConfiguration GCVirtualControllerElementUpdateBlock (GCVirtualControllerElementConfiguration configuration);
+
+	[NoTV, NoMac, NoWatch, iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
```
rolfbjarne, line 2003: Really minor, missing space:

```suggestion
	[BaseType (typeof (NSObject))]
```

rolfbjarne (approved):

src/gamecontroller.cs
tj-devel709: Good catch!

src/gamecontroller.cs
tj-devel709: Ah got it!

src/gamecontroller.cs
```diff
@@ -1922,4 +1996,50 @@ enum GCInputDirectional {
 		[Field ("GCInputDirectionalCardinalDpad")]
 		CardinalDpad,
 	}
+
+	delegate GCVirtualControllerElementConfiguration GCVirtualControllerElementUpdateBlock (GCVirtualControllerElementConfiguration configuration);
+
+	[NoTV, NoMac, NoWatch, iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
```
tj-devel709, line 2004: @rolfbjarne I know you already reviewed this PR but I just wanted to make sure that this line is correct. Also for others reviewing later:
I added this since there didn't appear to be an initializer for this interface and we have the `virtualControllerWithConfiguration:` that creates an instance

rolfbjarne: Yes, this looks entirely correct!

Also, if in doubt, it's better to put this attribute than to not to, because we can always add the default ctor later (but we can't remove it once we've made it a part of the stable API).

src/gamecontroller.cs
```diff
@@ -1922,4 +1996,50 @@ enum GCInputDirectional {
...
+	interface GCVirtualController
+	{
+		[Static]
+		[Export ("virtualControllerWithConfiguration:")]
+		GCVirtualController CreateVirtualController (GCVirtualControllerConfiguration configuration);
```
mandel-macaque, line 2009: Minor, we use the pattern *Create* in all other APIs, can you match it?

src/gamecontroller.cs
```diff
@@ -1922,4 +1996,50 @@ enum GCInputDirectional {
...
+
+		[Export ("connectWithReply:")]
+		void Connect ([NullAllowed] Action<NSError> reply);
```
mandel-macaque, line 2012: Could this be made async?

mandel-macaque (changes requested):

tj-devel709:
@mandel-macaque Updated!

mandel-macaque (approved):

PR #12016, opened 2021-06-24
[dotnet] use copyOnly for .png files in templates

jonathanpeppers:
Context: https://github.com/dotnet/templating/issues/3325
Context: https://github.com/dotnet/templating/wiki/Reference-for-template.json#content-manipulation

In current .NET 6 Preview 6 builds, there is an issue if a template
includes a binary file larger than ~8kb, it seems to get truncated when
`dotnet new` extracts the template.

A workaround is to use the `copyOnly` feature for binary files. Really,
we should be doing this anyway, because otherwise the templating system
considers replacing *text* in these binary files. It improves
performance to do this and would hopefully prevent a future bug of
random bytes getting replaced.

### DIFF
--- a/dotnet/Microsoft.MacCatalyst.Templates/maccatalyst/.template.config/template.json
+++ b/dotnet/Microsoft.MacCatalyst.Templates/maccatalyst/.template.config/template.json
@@ -11,6 +11,13 @@
     "type": "project"
   },
   "sourceName": "MacCatalystApp1",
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
We'll probably want to also backport this to Preview 6.

rolfbjarne (approved):

rolfbjarne:

PR #12022, opened 2021-06-25
[EventKit] Updates for Xcode13 Beta1

tj-devel709:
I was a little confused about this Frameworks MacCatalyst's backlog.

Possibly just an output
* !missing-enum! EKAlarmType not bound
    * This one is contained inside a manual binding and I do not really see other availability attributes
        * https://github.com/xamarin/xamarin-macios/blob/2b7dc07d20fc06344d7f164af8934f5e755572c2/src/EventKit/EKEnums.cs#L180
    * Also the only place I see it used is here: https://github.com/xamarin/xamarin-macios/blob/9a0cd6182b7086b238434ea5f857d65f5559fe2d/src/eventkit.cs#L204
        * So I feel that it is used as an output only and therefore does not need attributes

This group below belongs to EKAlarm which the web docs say support MacCatalyst 13.0 but the selectors themselves do not say they support MacCatalyst in the webdocs
* !missing-selector! EKAlarm::emailAddress not bound
* !missing-selector! EKAlarm::setEmailAddress: not bound
* !missing-selector! EKAlarm::setSoundName: not bound
* !missing-selector! EKAlarm::setUrl: not bound
* !missing-selector! EKAlarm::soundName not bound
* !missing-selector! EKAlarm::type not bound
* !missing-selector! EKAlarm::url not bound

These two are deprecated
* !missing-selector! EKEvent::birthdayPersonUniqueID not bound
* !missing-selector! EKEventStore::initWithAccessToEntityTypes: not bound

Similar to the EKAlarm selectors, EKEventStore says that it supports MacCatalyst but initWithSources does not say it supports MacCatalyst
* !missing-selector! EKEventStore::initWithSources: not bound


src/eventkit.cs
```diff
@@ -611,6 +616,7 @@ interface EKEventStore {
 #if MONOMAC
 		[Deprecated (PlatformName.MacOSX, 10, 9)]
 		[Export ("initWithAccessToEntityTypes:")]
+		[NoMacCatalyst]
```
mandel-macaque, line 619: @rolfbjarne  @spouliot @tj-devel709  if that NoMacCatalust is inside the MONOMAC, does it make any effect? Should it be with the other attrs?

src/eventkit.cs
```diff
@@ -657,4 +663,68 @@ interface EKReminder {
...
+		string ConferenceDetails { get; }
+	}
+
+	[Mac (12,0), iOS (15,0), Watch (8,0), MacCatalyst (15,0), NoTV]
+	[BaseType (typeof (NSObject))]
+	interface EKVirtualConferenceProvider : NSExtensionRequestHandling
+	{
+		[Async]
+		[Export ("fetchAvailableRoomTypesWithCompletionHandler:")]
+		void FetchAvailableRoomTypes (Action<NSArray<EKVirtualConferenceRoomTypeDescriptor>, NSError> completionHandler);
```
mandel-macaque, line 692: Can we move the Action to a delegate so that is better documented for the users? delegates can take names for the attrs.

src/eventkit.cs
```diff
@@ -657,4 +663,68 @@ interface EKReminder {
...
+
+		[Async]
+		[Export ("fetchVirtualConferenceForIdentifier:completionHandler:")]
+		void FetchVirtualConference (string identifier, Action<EKVirtualConferenceDescriptor, NSError> completionHandler);
```
mandel-macaque, line 696: Same.

mandel-macaque (changes requested):
Small fix, also a question regarding the attr.

tj-devel709:
@mandel-macaque I made an attempt with the delegate, let me know if it is sufficient haha

src/eventkit.cs
```diff
@@ -683,17 +683,19 @@ interface EKVirtualConferenceDescriptor
 		string ConferenceDetails { get; }
 	}
 
+	delegate void VirtualConferenceHandler (EKVirtualConferenceDescriptor virtualConferenceDescriptor, NSError error);
```
mandel-macaque, line 686: This is the reason I asked for it, there should be to diff ones, if you look at:
```csharp
void FetchAvailableRoomTypes (Action<NSArray<EKVirtualConferenceRoomTypeDescriptor>, NSError> completionHandler);
void FetchVirtualConference (string identifier, Action<EKVirtualConferenceDescriptor, NSError> completionHandler);
```

The parameters are diff, you need to create two diff ones. one for FetchAvailableRoomTypes with EKVirtualConferenceRoomTypeDescriptor as a first param and a second one for FetchVirtualConference with EKVirtualConferenceDescriptor as the first param. 
```
EKVirtualConferenceRoomTypeDescriptor != EKVirtualConferenceDescriptor
```

mandel-macaque (changes requested):

src/eventkit.cs
tj-devel709: Ah good catch, I completely missed that!

src/eventkit.cs
rolfbjarne: No, it doesn't not have effect inside an `#if MONOMAC`

rolfbjarne (approved):

src/eventkit.cs
mandel-macaque: @tj-devel709  that means that the attr has to be outside the MONOMAC ;) (I suspected so)

src/eventkit.cs
```diff
@@ -200,17 +200,21 @@ interface EKAlarm : NSCopying {
 		EKAlarmProximity Proximity { get; set;  }
 
 #if MONOMAC
+		[NoMacCatalyst]
```
rolfbjarne, line 203: There are numerous `[NoMacCatalyst]` attributes that are inside `#if MONOMAC`, they're all redundant.

rolfbjarne (approved):

mandel-macaque (approved):
Approving BUT the redundant attributes should be removed.

src/eventkit.cs
```diff
@@ -657,4 +657,71 @@ interface EKReminder {
...
+		string Title { get; }
+
+		[Export ("identifier")]
+		string Identifier { get; }
+	}
+
+	[Mac (12,0), iOS (15,0), Watch (8,0), MacCatalyst (15,0), NoTV]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface EKVirtualConferenceURLDescriptor
```
dalexsoto, line 715:
```suggestion
	[Mac (12,0), iOS (15,0), Watch (8,0), MacCatalyst (15,0), NoTV]
	[BaseType (typeof (NSObject), Name = "EKVirtualConferenceURLDescriptor")]
	[DisableDefaultCtor]
	interface EKVirtualConferenceUrlDescriptor
```

dalexsoto: `URL` => `Url` 3 letters acronym

src/eventkit.cs
```diff
@@ -657,4 +657,71 @@ interface EKReminder {
...
 	}
+
+	[Mac (12,0), iOS (15,0), Watch (8,0), MacCatalyst (15,0), NoTV]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface EKVirtualConferenceDescriptor
+	{
+		[Export ("initWithTitle:URLDescriptors:conferenceDetails:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([NullAllowed] string title, EKVirtualConferenceURLDescriptor[] urlDescriptors, [NullAllowed] string conferenceDetails);
```
dalexsoto, line 668:
```suggestion
		IntPtr Constructor ([NullAllowed] string title, EKVirtualConferenceUrlDescriptor[] urlDescriptors, [NullAllowed] string conferenceDetails);
```

dalexsoto (changes requested):
Small issue

tj-devel709:
Unrelated test failure: https://github.com/xamarin/maccore/issues/2443

PR #12026, opened 2021-06-25
[ReplayKit] Updates for Xcode13 Beta 1

tj-devel709:


src/replaykit.cs
```diff
@@ -136,6 +136,21 @@ interface RPScreenRecorder {
 		[Async]
 		[Export ("stopRecordingWithOutputURL:completionHandler:")]
 		void StopRecording (NSUrl url, [NullAllowed] Action<NSError> completionHandler);
+
+		[Async]
+		[NoTV, NoWatch, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Export ("exportClipToURL:duration:completionHandler:")]
+		void ExportClip (NSUrl url, double duration, [NullAllowed] Action<NSError> completionHandler);
```
tj-devel709, line 143: I'm hesitant about changing this ones name from ExportClipToUrl to ExportClip. 
NSUrl is the first argument, but does this name change make it more confusing to use?

src/ReplayKit/RPEnums.cs
```diff
@@ -53,6 +53,7 @@ public enum RPRecordingError : long {
 		FailedToStartCaptureStack = -5833,
 		InvalidParameter = -5834,
 		FilePermissions = -5835,
+		ExportClipToURLInProgress = -5836,
```
rolfbjarne, line 56:
```suggestion
		ExportClipToUrlInProgress = -5836,
```

src/replaykit.cs
rolfbjarne: When in doubt, you can check the Swift naming, in this case it's `exportClip` (https://developer.apple.com/documentation/replaykit/rpscreenrecorder/3752130-exportclip), so I think this is totally fine.

rolfbjarne (changes requested):

mandel-macaque (approved):

src/replaykit.cs
tj-devel709: Makes sense! Thanks!

rolfbjarne (approved):

dalexsoto (approved):

PR #12032, opened 2021-06-28
[CI] Changing Cron Job's Occurrence

tj-devel709:
This change will trigger the scheduled build at 1800 UTC (8:00 pm in Spain, 2 pm EST) on every day-of-week from Monday through Friday and also on Sunday. This will give me more chances to debug what is going on with the OneLocBuild inputs!


mandel-macaque (approved):

mandel-macaque:
Merging, else we will miss the first cycle.

PR #12035, opened 2021-06-29
[dotnet] shorten workload ids

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/6045

We are aiming to simplify the .NET workload ids to short names like:

    dotnet workload install maui
    dotnet workload install android
    dotnet workload install ios
    dotnet workload install maccatalyst
    dotnet workload install macos
    dotnet workload install tvos

Updated the workload ids here, as well as the `.stamp-install-workloads`
make target.

Other changes:

* Updated descriptions to match Android. Still waiting on an *official*
description we should be using for each workload.

* Fixed trailing `,` that showed up as JSON errors in VS Code.


jonathanpeppers:
FYI this is for main/preview 7 only.

rolfbjarne (approved):

mandel-macaque (approved):

PR #12038, opened 2021-06-29
[iTunesLibrary] Update bindings for Xcode 13.0 beta 1

rachelkang:


src/ituneslibrary.cs
```diff
@@ -327,6 +328,11 @@ interface ITLibPlaylist
 
 		[Export ("kind", ArgumentSemantic.Assign)]
 		ITLibPlaylistKind Kind { get; }
+
+		[Introduced (PlatformName.MacOSX, 10, 6, message: "Use 'isPrimary' instead.")]
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'isPrimary' instead.")]
+		[Export ("master")]
+		bool Master { [Bind ("isMaster")] get; }
```
rolfbjarne, line 335: I see that you've moved the `Master` declaration here from above - but don't move code around unless really necessary, because it shows up as "removed+added" in the diff, which makes the diff harder to review.

On a more general note, changes that don't really change anything also makes it harder to run "git blame" to figure out when/what something happened, because the first thing that will come up is the commit that moved the code, which is not what you're looking for.

rolfbjarne (changes requested):

src/ituneslibrary.cs
mandel-macaque: To add to that, if we where working with a xcode13 branch, it would make my life harder when doing merges and back ports :/

mandel-macaque (approved):
Approving once the master move is undone.

src/ituneslibrary.cs
rachelkang: understood! fixed :)

rolfbjarne (approved):
👍

tj-devel709 (approved):

src/ituneslibrary.cs
```diff
@@ -307,9 +307,15 @@ interface ITLibPlaylist
 		[Export ("name")]
 		string Name { get; }
 
+		[Introduced (PlatformName.MacOSX, 10, 6, message: "Use 'isPrimary' instead.")]
```
dalexsoto, line 310:
```suggestion
		[Introduced (PlatformName.MacOSX, 10, 6, message: "Use 'Primary' instead.")]
```

src/ituneslibrary.cs
```diff
@@ -307,9 +307,15 @@ interface ITLibPlaylist
...
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'isPrimary' instead.")]
```
dalexsoto, line 311:
```suggestion
		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'Primary' instead.")]
```

dalexsoto (approved):
After these two nits

dalexsoto:
We use the .NET name in the message strings, this is the reason behind the change `isPrimary` to `Primary `

src/iTunesLibrary/Enums.cs
```diff
@@ -301,10 +301,15 @@ public enum ITLibPlaylistProperty {
 		AllItemsPlaylist,
 		[Field ("ITLibPlaylistPropertyDistinguisedKind")]
 		DistinguisedKind,
+		[Introduced (PlatformName.MacOSX, 10, 6, message: "Use 'ITLibPlaylistPropertyPrimary' instead.")]
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'ITLibPlaylistPropertyPrimary' instead.")]
```
dalexsoto, line 305: This one is also not needed

```suggestion
		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'ITLibPlaylistPropertyPrimary' instead.")]
```

rachelkang:
intro and xtro are all green!

tj-devel709:
This one looks good to "Squash and merge"!

rachelkang:
Test failure unrelated - https://github.com/xamarin/maccore/issues/2443

PR #12050, opened 2021-06-30
[builds] Use the dotnet-install.sh script to install .NET locally.

rolfbjarne:
Hopefully this will solve issues with random curl errors (56) that happens
when trying to download .NET (and which break the build randomly).

### DIFF
--- a/builds/.gitignore
+++ b/builds/.gitignore
@@ -1,3 +1,4 @@
+dotnet-install.sh
 BundledNETCorePlatformsPackageVersion.txt
 downloads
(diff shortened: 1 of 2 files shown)


mandel-macaque (approved):

dalexsoto (approved):
❤️

mandel-macaque:
Unrelated failure: https://github.com/xamarin/maccore/issues/2414

mandel-macaque:
Landing so that we get less issues in CI.

PR #12068, opened 2021-07-01
[CoreAudioTypes] Updates for Xcode13 Beta2

tj-devel709:


mandel-macaque (approved):

src/AudioToolbox/AudioType.cs
```diff
@@ -470,6 +470,8 @@ public enum AudioChannelLabel : int { // UInt32 AudioChannelLabel
 		HoaAcn14               = (2 << 16) | 14,
 		HoaAcn15               = (2 << 16) | 15,
 		HoaAcn65024            = (2 << 16) | 65024,
+		HoaSn3d	               = HoaAcn0,
```
rachelkang, line 473: Why is it `HoaAcn0` and not `kAudioChannelLabel_HOA_ACN_0` ?
I also can't find these on the [web docs](https://developer.apple.com/documentation/coreaudiotypes/1572101-audio_channel_layout_tags?language=objc), hmm

tj-devel709: Yeah the full name is kAudioChannelLabel_HOA_ACN_0 but the enum elements that were already here were shortened to things like `HoaAcn0` I guess just for easier usability!

PR #12075, opened 2021-07-01
[CI] Adding GithubActions[bot] to Bypass Label Verification

tj-devel709:
Although the github actions bot successfully automatically adds 2 labels to this PR, the PR check github action did not accept it. When I added a new label (or perhaps removed one and added it back) and then ran that check, it passed.

I think somewhere along the formation of the PR, the check is missing it. I propose adding github actions bot to the ignore labels portion


mandel-macaque:
Should not be required, but nothing bad to add it.

PR #12081, opened 2021-07-06
[MapKit] Update bindings for Xcode 13.0 beta 1 and beta 2

rachelkang:
Catalyst-related edits were necessary to make xtro happy


tests/xtro-sharpie/MacCatalyst-MapKit.ignore
```diff
@@ -0,0 +1,11 @@
...
+### MKOverlayRenderer.h
+### MKOverlayView.h (based on UIView so not in OSX headers list)
+!missing-pinvoke! MKRoadWidthAtZoomScale is not bound
+
+
+## unsorted
+
+!missing-field! MKAnnotationCalloutInfoDidChangeNotification not bound
+!missing-protocol-conformance! MKUserLocation should conform to MKAnnotation
+!missing-protocol-member! MKOverlay::coordinate not found
```
rachelkang, line 11: copied iOS-MapKit.ignore file contents - all are relevant to MacCatalyst as well (wrt .todo file and xtro)

tj-devel709, line 11: Why is this one ignored?

src/mapkit.cs
```diff
@@ -769,12 +769,15 @@ interface MKMapViewDelegate {
 		[Export ("mapViewDidChangeVisibleRegion:")]
 		void DidChangeVisibleRegion (MKMapView mapView);
 	}
-		
+
```
tj-devel709, line 772: Small, but i would revert changes like this to keep the blame more neat

tests/xtro-sharpie/MacCatalyst-MapKit.ignore
rachelkang: I'm not sure, but I just copied the contents of the ios ignore file (It fails on xtro otherwise)

src/mapkit.cs
rachelkang: thanks TJ! I'm not sure how to do this now, but will be sure to look out for these in the future

mandel-macaque (approved):

src/mapkit.cs
mandel-macaque: Will land it because I don't want to wait for a build, but you can do something like this:
```bash
git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero -
```

mandel-macaque (approved):

PR #12086, opened 2021-07-07
[Speech] Update bindings for XCode 13.0 beta 1

rachelkang:


tests/xtro-sharpie/iOS-Speech.todo
```diff
@@ -1,2 +0,0 @@
-!deprecated-attribute-missing! SFSpeechRecognitionRequest::interactionIdentifier missing a [Deprecated] attribute
-!deprecated-attribute-missing! SFSpeechRecognitionRequest::setInteractionIdentifier: missing a [Deprecated] attribute
```
tj-devel709, line 2: I think you should be able to delete this file if there is not anything else in there!

tj-devel709 (approved):

mandel-macaque (changes requested):
As @tj-devel709 mentioned, empty files should be removed from xtro.

tests/xtro-sharpie/iOS-Speech.todo
rachelkang: I had no idea - thanks for catching this!

mandel-macaque (approved):

PR #12091, opened 2021-07-08
[GameKit] Updates for Xcode13 Beta 1-2

tj-devel709:


src/gamekit.cs
```diff
@@ -853,6 +853,22 @@ interface GKLocalPlayer
...
+		[NoWatch, NoTV, NoMac, iOS (15,0), NoMacCatalyst]
+		[Export ("presentFriendRequestCreatorFromViewController:error:")]
+		bool PresentFriendRequestCreator (UIViewController viewController, [NullAllowed] out NSError error);
+#endif
+
+#if MONOMAC
+		[NoWatch, NoTV, NoiOS, Mac (12,0), NoMacCatalyst]
+		[Export ("presentFriendRequestCreatorFromWindow:error:")]
+		bool PresentFriendRequestCreatorFromWindow ([NullAllowed] NSWindow window, [NullAllowed] out NSError error);
+#endif
```
mandel-macaque, line 870: Is the #if needed? Also, also, you have a #if !MONOMAC and a #if MONOMAC what to use a:
```suggestion
#if MONOMAC
		[NoWatch, NoTV, NoiOS, Mac (12,0), NoMacCatalyst]
		[Export ("presentFriendRequestCreatorFromWindow:error:")]
		bool PresentFriendRequestCreatorFromWindow ([NullAllowed] NSWindow window, [NullAllowed] out NSError error);
#else
		[NoWatch, NoTV, NoMac, iOS (15,0), NoMacCatalyst]
		[Export ("presentFriendRequestCreatorFromViewController:error:")]
		bool PresentFriendRequestCreator (UIViewController viewController, [NullAllowed] out NSError error);
#endif
```

mandel-macaque (changes requested):

src/gamekit.cs
tj-devel709: Is it not needed? I think I got a build error when I did not use the #ifs.
Also these two methods are not related so I thought the separate #ifs would be better in case these were ever moved or anything?

mandel-macaque: 1. What error did you get?
2. No, we want to reduce the number of #if else it is very hard to review and confusing. Preprocessor directives make things harder.

tj-devel709: Okay, I'll build again and see which error it was!

tj-devel709: The error:
```
gamekit.cs(866,61): error CS0246: The type or namespace name 'NSWindow' could 
not be found (are you missing a using directive or an assembly reference?)
```

mandel-macaque: Hm, that errors is coming when you compile for iOS, which makes sense since NSWindow is part of AppKit and not present on iOS. So, the two methods are related, and I would do the following:

1. use the #if/#else/#endif
2. Rename PresentFriendRequestCreatorFromWindow to PresentFriendRequestCreator since FromWindow is redundant. 

Both methods are probably the same one but working around the fact that UIKit and AppKit are not the same.

tj-devel709: Ah okay that makes sense thanks! They do look related and I just missed it!

mandel-macaque:
@tj-devel709 **do not merge** until the cecil tests pass.

src/GameKit/GameKit.cs
```diff
@@ -144,6 +144,8 @@ public enum GKError : long {
 		FriendListDescriptionMissing = 100,
 		FriendListRestricted = 101,
 		FriendListDenied = 102,
+		[iOS (15,0), Mac (12,0), MacCatalyst (15,0), Watch (8,0), TV (15,0)]
```
rolfbjarne, line 147: We usually don't add availability information on enum values that are errors, it doesn't really make much sense, since users aren't really using these values.

src/gamekit.cs
```diff
@@ -853,6 +853,21 @@ interface GKLocalPlayer
 		Action<NSViewController, NSError> AuthenticateHandler { get; set; }
 #endif
 
+		[NoWatch, NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Export ("isPresentingFriendRequestViewController")]
+		bool IsPresentingFriendRequestViewController { get; }
+
+#if !MONOMAC
```
rolfbjarne, line 860: You can add something like this at the top of the file:

```csharp
using NSWindow = Foundation.NSObject;
```

here: https://github.com/xamarin/xamarin-macios/blob/a7c9099153bb34bee78a26f925fa0fc545c2a4c0/src/gamekit.cs#L26-L29

This way the API declaration will compile just fine without the conditional logic.

You shouldn't need something for `UIViewController`, since it's already there:

https://github.com/xamarin/xamarin-macios/blob/a7c9099153bb34bee78a26f925fa0fc545c2a4c0/src/gamekit.cs#L25

rolfbjarne (changes requested):

src/GameKit/GameKit.cs
tj-devel709: Okay that makes sense! Thanks!

src/gamekit.cs
tj-devel709: Okay that makes sense why the UIViewController was not giving me errors! Thanks!

rolfbjarne (approved):

mandel-macaque (approved):

PR #12093, opened 2021-07-09
[dotnet] no default $(RuntimeIdentifier) for class libraries

jonathanpeppers:
Context: https://github.com/dotnet/maui/pull/1578

In the latest xamarin-macios bump, the dotnet/maui build started
failing because the `$(OutputPath)` of all the projects changed from:

    bin/Debug/net6.0-ios/
    bin/Debug/net6.0-maccatalyst/

To:

    bin/Debug/net6.0-ios/iossimulator-x64/
    bin/Debug/net6.0-maccatalyst/maccatalyst-x64/

Class library projects appear to have `$(RuntimeIdentifier)` set by
default. I don't think this is needed, because class libraries do not
need to be compiled as `--self-contained` apps.

On the Android side, we only set a default `$(RuntimeIdentifiers)` for
application projects:

https://github.com/xamarin/xamarin-android/blob/7c5fab13329ee898fb1562f83576a7ca881f2881/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L49-L61

I think the correct change here is to check:

    '$(OutputType)' == 'Exe' Or '$(IsAppExtension)' == 'true'

So Apple platforms only put a default `$(RuntimeIdentifier)` for
applications or extensions.


rolfbjarne (approved):

rolfbjarne:

PR #12096, opened 2021-07-09
[IOSurface] Updates for Xcode13 Beta 1

tj-devel709:
I don't feel too confident with this one.
I added the missing p-invokes to the ignore since they are ignored in other platforms.

I added the two fields. They say they are supported in older versions and the iOS 11.4 intro test did not complain about them, but a good amount of other IOSurface* fields are in the common ignore and I am not sure why.

### DIFF
--- a/src/iosurface.cs
+++ b/src/iosurface.cs
@@ -228,5 +228,13 @@ namespace IOSurface {
 		[iOS (11,0)][Mac (10,13)]
 		[Internal, Export ("setPurgeable:oldState:")]
 		int _SetPurgeable (IOSurfacePurgeabilityState newState, IntPtr oldStatePtr);
+
+		[Watch (4,0), TV (11,0), Mac (10,6), iOS (11,0), MacCatalyst (13,0)]
+		[Field ("kIOSurfaceColorSpace")]
+		NSString kIOSurfaceColorSpace { get; }
+
+		[Watch (4,0), TV (11,0), Mac (10,6), iOS (11,0), MacCatalyst (13,0)]
+		[Field ("kIOSurfaceICCProfile")]
+		NSString kIOSurfaceICCProfile { get; }
 	}
 }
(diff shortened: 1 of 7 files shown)


src/iosurface.cs
```diff
@@ -228,5 +228,13 @@ interface IOSurface : NSSecureCoding
 		[iOS (11,0)][Mac (10,13)]
 		[Internal, Export ("setPurgeable:oldState:")]
 		int _SetPurgeable (IOSurfacePurgeabilityState newState, IntPtr oldStatePtr);
+
+		[Watch (4,0), TV (11,0), Mac (10,6), iOS (11,0), MacCatalyst (13,0)]
+		[Field ("kIOSurfaceColorSpace")]
+		NSString kIOSurfaceColorSpace { get; }
```
mandel-macaque, line 234: kIOSurfaceColorSpace is not correct. Those are global in objc and that is why they add the framework name and also that k which stands for key.

src/iosurface.cs
```diff
@@ -228,5 +228,13 @@ interface IOSurface : NSSecureCoding
...
+
+		[Watch (4,0), TV (11,0), Mac (10,6), iOS (11,0), MacCatalyst (13,0)]
+		[Field ("kIOSurfaceICCProfile")]
+		NSString kIOSurfaceICCProfile { get; }
```
mandel-macaque, line 238: Same.

mandel-macaque (changes requested):

rolfbjarne (changes requested):
Just add the two new fields and P/Invokes to the common ignore file, there are a lot of other API that we haven't bound in this framework, and which we'll probably have to bind first for any of these to make sense.

rolfbjarne (approved):

tj-devel709:
Unrelated Test Failure: https://github.com/xamarin/maccore/issues/2443

mandel-macaque (approved):

PR #12102, opened 2021-07-12
[SafariServices] Update to Xcode13 beta1

tj-devel709:


rolfbjarne (approved):

tj-devel709:
Whoops, forgot to delete the todos

src/safariservices.cs
```diff
@@ -116,6 +117,11 @@ interface SFSafariViewController {
 		[iOS (11,0)]
 		[Export ("dismissButtonStyle", ArgumentSemantic.Assign)]
 		SFSafariViewControllerDismissButtonStyle DismissButtonStyle { get; set; }
+
+		[iOS (15,0), MacCatalyst (15,0)]
+		[Static]
+		[Export ("prewarmConnectionsToURLs:")]
+		SFSafariViewControllerPrewarmingToken PrewarmConnections (NSUrl[] Urls);
```
spouliot, line 124: parameter names are camel cased -> `urls`

src/safariservices.cs
```diff
@@ -456,4 +467,38 @@ interface SFUniversalLink {
 		bool Enabled { [Bind ("isEnabled")] get; set; }
 	}
 #endif
+
+	[Static]
+	[iOS (15,0), Mac (11,0), MacCatalyst (15,0), NoTV, NoWatch]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 473: this should work without being an `NSObject` subclass
since those have impact (registration) it's better to avoid them, unless needed

spouliot (changes requested):

tests/xtro-sharpie/MacCatalyst-SafariServices.ignore
```diff
@@ -0,0 +1,2 @@
+## for compatibility we have extra code that return `false` if `url` is `null`
+!extra-null-allowed! 'System.Boolean SafariServices.SSReadingList::SupportsUrl(Foundation.NSUrl)' has a extraneous [NullAllowed] on parameter #0
```
mandel-macaque, line 2: This is a copy&paste from the iOS ignore, since it issuing to happen in all platforms (they all share code) it should be moved to the common file rather than having the same string in both files (and same explanation).

From the iOS todo:
```
iOS-SafariServices.ignore-## for compatibility we have extra code that return `false` if `url` is `null`
iOS-SafariServices.ignore:!extra-null-allowed! 'System.Boolean SafariServices.SSReadingList::SupportsUrl(Foundation.NSUrl)' has a extraneous [NullAllowed] on parameter #0
```

mandel-macaque (changes requested):
Please merge with main too.

src/safariservices.cs
tj-devel709: Okay! What is a way to know when to use them? Do I not use it in this interface just contains fields?

spouliot: If it's a `static` type and it's **not** a type that exists in Apple SDK then assume, by default, that it's not needed (to have it subclass `NSObject`).

spouliot (approved):

src/safariservices.cs
```diff
@@ -456,4 +467,37 @@ interface SFUniversalLink {
...
+		UIImage TemplateImage { get; }
+
+		[NullAllowed, Export ("extensionIdentifier")]
+		string ExtensionIdentifier { get; }
+	}
+
+	[iOS (15,0), MacCatalyst (15,0), NoMac, NoTV, NoWatch]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface SFSafariViewControllerPrewarmingToken : NSCoding, NSSecureCoding
```
tj-devel709, line 498: Note to TJ,
These protocols may need to be removed from here and this interface may need to be documented in iOSApiProtocolTest.cs

tj-devel709:
Updating to fix unrelated failure: https://github.com/xamarin/maccore/issues/2472

tj-devel709:
Unrelated Test Failures: https://github.com/xamarin/maccore/issues/2443 and https://github.com/xamarin/maccore/issues/2471

tj-devel709:
Fixing merge conflicts ^

tj-devel709:
Unrelated Test Failures:
monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2471
link all/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414

PR #12104, opened 2021-07-13
[runtime] Set the current directory to the root directory of the app bundle for all platforms in .NET.

rolfbjarne:
To have consistent behavior in .NET, set the current directory to the root of
the app bundle for all platforms.

This is a breaking change for legacy Xamarin.Mac, which used to set the
current directory to the Contents/Resources subdirectory, but there's a simple
workaround for customers that depend on the old behavior (change it in Main
themselves), and I believe the consistent experience across platforms warrants
this change.

Note that we already had a breaking change here for macOS/.NET: we were
(unintentionally) setting the current directory to the Contents/MonoBundle
directory, which neither matched mobile platforms, nor the legacy Xamarin.Mac
behavior.

This solves the problem of what to do for Mac Catalyst apps, because there's
no need to choose between the macOS or the mobile behavior, since they're the
same.

This required changing the launch of macOS apps using CoreCLR to pass the full
path to the entry assembly, since the entry assembly isn't in the current
directory anymore.

### DIFF
--- a/runtime/coreclr-bridge.m
+++ b/runtime/coreclr-bridge.m
@@ -602,18 +602,18 @@ mono_jit_exec (MonoDomain * domain, MonoAssembly * assembly, int argc, const cha
 {
 	unsigned int exitCode = 0;
 
-	char *assemblyName = xamarin_bridge_get_assembly_name (assembly->gchandle);
+	char *assemblyPath = xamarin_bridge_get_assembly_location (assembly->gchandle);
 
-	LOG_CORECLR (stderr, "mono_jit_exec (%p, %p, %i, %p) => EXECUTING %s\n", domain, assembly, argc, argv, assemblyName);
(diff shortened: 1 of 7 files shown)


rolfbjarne:
Test failures are unrelated

* monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2451
* monotouch-test/tvOS - simulator/Debug: https://github.com/xamarin/maccore/issues/1605

mandel-macaque (approved):
+1 but lets ping VsMac just in case :)

stephen-hawley (approved):

rolfbjarne:
CC @emaf just in case this is something that would affect you (I don't think it would though, but just to be sure).

rolfbjarne:
CC @Therzok is this something that would affect you?

PR #12106, opened 2021-07-13
[runtime] Add logging when a P/Invoke can't be resolved in xamarin_pinvoke_override.

rolfbjarne:
Return early when we're not going to try resolving anything, which means that
if we didn't find something by the end, we know that it's because we failed
(and not because we weren't supposed to try), and we log that.

This makes it easier to diagnose a few failure conditions.


tj-devel709 (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2464).

PR #12121, opened 2021-07-14
[SceneKit] Update bindings for Xcode13.0 beta 1 & beta 2

rachelkang:


src/scenekit.cs
```diff
@@ -3991,6 +3997,11 @@ interface SCNShadable {
 		[Wrap ("WeakShaderModifiers")]
 		SCNShaderModifiers ShaderModifiers { get; set; }
 
+		[Mac (12,0), iOS (15,0), TV (15,0), Watch (8,0)]
+		[NullAllowed] // by default this property is null
+		[Export ("minimumLanguageVersion", ArgumentSemantic.Retain)]
+		NSNumber MinimumLanguageVersion { get; set; }
```
rachelkang, line 4003: Not sure if this should be a different numeric type - and if so, which one? int, float, nint, nfloat, etc...

src/scenekit.cs
```diff
@@ -160,6 +160,12 @@ interface SCNAnimatable {
 		[Export ("removeAllAnimations")]
 		void RemoveAllAnimations ();
 
+		[Introduced (PlatformName.MacCatalyst, 15, 0)]
```
rolfbjarne, line 163: No big deal, but you can use the shorter `[MacCatalyst (15, 0)]`

src/scenekit.cs
```diff
@@ -160,6 +160,12 @@ interface SCNAnimatable {
...
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+		[Abstract]
```
rolfbjarne, line 165: Adding an `[Abstract]` member to an existing protocol is a breaking change.

What we do is:

1. Add the `[Abstract]` inside a `#if XAMCORE_4_0` block
2. Add the xtro failure to the corresponding .ignore file, with a comment explaining why.

src/scenekit.cs
rolfbjarne: Just a FYI that that would require using a `[BindAs (typeof (<numeric type>?)]`.

rolfbjarne (changes requested):

src/scenekit.cs
rachelkang: hmm so is it fine as NSNumber for now? or am I misunderstanding?

rolfbjarne: The header says:

> @abstract The minimum language version required to interpret the shadable source code (wrapped MTLLanguageVersion). Defaults to nil.

so this should be a `[BindAs (typeof (MTLLanguageVersion?)]`

rolfbjarne (changes requested):

mandel-macaque (approved):
Approving, please address @rolfbjarne comment regarding the binders.

tj-devel709 (approved):

src/SceneKit/Defs.cs
```diff
@@ -283,7 +283,7 @@ public enum SCNAntialiasingMode : ulong {
 		None,
 		Multisampling2X,
 		Multisampling4X,
-#if MONOMAC
+#if MONOMAC || __MACCATALYST__
```
spouliot, line 286: Is catalyst support new new in Xcode 13 ?

In that case there should be a `[MacCatalyst (15,0)]` on each member

tests/xtro-sharpie/MacCatalyst-SceneKit.ignore
```diff
@@ -0,0 +1,8 @@
+# deprecated or not used on catalyst
+!missing-protocol-member! SCNProgramDelegate::programIsOpaque: not found
+!missing-selector! SCNMaterialProperty::borderColor not bound
+!missing-selector! SCNMaterialProperty::setBorderColor: not bound
+!missing-selector! SCNRenderer::render not bound
+
+# adding abstract member to an existing type would be a breaking change
+!incorrect-protocol-member! SCNAnimatable::removeAllAnimationsWithBlendOutDuration: is REQUIRED and should be abstract
```
spouliot, line 8: when something needs to be annotated for all platforms then you should add it to the `common-*.ignore` file

src/scenekit.cs
dalexsoto: Unfortunately this is inside a protocol so BindAs is not an option, you will need to keep `NSNumber`

```
	[Protocol, Model]
	interface SCNShadable {
```

src/SceneKit/Defs.cs
rachelkang: No, it's not new new! I think I only got this because of the xtro - header files don't mention it, and web docs indicate 13.1 https://developer.apple.com/documentation/scenekit/scnantialiasingmode

tests/xtro-sharpie/MacCatalyst-SceneKit.ignore
rachelkang: got it, thank you!

spouliot (approved):

rolfbjarne (approved):

mandel-macaque:
Unrelated test failure: https://github.com/xamarin/maccore/issues/2443

dalexsoto:
Beta 4 is out, we should wait for the bump PR to land

dalexsoto (approved):

mandel-macaque:
Diff failures, with looks like https://github.com/xamarin/maccore/issues/2414

PR #12133, opened 2021-07-16
[xharness] Handle exceptions that occur while listing device candidates gracefully.

rolfbjarne:
This fixes a problem where the html report wouldn't be re-generated if there
was a problem loading a device (which for instance could happen if xharness
was unable to create a required simulator).

### DIFF
--- a/tests/xharness/Jenkins/Reports/HtmlReportWriter.cs
+++ b/tests/xharness/Jenkins/Reports/HtmlReportWriter.cs
@@ -405,8 +405,12 @@ namespace Xharness.Jenkins.Reports {
 									candidates = (runTest as RunSimulatorTask)?.Candidates;
 								if (candidates != null) {
 									writer.WriteLine ($"Candidate devices:<br />");
-									foreach (var candidate in candidates)
-										writer.WriteLine ($"&nbsp;&nbsp;&nbsp;&nbsp;{candidate.Name} (Version: {candidate.OSVersion})<br />");
+									try {
+										foreach (var candidate in candidates)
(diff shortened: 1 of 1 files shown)


tj-devel709 (approved):

tests/xharness/Jenkins/Reports/HtmlReportWriter.cs
```diff
@@ -405,8 +405,12 @@ <li>Toggle visibility
 									candidates = (runTest as RunSimulatorTask)?.Candidates;
 								if (candidates != null) {
 									writer.WriteLine ($"Candidate devices:<br />");
-									foreach (var candidate in candidates)
-										writer.WriteLine ($"&nbsp;&nbsp;&nbsp;&nbsp;{candidate.Name} (Version: {candidate.OSVersion})<br />");
+									try {
+										foreach (var candidate in candidates)
+											writer.WriteLine ($"&nbsp;&nbsp;&nbsp;&nbsp;{candidate.Name} (Version: {candidate.OSVersion})<br />");
```
spouliot, line 410: Moving the `try catch` inside the `foreach` would list all the _valid_ candidates - not just the first ones before a bad entry.

Not sure if that can happen in real-life ?!? or if it would be useful when debugging such issue...

spouliot (approved):

mandel-macaque (approved):

tests/xharness/Jenkins/Reports/HtmlReportWriter.cs
rolfbjarne: It's the iteration of the `candidates` variable that throws the exception, so if I put the `try catch` inside, it won't catch anything.

PR #12135, opened 2021-07-16
[src] Share the NSTextAttachment[Container] implementations between AppKit and UIKit.

rolfbjarne:


src/xkit.cs
```diff
@@ -2090,4 +2093,78 @@ interface NSLayoutConstraint
...
+	[Mac (10,11)]
+	[MacCatalyst (13,0)]
+	[Model]
+	[Protocol]
+	[BaseType (typeof (NSObject))]
+	partial interface NSTextAttachmentContainer {
+		[Abstract]
+		[Export ("imageForBounds:textContainer:characterIndex:")]
+		[return: NullAllowed]
+#if MONOMAC && !XAMCORE_4_0
```
Therzok, line 2107: Is the macOS target not going to work anymore in XAMCORE 4?

Looks like it would generate UIImage

Therzok: Or was this supposed to return `Image` as it's alias'd above

rolfbjarne: Fixed to use the alias'ed definition above. Thanks!

rolfbjarne:

spouliot (approved):
API diff for macOS shows some modifications for parameter names - but we generally prefer consistence (and sharing) over strict binary compatibility

PR #12136, opened 2021-07-16
[dotnet/msbuild] Add support for using LLVM to build .NET apps. Fixes #11379.

### LINKED ISSUE #11379: .NET: Implement support for AOT-compiling using LLVM
https://github.com/xamarin/xamarin-macios/blob/b4b8ee735c60bf897c87e71057fd6f2ab11206d6/tools/dotnet-linker/Steps/ComputeAOTArguments.cs#L40-L41

This should be just a matter of copying/porting whatever mtouch does when using LLVM to dotnet-linker.

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/issues/11379.


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -703,7 +705,7 @@
 			Condition="'$(_SdkIsSimulator)' != 'true' And '$(_PlatformName)' != 'macOS'"
 			DependsOnTargets="_ComputeVariables"
 			Inputs="@(_AssembliesToAOT)"
-			Outputs="@(_AssembliesToAOT -> '$(_AOTOutputDirectory)%(Arch)\%(Filename)%(Extension).o')">
+			Outputs="@(_AssembliesToAOT -> '%(ObjectFile)');@(_AssembliesToAOT -> '%(LLVMFile)');">
```
emaf, line 708: @rolfbjarne are both ObjectFile and LLVMFile ITaskItem outputs of any task? I'm assuming those files will be produced by a task executed on macOS, so we need those to be an output prop of a task so they are copied to Windows to make the Inputs/Outputs check work, otherwise this target will always be executed from Windows.

spouliot (approved):
That does not seems to fulfill the request of having LLVM by default for release (non-debug) builds.

It might be elsewhere... I'm a bit behind on the changes :)

rolfbjarne:
> That does not seems to fulfill the request of having LLVM by default for release (non-debug) builds.

Before we can make it the default, we have to implement it :) Making it the default once it's working is easy enough.

spouliot:
> we have to implement it

hmm... I thought this PR was the (last) missing piece.

dotnet/targets/Xamarin.Shared.Sdk.targets
rolfbjarne: @emaf I changed this to add these files to `FileWrites`: https://github.com/xamarin/xamarin-macios/pull/12136/commits/d59c218a711ffcb89b1c4541ef473b3d5795ebbd, would that work?

rolfbjarne:
> > we have to implement it
> 
> hmm... I thought this PR was the (last) missing piece.

I've created #12147 to track making LLVM the default for release builds.

dotnet/targets/Xamarin.Shared.Sdk.targets
emaf: Yeah, I think that should work!

emaf (approved):

PR #12145, opened 2021-07-19
[ImageIO] Add support for Xcode 13 beta 3.

mandel-macaque:
@spouliot the headers say iOS 19 which makes no sense (although is good to know apple is thinking that far ahead ;) So I added 15, I wonder if adding 19 and wait for the next beta is the correct way to go.

### DIFF
--- a/src/imageio.cs
+++ b/src/imageio.cs
@@ -576,6 +576,9 @@ namespace ImageIO {
 		NSString PNGSoftware { get; }
 		[Field ("kCGImagePropertyPNGTitle")]
 		NSString PNGTitle { get; }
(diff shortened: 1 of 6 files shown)


rolfbjarne (approved):

spouliot (approved):
> So I added 15, I wonder if adding 19 and wait for the next beta is the correct way to go.

No need. Apple can either:
* correct this (to 15) before release (might be after Xcode 13.0 is shipped)
  * that would require another commit (or two, if after 13.0)
* remove it
  * that would require another commit

Using `15,0` means you're likely avoiding another commit in the future, one that could be forgotten (if done after Xcode 13 is stable). Adding `19,0` would also require _incorrect_ changes to the version (Xcode/OS) mapping code used for testing.

PR #12146, opened 2021-07-20
[AppKit/UIKit] Share the implementation of some classes between AppKit and UIKit.

rolfbjarne:
Fourth group of classes (there will be more!).

The only breaking changes should be renamed parameters (for macOS, to match
the other platforms).

This PR might be easier to review commit-by-commit.


spouliot (approved):
diffs looks fine
but xtro unhappy

rolfbjarne:
> diffs looks fine
> but xtro unhappy

I had it fixed locally, but forgot to push 🤦

src/xkit.cs
```diff
@@ -2167,4 +2168,476 @@ partial interface NSTextAttachment : NSTextAttachmentContainer, NSSecureCoding
...
+#if MONOMAC && !XAMCORE_4_0
+		[Export ("initWithString:")]
+		IntPtr Constructor (string str);
+#endif
+
+		[Export ("layoutManagers")]
+#if MONOMAC || XAMCORE_4_0
+		NSLayoutManager [] LayoutManagers { get; }
+#else
+		NSObject [] LayoutManagers { get; }
```
mandel-macaque, line 2186: Why not do the using NSLayoutManager  = System.NSObject[] in a #if at the start of the file?

mandel-macaque (approved):

src/xkit.cs
rolfbjarne: We could do that, but the "using ..." pattern is usually used for types that don't exist on a particular platform to make the rest of the code compile, while in this case it's a binding mistake/improvement, which is only a problem in this particular location, so you'd end up with something like this at the top of the file:

```csharp
#if MONOMAC || XAMCORE_4_0
using LayoutManagersReturnTypeForNSTextStorage = Foundation.NSLayoutManager[];
#else
using LayoutManagersReturnTypeForNSTextStorage = Foundation.NSObject[];
```

and I don't think that looks much better.

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

* monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2443
* link sdk/Mac Catalyst/Debug: https://github.com/xamarin/maccore/issues/2414

PR #12152, opened 2021-07-20
[StoreKit] Update bindings for Xcode 13.0 beta 1

rachelkang:
No beta 2 or beta 3 diffs


src/storekit.cs
```diff
@@ -659,6 +659,10 @@ interface SKStoreProductParameterKey
 		[MacCatalyst (14,0)]
 		[Field ("SKStoreProductParameterAdNetworkVersion")]
 		NSString AdNetworkVersion { get; }
+		
+		[Mac (12, 0), iOS (15, 0), MacCatalyst (15, 0)]
```
spouliot, line 663: that needs a `[NoTV]` if it's not available on tvOS
otherwise it will default to the minimum value (9.0) since it will fallback to the type definition (which is supported on tvOS)

There's no need for a `[NoWatch]` here since it's already present at the type-level

src/storekit.cs
```diff
@@ -774,7 +778,7 @@ interface SKCloudServiceController {
 		[Static]
 		[Async]
 		[Export ("requestAuthorization:")]
-		void RequestAuthorization (Action<SKCloudServiceAuthorizationStatus> handler);
+		void RequestAuthorization (Action<SKCloudServiceAuthorizationStatus> completionHandler);
```
spouliot, line 781: it's better but it's a breaking change - it's not the worse kind but it's better avoided unless something else force us to make that change

src/storekit.cs
```diff
@@ -1076,6 +1080,16 @@ interface SKOverlayAppConfiguration {
...
 
+		[iOS (15, 0), MacCatalyst (15, 0)]
+		[NullAllowed]
+		[Export ("customProductPageIdentifier", ArgumentSemantic.Retain)]
+		string CustomProductPageIdentifier { get; set; }
+
+		[iOS (15, 0), MacCatalyst (15, 0)]
+		[NullAllowed]
+		[Export ("latestReleaseID", ArgumentSemantic.Retain)]
+		string LatestReleaseID { get; set; }
```
spouliot, line 1091: `Id` -> it's no an acronym

src/storekit.cs
```diff
@@ -1105,6 +1119,16 @@ interface SKOverlayAppClipConfiguration {
...
```
spouliot, line 1130: same `Id`

spouliot (changes requested):

rolfbjarne (approved):

mandel-macaque (approved):
+1 after the changes requested by @spouliot

tj-devel709 (approved):

src/storekit.cs
rachelkang: header files don't say anything about TV, but the [web docs](https://developer.apple.com/documentation/storekit/skstoreproductparametercustomproductpageidentifier?changes=_4&language=objc) indicate availability, so I'm adding it in!

spouliot (approved):

dalexsoto (approved):

PR #12153, opened 2021-07-20
[CoreBluetooh] Add Xcode 13 beta 3 support.

mandel-macaque:


src/CoreBluetooth/Enums.cs
```diff
@@ -180,7 +180,7 @@ public enum CBConnectionEvent : long {
 		Connected = 1,
 	}
 
-	[Flags, iOS (13,0), TV (13,0), Watch (6,0), NoMac]
+	[Flags, iOS (13,0), TV (13,0), Watch (6,0), Mac (12,0)]
```
spouliot, line 183: it's not really available, just noise (and extra metadata for us), as the only API that use it is not available on macOS

```
		[iOS (13,0), TV (13,0), Watch (6,0), NoMac]
		[Static]
		[Export ("supportsFeatures:")]
		bool SupportsFeatures (CBCentralManagerFeature features);
```

let's just add that into the `macOS-CoreBluetooth.ignore` file instead

rolfbjarne (approved):

spouliot (changes requested):

spouliot (approved):

tj-devel709 (approved):

mandel-macaque:

PR #12157, opened 2021-07-20
[CI] Add Prepare Release and VS Insertion stages (again)

pjcollins:
Context: xamarin/yaml-templates#117

Updates the .NET 6 NuGet packaging steps to exclude package metadata,
as the .msi conversion tooling does not process .nupkg file names with
the `+sha.commit` metadata.

Two new stages have been added to facilitate the Visual Studio setup
authoring process.

The first stage named "Prepare Release" will sign the .NET 6 NuGet
package content (inside and out), convert relevant packages to .msi
installers, generate Visual Studio manifests for the .msi installers,
and push the signed packages to the `xamarin-impl` feed.

The new `SignList.xml` file is required for our NuGet signing templates.

The new `xamarin-workload.props` file contains version information
and other metadata required to generate a Visual Studio manifest.

The second stage starts with a [manual validation task][0].  This task
will pause and wait for someone to click a "Resume" or "Reject" button
that will appear on the pipeline UI.  This task is configured to be
rejected after waiting for two days, but it can be manually re-ran at a
later date if we want to trigger VS insertion for an older build.

If the manual validation task is approved, a VS Drop will be created
containing all .NET 6 .msi files.  This Drop URL can then be used to
update our component versions in Visual Studio.  This last piece is
currently manual as we will initially be introducing new components,
however we should be able to automate VS PR creation in the future.

Commit 09f911b504ec9b803eb6d725cca14cc7d3ed986e missed adding the
PR build check condition to a step in `sign-and-notarized.yml`, causing
PR builds from forks to fail.  We can fix this by adding in the missing
condition.

[0]: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/manual-validation?view=azure-devops&tabs=yaml


mandel-macaque:
PR is against a fork so we should see if this time works

rolfbjarne (approved):

mandel-macaque (approved):
PR build worked from fork which was the issue we had in the previos attempt.

PR #12163, opened 2021-07-21
[Photos] Update bindings for Xcode 13.0 beta 1 and beta 2

rachelkang:
No beta 3 updates


src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
 	[BaseType (typeof (PHPhotoLibrary))]
 	interface PHPhotoLibrary_CloudIdentifiers {
 
+		[TV (15,0), Mac (12,0), iOS (15,0)]
+		[Export ("localIdentifierMappingsForCloudIdentifiers:")]
+		NSDictionary<PHCloudIdentifier, PHLocalIdentifierMapping> LocalIdentifierMappingsForCloudIdentifiers (PHCloudIdentifier[] cloudIdentifiers);
+
+		[TV (15,0), Mac (12,0), iOS (15,0)]
+		[Export ("cloudIdentifierMappingsForLocalIdentifiers:")]
+		NSDictionary<NSString, PHCloudIdentifierMapping> CloudIdentifierMappingsForLocalIdentifiers (string[] localIdentifiers);
```
mandel-macaque, line 1188: You need to add MacCatalyst (15,0) else it will be using the default min version.

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
+
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'localIdentifierMappingsForCloudIdentifiers:' instead.")]
 		[Export ("localIdentifiersForCloudIdentifiers:")]
 		string[] GetLocalIdentifiers (PHCloudIdentifier[] cloudIdentifiers);
```
mandel-macaque, line 1192: Do we want this to be added in MacCatalyst if it is deprecated, We should add NotMacCatalyst if we are deprecating. Is it deprecated on iOS and Tv?

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'cloudIdentifierMappingsForCloudIdentifiers:' instead.")]
```
mandel-macaque, line 1194: Same.

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
 		[Export ("cloudIdentifiersForLocalIdentifiers:")]
 		PHCloudIdentifier[] GetCloudIdentifiers (string[] localIdentifiers);
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosErrorIdentifierNotFound' instead.")]
```
mandel-macaque, line 1198: Same

mandel-macaque (changes requested):

src/Photos/Enums.cs
```diff
@@ -356,7 +358,9 @@ public enum PHProjectSectionType : long {
 	[Native]
 	[ErrorDomain ("PHLivePhotoEditingErrorDomain")]
 	public enum PHLivePhotoEditingError : long {
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use 'InternalError' instead.")]
```
spouliot, line 361: the suggested alternative is inside a different enum, which might be hard to use. Sadly not much we can do about it :(

However we should name that enum `Use 'PHPhotosError.InternalError' instead.`

src/Photos/Enums.cs
```diff
@@ -356,7 +358,9 @@ public enum PHProjectSectionType : long {
...
 		Unknown,
+		[Deprecated (PlatformName.MacOSX, 10, 15, message: "Use 'UserCancelled' instead.")]
```
spouliot, line 363: same, prefix value with `PHPhotosError.`

src/Photos/Enums.cs
```diff
@@ -378,12 +382,35 @@ public enum FigExifCustomRenderedValue : short {
...
 	[Native]
 	public enum PHPhotosError : long {
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'InternalError' instead.")]
 		Invalid = -1,
+		InternalError = -1,
 		UserCancelled = 3072,
 		LibraryVolumeOffline = 3114,
 		RelinquishingLibraryBundleToWriter = 3142,
 		SwitchingSystemPhotoLibrary = 3143,
+		[TV (14,0), Mac (11,0), iOS (14,0)]
```
spouliot, line 392: It's a special case but we do not annotate error codes with OS versions. 

Why ? They are are output-only and, as enum values, they do not have to exists to be used. E.g.

```csharp
var error = CallApi ();
switch (error)
case PHPhotosError. NetworkAccessRequired:
    // it does not make sense to add a version check here
    // and we don't want compiler/tooling warnings for them if we don't 
    Console.WriteLine ("Network Access Required");
    break;
// ... more cases ...
default:
    Console.WriteLine ($"Error {error}");
    break;
}
```

It's also a bunch of (not really useful) metadata to include inside platform assemblies.

src/photos.cs
```diff
@@ -146,9 +146,14 @@ interface PHAsset {
 		[Export ("playbackStyle", ArgumentSemantic.Assign)]
 		PHAssetPlaybackStyle PlaybackStyle { get; }
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosErrorIdentifierNotFound' instead.")]
```
spouliot, line 149: missing dot `.` -> `PHPhotosError.IdentifierNotFound`

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
 		[TV (13,0), Mac (10,15), iOS (13,0)]
 		[Export ("unregisterAvailabilityObserver:")]
 		void Unregister (IPHPhotoLibraryAvailabilityObserver observer);
+
+		[TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Field ("PHLocalIdentifiersErrorKey")]
+		NSString PHLocalIdentifiersErrorKey { get; }
```
spouliot, line 1173: not sure if it's the best place for that key
but it's not named correctly, IOW you do not want to use `PHPhotoLibrary.PHLocalIdentifiersErrorKey`
minimally remove `PH` prefix
also check if other constants exists - there might be a better place for this one (or not)

src/photos.cs
spouliot: yup, it looks like this needs a `[NoTV][NoiOS][NoMacCatalyst]` since they are unlikely to be both added and deprecated
and if that's the case then it's generally better not to include them anyway

src/photos.cs
```diff
@@ -1406,13 +1424,15 @@ interface PHProjectChangeRequest {
 	}
 
 	[Mac (10,13)]
-	[NoiOS][NoTV]
-	[Unavailable (PlatformName.MacCatalyst)]
+	[TV (15,0)]
+	[iOS (15,0)]
 	[Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 1429: that needs to be removed as you already removed the `[Unavailable (PlatformName.MacCatalyst)]`

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
+		[TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Field ("PHLocalIdentifiersErrorKey")]
+		NSString PHLocalIdentifiersErrorKey { get; }
 	}
 
 	[Mac (10,13)]
-	[NoTV][NoiOS]
 	[Category]
-	[Unavailable (PlatformName.MacCatalyst)]
 	[Advice ("This API is not available when using UIKit on macOS.")]
```
spouliot, line 1178: that needs to be removed as you already removed the `[Unavailable (PlatformName.MacCatalyst)]`

src/photos.cs
```diff
@@ -1406,13 +1424,15 @@ interface PHProjectChangeRequest {
...
-	[NoiOS][NoTV]
-	[Unavailable (PlatformName.MacCatalyst)]
+	[TV (15,0)]
+	[iOS (15,0)]
 	[Advice ("This API is not available when using UIKit on macOS.")]
 	[BaseType (typeof (NSObject))]
 	[DisableDefaultCtor]
 	interface PHCloudIdentifier : NSSecureCoding {
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosErrorIdentifierNotFound' instead.")]
```
spouliot, line 1434: missing dot `.`

spouliot (changes requested):

src/photos.cs
mandel-macaque: WE already did that :) That key is only for when the indentifer is not matched and we noticed that the only Object with methods for identifiers is this class, so it would be something like:
```csharp
PHPhotoLibrary.LocalIdentifierMappingsForCloudIdentifiers(.., out var error);
// access the error info
error.UserInfo[PHPhotoLibrary.LocalIdentifiersErrorKey]
```

Yet, I did not know that @rachelkang was going to add those two methods as a Category in a diff object, I would have added them in the library and not in PHPhotoLibrary_CloudIdentifiers. If we want a diff class for the category, lets move the key there, else lets move the category methods to  PHPhotoLibrary. My preference is  PHPhotoLibrary but I'm happy either way.

spouliot: sounds good - but the
> remove `PH` prefix
remains :)

src/photos.cs
rachelkang: The [Advice]? Doesn't that apply to macOS?

src/Photos/Enums.cs
rachelkang: that makes sense, thank you! removing now

src/photos.cs
rachelkang: removed the prefix!

Let me know if I should definitively move the key to PHPhotoLibrary_CloudIdentifiers or move the category methods to PHPhotoLibrary. Moving neither for now, although I think it would make sense to move the key to PHPhotoLibrary_CloudIdentifiers where the category methods are. I put the category methods there because the older versions of them (deprecated) also live there

src/Photos/Enums.cs
```diff
@@ -378,12 +382,35 @@ public enum FigExifCustomRenderedValue : short {
 	[TV (13,0), Mac (10,15), iOS (13,0)]
 	[Native]
 	public enum PHPhotosError : long {
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'InternalError' instead.")]
 		Invalid = -1,
```
rolfbjarne, line 386: No big deal, but you can add a `#if !XAMCORE_4_0` around the Invalid value.

src/photos.cs
```diff
@@ -1162,22 +1167,35 @@ interface PHPhotoLibrary {
...
 		[Export ("unregisterAvailabilityObserver:")]
 		void Unregister (IPHPhotoLibraryAvailabilityObserver observer);
+
+		[TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Field ("PHLocalIdentifiersErrorKey")]
+		NSString PHLocalIdentifiersErrorKey { get; }
 	}
 
 	[Mac (10,13)]
-	[NoTV][NoiOS]
```
rolfbjarne, line 1168: If the category has been added to iOS and tvOS now, you'll need the corresponding availability attributes (and that way you won't need them on each member either).

src/photos.cs
rolfbjarne: "UIKit on macOS" => Mac Catalyst (it's an old name).

rolfbjarne (changes requested):

src/photos.cs
rachelkang: ah, good to know! thanks!

src/Photos/Enums.cs
rachelkang: ok! adding it in, but curious - why does that help here?

src/Photos/Enums.cs
```diff
@@ -378,12 +382,26 @@ public enum FigExifCustomRenderedValue : short {
 	[TV (13,0), Mac (10,15), iOS (13,0)]
 	[Native]
 	public enum PHPhotosError : long {
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'InternalError' instead.")]
+#if !XAMCORE_4_0
 		Invalid = -1,
+#endif
+		InternalError = -1,
```
tj-devel709, line 389: Will having 2 enums with the value -1 cause issues?

src/photos.cs
```diff
@@ -1423,4 +1445,28 @@ interface PHCloudIdentifier : NSSecureCoding {
 		[Export ("initWithStringValue:")]
 		IntPtr Constructor (string stringValue);
 	}
+
+	[TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
```
tj-devel709, line 1450:
```suggestion
	[BaseType (typeof (NSObject))]
```
small

src/photos.cs
```diff
@@ -1423,4 +1445,28 @@ interface PHCloudIdentifier : NSSecureCoding {
...
+	{
+		[NullAllowed, Export ("cloudIdentifier")]
+		PHCloudIdentifier CloudIdentifier { get; }
+
+		[NullAllowed, Export ("error")]
+		NSError Error { get; }
+	}
+
+	[TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
```
tj-devel709, line 1462:
```suggestion
	[BaseType (typeof (NSObject))]
```
small also

src/photos.cs
```diff
@@ -146,9 +146,14 @@ interface PHAsset {
 		[Export ("playbackStyle", ArgumentSemantic.Assign)]
 		PHAssetPlaybackStyle PlaybackStyle { get; }
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosError.IdentifierNotFound' instead.")]
 		[NoTV][NoiOS]
 		[Field ("PHLocalIdentifierNotFound")]
 		NSString LocalIdentifierNotFound { get; }
+
+		[TV (15, 0), Mac (12, 0), iOS (15, 0), MacCatalyst (15,0)]
```
tj-devel709, line 154: Extra extra small but maybe try to be consistent with the spacing in the attribute versions

src/Photos/Enums.cs
rolfbjarne: Nope, it's totally valid to have multiple enum fields with the same numerical value.

src/Photos/Enums.cs
rolfbjarne: Just so that whenever we can break API (and define XAMCORE_4_0), we don't keep a deprecated enum value for which there's an alternative (with no drawbacks whatsoever).

src/Photos/Enums.cs
rolfbjarne: Oh wait, the [Deprecated] attribute goes inside the #if as well.

rolfbjarne (changes requested):

src/photos.cs
rachelkang: nice catch, thanks!

src/photos.cs
rachelkang: fixed! having seen both spacing styles all over (even in this file), I'm curious if this is the actual stylistic guideline we should stick with?

rolfbjarne (approved):

src/photos.cs
```diff
@@ -1162,22 +1167,38 @@ interface PHPhotoLibrary {
...
+	[TV (15,0), iOS (15,0), MacCatalyst (15,0)]
 	[Category]
-	[Unavailable (PlatformName.MacCatalyst)]
-	[Advice ("This API is not available when using UIKit on macOS.")]
 	[BaseType (typeof (PHPhotoLibrary))]
 	interface PHPhotoLibrary_CloudIdentifiers {
 
+		[Mac (12,0)]
+		[Export ("localIdentifierMappingsForCloudIdentifiers:")]
+		NSDictionary<PHCloudIdentifier, PHLocalIdentifierMapping> LocalIdentifierMappingsForCloudIdentifiers (PHCloudIdentifier[] cloudIdentifiers);
```
spouliot, line 1184: -> `GetLocalIdentifierMappings`

* prefix with `Get` so there's a verb/action in the method name
* remove `ForCloudIdentifiers` suffix, it's the parameter

src/photos.cs
```diff
@@ -1162,22 +1167,38 @@ interface PHPhotoLibrary {
...
+
+		[Mac (12,0)]
+		[Export ("cloudIdentifierMappingsForLocalIdentifiers:")]
+		NSDictionary<NSString, PHCloudIdentifierMapping> CloudIdentifierMappingsForLocalIdentifiers (string[] localIdentifiers);
```
spouliot, line 1188: -> `GetCloudIdentifierMappings` for the same reasons as above

src/photos.cs
```diff
@@ -1406,13 +1427,14 @@ interface PHProjectChangeRequest {
...
-	[Unavailable (PlatformName.MacCatalyst)]
-	[Advice ("This API is not available when using UIKit on macOS.")]
+	[TV (15,0)]
+	[iOS (15,0)]
 	[BaseType (typeof (NSObject))]
 	[DisableDefaultCtor]
 	interface PHCloudIdentifier : NSSecureCoding {
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosError.IdentifierNotFound' instead.")]
+		[NoTV, NoiOS]
```
spouliot, line 1437: you can add `[NoMacCatalyst]` like the cases above

src/photos.cs
```diff
@@ -146,9 +146,14 @@ interface PHAsset {
 		[Export ("playbackStyle", ArgumentSemantic.Assign)]
 		PHAssetPlaybackStyle PlaybackStyle { get; }
 
+		[Deprecated (PlatformName.MacOSX, 12, 0, message: "Use 'PHPhotosError.IdentifierNotFound' instead.")]
 		[NoTV][NoiOS]
```
spouliot, line 150: you can add `[NoMacCatalyst]` like the cases below

src/Photos/Enums.cs
```diff
@@ -356,7 +358,9 @@ public enum PHProjectSectionType : long {
 	[Native]
```
spouliot, line 358: you can add `[NoMacCatalyst]` since it's deprecated

src/Photos/Enums.cs
```diff
@@ -182,6 +182,8 @@ public enum PHAssetCollectionSubtype : long {
 		SmartAlbumLongExposures = 215,
 		[iOS (13,0)][TV(13,0)][Mac (10,15)]
 		SmartAlbumUnableToUpload = 216,
+		[iOS (15,0), TV (15,0), Mac (12,0)]
```
spouliot, line 185: looks like it should also have `[MacCatalyst (15,0)]`

spouliot (changes requested):

spouliot (approved):

mandel-macaque (approved):

Issue #12150, opened 2021-07-20
Attribute-based code injection for bindings should not, by default, be considered optimizable


spouliot:
### Steps to Reproduce

1. Create (or update) a binding project
2. Add `[Dispose ("...")]` on a member
3. Make `...` do some weird stuff

note: `SnippetAttribute` is the base class of `PreSnippetAttribute`, `PrologueSnippetAttribute`, `PostSnippetAttribute` and `DisposeAttribute`.

However the generated code (for the member) will be decorated with:

```csharp
[BindingImpl (BindingImplOptions.GeneratedCode | BindingImplOptions.Optimizable)]
```

* `BindingImplOptions.GeneratedCode` is partially true. It's generated but the author decided to inject the `...` custom code.

* `BindingImplOptions.Optimizable` is unknown. It might be, it might not - and that might change over time.

### Expected Behavior

Binding code should work, optimized or not.

The correct (safe) way to handle this is to default to `[BindingImpl (BindingImplOptions.GeneratedCode)]` when user code is included. It's _partially_ [re-]generated and the additional code must satisfy any of the runtime requirement of the method.

However we can't say (automatically) that it's optimizable and, by default, it should not be (even more as those optimizations might change so it's basically only safe to assume optimizable inside platform code, i.e. not user code).

Those attributes should be extended to have an optional `bool optimizable` property (default being false).

### Actual Behavior

`BindingImplOptions.Optimizable` is present - but the user code might not be optimizable and that can make the tooling do incorrect things (breaking the build or create issues at runtime).

### Environment

`main`

PR #12165, opened 2021-07-21
[generator] Do not generate `BindingImplOptions.Optimizable` when code snippets are provided

spouliot:
unless the snippet-based attribute also set the `Optimizable` property.

IOW this is now required to _opt-in_ (to optimize) instead of default,
with no ~easy~ way to turn it off.

Re-enabled each of the `[Dispose]` cases after protecting the required
extra calls they make.

Fix https://github.com/xamarin/xamarin-macios/issues/12150


spouliot:
Some tests needs to be added (none seems to exists for the snippets).

src/CoreAnimation/CALayer.cs
```diff
@@ -103,6 +103,8 @@ void SetCALayerDelegate (CALayerDelegate? value)
 			}
 		}
 
+		// Note: preserving this member allows us to re-enable the `Optimizable` binding flag
```
rolfbjarne, line 106: Why would preserving the OnDispose method make the calling method optimizable? Or am I missing something?

rolfbjarne (approved):

src/CoreAnimation/CALayer.cs
spouliot: The (dotnet) processing for (`Optimizable`)`Dispose` method body is skipped to avoid marking the backing fields.

When the bodies are re-introduced it's too late [1] to mark their content again. That works well for purely generated `Dispose` methods (since the `base.Dispose` is part of `IDisposable` it's already marked) and we check if the fields are marked (or not).

However if code (like calls to other methods) are injected then they might not be marked - so those extra pieces (like `OnDispose` here) have to be preserved (or the linker won't be able to save the _invalid_ assembly).

[1] there's something else I want to try... but it's more complex/time consuming than fixing the 3 usages we currently require

dalexsoto (approved):

PR #12181, opened 2021-07-22
[SoundAnalysis] Updates for Xcode13 Beta1

tj-devel709:


src/soundanalysis.cs
```diff
@@ -33,6 +34,14 @@ enum SNErrorCode : long {
 		InvalidFile,
 	}
 
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
```
spouliot, line 37: is this also new in MacCatalyst ?
if not -> `[NoMacCalayst]`
if so -> [MacCatalyst (15,0)]`

src/soundanalysis.cs
```diff
@@ -123,6 +137,22 @@ interface SNClassifySoundRequest : SNRequest {
...
+		[Export ("initWithClassifierIdentifier:error:")]
+		IntPtr Constructor (string classifierIdentifier, [NullAllowed] out NSError error);
+
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Export ("knownClassifications", ArgumentSemantic.Copy)]
+		string[] KnownClassifications { get; }
+
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Export ("windowDuration", ArgumentSemantic.Assign)]
+		CMTime WindowDuration { get; set; }
```
spouliot, line 151: That won't work, as expected, with watchOS since you added the following on the top
```
#if WATCH
 using CMTimeRange = Foundation.NSObject;
 using CMTime = Foundation.NSObject;
```

src/soundanalysis.cs
```diff
@@ -153,4 +183,31 @@ interface SNResultsObserving {
...
+		IntPtr Constructor (NSValue[] enumeratedDurations);
+
+		[Export ("initWithDurationRange:")]
+		IntPtr Constructor (CMTimeRange durationRange);
+
+		[Export ("type", ArgumentSemantic.Assign)]
+		SNTimeDurationConstraintType Type { get; }
+
+		[Export ("enumeratedDurations", ArgumentSemantic.Strong)]
+		NSValue[] EnumeratedDurations { get; }
```
rolfbjarne, line 208: The `enumeratedDurations` property is defined to be "an array of CMTime structures (boxed in NSValue instances)", so you can use `BindAs` here to expose it as an array of CMTime.

rolfbjarne (changes requested):

src/soundanalysis.cs
```diff
@@ -153,4 +183,31 @@ interface SNResultsObserving {
...
+		Version1,
+	}
+
+	[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface SNTimeDurationConstraint /* privately conforms to NSCoding, NSCopying, and NSSecureCoding */
+	{
+		[Export ("initWithEnumeratedDurations:")]
+		IntPtr Constructor (NSValue[] enumeratedDurations);
```
spouliot, line 199: `@param enumeratedDurations A discrete set of duration values (represented as CMTime values boxed in NSValue instances) permitted by this constraint.`

Have you tried `[BindAs]` ? not sure if it handle this case...

src/soundanalysis.cs
spouliot, line 208: same

spouliot (changes requested):

src/soundanalysis.cs
tj-devel709: Yeah I tried this and showed @dalexsoto and he came to the conclusion that CMTime arrays are not supported yet

src/soundanalysis.cs
tj-devel709: I tried this and showed @dalexsoto and he came to the conclusion that CMTime arrays are not supported yet

src/soundanalysis.cs
tj-devel709: Good call, this should support MacCatalyst!

src/soundanalysis.cs
rolfbjarne: We have tests for BindAs with CMTime arrays:

https://github.com/xamarin/xamarin-macios/blob/ffc17b6ce287620b7cd5785524ebf607d69b09c1/tests/generator/bindastests.cs#L102-L104

tj-devel709: I'll give it another shot!

src/soundanalysis.cs
tj-devel709: Okay there was a comment above it from 2 years ago that said remove the 
```
#if WATCH
 using CMTimeRange = Foundation.NSObject;
```
 when watch is enabled in CoreMedia, which the headers appear to be. So I removed this #if else.
 But I should check in xcode and make sure I can call CoreMedia in a watch project.

tj-devel709: Okay, looks like I can call CoreMedia in a watch project!

src/soundanalysis.cs
tj-devel709: So after making this change: https://github.com/xamarin/xamarin-macios/pull/12181/files#diff-771cfc4720d50e07d24a25397a8721722a43b16db581b2cb4259de08f0a61414L15

Adding these BindAs's gives me these errors: https://gist.github.com/tj-devel709/7a2427e85b5fdd310505638decd8b34d
I am pretty confused on why it is trying to convert from CoreMedia.CMTime[] to Foundation.NSObject though

rolfbjarne: Can you gist `build/watch/watch/SoundAnalysis/SNTimeDurationConstraint.g.cs` too?

tj-devel709: https://gist.github.com/tj-devel709/dfea500a2c3bfa0c316523e7ce2b74b6

rolfbjarne: I believe that's because `CMTime` didn't used to be in watchOS, so we defined it to be `NSObject`, like this:

https://github.com/xamarin/xamarin-macios/blob/478b58897fc0b30920ac225750c1d4d4f20343fd/src/foundation.cs#L88

That wasn't a problem while `CMTime` wasn't actually used in any watchOS API, but now that it is (and also because `CMTime` was introduced to watchOS in watchOS 6.0), the fact that we define it to be `NSObject` is causing these weird errors.

The fix is to expose `CMTime` in watchOS as well, and remove these hacks defining `CMTime` as `NSObject` - and the same is true for `CMTimeMapping` and `CMTimeRange` just below:

https://github.com/xamarin/xamarin-macios/blob/478b58897fc0b30920ac225750c1d4d4f20343fd/src/foundation.cs#L89-L90

If would probably be best to do that in a different PR, and once that PR is merged into main, then merge main into this PR, and the BindAs attribute change should compile just fine.

tj-devel709: Thanks for the rundown! I will create that PR now!

tj-devel709: Okay, PR is now created: https://github.com/xamarin/xamarin-macios/pull/12280

mandel-macaque: I have fixed this issue in the last push.

mandel-macaque (approved):

spouliot (approved):

rolfbjarne (approved):

PR #12185, opened 2021-07-22
[AppKit] Move the rest of the enums to the API definition.

rolfbjarne:
This makes it easier to include/exclude certain enums in Mac Catalyst (since
we can use [NoMacCatalyst] instead of littering the files with "#if
__MACCATALYST__").

There should be no API changes at all.


spouliot (commented):
if there's only enums then APPKIT_API_SOURCE could be used, right ?

rolfbjarne:
> if there's only enums then APPKIT_API_SOURCE could be used, right ?

Hm, interesting idea.

It's not only enums in AppKit/Enums.cs: 

https://github.com/xamarin/xamarin-macios/blob/14ed8e4305a25eff79ad25e7bcf8155fcf9b290a/src/AppKit/Enums.cs#L31-L36

but that could be moved out of the file instead.

I'll try and see what happens.

tj-devel709 (approved):

rolfbjarne:
Superseded by #12191.

PR #12189, opened 2021-07-23
[ShazamKit] Updates Xcode13 beta1

tj-devel709:
I am hitting this test failure running Intro locally on iOS and I am not sure how to handle this
image


src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+
+		[Export ("addCustomCatalogFromURL:error:")]
+		bool Add (NSUrl url, [NullAllowed] out NSError error);
+
+		[Export ("writeToURL:error:")]
+		bool WriteL (NSUrl url, [NullAllowed] out NSError error);
+
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
+		[Export ("new")]
+		[return: Release]
```
tj-devel709, line 85: Xtro suggested this. Want to make sure this gets seen!

rolfbjarne (changes requested):

rolfbjarne:
I think @mandel-macaque had a similar intro test failure, but I don't remember what the solution was.

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+	{
+		[Export ("mediaItems", ArgumentSemantic.Strong)]
+		SHMatchedMediaItem[] MediaItems { get; }
+
+		[Export ("querySignature", ArgumentSemantic.Strong)]
+		SHSignature QuerySignature { get; }
+	}
+
+	[iOS (15,0), Mac (12,0), Watch (8,0), TV (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (SHMediaItem))]
```
spouliot, line 102: there's no way to set the values for the properties, that's usually because the instance are created by the OS (and not the user code). 
IOW `	[DisableDefaultCtor]`

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+
+		[NullAllowed, Export ("isrc")]
+		string Isrc { get; }
+
+		[Export ("mediaItemWithProperties:")]
+		SHMediaItem MediaItem (NSDictionary<NSString, NSObject> properties);
+
+		[Async]
+		[Export ("fetchMediaItemWithShazamID:completionHandler:")]
+		void FetchMediaItem (string shazamID, Action<SHMediaItem, NSError> completionHandler);
```
spouliot, line 161: -> `shazamId`

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+
+		[Export ("valueForProperty:")]
+		NSObject Value (string property);
```
spouliot, line 164: -> `GetValue`
in C# methods should have a verb/action

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+		NSObject ObjectForKeyedSubscript (string key);
+	}
+
+	[iOS (15,0), Mac (12,0), Watch (8,0), TV (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface SHMediaLibrary
+	{
+		[Export ("defaultLibrary", ArgumentSemantic.Strong)]
+		SHMediaLibrary DefaultLibrary { get; }
```
spouliot, line 176: missing `[Static]` ?
https://developer.apple.com/documentation/shazamkit/shmedialibrary/3747200-defaultlibrary?language=objc shows `class`

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+
+		[Async]
+		[Export ("addMediaItems:completionHandler:")]
+		void Add (SHMediaItem[] mediaItems, Action<NSError> completionHandler);
+	}
+
+	[iOS (15,0), Mac (12,0), Watch (8,0), TV (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 184: -> `[DisableDefaultCtor]` ?
the `.ctor` does not accept a `null` value and `Catalog` is read-only

src/shazamkit.cs
```diff
@@ -0,0 +1,250 @@
...
+
+		[Export ("matchStreamingBuffer:atTime:")]
+		void MatchStreamingBuffer (AVAudioPcmBuffer buffer, [NullAllowed] AVAudioTime time);
+
+		[Export ("matchSignature:")]
+		void Match (SHSignature signature);
+	}
+
+	[iOS (15,0), Mac (12,0), Watch (8,0), TV (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 208: same, `[DisableDefaultCtor]`
the `.ctor` does not allow `null` `dataRepresentation`
and the `DataRepresentation` property does not allow null, nor allow you to set your own value

tests/xtro-sharpie/common-ShazamKit.ignore
```diff
@@ -0,0 +1,5 @@
+# Intro says these selectors should not be Static
+!missing-selector! +SHMediaItem::fetchMediaItemWithShazamID:completionHandler: not bound
+!missing-selector! +SHMediaItem::mediaItemWithProperties: not bound
+!missing-selector! +SHMediaLibrary::defaultLibrary not bound
+!missing-selector! +SHSignature::signatureWithDataRepresentation:error: not bound
```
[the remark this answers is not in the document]
spouliot: please file an issue/feedback with Apple
`+SHMediaLibrary::defaultLibrary` is a common pattern and only make sense if it's static

spouliot (changes requested):
Intro failure wrt `SHCatalog` is likely because it's an abstract base class
https://developer.apple.com/documentation/shazamkit/shcatalog?changes=_4&language=objc

src/shazamkit.cs
tj-devel709: This is what I am currently confused about.
I had static on these selectors:
```
SHMediaItem::fetchMediaItemWithShazamID:completionHandler:
SHMediaItem::mediaItemWithProperties:
SHMediaLibrary::defaultLibrary
SHSignature::signatureWithDataRepresentation:error:
```
but Intro on iOS complains and says that these should not actually be static.
So I removed them and add to the ignore file, but then the TV intro says that they should be static?

src/shazamkit.cs
```diff
@@ -0,0 +1,253 @@
...
+		[NullAllowed, Export ("artworkURL", ArgumentSemantic.Strong)]
+		NSUrl ArtworkUrl { get; }
+
+		[NullAllowed, Export ("videoURL", ArgumentSemantic.Strong)]
+		NSUrl VideoUrl { get; }
+
+		[Export ("explicitContent")]
+		bool ExplicitContent { get; }
+
+		[NullAllowed, Export ("isrc")]
```
mandel-macaque, line 154: Possible, but I think @spouliot had a point in leaving the export alone rather than combining it with other attributes.

src/shazamkit.cs
mandel-macaque: Definitely static, what is the intro erro on iOS?

mandel-macaque (changes requested):
Small fixes.

src/shazamkit.cs
tj-devel709:
```suggestion
		[NullAllowed] [Export ("isrc")]
```
Do you mean doing something like this?

tj-devel709: Do you mean doing something like ```[NullAllowed][Export ("isrc")]```?

src/shazamkit.cs
tj-devel709: Let me pull it up!

tj-devel709: image

spouliot: Try using them in an ObjC app
They might just not be "responding" but might work at runtime
If they don't work file an issue with Apple and attach your sample

src/shazamkit.cs
mandel-macaque: Like all the other bindings, one line for just the export.

src/shazamkit.cs
tj-devel709: So I tried these out [here](https://gist.github.com/tj-devel709/2911a893c5105c31b307900221ca657d) 
and it looks like SHCustomCatalog::new and SHMediaLibrary::defaultLibrary are working, but the rest are not recognized!
I'll create an issue with Apple

tj-devel709: Just to confirm, I should file it here? www.feedbackassistant.apple.com

spouliot: @tj-devel709 yes

tj-devel709: Okay, I have filed an issue with apple!

tj-devel709: If useful here, this is the sample I gave to Apple.
[ShazamStaticTest.zip](https://github.com/xamarin/xamarin-macios/files/6880406/ShazamStaticTest.zip)

tj-devel709: Should I make comments about the ones that are not behaving as static for now or wait until the next beta and see if the issue is resolved?

spouliot: b4 is out now :)
but, to get things rolling, you can always comment what's incorrect (or unsure) and file an issue (and add it to the checklist)

src/shazamkit.cs
```diff
@@ -0,0 +1,268 @@
...
+		[Export ("explicitContent")]
+		bool ExplicitContent { get; }
+
+		[NullAllowed]
+		[Export ("isrc")]
+		string Isrc { get; }
+
+		[Static]
+		[Export ("mediaItemWithProperties:")]
+		SHMediaItem MediaItem (NSDictionary<NSString, NSObject> properties);
```
mandel-macaque, line 170: Missing a verb, all methods in c# should start with one.

src/shazamkit.cs
```diff
@@ -0,0 +1,268 @@
...
+		ISHSessionDelegate Delegate { get; set; }
+
+		[NullAllowed, Export ("delegate", ArgumentSemantic.Weak)]
+		NSObject WeakDelegate { get; set; }
+
+		[Export ("initWithCatalog:")]
+		IntPtr Constructor (SHCatalog catalog);
+
+		[Export ("matchStreamingBuffer:atTime:")]
+		void MatchStreamingBuffer (AVAudioPcmBuffer buffer, [NullAllowed] AVAudioTime time);
```
mandel-macaque, line 217: Maybe move to Match, like the other method.

tests/introspection/ApiSelectorTest.cs
```diff
@@ -840,6 +840,35 @@ protected virtual bool Skip (Type type, string selectorName)
 				if (selectorName == "waitUntilExit")
 					return true;
 				break;
+			case "SHCustomCatalog":
+				// ShazamKit has some issues running on simulator
+				switch (selectorName) {
+				case "new":
+					return true;
+				}
+				break;
```
mandel-macaque, line 849: If you look at others, we have code that will return true only if in the simulator, you should do the same.

tests/introspection/ApiSelectorTest.cs
```diff
@@ -840,6 +840,35 @@ protected virtual bool Skip (Type type, string selectorName)
...
+			case "SHMediaItem":
+				// ShazamKit has some issues running on simulator
+				switch (selectorName) {
+				case "fetchMediaItemWithShazamID:completionHandler:":
+				case "mediaItemWithProperties:":
+					return true;
```
mandel-macaque, line 855: same.

tests/introspection/ApiSelectorTest.cs
```diff
@@ -840,6 +840,35 @@ protected virtual bool Skip (Type type, string selectorName)
...
+				}
+				break;
+			case "SHMediaLibrary":
+				// ShazamKit has some issues running on simulator
+				switch (selectorName) {
+				case "defaultLibrary":
+					return true;
```
mandel-macaque, line 862: same.

tests/introspection/ApiSelectorTest.cs
```diff
@@ -840,6 +840,35 @@ protected virtual bool Skip (Type type, string selectorName)
...
+				}
+				break;
+			case "SHSignature":
+				// ShazamKit has some issues running on simulator
+				switch (selectorName) {
+				case "signatureWithDataRepresentation:error:":
+					return true;
```
mandel-macaque, line 869: same

mandel-macaque (changes requested):

tests/introspection/ApiSelectorTest.cs
tj-devel709: Ah good call, thank you!

src/shazamkit.cs
tj-devel709: Whoops! Good call!

src/shazamkit.cs
tj-devel709: I was very wrong. I just was not calling the methods the correct way in xcode!

mandel-macaque: @tj-devel709 so the methods do work, correct? we should close the maccore issue you created then.

mandel-macaque (approved):
Conflicts need to be fixed.

tests/introspection/ApiSelectorTest.cs
```diff
@@ -840,12 +840,14 @@ protected virtual bool Skip (Type type, string selectorName)
 				if (selectorName == "waitUntilExit")
 					return true;
 				break;
+#if !__MACOS__
```
mandel-macaque, line 843: If this is just on iOS, this is not the correct file, it should be in tests/introspection/iOS/iOSApiSelectorTest.cs

mandel-macaque (changes requested):

tj-devel709:
> I think @mandel-macaque had a similar intro test failure, but I don't remember what the solution was.

@mandel-macaque Does the intro failure at the very top of this PR look familiar to you?

tests/introspection/ApiSelectorTest.cs
tj-devel709: I believe that it was having issues with TV and Watch simulators as well

mandel-macaque: watchOS and tv will go in the iOS entry since they are generated from it.

mandel-macaque:
@tj-devel709 if we are talking about: 
```
VerifyClassPtr: class_ptr and RegisterAttribute are different: SHCatalog
```

The fix is to just run it the device and not the simulator, on the device you will have no issues. Have you ran the tests on a device before pushing to the branch?

tools/common/Frameworks.cs
```diff
@@ -427,6 +428,7 @@ public static Frameworks CreateiOSFrameworks (bool is_simulator_build)
 				{ "Chip", "CHIP", new Version (15, 0), NotAvailableInSimulator /* no headers in beta 2 */ },
 				{ "Phase", "PHASE", new Version (15,0), NotAvailableInSimulator /* no headers in beta 2 */ },
 				{ "OSLog", "OSLog", 15,0 },
+				{ "ShazamKit", "ShazamKit", 15,0 },
```
mandel-macaque, line 431: Look at the CHIP framework, we need to add the NotAvailableInSimulator version.

tools/common/Frameworks.cs
```diff
@@ -498,6 +500,7 @@ public static Frameworks GetwatchOSFrameworks (bool is_simulator_build)
 				{ "Chip", "CHIP", new Version (8, 0), NotAvailableInSimulator /* no headers in beta 2 */ },
 				{ "NearbyInteraction", "NearbyInteraction", 8,0 },
 				{ "OSLog", "OSLog", 8,0 },
+				{ "ShazamKit", "ShazamKit", 8,0 },
```
mandel-macaque, line 503: same

tools/common/Frameworks.cs
```diff
@@ -589,6 +592,7 @@ public static Frameworks GetwatchOSFrameworks (bool is_simulator_build)
 
 					{ "Chip", "CHIP", new Version (15, 0), NotAvailableInSimulator /* no headers in beta 2 */ },
 					{ "OSLog", "OSLog", 15,0 },
+					{ "ShazamKit", "ShazamKit", 15,0 },
```
mandel-macaque, line 595: same

mandel-macaque (changes requested):

rolfbjarne (approved):

dalexsoto (approved):

mandel-macaque (approved):

src/shazamkit.cs
```diff
@@ -0,0 +1,267 @@
...
+		bool Add (SHSignature signature, SHMediaItem[] mediaItems, [NullAllowed] out NSError error);
+
+		[Export ("addCustomCatalogFromURL:error:")]
+		bool Add (NSUrl url, [NullAllowed] out NSError error);
+
+		[Export ("writeToURL:error:")]
+		bool Write (NSUrl url, [NullAllowed] out NSError error);
+
+		[Static]
+		[Watch (8,0), TV (15,0), Mac (12,0), iOS (15,0)]
```
spouliot, line 84: ^ that's a (incomplete) duplicate of the attributes on the type

spouliot (approved):

src/shazamkit.cs
tj-devel709: Good catch! Removing now!

tj-devel709:
Unrelated test failue: https://github.com/xamarin/maccore/issues/2443

PR #12191, opened 2021-07-23
[AppKit] Make the files containing enums API source files instead of Core source files.

rolfbjarne:
This makes it easier to include/exclude certain enums in Mac Catalyst (since
we can use [NoMacCatalyst] instead of littering the files with "#if
MACCATALYST").

It required moving a little bit of unrelated code out of the enums source
files, but this is a much smaller change than moving all of the enums out of
these files (see #12185 for how that ends up).

### DIFF
--- a/src/AppKit/Compat.cs
+++ b/src/AppKit/Compat.cs
@@ -40,4 +40,13 @@ namespace AppKit {
 		}
 #endif
 	}
+
+#if !XAMCORE_4_0
+	public static class NSFileTypeForHFSTypeCode
+	{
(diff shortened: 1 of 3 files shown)


dalexsoto (approved):

rolfbjarne:
Test failures are unrelated

* monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2471
* framework-test/Mac Catalyst/Debug: https://github.com/xamarin/maccore/issues/2414

spouliot (approved):
👍

PR #12192, opened 2021-07-23
[monotouch-test] Fix CalendarTest.TestOrdinality to use UTC instead of local datetimes. Fixes #xamarin/maccore@2471.

rolfbjarne:
This is a bit of a shot in the dark, because I can't reproduce the issue, however:

* The test only fails on Mac Catalyst / .NET.
* The only place we're missing globalization data (such as time zone info) is on Mac Catalyst / .NET.
* The test needs time zone info, because it's converting a date from the local time zone to UTC.
* The test sometimes succeeds, sometimes fails (on the bots).
* I'm not in the same time zone as the bots.

My guess is that the lack of time zone info means that at certain times of the
day the test will fail, and exactly which times depend on the current time
zone where the test is executed.

Modifying the test to use UTC datetimes instead of local datetimes should make
the problem independent of the local time zone, and always have the same
behavior. And the test still passes for me, so I'm hoping it'll fix the test
on the bots too.

Fixes https://github.com/xamarin/maccore/issues/2471.


dalexsoto (approved):

spouliot (approved):

rolfbjarne:
Of course now it fails everywhere _except_ Mac Catalyst on .NET 😒

rolfbjarne:

Issue #6573, opened 2019-07-12
[Foundation] NSWritingDirection puzzling situation...


dalexsoto:
Apple decided to rename `UITextWritingDirection` to `NSWritingDirection` in iOS 13 [see docs](https://developer.apple.com/documentation/uikit/uitextwritingdirection), this uncovered a very puzzling situation, it turns out `NSWritingDirection` already exists in

AppKit 

https://github.com/xamarin/xamarin-macios/blob/06426bd7a0c5e371dc14bddf00ab68bbeaecfd9f/src/Foundation/Enum.cs#L793-L801

and Foundation

https://github.com/xamarin/xamarin-macios/blob/7b83195937b96d22b3fe9e83d13818da4e99c9b8/src/AppKit/Enums.cs#L1292-L1298

To make things worse the AppKit version has 2 extra members for no apparent reason ([see docs](https://developer.apple.com/documentation/appkit/nswritingdirection)).

Quoting @spouliot 
> @dalexsoto problem will be a new framework using `NSWritingDirection` - it will create incompatible API between platforms (if using facades) because of namespaces

So this is something we'll need to fix by:

1. Keeping `NSWritingDirection` inside Foundation. #12199
2. Remove deprecation notice from `Foundation` and deprecate the `AppKit` one. #12199
3. Find all API using `AppKit.NSWritingDirection` and adjust accordingly
4. Figure out if we want to keep those 2 extra members from `Appkit.NSWritingDirection` in `Foundation.NSWritingDirection` but make sure we **exclude** them from iOS, watchOS and tvOS if so.
5. Find all API that uses `UITextWritingDirection` and adjust it to use `Foundation.NSWritingDirection`

PR #12199, opened 2021-07-23
[AppKit/UIKit/Foundation] Use NSWritingDirection from Foundation. Partial fix for #6573.

rolfbjarne:
Make the NSWritingDirection enum from Foundation the main version, and
deprecate the others (AppKit.NSWritingDirection,
UIKit.UITextWritingDirection).

This makes things a bit easier with Mac Catalyst, which ships both AppKit and UIKit.

Ref: https://github.com/xamarin/xamarin-macios/issues/6573

### DIFF
--- a/src/AppKit/Enums.cs
+++ b/src/AppKit/Enums.cs
@@ -1225,16 +1225,21 @@ namespace AppKit {
 		Natural = 4
 	}
 
+#if !XAMCORE_4_0 && MONOMAC
+	// Not hard deprecating now but until XAMCORE_4_0 happens or we can
+	// properly fix all the API using this.
+	// see: https://github.com/xamarin/xamarin-macios/issues/6573
+	// [Obsolete ("Use NSWritingDirection in Foundation instead.")]
 	[Flags]
 	[Native]
 	public enum NSWritingDirection : long {
 		Natural = -1, LeftToRight, RightToLeft,
+		[Obsolete ("Use 'LeftToRight' instead.")]
 		Embedding = 0,
-#if !XAMCORE_4_0
 		[Obsolete ("This API is not available on this platform.")]
 		Override = 2,
-#endif
 	}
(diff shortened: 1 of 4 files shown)


src/UIKit/UIEnums.cs
```diff
@@ -1155,7 +1155,7 @@ public enum UITextDirection : long {
 	// Not hard deprecating now but until XAMCORE_4_0 happens or we can
 	// properly fix all the API using this.
 	// see: https://github.com/xamarin/xamarin-macios/issues/6573
-	// Use Foundation.NSWritingDirection instead.
+	// [Obsolete ("Use NSWritingDirection in Foundation instead.")]
```
spouliot, line 1158: uncomment ?

src/AppKit/Enums.cs
```diff
@@ -1225,16 +1225,21 @@ public enum NSTextAlignment : ulong {
 		Natural = 4
 	}
 
+#if !XAMCORE_4_0 && MONOMAC
+	// Not hard deprecating now but until XAMCORE_4_0 happens or we can
+	// properly fix all the API using this.
+	// see: https://github.com/xamarin/xamarin-macios/issues/6573
+	// [Obsolete ("Use NSWritingDirection in Foundation instead.")]
```
spouliot, line 1232: uncomment ?

spouliot (approved):

src/UIKit/UIEnums.cs
rolfbjarne: We have APIs that expose this enum (and the one below), so if we uncomment this, then every consumer of those APIs will get this warning (and there's no way to fix them except ignoring the warning).

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2471).

PR #12200, opened 2021-07-23
[msbuild] Unify the Archive task between iOS and Mac.

rolfbjarne:
The iOS version and the Mac version were slightly different in that they were
adding different things to the archive, but both seemed to be resilient to
those files not existing, so I just merged both implementations to try to add
everything to the archive.


emaf (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2471).

PR #12043, opened 2021-06-29
[AuthenticationServices] Updates for Xcode13 Beta3

tj-devel709:
Hi team, I am having a rough time with this binding..

Here are the Xtro errors I am currently getting:
On iOS, Mac, and MacCatalyst: https://gist.github.com/tj-devel709/b70a6205ff785d0faecede874f6f0d22
On tvOS: !unknown-type! ASAuthorizationCustomMethod bound

## The p-invoke
- !missing-pinvoke! ASAuthorizationAllSupportedPublicKeyCredentialDescriptorTransports is not bound

I attempted to create the partial class 'PublicPrivateKeyAuthentication' and added it to frameworks.sources.

- Where can I find information about what should go in the DllIimport?
- From the Xtro Error, it looks like it is not recognizing this class. Is there something else I am missing?

## The Protocol Conformance (Ignore this one for right now, need to talk to Alex about specifics in 1:1!)

- !missing-protocol-conformance! ASAuthorizationPlatformPublicKeyCredentialAssertionRequest should conform to ASAuthorizationPublicKeyCredentialAssertionRequest
- !missing-protocol-conformance! ASAuthorizationPlatformPublicKeyCredentialDescriptor should conform to ASAuthorizationPublicKeyCredentialDescriptor
- !missing-protocol-conformance! ASAuthorizationSecurityKeyPublicKeyCredentialAssertionRequest should conform to ASAuthorizationPublicKeyCredentialAssertionRequest
- !missing-protocol-conformance! ASAuthorizationSecurityKeyPublicKeyCredentialDescriptor should conform to ASAuthorizationPublicKeyCredentialDescriptor

I created empty interfaces of these protocol interfaces and tried to use them, but it doesn't seem to solve this issue.

## The Unknown Types
- !unknown-type! ASAuthorizationPublicKeyCredentialAttestationKind bound
- !unknown-type! ASAuthorizationPublicKeyCredentialResidentKeyPreference bound
- !unknown-type! ASAuthorizationPublicKeyCredentialUserVerificationPreference bound
- !unknown-type! ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport bound
- !unknown-type! ASAuthorizationCustomMethod

I created enums for these elements following the names xcode gives. For example, 
```
API_AVAILABLE(macos(12.0), ios(15.0)) API_UNAVAILABLE(watchos)
[...]

Issue #11883, opened 2021-06-10
Xcode 13: list of changes to be reviewed/reverted before final release


rolfbjarne:
* [x] ~[[SimInstaller] Xcode13.0 cannot install simulators](https://github.com/xamarin/xamarin-macios/issues/11881)~
* [ ] [WKPreferences isTextInteractionEnabled is not correctly bound on xcode13.](https://github.com/xamarin/xamarin-macios/issues/11955)
* [ ] [[AuthenticationServices] StronglyTypedNSArray Recommendation](https://github.com/xamarin/xamarin-macios/issues/12044)
* [x] [[AuthenticationServices] SecKey Not Usable inside AuthenticationServices Binding](https://github.com/xamarin/maccore/issues/2482)
* [ ] [[QuickLook] QLPreviewSceneActivationConfiguration not yet bound in Xcode13](https://github.com/xamarin/xamarin-macios/issues/12335)
* [ ] [[AudioToolbox] Some AudioToolbox APIs not yet implemented in Xcode13](https://github.com/xamarin/xamarin-macios/issues/12489)

dalexsoto:
Nothing here was a breaking change but enhancements, @mandel-macaque and myself reviewed the list and decided this issue is good to go.

PR #12203, opened 2021-07-23
[AuthenticationServices] Updates for Xcode13 Beta3

tj-devel709:
Closed https://github.com/xamarin/xamarin-macios/pull/12043 in favor of this PR.
I tried to squash and rebase on top of main for this PR so hopefully I didn't miss any commits!


src/AuthenticationServices/PublicPrivateKeyAuthentication.cs
```diff
@@ -0,0 +1,33 @@
...
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+
+#nullable enable
+
+namespace AuthenticationServices {
+
+    [NoWatch, NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	public static class PublicPrivateKeyAuthentication {
```
rolfbjarne, line 23: Indentation is off here.

rolfbjarne (approved):

src/authenticationservices.cs
```diff
@@ -623,6 +636,10 @@ interface ASAuthorizationSingleSignOnCredential : ASAuthorizationCredential {
 
 		[NullAllowed, Export ("authenticatedResponse", ArgumentSemantic.Copy)]
 		NSHttpUrlResponse AuthenticatedResponse { get; }
+
+		[Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+		[Export ("privateKeys")]
+		NSObject[] PrivateKeys { get; }
```
spouliot, line 642:
```
+/*! @abstract Private SecKeys returned from the AuthenticationServices extension.
+ */
+@property (nonatomic, readonly) NSArray *privateKeys API_AVAILABLE(ios(15.0), macos(12.0)) API_UNAVAILABLE(tvos, watchos);
```

`SecKey` is a type - so it should be an array of them

src/authenticationservices.cs
```diff
@@ -651,6 +668,10 @@ interface ASAuthorizationSingleSignOnRequest {
 
 		[Export ("authorizationOptions", ArgumentSemantic.Copy)]
 		NSUrlQueryItem[] AuthorizationOptions { get; set; }
+
+		[NoWatch, NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
```
spouliot, line 672: minor: there's already a `[NoWatch, NoTV]` on the type so it's not needed on the members

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+		[Export ("signature", ArgumentSemantic.Copy)]
+		NSData Signature { get; }
+	}
+
+	[NoWatch, Mac (12,0), iOS (15,0), MacCatalyst (15,0), NoTV]
+	[BaseType (typeof (ASAuthorizationRequest))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationPlatformPublicKeyCredentialAssertionRequest : ASAuthorizationPublicKeyCredentialAssertionRequest
+	{
+		[Sealed]
```
spouliot, line 1029: why ? iow worth a comment

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+		[Export ("allowedCredentials", ArgumentSemantic.Copy)]
+		ASAuthorizationPlatformPublicKeyCredentialDescriptor[] PlatformAllowedCredentials { get; set; }
+	}
+
+	[NoWatch, NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (ASAuthorizationRequest))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationSecurityKeyPublicKeyCredentialAssertionRequest : ASAuthorizationPublicKeyCredentialAssertionRequest
+	{
+		[Sealed]
```
spouliot, line 1039: same

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationPlatformPublicKeyCredentialProvider : ASAuthorizationProvider
+	{
+		[Export ("initWithRelyingPartyIdentifier:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (string relyingPartyIdentifier);
+
+		[Export ("createCredentialRegistrationRequestWithChallenge:name:userID:")]
+		ASAuthorizationPlatformPublicKeyCredentialRegistrationRequest CreateCredentialRegistrationRequestWithChallenge (NSData challenge, string name, NSData userId);
```
spouliot, line 1153: remove `WithChallenge` suffix
that's the first argument (not the method name)

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+
+		[Export ("createCredentialAssertionRequestWithChallenge:")]
+		ASAuthorizationPlatformPublicKeyCredentialAssertionRequest CreateCredentialAssertionRequestWithChallenge (NSData challenge);
```
spouliot, line 1156: remove `WithChallenge` suffix
that's the first argument (not the method name)

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+		NSDictionary<NSString, NSString> HttpAuthorizationHeaders { get; set; }
+
+		[NullAllowed, Export ("httpResponse", ArgumentSemantic.Copy)]
+		NSHttpUrlResponse HttpResponse { get; set; }
+
+		[NullAllowed, Export ("httpBody", ArgumentSemantic.Assign)]
+		NSData HttpBody { get; set; }
+
+		[Export ("privateKeys", ArgumentSemantic.Assign)]
+		NSObject[] PrivateKeys { get; set; }
```
spouliot, line 1183: likely `SecKey[]`

src/authenticationservices.cs
```diff
@@ -930,4 +951,342 @@ interface ASAccountAuthenticationModificationViewController {
...
+	[NoWatch, NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface ASAuthorizationPublicKeyCredentialParameters : NSSecureCoding, NSCopying
+	{
+		[Export ("initWithAlgorithm:")]
+		IntPtr Constructor (nint algorithm);
+
+		[Export ("algorithm")]
+		nint Algorithm { get; }
```
spouliot, line 1195:
```
+@property (nonatomic, readonly) ASCOSEAlgorithmIdentifier algorithm;
```

why not use the type ?

src/AuthenticationServices/PublicPrivateKeyAuthentication.cs
```diff
@@ -0,0 +1,33 @@
...
+        [DllImport (Constants.AuthenticationServicesLibrary)]
+        static extern /* NSString[] */ IntPtr ASAuthorizationAllSupportedPublicKeyCredentialDescriptorTransports ();
+
+        public static /* ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport[] */ NSString[] GetAllSupportedPublicKeyCredentialDescriptorTransports () {
+            return NSArray.ArrayFromHandle<NSString> (ASAuthorizationAllSupportedPublicKeyCredentialDescriptorTransports ());
+        }
```
spouliot, line 29: you have `ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport` defined
there's no point in exposing the list (only) as `NSString`

spouliot (changes requested):

src/AuthenticationServices/PublicPrivateKeyAuthentication.cs
tj-devel709: Whoops, was using spaces!

src/AuthenticationServices/PublicPrivateKeyAuthentication.cs
tj-devel709: Are you suggesting to do something like this?
```
public static ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport[] GetAllSupportedPublicKeyCredentialDescriptorTransports () {
	return NSArray.ArrayFromHandle<ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport> (ASAuthorizationAllSupportedPublicKeyCredentialDescriptorTransports ());
}
```

src/authenticationservices.cs
tj-devel709: Hmm, so I see that SecKey is created in this manual class here src/Security/Certificate.cs. 
The namespace is Security, but when I add 'using Security' the compiler still says that it cannot find the reference.

src/authenticationservices.cs
tj-devel709: I cannot find this type in our bindings using grep

spouliot: it might be new... not sure why xtro did not pick it up
you need to track the header files to see/learn why :)

src/AuthenticationServices/PublicPrivateKeyAuthentication.cs
spouliot: For the API signature ? yes
Not sure the implementation will compile...

src/authenticationservices.cs
tj-devel709: So I actually don't who told me to add the Sealed, perhaps @dalexsoto ?
But when I remove it, I get these build errors: https://gist.github.com/tj-devel709/d459d9e717a186486deab6479f57c050

spouliot: That error is because you're using the same selector for multiple methods/properties.

Why is that called `PlatformAllowedCredentials` if the selector is `allowedCredentials` ?

src/authenticationservices.cs
spouliot: That might require some changes in the generator. It does not seems to allow `SecKey` at the moment.

Still it's important because `SecKey` is **not** a subclass of `NSObject`, so the binding are incorrect without that fix.

src/authenticationservices.cs
tj-devel709: Oh I just remembered what is going on! Alex did help me with this one a few weeks ago!

So there are two different classes that use this similar "allowedCredentials" header: `ASAuthorizationPlatformPublicKeyCredentialAssertionRequest` and `ASAuthorizationSecurityKeyPublicKeyCredentialAssertionRequest`

The return type of the former is `ASAuthorizationPlatformPublicKeyCredentialDescriptor[]`
and the return type of the second one is `ASAuthorizationSecurityKeyPublicKeyCredentialDescriptor[]`

but they both are overriding a property from the protocol `ASAuthorizationPublicKeyCredentialAssertionRequest` 
which has the base property "allowedCredentials" that is of type `IASAuthorizationPublicKeyCredentialDescriptor[]`

So to get past this typing confusion, Alex suggested giving them different names and sealing the properties in the child classes. I definitely should have made a comment because I had already forgotten about this.

src/authenticationservices.cs
tj-devel709: Hmm okay so the headers didn't really reveal anything to me: ```@abstract A COSE algorithm indentifier.```
 but looking at the web docs and looking up what ASCOSEAlgorithmIdentifier reveals that it is a type alias and has this:
```typedef NSInteger ASCOSEAlgorithmIdentifier;```
But I am not sure if this is any better?
https://developer.apple.com/documentation/authenticationservices/ascosealgorithmidentifier?language=objc

Comment (tj-devel709) [reply src/authenticationservices.cs:1198]: I'll keep looking at this after the meeting though

File: src/authenticationservices.cs (hunk 4 above)
Comment (spouliot) [reply src/authenticationservices.cs:1031]: That used to be a problem but C# now have covariant returns
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns

I wonder if _just_ using `[Override]` would work...
or if more change are needed

The main _problem_ with this rename-approach is that it's not easily discoverable by developers - so they might still use the original (base) one and have to deal with casts

File: src/authenticationservices.cs (hunk 9 above)
Comment (spouliot) [reply src/authenticationservices.cs:1198]: It's not how most enums are defined - and it seems sharpie does not process them when defined this way.

```
API_AVAILABLE(macos(12.0), ios(15.0)) API_UNAVAILABLE(watchos)
typedef NSInteger ASCOSEAlgorithmIdentifier NS_TYPED_EXTENSIBLE_ENUM;

/*! @abstract The COSE algorithm identifier representing ECDSA with SHA-256.
 */
API_AVAILABLE(macos(12.0), ios(15.0)) API_UNAVAILABLE(watchos)
static ASCOSEAlgorithmIdentifier const ASCOSEAlgorithmIdentifierES256 = -7;
```

/Applications/Xcode_13.0.0-beta3.app//Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/AuthenticationServices.framework/Headers/ASCOSEConstants.h 

Also check for `ASCOSEEllipticCurveIdentifier` since it might suffer from the same `nine` issue

File: src/authenticationservices.cs (hunk 2 above)
Comment (tj-devel709) [reply src/authenticationservices.cs:643]: Should I create an issue and link that in the binding?

File: src/authenticationservices.cs (hunk 9 above)
Comment (tj-devel709) [reply src/authenticationservices.cs:1198]: AH thanks for finding this! I will grep the header files next time!

Comment (tj-devel709) [reply src/authenticationservices.cs:1198]: Interestingly, ASCOSEEllipticCurveIdentifier is only declared but not used

File: src/AuthenticationServices/PublicPrivateKeyAuthentication.cs (hunk 10 above)
Comment (tj-devel709) [reply src/AuthenticationServices/PublicPrivateKeyAuthentication.cs]: it does not currently compile since ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport is an enum and it complains saying that it needs to be a reference type: 
```
error CS0452: The type 'ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport' must be a reference type in order to use it as parameter 'T' in the generic type or method 'NSArray.ArrayFromHandle<T>(IntPtr)'
```
Should I try turning the enum into an interface so that it can be a reference type?

File: src/authenticationservices.cs (hunk 4 above)
Comment (tj-devel709) [reply src/authenticationservices.cs:1031]: Using just the Override attribute gives me the following [errors](https://gist.github.com/tj-devel709/c4f0b6a35c8fe82332904237032e98e2)

I was wondering if maybe the protocol interface method needs a virtual attribute?
```
[The Sealed Attribute] Instructs the generator to flag the generated method as sealed. 
If this attribute is not specified, the default is to generate a virtual method 
(either a virtual method, an abstract method or an override depending on how other attributes are used).
```
This method is marked abstract and is maybe not virtual by default because of that?

File: src/authenticationservices.cs (hunk 9 above)
Comment (spouliot) [reply src/authenticationservices.cs]: some new (b4?) API might eventually use it
but no need to bind it until we have a need for it :)

File: src/authenticationservices.cs (hunk 2 above)
Comment (spouliot) [reply src/authenticationservices.cs:643]: yes, comment the bindings and create an issue
also add it to the checklist for xcode13

File: src/authenticationservices.cs (hunk 4 above)
Comment (spouliot) [reply src/authenticationservices.cs:1043]: it's not a big deal, let's land this as-is (unless someone else objects) and I might have a have a quick look

File: src/AuthenticationServices/PublicPrivateKeyAuthentication.cs (hunk 10 above)
Comment (spouliot) [reply src/AuthenticationServices/PublicPrivateKeyAuthentication.cs]: no, you just need a bit more code
You already get an array of `NSString` and you can create an array of the `enum` and convert each string constant into it's enum value

File: src/authenticationservices.cs [hunk 11]
```diff
@@ -1002,6 +1003,18 @@ enum ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport {
...
+	[Native]
+	enum ASCOSEAlgorithmIdentifier : long {
+		Es256 = -7,
+	}
+
+	[Mac (12,0), iOS (15,0), MacCatalyst (15,0), NoWatch, NoTV]
+	[Native]
+	enum ASCOSEEllipticCurveIdentifier : long {
+		P256 = 1,
+	}
```
spouliot, line 1016: ^ comment it
why ? it's not yet used in any API and it could possibly change before it's used (and that would be a breaking change for us)

src/authenticationservices.cs
```diff
@@ -1002,6 +1003,18 @@ enum ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport {
 		Bluetooth,
 	}
 
+	[Mac (12,0), iOS (15,0), MacCatalyst (15,0), NoWatch, NoTV]
+	[Native]
+	enum ASCOSEAlgorithmIdentifier : long {
```
spouliot, line 1008: `ASCoseAlgorithmIdentifier`

src/authenticationservices.cs
```diff
@@ -1002,6 +1003,18 @@ enum ASAuthorizationSecurityKeyPublicKeyCredentialDescriptorTransport {
...
+		Es256 = -7,
```
spouliot, line 1009: but `ES256` since two-letters acronyms remains uppercase in .net
3+ letters acronyms become Pascal-cased

spouliot (changes requested):

src/authenticationservices.cs
tj-devel709: Clarification: leave a comment saying this should be SecKey or comment out this whole property and leave a note there about it?

spouliot: * Comment the property binding
* Add a link to the issue you created inside the comment
* Add a link to the issue to https://github.com/xamarin/xamarin-macios/issues/11883

Why ? 
* It's not usable as an `NSObject` (since `SecKey` is not a subclass of it)
* if we forget it (or lack time to update) then we're not creating a future breaking change

It's always easier to add (later) than to remove :) 
IOW never add something "new" that's not right / needs to be fixed

mandel-macaque (approved):

spouliot (approved):

src/authenticationservices.cs
mandel-macaque: Not an issue with the generator. SecKey was in a #if COREBUILD moving the #if to provide the minimum fixed the problem. I have pushed the fix.

mandel-macaque:
Unrelated issues:

* https://github.com/xamarin/maccore/issues/2414 
* https://github.com/xamarin/maccore/issues/2443

PR #12208, opened 2021-07-23
[generator] Skip protocols that aren't available on the current platform.

rolfbjarne:
That way we can do this in api definitions:

    interface MyType : MyProtocol { }

    [Protocol]
    [NoMacCatalyst]
    interface MyProtocol {}

instead of this:

    interface MyType
    #if __MACOS__
        : MyProtocol
    #endif
    {
    }

    [Protocol]
    [NoMacCatalyst]
    interface MyProtocol {}

for a type that implements a protocol on macOS, but not on Mac Catalyst.


spouliot (changes requested):
looks good but should come with a unit test

mandel-macaque (approved):
Looks good, I do agree about the test.

dalexsoto (approved):
+3 on the test :P

spouliot (approved):

PR #12230, opened 2021-07-26
[CI] Run tests on M1 machines with BS.

mandel-macaque:


tools/devops/automation/build-pipeline.yml
```diff
@@ -84,25 +84,39 @@ parameters:
...
     },
     {
       stageName: 'mac_10_14',
       displayName: 'Mac Mojave (10.14)',
       macPool: 'Hosted Mac Internal Mojave',
-      osVersion: '10.14',
       statusContext: 'Mac Mojave (10.14)',
-      checkDemands: false
+      demands: [
+        "Agent.OS -equals Darwin",
```
spouliot, line 99: does not it need something like `"Agent.OSVersion -equals '10.14'"` ? 
or is the `macPool` enough to identify them ?

tools/devops/automation/build-pipeline.yml
```diff
@@ -84,25 +84,39 @@ parameters:
...
     },
     {
       stageName: 'mac_10_13',
       displayName: 'Mac High Sierra (10.13)',
       macPool: 'Hosted Mac Internal',
-      osVersion: '10.13',
       statusContext: 'Mac High Sierra (10.13)',
-      checkDemands: false
+      demands: [
+        "Agent.OS -equals Darwin",
```
spouliot, line 108: same

spouliot (approved):

tools/devops/automation/build-pipeline.yml
mandel-macaque: In this case, we do not need to do it because the agent did not set it, meaning that if we add it we never get a match (bad configured agent). But the pools only contain machines of that version.

tools/devops/automation/build-pipeline.yml
mandel-macaque: Same, bad configured bots :/

dalexsoto (approved):

rolfbjarne (approved):
Looks good, except that there doesn't seem to be any results from the M1 bot.

PR #12234, opened 2021-07-26
[tests] Fix introspection on macOS 12

spouliot:
This was fixed for iOS 15 but also happens on macOS 12 so the case
was moved into the common, shared test sources.


dalexsoto (approved):

mandel-macaque (approved):

spouliot:
Failure not related to PR -> https://github.com/xamarin/maccore/issues/2443

PR #12241, opened 2021-07-27
[monotouch-test] No need to ignore NSTimeZoneTest.All_28300 anymore.

rolfbjarne:


spouliot (approved):
still needed ?

rolfbjarne:
> still needed ?

No, but now the Ignore can be completely removed, so I changed this PR to do that instead.

rolfbjarne:

PR #12242, opened 2021-07-27
[dotnet] Make it clearer that the linker is off by default for Mac Catalyst (like it already is for macOS).

rolfbjarne:
There's no real change here, the code did exactly this in before, but it was a
side effect of '_SdkIsSimulator' being 'true' for Mac Catalyst, which isn't
obvious. Now the logic is explicit for all platforms.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -872,9 +872,9 @@
 
 	<Target Name="_ComputeDefaultLinkMode" DependsOnTargets="_DetectSdkLocations">
 		<PropertyGroup>
-			<_DefaultLinkMode Condition="'$(_PlatformName)' == 'macOS'">None</_DefaultLinkMode> <!-- Linking is off by default for macOS apps -->
-			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_SdkIsSimulator)' == 'true'">None</_DefaultLinkMode> <!-- Linking is off by default in the simulator -->
-			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_SdkIsSimulator)' != 'true'">SdkOnly</_DefaultLinkMode> <!-- Linking is SdkOnly for iOS/tvOS/watchOS apps on device -->
+			<_DefaultLinkMode Condition="'$(_PlatformName)' == 'macOS' Or '$(_PlatformName)' == 'MacCatalyst'">None</_DefaultLinkMode> <!-- Linking is off by default for macOS and Mac Catalyst apps -->
+			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_PlatformName)' == 'MacCatalyst' And '$(_SdkIsSimulator)' == 'true'">None</_DefaultLinkMode> <!-- Linking is off by default in the simulator -->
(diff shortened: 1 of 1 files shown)


dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -872,9 +872,9 @@
...
 		<PropertyGroup>
-			<_DefaultLinkMode Condition="'$(_PlatformName)' == 'macOS'">None</_DefaultLinkMode> <!-- Linking is off by default for macOS apps -->
-			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_SdkIsSimulator)' == 'true'">None</_DefaultLinkMode> <!-- Linking is off by default in the simulator -->
-			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_SdkIsSimulator)' != 'true'">SdkOnly</_DefaultLinkMode> <!-- Linking is SdkOnly for iOS/tvOS/watchOS apps on device -->
+			<_DefaultLinkMode Condition="'$(_PlatformName)' == 'macOS' Or '$(_PlatformName)' == 'MacCatalyst'">None</_DefaultLinkMode> <!-- Linking is off by default for macOS and Mac Catalyst apps -->
+			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_PlatformName)' == 'MacCatalyst' And '$(_SdkIsSimulator)' == 'true'">None</_DefaultLinkMode> <!-- Linking is off by default in the simulator -->
+			<_DefaultLinkMode Condition="'$(_PlatformName)' != 'macOS' And '$(_PlatformName)' == 'MacCatalyst' Abd '$(_SdkIsSimulator)' != 'true'">SdkOnly</_DefaultLinkMode> <!-- Linking is SdkOnly for iOS/tvOS/watchOS apps on device -->
```
spouliot, line 877: `Adb` won't work :)

spouliot: should not it be `'$(_PlatformName)' != 'MacCatalyst'`
since it's for everything but Mac* ?

spouliot (changes requested):

spouliot:
I'm not sure if that not-linking default is still the right choice considering that AOT is required (at least on ARM targets) and the build will become very long...

App size will also be quite larger, but that's likely not an issue (and it can be changed to linking if/when it becomes one)

rolfbjarne:
> I'm not sure if that not-linking default is still the right choice considering that AOT is required (at least on ARM targets) and the build will become very long...

The interpreter still works, and that's the default (it still requires AOT-compiling System.Private.CoreLib.dll, but that's it).

spouliot:
> > I'm not sure if that not-linking default is still the right choice considering that AOT is required (at least on ARM targets) and the build will become very long...
> 
> The interpreter still works, and that's the default (it still requires AOT-compiling System.Private.CoreLib.dll, but that's it).

That (interpreter) make sense for debug builds but I was thinking about release builds

spouliot (approved):

rolfbjarne:

rolfbjarne:
> > > I'm not sure if that not-linking default is still the right choice considering that AOT is required (at least on ARM targets) and the build will become very long...
> > 
> > 
> > The interpreter still works, and that's the default (it still requires AOT-compiling System.Private.CoreLib.dll, but that's it).
> 
> That (interpreter) make sense for debug builds but I was thinking about release builds

Yes, I'm not sure what's the best default behavior here. I filed #12264 so that we can make a decision.

PR #12250, opened 2021-07-27
[CI] Do not even try to do net6 publishing on not main builds.

mandel-macaque:
Lets not create steps using templates that will not be ran anyway. This removes those 2 extra steps when we are building on a PR.

Example of a internal build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5025066&view=results
External build (this PR :) )

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -243,8 +243,9 @@ stages:
         xqaCertPass: $(xqa--certificates--password)
         enableDotnet: ${{ parameters.enableDotnet }}
 
-# .NET 6 Release Prep and VS Insertion Stages
-- template: templates/release/vs-insertion-prep.yml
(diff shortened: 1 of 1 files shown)


rolfbjarne (approved):

tools/devops/automation/build-pipeline.yml
```diff
@@ -243,8 +243,9 @@ stages:
         xqaCertPass: $(xqa--certificates--password)
         enableDotnet: ${{ parameters.enableDotnet }}
 
-# .NET 6 Release Prep and VS Insertion Stages
-- template: templates/release/vs-insertion-prep.yml
+# .NET 6 Release Prep and VS Insertion Stages, only execute them when the build reason is not a PR or a schedule build fro OneLoc
```
mandel-macaque, line 246:
```suggestion
# .NET 6 Release Prep and VS Insertion Stages, only execute them when the build reason is not a PR or a schedule build from OneLoc
```

spouliot (approved):

PR #12273, opened 2021-07-28
[Xcode13] Bump for Xcode 13 beta4.

mandel-macaque:

PR #12274, opened 2021-07-28
[xtro] Append the lines and then sort them to get everything grouped better.

mandel-macaque:
@spouliot  @rolfbjarne I don't know how I feel abut this. On one hand, is easy to just do sort -o filename filename. On the other hand I could have done everything in c#.


spouliot (approved):
👍 
it's not a time-sensitive operation, nor is it part of the build

dalexsoto (approved):

spouliot:
https://github.com/xamarin/xamarin-macios/pull/12273#discussion_r678508129

seems this might end up creating a noisy battle between sorts :(

PR #12305, opened 2021-07-30
[src] Merge two implementations of converting a FourCC to a string.

rolfbjarne:


mandel-macaque (approved):

dalexsoto (approved):

src/Foundation/NSScriptCommandDescription.cs
```diff
@@ -26,15 +26,6 @@ public partial class NSScriptCommandDescription {
...
-		static string ToFourCCString (int value)
-		{
-			return new string (new char [] {
-				(char) (byte) (value >> 24),
-				(char) (byte) (value >> 16),
-				(char) (byte) (value >> 8),
-				(char) (byte) value });
-		}
-		
 		static int ToIntValue (string fourCC)
```
dalexsoto, line 29: Should we do the same for the reverse one?

PR #12309, opened 2021-07-30
[runtime] Implement computing and passing the NATIVE_DLL_SEARCH_DIRECTORIES runtime property. Fixes #10504.

### LINKED ISSUE #10504: Xamarin iOS should call monovm_initialize for .NET 6

rolfbjarne:
This adds support to compute the NATIVE_DLL_SEARCH_DIRECTORIES value and pass
it to the runtime. It's the last property listed in #10504, so this fixes that
issue.

Fixes https://github.com/xamarin/xamarin-macios/issues/10504

### DIFF
--- a/runtime/runtime.m
+++ b/runtime/runtime.m
@@ -2450,6 +2450,39 @@ xamarin_compute_trusted_platform_assemblies ()
 	return rv;
 }
 
+char *
+xamarin_compute_native_dll_search_directories ()
+{
+	const char *bundle_path = xamarin_get_bundle_path ();
+
+	NSMutableArray<NSString *> *directories = [NSMutableArray array];
+
+	// Native libraries might be in the app bundle
+	[directories addObject: [NSString stringWithUTF8String: bundle_path]];
+	// They won't be in the runtimeidentifier-specific directory (because they get lipo'ed into a fat file instead)
+	// However, might also be in the Resources/lib directory
+	[directories addObject: [[[NSBundle mainBundle] resourcePath] stringByAppendingPathComponent: @"lib"]];
+
+	// Missing:
+	// * The parent app bundle if launched from an app extension. This requires adding app extension tests, which we currently don't have.
+
+	// Remove the ones that don't exist
+	NSFileManager *manager = [NSFileManager defaultManager];
+	for (int i = (int) [directories count] - 1; i >= 0; i--) {
+		NSString *dir = [directories objectAtIndex: (NSUInteger) i];
+		BOOL isDirectory;
+		if ([manager fileExistsAtPath: dir isDirectory: &isDirectory] && isDirectory)
+			continue;
+
+		[directories removeObjectAtIndex: (NSUInteger) i];
+	}
+
+	// Join them all together with a colon separating them
+	NSString *joined = [directories componentsJoinedByString: @":"];
+	char *rv = xamarin_strdup_printf ("%s", [joined UTF8String]);
+	return rv;
+}
+
 void
 xamarin_vm_initialize ()
 {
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):

runtime/runtime.m
```diff
@@ -2450,6 +2450,39 @@ -(void) xamarinSetFlags: (enum XamarinGCHandleFlags) flags;
...
+		BOOL isDirectory;
+		if ([manager fileExistsAtPath: dir isDirectory: &isDirectory] && isDirectory)
+			continue;
+
+		[directories removeObjectAtIndex: (NSUInteger) i];
+	}
+
+	// Join them all together with a colon separating them
+	NSString *joined = [directories componentsJoinedByString: @":"];
+	char *rv = xamarin_strdup_printf ("%s", [joined UTF8String]);
```
Therzok, line 2482: Maybe use this instead? https://developer.apple.com/documentation/foundation/nsstring/1408489-cstringusingencoding?language=objc

rolfbjarne: The return value from that function is only valid for as long as the NSString is alive: "The returned C string is guaranteed to be valid only until either the receiver is freed". I could use [getCString:maxLength:encoding:](https://developer.apple.com/documentation/foundation/nsstring/1415702-getcstring?language=objc) instead, but that ends up making the code somewhat longer and more complex, and this is a one-time thing at startup, so I don't think it matters all that much.

rolfbjarne:

PR #12344, opened 2021-08-04
[runtime] Add support for logging exceptions that go through exception marshalling using an environment variable. Fixes #12343.

### LINKED ISSUE #12343: Make it possible to log marshaled exceptions from the command line using an environment variable.
Sometimes exceptions (either managed or native) are thrown before any user code has executed, and in that case our [exception events](https://docs.microsoft.com/en-us/xamarin/ios/platform/exception-marshaling#events) are not useful, because user code won't have had a chance to attach handlers to those events.

The idea would be to set an environment variable that would cause information about the exceptions to be printed to the system log.

rolfbjarne:
Setting the XAMARIN_LOG_MARSHALLED_EXCEPTIONS environment variable will now
make us print all exceptions that go through exception marshalling, and how
we'll handle them.

Fixes https://github.com/xamarin/xamarin-macios/issues/12343.


spouliot (approved):

dalexsoto (approved):

rolfbjarne:

PR #12352, opened 2021-08-05
[NetworkExtension] Add Xcode 13 beta 4 support.

mandel-macaque:


rolfbjarne (approved):

tests/xtro-sharpie/MacCatalyst-NetworkExtension.ignore
```diff
@@ -0,0 +1 @@
+!duplicate-type-name! NWPathStatus enum exists as both NetworkExtension.NWPathStatus and Network.NWPathStatus
```
spouliot, line 1: weird, that type has not changed in the PR

spouliot (approved):

tests/xtro-sharpie/MacCatalyst-NetworkExtension.ignore
mandel-macaque: Was not present in catalyst, was in macOS. And the ignore of that type is in the macOS ignore, that's why we had to add it for catalyst

spouliot: yeah - but you did not change `src/frameworks.sources` (or any other files) in that PR, so
if it was present before, then it should have been an error in xtro
if it was not, then it was not added, and should not be reported

mandel-macaque:
Unrelated error: https://github.com/xamarin/maccore/issues/2414

PR #12334, opened 2021-08-03
[dotnet] Remove workaround for private symbols for AOT.

rolfbjarne:

PR #12367, opened 2021-08-05
[net6] Fixes LinkNativeCode task execution from Windows

emaf:
This is an improvement on how we detect if the file that is going to be copied to the Mac is coming from the output path or not. Copying files from the Windows output dir to the Mac will make the build fail.

There were two problems:
- The paths we were comparing could contain different path separators.
- The item we are trying to copy may be in a folder that is inside the output dir.

For example:
- OutputFile = `obj\Debug\net6.0-ios\iossimulator-x64\nativelibraries/Foo`
- Item to copy: `obj/Debug/net6.0-ios/iossimulator-x64/nativelibraries/aot-output/x86_64/System.Private.CoreLib.dll.o`


rolfbjarne (approved):

rolfbjarne:
Test failures are unrelated

* monotouch-test/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2443
* DotNet tests: fixed in #12334.

PR #12368, opened 2021-08-06
[generator] Use new `CFArray.StringArrayFromHandle` instead of the existing `NSArray` API

spouliot:
This is another example that p/invokes are much faster than calling
selectors.

Beside the generator the manual bindings were updated to use the newer
API and the old one was decorated as `[Obsolete]`.

|                        Method |          name |                value |            Mean |           Error |        StdDev |
|------------------------------ |-------------- |--------------------- |----------------:|----------------:|--------------:|
| CFArray_StringArrayFromHandle |         empty |                   () |        123.9 ns |        68.92 ns |       3.78 ns |
| NSArray_StringArrayFromHandle |         empty |                   () |      1,422.6 ns |        25.83 ns |       1.42 ns |
| CFArray_StringArrayFromHandle |           few |    (  (...).") [108] |      1,885.2 ns |        46.37 ns |       2.54 ns |
| NSArray_StringArrayFromHandle |           few |    (  (...).") [108] |      8,530.0 ns |       594.40 ns |      32.58 ns |
| CFArray_StringArrayFromHandle | large_mutable |    ((...)) [8419330] | 15,821,101.0 ns | 4,803,631.19 ns | 263,303.23 ns |
| NSArray_StringArrayFromHandle | large_mutable |    ((...)) [8419330] | 22,823,871.9 ns | 6,589,380.43 ns | 361,186.18 ns |
| CFArray_StringArrayFromHandle |       mutable |   (   (...)9e") [54] |        867.4 ns |        59.23 ns |       3.25 ns |
| NSArray_StringArrayFromHandle |       mutable |   (   (...)9e") [54] |      4,939.6 ns |       203.28 ns |      11.14 ns |

note: `NSArray.StringArrayFromHandle` was already using (p/invoke-based)
`CFString.FromHandle` instead of (selector-based) `NSString.FromHandle`
to create the managed `string` instances inside the `string[]`.

### DIFF
--- a/src/CoreFoundation/CFArray.cs
+++ b/src/CoreFoundation/CFArray.cs
@@ -128,6 +128,20 @@ namespace CoreFoundation {
 		extern static CFArrayRef CFArrayCreateCopy (CFAllocatorRef allocator, CFArrayRef theArray);
 
 		internal CFArray Clone () => new CFArray (CFArrayCreateCopy (IntPtr.Zero, GetCheckedHandle ()), true);
+
+		// identical signature to NSArray API
+		static public string?[]? StringArrayFromHandle (IntPtr handle)
+		{
+			if (handle == IntPtr.Zero)
+				return null;
+
+			var c = CFArrayGetCount (handle);
+			string?[] ret = new string [c];
+
+			for (nint i = 0; i < c; i++)
+				ret [i] = CFString.FromHandle (CFArrayGetValueAtIndex (handle, i));
+			return ret;
+		}
 	}
 }
 
--- a/src/AssetsLibrary/ALAsset.cs
+++ b/src/AssetsLibrary/ALAsset.cs
@@ -10,6 +10,7 @@
 using ObjCRuntime;
 using Foundation;
 using CoreGraphics;
+using CoreFoundation;
 using CoreLocation;
 using UIKit;
 using MediaPlayer;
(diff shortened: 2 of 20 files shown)


dalexsoto (approved):

src/CoreFoundation/CFArray.cs
```diff
@@ -128,6 +128,20 @@ internal static nint GetCount (IntPtr array)
...
+		static public string?[]? StringArrayFromHandle (IntPtr handle)
+		{
+			if (handle == IntPtr.Zero)
+				return null;
+
+			var c = CFArrayGetCount (handle);
+			string?[] ret = new string [c];
+
+			for (nint i = 0; i < c; i++)
+				ret [i] = CFString.FromHandle (CFArrayGetValueAtIndex (handle, i));
```
rolfbjarne, line 142: I wonder if it would be faster to call `CFArrayGetValues` to get all the values at the same time (or chunks of them at least)?

rolfbjarne (approved):

src/CoreFoundation/CFArray.cs
spouliot: Good catch! Most probably, IOW when the element count is large enough to offset the temporary buffer allocation.

I'll try it (on another night) for `ArrayFromHandle` and update `StringArrayFromHandle` if that's the case.

spouliot:
Failure is unrelated to PR -> https://github.com/xamarin/maccore/issues/2451

PR #12383, opened 2021-08-09
[monotouch-test] Ignore ReadingListTest.DefaultReadingList.

rolfbjarne:
This test adds entries to the global reading list in Safari, and there's no
API for the test to clean up after itself, so the reading list becomes an
endless list of entries after a whlie.

Beside the fact that this is somewhat annoying for people who actually use
their reading lists, it can also can end up consuming a significant amount of
space on the hard drive, because Safari will download each site to be
available offline.


dalexsoto (approved):

rolfbjarne:

PR #12396, opened 2021-08-10
[dotnet/msbuild] Fix publishing user frameworks and dynamic libraries from binding projects.

rolfbjarne:
We extract frameworks from third-party libraries when running the linker, and
we need to store this information on disk and the reload it after the linker
has executed, and add it to the existing MSBuild variables that keep track of
the user frameworks and dynamic libraries that have to be copied to the app
bundle.

Fixes the framework-test tests.


spouliot (approved):

emaf (approved):

rolfbjarne:
Test run is green, except that it failed to report back to GitHub.

[index.html.txt](https://github.com/xamarin/xamarin-macios/files/6966756/index.html.txt)

PR #12401, opened 2021-08-10
[src] Ensure complete (error) coverage for nullability warnings

spouliot:
Use the _recent_ [1] `nullable` to ensure we treat all nullable warnings
as errors. That list can (and has) change[d] over time...

Fix one case, `CVTime`, that was (only) a warning in our build.

Also fix `Makefile` so we are not giving the `warnaserror` for nullable
to the generator. It does not use them (different codes) and does not
accept string values (which breaks the build). IOW it's the generated
code that must be nullable-aware, not the input to the generator.

[1] https://github.com/dotnet/roslyn/issues/41605


dalexsoto (approved):

mandel-macaque (approved):

stephen-hawley (approved):

spouliot:
Failure unrelated to PR -> https://github.com/xamarin/maccore/issues/1605

PR #12405, opened 2021-08-10
[MailKit] Introducing in Xcode13 Beta 5

tj-devel709:
I have a few comments/questions listed down below that I would like help with!


src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Export ("mailComposeSessionDidEnd:")]
+		void MailComposeSessionDidEnd (MEComposeSession session);
+
+		[Abstract]
+		[Export ("viewControllerForSession:")]
+		MEExtensionViewController GetViewController (MEComposeSession session);
+
+		[Async]
+		[Export ("session:annotateAddressesWithCompletionHandler:")]
+		void AnnotateAddress (MEComposeSession session, SessionHandler completionHandler);
```
tj-devel709, line 329: Not sure if this delegate really helps since I am only naming one parameter..

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Advice ("'MEAddressAnnotation' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface MEAddressAnnotation : NSSecureCoding
+	{
+		[Static]
+		[Export ("errorWithLocalizedDescription:")]
+		MEAddressAnnotation AnnotateError (string localizedDescription);
```
tj-devel709, line 232: These were originally "Error", "Warning", and "Success" but I tried to start them with verbs. Is there a better choice for this?

src/frameworks.sources
```diff
@@ -2256,6 +2257,7 @@ MACCATALYST_FRAMEWORKS = \
 	JavaScriptCore \
 	LinkPresentation \
 	LocalAuthentication \
+	MailKit \
```
spouliot, line 2260: why ? it seems types are all decorated with `[NoMacCatalyst]` ?

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
+using System;
+using CoreGraphics;
+using Foundation;
+using ObjCRuntime;
+using AVFoundation;
+using AppKit;
+
+namespace MailKit {
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Native, Advice ("'MEComposeSessionErrorCode' is not available in UIKit on macOS.")]
```
spouliot, line 10: You can removed all `[Advice ("'MEComposeSessionErrorCode' is not available in UIKit on macOS.")]`
That was needed before we had `[NoMacCatalyst]` attributes

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Abstract]
+		[Export ("extensionViewControllerForMessageSigners:")]
+		[return: NullAllowed]
+		MEExtensionViewController ExtensionViewController (MEMessageSigner[] messageSigners);
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[DisableDefaultCtor]
+	[BaseType (typeof (NSViewController))]
+	interface MEExtensionViewController
```
tj-devel709, line 420: Introspection on Mac gives this error:
```
14:36:15.5618610 	[FAIL] DesignatedInitializer :   1 potential errors found in 751 designated initializer validated
14:36:15.5619810   Expected: 0
14:36:15.5619930   But was:  1
14:36:15.5620130 
14:36:15.5620240 [FAIL] MailKit.MEExtensionViewController should re-expose NSViewController::.ctor(System.String, Foundation.NSBundle)
14:36:15.5620320 
14:36:15.5624080 		  at Introspection.ApiCtorInitTest.DesignatedInitializer () [0x0014b] in /Users/tjlambert/Documents/Xcode13/xamarin-macios/tests/introspection/ApiCtorInitTest.cs:335 

```

This seems like something that should be ignored in the test?

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Advice ("'MEComposeSession' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface MEComposeSession : NSSecureCoding
+	{
+		[Export ("sessionID", ArgumentSemantic.Strong)]
+		NSUuid SessionID { get; }
```
spouliot, line 250: `Id` since it's not an acronym

src/mailkit.cs
spouliot: not sure, but the name of the custom delegate should be information
`SessionHandler` is true, but not really informative of that's expected
maybe `MEAnnotationHandler` ?

but then `<Action<NSDictionary<MEEmailAddress, MEAddressAnnotation>> completionHandler` seems rather obvious...

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[NullAllowed, Export ("addressString")]
+		string AddressString { get; }
+	}
+
+	delegate void SessionHandler (NSDictionary<MEEmailAddress, MEAddressAnnotation> sessionDictionary);
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEComposeSessionHandler' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	interface MEComposeSessionHandler
```
spouliot, line 313: remove the `[BaseType]` since it's a `[Protocol]` (and there's no `[Model]` on it since it's not an ObjC delegate)

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+		[Abstract]
+		[Export ("mailComposeSessionDidEnd:")]
+		void MailComposeSessionDidEnd (MEComposeSession session);
+
+		[Abstract]
+		[Export ("viewControllerForSession:")]
+		MEExtensionViewController GetViewController (MEComposeSession session);
+
+		[Async]
```
spouliot, line 327: can't put `Async` on protocols
conforming to this member just would not work at runtime
and since it's being called by the OS then it would not be helpful

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Export ("session:annotateAddressesWithCompletionHandler:")]
+		void AnnotateAddress (MEComposeSession session, SessionHandler completionHandler);
+
+		[Async]
```
spouliot, line 331: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Export ("session:canSendMessageWithCompletionHandler:")]
+		void AllowMessageSend (MEComposeSession session, Action<NSError> completionHandler);
+
+		[Export ("additionalHeadersForSession:")]
+		NSDictionary<NSString, NSArray<NSString>> GetAdditionalHeaders (MEComposeSession session);
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEContentBlocker' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 341: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	interface MEContentBlocker
+	{
+		[Abstract]
+		[Export ("contentRulesJSON")]
+		NSData ContentRulesJson { get; }
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEExtension' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 351: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	interface MEExtension
+	{
+		[Export ("handlerForComposeSession:")]
+		MEComposeSessionHandler HandlerForComposeSession (MEComposeSession session);
```
spouliot, line 355: prefix with a verb/action, e.g. `GetHandlerForComposeSession`

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+		[Export ("handlerForMessageActions")]
+		MEMessageActionHandler HandlerForMessageActions { get; }
```
spouliot, line 358: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+		[Export ("handlerForContentBlocker")]
+		MEContentBlocker HandlerForContentBlocker { get; }
```
spouliot, line 361: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+		[Export ("handlerForMessageSecurity")]
+		IMEMessageSecurityHandler HandlerForMessageSecurity { get; }
```
spouliot, line 364: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEMessageActionHandler' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 369: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	interface MEMessageActionHandler
+	{
+		[Abstract]
+		[Export ("decideActionForMessage:completionHandler:")]
+		void CompletionHandler (MEMessage message, Action<MEMessageActionDecision> completionHandler);
```
spouliot, line 374: wrong name, it's the `decideAction` that's the important part of the selector name
`ForMessage:completionHandler:` are the parameters

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+
+		[Export ("requiredHeaders", ArgumentSemantic.Copy)]
+		string[] RequiredHeaders { get; }
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEMessageDecoder' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 382: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	{
+		[Abstract]
+		[Export ("decodedMessageForMessageData:")]
+		[return: NullAllowed]
+		MEDecodedMessage DecodedMessage (NSData data);
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEMessageEncoder' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
```
spouliot, line 393: same

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+	interface IMEMessageSecurityHandler {}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Protocol, Advice ("'MEMessageSecurityHandler' is not available in UIKit on macOS.")]
+	interface MEMessageSecurityHandler : MEMessageEncoder, MEMessageDecoder
+	{
+		[Abstract]
+		[Export ("extensionViewControllerForMessageSigners:")]
+		[return: NullAllowed]
+		MEExtensionViewController ExtensionViewController (MEMessageSigner[] messageSigners);
```
spouliot, line 414: prefix with verb/action, e.g. `Get*`

spouliot (changes requested):

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Export ("dateSent", ArgumentSemantic.Copy)]
+		[NullAllowed]
+		NSDate DateSent { get; }
+
+		[Export ("dateReceived", ArgumentSemantic.Copy)]
+		[NullAllowed]
+		NSDate DateReceived { get; }
+
+		[NullAllowed, Export ("headers", ArgumentSemantic.Copy)]
+		NSDictionary<NSString, NSArray<NSString>> Headers { get; }
```
mandel-macaque, line 81: Can this be a smart dictionary ?

src/mailkit.cs
```diff
@@ -0,0 +1,423 @@
...
+		[Static]
+		[Export ("setColorActionWithColor:")]
+		MEMessageAction SetColor (MEMessageActionMessageColor color);
+	}
+
+	[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
+	[Advice ("'MEMessageActionDecision' is not available in UIKit on macOS.")]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface MEMessageActionDecision : NSSecureCoding
```
mandel-macaque, line 130: Static class? All methods/properties are static and we have no contractor.

src/mailkit.cs
mandel-macaque: Why annotate? I'm not native, but annotate to me is:
```
add notes to (a text or diagram) giving explanation or comment.
```
What about GetLocalizedError () etc

mandel-macaque (changes requested):

tools/common/Frameworks.cs
```diff
@@ -264,6 +264,7 @@ public Framework Find (string framework)
 					{ "Chip", "CHIP", 12, 0 },
 					{ "MetricKit", 12, 0 },
 					{ "Phase", "PHASE", 12, 0 },
+					{ "MailKit", "MailKit", 12, 0 },
```
mandel-macaque, line 267: Alphabetical order please.

src/mailkit.cs
rolfbjarne: From the documentation about [MEExtensionViewController](https://developer.apple.com/documentation/mailkit/meextensionviewcontroller?language=objc), app developers are supposed to subclass this class to provide their own implementation. This means that there must be constructors available in this class for subclasses to call. In Objective-C constructors are automatically available to all subclasses (not only direct ones), which means that the constructors defined on `NSViewController` are available for this class as well (that's what the failing test is talking about).

In the end, it seems like we should add two constructors: the default one, and the one from the error message, and make both protected (which effectively makes this class `abstract`, because you can only instantiate it if you subclass it - which is what the documentation implies should be the case):

```csharp
[NoWatch, NoTV, NoiOS, NoMacCatalyst, Mac (12,0)]
[DefaultCtorVisibility (Visibility.Protected)]
[BaseType (typeof (NSViewController))]
interface MEExtensionViewController
{
		[DesignatedInitializer]
		[Protected]
		[Export ("initWithNibName:bundle:")]
		IntPtr Constructor ([NullAllowed] string nibNameOrNull, [NullAllowed] NSBundle nibBundleOrNull);
}
```

rolfbjarne (changes requested):

src/mailkit.cs
tj-devel709: I'll just leave the action in!

src/mailkit.cs
tj-devel709: Thank you! I do (now) recall Rachel asking about this and I will try to remember this for next time!

src/frameworks.sources
tj-devel709: I was a little confused about this, but this makes sense not to add it, thanks!

src/mailkit.cs
tj-devel709: Do I need the verb/action prefix for getters as well?

src/mailkit.cs
tj-devel709: Okay I think these are better names! I was getting confused looking at the web docs, but this makes sense!

src/mailkit.cs
tj-devel709: Sorry, what do you mean by smart dictionary?

src/mailkit.cs
dalexsoto: This can't be a static class heh

mandel-macaque: And it should not, i'm wrong we get it from https://developer.apple.com/documentation/mailkit/memessageactionhandler/3783568-decideactionformessage?changes=_3&language=objc

So we need to be able to instantiate the class.

src/mailkit.cs
mandel-macaque: Sorry, strong dictionary: https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#strongly-typed-dictionaries

mandel-macaque: It might be the case it can't since headers can be anything in the mail protocol, so lets ignore it for now

src/mailkit.cs
tj-devel709: When I made the interface static, I got these errors in the build: https://gist.github.com/tj-devel709/a4500a4835c99831d0d2f7883f92fe5d

tj-devel709: Oops, sorry didn't refresh before I posted that ^

spouliot (approved):

rolfbjarne (approved):

mandel-macaque (approved):

dalexsoto (approved):

tj-devel709:
Failing tests are unrelated networking issues!

PR #12407, opened 2021-08-11
[src] Use [NoMacCatalyst] everywhere.

rolfbjarne:
Instead of this:

    [Unavailable (PlatformName.MacCatalyst)]

do this:

    [NoMacCatalyst]

Except when the API is actually available (in which case add the correct availability attribute), or in manual binding code (where we still have to use `[Unavailable (...)]`, although not the `[Advice (...)]`).

### DIFF
--- a/src/AudioUnit/AudioComponentDescription.cs
+++ b/src/AudioUnit/AudioComponentDescription.cs
@@ -138,7 +138,6 @@ namespace AudioUnit
 #endif
 		[Mac (10,15)]
 		[Unavailable (PlatformName.MacCatalyst)]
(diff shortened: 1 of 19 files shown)


spouliot (approved):

mandel-macaque (approved):

dalexsoto (approved):

mandel-macaque (approved):

tj-devel709 (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2492).

PR #12421, opened 2021-08-12
[dotnet] Fix exclusion of mono components.

rolfbjarne:
'_MonoRuntimePackPath' already contains a trailing slash, so adding another
one made two, and that was too much for MSBuild to realize that it was really
the same file as the one with only one slash in the path, and those files
weren't removed from the build correctly.

This saves ~300kb from the app size comparison:

Before: https://gist.github.com/rolfbjarne/812f3ded1cf1a8bea45c7ca9b42bf95b
After: https://gist.github.com/rolfbjarne/22b23db0a162e9b9e366392b67063907


spouliot (approved):

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

* introspection/Mac Catalyst/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414
* monotouch-test/tvOS - simulator/Debug (LinkSdk) [dotnet]: https://github.com/xamarin/maccore/issues/1605

PR #12424, opened 2021-08-12
[msbuild] Unify the iOS and macOS versions of the IBTool task.

rolfbjarne:
* Have a single implementation of AutoActivateCustomFonts.
* Share the GetTargetDevices implementations between ACTool and IBTool, after removing
  a condition for Xcode 6.0 (which we don't support anymore, so that check could
  be removed) the implementations were identical.


dalexsoto (approved):

spouliot (approved):

stephen-hawley (approved):

emaf (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2304).

PR #12433, opened 2021-08-13
[msbuild] Add a property that specifies the final path to the Info.plist in the app bundle

rolfbjarne:
* Add an '_AppBundleManifest' property that specifies the final path to the
  Info.plist in the app bundle.
* Remove the '_AppBundleManifestPath' property, it's not used anywhere.
* Adjust the CompileAppManifest task to take the final path to the Info.plist,
  instead of computing it and returning it. This way the CompileAppManifest
  task does not output anything back into MSBuild (which is important, because
  the CompileAppManifest task won't execute if the output file is up-to-date,
  and if it's not executed, any output properties won't be set either).

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
@@ -35,6 +35,7 @@ namespace Xamarin.MacDev.Tasks
 		public string BundleIdentifier { get; set; }
 
 		[Required]
+		[Output] // This is required to create an empty file on Windows for the Input/Outputs check.
 		public string CompiledAppManifest { get; set; }
 
 		[Required]
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
@@ -29,13 +29,13 @@ namespace Xamarin.MacDev.Tasks
 		public string AppManifest { get; set; }
 
 		[Required]
-		public string AppManifestBundleDirectory { get; set; }
-
-		[Required]
 		public string AssemblyName { get; set; }
 
 		[Required]
 		public string BundleIdentifier { get; set; }
+
+		[Required]
+		public string CompiledAppManifest { get; set; }
 
 		[Required]
 		public bool Debug { get; set; }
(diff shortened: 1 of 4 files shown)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
```diff
@@ -73,13 +73,6 @@ public abstract class CompileAppManifestTaskBase : XamarinTask
 		public string TargetArchitectures { get; set; }
 		#endregion
 
-		#region Outputs
-
-		[Output]
-		public ITaskItem CompiledAppManifest { get; set; }
```
emaf, line 79: This output property was creating an empty Info.plist file on Windows for the Inputs/Outpus MSBuild check. If we don't want to have output props on this task we can run the `Touch` task after `CompileAppManifest`.

In my opinion having an ITaskItem output prop is still the best option, that way we ensure every time this task is run the file will be created on Windows.

emaf (changes requested):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
rolfbjarne: @emaf I made it an [Output] property:

https://github.com/xamarin/xamarin-macios/blob/d02e405963d3ec367fba2a583a04537648dd7b1d/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs#L37-L39

would that work?

emaf: Yes, that should work, but the type should be `ITaskItem` :)

dalexsoto (approved):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
rolfbjarne: @emaf ok, changed!

spouliot (approved):

rolfbjarne:
Test failures are unrelated:

* framework-test/Mac Catalyst/Debug: https://github.com/xamarin/maccore/issues/2414
* monotouch-test/tvOS - simulator/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2304

emaf (approved):

PR #12451, opened 2021-08-16
[HomeKit] Updates for Xcode13 Beta 5

tj-devel709:
Not 100% sure if this is correct. 
The headers showed that they are now deprecated, but Sharpie and the web docs do not reflect this. 
I will test in Xcode to be sure!

### DIFF
--- a/src/homekit.cs
+++ b/src/homekit.cs
@@ -85,6 +85,7 @@ namespace HomeKit {
 
 		[iOS (13,0), NoWatch, NoTV, NoMac]
 		[NoMacCatalyst]
+		[Deprecated (PlatformName.iOS, 15, 0, message: "This method is no longer supported.")]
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):
Looking at the headers, it is correct.

spouliot (approved):
Not sure why sharpie did not pick it up. Likely because the macro is different from others (each framework like to have it's own or re-define them) and did not end up inside the `.pch` file.

In any case the attributes will not cause a breaking change if we have to update them later, so having them in sync with Xcode headers is fine :-)

mandel-macaque:
The failing tests is a network one, we have been having several of those in the CI in recent days:
```
TestNSUrlSessionHandlerCookies : Failed to get native cookies 
```

PR #12454, opened 2021-08-16
[Carplay] Add support for Xcode 13 beta 5 and fix catalyst attrs.

mandel-macaque:


mandel-macaque:
Failure is not related: https://github.com/xamarin/maccore/issues/2493

rolfbjarne (approved):

src/carplay.cs
```diff
@@ -1907,4 +1925,22 @@ interface CPInformationRatingItem
 		[NullAllowed, Export ("maximumRating")]
 		NSNumber MaximumRating { get; }
 	}
+
+	[NoWatch, NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
+	interface CPAssistantCellConfiguration : NSSecureCoding
```
spouliot, line 1931: -> `[DisableDefaultCtor]`
there's no way to change the default, empty/null, value if `init` is used - so it's unusable

spouliot (changes requested):

dalexsoto (approved):

src/carplay.cs
```diff
@@ -1907,4 +1925,22 @@ interface CPInformationRatingItem
 		[NullAllowed, Export ("maximumRating")]
 		NSNumber MaximumRating { get; }
 	}
+
+	[NoWatch, NoTV, NoMac, iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
```
mandel-macaque, line 1930:
```suggestion
	[BaseType (typeof(NSObject))]
	[DisableDefaultCtor]
```

spouliot (approved):

PR #12461, opened 2021-08-17
[msbuild] Refactor the logic to compute the target devices for IBTool/ACTool.

rolfbjarne:
Refactor the logic to compute the target devices for IBTool/ACTool, so that it
doesn't depend on the existence of the input app manifest, but instead uses
values that either comes from the app manifest, or have a default.

This becomes an important distinction when we add support for multiple partial
app manifests, or none at all (where all the required values come from MSBuild
properties).


spouliot (approved):

dalexsoto (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated (will be fixed with #12470).

emaf (approved):

PR #12463, opened 2021-08-17
[CoreAudio] Updates for Xcode13 Beta 3

tj-devel709:


tj-devel709:
Sorry, I don't think this one is actually ready for review. I think there are some things in the headers that were not mentioned by xtro that need updating!

PR #12465, opened 2021-08-17
[CoreAudio] Updates for Xcode13 Beta 5

tj-devel709:
There were a few changes in the headers that I believe are not bound by us:
- kAudioAggregateDevicePropertyMasterSubDevice -> kAudioAggregateDevicePropertyMainSubDevice
- mMasterHostTicksPerFrame -> mMainHostTicksPerFrame | inside AudioServerPlugInIOCycleInfo (confirmed in office hours)
- kAudioHardwareNotReadyError
- kAudioHardwarePropertyProcessIsMaster -> kAudioHardwarePropertyProcessIsMain
    - although it seems as though the AudioObjectPropertySelector holds the values for kAudioHardwareProperty as well?


src/AudioUnit/AUEnums.cs
```diff
@@ -187,7 +187,9 @@ public enum AudioObjectPropertySelector : uint
 		TranslateUIDToBox = 1969841250, // 'uidb'
 		ClockDeviceList = 1668049699, //'clk#'
 		TranslateUidToClockDevice = 1969841251, // 'uidc',
+		[Obsolete ("Use the 'ProcessIsMain' element instead.")]
```
mandel-macaque, line 190: You are missing the deprecated attrs:

```
[Deprecated (PlatformName.TvOS, 15,0)]
```
you will have to add them for each of the supported platforms.

src/AudioUnit/AUEnums.cs
```diff
@@ -187,7 +187,9 @@ public enum AudioObjectPropertySelector : uint
...
 		ProcessIsMaster = 1835103092, // 'mast'
+		ProcessIsMain = 1835100526, // 'main'
```
mandel-macaque, line 192: AFAIK we should say when it was added.

src/AudioUnit/AudioUnit.cs
```diff
@@ -403,7 +409,7 @@ public static uint GetCurrentInputDevice ()
 			var theAddress = new AudioObjectPropertyAddress (
 				AudioObjectPropertySelector.DefaultInputDevice,
 				AudioObjectPropertyScope.Global,
-				AudioObjectPropertyElement.Master);
+				AudioObjectPropertyElement.Main);
```
mandel-macaque, line 412: @spouliot  do we trust apple to do the correct things, or should we just use master ?

mandel-macaque (changes requested):

src/AudioUnit/AUEnums.cs
```diff
@@ -212,7 +215,9 @@ public enum AudioObjectPropertyScope : uint
 
 	public enum AudioObjectPropertyElement : uint
 	{
+		[Obsolete ("Use the 'Main' element instead.")]
```
spouliot, line 218: you can wrap this inside a

```
#if !XAMCORE_4_0
...
#endif
```

so a new, future profile won't require extra work to remove obsolete code.

src/AudioUnit/AudioUnit.cs
```diff
@@ -389,10 +390,15 @@ public uint GetCurrentDevice (AudioUnitScopeType scope, uint audioUnitElement =
 #if !XAMCORE_3_0 || MONOMAC
 #if !MONOMAC
 		[Obsolete ("This API is not available on iOS.")]
+#endif
+#if NET
+    [SupportedOSPlatform ("maccatalyst15.0")]
+#else
+    [MacCatalyst (15,0)]
```
spouliot, line 397: indent is incorrect, use tabs

src/AudioUnit/AudioUnit.cs
spouliot: both enum members have the same `0` value so there's no risk for this one

src/AudioUnit/AUEnums.cs
```diff
@@ -200,6 +202,7 @@ public enum AudioObjectPropertySelector : uint
 		ActualSampleRate = 1634955892,// 'asrt',
 		ClockDevice = 1634755428, // 'apcd',
 		IOThreadOSWorkgroup = 1869838183, // 'oswg'
+		ProcessMute = 1634758765, // 'appm'
```
spouliot, line 205: ^ needs availability attributes - for both `NET` and legacy

src/AudioUnit/AUEnums.cs
spouliot: unlike the other enum this one has a different value so it's useful to know when it should be avoided

spouliot (changes requested):

rolfbjarne (approved):

src/AudioUnit/AUEnums.cs
tj-devel709: With a comment?

tj-devel709:
@spouliot and others
When I checked the apidiff, `ProcessIsMaster` changed to `ProcessIsMain` and `ProcessMute` was added only in the iOS diff and mac diff. These were added in MacCatalyst, and nothing happened in tv and watch.
Therefore I am concluding these are only available in ios, mac, and mac catalyst

src/AudioUnit/AudioUnit.cs
```diff
@@ -389,10 +390,15 @@ public uint GetCurrentDevice (AudioUnitScopeType scope, uint audioUnitElement =
 #if !XAMCORE_3_0 || MONOMAC
 #if !MONOMAC
 		[Obsolete ("This API is not available on iOS.")]
+#endif
+#if NET
+	[SupportedOSPlatform ("maccatalyst15.0")]
+#else
+	[MacCatalyst (15,0)]
```
spouliot, line 397: indent looks wrong

src/AudioUnit/AUEnums.cs
spouliot: if it's not available in tvOS then it needs `[NoTV]` and `[SupportedOSPlatform ("tvos")]`
unless the enum type itself is not available ?

otherwise it implies availability for tvOS 9 (for legacy) or tvOS 10 (for net6)

spouliot (changes requested):

spouliot (approved):

mandel-macaque (approved):

PR #12473, opened 2021-08-18
[generator] Simplify native enums logic.

rolfbjarne:
Instead of generating one native P/Invoke signature with an int parameter and
another with a long parameter for methods that take [Native] enums, generate a
single nint parameter (and the same for the unsigned version).

This simplifies both the generator code and the generated code. The generator
diff contains *a lot* of changes like this:

    -	if (IntPtr.Size == 8) {
    -		ret = (ARAppClipCodeUrlDecodingState) global::ObjCRuntime.Messaging.Int64_objc_msgSend (this.Handle, Selector.GetHandle ("urlDecodingState"));
    -	} else {
    -		ret = (ARAppClipCodeUrlDecodingState) global::ObjCRuntime.Messaging.int_objc_msgSend (this.Handle, Selector.GetHandle ("urlDecodingState"));
    -	}
    +	ret = (ARKit.ARAppClipCodeUrlDecodingState) (long) global::ObjCRuntime.Messaging.nint_objc_msgSend (this.Handle, Selector.GetHandle ("urlDecodingState"));

An unlinked Xamarin.iOS.dll is ~300kb smaller (once linked the difference
should be minimal though).

I also made the min/max detection logic (check for int32.MinValue/MaxValue and
convert to int64.MinValue/MaxValue) specific to ARCH_32, since we don't need
it in 64-bit code.


tests/generator/tests/nativeenum.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using Foundation;
+using ObjCRuntime;
+
+namespace NS {
+	[Native]
+	public enum MyEnum6 : long {
```
rolfbjarne, line 7: It looks weird to start at "6", but I have more local tests that will fill in the missing numbering 😄

src/generator.cs
```diff
@@ -1057,8 +1052,8 @@ public void RegisterMethodName (string method_name)
 	//
 	// Helpers
 	//
-	string MakeSig (MethodInfo mi, bool stret, bool aligned = false, EnumMode enum_mode = EnumMode.Compat ) { return MakeSig ("objc_msgSend", stret, mi, aligned, enum_mode); }
-	string MakeSuperSig (MethodInfo mi, bool stret, bool aligned = false, EnumMode enum_mode = EnumMode.Compat) { return MakeSig ("objc_msgSendSuper", stret, mi, aligned, enum_mode); }
+	string MakeSig (MethodInfo mi, bool stret, bool aligned = false ) { return MakeSig ("objc_msgSend", stret, mi, aligned); }
```
spouliot, line 1055: minor: extra space added after `false`

spouliot (approved):
nice one ❤️ 

I once looked to see if we could share more p/invoke signatures (same size arguments) but totally missed that one!

dalexsoto (approved):

spouliot (approved):

PR #12474, opened 2021-08-18
[msbuild] Add a task to compute the path to the AOT compiler.

rolfbjarne:
The MonoAotCrossCompiler item group is empty when executing remotely from
Windows, so in that case we evaluate a csproj on the Mac side to compute the
AOT compiler path instead of relying in the task input from Windows.

Ref: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1362367


lambdageek:
@rolfbjarne For `UseLLVM`, How do you set LLVMPath? Those files are also on the Mac in the runtime cross-compiler pack, I believe.

dotnet/targets/Xamarin.Shared.Sdk.targets
```diff
@@ -783,14 +784,26 @@
...
+			SessionId="$(BuildSessionId)"
+			Condition="'$(IsMacEnabled)' == 'true'"
+			MonoAotCrossCompiler="@(MonoAotCrossCompiler)"
+			RuntimeIdentifier="$(RuntimeIdentifier)"
+			TargetFrameworkMoniker="$(_ComputedTargetFrameworkMoniker)"
+		>
+			<Output TaskParameter="AotCompiler" PropertyName="_AOTCompiler" />
+		</FindAotCompiler>
+
+		<Error Text="The AOT compiler '$(_AOTCompiler)' does not exist." Condition="!Exists ($(_AOTCompiler))" />
```
emaf, line 798: This won't work from Windows, because the AOT Compiler won't exist there. Probably this check should be done from inside the FindAotCompiler task.

emaf (changes requested):

rolfbjarne:
@lambdageek 

> @rolfbjarne For `UseLLVM`, How do you set LLVMPath? Those files are also on the Mac in the runtime cross-compiler pack, I believe.

We compute it from the path to the AOT compiler (we assume it's the directory where the AOT compiler is).

> (I think you use the MonoAOTCompiler task from the runtime SDK, right?)

No, we haven't switched yet (but we still need the path to the LLVM compiler, so your question still applies).

emaf (approved):

dalexsoto (approved):

PR #12485, opened 2021-08-19
[msbuild] Rename the GetMinimumOSVersion task to ReadAppManifest and make it read more properties from the app manifest.

rolfbjarne:
* Read CFBundleDisplayName and CFBundleVersion and use them in the
  _CompileITunesMetadata task.

* Read numerous other app manifest values and pass them to the ACTool and
  IBTool tasks.

This makes it possible to not parse the Info.plist in these tasks, which will
become more complicated in the future, when we might either not have an
Info.plist, or have many partial ones.

Also enable nullability.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ReadAppManifestTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ReadAppManifestTaskBase.cs
@@ -15,6 +15,12 @@ namespace Xamarin.MacDev.Tasks {
 		public string? SdkVersion { get; set; }
 
 		[Output]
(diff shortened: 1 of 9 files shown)


emaf (approved):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ReadAppManifestTaskBase.cs
```diff
@@ -0,0 +1,86 @@
+#nullable enable
```
spouliot, line 1: ❤️

spouliot (approved):

dalexsoto (approved):

PR #12495, opened 2021-08-19
[iAD] Updates for DotNet Attributes

tj-devel709:


rolfbjarne (approved):

mandel-macaque (approved):

dalexsoto (approved):

src/iAd/ADCompat.cs
```diff
@@ -289,22 +289,50 @@ public virtual void WillLoad(ADBannerView bannerView)
 	// some of this API is still provided
 	public partial class ADClient : NSObject {
 
-		[Obsoleted (PlatformName.iOS, 15, 0, PlatformArchitecture.None, Constants.iAdRemoved)]
```
spouliot, line 292: why to you change the existing API ?
has something changed in the headers ?

if not then the legacy attributes should remain identical - this is what will be used to review the PRs

src/iAd/ADCompat.cs
```diff
@@ -601,7 +629,11 @@ public static bool ShouldPresentInterstitialAd (this UIViewController This)
 #endif
 	public static class IAdPreroll {
 
-		[Introduced (PlatformName.iOS, 8, 0, PlatformArchitecture.All, null)]
+		#if !NET
+				[iOS (8,0)]
+		#else
+				[SupportedOSPlatform ("ios8.0")]
+		#endif
```
spouliot, line 636: un-indent the `#if`
and let the attributes be on their normal indentation

spouliot (changes requested):

src/iAd/ADCompat.cs
tj-devel709: Since these used to just be obsolete, I should not have the deprecated nor the Unsupported right? I should just have the obsolete?

spouliot: The `[UnsuppotedOSPlatform]` with the version number is fine, it helps the new analyzer.

spouliot: However the `[Deprecated]` should not be added (unless the header changes)

and `[Obsoloted]` (with a `d` suffix) should not be used in the `NET` section.

src/iAd/ADCompat.cs
```diff
@@ -289,22 +289,50 @@ public virtual void WillLoad(ADBannerView bannerView)
...
 		}
 
-		[Obsoleted (PlatformName.iOS, 15, 0, PlatformArchitecture.None, Constants.iAdRemoved)]
+#if !NET
+		[Deprecated (PlatformName.iOS, 15, 0, message: "The iAd framework has been removed from iOS")]
+#else
+		[UnsupportedOSPlatform ("ios15.0")]
+#if IOS
+		[Obsolete ("Starting with ios15.0 The iAd framework has been removed from iOS", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
+#endif
```
spouliot, line 310: same here

src/iAd/ADCompat.cs
```diff
@@ -289,22 +289,50 @@ public virtual void WillLoad(ADBannerView bannerView)
...
```
spouliot, line 322: same here

src/iAd/ADCompat.cs
```diff
@@ -289,22 +289,50 @@ public virtual void WillLoad(ADBannerView bannerView)
...
```
spouliot, line 334: same here

src/iAd/ADCompat.cs
```diff
@@ -601,7 +629,11 @@ public static bool ShouldPresentInterstitialAd (this UIViewController This)
 #endif
 	public static class IAdPreroll {
 
-		[Introduced (PlatformName.iOS, 8, 0, PlatformArchitecture.All, null)]
+#if !NET
+		[iOS (8,0)]
+#else
+		[SupportedOSPlatform ("ios8.0")]
```
spouliot, line 635: 8.0 < dot net minimum (10)

src/iAd/ADCompat.cs
```diff
@@ -601,7 +629,11 @@ public static bool ShouldPresentInterstitialAd (this UIViewController This)
 #endif
 	public static class IAdPreroll {
 
-		[Introduced (PlatformName.iOS, 8, 0, PlatformArchitecture.All, null)]
+#if !NET
+		[iOS (8,0)]
```
spouliot, line 633: correct, but it would be best not to change the existing code to ease reviews

spouliot (changes requested):

spouliot (approved):

PR #12501, opened 2021-08-20
[dotnet] Updates the path to illink when building from Windows

emaf:
The directory that contains illink.dll was renamed to net6.0


rolfbjarne (approved):

rolfbjarne:
Test failures are all unrelated network failures (which will be fixed with #12509).

PR #12503, opened 2021-08-20
Correct Xamarin.Mac profiling properties

Therzok:
Seems like it was broken as of https://github.com/xamarin/xamarin-macios/commit/7c15428fc20d9ebecb48f999654b701707461c72

If MtouchProfiling is not set for Xamarin.Mac apps, it would not have a value. This behavioural change was also not replicated in the IDE, which still sets `Profiling`.

Change the property so it inherits MtouchProfiling if Profiling is not set and detauls to false.


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2160).

rolfbjarne (approved):

emaf (approved):

PR #12512, opened 2021-08-23
[xharness] Move loading old test results into the background.

rolfbjarne:
At startup, we load old test results that may exist on disk. However, this may
take a while, especially since the disk is usually quite busy already with
everything else we do at startup, so don't wait for this to finish before
displaying the UI.


spouliot (approved):

rolfbjarne:
Test failures are all unrelated, and fixed by #12509.

PR #12516, opened 2021-08-23
[msbuild/dotnet] Rework how app manifests are created.

rolfbjarne:
How we create the app manifest (Info.plist) has to be modified so that we can
add support for getting all the values from MSBuild properties (i.e. no
Info.plist in the project), as well as having multiple partial app manifests
that all get merged into the final app manifest.

Here's the new process:

1. The user can specify values in multiple ways:

   * An Info.plist in their project file (as always, by using a `None` item
     with filename "Info.plist" or with a `Link` metadata with filename
     "Info.plist"). We find this Info.plist in the DetectAppManifest target.
   * A partial plist in their project (using the `PartialAppManifest` item
     group)
   * Some MSBuild properties can also add values.

   The general precedence is: MSBuild properties can be overridden by the
   Info.plist, which can be overridden by a partial plist.

2. In the `CompileAppManifest` target we get all the inputs from above, and
   compute a temporary app manifest, which is written to a temporary output
   file.

3. In the `ReadAppManifest` target, we read the temporary output file and
   outputs numerous MSBuild properties (most of them private)

4. We run other targets that may add more entries to the final app manifest
   (these tasks might depend on the values from `ReadAppManifest`). These
   entries are written to partial plists, and added to the
   _PostCompilePartialAppManifest item group.

   Currently the targets that can add more entries to the app manifest are
   _CompileImageAssets and  _CompileCoreMLModels.

5. In the new `WriteAppManifest` target, we read the temporary output file
   from `ReadAppManifest` + any `_PostCompilePartialAppManifest` files and
   merge them all together to get the final Info.plist.

This also required moving the computation of CFBundleIdentifier from the
DetectSigningIdentity task to the CompileAppManifest task, which also meant
reordering these two tasks, so that the DetectSigningIdentity task is executed
after the CompileAppManifest task (technically after the ReadAppManifest
task), because the DetectSigningIdentity task needs to know the bundle
identifier.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
@@ -1,0 +1,51 @@
+using System;
+using System.IO;
+using System.Linq;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+using Xamarin.Localization.MSBuild;
+using Xamarin.Utils;
+
+namespace Xamarin.MacDev.Tasks
+{
+	public abstract class WriteAppManifestTaskBase : XamarinTask
+	{
+		#region Inputs
+
+		[Required]
+		public string AppBundleManifest { get; set; }
+
+		[Required]
+		public ITaskItem[] AppManifests { get; set; }
+
+		#endregion
+
+		public override bool Execute ()
+		{
+			PDictionary plist;
+
+			var firstManifest = AppManifests [0].ItemSpec;
+			try {
+				plist = PDictionary.FromFile (firstManifest);
+			} catch (Exception ex) {
+				Log.LogError (null, null, null, firstManifest, 0, 0, 0, 0, MSBStrings.E0010, ex.Message);
+				return false;
+			}
+
+			if (AppManifests.Length >1)
+				CompileAppManifestTaskBase.MergePartialPLists (this, plist, AppManifests.Skip (1));
+
+			// Remove any IDE specific keys we don't want in the final app manifest.
+			plist.Remove (ManifestKeys.XSLaunchImageAssets);
+			plist.Remove (ManifestKeys.XSAppIconAssets);
+
+			// write the resulting app manifest
+			Directory.CreateDirectory (Path.GetDirectoryName (AppBundleManifest));
+			plist.Save (AppBundleManifest, true, true);
+
+			return !Log.HasLoggedErrors;
+		}
+	}
+}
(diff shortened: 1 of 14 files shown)


msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.IO;
+using System.Linq;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+using Xamarin.Localization.MSBuild;
+using Xamarin.Utils;
+
```
spouliot, line 10: it would be nice, for new files, to enable nullable

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
```diff
@@ -0,0 +1,51 @@
...
+			var firstManifest = AppManifests [0].ItemSpec;
+			try {
+				plist = PDictionary.FromFile (firstManifest);
+			} catch (Exception ex) {
+				Log.LogError (null, null, null, firstManifest, 0, 0, 0, 0, MSBStrings.E0010, ex.Message);
+				return false;
+			}
+
+			if (AppManifests.Length >1)
```
spouliot, line 37: minor: missing space

msbuild/Xamarin.MacDev.Tasks/Tasks/WriteAppManifest.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.Build.Framework;
+using Xamarin.Messaging.Build.Client;
+
```
spouliot, line 3: same, wrt nullable

spouliot (approved):
huge update and affect is not limited to net6
c.c. @dalexsoto worth 👀 on the samples results

chamons (approved):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
```diff
@@ -0,0 +1,51 @@
...
+using Xamarin.Localization.MSBuild;
+
+namespace Xamarin.MacDev.Tasks
+{
+	public abstract class WriteAppManifestTaskBase : XamarinTask
+	{
+		#region Inputs
+
+		[Required]
+		public string? AppBundleManifest { get; set; }
```
emaf, line 18: This needs to be an Output property of the ITaskItem type, so the file is also created on Windows.

emaf (changes requested):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/WriteAppManifestTaskBase.cs
rolfbjarne: Fixed!

emaf (approved):

PR #12533, opened 2021-08-24
[ThreadNetwork] Add new framework Xcode 13 beta 5.

mandel-macaque:


src/threadnetwork.cs
```diff
@@ -0,0 +1,74 @@
...
+	[BaseType (typeof (NSObject))]
+	interface THClient
+	{
+		[Async]
+		[Export ("retrieveAllCredentials:")]
+		void RetrieveAllCredentials (Action<NSSet<THCredentials>, NSError> completion);
+
+		[Async]
+		[Export ("deleteCredentialsForBorderAgent:completion:")]
+		void DeleteCredentialsForBorderAgent (NSData borderAgentId, Action<NSError> completion);
```
spouliot, line 19: remove `ForBorderAgent` since it's the first parameter, not the action

src/threadnetwork.cs
```diff
@@ -0,0 +1,74 @@
...
+
+		[Async]
+		[Export ("retrieveCredentialsForBorderAgent:completion:")]
+		void RetrieveCredentialsForBorderAgent (NSData borderAgentId, Action<THCredentials, NSError> completion);
```
spouliot, line 23: remove `ForBorderAgent` since it's the first parameter, not the action

src/threadnetwork.cs
```diff
@@ -0,0 +1,74 @@
...
+
+		[Async]
+		[Export ("storeCredentialsForBorderAgent:activeOperationalDataSet:completion:")]
+		void StoreCredentialsForBorderAgent (NSData borderAgentId, NSData activeOperationalDataSet, Action<NSError> completion);
```
spouliot, line 27: remove `ForBorderAgent` since it's the first parameter, not the action

src/threadnetwork.cs
```diff
@@ -0,0 +1,74 @@
...
+
+		[Async]
+		[Export ("retrievePreferredCredentials:")]
+		void RetrievePreferredCredentials (Action<THCredentials, NSError> completion);
+
+		[Async]
+		[Export ("retrieveCredentialsForExtendedPANID:completion:")]
+		void RetrieveCredentialsForExtendedPanid (NSData extendedPanid, Action<THCredentials, NSError> completion);
```
spouliot, line 35: remove `ForExtendedPanid` since it's the first parameter, not the action

spouliot: looking at context / doc [1], the first argument should be `extendedPanId` (uppercase 'I')

[1] _Requests Thread credentials for an extended Personal Area Network (PAN) ID._

spouliot (changes requested):

src/threadnetwork.cs
mandel-macaque: I don't think so, there is a reason for the 'ForBorderAgent'. We have another method called 'RetrieveCredentialsForExtendedPanId' which takes a NsData too, I fear that people will either confuse the NSData between a PanId and a BorderId, so I prefer to be explicit.

src/threadnetwork.cs
mandel-macaque: same.

src/threadnetwork.cs
mandel-macaque: I missed that.

mandel-macaque, line 35:
```suggestion
		void RetrieveCredentialsForExtendedPanId (NSData extendedPanId, Action<THCredentials, NSError> completion);
```

src/threadnetwork.cs
spouliot: make sense

I'd usually check the swift names - but it looks like they are not (yet?) on the web doc

spouliot (approved):

dalexsoto (approved):

src/threadnetwork.cs
```diff
@@ -0,0 +1,74 @@
...
+	[iOS (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface THCredentials : NSSecureCoding
+	{
+		[NullAllowed, Export ("networkName")]
+		string NetworkName { get; }
+
+		[NullAllowed, Export ("extendedPANID")]
+		NSData ExtendedPanid { get; }
```
rolfbjarne, line 47:
```suggestion
		NSData ExtendedPanId { get; }
```

rolfbjarne (approved):

Issue #12569, opened 2021-08-27
introspection failures: [FAIL] [Rule 3] Missing '.' in attribute's message: "Starting with ios15.0 The iAd framework has been removed from iOS" - Member name: ErrorDomain, Type: ADClient


rolfbjarne:
Introspection fails on iOS 64-bit simulator for .NET:

```
[FAIL] [Rule 3] Missing '.' in attribute's message: "Starting with ios15.0 The iAd framework has been removed from iOS" - Member name: ErrorDomain, Type: ADClient
[FAIL] [Rule 3] Missing '.' in attribute's message: "Starting with ios15.0 The iAd framework has been removed from iOS" - Member name: DetermineAppInstallationAttribution, Type: ADClient
[FAIL] [Rule 3] Missing '.' in attribute's message: "Starting with ios15.0 The iAd framework has been removed from iOS" - Member name: LookupAdConversionDetails, Type: ADClient
[FAIL] [Rule 3] Missing '.' in attribute's message: "Starting with ios15.0 The iAd framework has been removed from iOS" - Member name: LookupAdConversionDetailsAsync, Type: ADClient
[FAIL] AttributeTypoTest : Attributes have typos!
```

Example: https://github.com/xamarin/xamarin-macios/pull/12516#issuecomment-906833797

This was probably introduced in f0373ee4f7140a70138fc919baf3fc3381656ef7.

PR #12570, opened 2021-08-27
[msbuild] Make the ParseBundlerArgumentsTaskBase take the verbosity as input as well.

rolfbjarne:
This way we'll honor a _BundlerVerbosity value passed on the command line.

This fixes an inconvenience when we have when we want to increase verbosity
for a test run, which can be done by setting 'MtouchExtraArgs=-v'. This works
fine until the test we want to run wants to use MtouchExtraArgs for its own
purposes (1). If we set MtouchExtraArgs on the command line, then that value
will be global and won't be able to be overridden by the test, and the test
will fail.

With this fix we can pass _BundlerVerbosity=1 directly when building the test
to avoid the whole problem.

1. For instance, most tests pass an argument in MtouchExtraArgs to use
dlsym for NUnit.Framework.dll, which contains a P/Invoke to a native function
that doesn't exist - see the tests/nunit.framework.targets file.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ParseBundlerArgumentsTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/ParseBundlerArgumentsTaskBase.cs
@@ -34,6 +34,7 @@ namespace Xamarin.MacDev.Tasks {
 		[Output]
 		public string Registrar { get; set; }
 
+		// This is input too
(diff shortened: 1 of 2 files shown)


spouliot (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated (both due to https://github.com/xamarin/xamarin-macios/issues/12569).

PR #12580, opened 2021-08-30
Generated msi files for all packages

mattleibow:
Since both tvOS and macOS are needed on Windows, we also need the .msi versions.

### DIFF
--- a/tools/devops/automation/templates/release/vs-insertion-prep.yml
+++ b/tools/devops/automation/templates/release/vs-insertion-prep.yml
@@ -79,10 +79,5 @@ stages:
   parameters:
     dependsOn: prepare_release
     symbolArtifactName: nuget-signed
-    symbolArtifactPatterns: |
-      Microsoft.NET.Sdk.iOS.Manifest*.nupkg
-      Microsoft.NET.Sdk.MacCatalyst.Manifest*.nupkg
-      Microsoft.iOS*.nupkg
-      Microsoft.MacCatalyst*.nupkg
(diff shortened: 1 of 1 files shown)


spouliot (changes requested):
You should not commit to `release/6.0.1xx-rc.1` without
* a previous commit to `main`
* an explanation why this does not apply to `main`

pjcollins (changes requested):
If we want to do this we need to also remove the filters from https://github.com/xamarin/xamarin-macios/blob/main/tools/devops/automation/templates/release/vs-insertion-prep.yml#L29-L33. However, this may cause the msi conversion task to fail as some packs from dependencies (especially in runtime?) may only be supported on macOS.

This will also cause VS setup authoring to be generated for macOS and tvOS workloads, which means that these two workloads will end up in our next VS insertion. It's possible that these changes would create a need for corresponding dotnet/runtime build infra changes before we would be able to insert into VS again...

Finally, this opens up a broader question about what the experience for these workloads on Windows should look like, and I think we should have an answer/plan for that before making this change.

I think the better short term fix would be to make these workloads only supported on macOS, until some of the questions above are sorted:
https://github.com/dotnet/runtime/blob/2a3141a5a202e8b86d52b53d346fce145f68d7d1/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.json.in#L97

mattleibow:
I recreated the PR on main: https://github.com/xamarin/xamarin-macios/pull/12581

PR #12581, opened 2021-08-30
Generate msi files and VS setup authoring for all .NET 6 platforms

mattleibow:
Since both tvOS and macOS are needed on Windows, we also need the .msi versions.

tvOS and macOS are needed on Windows so that we can built multi-targeted libraries.

This also needs to go to RC 1

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -128,7 +128,7 @@ resources:
   - repository: templates
     type: github
     name: xamarin/yaml-templates
-    ref: refs/heads/main
+    ref: refs/heads/pjc/shorten-manifests
     endpoint: xamarin
 
   - repository: maccore
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -8,7 +8,7 @@
 parameters:
 - name: runTests
   type: boolean
-  default: true
+  default: false
 
 - name: runDeviceTests
   type: boolean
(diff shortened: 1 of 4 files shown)


mandel-macaque (approved):
+1 I have pushed to origins to see it running in the internal pool which is diff to the external one.

pjcollins:
I made a change last week to only run signing/conversion on main/release branches. @mattleibow with your permission I will push some changes here to try to test the msi generation changes and a VS build as well.

mattleibow:
For reference of how I use this today, it is mainly for libraries, but I do this: https://github.com/mono/SkiaSharp/blob/main/binding/SkiaSharp/SkiaSharp.csproj#L4

So nice to have one platform to build everything.

dalexsoto:
Just parsed it again this still needs work

tools/devops/automation/build-pipeline.yml
```diff
@@ -258,7 +258,7 @@ stages:
         enableDotnet: ${{ parameters.enableDotnet }}
 
 # .NET 6 Release Prep and VS Insertion Stages, only execute them when the build comes from an official branch and is not a schedule build from OneLoc
-- ${{ if and(ne(variables['Build.Reason'], 'Schedule'), or(eq(variables['Build.SourceBranch'], 'refs/heads/main'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'))) }}:
+- ${{ if and(ne(variables['Build.Reason'], 'Schedule'), or(eq(variables['Build.SourceBranch'], 'refs/heads/mattleibow/tvos-msi'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'))) }}:
```
spouliot, line 261: ^ can't be committed as-is

spouliot (changes requested):

tools/devops/automation/build-pipeline.yml
pjcollins: I converted this back to a draft for testing, we'll need .msi conversion and vs insertion to run to properly test this. I'll revert the testing changes if this does end up working.

pjcollins:
Tested with https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/348300, VS is building and there aren't any new sign/symbol issues that would block so this should be ready. The one symbol failure on the VS PR is unrelated, and due this branch missing https://github.com/xamarin/xamarin-macios/commit/12d8b37f4f876e4f8fe34bbbe802ea073166ed6a.

dalexsoto (approved):

PR #12590, opened 2021-08-30
[Network] Add Support for DotNet Attributes

tj-devel709:


src/Network/NWEthernetChannel.cs
```diff
@@ -24,10 +25,16 @@
 
 namespace Network {
 
+#if !NET
 	[NoWatch, NoTV, NoiOS, Mac (10,15)]
+#endif
```
mandel-macaque, line 30: Missing mac10.15 for net core, looking at the min versions I have: [assembly: SupportedOSPlatform ("macos10.14")]

mandel-macaque (changes requested):

src/Network/NWEthernetChannel.cs
spouliot: it's fine, can't annotate `delegate` on dotnet

spouliot (approved):

src/Network/NWEthernetChannel.cs
mandel-macaque: that's lame

mandel-macaque (approved):

PR #12592, opened 2021-08-30
[CoreText] Add Support for DotNet Attributes

tj-devel709:


src/CoreText/CTFontManager.cs
```diff
@@ -44,20 +45,46 @@ namespace CoreText {
...
 		Persistent = 2,
+#if !NET
 		[NoiOS][NoTV][NoWatch]
+#else
+		[UnsupportedOSPlatform ("tvos")]
+		[UnsupportedOSPlatform ("ios")]
+		[UnsupportedOSPlatform ("maccatalyst")]
+		[UnsupportedOSPlatform ("macos")]
+#endif
 		Session = 3,
```
mandel-macaque, line 63: I do not believe it was removed from mac os, and I wrong?

src/CoreText/CTFontManager.cs
```diff
@@ -44,20 +45,46 @@ namespace CoreText {
...
 #if !XAMCORE_4_0
+#if !NET
 		[NoiOS][NoTV][NoWatch] // historically not available under the old name
+#else
+		[UnsupportedOSPlatform ("tvos")]
+		[UnsupportedOSPlatform ("ios")]
+		[UnsupportedOSPlatform ("maccatalyst")]
+		[UnsupportedOSPlatform ("macos")]
+#endif
 		User = Persistent,
```
mandel-macaque, line 73: Same. what happened with mac?

src/CoreText/CTFontManager.cs
```diff
@@ -68,18 +95,31 @@ public partial class CTFontManager {
...
+#if !NET
 		[Obsolete ("API not available on iOS, it will always return false.")]
 		[Deprecated (PlatformName.MacOSX, 10, 6)]
 		[Unavailable (PlatformName.iOS)]
+#else
+		[UnsupportedOSPlatform ("ios")]
+		[UnsupportedOSPlatform ("macos10.6")]
+#if MONOMAC
+		[Obsolete ("Starting with macos10.6 API not available on iOS, it will always return false.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
+#endif
```
mandel-macaque, line 121: on iOS but we have a #if MONOMAC is that correct?

src/CoreText/CTFontManager.cs
```diff
@@ -156,7 +215,9 @@ static NSArray EnsureNonNullArray (object [] items, string name)
 			}
 		}
 
+#if !NET
 		[Watch (6,0), TV (13,0), Mac (10,15), iOS (13,0)]
+#endif
```
mandel-macaque, line 220: Missing 10.15 for NET since the min is 10.14.

mandel-macaque (changes requested):

src/CoreText/CTFontManager.cs
spouliot: The old message was wrong since the API always return `false`, even if macOS was lower than 10.6 (as we're not calling anything)

src/CoreText/CTFontManager.cs
spouliot: @mandel-macaque that's fine, the dotnet attribute can't be applied to delegates - which is half true (the delegate can be used whatever the OS version) and half false (you can keep looking for an available API that can consume it)

spouliot (approved):

dalexsoto (approved):

src/CoreText/CTFontManager.cs
tj-devel709: Good catch!

src/CoreText/CTFontManager.cs
tj-devel709: Also good catch!

src/CoreText/CTFontManager.cs
tj-devel709: Should I change the obsolete message in dotnet to be 
"Starting with macos10.6 API will always return false."
?

spouliot: The API always `return false` irrespective of the OS or platform
Make this a single, OS-less and version-less, `[Obsolete]`

mandel-macaque (approved):

tj-devel709:
The test failures are unrelated: https://github.com/xamarin/maccore/issues/2414

PR #12603, opened 2021-08-31
[Foundation] Add Support for DotNet Attributes

tj-devel709:


src/Foundation/NSMetadataItem.cs
```diff
@@ -162,1053 +169,1347 @@ bool GetBool (NSString key)
 			}
 		}
 
-		[iOS (7,0)][Mac (10,9)]
```
tj-devel709, line 165: Should I keep this in? I removed it since the class has the same attributes, but it makes changes to the diff

src/Foundation/NSMetadataItem.cs
```diff
@@ -162,1053 +169,1347 @@ bool GetBool (NSString key)
...
 		[iOS (8,0)][Mac (10,10)]
+#endif
 		public bool UbiquitousItemIsExternalDocument {
 			get {
 				return GetBool (NSMetadataQuery.UbiquitousItemIsExternalDocumentKey);
 			}
 		}
 
+#if !NET
 		[iOS (8,0)][Mac (10,9)]
```
tj-devel709, line 205: Mac is already included in the class. Is it worth removing?

src/Foundation/NSMetadataItem.cs
spouliot: good catch! it's fine to remove

src/Foundation/NSMetadataItem.cs
spouliot: yes

spouliot (approved):

mandel-macaque (approved):

src/Foundation/NSMetadataItem.cs
tj-devel709: Wait.. I think I just got confused. I don't think these are actually exposed by the class and do need to be included

PR #12628, opened 2021-09-03
[UIKit] Add Support for DotNet Attributes

tj-devel709:
Saved for last! Perhaps to be merged after the UIKit binding updates


spouliot (approved):

src/UIKit/UIStringDrawing.cs
```diff
@@ -6,87 +6,147 @@
...
 using ObjCRuntime;
+using System.Runtime.Versioning;
 
 namespace UIKit {
 	public unsafe static partial class UIStringDrawing  {
+#if !NET
 		[Deprecated (PlatformName.iOS, 7, 0, message : "Use NSString.DrawString(CGPoint, UIStringAttributes) instead.")]
+#else
+		[UnsupportedOSPlatform ("ios7.0")]
+		[Obsolete ("Starting with ios7.0 use NSString.DrawString(CGPoint, UIStringAttributes) instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
```
rolfbjarne, line 17: I wonder if we should just simplify this to:

```suggestion
		[UnsupportedOSPlatform ("ios")]
		[Obsolete ("Use NSString.DrawString(CGPoint, UIStringAttributes) instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
```

since we don't support any iOS version earlier than iOS 7.0 anyway.

spouliot: Two small issues with the proposal:

* it's not what the generator does for generated bindings (but that could be changed too)

* developers updating an old app (e.g. new language) could face dire-sounding warnings that they are using API that are not supported. I'd much rather know (be warned of) that I'm using something **very old** that something that's not working.

Whatever is decided it should be applied uniformly for all attributes (intro could easily check for "too old" versions).

dalexsoto (approved):

tj-devel709:
There is a bug that I will link here where Xharness tells us there is an error but the logs disagree. Running this locally, I was able to see that the Monotouch-test for Mac Catalyst Debug [Dotnet] passes making this failure irrelevant in this PR.

image

dalexsoto (approved):

mandel-macaque:
The fsharp failures is unrelated and has been fixed by @rolfbjarne already.

PR #12638, opened 2021-09-06
[dotnet] Support SupportedOSPlatformVersion. Fixes #12336.

### LINKED ISSUE #12336: Ensure `MinimumOSVersion` is is consistent with`SupportedOSPlatformVersion`
We should ensure that the `MinimumOSVersion` property in the Info.plist is consistent with the `SupportedOSPlatformVersion` MSBuild property. `SupportedOSPlatformVersion` is a standardized across platforms and controls the SDK's platform analyzers.

https://github.com/dotnet/designs/search?q=SupportedOSPlatformVersion

A few options:
* emit a build error if the `MinimumOSVersion` property and the `SupportedOSPlatformVersion` MSBuild property do not match
* merge the `SupportedOSPlatformVersion` MSBuild property into the `MinimumOSVersion` plist value during app build
* set the `SupportedOSPlatformVersion` MSBuild property from the `MinimumOSVersion` plist property in a build target at start of build. the downside would be that it would not be available during evaluation.

rolfbjarne:
* Add support for the SupportedOSPlatformVersion MSBuild property, and write
  it to the Info.plist for the corresponding minimum OS version.
* If there are any minimum OS version in the Info.plist, we'll now show an
  error if it doesn't match SupportedOSPlatformVersion.

This unfortunately means that if there's any minimum OS version in any
Info.plist, then that will most likely have to be moved to the
SupportedOSPlatformVersion property (or removed entirely if that's the right
choice), since it's unlikely to match the default value for
SupportedOSPlatformVersion. However, this was deemed to be the best option for
the future (it's a one-time pain during migration).

Also add new tests, update existing tests, and update the templates.

Fixes https://github.com/xamarin/xamarin-macios/issues/12336.

### DIFF
--- a/dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
+++ b/dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
@@ -31,10 +31,10 @@
     },
     "minOSVersion": {
       "type": "parameter",
-      "description": "Overrides MinimumOSVersion in the Info.plist",
+      "description": "Overrides SupportedOSPlatformVersion in the project file",
       "replaces": "minOSVersion",
       "datatype": "string",
-      "defaultValue": "11.2"
+      "defaultValue": "14.0"
     }
   },
   "defaultName": "tvOSApp1"
--- a/dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
+++ b/dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
@@ -34,7 +34,7 @@
       "description": "Overrides SupportedOSPlatformVersion in the project file",
       "replaces": "minOSVersion",
       "datatype": "string",
-      "defaultValue": "14.0"
+      "defaultValue": "11.2"
     }
   },
   "defaultName": "tvOSApp1"
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
@@ -183,7 +183,7 @@ namespace Xamarin.MacDev.Tasks
 				return false;
 			} else if (!string.IsNullOrEmpty (convertedSupportedOSPlatformVersion) && convertedSupportedOSPlatformVersion != minimumOSVersionInManifest) {
 				// SupportedOSPlatformVersion and the value in the Info.plist are not the same. This is an error.
-				Log.LogError (null, null, null, AppManifest, 0, 0, 0, 0, "The {0} value in the Info.plist ({1}) does not match the SupportedOSPlatformVersion value in the project file ({2}).", minimumVersionKey, minimumOSVersionInManifest, SupportedOSPlatformVersion); // FIXME: add tests + translated string.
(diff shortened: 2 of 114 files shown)


dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
```diff
@@ -31,10 +31,10 @@
     },
     "minOSVersion": {
       "type": "parameter",
-      "description": "Overrides MinimumOSVersion in the Info.plist",
+      "description": "Overrides SupportedOSPlatformVersion in the project file",
       "replaces": "minOSVersion",
       "datatype": "string",
-      "defaultValue": "11.2"
+      "defaultValue": "14.0"
```
spouliot, line 37: why ? it's still `11.2` for iOS
which I assume is to use the new `UIScene` API as they are newer than the `10.0` minimum ?

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
```diff
@@ -158,16 +172,25 @@ bool SetMinimumOSVersion (PDictionary plist)
...
+					minimumOSVersion = convertedSupportedOSPlatformVersion;
+				} else {
+					minimumOSVersion = SdkVersion;
+				}
 			} else if (!IAppleSdkVersion_Extensions.TryParse (minimumOSVersionInManifest, out var _)) {
 				Log.LogError (null, null, null, AppManifest, 0, 0, 0, 0, MSBStrings.E0011, minimumOSVersionInManifest);
 				return false;
+			} else if (!string.IsNullOrEmpty (convertedSupportedOSPlatformVersion) && convertedSupportedOSPlatformVersion != minimumOSVersionInManifest) {
+				// SupportedOSPlatformVersion and the value in the Info.plist are not the same. This is an error.
+				Log.LogError (null, null, null, AppManifest, 0, 0, 0, 0, "The {0} value in the Info.plist ({1}) does not match the SupportedOSPlatformVersion value in the project file ({2}).", minimumVersionKey, minimumOSVersionInManifest, SupportedOSPlatformVersion); // FIXME: add tests + translated string.
```
spouliot, line 186: `// FIXME: add tests + translated string.`

spouliot (changes requested):

dotnet/Templates/Microsoft.tvOS.Templates/tvos/.template.config/template.json
rolfbjarne: I'm not sure what happened here 🤔

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
rolfbjarne: 🤦

emaf (approved):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CompileAppManifestTaskBase.cs
rolfbjarne: @spouliot fixed now (I'd already written the test, so I just had to localize the message).

spouliot (approved):

dalexsoto (approved):

PR #12661, opened 2021-09-08
[msbuild] Don't require an Info.plist for .NET builds.

rolfbjarne:
We either have defaults or MSBuild properties for the most important values that
were previously required to be in the Info.plist, so now it doesn't make sense anymore
to require an Info.plist, when for simple cases it will just be an empty file.

This allows us to simplify some of our test projects and remove a few empty Info.plist files.


dalexsoto (approved):

chamons (approved):

emaf (approved):

rolfbjarne:
Looks like it got cancelled?

> ##[error]The operation was canceled.

PR #12666, opened 2021-09-08
[CoreFoundation] The native 'dispatch_queue_create_with_target' function is actually called 'dispatch_queue_create_with_target$V2'.

rolfbjarne:
Apple does this in their headers:

    #define DISPATCH_ALIAS_V2(sym)	 __asm__("_" #sym "$V2")

    dispatch_queue_t
    dispatch_queue_create_with_target(const char *_Nullable label,
    		dispatch_queue_attr_t _Nullable attr, dispatch_queue_t _Nullable target)
    		DISPATCH_ALIAS_V2(dispatch_queue_create_with_target);

Which means that the native compiler will call
'dispatch_queue_create_with_target$V2' when the source code says to call
'dispatch_queue_create_with_target'.

The only place I've run into this problem, is when building for tvOS (device),
and targetting exactly tvOS 10.0 (neither earlier or later), in which case the
linker fails:

    Undefined symbols for architecture arm64:
      "_dispatch_queue_create_with_target", referenced from:
          wrapper_managed_to_native_CoreFoundation_DispatchQueue_dispatch_queue_create_with_target_string_intptr_intptr in Xamarin.TVOS.dll.o

I filed this as a feedback with Apple some time ago [1], and Apple resolved it
as by design, saying "These symbols are renamed, please use the SDK."

Now I ran into it again with .NET, and it's become a bit more important, since
tvOS 10.0 is the earliest tvOS version we support, which means it'll be more
likely that customers use _exactly_ 10.0 as their target tvOS version. So I
looked into it again, and as far as I can tell, we can just call the '$V2'
variant instead of the original name everywhere.

Apple does the same thing for two other functions, but we haven't bound any of
those, so this only affects 'dispatch_queue_create_with_target' for us.

[1]: Bug ID 48076044: Can't reference 'dispatch_queue_create_with_target' when min tvOS version is exactly 10.0

### DIFF
--- a/src/CoreFoundation/Dispatch.cs
+++ b/src/CoreFoundation/Dispatch.cs
@@ -511,7 +511,7 @@ namespace CoreFoundation {
 		[TV (10,0)]
 		[Watch (3,0)]
 #endif
-		[DllImport (Constants.libcLibrary)]
+		[DllImport (Constants.libcLibrary, EntryPoint = "dispatch_queue_create_with_target$V2")]
 		extern static IntPtr dispatch_queue_create_with_target (string label, IntPtr attr, IntPtr target);
 
 		[DllImport (Constants.libcLibrary)]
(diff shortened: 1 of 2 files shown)


spouliot (approved):
V2 all the things!

dalexsoto (approved):
Approve v2

mandel-macaque (approved):
🤦

rolfbjarne:

PR #12678, opened 2021-09-09
[CI] In order to support APIScan we need to add the governance checks AFTER the build.

mandel-macaque:


tools/devops/automation/build-pipeline.yml
```diff
@@ -234,7 +234,8 @@ stages:
 
 - stage: build_packages
   displayName: 'Build'
-  dependsOn: []
+  dependsOn:
+  - build_packages
```
filipnavara, line 238: You can use `dependsOn` with cyclic reference to the same stage?!

mandel-macaque: nice catch 🤦

dalexsoto (approved):

rolfbjarne (approved):

mandel-macaque:
Failures re unrelated.

Issue #12640, opened 2021-09-07
[dotnet] LinkSdk.LinkSdkRegressionTest.SpecialFolder fails on tvOS


rolfbjarne:
Failure:

    LinkSdk.LinkSdkRegressionTest
        [FAIL] SpecialFolder :   /Users/builder/Library/Developer/CoreSimulator/Devices/76C018EB-3C82-46A5-BC38-B374E8ECABC6/data/Containers/Data/Application/57C91DB9-18F2-4B24-B112-B9BB7DC530ED/Library/Caches/Documents/.fonts
            Expected: True
            But was:  False
                at LinkSdk.LinkSdkRegressionTest.TestFolder (System.Environment+SpecialFolder folder, System.Boolean supported, System.Nullable`1[T] exists, System.Boolean readOnly) [0x00058] in /Users/builder/azdo/_work/1/s/xamarin-macios/tests/linker/ios/link sdk/LinkSdkRegressionTest.cs:851 
                at LinkSdk.LinkSdkRegressionTest.SpecialFolder () [0x0027c] in /Users/builder/azdo/_work/1/s/xamarin-macios/tests/linker/ios/link sdk/LinkSdkRegressionTest.cs:952 

Ref: https://github.com/xamarin/xamarin-macios/pull/12571#commitcomment-56023692

This was caused by: https://github.com/dotnet/runtime/commit/8da82f23adea445dd9d75a0ee5575fdae3cd8b8d

rolfbjarne:
This happens in the RC 1 branch as well: https://github.com/xamarin/xamarin-macios/pull/12633#issuecomment-914188584

filipnavara:
I would have expected it to be related to https://github.com/dotnet/runtime/commit/8da82f23adea445dd9d75a0ee5575fdae3cd8b8d (and other tvOS fixes made there). I will look into it.

rolfbjarne:
> I would have expected it to be related to [dotnet/runtime@8da82f2](https://github.com/dotnet/runtime/commit/8da82f23adea445dd9d75a0ee5575fdae3cd8b8d) (and other tvOS fixes made there). I will look into it.

It was a copy-paste mistake on my part, I meant to paste the commit you mention.

spouliot:
The _latest_ (after the committed fixes) failures did not happen locally, even after I reseted the simulator (and a `git clean`). Some state on the bot's simulator might have _permanently_ changed :(

rolfbjarne:
[...]

PR #12681, opened 2021-09-09
[generator] Use the newer, faster API alternatives when generating CIFilter subclasses

spouliot:


src/generator-filters.cs
```diff
@@ -353,10 +356,9 @@ void GenerateFilterSetter (string propertyType, string propertyName)
 			break;
 		case "string":
 			// NSString should not be added - it should be bound as a string
-			print ("using (var ns = new NSString (value))");
-			indent++;
-			print ("SetValue (\"{0}\", ns);", propertyName);
-			indent--;
+			print ("var ptr = CFString.CreateNative (value);");
+			print ($"SetHandle (\"{propertyName}\", ptr);");
```
rolfbjarne, line 360: Would it be better to add a `SetValue` overload that takes a string, and does the CFString creation there? It would result in less generated/duplicated code.

rolfbjarne (approved):

src/generator-filters.cs
spouliot: I like that. Let me have a look...

dalexsoto (approved):

mandel-macaque:
@spouliot  AFAIK we can land that.

mandel-macaque (approved):

spouliot:
Failures are unrelated to PR
* https://github.com/xamarin/maccore/issues/2451 (1)
* https://github.com/xamarin/xamarin-macios/issues/12640 (2)

PR #12698, opened 2021-09-10
[xharness] Set the EnableSGenConc value for all project configurations.

rolfbjarne:
* Update to use EnableSGenConc instead of MtouchEnableSGenConc (either works,
  the former is the new future).
* Set EnableSGenConc as a top-level property, so we make sure it's set no
  matter how the project is executed.

This makes it so that this setting work for .NET projects, where we're not using project configurations.


rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/xamarin-macios/issues/12640).

spouliot (approved):

PR #12701, opened 2021-09-10
[msbuild/dotnet] Make 'LinkSdk' the default linker mode for Mac Catalyst when building for Release, and 'None' the default when building for Debug. Fixes #12264.

### LINKED ISSUE #12264: Mac Catalyst: decide upon default linker behavior
Ref: https://github.com/xamarin/xamarin-macios/pull/12242

The current linker behavior for Mac Catalyst is to not link, neither debug nor release builds.

We should figure out if this is really what we want the default behavior to be.

Points to note:

* The JIT is not available on Mac Catalyst/ARM64. When building for ARM64 (either by itself, or when also building for X64), we enable the interpreter by default.
* We might decide to AOT compile release builds by default, not linking will increase the build time considerably.

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/pull/12264.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -1025,7 +1025,9 @@
 
 	<Target Name="_ComputeDefaultLinkMode" DependsOnTargets="_DetectSdkLocations">
 		<PropertyGroup>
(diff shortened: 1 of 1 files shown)


spouliot (approved):
and that answer https://github.com/xamarin/xamarin-macios/issues/12264

PR #12715, opened 2021-09-13
[src] Fix the library for a few P/Invokes.

rolfbjarne:
This matters because we use the library defined by the P/Invoke to determine
which frameworks to link with, and if we get it wrong, things like this may
happen:

    Undefined symbols for architecture arm64:
      "_AudioObjectGetPropertyData", referenced from:
          wrapper_managed_to_native_AudioUnit_AudioUnit_AudioObjectGetPropertyData_uint_AudioUnit_AudioObjectPropertyAddress__uint__intptr__uint__uint_ in Xamarin.MacCatalyst.dll.o

because we though that the AudioObjectGetPropertyData function was in the
AudioUnit framework, and not in CoreAudio where it really is, and thus we
didn't link with the CoreAudio framework.

Also add an introspection test to verify that the library our P/Invokes point
to is correct.

### DIFF
--- a/src/CoreFoundation/CFStream.cs
+++ b/src/CoreFoundation/CFStream.cs
@@ -239,7 +239,7 @@ namespace CoreFoundation {
 		[Obsolete ("Starting with macos12.0 use 'Network.framework' instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
 #endif
 #endif
-		[DllImport (Constants.CFNetworkLibrary)]
+		[DllImport (Constants.CoreFoundationLibrary)]
 		internal extern static void CFStreamCreatePairWithPeerSocketSignature (/* CFAllocatorRef */ IntPtr allocator, 
 			/* CFSocketSignature* */ ref CFSocketSignature sig, 
 			/* CFReadStreamRef* */ out IntPtr readStream, /* CFWriteStreamRef* */ out IntPtr writeStream);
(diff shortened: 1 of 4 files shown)


tests/introspection/ApiPInvokeTest.cs
```diff
@@ -238,15 +238,109 @@ protected void Check (Assembly a)
...
 		}
 
+		protected string ResolveLibrarySymlinks (string path)
+		{
+			var resolved = ((NSString) path).ResolveSymlinksInPath ().ToString ();
+			// ResolveSymlinksInPath will return the input if something goes wrong.
+			// Something usually goes wrong with system libraries: they don't actually exist on disk :/
+			// So add some custom logic to handle those cases.
+			resolved = resolved.Replace ("/Versions/A/", "/");
+			resolved = resolved.Replace ("/Versions/C/", "/");
```
spouliot, line 275: that implies there's a `B` version too :)

spouliot (approved):
👍

dalexsoto (approved):

src/CoreFoundation/CFStream.cs
```diff
@@ -239,7 +239,7 @@ public abstract class CFStream : CFType, INativeObject, IDisposable {
 		[Obsolete ("Starting with macos12.0 use 'Network.framework' instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
 #endif
 #endif
-		[DllImport (Constants.CFNetworkLibrary)]
+		[DllImport (Constants.CoreFoundationLibrary)]
```
mandel-macaque, line 242: This has been like that for about 6 years... 😵‍💫

tests/introspection/ApiPInvokeTest.cs
```diff
@@ -238,15 +238,109 @@ protected void Check (Assembly a)
...
+						// Verify that the P/Invoke points to the right library.
+						Dl_info info = default (Dl_info);
+						var found = dladdr (h, out info);
+						if (found != 0) {
+							// Resolve symlinks in both cases
+							var dllImportPath = ResolveLibrarySymlinks (path);
+							var foundLibrary = ResolveLibrarySymlinks (Marshal.PtrToStringAuto (info.dli_fname));
+							if (Skip (name, ref dllImportPath, ref foundLibrary)) {
+								// Skipped
+							} else if (foundLibrary != dllImportPath) {
```
mandel-macaque, line 253: Why not merging the if statement:
```suggestion
							if (!Skip (name, ref dllImportPath, ref foundLibrary) && (foundLibrary != dllImportPath) {
```

mandel-macaque (approved):

tests/introspection/ApiPInvokeTest.cs
rolfbjarne: Easier to add CWLs for debugging (I also find it easier to read).

rolfbjarne:

PR #12722, opened 2021-09-13
[TVUIKit] Update bindings to Xcode 13 Beta 5

dalexsoto:


src/tvuikit.cs
```diff
@@ -290,4 +300,166 @@ interface TVCollectionViewFullScreenLayout {
...
 	}
+
+	[TV (15,0), NoWatch, NoMac, NoiOS]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface TVMediaItemContentConfiguration : UIContentConfiguration, NSSecureCoding {
+
+		[Static]
+		[Export ("wideCellConfiguration")]
+		TVMediaItemContentConfiguration CreateWideCell ();
```
mandel-macaque, line 311: Mixed feelings about adding Configuration at the end.

mandel-macaque (approved):

src/tvuikit.cs
dalexsoto: Yeah I thought so but the return type is a configuration object and to add to it you have to type the whole type name because it is a static method

src/tvuikit.cs
```diff
@@ -290,4 +300,166 @@ interface TVCollectionViewFullScreenLayout {
...
+	[TV (15,0), NoWatch, NoMac, NoiOS]
+	[BaseType (typeof (UIView))]
+	[DisableDefaultCtor]
+	interface TVMediaItemContentView : UIContentView {
+
+		[Export ("initWithConfiguration:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (TVMediaItemContentConfiguration configuration);
+
+		// Conflicts with IUIContentView.Configuration property unfortunately
```
spouliot, line 350: hmm...

`UIContentConfiguration` is a protocol and `TVMediaItemContentConfiguration` conforms to it. So the `IUIContentView.Configuration` property is usable...

The setter is usable without any issue. However the getter is not optimal but...

It might be better, and more general, to add an helper/extension method like `GetConfiguration<T> ()` that would (internally) call the getter and return an instance of the request subclass of `IUIContentConfiguration`

src/tvuikit.cs
spouliot: `Configuration` suffix or `CreateForWideCell` ?
it's definitively not a wide cell that comes out :)

src/tvuikit.cs
```diff
@@ -290,4 +300,166 @@ interface TVCollectionViewFullScreenLayout {
...
+	}
+
+	[TV (15,0)]
+	[BaseType (typeof (NSObject))]
+	[DisableDefaultCtor]
+	interface TVMonogramContentConfiguration : UIContentConfiguration, NSSecureCoding {
+
+		[Static]
+		[Export ("cellConfiguration")]
+		TVMonogramContentConfiguration CreateCell ();
```
spouliot, line 411: same as the wide :)

src/tvuikit.cs
```diff
@@ -290,4 +300,166 @@ interface TVCollectionViewFullScreenLayout {
...
+	[TV (15,0)]
+	[BaseType (typeof (UIView))]
+	[DisableDefaultCtor]
+	interface TVMonogramContentView : UIContentView {
+
+		[Export ("initWithConfiguration:")]
+		[DesignatedInitializer]
+		IntPtr Constructor (TVMonogramContentConfiguration configuration);
+
+		// Conflicts with IUIContentView.Configuration property unfortunately
```
spouliot, line 441: same as the other one too

spouliot (approved):

src/tvuikit.cs
dalexsoto: Right, but is it really worthy creating the extension ‘ GetConfiguration<T> ()’ if you only have two places where it is used? I think by  just keeping the sealed GetConfiguration() on each subclass is a bit better since it avoids user guess work and any generics involvement

spouliot: I think so as it can be done, once, as an extension method on the interface (for the protocol) without requiring each type to add new API (hopefully with a consistent name over time) and with additional `[Export]` (that the registrar has to process).

rolfbjarne (approved):

tj-devel709 (approved):

PR #12739, opened 2021-09-15
[MMP] If we disable mac do not try to build for dotnet.

mandel-macaque:


dalexsoto (approved):

spouliot (commented):
not sure...
disabling XM means we can't build mlaunch

mandel-macaque:
@spouliot this is just if we disable it for those case in which we do not build XM due to a Xcode rollback. We did that last time with 12.5. We can build mlaunch in main.

PR #12742, opened 2021-09-15
[Xtro] Allow to run xtro with some platforms ignored.

mandel-macaque:
Do not add dependencies to targets that have been disabled via ifdef.

### DIFF
--- a/tests/xtro-sharpie/Makefile
+++ b/tests/xtro-sharpie/Makefile
@@ -138,6 +138,8 @@ report-short:
 	JENKINS_SERVER_COOKIE=1 make report
 
 ifdef INCLUDE_IOS
+PCH_FILES+=$(XIOS_PCH)
+
 classify-ios:
 	$(MONO) bin/Debug/xtro-sharpie.exe $(XIOS_PCH) $(XIOS) $(XIOS_GL)
 else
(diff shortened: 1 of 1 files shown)


spouliot (approved):

dalexsoto (approved):

tests/xtro-sharpie/Makefile
```diff
@@ -172,7 +172,7 @@ else
 classify-maccatalyst: ; @true
 endif
 
-classify: build $(XIOS_PCH) $(XWATCHOS_PCH) $(XTVOS_PCH) $(XMAC_PCH) $(XCAT_PCH)
+classify: build $(if $(INCLUDE_IOS), $(XIOS_PCH)) $(if $(INCLUDE_TVOS), $(XTVOS_PCH)) $(if $(INCLUDE_WATCH), $(XWATCHOS_PCH)) $(if $(INCLUDE_MAC), $(XMAC_PCH)) $(if $(INCLUDE_MACCATALYST), $(XCAT_PCH))
```
rolfbjarne, line 175: That's kind of unwieldy to read.

Maybe do something like this in each "ifdef INCLUDE_*" above:

```make
ifdef INCLUDE_IOS
PCH_FILES+=$(XIOS_PCH)
...
endif

ifdef INCLUDE_TVOS
PCH_FILES+=$(XTVOS_PCH)
...
endif

...
```

and then this becomes:

```suggestion
classify: build $(PCH_FILES)
```

rolfbjarne (changes requested):

rolfbjarne (approved):

tj-devel709 (approved):

PR #12760, opened 2021-09-16
[Xcode13] Bump to xcode 13 rc.

mandel-macaque:


tools/devops/automation/templates/build/build.yml
```diff
@@ -288,11 +288,6 @@ steps:
       CONFIGURE_FLAGS="--enable-xamarin"
     fi
 
-    if [[ "$EnableDotNet" == "True" ]]; then
```
mandel-macaque, line 291: This will need to be reverted like the other changes. not needed for the bump, but why build it if this branch is just for classic?

tests/xtro-sharpie/common-AppKit.ignore
```diff
@@ -1,857 +1,16 @@
 ## NSUserInterfaceItemIdentifier -> NS_TYPED_EXTENSIBLE_ENUM
```
rolfbjarne, line 1: Why are changes to `common-AppKit.ignore` needed if we're excluding macOS and Mac Catalyst? This will just complicate the merge back to main later.

rolfbjarne (changes requested):

tests/xtro-sharpie/common-AppKit.ignore
mandel-macaque: Auto sanitize did that. I'll take a look

mandel-macaque: Not updating the ignore will result in the following failures in xtro: https://gist.github.com/mandel-macaque/0e9384d87715fdc0f40604c5fd42cf35

mandel-macaque: The correct fix is to move the ignores, I have created https://github.com/xamarin/xamarin-macios/issues/12769 to track a more appropriate fix.

rolfbjarne (approved):

tj-devel709 (approved):

PR #12763, opened 2021-09-17
[tests] Attempt to work around an issue launching Mac Catalyst tests by using a unique bundle identifier for all tests.

rolfbjarne:


rolfbjarne:
This problem seems to have disappeared once we updated macs to Monterey.

PR #12764, opened 2021-09-17
[dotnet] Remove watchOS code and from our .NET support.

rolfbjarne:
It's not tested, and thus has probably already bitrotted. If we add support
for watchOS to .NET in the future, it would likely be easier to start from
scratch (copying some of the other platforms), than having incomplete and
bitrotted code.


dalexsoto (approved):

rolfbjarne:
Tests executed; what failed above was the publishing of the test results (they're still available for download from Azure DevOps).

There's a test failure, but it's unrelated (https://github.com/xamarin/maccore/issues/2414).

PR #12771, opened 2021-09-17
[Intents] Update bindings for Xcode 13.0 betas 1 through 5

rachelkang:
build and intro and xtro and I are all happy :)


src/Intents/INPerson.cs
```diff
@@ -0,0 +1,67 @@
...
+
+#if !NET
+	[Introduced (PlatformName.iOS, 15,0)]
+	[Introduced (PlatformName.MacOSX, 12,0)]
+	[Introduced (PlatformName.WatchOS, 8,0)]
+#else
+	[SupportedOSPlatform ("ios15.0")]
+	[SupportedOSPlatform ("macos12.0")]
+#endif //!NET
+        public enum INPersonType {
```
rolfbjarne, line 24: The indentation is off here, and it looks like the indentation is spaces too (when it should be tabs).

src/intents.cs
```diff
@@ -6022,23 +6086,19 @@ interface INSearchForPhotosIntentResponse {
 		NSNumber SearchResultsCount { get; set; }
 	}
 
-#if XAMCORE_4_0
-	[NoMac]
-#elif MONOMAC
-	[Obsoleted (PlatformName.MacOSX, 10,0, message: "Unavailable on macOS, will be removed in the future.")]
-#endif
 	[iOS (10, 0)]
```
rolfbjarne, line 6089: Shouldn't this type also get a `[Mac (12,0)]`?

src/intents.cs
```diff
@@ -7363,9 +7514,12 @@ interface INStringResolutionResult {
 		INStringResolutionResult GetConfirmationRequired (NSObject itemToConfirm, nint reason);
 	}
 
+ 	[Deprecated (PlatformName.iOS, 15, 0)]
+ 	[Deprecated (PlatformName.MacCatalyst, 15, 0)]
 	[iOS (10, 0)]
-	[Watch (3, 2)]
-	[NoTV, Mac (11,0)]
+	[NoWatch] // it says it's now NoWatch, but not deprecated? (there was previous watch availability)
```
rolfbjarne, line 7520: This is a breaking change, if the type was already present in an earlier version of Xamarin.iOS, we can't remove it now (we have to mark it as deprecated instead).

src/intents.cs
```diff
@@ -7595,10 +7749,20 @@ interface INWorkoutLocationTypeResolutionResult {
 	[BaseType (typeof (NSUserActivity))]
 	interface NSUserActivity_IntentsAdditions {
 
+		[Mac (12,0)]
```
rolfbjarne, line 7752: This is strange: this method already existed, but Apple marked it as available only in macOS 12.0 now?

rolfbjarne (changes requested):

src/Intents/INPerson.cs
rachelkang: ahh thanks for catching!

src/intents.cs
rachelkang: makes sense, thanks!

src/intents.cs
rachelkang: yep - this is the case for most of the updates in this PR. 

```diff
-@property (readonly, nullable, NS_NONATOMIC_IOSONLY) INInteraction *interaction API_AVAILABLE(ios(10.0), watchos(3.2), tvos(14.0)) API_UNAVAILABLE(macos);
+@property (readonly, nullable, NS_NONATOMIC_IOSONLY) INInteraction *interaction API_AVAILABLE(ios(10.0), macos(12.0), watchos(3.2), tvos(14.0));
```

src/intents.cs
rachelkang: yes, but there are some deprecated types that had earlier availability and deprecation in xamarin.ios. I wasn't sure if I could just delete those entirely, but assumed it might be a breaking change and that it would be better to just decorate everything else with Mac (12,0). Is there an alternative, better solution though?

src/intents.cs
rolfbjarne: Ah, I see, it was marked as unavailable in Apple's headers, but we never reflected that in our bindings (so our attributes were incorrect, and now they aren't anymore).

src/intents.cs
rolfbjarne: I'm not sure I understand. In all the other cases where you removed a similar block of availability attributes for macOS ([NoMac] in XAMCORE_4_0 + [Obsolete] in MONOMAC), you added a `[Mac (12, 0)]`. That's correct. However, this particular type didn't get a `[Mac (12, 0)]` - but you still removed the previous availability attributes. That effectively means that you're saying that this type was introduced in macOS in the first macOS version we support (macOS 10.9), and I don't think that's what you wanted :smile:

src/Intents/INPerson.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Intents
+{
+
+#if !TVOS
+	public partial class INPerson
+	{
+
+#if !NET
+	[Introduced (PlatformName.iOS, 15,0)]
```
rolfbjarne, line 17: The indentation is still off for all the attributes in this file (looks like they should be indented one more level).

rolfbjarne (changes requested):

src/intents.cs
rachelkang: ahh ok I misunderstood, but I see what you mean now!

src/Intents/INPerson.cs
```diff
@@ -0,0 +1,67 @@
...
+	[Introduced (PlatformName.MacOSX, 12,0)]
+	[Introduced (PlatformName.WatchOS, 8,0)]
+#else
+	[SupportedOSPlatform ("ios15.0")]
+	[SupportedOSPlatform ("macos12.0")]
+	#endif //!NET
```
tj-devel709, line 23: Looks like extra indentation here

src/intents.cs
```diff
@@ -12975,12 +13226,12 @@ interface INStartCallContactResolutionResult {
...
 
-	[Watch (6,0), NoTV, NoMac, iOS (13,0)]
+	[Watch (6,0), NoTV, Mac (12,0), iOS (13,0)]
 	[BaseType (typeof (INIntent))]
 	[DisableDefaultCtor]
-	interface INStartCallIntent {
-
-		[Watch (7,0), NoTV, NoMac, iOS (14,0)]
+	interface INStartCallIntent : UNNotificationContentProviding {
+
```
tj-devel709, line 13233: extra whitespace

tj-devel709:
I know Sebastien explained a little to me why, but I think I am still a little confused about something. It makes sense to me to turn a 'NoMac' to a 'Mac 12,0' but why are we changing the already available Mac attributes (ie. Mac 11.3) to 'Mac 12,0'?

rachelkang:
> I know Sebastien explained a little to me why, but I think I am still a little confused about something. It makes sense to me to turn a 'NoMac' to a 'Mac 12,0' but why are we changing the already available Mac attributes (ie. Mac 11,3) to 'Mac 12,0' inside src/intents.cs?

@tj-devel709 yeah, it's confusing! there are slightly different justifications based on the scenario, but for the most part, what happened was that the interface was originally NoMac and some properties inside of them appeared to have some earlier Mac availability. But if I understood Sebastien correctly, this was incorrect to begin with because "it should not exist in Xamarin.Mac if the type has [NoMac]"

tj-devel709:
Ah I see now! Thanks for clearing that up for me, @rachelkang!

rolfbjarne (approved):

src/intents.cs
```diff
@@ -6862,6 +6955,55 @@ interface INSetSeatSettingsInCarIntentResponse {
 		INSetSeatSettingsInCarIntentResponseCode Code { get; }
 	}
 
+	[Watch (8,0), NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(INIntent))]
```
dalexsoto, line 6959: Probably you want a `DisableDefaultCtor` here because there is no way to set `FocusStatus` unless the provided ctor is used

src/intents.cs
```diff
@@ -11804,6 +12009,49 @@ interface INFlightReservation : NSCopying, NSSecureCoding {
...
+	}
+
+	[Watch (8,0), NoTV, Mac (12,0), iOS (15,0), MacCatalyst (15,0)]
+	[BaseType (typeof(NSObject))]
+	[DisableDefaultCtor]
+	interface INFocusStatus : NSCopying, NSSecureCoding
+	{
+		[Export ("initWithIsFocused:")]
+		[DesignatedInitializer]
+		IntPtr Constructor ([NullAllowed] NSNumber isFocused);
```
dalexsoto, line 12029: You can use [BindAs](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#bindasattribute) here, see swift [docs](https://developer.apple.com/documentation/sirikit/infocusstatus/3821590-init)

```suggestion
		IntPtr Constructor ([NullAllowed] [BindAs (typeof (bool?))] NSNumber isFocused);
```

src/intents.cs
```diff
@@ -11804,6 +12009,49 @@ interface INFlightReservation : NSCopying, NSSecureCoding {
...
+
+		[Export ("isFocused", ArgumentSemantic.Copy)]
```
dalexsoto, line 12031: Same, you can use [BindAs](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/binding/binding-types-reference#bindasattribute) to provide a better binding here
```suggestion
		[Export ("isFocused", ArgumentSemantic.Copy)]
		[BindAs (typeof (bool?))]
```

dalexsoto (approved):
Awesome! Only minor things

dalexsoto:
Unrelated test failure 
* https://github.com/xamarin/maccore/issues/2414

PR #12776, opened 2021-09-20
[tests] Fix packaging of Xamarin.Mac tests.

rolfbjarne:
I recently deleted the generated makefile support for building and running our
test suites. It turned out that it was used for building the packaged
Xamarin.Mac tests, so it wasn't as unused as I thought.

So fix the building and packaging of Xamarin.Mac tests to not use the
(non-existent) makefile targets, but instead replicate it with manual make
code.

Also take the opportunity to add packaging and execution of the .NET versions
of these test suites we execute on other macOS versions (both for macOS and
the Mac Catalyst).


chamons (approved):

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2443).

PR #12791, opened 2021-09-21
[dotnet-linker] Check if both static and dynamic libraries from the Mono runtime matches P/Invoke libraries.

rolfbjarne:
When checking if we need to manually preserve a native symbol, we need to
check if a P/Invoke matches both static and dynamic libraries from the Mono
runtime, not just dynamic libraries.


mauroa (approved):

rolfbjarne:

PR #12831, opened 2021-09-24
[msbuild] Sign Xamarin.iOS.Tasks.dll and Xamarin.Mac.Tasks.dll. Fixes #9835.

### LINKED ISSUE #9835: [.NET 6] Change the identify of the Xamarin.iOS.Tasks.dll and Xamarin.Localization.MSBuild.dll so that they're not duplicated across packages.
We currently ship Xamarin.iOS.Tasks.dll and Xamarin.Localization.MSBuild.dll in three/four packages: Microsoft.[iOS|tvOS|watchOS|macOS].Sdk. This is problematic if the user uses different versions of these packages in the same solution, because only one of them will be loaded in the MSBuild process, and there might be incompatibilities.

This is a very similar issue as #9830, just for .NET instead of old-style Xamarin.iOS/Mac.

It's the same reason we use ILRepack to merge all the dependent assemblies into Xamarin.[iOS|Mac].Tasks.dll.

Potential solutions:

* Rename the assemblies according to the platform (say Xamarin.tvOS.Tasks.dll / Xamarin.tvOS.Localization.MSBuild.dll for instance).
* Sign each version with different keys? Not sure if this would work, but it might be the simplest.

---

Time estimate: 3 days.

rolfbjarne:
We need to strongname our MSBuild assemblies, so that different versions
can be loaded side-by-side (one example being having both a legacy and a
.NET project in the same solution).

This required setting a version for Xamarin.iOS.Tasks.dll and
Xamarin.Mac.Tasks.dll, otherwise strong-naming won't work properly (all
versions of an assembly would have the same identity).

Also sign the corresponding test assemblies, since they poke into the
internals of the task assemblies.

Fixes https://github.com/xamarin/xamarin-macios/issues/9835.


chamons (approved):
I have a foreboding feeling that this will come back again, as every time I've dealt with strong naming there was some issue.

PR looks great though.

dalexsoto (approved):

msbuild/Xamarin.Mac.Tasks/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+[assembly: AssemblyCompanyAttribute ("Microsoft Corp.")]
+[assembly: AssemblyFileVersionAttribute (VersionConstants.XamarinMacVersion)]
+[assembly: AssemblyInformationalVersionAttribute (VersionConstants.XamarinMacVersion + "." + VersionConstants.NuGetPrereleaseIdentifier + "+" + VersionConstants.NuGetBuildMetadata)]
```
mandel-macaque, line 7: I would have prefer string interpolation ;)

mandel-macaque (approved):

rolfbjarne:

mauroa:
Should we add the same for Xamarin.iOS.Windows.Tasks.csproj? It contains the HotRestart and Windows Tasks/Targets

mauroa (commented):
Should we add the same for Xamarin.iOS.Windows.Tasks.csproj? It contains the HotRestart and Windows Tasks/Targets

rolfbjarne:
> Should we add the same for Xamarin.iOS.Windows.Tasks.csproj? It contains the HotRestart and Windows Tasks/Targets

Yes, I think it would be a good idea. I think that can be done in a different PR though.

mauroa (approved):

PR #12832, opened 2021-09-24
[dotnet] Show an error if we must link, but linking has been disabled. Fixes #12372.

### LINKED ISSUE #12372: .NET: we don't support PublishTrimmed=false, so show an error or warning in that case
### Steps to Reproduce

```
dotnet build /p:PublishTrimmed=false
```

This doesn't produce a valid app bundle, so we should show a warning or error asking users to turn trimming on.

rolfbjarne:
For executable projects, we must run the linker (otherwise we won't produce
something that can be executed).

We'll enable the linker by default in these cases, but if the developer has
manually disabled the linker (if the linker fails to execute for any reason,
it will suggest that the developer disables it), then we should show an error
explaining why.

Fixes https://github.com/xamarin/xamarin-macios/issues/12372.


mandel-macaque (approved):

dalexsoto (approved):

tj-devel709 (approved):

rolfbjarne:

PR #12837, opened 2021-09-27
[Makefile] Remove warning for duplicate files.

mandel-macaque:
csc does not care about the order of the files, yet it does care about
duplicate files (warning is printed). One of the side effects of sort
implementation in gnu make is that it removes the duplicates and
therefore the warning about some files being included twice.


dalexsoto (approved):

rolfbjarne (approved):

mandel-macaque:
Un related failure: https://github.com/xamarin/maccore/issues/2519

PR #12861, opened 2021-09-28
[Metal] Fix a few issues in MTLDevice.

rolfbjarne:
* MTLCopyAllDevices returns a retained object, so we need to release it.
* MTLCopyAllDevicesWithObserver returns an observer (no need to provide one). This
  means we can obsolete the 'ref NSObject' overload (the API doesn't make sense),
  and instead add an 'out NSObject' overload.
* The returned observer from MTLCopyAllDevicesWithObserver is retained, so we must
  release it.
* The returned array from MTLCopyAllDevicesWithObserver is a retained object, so
  we need to release it.
* Simpify the supporting block code for the calls to MTLCopyAllDevicesWithObserver.
* Clean up the block we passed to MTLCopyAllDevicesWithObserver.

### DIFF
--- a/src/Metal/MTLDevice.cs
+++ b/src/Metal/MTLDevice.cs
@@ -61,42 +61,48 @@ namespace Metal {
 		public static IMTLDevice [] GetAllDevices ()
 		{
 			var rv = MTLCopyAllDevices ();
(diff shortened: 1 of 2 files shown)


mandel-macaque (approved):
We have a PR race! 🚗

PR #12886, opened 2021-09-30
[ApiDiffs] Creating Api Diffs for the Dotnet Assemblies

tj-devel709:
This PR creates the Api Diffs for the Dotnet Assemblies for three different scenarios:
* New Dotnet assemblies vs. Stable Dotnet assemblies
* New Dotnet assemblies vs. Stable Legacy assemblies
* New Dotnet iOS assembly vs. New Dotnet MacCatalyst assembly

Task is found here: https://github.com/xamarin/xamarin-macios/issues/10210

Next part of this is to include these new diffs into the CI!

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -41,6 +41,7 @@ APIDIFF_REFERENCES_DOTNET_iOS=https://bosstoragemirror.blob.core.windows.net/wre
 APIDIFF_REFERENCES_DOTNET_tvOS=https://bosstoragemirror.blob.core.windows.net/wrench/main/3c2c3d062daedd66ccd06ecb68192767ddf8e3c4/5315390/package/bundle.zip
 APIDIFF_REFERENCES_DOTNET_macOS=https://bosstoragemirror.blob.core.windows.net/wrench/main/3c2c3d062daedd66ccd06ecb68192767ddf8e3c4/5315390/package/bundle.zip
 APIDIFF_REFERENCES_DOTNET_MacCatalyst=https://bosstoragemirror.blob.core.windows.net/wrench/main/3c2c3d062daedd66ccd06ecb68192767ddf8e3c4/5315390/package/bundle.zip
+
 PACKAGE_HEAD_REV=$(shell git rev-parse HEAD)
 
 #
--- a/tools/apidiff/Makefile
+++ b/tools/apidiff/Makefile
@@ -603,3 +603,4 @@ jenkins-api-diff:
 	$(Q) mkdir -p $(JENKINS_RESULTS_DIRECTORY)/api-diff
 	$(Q) $(CP) *.html $(JENKINS_RESULTS_DIRECTORY)/api-diff
 	$(Q) $(CP) *.md $(JENKINS_RESULTS_DIRECTORY)/api-diff
+	$(Q) $(CP) -R diff $(JENKINS_RESULTS_DIRECTORY)/api-diff


tj-devel709:
Here is a zip of the Apidiff directory. You can download the zip and view the `index.html` file to see the new diffs and view the html and md files

tools/apidiff/Makefile
```diff
@@ -44,6 +44,8 @@ MAC_ASSEMBLIES     = $(foreach file,$(MONO_ASSEMBLIES),Xamarin.Mac/$(file))
 WATCHOS_ASSEMBLIES = $(foreach file,$(filter-out Mono.Data.Tds Mono.Security,$(MONO_ASSEMBLIES)),Xamarin.WatchOS/$(file)) $(WATCHOS_SRC_ASSEMBLIES)
 TVOS_ASSEMBLIES    = $(foreach file,$(MONO_ASSEMBLIES),Xamarin.TVOS/$(file))    $(TVOS_SRC_ASSEMBLIES)
 MACCATALYST_ASSEMBLIES = $(foreach file,$(MONO_ASSEMBLIES),Xamarin.MacCatalyst/$(file))    $(MACCATALYST_SRC_ASSEMBLIES)
+DOTNET_ASSEMBLIES = Microsoft.iOS.Ref/ref/net6.0/Xamarin.iOS Microsoft.tvOS.Ref/ref/net6.0/Xamarin.TVOS Microsoft.MacCatalyst.Ref/ref/net6.0/Xamarin.MacCatalyst Microsoft.macOS.Ref/ref/net6.0/Xamarin.Mac Microsoft.MacCatalyst.Ref/ref/net6.0/Xamarin.iOS
```
mandel-macaque, line 47: Can we have this in more than one line?

mandel-macaque (approved):
Small comment.

Make.config
```diff
@@ -36,8 +36,7 @@ $(TOP)/Make.config.inc: $(TOP)/Make.config $(TOP)/mk/mono.mk
 
 include $(TOP)/Make.versions
 
-APIDIFF_REFERENCES=https://bosstoragemirror.blob.core.windows.net/wrench/xcode12.5/ab40b131d3e87a96c3414447dcded5fec45bce36/4679032/package/bundle.zip
-
+APIDIFF_REFERENCES=https://bosstoragemirror.blob.core.windows.net/wrench/main/cd2867d44caa9485701b6b63bb6c38f771a767c9/5267313/package/bundle.zip
```
rolfbjarne, line 39: Unfortunately this part becomes a bit complicated.

We can't do this, because it will make the api diffs for legacy Xamarin wrong (in fact this should be bumped to the apidiff reference from the `xcode13-ios` branch now).

However, we can't bump to the bundle.zip from the `xcode13-ios` branch, because it won't contain binaries for Xamarin.Mac (nor the .NET assemblies for that matter).

We really need a different bundle.zip for every platform, but for legacy Xamarin that's out of scope for this PR.

But what is possible is a middle ground: use a different url for .NET assemblies, and even better would be to make it possible to use a different url for each platform, so something like this:

```make
APIDIFF_REFERENCES_DOTNET_iOS=...
APIDIFF_REFERENCES_DOTNET_tvOS=...
APIDIFF_REFERENCES_DOTNET_macOS=...
APIDIFF_REFERENCES_DOTNET_MacCatalyst=...
```

because it's entirely possible we'll end up in the same situation again at some point (different stable versions of iOS and macOS), and this way we'd have a solution since we could have different reference bundles for different platforms.

rolfbjarne (changes requested):
This looks very good so far!

tj-devel709:
@rolfbjarne I will be pausing work on this PR to work on creating multiple bundle links to use for the references!

rolfbjarne:
This looks much better now!

PR #12921, opened 2021-10-04
[tests] Modify a few test directories with multiple projects to have the projects in subfolders.

rolfbjarne:
This fixes a problem where the build from the different projects could stomp
on eachother in the obj/bin folders. It's technically possible to make this
work by setting up custom [Base][Intermediate][OutputPath] properties in the
project files, but the simplest solution is to just make sure there's no more
than a single project per directory.


dalexsoto (approved):

mandel-macaque (approved):

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2414)

PR #12936, opened 2021-10-05
[AppKit] Update Xcode 13.0 bindings for betas 1,2,3,4,5

rachelkang:


src/appkit.cs
```diff
@@ -13069,6 +13143,30 @@ partial interface NSScreen {
 		string LocalizedName { get; }
 	}
 
+	[NoMacCatalyst]
+	partial interface NSScreen
```
rachelkang, line 13147: Was unsure about what to do about this interface structure - there are a couple other partial interfaces with the same name, in the same file here. Not sure why they're all separated, but sharpie does the same, and diffs suggest the interface is new for some reason. So, I just did what the diffs/sharpie seem to indicate.

But I was wondering - why is it like that? Can't I just add these new properties to the partial interface above?

src/appkit.cs
```diff
@@ -1040,6 +1053,30 @@ interface NSApplicationDelegate {
 		[Deprecated (PlatformName.MacOSX, 11, 0, message: "Now optional on NSApplicationDelegate.")]
 		[Export ("application:delegateHandlesKey:"), DelegateName ("NSApplicationHandlesKey"), NoDefaultValue]
 		bool HandlesKey (NSApplication sender, string key);
+
+		[Mac (12,0)]
+		[IgnoredInDelegate]
+		[Export ("applicationSupportsSecureRestorableState:")]
+		bool ApplicationSupportsSecureRestorableState (NSApplication application);
```
rolfbjarne, line 1060: The 'application' suffix is often emitted from the managed name in this case, because it's somewhat redundant (look at the other methods in this class, they mostly do the same thing) - in the UIApplicationDelegate class the logic is about UIApplication, which means that spelling out 'Application' all the time feels a bit unnecessary:

```suggestion
		bool SupportsSecureRestorableState (NSApplication application);
```

src/appkit.cs
```diff
@@ -1040,6 +1053,30 @@ interface NSApplicationDelegate {
...
+
+		[Mac (12,0)]
+		[IgnoredInDelegate]
+		[Export ("application:handlerForIntent:")]
+		[return: NullAllowed]
+		NSObject GetApplication (NSApplication application, INIntent intent);
```
rolfbjarne, line 1066: This is somewhat of a similar case, the important part is what comes after 'application:', so this is a better managed name:

```suggestion
		NSObject GetHandler (NSApplication application, INIntent intent);
```

src/appkit.cs
```diff
@@ -1040,6 +1053,30 @@ interface NSApplicationDelegate {
...
+
+		[Mac (12,0)]
+		[IgnoredInDelegate]
+		[Export ("applicationShouldAutomaticallyLocalizeKeyEquivalents:")]
+		bool ApplicationShouldAutomaticallyLocalizeKeyEquivalents (NSApplication application);
```
rolfbjarne, line 1071: Same:

```suggestion
		bool ShouldAutomaticallyLocalizeKeyEquivalents (NSApplication application);
```

src/appkit.cs
```diff
@@ -1040,6 +1053,30 @@ interface NSApplicationDelegate {
...
+
+		[Mac (12,0)]
+		[Export ("applicationProtectedDataWillBecomeUnavailable:")]
+		void ApplicationProtectedDataWillBecomeUnavailable (NSNotification notification);
```
rolfbjarne, line 1075:
```suggestion
		void ProtectedDataWillBecomeUnavailable (NSNotification notification);
```

src/appkit.cs
```diff
@@ -1040,6 +1053,30 @@ interface NSApplicationDelegate {
...
+
+		[Mac (12,0)]
+		[Export ("applicationProtectedDataDidBecomeAvailable:")]
+		void ApplicationProtectedDataDidBecomeAvailable (NSNotification notification);
```
rolfbjarne, line 1079:
```suggestion
		void ProtectedDataDidBecomeAvailable (NSNotification notification);
```

src/appkit.cs
```diff
@@ -1565,6 +1602,10 @@ partial interface NSBitmapImageRep : NSSecureCoding {
 		[Field ("NSImageEXIFData")]
 		NSString EXIFData { get; }
 
+		[Mac (12,0)]
+		[Field ("NSImageIPTCData")]
+		NSString IPTCData { get; }
```
rolfbjarne, line 1607: [IPTC](https://en.wikipedia.org/wiki/IPTC_Information_Interchange_Model) is an acronym that's longer that 2 letters, so:

```suggestion
		NSString IptcData { get; }
```

src/appkit.cs
```diff
@@ -5748,6 +5799,11 @@ partial interface NSDocument /* : NSUserActivityRestoring radar://42781537 */ {
 		[Export ("restorableStateKeyPaths", ArgumentSemantic.Copy)]
 		string [] RestorableStateKeyPaths ();
 
+		[Mac (12,0)]
+		[Static]
+		[Export ("allowedClassesForRestorableStateKeyPath:")]
+		Class[] AllowedClassesForRestorableStateKeyPath (string keyPath);
```
rolfbjarne, line 5805: Methods should have a verb:

```suggestion
		Class[] GetAllowedClassesForRestorableStateKeyPath (string keyPath);
```

src/appkit.cs
```diff
@@ -12623,6 +12692,11 @@ partial interface NSResponder : NSCoding, NSTouchBarProvider /* radar://42781537
...
```
rolfbjarne, line 12698: Same:

```suggestion
		Class[] GetAllowedClassesForRestorableStateKeyPath (string keyPath);
```

src/appkit.cs
rolfbjarne: I don't know why sharpie does that, but you can just add it to any existing partial interfaces (at the very least it makes it less confusing for the next person).

src/appkit.cs
```diff
@@ -20196,9 +20326,10 @@ partial interface NSWindow : NSAnimatablePropertyContainer, NSUserInterfaceItemI
 	
 		[Export ("resizeFlags")]
 		nint ResizeFlags { get; }
-	
-		[Export ("keyDown:")]
-		void KeyDown (NSEvent  theEvent);
+
+		// Inherits from NSResponder
+		// [Export ("keyDown:")]
```
rolfbjarne, line 20331: Was this method just removed? If so, it's a breaking change

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
 		
 		[Export ("URLForApplicationToOpenURL:"), ThreadSafe]
 		NSUrl UrlForApplication (NSUrl url );
+
+		[Mac (12,0)]
+		[Export ("URLsForApplicationsWithBundleIdentifier:")]
+		NSUrl[] URLsForApplications (string bundleIdentifier);
```
rolfbjarne, line 21549: Verb + casing for acronym:

```suggestion
		NSUrl[] GetUrlsForApplications (string bundleIdentifier);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Mac (12,0)]
+		[Export ("URLsForApplicationsToOpenURL:")]
+		NSUrl[] URLsForApplications (NSUrl url);
```
rolfbjarne, line 21553:
```suggestion
		NSUrl[] GetUrlsForApplications (NSUrl url);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenContentTypeOfFileAtURL:completionHandler:")]
+		void SetDefaultApplicationToOpenContentType (NSUrl applicationURL, NSUrl url, [NullAllowed] Action<NSError> completionHandler);
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenURLsWithScheme:completionHandler:")]
+		void SetDefaultApplication (NSUrl applicationURL, string urlScheme, [NullAllowed] Action<NSError> completionHandler);
```
rolfbjarne, line 21563: I think this is better, just to match the pattern of the other ones:

```suggestion
		void SetDefaultApplicationToOpenUrls (NSUrl applicationUrl, string urlScheme, [NullAllowed] Action<NSError> completionHandler);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+		NSUrl[] URLsForApplications (string bundleIdentifier);
+
+		[Mac (12,0)]
+		[Export ("URLsForApplicationsToOpenURL:")]
+		NSUrl[] URLsForApplications (NSUrl url);
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenContentTypeOfFileAtURL:completionHandler:")]
+		void SetDefaultApplicationToOpenContentType (NSUrl applicationURL, NSUrl url, [NullAllowed] Action<NSError> completionHandler);
```
rolfbjarne, line 21558:
```suggestion
		void SetDefaultApplicationToOpenContentType (NSUrl applicationUrl, NSUrl url, [NullAllowed] Action<NSError> completionHandler);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenURLsWithScheme:completionHandler:")]
+		void SetDefaultApplication (NSUrl applicationURL, string urlScheme, [NullAllowed] Action<NSError> completionHandler);
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenFileAtURL:completionHandler:")]
+		void SetDefaultApplicationToOpenFile (NSUrl applicationURL, NSUrl url, [NullAllowed] Action<NSError> completionHandler);
```
rolfbjarne, line 21568:
```suggestion
		void SetDefaultApplicationToOpenFile (NSUrl applicationUrl, NSUrl url, [NullAllowed] Action<NSError> completionHandler);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Mac (12,0)]
+		[Export ("URLForApplicationToOpenContentType:")]
+		[return: NullAllowed]
+		NSUrl URLForApplication (UTType contentType);
```
rolfbjarne, line 21573: I think this have to be a bit more explicit to be understandable:

```suggestion
		NSUrl GetUrlForApplicationToOpenContentType (UTType contentType);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Mac (12,0)]
+		[Export ("URLsForApplicationsToOpenContentType:")]
+		NSUrl[] URLsForApplications (UTType contentType);
```
rolfbjarne, line 21577:
```suggestion
		NSUrl[] GetUrlsForApplicationsToOpenContentType (UTType contentType);
```

src/appkit.cs
```diff
@@ -21412,6 +21543,43 @@ interface NSWorkspace : NSWorkspaceAccessibilityExtensions {
...
+
+		[Async]
+		[Mac (12,0)]
+		[Export ("setDefaultApplicationAtURL:toOpenContentType:completionHandler:")]
+		void SetDefaultApplication (NSUrl applicationURL, UTType contentType, [NullAllowed] Action<NSError> completionHandler);
```
rolfbjarne, line 21582:
```suggestion
		void SetDefaultApplicationToOpenContentType (NSUrl applicationURL, UTType contentType, [NullAllowed] Action<NSError> completionHandler);
```

rolfbjarne (changes requested):
This looks great! I mostly just had a few comments about naming.

src/appkit.cs
rachelkang: sounds good, thanks!

src/appkit.cs
rachelkang: Yes, but the interface inherits from NSResponder, which still has keyDown:, so my understanding was that this was ok. I think @dalexsoto and/or @mandel-macaque suggested I comment this one out like this.

According to the diffs, it was simply removed from NSWindow. And the header files and web docs don't document its existence in NSWindow at all.

rachelkang:
@rolfbjarne thank you! updated based on your feedback

rolfbjarne (approved):

src/appkit.cs
```diff
@@ -12623,6 +12692,11 @@ partial interface NSResponder : NSCoding, NSTouchBarProvider /* radar://42781537
 		[Export ("restorableStateKeyPaths", ArgumentSemantic.Copy)]
 		string [] RestorableStateKeyPaths ();
 
+		[Mac (12,0)]
+		[Static]
+		[Export ("allowedClassesForRestorableStateKeyPath:")]
+		Class[] GetAllowedClassesForRestorableStateKeyPath (string keyPath);
```
tj-devel709, line 12698: Wondering if this name could exclude KeyPath since it is the parameter name.
Swift name in the docs is `allowedClasses(forRestorableStateKeyPath:)`
but would `GetAllowedClassesForRestorableState (string keyPath)` work here?

src/appkit.cs
```diff
@@ -17897,48 +17991,70 @@ interface NSTextDelegate {
...
+	}
 
-		//Detected properties
-		[Export ("attachment")][NullAllowed]
-		NSTextAttachment Attachment { get; set; }
+	[NoMacCatalyst]
+	[BaseType (typeof (NSCell))]
+	interface NSTextAttachmentCell : NSTextAttachmentCellProtocol {
+		[Export ("initImageCell:")]
+		IntPtr Constructor (NSImage  image);
```
tj-devel709, line 18054:
```suggestion
		IntPtr Constructor (NSImage image);
```nit

src/appkit.cs
rachelkang: thanks! I'll just simplify to `GetAllowedClasses` then - I think RestorableStateKeyPath is one thing, so it could be confusing to imply its for the restorablestate and not specify the keypath

PR #12937, opened 2021-10-06
[devops] Return a non-zero exit code if tests fail on older macOS versions.

rolfbjarne:
Return a non-zero exit code if tests fail on older macOS versions, but keep
running tests. This way the step shows up as orange if something fails (and
not green, which is confusing).


tj-devel709 (approved):

chamons (approved):
LGTM - FYI @mandel-macaque

PR #12990, opened 2021-10-14
[NetworkExtension] Updates for Catalyst

rachelkang:
Replaced MacCatalyst (15,0) with NoMacCatalyst for selectors that seem to have been misattributed, as there is no indication that they should have Catalyst availability.


tj-devel709:
Maybe let @mandel-macaque confirm that these were added by accident!

PR #13028, opened 2021-10-19
[Github Actions] Get rid of annoying failure emails

tj-devel709:
When merging something to main from a fork, this github action gets a little salty and results in emails being sent about it failing when we use `exit 1`

image
image

Let's just use `success ()` conditionals to continue running the github action or stop running the github action as needed.


tj-devel709:
Having trouble testing, working on that now

PR #13046, opened 2021-10-20
[Hot Restart] Copy the .NET app files and resources to the bundle before it gets signed

filipnavara:


rolfbjarne:

filipnavara:
It's a bit tricky to come up with the performance numbers for the closed-source code. Just to give a rough idea my own code for resource hashing was running at around 0.6s (+- 0.1s) for 230 MB app bundle and calculating both SHA-1 and SHA-256 simultaneously for each file (ie. avoiding double reads from the disk).

Using different rules for resource coverage turned out to be no-go. Apple has code to detect weak resource rules and fails the strict validation if they are used.

Finally, Apple [seems to be aware](https://github.com/apple-opensource/Security/blob/dccb4e7e9108a88fbc7f2983d5b4d228c1bb761d/OSX/libsecurity_codesigning/lib/SecStaticCode.cpp#L147-L153) of the fact that the resource envelope is not properly checked when the bundle is using embedded provisioning profile. It's not clear whether or when they will plug the hole and start verifying it.

mandel-macaque (approved):

rolfbjarne:
@filipnavara I'm sorry we dropped the ball on this PR, is it still something you'd like to get merged?

filipnavara:
I think there's still merit to the PR but it would need rebasing and retesting. I'm unlikely to do that soon so feel free to close it. 

(Not directly related but I also maintain https://github.com/filipnavara/CodeSign to sign iOS apps. It's licensed as MIT and I even had a wrapper that allowed it to be used directly in HotRestart. If you ever grow tired of maintaining your own signing code and catching up to iOS releases, I would be more than happy to offer assistance to migrate to the OSS code.)

rolfbjarne:
> I think there's still merit to the PR but it would need rebasing and retesting. I'm unlikely to do that soon so feel free to close it.

OK, thanks, and sorry again for dropping the ball here.

> (Not directly related but I also maintain [filipnavara/CodeSign](https://github.com/filipnavara/CodeSign) to sign iOS apps. It's licensed as MIT and I even had a wrapper that allowed it to be used directly in HotRestart. If you ever grow tired of maintaining your own signing code and catching up to iOS releases, I would be more than happy to offer assistance to migrate to the OSS code.)

@emaf @mauroa FYI

PR #13135, opened 2021-10-26
[Network] Implement all bindings up to Xcode 13.1

mandel-macaque:
QUIC methods are not tested because I do not have a server running QUIC and httpbin.org does not have support yet. Also curl does not have support in its stable version but has it in HEAD.

### DIFF
--- a/tests/monotouch-test/Network/NWConnectionGroupTest.cs
+++ b/tests/monotouch-test/Network/NWConnectionGroupTest.cs
@@ -1,3 +1,5 @@
+using System;
+
 using CoreFoundation;
 using Foundation;
 using Network;
--- a/tests/monotouch-test/Network/NWListenerTest.cs
+++ b/tests/monotouch-test/Network/NWListenerTest.cs
@@ -1,4 +1,6 @@
 #if !__WATCHOS__
+using System;
+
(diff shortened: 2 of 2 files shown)


src/Network/NWResolverConfig.cs
```diff
@@ -0,0 +1,47 @@
...
+				throw new ArgumentNullException (nameof (urlEndpoint));
+			switch (endpointType) {
+			case NWResolverConfigEndpointType.Https:
+				InitializeHandle (nw_resolver_config_create_https (urlEndpoint.Handle));
+				break;
+			case NWResolverConfigEndpointType.Tls:
+				InitializeHandle (nw_resolver_config_create_tls (urlEndpoint.Handle));
+				break;
+			default:
+				throw new ArgumentOutOfRangeException ("Unknown endpoint type.");
```
rolfbjarne, line 37:
```suggestion
				throw new ArgumentOutOfRangeException ($"Unknown endpoint type: {endpointType}");
```

rolfbjarne (approved):

dalexsoto (approved):

tj-devel709 (approved):

PR #13151, opened 2021-10-28
[PrintCore] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Subclass NativeObject to reuse object lifetime code.ç
* Remove the PrintCore.NativeInvoke class and put the P/Invokes in the
  PMPrintCoreBase class - no need for a separate class for the P/Invokes in
  this class.

### DIFF
--- a/src/PrintCore/PrintCore.cs
+++ b/src/PrintCore/PrintCore.cs
@@ -6,6 +6,9 @@
 //
 // Copyright 2016 Microsoft Inc
 //
+
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/PrintCore/PrintCore.cs
```diff
@@ -117,50 +98,34 @@ public static PMStatusCode TryCreate (out PMPrintSession session)
...
 				throw new ArgumentNullException (nameof (pageFormat));
-			PMSessionDefaultPageFormat (handle, pageFormat.Handle);
+			PMSessionDefaultPageFormat (Handle, pageFormat.Handle);
 		}
 		
 		[DllImport (Constants.PrintCoreLibrary)]
 		extern static PMStatusCode PMSessionCreatePrinterList (IntPtr printSession, out IntPtr printerListArray, out int index, out IntPtr printer);
-
-		internal static string [] FetchArray (IntPtr cfArrayString, bool owns)
+
```
tj-devel709, line 118: nit: extra tabs

tj-devel709 (approved):

src/PrintCore/PrintCore.cs
rolfbjarne:
```suggestion

```

PR #13152, opened 2021-10-28
[SearchKit] Subclass NativeObject + numerous other code updates

rolfbjarne:


rolfbjarne:
Test failures are unrelated

* introspection/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2494

PR #13157, opened 2021-10-28
[CGImageDestination] Subclass NativeObject + numerous other code updates

rolfbjarne:


rolfbjarne:

PR #13165, opened 2021-10-28
[CFRunLoop] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Make the (IntPtr, bool) constructor internal for .NET
* Make the CFRunLoopSourceCreate P/Invoke take a ref parameter to avoid manual
  memory handling.

### DIFF
--- a/src/CoreFoundation/CFSocket.cs
+++ b/src/CoreFoundation/CFSocket.cs
@@ -326,7 +326,7 @@ namespace CoreFoundation {
 				throw new CFSocketException (CFSocketError.Error);
 			gch = GCHandle.Alloc (this);
 
-			var source = new CFRunLoopSource (CFSocketCreateRunLoopSource (IntPtr.Zero, handle, 0));
+			var source = new CFRunLoopSource (CFSocketCreateRunLoopSource (IntPtr.Zero, handle, 0), true);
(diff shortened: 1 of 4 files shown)


src/CoreFoundation/CFSocket.cs
```diff
@@ -302,8 +302,11 @@ public CFSocket (AddressFamily family, SocketType type, ProtocolType proto, CFRu
 				var handle = CFSocketCreate (IntPtr.Zero, family, type, proto, (nuint) (ulong) cbTypes, OnCallback, ref ctx);
 				InitializeHandle (handle);
 
-			var source = new CFRunLoopSource (CFSocketCreateRunLoopSource (IntPtr.Zero, handle, 0), true);
-			loop.AddSource (source, CFRunLoop.ModeDefault);
+				var source = new CFRunLoopSource (CFSocketCreateRunLoopSource (IntPtr.Zero, handle, 0), true);
+				loop.AddSource (source, CFRunLoop.ModeDefault);
+			} catch {
+				gch.Free ();
```
Therzok, line 308:
```suggestion
				gch.Free ();
				throw;
```

tj-devel709 (approved):

PR #13175, opened 2021-10-28
[src] Remove the legacy build of Xamarin.MacCatalyst.dll

rolfbjarne:
We decided some time ago that Mac Catalyst will only be supported in .NET,
so building it for legacy Xamarin is wasteful.

This should speed up the src/ build somewhat.


mandel-macaque (approved):

tools/apidiff/Makefile
```diff
@@ -283,13 +268,6 @@ ifdef INCLUDE_TVOS
 		echo "<h2>Xamarin.TVOS New Dotnet vs Legacy Stable API diff is empty</h2>" >> $@; \
 	fi;
 endif
-ifdef INCLUDE_MACCATALYST
```
tj-devel709, line 286: This removes the dotnet/legacy html for MacCatalyst, but I think we would probably want to also remove the markdown for this dotnet/legacy MacCatlyst diff and remove the part from the CI that prints these to Github. This would also probably apply for the other removed diffs like maccat-api-diff.html etc. Is this something I should do next?

mandel-macaque: It should be done in the same PR else we will have commits with issue in the CI. Also, reverting to the initial status should be done in a single commit IMO.

rolfbjarne: @tj-devel709 yes, that's a great point, could you do that on my branch and just push your changes to this PR? That way we're doing it all in one PR.

That said, it'll probably be a little while until we can merge this, because we have to implement new xtro logic first.

PR #13180, opened 2021-10-29
[CV[Image|Pixel[Buffer] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Remove the internal (IntPtr) and default constructor.

### DIFF
--- a/src/CoreVideo/CVPixelBuffer.cs
+++ b/src/CoreVideo/CVPixelBuffer.cs
@@ -6,7 +6,6 @@
 // Copyright 2010 Novell, Inc
 // Copyright 2011-2014 Xamarin Inc
 //
-
 using System;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


mandel-macaque (approved):

src/CoreVideo/CVPixelBuffer.cs
```diff
@@ -6,6 +6,7 @@
 // Copyright 2010 Novell, Inc
 // Copyright 2011-2014 Xamarin Inc
 //
+
```
tj-devel709, line 9: Possibly extra newline

tj-devel709 (approved):

src/CoreVideo/CVPixelBuffer.cs
rolfbjarne:
```suggestion
```

PR #13186, opened 2021-10-29
[SecAccessControl] Subclass NativeObject + a few other code updates

rolfbjarne:
* The implementation of SecAccessControl was a bit unconventional: it would
  only create a native object when necessary. This breaks the usual pattern,
  and the original commit doesn't explain anything, so just make this class as
  any other classes (native instance created in the constructor).

### DIFF
--- a/src/Security/SecAccessControl.cs
+++ b/src/Security/SecAccessControl.cs
@@ -12,6 +12,8 @@
 // calling SecAddItem.
 //
 
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using ObjCRuntime;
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

src/Security/SecAccessControl.cs
```diff
@@ -98,56 +100,19 @@ public enum SecAccessControlCreateFlags : long {
...
-		internal SecAccessControl (IntPtr handle)
+		internal SecAccessControl (IntPtr handle, bool owns)
+			: base (handle, owns)
 		{
-			// note: the properties won't match reality
-			Handle = handle;
 		}
 
 		public SecAccessControl (SecAccessible accessible, SecAccessControlCreateFlags flags = SecAccessControlCreateFlags.UserPresence)
+			: base (SecAccessControlCreateWithFlags (IntPtr.Zero, KeysAccessible.FromSecAccessible (accessible), (nint) (long) flags, out var _), true)
```
tj-devel709, line 111: Interesting, is this double type casting? `(nint) (long) flags`

tj-devel709 (approved):

src/Security/SecAccessControl.cs
rolfbjarne: Yes, it's double type casting. `flags` is an enum, and we can't cast directly from an enum to `nint`, so we have to go through `long` first (which is the underlying type of the enum).

PR #13189, opened 2021-10-29
[CVPixelBufferPool] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Remove the internal (IntPtr) constructor.


tj-devel709 (approved):

rolfbjarne:

PR #13212, opened 2021-11-02
[CG[Mutable]ImageMetadata] Subclass NativeObject + numerous other code updates

rolfbjarne:


chamons (approved):

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:

PR #13213, opened 2021-11-02
[CGContextPDF] Enable nullability + a few other code updates

rolfbjarne:


mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne:

PR #13221, opened 2021-11-02
[CTFontCollection] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Use CFArray helper methods to create managed array from native CFArrays.

### DIFF
--- a/src/CoreText/CTFontCollection.cs
+++ b/src/CoreText/CTFontCollection.cs
@@ -26,6 +26,9 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
+
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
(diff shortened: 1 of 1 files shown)


src/CoreText/CTFontCollection.cs
```diff
@@ -162,13 +133,10 @@ public CTFontCollection WithFontDescriptors (CTFontDescriptor[] queryDescriptors
...
 		{
-			var cfArrayRef = CTFontCollectionCreateMatchingFontDescriptors (handle);
+			var cfArrayRef = CTFontCollectionCreateMatchingFontDescriptors (Handle);
 			if (cfArrayRef == IntPtr.Zero)
 				return Array.Empty <CTFontDescriptor> ();
-			var matches = NSArray.ArrayFromHandle (cfArrayRef,
-					fd => new CTFontDescriptor (fd, false));
-			CFObject.CFRelease (cfArrayRef);
-			return matches;
+			return CFArray.ArrayFromHandleFunc (cfArrayRef, fd => new CTFontDescriptor (fd, false), true)!;
```
dalexsoto, line 139: nice reminder of ArrayFromHandleFunc 👍

dalexsoto (approved):

rolfbjarne:

PR #13223, opened 2021-11-02
[CTLine] Subclass NativeObject + numerous other code updates

rolfbjarne:
* Simplify block creation code a bit.

### DIFF
--- a/src/CoreText/CTLine.cs
+++ b/src/CoreText/CTLine.cs
@@ -26,6 +26,9 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
+
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

src/CoreText/CTLine.cs
```diff
@@ -238,37 +204,32 @@ public nfloat GetOffsetForStringIndex (nint charIndex)
...
 		static unsafe void TrampolineEnumerate (IntPtr blockPtr, double offset, nint charIndex, bool leadingEdge, ref bool stop)
 		{
-			var block = (BlockLiteral *) blockPtr;
-                        var del = (CaretEdgeEnumerator) (block->Target);
-                        if (del != null)
-                                del (offset, charIndex, leadingEdge, ref stop);
- 		}
-		
+			var block = (BlockLiteral*) blockPtr;
+			var del = (CaretEdgeEnumerator) (block->Target);
```
mandel-macaque, line 215: Could we have used as and then do it in the if statement?

dalexsoto (approved):

src/CoreText/CTLine.cs
rolfbjarne: I simplified it a bit differently in the end.

PR #13231, opened 2021-11-02
[ObjCRuntime] Add a non-deprecated internal system-version checking API and use it everywhere.

rolfbjarne:
The PlatformHelper class is deprecated, so implement a new version that isn't
deprecated, and which shares a similar API between all platforms - the Check*
methods includes the name of the platform, because that makes it clearer which
version we're talking about from the call site. There's a quirk though:
there's no separate ChecktvOS or CheckMacCatalyst, because the system version
is the same as for iOS, so we can just use 'iOS'.

For macOS we can now use NSProcessInfo.ProcessInfo.OperatingSystemVersion to
determine the OS version, because it's supported in all versions of macOS we
support for .NET.

Fixes issues such as this when building with XAMCORE_4_0:

> CoreMedia/CMSync.cs(590,11): error CS0103: The name 'PlatformHelper' does not exist in the current context

### DIFF
--- a/src/AddressBook/ABAddressBook.cs
+++ b/src/AddressBook/ABAddressBook.cs
@@ -86,7 +86,7 @@ namespace AddressBook {
 		{
 #if __MACCATALYST__
 			// avoid TypeLoadException if used before macOS 11.x
-			if (!UIKit.UIDevice.CurrentDevice.CheckSystemVersion (14,0))
+			if (!SystemVersion.CheckiOS (14,0))
 				return;
 #endif
 			// ensure we can init. This is needed before iOS6 (as per doc).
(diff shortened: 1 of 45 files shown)


mandel-macaque (approved):

dalexsoto (approved):

chamons (approved):
❤️ it!

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

* Xtro/Mac: Fixed in #13271.
* monotouch-test/watchOS 32-bits - simulator/Release (all optimizations): https://github.com/xamarin/maccore/issues/581

PR #13243, opened 2021-11-03
[AppKit/Foundation/UIKit] Use Foundation.NSWritingDirection instead of UIKit.UITextWritingDirection and AppKit.NSWritingDirection for .NET. Fixes #6573.

### LINKED ISSUE #6573: [Foundation] NSWritingDirection puzzling situation...

rolfbjarne:
Fixes this when building with XAMCORE_4_0:

> uikit.cs(6775,3): error CS0246: The type or namespace name 'UITextWritingDirection' could not be found (are you missing a using directive or an assembly reference?)
> uikit.cs(6597,3): error CS0246: The type or namespace name 'UITextWritingDirection' could not be found (are you missing a using directive or an assembly reference?)
> uikit.cs(6601,41): error CS0246: The type or namespace name 'UITextWritingDirection' could not be found (are you missing a using directive or an assembly reference?)

Fixes https://github.com/xamarin/xamarin-macios/issues/6573.

### DIFF
--- a/src/AppKit/Enums.cs
+++ b/src/AppKit/Enums.cs
@@ -1313,11 +1313,9 @@ namespace AppKit {
 		Natural = 4
 	}
 
-#if !XAMCORE_4_0 && MONOMAC
-	// Not hard deprecating now but until XAMCORE_4_0 happens or we can
-	// properly fix all the API using this.
+#if !NET && MONOMAC
+	// Use Foundation.NSWritingDirection in .NET.
 	// see: https://github.com/xamarin/xamarin-macios/issues/6573
-	// [Obsolete ("Use NSWritingDirection in Foundation instead.")]
 	[Flags]
 	[Native]
 	public enum NSWritingDirection : long {
@@ -1327,7 +1325,7 @@ namespace AppKit {
 		[Obsolete ("This API is not available on this platform.")]
 		Override = 2,
 	}
-#endif // !XAMCORE_4_0 && MONOMAC
+#endif // !NET && MONOMAC
 
 	[NoMacCatalyst]
 #if !XAMCORE_4_0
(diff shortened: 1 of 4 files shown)


chamons (approved):

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated (all due to https://github.com/xamarin/maccore/issues/2525).

PR #13258, opened 2021-11-03
[AudioConverter] Enable nullability + a few other code updates

rolfbjarne:


mandel-macaque (approved):

rolfbjarne:

PR #13259, opened 2021-11-03
[CI] Use a vimage for 10.15 since we do not have actual bots.

mandel-macaque:
We do not have actual bots for 10.15 but we do have an image. So in that case we can state that the pool should be an image. Changes will ensure that the stage will find a device that does have the correct os version.

### DIFF
--- a/tools/devops/automation/build-pipeline.yml
+++ b/tools/devops/automation/build-pipeline.yml
@@ -72,7 +72,8 @@ parameters:
     {
       stageName: 'mac_10_15',
       displayName: 'Mac Catalina (10.15)',
-      macPool: 'VSEng-Xamarin-RedmondMacBuildPool-iOS-Trusted',
+      macPool: 'macOS-10.15',
+      useImage: true,
       statusContext: 'Mac Catalina (10.15)',
       demands: [
         "Agent.OS -equals Darwin",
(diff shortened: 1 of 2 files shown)


tools/devops/automation/build-pipeline.yml
```diff
@@ -83,6 +84,7 @@ parameters:
       stageName: 'mac_10_14',
       displayName: 'Mac Mojave (10.14)',
       macPool: 'Hosted Mac Internal Mojave',
```
mjbond-msft, line 86: The `Hosted Mac Internal Mojave` pool has been retired. Instead recommend using `internal-macos-10.14` where you'll set `useImage` to true

mandel-macaque: wil do a pr with that change, thx!

tools/devops/automation/build-pipeline.yml
```diff
@@ -92,6 +94,7 @@ parameters:
       stageName: 'mac_10_13',
       displayName: 'Mac High Sierra (10.13)',
       macPool: 'Hosted Mac Internal',
```
mjbond-msft, line 96: You can use the `internal-macos-10.13` image in place of the `Hosted Mac Internal` pool

mjbond-msft (approved):

rolfbjarne (approved):

mandel-macaque:
Merging since the tests we wanted to run do run: https://github.com/xamarin/xamarin-macios/pull/13259#commitcomment-59396262

PR #13266, opened 2021-11-03
[Security/Authorization] Enable nullability + a few other code updates

rolfbjarne:


rolfbjarne:
Test failures are unrelated:

* Xtro/Mac: Fixed in #13271.

PR #13271, opened 2021-11-04
[Xtro] Fix xtro after a bad commit.

mandel-macaque:
I missed a failure in https://github.com/xamarin/xamarin-macios/pull/13244

PR #13276, opened 2021-11-04
[AudioComponent] Enable nullability + a few other code updates

rolfbjarne:
* Make the 'handle' field private.


rolfbjarne:
Test failure is unrelated (fixed in #13271).

tj-devel709 (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated and all due to https://github.com/xamarin/maccore/issues/2525.

PR #13278, opened 2021-11-04
[AudioUnit] Enable nullability + a few other code updates

rolfbjarne:


rolfbjarne:
Test failure is unrelated

* xammac tests/Mac Modern/Debug: https://github.com/xamarin/maccore/issues/2525
* Xtro/Mac: fixed in #13271.

mandel-macaque (approved):

tj-devel709 (approved):

PR #13292, opened 2021-11-05
[api-diff] Bump to latest stable (xcode13.1 branch) + add support for different reference api urls for legacy iOS and Mac.

rolfbjarne:
Also fix make logic to only have a single rule per hash, which avoids a few make warnings about duplicate targets.

Update README with new releases.


tj-devel709 (approved):

dalexsoto (approved):

rolfbjarne:

PR #13303, opened 2021-11-08
[CI test] Test if Tls12 failure is due to outdated OS.

rolfbjarne:
Run CI on an older branch to check if that also fails - if it does, then it's not a recent regression in our code.

Ref: https://github.com/xamarin/maccore/issues/2525


rolfbjarne:
Theory proved: the test fails on older branches as well, so this is not a regression in our code, it likely needs an OS update.

PR #13306, opened 2021-11-08
[CGPath] Fix leak when creating mutable CGPaths. Fixes #11339.

### LINKED ISSUE #11339: Leaks coming in from CGPath
### Steps to Reproduce

1. Use CGPath.FromRect
2. Profile in instruments
3. See unmanaged memory is leaked

### Expected Behavior

CGPath instances are properly cleaned up.

### Actual Behavior

It seems that there are copies that are lost along the way. MakeMutable always copies and all the users of the API pass it objects which are resulting from `Create*` APIs, which Cocoa API guidelines say that they mean you own the reference.

https://github.com/xamarin/xamarin-macios/blob/99fa931a279a220cd93f7ea842b7043e0e574932/src/CoreGraphics/CGPath.cs#L527

rolfbjarne:
Make the MakeMutable function take into account that the input might have to
be released, offer a way to do so, and update all callsites.

Fixes https://github.com/xamarin/xamarin-macios/issues/11339.

### DIFF
--- a/src/CoreGraphics/CGPath.cs
+++ b/src/CoreGraphics/CGPath.cs
@@ -506,9 +506,11 @@ namespace CoreGraphics {
 			gch.Free ();
 		}
 
-		static CGPath MakeMutable (IntPtr source)
+		static CGPath MakeMutable (IntPtr source, bool owns)
 		{
 			var mutable = CGPathCreateMutableCopy (source);
+			if (owns)
+				CGPathRelease (source);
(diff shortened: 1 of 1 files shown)


Therzok (approved):

mandel-macaque (approved):

filipnavara (approved):

rolfbjarne:

PR #13307, opened 2021-11-08
[Security] Remove the SslCipherSuite enum from .NET. Fixes #11498.

### LINKED ISSUE #11498: SslCipherSuite size
Apple defines `SSLCipherSuite` as:

```c
/*
 * Defined as enum for debugging, but in the protocol
 * it is actually exactly two bytes
 */
#if ((TARGET_OS_IPHONE && !TARGET_OS_MACCATALYST) || (TARGET_OS_OSX && TARGET_CPU_ARM64))
/* 16-bit value on iOS */
typedef uint16_t SSLCipherSuite;
#else
/* 32-bit value elsewhere */
typedef uint32_t SSLCipherSuite;
#endif
```

Apple's documentation about [SSLCipherSuite says](https://www.dropbox.com/s/6nysapfy3vmie0w/Screen%20Shot%202021-05-11%20at%2014.46.37.png?dl=0) (including as a screenshot, because it's changed in the past):

Screen Shot 2021-05-11 at 14 46 37

1. The comments in the headers are wrong.
2. The comments do not match the documentation.
3. The documentation is also wrong.

The truth is:

|               | uint16_t | uint32_t |
|---------------|:--------:|:--------:|
| iOS           |    X     |          |
| tvOS          |    X     |          |
| watchOS       |    X     |          |
| Mac Catalyst  |          |     X    |
| macOS (x64)   |          |     X    |
| macOS (arm64) |    X     |          |

The question is how to model this in C#, and there are two ways:

## 1️⃣ Make our enum size match Apple's

Pros: 

* It matches Apple, which is always good.
* Less likely to run into unwelcome surprises in API updates in the future.

Cons:

* It requires us to build different assemblies for x64 and arm64 for macOS.
* It makes code that works on iOS not work on Mac Catalyst.

## 2️⃣ Choose an enum size for all platforms, and hide the underlying size differences from app developers

Pros:

* Unified managed API.
* Simpler build logic (for our platform assemblies)
* Faster build (for Xamarin.Mac: no need to build multiple versions).

Cons:

* More work on our side (for the C# code). We'd have to:
    * Make sure we don't use `SslCipherSuite` in any P/Invoke and instead use the correct underlying type based on the platform and architecture.
* App developers would run into the same size problem if they were to create P/Invokes with our enum type.
* More likely that Apple comes up with something that breaks us in the future.

---

Until macOS (arm64) I leaned towards option 1, since it was the simpler option for us to implement, and just one of several API differences between iOS and Mac Catalyst. With the difference between x64 and arm64 on macOS, I'm leaning towards option 2, since having different managed API depending on only the architecture is something we've been able to avoid in the past (re: native enums).

To recap: thumbs up for option 2️⃣, thumbs down for option 1️⃣  

---

This test ignore must be reverted when this is fixed: https://github.com/xamarin/xamarin-macios/pull/11613

rolfbjarne:
* Remove the SslCipherSuite enum from .NET, it's complicated to implement
  correctly on macOS for both x64 and arm64, and it's also obsolete, so just
  remove it.
* Change the type for NSUrlSessionTaskTransactionMetrics.NegotiatedTlsCipherSuite
  to be TlsCipherSuite instead of SslCipherSuite for .NET (this is in fact the
  correct value according to the headers).

Fixes https://github.com/xamarin/xamarin-macios/issues/11498.

### DIFF
--- a/src/Security/SecProtocolMetadata.cs
+++ b/src/Security/SecProtocolMetadata.cs
@@ -125,39 +125,17 @@ namespace Security {
 		[Deprecated (PlatformName.TvOS, 13,0, message: "Use 'NegotiatedTlsCipherSuite' instead.")]
 		[Deprecated (PlatformName.WatchOS, 6,0, message: "Use 'NegotiatedTlsCipherSuite' instead.")]
 		[Deprecated (PlatformName.MacOSX, 10,15, message: "Use 'NegotiatedTlsCipherSuite' instead.")]
-#else
-		[UnsupportedOSPlatform ("ios13.0")]
-		[UnsupportedOSPlatform ("tvos13.0")]
-		[UnsupportedOSPlatform ("macos10.15")]
-#if IOS
-		[Obsolete ("Starting with ios13.0 use 'NegotiatedTlsCipherSuite' instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
-#elif TVOS
-		[Obsolete ("Starting with tvos13.0 use 'NegotiatedTlsCipherSuite' instead.", DiagnosticId = "BI1234", UrlFormat = "https://github.com/xamarin/xamarin-macios/wiki/Obsolete")]
-#elif MONOMAC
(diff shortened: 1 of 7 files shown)


chamons (approved):

rolfbjarne:

PR #13304, opened 2021-11-08
[AudioToolbox] Fix nullability issue to fix broken build.

rolfbjarne:
Fixes this compile error:

> AudioToolbox/MusicTrack.cs(256,11): error CS8603: Possible null reference return.

PR #13308, opened 2021-11-08
FromSequence returns a nullable

StephaneDelcroix:
Had trouble compiling today's version of macios. It happens that
FromSequence returns a track created by CreateTrack, and that one
can be null, so I changed the signature of FromSequence to reflect
that fact.


mandel-macaque:
@StephaneDelcroix thanks! AFAIK we should have no issues with the tests, if we do I'll let you know.

mandel-macaque (approved):

rolfbjarne:
This is a duplicate of #13304, which we've already merged.

PR #13313, opened 2021-11-08
[CI] Provide the stage that does the different steps to prepare for an insertion.

mandel-macaque:


dalexsoto (approved):

rolfbjarne (approved):

mandel-macaque:
Merging, we will need to test the insertion via a release pipeline that copies the artefacts. I can't be done manually :/

PR #13316, opened 2021-11-09
[src] Introduce DisposableObject and use it.

rolfbjarne:
Add a new DisposableObject class, which is a new base class below
NativeObject, to be used for types that aren't reference counted, but still
disposable. NativeObject is now implemented on top of this type, adding
support for reference counting.

Also change several type to use DisposableObject as their base type, which
makes it possible to share a lot more code and delete a good chunk of
duplicated code.

### DIFF
--- a/src/ObjCRuntime/DisposableObject.cs
+++ b/src/ObjCRuntime/DisposableObject.cs
@@ -1,0 +1,95 @@
+//
+// DisposableObject.cs: A base class for for many native data types without assuming any particular lifecycle.
+
+// Authors:
+//   Rolf Bjarne Kvinge
+//
+// Copyright 2021 Microsoft Corp
+//
+using System;
+using System.Runtime.InteropServices;
+using ObjCRuntime;
+using Foundation;
+
+#nullable enable
+
+namespace ObjCRuntime {
+	//
+	// The DisposableObject class is intended to be a base class for many native data
+	// data types, without assuming any particular lifecycle (might be reference counted,
+	// might not be).
+	//
+	// It provides the common boilerplate for this kind of objects and the Dispose
+	// pattern.
+	//
+	public abstract class DisposableObject : INativeObject, IDisposable {
+		IntPtr handle;
+		readonly bool owns;
+
+		public IntPtr Handle {
+			get => handle;
+			protected set => InitializeHandle (value);
+		}
+
+		protected bool Owns { get; }
+
+		protected DisposableObject ()
+		{
+		}
+
+		protected DisposableObject (IntPtr handle, bool owns)
+			: this (handle, owns, true)
+		{
+		}
+
+		protected DisposableObject (IntPtr handle, bool owns, bool verify)
+		{
+			InitializeHandle (handle, verify);
+			this.owns = owns;
+		}
+
+		~DisposableObject ()
+		{
+			Dispose (false);
+		}
+
+		public void Dispose ()
+		{
+			Dispose (true);
+			GC.SuppressFinalize (this);
+		}
+
+		protected virtual void Dispose (bool disposing)
+		{
+			ClearHandle ();
+		}
+
+		protected void ClearHandle ()
+		{
+			handle = IntPtr.Zero;
+		}
+
+		void InitializeHandle (IntPtr handle, bool verify)
+		{
+#if !COREBUILD
+			if (verify && handle == IntPtr.Zero && Class.ThrowOnInitFailure) {
+				throw new Exception ($"Could not initialize an instance of the type '{GetType ().FullName}': handle is null.\n" +
+				    "It is possible to ignore this condition by setting ObjCRuntime.Class.ThrowOnInitFailure to false.");
+			}
+#endif
+			this.handle = handle;
+		}
+
+		protected virtual void InitializeHandle (IntPtr handle)
+		{
+			InitializeHandle (handle, true);
+		}
+
+		public IntPtr GetCheckedHandle ()
+		{
+			if (handle == IntPtr.Zero)
+				ObjCRuntime.ThrowHelper.ThrowObjectDisposedException (this);
+			return handle;
+		}
+	}
+}
--- a/src/AudioToolbox/AudioConverter.cs
+++ b/src/AudioToolbox/AudioConverter.cs
@@ -30,6 +30,7 @@
 
 using System;
 using System.Runtime.InteropServices;
+using CoreFoundation;
 using Foundation;
 using ObjCRuntime;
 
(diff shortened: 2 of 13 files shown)


rolfbjarne:
Test failures are unrelated:

* monotouch-test/Mac [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2525
* monotouch-test/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2525
* dont link/Mac Catalyst [dotnet]/Release [dotnet]: https://github.com/xamarin/maccore/issues/2414
* link all/Mac Catalyst [dotnet]/Release [dotnet]: https://github.com/xamarin/maccore/issues/2414
* xammac tests/Mac Modern/Debug: https://github.com/xamarin/maccore/issues/2525
* xammac tests/Mac Modern/Release: https://github.com/xamarin/maccore/issues/2525
* xammac tests/Mac Modern/Release (all optimizations): https://github.com/xamarin/maccore/issues/2525

dalexsoto (approved):
A very small nit that you can fix in another PR, no need to block for that on this one

src/ObjCRuntime/DisposableObject.cs
```diff
@@ -0,0 +1,95 @@
...
+
+		protected DisposableObject (IntPtr handle, bool owns)
+			: this (handle, owns, true)
+		{
+		}
+
+		protected DisposableObject (IntPtr handle, bool owns, bool verify)
+		{
+			InitializeHandle (handle, verify);
+			this.owns = owns;
```
Therzok, line 48: Shoudln't this SuppressFinalize if the object is not owned?

src/ObjCRuntime/DisposableObject.cs
```diff
@@ -0,0 +1,95 @@
...
+
+		public void Dispose ()
+		{
+			Dispose (true);
+			GC.SuppressFinalize (this);
+		}
+
+		protected virtual void Dispose (bool disposing)
+		{
+			ClearHandle ();
```
Therzok, line 64: What's the use of clearing the IntPtr in here?

Therzok, line 64: ~What's the use of clearing the IntPtr in here?~

Edit: It's for ObjectDisposedException.

src/ObjCRuntime/DisposableObject.cs
```diff
@@ -0,0 +1,95 @@
...
+				throw new Exception ($"Could not initialize an instance of the type '{GetType ().FullName}': handle is null.\n" +
+				    "It is possible to ignore this condition by setting ObjCRuntime.Class.ThrowOnInitFailure to false.");
+			}
+#endif
+			this.handle = handle;
+		}
+
+		protected virtual void InitializeHandle (IntPtr handle)
+		{
+			InitializeHandle (handle, true);
```
Therzok, line 85: If InitializeHandle is overrideable, shoudln't ClearHandle also be overrideable?

src/ObjCRuntime/DisposableObject.cs
```diff
@@ -0,0 +1,95 @@
...
+	// The DisposableObject class is intended to be a base class for many native data
+	// data types, without assuming any particular lifecycle (might be reference counted,
+	// might not be).
+	//
+	// It provides the common boilerplate for this kind of objects and the Dispose
+	// pattern.
+	//
+	public abstract class DisposableObject : INativeObject, IDisposable {
+		IntPtr handle;
+		readonly bool owns;
```
Therzok, line 27: This looks like SafeHandle :D

src/ObjCRuntime/DisposableObject.cs
rolfbjarne: That's an interesting optimization! But what happens if a subclass override Dispose to free native resources independent on whether the handle is owned or not?

src/ObjCRuntime/DisposableObject.cs
rolfbjarne: Yeah, you've got a point, but I'd rather not make it overridable until we have a specific use-case for it (we can always make it virtual later if we want to, but we can't do the opposite).

src/ObjCRuntime/DisposableObject.cs
rolfbjarne: > The `owns` parameter seems to put the burden on the subclasses of DisposableObject.

Not sure I understand what you mean here?

src/ObjCRuntime/DisposableObject.cs
Therzok: Could be documented as a remarks, because I don't believe that is generally the case :p

Therzok: Removing the GC.SuppressFinalize is an easier change to allow for more cases than adding it in the futur breaking compat :D

rolfbjarne: Actually it won't work: the NativeObject subclass will retain if `owns` is false - thus taking ownership - and it must still be collected.

Therzok: Doesn't that mean NativeObject should ReRegisterFinalize because it retains - thus owns?

src/ObjCRuntime/DisposableObject.cs
Therzok: I'm basing the domain knowledge on SafeHandle and its inner workings, because it seems to have almost the same purpose: https://source.dot.net/#System.Private.CoreLib/SafeHandle.cs,53

From my understanding, the base subclass should be responsable for the last object release and the initial registration for dispose/finalize.

The subclasses shouldn't care about `owns`, as it should automatically be tracked in the underlying DisposableObject - regardless of the tracking mechanism.

src/ObjCRuntime/DisposableObject.cs
Therzok: I think I'm missing some context here. If the ownership is taken by managed, then wouldn't `owns` be passed as true?

rolfbjarne: > Doesn't that mean NativeObject should ReRegisterFinalize because it retains - thus owns?

Yes, I guess something like that could work. That means that in the end `NativeObject` instances would never suppress their finalize method, it would only have any effect for other `DisposableObject` subclasses.

> I think I'm missing some context here. If the ownership is taken by managed, then wouldn't `owns` be passed as true?

The `owns` parameter has two slightly different meaning depending on the type:

Passing `owns=false` means:

* For `NativeObject` subclasses it means that the class must retain the handle. At that point the `Disposable.Owns` field is more of a historical artifact, since it's not used (we could set it to `true` if that makes more sense).
* For other `DisposableObject` subclasses it currently means that the instance does not own the Handle, and must not free it.

Passing `owns=true` means that the caller has ownership of the handle in question, and gives it to the `DisposableObject` instance. The Handle must be freed in Dispose; `NativeObject` will release it, other subclasses will have their own logic.

The main difference between `NativeObject` and other `DisposableObject` subclasses is that the latter represents handles that can't be retained/released.

PR #13320, opened 2021-11-09
[DictionaryContainer] Improve nullability + a few other code improvements.

rolfbjarne:
* Add a [NotNullWhen] attribute to the 'key' parameter for the NullCheckAndRemoveKey
  method, which means that if the method returns true, the 'key' parameter wasn't
  null, and thus we can remove numerous ignored null check in call sites ("key!"
  -> "key").
* Use CFString.CreateNative/ReleaseNative instead of other means to create native
  strings (the fastest and least memory hungry option).
* Remove unnecessary IntPtr.Zero checks before calling NSArray.ArrayFromHandle.
* Use the null-safe NativeObjectExtensions.GetHandle extension method to get the
  handle instead of checking for null (avoids some code duplication).

### DIFF
--- a/src/Foundation/DictionaryContainer.cs
+++ b/src/Foundation/DictionaryContainer.cs
@@ -27,6 +27,7 @@
 //
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 
 #if !COREBUILD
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

dalexsoto (approved):

chamons (approved):

rolfbjarne:

PR #13332, opened 2021-11-10
[src] Change all usages of old-style availability attributes to new-style availability attributes.

rolfbjarne:
This will allow us to remove the old-style attributes from .NET (since they won't be used in shared code).


rolfbjarne:
Test failures are unrelated (all due to https://github.com/xamarin/maccore/issues/2525).

Issue #12749, opened 2021-09-16
GeneratorTests.BGenTests.StackOverflow20696157: build Expected: 0 But was: 1


rolfbjarne:
Test failure:

```
=> GeneratorTests.BGenTests.StackOverflow20696157
GeneratorTests.BGenTests.StackOverflow20696157
	sof20696157.cs
[...]

PR #13336, opened 2021-11-10
[src] Use 'csc -refout' to generate reference assemblies instead of mono-cil-strip.

rolfbjarne:
mono-cil-strip is buggy, and may generate invalid assemblies.

Ref: https://github.com/dotnet/runtime/issues/61402#issuecomment-965753455

Fixes https://github.com/xamarin/xamarin-macios/issues/12749.


rolfbjarne:

dalexsoto (approved):
Just curious, is this something we need to revert once https://github.com/mono/mono/issues/21302 is fixed?

rolfbjarne:
> Just curious, is this something we need to revert once [mono/mono#21302](https://github.com/mono/mono/issues/21302) is fixed?

No, we won't have to revert anything, this new approach is the best option (iirc mcs/csc didn't have the -refout option when we started using mono-cil-strip for this purpose)

PR #13338, opened 2021-11-11
[compare-commits] Ignore *.binlog files when creating a diff of the generated files.

rolfbjarne:
Removes this from the generator diff, which happens on every build:

    diff --git a/build/common/bgen.exe.binlog b/build-new/common/bgen.exe.binlog
    index fc4ecd1fe..3b0503171 100644
    Binary files a/build/common/bgen.exe.binlog and b/build-new/common/bgen.exe.binlog differ


rolfbjarne:

dalexsoto (approved):

PR #13339, opened 2021-11-11
[NSData] Enable nullability + a few other code updates

rolfbjarne:
* Remove the internal (IntPtr, bool) constructor, it has non-standard semantics and it's confusing.

### DIFF
--- a/src/Security/Trust.cs
+++ b/src/Security/Trust.cs
@@ -337,9 +337,9 @@ namespace Security {
 #if !NET
 		[Mac (10,9)]
 #endif
-		public NSData GetExceptions ()
+		public NSData? GetExceptions ()
 		{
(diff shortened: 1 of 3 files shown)


rolfbjarne:

src/Security/Trust.cs
```diff
@@ -337,9 +337,9 @@ public SecKey GetKey ()
 #if !NET
 		[Mac (10,9)]
 #endif
-		public NSData GetExceptions ()
+		public NSData? GetExceptions ()
 		{
-			return new NSData (SecTrustCopyExceptions (GetCheckedHandle ()), false); // inverted boolean?
+			return Runtime.GetNSObject<NSData> (SecTrustCopyExceptions (GetCheckedHandle ()), true);
```
dalexsoto, line 342: Fun change 😄

dalexsoto (approved):

Issue #12945, opened 2021-10-06
Hot Restart Prebuilt bundle is compiled with outdated SDK


filipnavara:
VS 2022 Preview 4.1 shipped with .NET 6 RC1 SDK. However, it also shipped the prebuilt version of `Xamarin.PreBuilt.iOS.app` in `Microsoft.iOS.Windows.Sdk` workload pack that was compiled with Preview 7 SDK. This causes the application to load outdated version of some platform assemblies, such as System.Net.Http.dll. At very least I would expect the prebuilt version to match the shipped SDK. Ideally I would expect the workload pack to have up-to-date version on the dotnet6 feeds where I download the other workload packs.

Ref: https://github.com/xamarin/xamarin-macios/blob/2bc66b0cbebdebfb6ec1a71ff1713c101d66cdf0/msbuild/Xamarin.iOS.Tasks.Windows/Xamarin.iOS.Tasks.Windows.csproj#L33-L36

rolfbjarne:
CC @mauroa I believe you've been working on this?

PR #13342, opened 2021-11-11
[msbuild/dotnet] Build the Xamarin.PreBuilt.iOS app bundle.

rolfbjarne:
Build the Xamarin.PreBuilt.iOS app bundle instead of using a prebuilt bundle.

This makes sure that we're always using the latest BCL.

Some accurate build massaging was needed, because:

* To build the prebuilt app we need the iOS workload installed (into our build-local
  .NET installation).
* The iOS workload contains the Microsoft.iOS.Windows.Sdk pack.
* The Microsoft.iOS.Windows.Sdk pack contains the prebuilt app.

Thus we had a circular reference. Fortunately, the Microsoft.iOS.Windows.Sdk pack
is only required on Windows, which means we can break this circular reference by:

* Mark Microsoft.iOS.Windows.Sdk pack as only to be installed on Windows (unfortunately
  it seems we have to list the exact runtime identifiers for the platforms where
  to install the pack, so we can't do something like "win-*", but new variations
  of the "win-*" runtime identifier shouldn't show up all that often).
* Build the prebuilt app on macOS.

This way we don't need the Microsoft.iOS.Windows.Sdk pack when installing the iOS
workload locally.

The .NET build order is now:

* Build general sdk, runtime and ref packs for .NET.
* Build the prebuilt app.
* Build the Microsoft.iOS.Windows.Sdk pack.

### DIFF
--- a/dotnet/Makefile
+++ b/dotnet/Makefile
@@ -42,6 +42,9 @@ $(1)_WINDOWS_NUGET_TARGETS = \
 endef
 $(foreach platform,$(DOTNET_WINDOWS_PLATFORMS),$(eval $(call DefineWindowsTargets,$(platform))))
 
+iOS_WINDOWS_NUGET_TARGETS += \
+	$(DOTNET_DESTDIR)/Microsoft.iOS.Windows.Sdk/tools/msbuild/iOS/Xamarin.PreBuilt.iOS.app.zip \
+
 DIRECTORIES += \
 	$(DOTNET_NUPKG_DIR) \
 	$(DOTNET_PKG_DIR) \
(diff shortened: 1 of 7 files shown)


dalexsoto (approved):
Nice!

chamons (approved):

mauroa (approved):

emaf (approved):
This is awesome! Thank you, Rolf!

tj-devel709 (approved):

filipnavara:
Fixes https://github.com/xamarin/xamarin-macios/issues/12945

rolfbjarne:
Test failures are unrelated:

* link sdk/iOS Unified 64-bits - simulator/Debug: https://github.com/xamarin/maccore/issues/2529
* Generator tests/NUnit: https://github.com/xamarin/xamarin-macios/issues/12749

rolfbjarne:
@mauroa / @emaf could you verify that this is actually working?

PR #13343, opened 2021-11-11
[mtouch] Fix logic to determine whether we're using Mono.framework on watchOS.

rolfbjarne:
The previous "8.0" sdk version check is incorrect (that's "iOS 8.0", the first
iOS version to add support for user frameworks). However, it started kicking
in for watchOS 8.0, and now suddenly we're trying to build all watchOS
extensions with Mono.framework, which is not what we want.

Instead modify the code to:

* Split by platform, keep the existing behavior for iOS, tvOS and Mac Catalyst
  (since "8.0" is the correct version check for all those platforms).
* Use the Mono framework in watchOS if either of the following are true:
	* In the main watchOS extension if it has nested extensions.
	* In any other extensions (because they'll be nested extensions).

Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1430408

### DIFF
--- a/tools/mtouch/Application.mtouch.cs
+++ b/tools/mtouch/Application.mtouch.cs
@@ -798,13 +798,32 @@ namespace Xamarin.Bundler {
 			if (!enable_msym.HasValue)
 				enable_msym = !EnableDebug && IsDeviceBuild;
 
-			if (!UseMonoFramework.HasValue && DeploymentTarget >= new Version (8, 0)) {
-				if (IsExtension) {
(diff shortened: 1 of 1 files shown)


chamons (approved):

tj-devel709 (approved):

mandel-macaque:
funky bot, needed a re-build.

mandel-macaque (approved):

dalexsoto (approved):

Issue #10030, opened 2020-11-03
[.NET 6] Check which files should be included by default for asset catalogs


rolfbjarne:
According to Apple's documentation (https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_ref-Asset_Catalog_Format/AssetTypes.html), it seems that pretty much any filetype could be included inside an \*.xcassets directory (the "Data Set Type" can contain "files in any format except binary executable code").

We currently only include by default \*.pdf, \*.jpg, \*.png and \*.json files: https://github.com/xamarin/xamarin-macios/blob/main/dotnet/DefaultCompilationIncludes.md#asset-catalogs

We should add tests and possibly update the files included by default to include all files (except `.DS_Store`?) inside \*.xcassets directories.

rolfbjarne:
Fixed in https://github.com/xamarin/xamarin-macios/pull/13346.

PR #13346, opened 2021-11-11
[Dotnet] Automatically Allow Assets of all types and Test

tj-devel709:
This PR addresses this issue: https://github.com/xamarin/xamarin-macios/issues/10030

In order to build a dotnet project on the VSMac2022 Preview, I copied the xamarin-macios/tests/dotnet/AppWithResources solution to start. 

**There are a ton of "new" files so I will highlight the important ones**:
I made this change here to not only accept PDFs, JPGs, PNGs, and JSON files but any file type other than DS_Store files in the format that the Assets.xcassets expects similar to the [format Xcode uses](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_ref-Asset_Catalog_Format/AssetTypes.html). 

Then the test resides here:
- I first delete any contents inside the Assets.xcassets directory inside the project
- Recopy the Assets into the project 
- Clean and Build the project
- Get the assets binary file (assets.car) and process it using "xcrun --sdk iphoneos assetutil --info"
- Turn the output to a JsonDocument and search for the expected json objects stored inside XCAssetTargets


tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+					myProcess.StartInfo.UseShellExecute = false;
+					myProcess.StartInfo.RedirectStandardOutput = true;
+					myProcess.StartInfo.Arguments = $"--sdk iphoneos assetutil --info {assetsPath}";
+					myProcess.StartInfo.FileName = "xcrun";
+					myProcess.StartInfo.CreateNoWindow = true;
+					myProcess.Start ();
+
+					string output = myProcess.StandardOutput.ReadToEnd ();
+					StreamReader reader = myProcess.StandardOutput;
+					myProcess.WaitForExit ();
```
mandel-macaque, line 779: This can block the test FOR EVER if the process that is being executed does not exit, you need to use https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=net-5.0#System_Diagnostics_Process_WaitForExit_System_Int32_ and pass a timeout. Then check the result if it is true of false and assert for it to be false.

You can do that in several ways, use an out parameter, to state if the process exited correctly or move the return type to JsonDocument? and assert is not null.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			}
+			return foundElements;
+		}
+
+		string IsTarget (JsonElement item) {
+			if (item.TryGetProperty ("AssetType", out var assetType)) {
+				foreach (var target in XCAssetTargets) {
+					var result = IsTarget (item, assetType, target);
+					if (result is not null)
+						return result;
```
mandel-macaque, line 808: return result? You already check in the caller if it is null, and null seems to be acceptable

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -4,6 +4,8 @@
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Diagnostics;
+using System.Text.Json;
 
 using Mono.Cecil;
```
mandel-macaque, line 11: Add nullable enabled please. There are going to be a fair amount of warnings in this code.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+
+		void CopyAssets (string project_path)
+		{
+			DirectoryInfo testingAssetsDir = new DirectoryInfo (Path.Combine (project_path, @"../../TestingAssets"));
+			string xcassetsDir = Path.Combine (project_path, @"../Assets.xcassets");
+
+			if (!testingAssetsDir.Exists)
+				throw new DirectoryNotFoundException ($"TestingAssets directory does not exist or could not be found: {testingAssetsDir}");
+
+			if (!Directory.Exists (xcassetsDir))
```
mandel-macaque, line 752: What happen if it exists? Use the old one? That is going to be error prone.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+
+		void DeleteAssets (string project_path)
+		{
+			string xcassetsDir = Path.Combine (project_path, @"../Assets.xcassets");
+			try {
+				if (Directory.Exists (xcassetsDir))
+					Directory.Delete (xcassetsDir, true);
+			}
+			catch (Exception e){
+				Console.WriteLine($"Deleting Assets.xcassets failed: {e.Message}");
```
mandel-macaque, line 740: A console, no error no nothing. This is dangerous, if the exception is thrown, do not just catch and print. Either surface the exception, no tan issue, test fails we need to fix the test.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+		[TestCase (ApplePlatform.MacOSX, "osx-x64")]
+		[TestCase (ApplePlatform.MacOSX, "osx-arm64;osx-x64")] // https://github.com/xamarin/xamarin-macios/issues/12410
+		public void AppWithXCAssets (ApplePlatform platform, string runtimeIdentifiers)
+		{
+			var project = "AppWithXCAssets";
+			Configuration.IgnoreIfIgnoredPlatform (platform);
+
+			var project_path = GetProjectPath (project, runtimeIdentifiers: runtimeIdentifiers, platform: platform, out var appPath);
+
+			DeleteAssets (project_path);
```
mandel-macaque, line 701: Are the assets shared between the projects? If they are, the sis going to make it hard to debug.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+				file.CopyTo (tempPath, false);
+			}
+			foreach (DirectoryInfo subDir in dir.GetDirectories ()){
+				CopyDirectoryContents (subDir, Path.Combine (DestDirectory, subDir.Name));
+			}
+		}
+
+		JsonDocument ProcessAssets (string assetsPath) {
+			try {
+				using (Process myProcess = new Process ()) {
```
rolfbjarne, line 769: We have helper API to run a process and capture the output.

Something like this:

```csharp
var output = new StringBuilder ();
var executable = "xcrun";
var arguments = new string [] { "--sdk", "phones", "assetutil", "--info", assetPath };
var rv = Execution.RunWithStringBuildersAsync (executable, arguments, standardOutput: output, standardError: output, timeout: TimeSpan.FromSeconds (15)).Result;
Assert.AreEqual (0, rv.ExitCode, "ExitCode");

return JsonDocument.Parse (output);
```

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +684,173 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+					string output = myProcess.StandardOutput.ReadToEnd ();
+					StreamReader reader = myProcess.StandardOutput;
+					myProcess.WaitForExit ();
+
+					return JsonDocument.Parse (output);
+				}
+			}
+			catch (Exception e) {
+				Console.WriteLine($"Couldn't process: {e.Message}");
+			}
```
rolfbjarne, line 786: As Manuel said, don't swallow exceptions, it might hide test failures.

tests/dotnet/UnitTests/ProjectTest.cs
rolfbjarne: I'm guessing you're deleting and copying assets to avoid duplicating them for all the platforms in every project directory - but does it work if you create a symlink instead? So in each project directory, there would be an Assets.xcassets symlink pointing to the parent folder's TestingAssets directory.

That would make it possible to remove all the delete+copy code.

rolfbjarne (changes requested):

tj-devel709:
@rolfbjarne While running these tests locally, I realized that the 
`TestCase (ApplePlatform.iOS, "iossimulator-x64"`and `TestCase (ApplePlatform.TVOS, "tvossimulator-x64"` 
pass, but `TestCase (ApplePlatform.MacCatalyst, "maccatalyst-x64"` and `TestCase (ApplePlatform.MacOSX, "osx-x64"` fails.

I copied most of the test logic from the AppWithResources project and I don't see anything sticking out as to why the mac and maccatalyst tests would not be passing (they both pass when I run the AppWithResources test). They are both getting this error from this line:
`DotNet.AssertBuild (project_path, GetDefaultProperties (runtimeIdentifiers));` 

Is there anything that pops out to you as to why this might be happening?

rolfbjarne:
@tj-devel709 you'd have to look at the output from the test; it's probably a build failure of some sort. The test will create a binlog for the build (the path is printed to the output), which you can view to see where things went wrong.

tests/dotnet/UnitTests/ProjectTest.cs
tj-devel709: Ah I do this since that foreach right above that line will return null multiple times while it is searching for useful assets and we only want to add it to the foundElements hashset when it finds a useful one

tj-devel709: I only want to return the result here if it is not null since it will be going through the foreach loop and checking if the current JsonElement matches the XCAssetTarget we are looking at in that iteration. If not, I want it to continue through all the XCAssetTargets and only return when it finds it, or if it does not find it.

tj-devel709:
> @tj-devel709 you'd have to look at the output from the test; it's probably a build failure of some sort. The test will create a binlog for the build (the path is printed to the output), which you can view to see where things went wrong.

One of the svg files were too new of a version! I never would have guessed that haha thanks

tests/dotnet/UnitTests/ProjectTest.cs
tj-devel709: @mandel-macaque  @rolfbjarne 
So my idea behind the copying and deleting was that we could replicate someone adding new assets to a project each time the test runs. I thought this would be good in case the project already included the assets in a previous run and we change something that does not allow them to be included later. Maybe this is not a concern? 

And currently, the directory that they copy from is in `ProjectWithXCAssets/` and they are copying into `ProjectWithXCAssets/{platform}/`. I didn't think this would be an issue, but after talking to @mandel-macaque, there are two testcases with each platform ("iossimulator-x64" and "ios-arm64;ios-arm") which share the `ProjectWithXCAssets/iOS/` project path for example.

I don't know much about symlinks, but I will see if that can work here!

rolfbjarne: > So my idea behind the copying and deleting was that we could replicate someone adding new assets to a project each time the test runs.

That's not a bad idea, but it's a separate test, and would have to be written differently since the tests are by default started from a clean project directory:

1. Build the project
2. Add assets
3. Build again to see what happens.

> Edit: Does my first concern even need to be addressed?

No

> If not then the Symlink idea is just a way to not copy the assets into each project, correct?

Correct, it would just be a way to not quadruplicate all the assets (git is smart and it doesn't matter all that much, but if it works, why not?)

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -135,9 +139,9 @@ public void BuildEmbeddedResourcesTest (string platform)
 			Assert.That (ad.MainModule.Resources.Count, Is.EqualTo (1), "1 resource");
 			Assert.That (ad.MainModule.Resources [0].Name, Is.EqualTo ("EmbeddedResources.Welcome.resources"), "libtest.a");
 			var asm_dir = Path.GetDirectoryName (asm);
-			Assert.That (Path.Combine (asm_dir, "en-AU", "EmbeddedResources.resources.dll"), Does.Exist, "en-AU");
-			Assert.That (Path.Combine (asm_dir, "de", "EmbeddedResources.resources.dll"), Does.Exist, "de");
-			Assert.That (Path.Combine (asm_dir, "es", "EmbeddedResources.resources.dll"), Does.Exist, "es");
+			Assert.That (Path.Combine (asm_dir!, "en-AU", "EmbeddedResources.resources.dll"), Does.Exist, "en-AU");
```
rolfbjarne, line 142: Changing the line above to:

```csharp
var asm_dir = Path.GetDirectoryName (asm)!;
```

would be a bit simpler

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -312,11 +316,11 @@ public void BuildInterdependentBindingProjects (string platform)
 				var ad = AssemblyDefinition.ReadAssembly (asm, new ReaderParameters { ReadingMode = ReadingMode.Deferred });
 				Assert.That (ad.MainModule.Resources.Count, Is.EqualTo (0), "0 resources for interdependent-binding-projects.dll");
 
-				var ad1 = AssemblyDefinition.ReadAssembly (Path.Combine (asmDir, "bindings-test.dll"), new ReaderParameters { ReadingMode = ReadingMode.Deferred });
+				var ad1 = AssemblyDefinition.ReadAssembly (Path.Combine (asmDir!, "bindings-test.dll"), new ReaderParameters { ReadingMode = ReadingMode.Deferred });
```
rolfbjarne, line 319: Changing the `asmDir` declaration to the following is simpler:

```csharp
var asmDir = Path.GetDirectoryName (asm)!;
```

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,164 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			new XCAssetTarget ("Data", "Name", new string [] { "BmpImageDataTest", "JsonDataTest", "DS_StoreDataTest",
+				"DngImageDataTest", "EpsImageDataTest", "TiffImageDataTest" }),
+
+			new XCAssetTarget ("Color", "Name", new string [] { "ColorTest" }),
+
+			new XCAssetTarget ("Contents", "Name", new string [] { "SpritesTest" }),
+
+			new XCAssetTarget ("Texture Rendition", "Name", new string [] { "TextureTest" }),
+
+			new XCAssetTarget ("Vector", "RenditionName", new string [] { "samplepdf.pdf", "loopsvg.svg" }),
```
mandel-macaque, line 844: Lay manual would re-write this to type less due to his fat fingers, no need to add the class fin the new, it can be inferred, same for the array.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,164 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+		void DeleteAssets (string project_path)
+		{
+			string xcassetsDir = Path.Combine (project_path, @"../Assets.xcassets");
+			try {
+				if (Directory.Exists (xcassetsDir))
+					Directory.Delete (xcassetsDir, true);
+			}
+			catch (Exception e){
+				Assert.Fail ($"Deleting Assets.xcassets failed: {e.Message}");
+			}
```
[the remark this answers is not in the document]
tj-devel709: Hmm so would this look like removing the try catch and just having the if statement without any Assert here?

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,164 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+		JsonDocument ProcessAssets (string assetsPath) {
+			var output = new StringBuilder ();
+			var executable = "xcrun";
+			var arguments = new string [] { "--sdk", "iphoneos", "assetutil", "--info", assetsPath };
+			var rv = Execution.RunWithStringBuildersAsync (executable, arguments, standardOutput: output, standardError: output, timeout: TimeSpan.FromSeconds (120)).Result;
+			Assert.AreEqual (0, rv.ExitCode, "ExitCode");
+
+			// This Execution call produces an output with an objc warning. We just want the json below it.
+			if (output.ToString ().StartsWith ("objc"))
+				output.Remove (0, output.ToString ().IndexOf ('\n'));
```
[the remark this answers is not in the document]
tj-devel709: I added a string var instead, is this better? 
Tried inspecting in SharpLab but not too sure...
I guess it's a tradeoff between storing another string var vs calling .ToString () two more times but it does seem more concise with the string var

tests/dotnet/UnitTests/ProjectTest.cs
tj-devel709: The XCAssetTarget was fine being left off, but the string [] complained when removed!

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,164 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			}
+			foreach (DirectoryInfo subDir in dir.GetDirectories ()){
+				CopyDirectoryContents (subDir, Path.Combine (DestDirectory, subDir.Name));
+			}
+		}
+
+		JsonDocument ProcessAssets (string assetsPath) {
+			var output = new StringBuilder ();
+			var executable = "xcrun";
+			var arguments = new string [] { "--sdk", "iphoneos", "assetutil", "--info", assetsPath };
```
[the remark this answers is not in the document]
tj-devel709: @mandel-macaque Two Questions with this:
1) Should MacCatalyst use iphoneos15.0?
2) Is there an easy way to get the most updated SDK Version number? For example, if I hard code iphoneos to be 15.0, next update will need to be changed manually, or would I need to run this `xcodebuild -showsdks` command in another process and grep the outputs?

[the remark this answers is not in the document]
rolfbjarne: 1. I believe Mac Catalyst uses `macosx12.0`, but you can check that locally to see if it works.
2. The Xamarin.Tests.Configuration class has version variables you can use: 

https://github.com/xamarin/xamarin-macios/blob/71371d6446db2e6624e654947524d4429ed0ee6c/tests/common/Configuration.cs#L25-L28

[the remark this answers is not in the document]
tj-devel709: Oh awesome thanks!!

tests/dotnet/UnitTests/ProjectTest.cs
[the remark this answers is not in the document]
mandel-macaque: Yes, do remove the try and the catch and do not assert, the exception will be a failure. Catching an exception and not providing all the information is a very bad practice.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,205 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+					Directory.Delete (xcassetsDir, true);
+			}
+			catch (Exception e){
+				Assert.Fail ($"Deleting Assets.xcassets failed: {e.Message}");
+			}
+		}
+
+		void CopyAssets (string project_path)
+		{
+			DirectoryInfo testingAssetsDir = new DirectoryInfo (Path.Combine (project_path, @"../../TestingAssets"));
```
mandel-macaque, line 784: Why the need of a verbatim string literal?

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,205 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			DirectoryInfo xcassetsDir = new DirectoryInfo (Path.Combine (project_path, @"../Assets.xcassets"));
+
+			Assert.That (testingAssetsDir, Does.Exist);
```
mandel-macaque, line 787: Needs a message since there are two asserts in the method.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,205 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			if (!Directory.Exists (xcassetsDir.FullName))
+				MakeSymlinks (testingAssetsDir.FullName, xcassetsDir.FullName);
+			Assert.That (xcassetsDir, Does.Exist);
```
mandel-macaque, line 790: Same missing message.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,205 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+			if (assetType.ToString () == target.AssetType) {
+				if (item.TryGetProperty (target.CategoryName, out var value)) {
+					if (target.Values.Contains (value.ToString ()))
+						return string.Concat (assetType.ToString (), ".", value.ToString ());
+				}
+			}
+			return null;
+		}
+
+		class XCAssetTarget {
```
mandel-macaque, line 861: could we use a struct?

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -682,6 +686,205 @@ public void AppWithResources (ApplePlatform platform, string runtimeIdentifiers)
...
+						return result;
+				}
+			}
+			return null;
+		}
+
+		string? GetTarget (JsonElement item, JsonElement assetType, XCAssetTarget target)
+		{
+			if (assetType.ToString () == target.AssetType) {
+				if (item.TryGetProperty (target.CategoryName, out var value)) {
```
mandel-macaque, line 853: This could be combined in a single if.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/ProjectTest.cs
tj-devel709: Ah I just used it somewhere else and used it here too. I'll remove it!

tests/dotnet/UnitTests/ProjectTest.cs
tj-devel709: Here's a snippet of what Microsoft docs say:
```
❌ AVOID defining a struct unless the type has all of the following characteristics:

- It logically represents a single value, similar to primitive types (int, double, etc.).
- It has an instance size under 16 bytes.
- It is immutable.
- It will not have to be boxed frequently.

In all other cases, you should define your types as classes.
```
Since there are string arrays inside the XCAssetTarget, I am thinking that the instance size will be much greater than 16 bytes in most of the uses if I understand that correctly. Therefore I am leaning towards a class?

mandel-macaque: Not of terrible importance to switch to a struct perse but you are not correctly understanding how the array is stored. Please read: https://stackoverflow.com/questions/1113819/arrays-heap-and-stack-and-value-types and pay attention at some nice graphics at the bottom of the post.

mandel-macaque: When don't care that much in the tests, leave as it is ;)

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -770,8 +770,8 @@ void ConfigureAssets (string project_path, string runtimeIdentifiers, bool start
 		void DeleteAssets (string project_path)
 		{
 			string xcassetsDir = Path.Combine (project_path, "../Assets.xcassets");
-			if (Directory.Exists (xcassetsDir))
-				Directory.Delete (xcassetsDir, true);
+			if (File.Exists (xcassetsDir))
+				File.Delete (xcassetsDir);
```
mandel-macaque, line 774: @tj-devel709  as mentioned in our chat. File.Delete will not care if the file is not present. That check is redundant.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,231 @@
...
+		void DeleteAssets (string project_path)
+		{
+			string xcassetsDir = Path.Combine (project_path, "../Assets.xcassets");
+			File.Delete (xcassetsDir);
+		}
+
+		void CopyAssets (string project_path)
+		{
+			DirectoryInfo testingAssetsDir = new DirectoryInfo (Path.Combine (project_path, "../../TestingAssets"));
+			DirectoryInfo xcassetsDir = new DirectoryInfo (Path.Combine (project_path, "../Assets.xcassets"));
```
mandel-macaque, line 127: same, you can use var and have all the code look the same.

mandel-macaque (changes requested):
Looking much more cleaner in the new fixture. Also less error prone ( we can talk about what happens with an Assert that fails offline).

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,231 @@
...
+			this.sdkVersion = sdkVersion;
+		}
+
+		[SetUp]
+		public void Init ()
+		{
+			var project = "AppWithXCAssets";
+			Configuration.IgnoreIfIgnoredPlatform (platform);
+			project_path = GetProjectPath (project, runtimeIdentifiers: runtimeIdentifiers, platform: platform, out var appPath);
+			this.appPath = appPath;
```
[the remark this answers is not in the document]
tj-devel709: I added this here since the appPath will be needed for the tests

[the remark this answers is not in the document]
mandel-macaque: that is not my comment. In you code you have two variable with the same name with different scopes when you do not need to.

You have:

* appPath:  method variable
* appPath: instance variable

the out parameter of the method can take the instance variable meaning you can remove the line:
```csharp
this.appPath = appPath;
```

Also, know that I re-reading the code, please use cameCase in all variable and do not use sname_case. That is coming from the code you copied from the old test and is not the correct mono coding style.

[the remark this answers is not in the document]
tj-devel709: Ah! So it will assign the instance variable automatically?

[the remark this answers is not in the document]
tj-devel709: Oh I see, you removed the var from 'out var appPath'!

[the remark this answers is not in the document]
mandel-macaque: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,218 @@
...
+			this.sdkVersion = sdkVersion;
+		}
+
+		[SetUp]
+		public void Init ()
+		{
+			var project = "AppWithXCAssets";
+			Configuration.IgnoreIfIgnoredPlatform (platform);
+			project_path = GetProjectPath (project, runtimeIdentifiers: runtimeIdentifiers, platform: platform, out appPath);
+			DeleteAssets (project_path);
```
mandel-macaque, line 48: Why is delete assets in the setup? What is the goal? Should it be in a tear down method too? Unit tests should leave the system as they found it.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/AssetsTest.cs
tj-devel709: I thought the point of adding the setup was to delete the assets in the setup and not have to do it inside the tests. Otherwise, I don't see the point in having the setup since it would just be creating the project_path that we need to now use instance variables for. I guess the delete can happen in the tear down as well and then the delete in the setup would just make sure it is not there, the test would add it, and then the clean up would remove it and then it would be the same as before we started the test!

mandel-macaque: The issue is a follows:

1. Call setup. Set the project paths etc, configure the assets (which you are doing in the test)
2. Execute the code to test and assert expectations. Do not do any configuration here, that's part of the setup.
3. Tear down and delete de assets.

I don't understand why you call delete, since nothing has been configured at that point, and any left overs have to be cleaned in the tear down. The 3 steps I have described are the basics of unit testing.

chamons (commented):
There are a number of diffs like this:

```
+++ b/tests/dotnet/AppWithXCAssets/tvOS/Assets.xcassets
@@ -0,0 +1 @@
+/Users/tjlambert/Microsoft/FileTypes/xamarin-macios/tests/dotnet/AppWithXCAssets/TestingAssets
\ No newline at end of file
```

Why do we have files with your local path?

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,223 @@
...
+			this.runtimeIdentifiers = runtimeIdentifiers;
+			this.sdkVersion = sdkVersion;
+		}
+
+		[SetUp]
+		public void Init ()
+		{
+			var project = "AppWithXCAssets";
+			Configuration.IgnoreIfIgnoredPlatform (platform);
+			project_path = GetProjectPath (project, runtimeIdentifiers: runtimeIdentifiers, platform: platform, out appPath);
```
mandel-macaque, line 47: Do not use snake casing.

mandel-macaque (changes requested):

tests/dotnet/UnitTests/AssetsTest.cs
mandel-macaque: An important detail to remember is that the assert methods form nunit use exceptions, that means that the execution of the test stops as soon and there is a failing assertion. 

If the steps that need to clean the code are after the asserts and we have a failure, the clean steps won’t be executed. That forces the setup to be written with the assumption of a dirty workspace which is incorrect.

For reference: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Assert.cs#L380

That is one of the main reasons I asked you to move to use a setup and a tear down. Not only to make the code much cleaner (as @chamons  wonderfuly explained) but to avoid  tests to have side effects for subsequent tests. What you were experiencing with the previous implementation, in which you did not know if you had to delete a symlink, is rooted in that your tests were not correctly isolated from each other. 

Calling delete assets in the setup was/is a red flag. The same with the method that configured the assets, those should be in the setup method too.

tj-devel709: @chamons @mandel-macaque AH okay thank you both for the explanations! I will get right on this :)

tj-devel709:
> There are a number of diffs like this:
> 
> ```
> +++ b/tests/dotnet/AppWithXCAssets/tvOS/Assets.xcassets
> @@ -0,0 +1 @@
> +/Users/tjlambert/Microsoft/FileTypes/xamarin-macios/tests/dotnet/AppWithXCAssets/TestingAssets
> \ No newline at end of file
> ```
> 
> Why do we have files with your local path?

@chamons Ah interesting... These were symlinks that are removed now. This is the only place that I see this, did you catch these on anything other than the Assets.xcassets files?

mandel-macaque (changes requested):
Nearly completed, small tiny fixes.

tests/dotnet/UnitTests/ProjectTest.cs
```diff
@@ -4,6 +4,8 @@
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Diagnostics;
+using System.Text.Json;
```
mandel-macaque, line 8: I'm sure System.Text.Json is not longer used in the test. @tj-devel709  I have approved the PR but please remove it.

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,221 @@
...
+			var doc = ProcessAssets (assetsCar, GetFullSdkVersion ());
+			Assert.IsNotNull (doc, "There was an issue processing the asset binary.");
+
+			var foundAssets = FindAssets (doc);
+
+			// Seems the 2 vectors are not being consumed in MacCatalyst but they still appear in the image Datasets
+			var TotalUniqueAssets = platform == ApplePlatform.MacCatalyst ? 14 : 16;
+
+			Assert.AreEqual (TotalUniqueAssets, foundAssets.Count, "Wrong number of assets found");
+			Assert.That (foundAssets, Has.No.Member ("Data.DS_StoreDataTest"), "DS_Store files should not be included.");
```
rolfbjarne, line 82: I think it would be better to assert that the entire set of assets is as expected.

So list all the expected assets (in a hashset for instance), and then assert that the expected assets and the found assets all have the same paths.

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,221 @@
...
+			var output = new StringBuilder ();
+			var executable = "xcrun";
+			var arguments = new string [] { "--sdk", sdkVersion, "assetutil", "--info", assetsPath };
+			var rv = Execution.RunWithStringBuildersAsync (executable, arguments, standardOutput: output, standardError: output, timeout: TimeSpan.FromSeconds (120)).Result;
+			Assert.AreEqual (0, rv.ExitCode, $"Processing Assets Error: {rv.StandardError}. Unexpected ExitCode");
+			var s = output.ToString ();
+
+			// This Execution call produces an output with an objc warning. We just want the json below it.
+			if (s.StartsWith ("objc"))
+				output.Remove (0, s.IndexOf (Environment.NewLine) + 1);
```
rolfbjarne, line 153: You can just create a new string like this:

```suggestion
				s = s.Substring (s.IndexOf (Environment.NewLine) + 1);
```

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,221 @@
...
+
+			return JsonDocument.Parse (output.ToString ());
```
rolfbjarne, line 155: and then pass `s` here:

```suggestion
			return JsonDocument.Parse (s);
```

This avoids modifying both the StringBuilder and then create a new string.

rolfbjarne (changes requested):

tests/dotnet/UnitTests/AssetsTest.cs
mandel-macaque: I don't know how i missed this 🤦

tests/dotnet/UnitTests/AssetsTest.cs
```diff
@@ -0,0 +1,221 @@
...
+			if (isStartingWithAssets)
+				CopyAssets ();
+
+			Clean (projectPath);
+
+			DotNet.AssertBuild (projectPath, GetDefaultProperties (runtimeIdentifiers));
+			if (!isStartingWithAssets) {
+				CopyAssets ();
+				// Building the project twice without cleaning in between fails: https://github.com/xamarin/maccore/issues/2530
+				Clean (projectPath);
```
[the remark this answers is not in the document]
tj-devel709: So I merged that other PR and merged main into this PR, but it still did not work. While tracing, it seems that the new assets are not being compiled on the second build if I build, add assets, build but they get compiled if I build, add assets, clean, build. New issue is here: https://github.com/xamarin/maccore/issues/2534

rolfbjarne (approved):

PR #13356, opened 2021-11-12
[src] Add a new ObjCRuntime.NativeHandle type to represent native handles for .NET. Fixes #13126.

### LINKED ISSUE #13126: .NET: Use a different type than IntPtr for our handles
Current status is that we're using `IntPtr` for all our native handles. This is troublesome in .NET, because we won't be able to distinguish between pointer-sized numbers (`nint`) and plain pointers/handles, so as part of the breaking changes that are coming in .NET (#13087), we've decided to use a different type than `IntPtr` going forward.

Fundamentally there are two options:

1. A custom struct that contains a single IntPtr field - the purpose is for everything to behave the same, but have a different type than `IntPtr`.
2. A custom `SafeHandle` subclass (or more than one).

### Advantages

* No additional memory usage.
* Shouldn't have any memory management surprises, since it should behave the same.

### Disadvantages

* There might be marshaling problems on some platforms if we use the custom struct in P/Invokes, since a primitive type is not necessarily always marshaled the same way a struct. It'll be possible to avoid this though by always using IntPtr in P/Invokes.

### Advantages

* Allows for more advanced memory management scenarios.
* Seems to be the recommended solution in .NET.

### Disadvantages

* Since SafeHandle is a class, we'd double the number of managed instances (an additional SafeHandle instance for every NSObject instance).
* The memory management will be different, so there might be unpleasant surprises.

rolfbjarne:
Add a new struct, ObjCRuntime.NativeHandle, which will be used to represent
native handles for .NET (instead of using IntPtr). The main purpose is to be
able to use 'nint' as a number in API while not being prevented from using
native handles as well.

One example is NSMutableString, which has a constructor that takes a single
'nint capacity' parameter. With this change, we'll also be able to have a
constructor that takes a native handle in .NET - otherwise we'd have two
constructors with the same signature, because a C# 'nint' is just an 'IntPtr'.

This change required numerous changes pretty much everywhere. The work is
split up in commits as well as I was able to, and each commit explains what it
does.

Fixes https://github.com/xamarin/xamarin-macios/issues/13126.

This pull request is best reviewed commit-by-commit, with a cup of ☕️ or 🫖 in
hand, and a good chunk of free time.

### DIFF
--- a/src/ObjCRuntime/NativeHandle.cs
+++ b/src/ObjCRuntime/NativeHandle.cs
@@ -1,0 +1,92 @@
+#if NET
+
+using System;
+using System.Collections.Generic;
+
+#nullable enable
+
+namespace ObjCRuntime {
+	public readonly struct NativeHandle : IEquatable<NativeHandle> {
+		readonly IntPtr handle;
+
+		public IntPtr Handle {
+			get { return handle; }
+		}
+
+		public static NativeHandle Zero = default (NativeHandle);
+
+		public NativeHandle (IntPtr handle)
+		{
+			this.handle = handle;
+		}
+
+		public static bool operator == (NativeHandle left, IntPtr right)
+		{
+			return left.handle == right;
+		}
+
+		public static bool operator == (NativeHandle left, NativeHandle right)
+		{
+			return left.handle == right.handle;
+		}
+
+		public static bool operator == (IntPtr left, NativeHandle right)
+		{
+			return left == right.Handle;
+		}
+
+		public static bool operator != (NativeHandle left, IntPtr right)
+		{
+			return left.handle != right;
+		}
+
+		public static bool operator != (IntPtr left, NativeHandle right)
+		{
+			return left != right.Handle;
+		}
+
+		public static bool operator != (NativeHandle left, NativeHandle right)
+		{
+			return left.handle != right.Handle;
+		}
+
+		// Should this be made explicit? The JIT seems to optimize conversions away to
+		// treat everything as a plain IntPtr, so I'm not sure there's any reason
+		// to not keep it implicit.
+		public static implicit operator IntPtr (NativeHandle value)
+		{
+			return value.Handle;
+		}
+
+		// Should this be made explicit? The JIT seems to optimize conversions away to
+		// treat everything as a plain IntPtr, so I'm not sure there's any reason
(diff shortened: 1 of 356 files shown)


src/ObjCRuntime/NativeHandle.cs
```diff
@@ -0,0 +1,79 @@
+#if NET
+
+using System;
+
+namespace ObjCRuntime {
+	public readonly struct NativeHandle {
```
Therzok, line 6: IEquatable<NativeHandle>?

rolfbjarne: Good idea!

src/ObjCRuntime/NativeHandle.cs
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		public static bool operator != (NativeHandle left, NativeHandle right)
+		{
+			return left.handle != right.Handle;
+		}
+
+		// Should this be made explicit? The JIT seems to optimize conversions away to
+		// treat everything as a plain IntPtr, so I'm not sure there's any reason
+		// to not keep it implicit.
```
dalexsoto, line 55: 👍 agreed on keeping it implicit

src/ObjCRuntime/Protocol.cs
```diff
@@ -34,14 +38,21 @@ public Protocol (IntPtr handle)
 			this.handle = handle;
```
dalexsoto, line 38: Do we want to keep this one for .NET?

dalexsoto (approved):
From this first Review it looks great! ❤️

src/ObjCRuntime/Protocol.cs
rolfbjarne: IIRC I tried to remove it, ran into a problem, and had to put it back in. However, I can't remember exactly the scenario and I might be getting confused with something else, so I've pushed this change to see what CI says about it.

chamons (approved):
Talk about an massive PR of doom, great work.

Love the test and rest of the change looks good to me.

@dalexsoto - Not sure if we want to run some sort of additional manual testing on this before/after it goes it, this the definition of large blast radius PR.

rolfbjarne:
Running tests on device didn't find anything related to this PR.

PR #13360, opened 2021-11-15
[tests] Automatically run/exit tests executed from the command line.

rolfbjarne:
This is already the default for macOS apps, but this way it behaves the same
for Mac Catalyst as well.


dalexsoto (approved):

rolfbjarne:

PR #13366, opened 2021-11-15
[xtro] Run separately for .NET assemblies. Fixes #13236.

### LINKED ISSUE #13236: [xtro] Run xtro on the .NET version of our assemblies
We're currently running xtro on our legacy Xamarin assemblies.

As the API bindings diverge, we need to run them on the .NET version of those assemblies as well.

rolfbjarne:
Run xtro on the .NET assemblies, since they have a different API surface now.

This also means a new set of \*.todo and \*.ignore files (which I put in the api-annotations subdirectory).

Fixes https://github.com/xamarin/xamarin-macios/issues/13236.


rolfbjarne:
This needs a bit more work

PR #13383, opened 2021-11-17
[xtro] Skip numerous validations for deprecated API.

rolfbjarne:
There are numerous checks that don't make much sense to report for deprecated
API, so skip those. This also required updating a few .ignore and .todo files.


tests/xtro-sharpie/Helpers.cs
```diff
@@ -209,6 +209,62 @@ public static bool IsAvailable (this Decl decl)
...
+					category = (decl.DeclContext as ObjCCategoryDecl);
+					if (category != null)
+						result = category.ClassInterface.IsDeprecated (Platform);
+				}
+			}
+
+			// but right now most frameworks consider tvOS, watchOS, and catalyst like iOS unless 
+			// decorated otherwise so we must check again if we do not get a definitve answer
+			if ((result == null) && ((Platform == Platforms.tvOS) || (Platform == Platforms.watchOS) || (Platform == Platforms.MacCatalyst)))
+				result = decl.IsDeprecated (Platforms.iOS);
```
mandel-macaque, line 235: I don't understand why this logic was not moved to the other IsDeprecated method and then simply not return a bool? form there.

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -239,8 +243,10 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
 							log = !managed_unsigned_values.ContainsKey (UInt64.MaxValue);
 							managed_unsigned_values.Remove (UInt64.MaxValue);
 						}
-						if (log)
-							Log.On (framework).Add ($"!missing-enum-value! {type.Name} native value {native_unsigned_values [value]} = {value} not bound");
+						if (log) {
+							if (!native_unsigned_values [value].Deprecated && !decl.IsDeprecated ())
```
mandel-macaque, line 247: Do we need the nested if? Why not a single one like we had but with:
```suggestion
						if (log && !native_unsigned_values [value].Deprecated && !decl.IsDeprecated ())
```

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -189,15 +192,16 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
...
 					// assume, sequentially assigned (in case next `value.InitExpr` is null)
 					n++;
 				}
 
 				foreach (var value in native_signed_values.Keys) {
-					if (!managed_signed_values.ContainsKey (value))
-						Log.On (framework).Add ($"!missing-enum-value! {type.Name} native value {native_signed_values [value]} = {value} not bound");
-					else
+					if (!managed_signed_values.ContainsKey (value)) {
+						if (!native_signed_values [value].Deprecated && !decl.IsDeprecated ())
```
mandel-macaque, line 202: why not a single if?

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -131,12 +132,14 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
...
-			if (native) {
-				if (!IsNative (type))
-					Log.On (framework).Add ($"!missing-enum-native! {name}");
-			} else {
-				if (IsNative (type))
-					Log.On (framework).Add ($"!extra-enum-native! {name}");
+			if (!decl.IsDeprecated ()) {
+				if (native) {
+					if (!IsNative (type))
+						Log.On (framework).Add ($"!missing-enum-native! {name}");
```
mandel-macaque, line 138: Single if?

tests/xtro-sharpie/EnumCheck.cs
```diff
@@ -131,12 +132,14 @@ public override void VisitEnumDecl (EnumDecl decl, VisitKind visitKind)
...
+				} else {
```
mandel-macaque, line 139: could we do an else if?

mandel-macaque (changes requested):

tests/xtro-sharpie/EnumCheck.cs
rolfbjarne: This only shows up as changed because the code is re-indented (I added another if at the top). If you ignore whitespace in the diff this doesn't show up.

tests/xtro-sharpie/EnumCheck.cs
rolfbjarne: That changes the logic - we'll hit the else branch below in more cases.

tests/xtro-sharpie/Helpers.cs
rolfbjarne: This method is basically a copy of the `IsAvailable` methods above - but in any case, moving this logic inside the other IsDeprecated method and returning a plain `bool` would break the logic just above here wrt categories

tests/xtro-sharpie/EnumCheck.cs
rolfbjarne: Same: this is just re-indented code.

rolfbjarne:

mandel-macaque (approved):

PR #13386, opened 2021-11-17
Use the MANIFEST_VERSION_BAND for the manifests

mattleibow:
It appears that the package IDs for the manifests retain the main .NET version band, such as 100 and 200, and the packs use the full version of 101 or 203.

This PR just uses the version band for the manifest packages.


rolfbjarne:

rolfbjarne (approved):

PR #13402, opened 2021-11-19
[GameController] Use existing GameController types instead of OpenTK types for a few property getters for .NET.

rolfbjarne:
These all have equivalent SetX functions that already take the corresponding
GameController type, so this is just making the API consistent.

We could also make these getters + the corresponding Set* method a readwrite
property, but this way is exactly how Apple did it (both in Swift and
Objective-C).


tj-devel709 (approved):

rolfbjarne:

PR #13408, opened 2021-11-19
[xtro-sharpie] Add support for specifying the output directory.

rolfbjarne:
This makes it possible to have a different output directory for .NET assemblies.


tj-devel709 (approved):

chamons (approved):
Mono.Options all the things!

PR #13420, opened 2021-11-22
[xtro-sanity] Pass the included platforms as command line arguments.

rolfbjarne:
This avoids the need to compute them, and makes it possible to have multiple sets
of platforms.


mandel-macaque (approved):

rolfbjarne:

PR #13431, opened 2021-11-23
[Security] Add kSecUseNoAuthenticationUI for macOS.

rolfbjarne:
The kSecUseNoAuthenticationUI field is available on macOS, just deprecated,
and I don't see how having the field bound would be a breaking change like the
comment says (git history seems to indicate it's referring to the
Classic/Unified switchover - i.e. quite outdated), so add the field to macOS.


rolfbjarne:

tj-devel709 (approved):

dalexsoto (approved):

PR #13435, opened 2021-11-23
[src] Don't use deprecated API.

rolfbjarne:
Use the non-deprecated version, which will work in both legacy Xamarin and .NET.


rolfbjarne:

PR #13444, opened 2021-11-24
[src] Adjust visibility of numerous (IntPtr) and (IntPtr, bool) constructors

rolfbjarne:
* There should be no public (IntPtr) constructor, at most there should be a
  protected one for NSObject subclasses. There should be no such constructors
  for classes that just implement INativeObject.
* There should be no public (IntPtr, bool) constructor at all, they should all
  be internal.
* If changing the visibility or removing the ctor is a breaking change, then
  make it happen for .NET only.


dalexsoto (approved):

rolfbjarne:
Test failures are unrelated:

PR #13445, opened 2021-11-25
[xtro] Don't report missing availability attributes for enums representing errors.

rolfbjarne:
* We don't want availability attributes for enums that represent errors,
  because the code would end up unnecessarily complex to avoid warnings.
* We even have other tests to verify as much (cecil tests).


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/1440).

dalexsoto (approved):

PR #13449, opened 2021-11-25
[ObjCRuntime] There's no need to look up our own functions dynamically for Xamarin.Mac in .NET. Fixes #13398.

### LINKED ISSUE #13398: [.NET] Improve error message when Xamarin.Mac thinks it can't find libxammac.dylib
Ref: https://github.com/xamarin/xamarin-macios/issues/13392

The error message is confusing for .NET (we don't ship a libxammac.dylib, it's a libxamarin.a linked into the main executable).

We can probably improve the managed code as well (use a direct P/Invoke, etc.)

rolfbjarne:
We'll always be a self-contained app, with our libxamarin code linked
statically into the executable, so we can use plain P/Invokes.

Fixes https://github.com/xamarin/xamarin-macios/issues/13398.

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -1846,7 +1846,7 @@ namespace ObjCRuntime {
 		}
 
 		[EditorBrowsable (EditorBrowsableState.Never)]
-#if MONOMAC
+#if MONOMAC && !NET
 		public static void ReleaseBlockOnMainThread (IntPtr block)
 		{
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):
Makes sense for us but is this also true for VSMac now? Just worried about their extensions model

rolfbjarne:
> Makes sense for us but is this also true for VSMac now?

Yes, this is true for all our platforms in .NET, even VSMac.

PR #13451, opened 2021-11-25
[AVFoundation] Fix availability for a few enums

rolfbjarne:
None of these are deprecated, so no reason for them to not exist in .NET.

Also update the availability attributes according to current headers.


dalexsoto (approved):

rolfbjarne:

PR #13455, opened 2021-11-25
[src] Remove the Xamarin.iOS.dll reference assembly for Mac Catalyst.

rolfbjarne:
Xamarin.iOS.dll won't be compatible with Mac Catalyst in .NET 6 (because there
won't be any backwards compatibility), so we don't need the assembly that has
type forwarders to Xamarin.MacCatalyst.dll.


dalexsoto (approved):

mandel-macaque (approved):

rolfbjarne:

PR #13456, opened 2021-11-25
[introspection] Fix the ApiFrameworkTest.Simlauncher test to fail in case of errors.

rolfbjarne:
* Assert at the end of the test if there were any errors.
* Don't run this test for .NET, since we don't use a simlauncher in .NET.


rolfbjarne:

dalexsoto (approved):

PR #13461, opened 2021-11-26
[msbuild] Don't use custom logic to handle facades in .NET.

rolfbjarne:
It's not needed, and may in some cases do the wrong thing: for instance, this
logic will automatically reference any *.dll files in the project directory,
which breaks the build if those *.dll files aren't actually assemblies.


dalexsoto (approved):

rolfbjarne:
Tes failures are unrelated (https://github.com/xamarin/maccore/issues/2197).

PR #13470, opened 2021-11-30
[tests] Fix checking availability for .NET.

rolfbjarne:
Change the logic to detect if an API is available to:

* First check if there are any applicable UnavailableOSPlatform attributes,
  and only if an applicable attribute is found, then state that the API is
  unavailable (we can't ascertain that an API is available from an
  UnavailableOSPlatform attribute, only that it's unavailable).
* Once we know there are no applicable UnavailableOSPlatform attributes, we go
  on to check for applicable SupportedOSPlatform attributes, and if one is
  found, then we can say whether the API is available or not.
* If neither attributes were found, and we're building for Mac Catalyst, then
  repeat the two above checks for iOS instead.
* If still nothing, then assume the API is available (while incorrect, it's
  how our attributes are currently implemented).

This fixes introspection showing numerous test failures on older OS versions,
because we were detecting availability wrong - we were assuming that if
there's an UnavailableOSPlatform attribute whose version didn't match the OS
version, that the API was available (test case that proves this logic is
incorrect: OS version = 1.0, API introduced in 2.0, API unavailable in 3.0
- we'd detect that OS version 1.0 < unavailable in 3.0, and say "yay, we're
not unavailable, so we must be available!").

### DIFF
--- a/tests/common/PlatformInfo.cs
+++ b/tests/common/PlatformInfo.cs
@@ -153,28 +153,28 @@ namespace Xamarin.Tests
 		}
 
 #if NET
-		public static bool IsAvailable (this IEnumerable<OSPlatformAttribute> attributes, PlatformInfo targetPlatform)
+		static IEnumerable<(OSPlatformAttribute Attribute, ApplePlatform Platform, Version Version)> ParseAttributes (IEnumerable<OSPlatformAttribute> attributes)
 		{
-			// Sort the attributes so that maccatalyst attributes are processed before ios attributes
-			// This way we can easily fall back to ios for maccatalyst
-			attributes = attributes.OrderBy (v => v.PlatformName.ToLowerInvariant ()).Reverse ();
-
 			foreach (var attr in attributes) {
 				if (!attr.TryParse (out ApplePlatform? platform, out var version))
 					continue;
 
-				if (targetPlatform.Name == ApplePlatform.MacCatalyst) {
-					if (platform != ApplePlatform.iOS && platform != ApplePlatform.MacCatalyst)
-						continue;
-				} else if (platform != targetPlatform.Name) {
-					continue;
-				}
+				yield return new (attr, platform.Value, version);
+			}
+		}
 
-				if (attr is UnsupportedOSPlatformAttribute)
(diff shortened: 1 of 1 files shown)


tj-devel709 (approved):

tests/common/PlatformInfo.cs
```diff
@@ -187,6 +187,47 @@ public static bool IsAvailable (this IEnumerable<OSPlatformAttribute> attributes
...
+			// First we check for any unsupported attributes, and only once we know that there aren't any unsupported
+			// attributes, we check for supported attributes. Otherwise we might determine that an API is available
+			// if we check the supported attribute first for OS=3.0 here:
+			//     [SupportedOSPlatform ("1.0")
+			//     [UnsupportedOSPlatform ("2.0")
+			// if we run into the SupportedOSPlatform attribute first.
+			foreach (var parsedAttr in attributes) {
+				var attr = parsedAttr.Attribute;
+				var platform = parsedAttr.Platform;
+				var version = parsedAttr.Version;
```
mandel-macaque, line 202: You can have nicer code deconstructing the attributes tuple already in the foreach, no need to add the variables:
```suggestion
			foreach (var (attr, platform, version) in attributes) {
				var attr = parsedAttr.Attribute;
				var platform = parsedAttr.Platform;
				var version = parsedAttr.Version;
```

mandel-macaque (approved):

rolfbjarne:

rolfbjarne:
Test failures are unrelated:

* fsharp/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414
* monotouch-test/watchOS 32-bits - simulator/Debug: https://github.com/xamarin/maccore/issues/581
* monotouch-test/watchOS 32-bits - simulator/Debug (LinkSdk): https://github.com/xamarin/maccore/issues/581
* monotouch-test/watchOS 32-bits - simulator/Debug (static registrar): https://github.com/xamarin/maccore/issues/581
* monotouch-test/watchOS 32-bits - simulator/Release (all optimizations): https://github.com/xamarin/maccore/issues/581

PR #13475, opened 2021-12-01
[dotnet/tools] Add support for building for ARM64 simulator architectures. Fixes #12820.

### LINKED ISSUE #12820: .NET: Add support for iossimulator-arm64 and tvossimulator-arm64
Title says it all, they are currently not enabled in workload manifests and explicitly forbidden by MSBuild checks.

rolfbjarne:
* Build our runtime for iossimulator-arm64 and tvossimulator-arm64.
* Adjust the necessary code to ensure that we're actually building for the
  simulator even if we're building for ARM64 (which previously has always been
  a device architecture).
* Build test libraries for *simulator-arm64.
* Add xharness support.
* Add ARM64 simulator variations of monotouch-test and introspection to
  xharness.

A few important points:

* The JIT isn't supported, so we're using the AOT compiler. This makes the
  build slower than for x86-64.
* We have to use 'dlsym' to resolve native functions, because Apple ships
  numerous P/Invokes that don't exist in the simulator, so we can't resolve
  the statically at build time, because we end up with native linker errors.

Fixes https://github.com/xamarin/xamarin-macios/issues/12820.


emaf (approved):

mandel-macaque (approved):

rolfbjarne:

PR #13479, opened 2021-12-01
[msbuild] Improve logic to clean up the app bundle for macOS and Mac Catalyst apps before signing.

rolfbjarne:
There can't be any files in the root directory of the app bundle for macOS and
Mac Catalyst, otherwise code signing will fail. The problem is that Mono will
create a crash report in the current directory if the process crashes, and the
current directory is the root directory of the app bundle, which means that if
running an app crashes, the next build will likely fail because of the crash
report.

We had logic to detect this and remove any crash reports, but our crash report
detection pattern wasn't good enough and let some files through. This PR
updates that pattern, and also improves the code to report warnings for any
other files in the app bundle's root directory.


chamons (approved):
Love the test inclusion.

mandel-macaque (approved):

rolfbjarne:

tj-devel709 (approved):

dalexsoto (approved):

mauroa (approved):

emaf (approved):

PR #13487, opened 2021-12-02
[monotouch-test] Ignore a few tests in non-ARM64 simulators.

rolfbjarne:
Some tests fail when running on an M1, but in a x64_86 mode, so just ignore
those unless we're running on ARM64 (this will currently exclude them on
x86_64 hardware too, but that'll eventually not be a problem anymore when
there's no more x86_64 hardware, and just checking for ARM64 is easier than
checking for x86_64 mode on an ARM64 CPU).

### DIFF
--- a/tests/common/TestRuntime.cs
+++ b/tests/common/TestRuntime.cs
@@ -154,6 +154,16 @@ partial class TestRuntime
 #endif
 	}
 
+	public static void AssertIfSimulatorThenARM64 ()
+	{
+#if !__MACOS__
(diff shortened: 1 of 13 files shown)


mandel-macaque (approved):

mandel-macaque:
The TestRuntime class has more and more useful methods!  Time to do a nuget! ;)

dalexsoto (approved):

mandel-macaque:
Unrelated issue: https://github.com/xamarin/maccore/issues/2414

chamons (approved):

PR #13490, opened 2021-12-03
[src] Remove System.nint and System.nuint from .NET. Fixes #10508.

### LINKED ISSUE #10508: Deprecate and remove Xamarin types which conflict with predefined .NET types
As part of .NET 6 migration existing types that are not compatible or serve for exactly the same purpose should be removed or at least deprecated to allow Xamarin SDK to work with all .NET libraries.

### nint/nuint types

These types in Xamarin have a bit different semantic than in C# and I think the best approach would be to remove them from public API. I didn't try what happens when C# compiler see them defined as types but they should certainly not be prefered types over C# built-in version.

### CLong/CULong/NFloat types

The existing types which serve exactly the same needs should be deprecated to allow easier sharing for .NET libraries.

---

Time estimate: depends on what we decide to do. Worst case is probably ~4 weeks to remove the Xamarin types and deal with the fallout, best case is no time at all if we decide to do nothing.

rolfbjarne:
* Remove System.nint and System.nuint from .NET
* Add support for C#'s nint/nuint types to the generator.
* Accept IntPtr/UIntPtr as target types for BindAs attributes for NSNumber conversions.
* Fix a few APIs to take/return NativeHandle instead of IntPtr.

Fixes https://github.com/xamarin/xamarin-macios/issues/10508.

This PR is probably best reviewed commit-by-commit.


mandel-macaque (approved):

chamons (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated

* monotouch-test/Mac [dotnet]/Debug (static registrar) [dotnet]: https://github.com/xamarin/maccore/issues/2527
* link sdk/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414

PR #13498, opened 2021-12-06
Compare the `TestExecutingResult.TimedOut` flag only

premun:
This is in reaction to https://github.com/dotnet/xharness/pull/786 which adds a new sub-result to preserve behavior when the bump happens.
This is the only occurrence where we're comparing `TimedOut` via equality.


mandel-macaque (approved):

dalexsoto (approved):

mandel-macaque:

Issue #13516, opened 2021-12-09
[tests] Cecil tests failures


rolfbjarne:
Test failures (from https://github.com/xamarin/xamarin-macios/pull/13504#issuecomment-989854314):

* Cecil.Tests.ConstructorTest.INativeObjectIntPtrConstructorDoesNotOwnHandle(iOS): System.IO.DirectoryNotFoundException : Could not find a part of the path "/Users/builder/azdo/_work/1/s/xamarin-macios/_build/Microsoft.iOS.Runtime.'ios-arm/runtimes/'ios-arm/lib/net6.0/Xamarin.iOS.dll".
* Cecil.Tests.ConstructorTest.INativeObjectIntPtrConstructorDoesNotOwnHandle(TVOS): System.IO.DirectoryNotFoundException : Could not find a part of the path "/Users/builder/azdo/_work/1/s/xamarin-macios/_build/Microsoft.tvOS.Runtime.'tvos-arm64/runtimes/'tvos-arm64/lib/net6.0/Xamarin.TVOS.dll".
* Cecil.Tests.ConstructorTest.INativeObjectIntPtrConstructorDoesNotOwnHandle(MacCatalyst): System.IO.DirectoryNotFoundException : Could not find a part of the path "/Users/builder/azdo/_work/1/s/xamarin-macios/_build/Microsoft.MacCatalyst.Runtime.'maccatalyst-x64/runtimes/'maccatalyst-x64/lib/net6.0/Xamarin.MacCatalyst.dll".
* Cecil.Tests.ConstructorTest.INativeObjectIntPtrConstructorDoesNotOwnHandle(MacOSX): System.IO.DirectoryNotFoundException : Could not find a part of the path "/Users/builder/azdo/_work/1/s/xamarin-macios/_build/Microsoft.macOS.Runtime.'osx-x64/runtimes/'osx-x64/lib/net6.0/Xamarin.Mac.dll".

Looks like this is a result of https://github.com/xamarin/xamarin-macios/pull/13511.

CC @mandel-macaque

PR #13525, opened 2021-12-09
[src] Tweak makefile to try to avoid confusing make.

rolfbjarne:
Make can get confused if a rule doesn't update the timestamp of a target file,
so make sure the target file's timestamp is updated in a few cases.

This hopefully fixes an infinite loop inside parallel make I've been running into.


mandel-macaque (approved):

rolfbjarne:
Test failures are unrelated:

* Cecil-based tests/NUnit: #13516

PR #13536, opened 2021-12-10
[src] Modify some logic for Mac Catalyst that tests the current runtime architecture (Runtime.Arch) to behave like macOS instead (which doesn't check Runtime.Arch).

rolfbjarne:


tj-devel709 (approved):

mandel-macaque (approved):

rolfbjarne:
Test failure is unrelated (#13547).

Issue #13531, opened 2021-12-10
Terminating app due to uncaught exception 'System.ArgumentException', reason: 'Object of type 'Foundation.NSUrlSessionTask' cannot be converted to type 'Foundation.NSUrlSessionDataTask'.'


rolfbjarne:
There seems to be a random test failure that goes like this:

    MonoTests.System.Net.Http.MessageHandlerTest.AcceptSslCertificatesServicePointManager
        [PASS] AcceptSslCertificatesServicePointManager(System.Net.Http.HttpClientHandler)
        2021-12-09 13:40:07.861 monotouchtest[83307:12887188] *** Terminating app due to uncaught exception 'System.ArgumentException', reason: 'Object of type 'Foundation.NSUrlSessionTask' cannot be converted to type 'Foundation.NSUrlSessionDataTask'.'
        *** First throw call stack:
        (
        	0   CoreFoundation                      0x00007fff2064a1db __exceptionPreprocess + 242
        	1   libobjc.A.dylib                     0x00007fff20383d92 objc_exception_throw + 48
        	2   monotouchtest                       0x00000001092b12c2 xamarin_process_managed_exception + 866
        	3   monotouchtest                       0x000000010943e335 _ZL32native_to_managed_trampoline_637P11objc_objectP13objc_selectorPP11_MonoObjectS0_S0_S0_S0_j + 1621
        	4   monotouchtest                       0x000000010943dcd0 -[Foundation_NSUrlSessionHandler_NSUrlSessionHandlerDelegate URLSession:dataTask:willCacheResponse:completionHandler:] + 80
        	5   CFNetwork                           0x00007fff24a23c22 CFHTTPCookieStorageUnscheduleFromRunLoop + 162223
        	6   libdispatch.dylib                   0x00007fff2032d623 _dispatch_call_block_and_release + 12
        	7   libdispatch.dylib                   0x00007fff2032e806 _dispatch_client_callout + 8
        	8   libdispatch.dylib                   0x00007fff203345ea _dispatch_lane_serial_drain + 606
        	9   libdispatch.dylib                   0x00007fff203350e0 _dispatch_lane_invoke + 417
        	10  libdispatch.dylib                   0x00007fff2033ec0d _dispatch_workloop_worker_thread + 811
        	11  libsystem_pthread.dylib             0x00007fff204d545d _pthread_wqthread + 314
        	12  libsystem_pthread.dylib             0x00007fff204d442f start_wqthread + 15
        )

[...]

PR #13541, opened 2021-12-13
[dotnet] Pass -dead_strip to the native linker when we can.

rolfbjarne:
Pass -dead_strip to the native linker like we do for legacy Xamarin:

* If there are no custom linker arguments.
* If all third-party bindings in the app has SmartLink = true (this doesn't
  show up in the PR, but the logic exists for legacy Xamarin and is already
  executed for .NET, the resulting Application.DeadStrip value just wasn't
  taken into account).

This shrinks the app size a bot for a Hello World app:

* Before:     10.659.731 (https://gist.github.com/rolfbjarne/b5892a5c7fb8663d38e2b69f67bce90c)
* After:       9.940.240 (https://gist.github.com/rolfbjarne/8404394180fb9971bd2f1475b747c70a)
* Difference:   -719.491 (-6.7 %)

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -431,6 +431,7 @@
 				@(_BundlerDlsym -> 'Dlsym=%(Identity)')
 				Debug=$(_BundlerDebug)
 				DeploymentTarget=$(_MinimumOSVersion)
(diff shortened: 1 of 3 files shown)


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/xamarin-macios/issues/13531).

chamons (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/xamarin-macios/issues/13531).

PR #13543, opened 2021-12-13
[dotnet] Honor 'TrimMode' to specify linker behavior if LinkMode/MtouchLink aren't set. Fixes #13518.

### LINKED ISSUE #13518: iOS always changes public TrimMode property
iOS always changes public TrimMode property which breaks documented way how to enable trimming for any .NET6 application (see https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-options#trimming-granularity).

Incorrect logic is located at https://github.com/xamarin/xamarin-macios/blob/80dec2841faea6e22ab10622c1fe95153dd023dd/dotnet/targets/Xamarin.Shared.Sdk.targets#L467-L470

This logic should probably run only if `TrimMode` is not already set by user.

rolfbjarne:
* Change dotnet-linker to only care about whether we're actually trimming anything or not.
* Allow LinkMde/MtouchLink to not be set if TrimMode is set.
* Detect if any assemblies are linked or not by checking the global TrimMode
  property + any TrimMode properties on assemblies.

Fixes https://github.com/xamarin/xamarin-macios/issues/13518.


chamons (approved):

mandel-macaque (approved):

tj-devel709 (approved):

rolfbjarne:

PR #13546, opened 2021-12-13
[msbuild] Fix merging of apps with symlinks to directories.

rolfbjarne:
Symlinks to directories are treated the same as other symlinks (as files), not
as directories. This way we don't end up re-creating a directory hierarchy
when we only have to create a symlink.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/MergeAppBundlesTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/MergeAppBundlesTaskBase.cs
@@ -282,6 +282,11 @@ namespace Xamarin.MacDev.Tasks {
 					BundlePath = fullInput,
 					SpecificSubdirectory = specificSubdirectory,
 				};
+				// Remove any files inside directories which are symlinks (we only need to process the symlink itself)
+				var symlinkDirectories = files.Where (v => PathUtils.IsSymlink (v) && Directory.Exists (v));
+				if (symlinkDirectories.Any ()) {
+					files = files.Where (file => !symlinkDirectories.Any (dir => file.StartsWith (dir + Path.DirectorySeparatorChar))).ToArray ();
+				}
 				foreach (var file in files) {
 					var relativePath = file.Substring (fullInput.Length + 1);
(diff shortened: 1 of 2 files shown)


mauroa (approved):

chamons (approved):

mandel-macaque (approved):

rolfbjarne:
Test failures are unrelated:

* monotouch-test/Mac [dotnet]/Debug (static registrar) [dotnet]: https://github.com/xamarin/xamarin-macios/issues/13531
* MTouch tests/NUnit: #13547.

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/MergeAppBundlesTaskBase.cs
```diff
@@ -282,6 +282,11 @@ bool MergeAppBundles ()
 					BundlePath = fullInput,
 					SpecificSubdirectory = specificSubdirectory,
 				};
+				// Remove any files inside directories which are symlinks (we only need to process the symlink itself)
+				var symlinkDirectories = files.Where (v => PathUtils.IsSymlink (v) && Directory.Exists (v));
+				if (symlinkDirectories.Any ()) {
+					files = files.Where (file => !symlinkDirectories.Any (dir => file.StartsWith (dir + Path.DirectorySeparatorChar))).ToArray ();
```
Therzok, line 288: Wouldn't it be better to have a hashset with a custom comparer here (avoids allocations of all the strings and closures)? In example, are these tasks run on Unix only?

rolfbjarne: Yes, this algorithm isn't the fastest. That said, it's not a very huge dataset either (usually a few hundred files at most), so I'm not sure how much of a difference a faster (and more complex) algorithm will be, and if it'll be worth the time spent implementing it.

Therzok: Those hundreds of files across many builds called all over CI bots and stuff do add up :D

PR #13550, opened 2021-12-13
[msbuild] Rework how app manifest detection works

rolfbjarne:
Our current behavior is to detect any None, BundleResource or Content item
that's named 'Info.plist', and assume that's the app manifest.

That doesn't quite work when we end up with multiple 'Info.plist' entries in any
of those item groups (one example being a framework as a BundleResource - all frameworks
have an Info.plist, and there's no good way to distinguish what the developer's intention
was).

So:

1. Implement a 'AppManifestDetectionEnabled' property to disable automatic app manifest
   detection.

2. Add a public 'AppBundleManifest' property that specifies the app manifest
   (this is just a renamed version of our previously private '_AppManifest' property).

This makes it possible for app developers to:

* Disable automatic app manifest detection.
* Still have an app manifest by specifying it manually.
* Disable automatic app manifest detection, but also not specify an app manifest
  manually (so no custom app manifest at all).

Also:

* Rename '_AppBundleManifest' to '_AppBundleManifestPath' to make it less confusing
  with the new 'AppBundleManifest' property.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -1205,7 +1205,7 @@
 		<GetMlaunchArguments
 			SessionId="$(BuildSessionId)"
 			AppBundlePath="$(_AppBundlePath)"
-			AppManifestPath="$(_AppBundleManifest)"
+			AppManifestPath="$(_AppBundleManifestPath)"
 			DeviceName="$(_DeviceName)"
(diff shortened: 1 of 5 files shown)


chamons (approved):

rolfbjarne:
Test failure is unrelated (#13547).

tj-devel709 (approved):

mauroa (approved):

emaf (approved):

PR #13562, opened 2021-12-14
[dotnet] Remove Runtime.Arch and ObjCRuntime.Arch from Mac Catalyst. Fixes #10312.

### LINKED ISSUE #10312: Mac Catalyst: should we have Runtime.Arch? If so, what should it return?
Question:

* What do we do with regards to the Runtime.Arch property for Mac Catalyst?

Possibilities:

1) Remove it.
    Con:  existing code won't compile / increases binary incompatibility with iOS.
2) Keep it, and return:
    2.1) Simulator
    2.2) Device
    2.3) A third value (`MacCatalyst`?)

---

Time estimate (for Sebastien's proposal [here](https://github.com/xamarin/xamarin-macios/issues/10312#issuecomment-747647252)): 4 weeks

rolfbjarne:
Remove Runtime.Arch and ObjCRuntime.Arch from Mac Catalyst, because they don't
apply for a Mac Catalyst app (which is neither a simulator environment, nor a
device environment).

This means that code using these APIs will have to be re-evaluated to
determine what's the correct behavior for Mac Catalyst.

Also update our tests accordingly.

Fixes https://github.com/xamarin/xamarin-macios/issues/10312.

### DIFF
--- a/dotnet/BreakingChanges.md
+++ b/dotnet/BreakingChanges.md
@@ -41,3 +41,10 @@ without this type change would be impossible to expose correctly.
 
 There are implicit conversions between `System.IntPtr` and
 `ObjCRuntime.NativeHandle`, so most code should compile without changes.
+
+## The ObjCRuntime.Arch enum and the Runtime.Arch property have been removed.
+
+These APIs are used to determine whether we're executing in the simulator or
+on a device. Neither apply to a Mac Catalyst app, so they've been removed.
+
+Any code that these APIs will have to be ported to not use these APIs.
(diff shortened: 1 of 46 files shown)


mandel-macaque:
@rolfbjarne  we need to check this by hand, device tests are not working atm until I can work on them post-vacation.

mandel-macaque (approved):

rolfbjarne:

rolfbjarne:
> @rolfbjarne we need to check this by hand, device tests are not working atm until I can work on them post-vacation.

I ran the device tests locally, and didn't run into any problems.

PR #13565, opened 2021-12-14
[VTCompressionSession] Enable nullability + a few other code updates

rolfbjarne:
* Remove unnecessary destructor and Dispose method.
* Simplify some block code.


mandel-macaque (approved):

rolfbjarne:

PR #13568, opened 2021-12-14
[VTUtilities] Enable nullability + a few other code updates.

rolfbjarne:
* Use 'GetCheckedHandle' to check for disposed objects instead of doing it
  manually.


rolfbjarne:

PR #13572, opened 2021-12-14
Fix bgen CLI command for macos net6 TFM

drasticactions:
If you try and run bgen with a target framework of `'.NETCoreApp,Version=v6.0,Profile=macos'`, you will get a cryptic error.

```
drasticactions@DrasticnoMacBook-Pro % /usr/local/share/dotnet/packs/Microsoft.macOS.Sdk/12.0.101-preview.12.551/tools/bin/bgen --target-framework=.NETCoreApp,Version=v6.0,Profile=macos
error BI1053: bgen: Internal error: unknown target framework '.NETCoreApp,Version=v6.0,Profile=macos'.
```

If you give bgen an invalid framework, it will list this TFM as support, so it should work. The reason it doesn't is because of a missing `if` clause in the CLI. If you don't add a `baselibdll`, the code falls through to where it will always throw on this TFM. We should be able to do the same thing that's being done for the other net6 TFMs and add it to the list here. Now, it should work without needing to add the BaseLib location.


src/btouch.cs
```diff
@@ -371,6 +371,8 @@ int Main3 (string [] args)
 					baselibdll = Path.Combine (GetSDKRoot (), "lib", "reference", "mobile", "Xamarin.Mac.dll");
 				else if (target_framework == TargetFramework.Xamarin_Mac_4_5_Full || target_framework == TargetFramework.Xamarin_Mac_4_5_System)
 					baselibdll = Path.Combine (GetSDKRoot (), "lib", "reference", "full", "Xamarin.Mac.dll");
+				else if (target_framework == TargetFramework.DotNet_5_0_macOS)
+					baselibdll = Path.Combine (GetSDKRoot (), "lib/mono/Xamarin.Mac/Xamarin.Mac.dll");
```
mandel-macaque, line 375: Please, do the same as with the other lines that generate the baselibdll:
```suggestion
					baselibdll = Path.Combine (GetSDKRoot (), "lib", "mono", "Xamarin.Mac", "Xamarin.Mac.dll");
```

@rolfbjarne  do we have unit tests for this?

mandel-macaque (changes requested):

mandel-macaque:
Since you did use a branch and not a fork, all tests are going to be ran

src/btouch.cs
mandel-macaque: @drasticactions you did a commit and you missed this one ;)

drasticactions: I was following the pattern for what the other TFMs had, since those had hard paths too. I can convert them.

drasticactions: Changed with https://github.com/xamarin/xamarin-macios/pull/13572/commits/d5548e267ea7cde5cacf6098c75136a79546c46e, I didn't change the other ones since they were already like that.

mandel-macaque: I'll make sure the others are correct updated, thx!

src/btouch.cs
```diff
@@ -116,6 +116,8 @@ string GetAttributeLibraryPath ()
 				return Path.Combine (GetSDKRoot (), "lib", "bgen", "Xamarin.Mac-full.BindingAttributes.dll");
 			} else if (target_framework == TargetFramework.Xamarin_Mac_2_0_Mobile) {
 				return Path.Combine (GetSDKRoot (), "lib", "bgen", "Xamarin.Mac-mobile.BindingAttributes.dll");
+			} else if (target_framework == TargetFramework.DotNet_5_0_macOS) {
+				return Path.Combine (GetSDKRoot (), "lib", "bgen", "Xamarin.Mac-full.BindingAttributes.dll");
```
rolfbjarne, line 120: That's the wrong path for .NET, this is the correct one:

```suggestion
				return Path.Combine (GetSDKRoot (), "lib", "Xamarin.Apple.BindingAttributes.dll");
```

and that's the right value for all platforms, so this can go at the top of the method, something like this:

```csharp
if (IsDotNet)
    return Path.Combine (GetSDKRoot (), "lib", "Xamarin.Apple.BindingAttributes.dll");
```

rolfbjarne (changes requested):

rolfbjarne (approved):

rolfbjarne:

rolfbjarne:
@mandel-macaque can you re-review?

mandel-macaque (approved):

mandel-macaque:
@rolfbjarne  lets merge if green

Issue #13087, opened 2021-10-22
[Announcement] Breaking changes in .NET 6 for iOS, tvOS and macOS


rolfbjarne:
In order to fully support the new native types in C# (nint, nuint), we’ve unfortunately realized that we can’t keep compatibility with existing code or assemblies built for Xamarin.iOS or Xamarin.Mac. 

The consequences are as follows when upgrading a project to .NET 6: 
 
1. All code must be recompiled to support .NET 6. Existing assemblies (such as NuGets built for the old TargetFrameworkIdentifier xamarinios10) won’t work and won’t be supported. Non-xamarin specific assets for net4.x, netstandard, netcoreapp, net5.0+, etc. will work fine however. 
2. Existing code might need to be modified. We’ll post a more detailed document with examples of code changes that might be required at a later point. 

In addition, we’re taking the opportunity to fix a few mistakes we’ve made in the past, that we’ve been unable able to address because they were breaking changes. 

These are the main changes: 

1. Removal of the [System.nint](https://docs.microsoft.com/en-us/dotnet/api/system.nint?view=xamarin-ios-sdk-12) and [System.nuint](https://docs.microsoft.com/en-us/dotnet/api/system.nuint?view=xamarin-ios-sdk-12) types. We’re going to use the `nint` and `nuint` types [as defined by C#](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nint-nuint) (they are [System.IntPtr](https://docs.microsoft.com/en-us/dotnet/api/system.intptr?view=net-5.0) and [System.UIntPtr](https://docs.microsoft.com/en-us/dotnet/api/system.uintptr?view=net-5.0) under the hood). 
2. Removal of the [System.nfloat](https://docs.microsoft.com/en-us/dotnet/api/system.nfloat?view=xamarin-ios-sdk-12) type. We're going to use the [System.Runtime.InteropServices.NFloat](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nfloat?view=net-6.0) type instead. ~For the [System.nfloat](https://docs.microsoft.com/en-us/dotnet/api/system.nfloat?view=xamarin-ios-sdk-12) type we’re probably going to keep the type [1], but move it to a different namespace.~
[...]

PR #2070, opened 2017-05-08
Fix CMSampleBuffer owns: bug introduced in commit e879678.

RogerHardiman:
Note - I'm still working through the process of building and testing, but wanted to get the PR in for this tiny fix that was introduced with a change in Sept 2016.

PR #13589, opened 2021-12-16
[VTCompressionSession] Fix broken Create overloads for .NET.

rolfbjarne:
Once upon a time there was a single VTCompressionSession.Create method, which
was [driving users insane][1] - they had to manually call CFRetain to avoid
crashes! What an abomination!!

Insane users are clearly not happy users, and we wanted happy users, so time
and effort went into creating a solution: a new Create overload was devised
and [implemented][1], taking extreme care to not break our brave and insane
existing users who had to manually call CFRetain. Because the fix would break
existing users - the now extraneous CFRetain would mean that their apps would
leak memory. *A lot* of it! That's bad, so we decided to make sure that didn't
happen.

Of course, dear old Murhpy wanted a say, so the new Create overload didn't do
as intended. In fact, it had the same insane behavior the old Create overload
had! Ops.

But Murphy decided to have even more fun: the changes were so buggy, that they
in fact fixed the old Create overload! Which from now on wouldn't require the
horrendous manual CFRetain calls... and effectively introducing the leak the
fix was trying so hard to not introduce.

Oh dear Murphy.

Of course he had another trick up his sleeve: in our extreme efforts to help
our users, we added an Obsolete attribute that would tell people to use the
new Create overload.

Let that sink in for a moment: we had an Obsolete attribute on a function that
was (now) perfectly fine, telling users to use a function that was broken.

To get the correct behavior, users would now have to to remove their manual
CFRetain calls, and ignore the obsolete warning on the old (and correct)
Create overload which told users to use the new (and buggy) Create overload.

In other words: still insanity, just a slightly different flavor.

Murphy had a field day!

Time went by, and eventually a sane enough user [reported the insanity to
us][2]. Even better: the user actually provided a fix! Truly, we have some
amazing users.

Unfortunately, the user didn't have access to our code history, and thus was
obviously not able to see the whole picture, and the fix ended up being
incorrect.

Unrelated lesson learned: don't forget your history, otherwise you'll end up
repeating mistakes from the past.

So now came the problem: how to fix all the APIs? In a way that didn't make
our users' existing apps just suddenly start crashing or leaking?

There really was no way, so nothing really happened for quite a while.

Then, an opportunity presented itself: we'd be able to do [widespread breaking
changes][3].

So, hoping that Murphy stays away this sunny winter day, I'm changing both the
new and the old Create overloads to do the right thing. But only in .NET,
where we can do breaking changes! Or at least that's my intention. I've tried
to stave off our dear old friend by adding his arch enemy: unit tests. Which,
of course, Murphy couldn't stay away from, but it seems adding a few
Thread.Sleep calls makes him bored enough to stay away. Hopefully for good...

[1]: https://github.com/xamarin/maccore/commit/66c50b9a176c6edfa8e633bfd9b54831ebddc9e5
[2]: https://github.com/xamarin/xamarin-macios/pull/2070
[3]: https://github.com/xamarin/xamarin-macios/issues/13087

### DIFF
--- a/src/VideoToolbox/VTCompressionSession.cs
+++ b/src/VideoToolbox/VTCompressionSession.cs
@@ -86,19 +86,16 @@ namespace VideoToolbox {
 			}
 		}
 
-#if NET
-		[UnmanagedCallersOnly]
-#else
+#if !NET
 #if !MONOMAC
 		[MonoPInvokeCallback (typeof (CompressionOutputCallback))]
-#endif
 #endif
 		static void CompressionCallback (IntPtr outputCallbackClosure, IntPtr sourceFrame, VTStatus status, VTEncodeInfoFlags infoFlags, IntPtr cmSampleBufferPtr)
 		{
 			CompressionCallback (outputCallbackClosure, sourceFrame, status, infoFlags, cmSampleBufferPtr, true);
 		}
+#endif // !NET
 
-		[Obsolete ("This overload requires that the provided compressionOutputCallback manually CFRetain the passed CMSampleBuffer, use Create(int,int,CMVideoCodecType,VTCompressionOutputCallback,VTVideoEncoderSpecification,CVPixelBufferAttributes) variant instead which does not have that requirement.")]
 		public static VTCompressionSession? Create (int width, int height, CMVideoCodecType codecType,
 			VTCompressionOutputCallback compressionOutputCallback,
 			VTVideoEncoderSpecification? encoderSpecification = null, // hardware acceleration is default behavior on iOS. no opt-in required.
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):
This was definitely one of those histories that you can tell you grandsons 😄

chamons (approved):
We really should remove Murphy's commit privs, what a jerk.

tj-devel709:
Me looking for who Murphy is

chamons:
https://en.wikipedia.org/wiki/Murphy%27s_law

See also https://en.wikipedia.org/wiki/Finagle%27s_law

rolfbjarne:

PR #13598, opened 2021-12-17
[dotnet] Don't add FileNativeReferences to the main libraries to link with. Fixes #13503.

### LINKED ISSUE #13503: _LinkNativeExecutable target fails running net6.0-ios for simulator and device with native reference, works with older Xamarin.iOS
_From @tmarkovski on Sat, 04 Dec 2021 22:48:49 GMT_

### Description

I'm trying to run a basic `dotnet new ios` app with a native library linked. The build fails at target `_LinkNativeExecutable` with the error

```
ld: 17954 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The same library works just fine with Xamarin.iOS target using stable VS for Mac.
The same library, compiled for MacCatalyst target works well with net6.0-maccatalyst, identical code as ios.

My test projects are in this repo, which also includes the static libraries.
https://github.com/trinsic-id/okapi/tree/test-ios/example

Running `dotnet run` in the `IosApp` should attempt to run the app for simulator and produce the above error.
Running the other two projects (`MacCatalystApp` and `XamarinIosApp`) succeeds - the native library is linked correctly and invoked on app start.

The libraries only support `x64` architectures, as I'm running intel mac. Check with `lipo -info <library>.a`

### Reproduction Steps

Clone this repo at the specified branch 

https://github.com/trinsic-id/okapi/tree/test-ios/example

Run `dotnet run` from the `IosApp` folder

### Expected behavior

App should be run and static library linked correctly

### Actual behavior

Linker error 

```
ld: 17954 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Regression?

Yes. The same library works well with Xamarin.IOS target. See project under `XamarionIosApp` in the linked repo.

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

_Copied from original issue dotnet/runtime#62410_

rolfbjarne:
Don't add FileNativeReferences to the main libraries to link with, because we
pass that list of main libraries to the LinkNativeCodeTask, and we're already
passing the FileNativeReferences for a different task parameter.

This means that we end up adding the file native reference twice to the linker
arguments, and that's wasteful. It can also cause problems if those linker
arguments aren't always computed in the same way (once as a relative path,
once as an absolute path for instance).

Fixes https://github.com/xamarin/xamarin-macios/issues/13503.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -1062,8 +1062,6 @@
 			<_XamarinMainLibraries Include="$(_XamarinNativeLibraryDirectory)/$(_LibXamarinName)" />
 			<!-- Link with the libraries shipped with the mono runtime pack -->
 			<_XamarinMainLibraries Include="@(_MonoLibrary)" />
-			<!-- Link with static libraries from NativeReference items ((xc)frameworks are handled in the _ComputeFrameworkFilesToPublish target, and dynamic libraries are handled in the _ComputeDynamicLibrariesToPublish target) -->
-			<_XamarinMainLibraries Include="@(_FileNativeReference)" />
 
 			<!-- The frameworks we need to link with (both weakly and normally) -->
 			<_NativeExecutableFrameworks Include="@(_LinkerFrameworks)" />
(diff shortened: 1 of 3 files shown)


filipnavara (approved):

chamons (approved):
❤️ the test.

PR #13599, opened 2021-12-17
[Foundation] Make NSStream.[Get|Set]Property abstract in .NET.

rolfbjarne:
These methods must be overridden, otherwise the app might crash, so making
them abstract forces developers to do the right thing.

Ref: https://bugzilla.xamarin.com/show_bug.cgi?id=50891


rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2443).

PR #13629, opened 2021-12-22
[hotrestart] Fixed MSBuild outputs when using $(AppBundleManifest)

mauroa:
We only want the file name of the app manifest path to define the outputs because the app manifest will be always located at the app bundle root level.
For MAUI, the app manifest source file could be located on a sub-folder (like \Platforms\iOS\Info.plist) so if we take the full value of the property we will end up looking the manifest in the wrong place inside the app bundle, causing the inputs/outputs checks to fail and forcing many tasks to always run (like code sign), also causing unwanted behaviors like breaking incremental builds and incremental deployments


chamons (approved):

emaf (approved):

rolfbjarne (approved):
> because the app manifest will be always located at the app bundle root level

FYI that's not true for Mac and Mac Catalyst (but it is true for iOS, tvOS and watchOS).

I don't think we support neither from Windows right now, so unless I'm wrong this should not be an issue.

mauroa:
> > because the app manifest will be always located at the app bundle root level
> 
> FYI that's not true for Mac and Mac Catalyst (but it is true for iOS, tvOS and watchOS).
> 
> I don't think we support neither from Windows right now, so unless I'm wrong this should not be an issue.

Thanks for the clarification. Right, Hot Restart is only supported for iOS so the assumption should be enough

mauroa:
@rolfbjarne can we merge this PR or should we re-run all the checks?

dalexsoto:
@mauroa Rolf is still out, I have not seen that failure before so just to be on the safe side I re-merged main into your PR

PR #13631, opened 2021-12-22
[VideoToolbox] Add a VTCompressionSession.EncodeFrame overload that takes a CVImageBuffer instead of a handle to one.

rolfbjarne:
This makes the API nicer to use.


rolfbjarne:

PR #13643, opened 2021-12-30
[main] Add Xcode 13.2 Support

dalexsoto:


src/CallKit/CXCompat.cs
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	[Obsolete (Constants.UnavailableOnMacOS)]
+	[Native]
+	public enum CXErrorCodeIncomingCallError : long {
+		Unknown = 0,
+		Unentitled = 1,
+		CallUuidAlreadyExists = 2,
+		FilteredByDoNotDisturb = 3,
+		FilteredByBlockList = 4
```
mandel-macaque, line 400: Not, missing ,

mandel-macaque (approved):
Minor comment

src/CallKit/CXCompat.cs
dalexsoto: This is a back port so I'll fix it in another PR

PR #13645, opened 2022-01-04
Replaced DotNetZip with System.IO.Compression.ZipFile in Xamarin.iOS.…

mauroa:
…Tasks.Windows

### DIFF
--- a/msbuild/Xamarin.iOS.Tasks.Windows/Tasks/UnpackHotRestartFrameworks.cs
+++ b/msbuild/Xamarin.iOS.Tasks.Windows/Tasks/UnpackHotRestartFrameworks.cs
@@ -1,9 +1,9 @@
-using Ionic.Zip;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 using Mono.Cecil;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Compression;
 using System.Linq;
 
 namespace Xamarin.iOS.HotRestart.Tasks {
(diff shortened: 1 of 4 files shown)


emaf (approved):

chamons (commented):
What triggered this work?

Seems reasonable, I'd like @rolfbjarne approval.

dalexsoto:
Does this touch anything done in the Mac? and if so was it tested there? I can't find it where but IIRC there was a reason (discussion) for not using what was provided by the SDK and use `DotNetZip` instead

mauroa:
@dalexsoto @chamons this is a change that only affects the Windows pack and we only use it for unzipping purposes. DotNetZip was throwing some errors when running builds programmatically from the XamarinVS extension with the new project system, so that makes the archive process to fail in VS. This is an attempt to see if unzipping with System.IO.Compression can get rid of those errors since we haven't been able to find out the root cause of the DotNetZip problems. I'm testing the Windows pack produced by this CI build to validate if everything works as expected

mandel-macaque:
@chamons  @dalexsoto  we should get that Ionic library out of or repo, this is a good move, is the only part of the project that uses it

chamons (approved):

mandel-macaque (approved):

rolfbjarne (approved):

rolfbjarne:
@mauroa did you test the PR to make sure it works? If so, feel free to merge.

mauroa:
Closing as this change will be included soon with other related changes

PR #13650, opened 2022-01-05
Ensure we remotely copy all project pdbs and not only the main app one

mauroa:
With the previous approach we were only copying the main app pdb to the Mac instead of all the involved pdb files (main app and referenced projects). This was causing the class libraries to not load the symbols and not hit any breakpoints while debugging


chamons (approved):
👍 - Thought I'd love a test confirming this, and I'd love @rolfbjarne's 👀  as well.

mauroa:
@chamons this is a change for the Windows pack only and unfortunately we are not testing remote scenarios here since we don't have the infrastructure ready (a Windows machine paired with a Mac)

mandel-macaque (approved):

dalexsoto (approved):

emaf (approved):

PR #13653, opened 2022-01-06
[CI] Fix mac tests yaml error

mandel-macaque:
The fact that '|' is missing results in a posh error from a cmdlet complaining.


chamons (approved):

tj-devel709 (approved):

dalexsoto (approved):
that always gets us

PR #13656, opened 2022-01-06
Generate Software Bill of Materials (SBOM) manifest

mjbond-msft:
Related work item: VS #[1458467](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1458467)

Per Executive Order (EO) produce a Software Bill of Materials (SBOM) capturing the produced package files using a dedicated task
https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/ado-sbom-generator

SBOM manifest generation only runs as part of a CI build since the steps are integrated with the `Upload packages to Azure`  job

As a result of this change you will find an artifact named `sbom` attached to each build.  Within that artifact is a `manifest.json` file under a `_manifest` directory capturing all of the files that constitute the `Software Bill of Materials` 

The `sbom` steps capture the *.nupkg, *.pkg and *.msi package files published by the build

Test build having the `Upload packages to Azure` step enabled
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5720455&view=results


rolfbjarne (commented):
I can't see any additional artifacts here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5674947&view=artifacts&pathAsName=false&type=publishedArtifacts?

Also, do we want to capture the nugets we're going to start shipping soon?

mandel-macaque:
@rolfbjarne  @mjbond-msft you don't see the artefacts because those pifs are not build on PRs as per the condition:

condition: and(succeeded(), contains (dependencies.build.outputs['configuration.BuildPkgs'], 'True'), not(contains (variables['Build.DefinitionName'], 'private'))) #

You can see that the estire job is skipped. I don't believe we need to create the sbom for PRS, right?

mjbond-msft:
> You can see that the entire job is skipped. I don't believe we need to create the sbom for PRS, right?

Indeed. The SBOM generation is technically only needed for CI builds to capture shipping artifacts. However, it would be good to have it for PRs too so that the task and shared template it depends on get run (and verified) for each PR build (i.e., verify the SBOM step continually works and doesn't present any surprises at the CI). But given it's integrated with the `Upload packages to Azure` job it is bound by the `Upload packages` job condition

mandel-macaque:
yes, I don't thing we want to do the long step of the pkgs, but we can reconsider. We should catch the error in CI ideally

rolfbjarne (approved):

mandel-macaque (approved):

PR #13721, opened 2022-01-14
[msbuild] Touch all files when copying an app extension into the app.

rolfbjarne:
Touch all files when copying an app extension into the app.

This fixes an issue where we'd in some cases run into this scenario:

* Build once (successfully), and sign the app (and app extension), including all dylibs.
* We build again, copy the (unsigned) app extension into the app again, but
  not sign dylibs because the timestamp went back into the past.

### DIFF
--- a/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/DittoTaskBase.cs
+++ b/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/DittoTaskBase.cs
@@ -18,6 +18,7 @@ namespace Xamarin.MacDev.Tasks
 		[Output]
 		public ITaskItem? Destination { get; set; }
 
+		public bool TouchDestinationFiles { get; set; }
 		#endregion
 
 		protected override string ToolName {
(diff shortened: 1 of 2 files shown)


chamons (approved):

tj-devel709 (approved):
Oh awesome, I had to do this because the symlink directory content was stale and MSBuild was not picking up on it as well. I didn't know there was a method we could call that could touch files!

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/maccore/issues/2539).

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/DittoTaskBase.cs
```diff
@@ -44,6 +45,20 @@ protected override string GenerateCommandLineCommands ()
 			return args.ToString ();
 		}
 
+		public override bool Execute ()
+		{
+			if (!base.Execute ())
+				return false;
+
+			if (TouchDestinationFiles) {
+				foreach (var file in Directory.EnumerateFiles (Destination!.ItemSpec, "*", SearchOption.AllDirectories)) {
```
emaf, line 54: @rolfbjarne This won't touch files on Windows because it'll be executed remotely, but I'm not 100% if the files you're trying to update here exist on Windows though. If those are part of a MSBuild input/output check most probably they do exist on Windows.

rolfbjarne: The "is file up-to-date" check that runs into this scenario is in our own code:

https://github.com/xamarin/xamarin-macios/blob/a9c2caa5710a76783ac82c0a589020477e7bc9c9/msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs#L87-L101

So I don't think we need to replicate this on Windows.

emaf: Oh, awesome! Yep, in that case we don't need this to happen on Windows :)

PR #13722, opened 2022-01-14
[tests] Only test recording callbacks if we have a microphone.

rolfbjarne:
Try to fix https://github.com/xamarin/maccore/issues/2539 by checking if we
have a microphone before we try to test recording callbacks.


tj-devel709 (approved):

rolfbjarne:
Closing in favor of #13730.

PR #13727, opened 2022-01-14
[Intents] Remove Intents types from macOS in .NET that were added by mistake.

rolfbjarne:


src/intents.cs
```diff
@@ -373,7 +373,7 @@ public enum INMessageAttribute : long {
 		Played,
 	}
 
-#if XAMCORE_4_0
+#if NEt
```
tj-devel709, line 376:
```suggestion
#if NET
```

rolfbjarne: Good catch!

mandel-macaque (approved):

dalexsoto (approved):

PR #13747, opened 2022-01-16
[uikit] Use a code-first approach to app-level debugging options

spouliot:
Change `CheckForIllegalCrossThreadCalls` and `CheckForEventAndDelegateMismatches`
from fields to properties.

This has two main benefits:

1. Setting them inside code (and part of the templates) makes the
features easier to discover and easier to tweak the conditions to enable
or disable them. No extra arguments hidden in project files :)

2. This allows the linker to remove the properties (when disabled) since
they are not set from the static constructor (like the current fields)

Note: this is a draft, what's missing are

1. update all templates (ideally including MAUI ones in a different PR)

2. update the linker code to detect the properties (if never set) to
remove the feature. No more settings would be needed

### DIFF
--- a/dotnet/Templates/Microsoft.iOS.Templates/ios/Main.cs
+++ b/dotnet/Templates/Microsoft.iOS.Templates/ios/Main.cs
@@ -1,5 +1,12 @@
 using iOSApp1;
 
+#if DEBUG
+// Enable runtime checks to ensure some types are only used from the main (UI) thread
+UIApplication.CheckForIllegalCrossThreadCalls = true;
+// Enable runtime checks to ensure ObjC delegates and events (using their own delegates) are not mixed
+UIApplication.CheckForEventAndDelegateMismatches = true;
+#endif
+
 // This is the main entry point of the application.
 // If you want to use a different Application Delegate class from "AppDelegate"
 // you can specify it here.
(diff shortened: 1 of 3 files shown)


dotnet/Templates/Microsoft.iOS.Templates/ios/Main.cs
```diff
@@ -1,5 +1,12 @@
 using iOSApp1;
 
+#if DEBUG
```
rolfbjarne, line 3: I think we can skip updating templates if:

* We have a private _check_for_... field we set when we generate our unmanaged main function (or somehow set this value during build time) - and set it to `true` for debug builds and `false` otherwise.
* In the linker detect if the public CheckFor... setter has been preserved. If not, we can treat the getter as a constant value (because we'll know what we set the private field to).

spouliot: Yes, removing the fields (or properties) is easy (many ways) at least if they are not set from the `.cctor`. But the savings(of two fields) are very minimal since the rest of the logic is already removed. It's mostly a nice side effect :-)

However this keeps the features (and logic) into the magic/unseen land, which is bad for feature discovery. Even when documented a feature is only useful if you know about it (and for .net it won't be the old `mtouch` additional arguments to override them). Having the templates makes them obvious, at least more than finding the new (was it added) correct msbuild syntax.

rolfbjarne:
It's too late to get this change in .NET, so I've created an issue for it for the next time we can break the API: https://github.com/xamarin/xamarin-macios/issues/15428

PR #13755, opened 2022-01-17
[Foundation] Remove compat overloads of methods taking a NSUrlSessionPendingTasks delegate in .NET.

rolfbjarne:
We don't need compat logic in .NET anymore.


dalexsoto (approved):

rolfbjarne:
Red test is unrelated (#13768).

PR #13756, opened 2022-01-17
[Foundation] Use a better version of NSItemProviderCompletionHandler for .NET.

rolfbjarne:


src/foundation.cs
```diff
@@ -140,7 +140,7 @@ namespace Foundation
 
 	delegate bool NSEnumerateErrorHandler (NSUrl url, NSError error);
 	delegate void NSMetadataQueryEnumerationCallback (NSObject result, nuint idx, ref bool stop);
-#if XAMCORE_4_0
+#if NET
 	delegate void NSItemProviderCompletionHandler (INSSecureCoding item, NSError error);
```
dalexsoto, line 144: I think we should keep the context on the item, according to [docs](https://developer.apple.com/documentation/foundation/nsitemprovidercompletionhandler?language=objc) it is the item to be loaded 
```suggestion
	delegate void NSItemProviderCompletionHandler (INSSecureCoding itemToBeLoaded, NSError error);
```

dalexsoto (approved):

rolfbjarne:
Red test is unrelated (https://github.com/xamarin/maccore/issues/2443).

src/foundation.cs
rolfbjarne: https://github.com/xamarin/xamarin-macios/pull/13766

PR #13761, opened 2022-01-17
[CFNetwork] Move the CFHost and CFHTTP* types to the CFNetwork namespace in .NET.

rolfbjarne:
These types come from the CFNetwork.framework, but for some reason they were bound inside CoreServices many years ago.

This resolves a potential issue where we might end up linking with the
CoreServices framework instead of CFNetwork framework (because we use the
namespace of types to determine which framework they belong to).

### DIFF
--- a/dotnet/BreakingChanges.md
+++ b/dotnet/BreakingChanges.md
@@ -85,3 +85,15 @@ interacting with the platform.
 In .NET, we've changed the managed SCNMatrix4 to be a column-major matrix, to
 match the native version. This means that any transposing that's currently
 done when accessing Apple APIs has to be undone.
+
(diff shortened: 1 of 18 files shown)


dalexsoto (approved):
Do you know if there are any implications for the `CFNetworkHandler` due to this change?

rolfbjarne:
> Do you know if there are any implications for the `CFNetworkHandler` due to this change?

I looked around a bit, and couldn't find anything that would break, so I think it won't affect `CFNetworkHandler`.

tj-devel709 (approved):

chamons (approved):

rolfbjarne:
Red tests are unrelated:

* introspection/Mac Modern/Debug: https://github.com/xamarin/maccore/issues/2540

PR #13782, opened 2022-01-18
[UIKit] Add missing nullability fixes #13683

### LINKED ISSUE #13683: UIApplication.KeyWindow is not marked as nullable
### Steps to Reproduce

1. Use the code `var window = UIApplication.SharedApplication.KeyWindow` with `#nullable enable`.

### Expected Behavior

`window` is marked as maybe null.

### Actual Behavior

`window` is marked as not null.

### More info

I believe it should be annotated with `[NullAllowed]`, although I couldn't actually find it in the binding source so I don't know if something more complicated is happening with it.

This is based on the Apple docs for `keyWindow`: https://developer.apple.com/documentation/uikit/uiapplication/1622924-keywindow, and also on the fact I've just had to fix a null reference bug due to accessing the key window too early.

mandel-macaque:


rolfbjarne (approved):

src/uikit.cs
```diff
@@ -2140,6 +2140,7 @@ interface UIApplication {
 
 		[Deprecated (PlatformName.iOS, 13, 0, message: "Should not be used for applications that support multiple scenes because it returns a key window across all connected scenes.")]
 		[Deprecated (PlatformName.TvOS, 13, 0, message: "Should not be used for applications that support multiple scenes because it returns a key window across all connected scenes.")]
+		[NullAllowed]
```
spouliot, line 2143: weird that xtro did not complain
and does not list that property

mandel-macaque: I agreee, I did run xtro before the commit and no complains were given 🐟

PR #13835, opened 2022-01-21
[CI] Unify all the setup steps in a template.

mandel-macaque:
Same code is used in both, unify it to make things cleaner.


rolfbjarne (approved):

tools/devops/automation/templates/common/setup.yml
```diff
@@ -0,0 +1,85 @@
+# Template that does all the bolier plate needed to build and execute tests on macOS bots
```
mjbond-msft, line 1: spelling: bolier -> boiler

mandel-macaque:
```suggestion
# Template that does all the boiler plate needed to build and execute tests on macOS bots
```

tj-devel709 (approved):

PR #13850, opened 2022-01-22
[foundation] add `NSMutableDictionary` helper to avoid creating extra `NSString` instances

spouliot:
That makes it easier on the caller, e.g. `CGContextPDF`, to avoid
creating short-lived `NSString` instance without requiring a lot
of code.

Downside: it becomes ambiguious if `NSString` is used, e.g.
`SecCertificate` for the compiler to decide between `NSObject` and
`string` overloads (but it's easy to fix and has no other impact).

### DIFF
--- a/src/Foundation/NSMutableDictionary.cs
+++ b/src/Foundation/NSMutableDictionary.cs
@@ -25,6 +25,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 
+using CoreFoundation;
 using ObjCRuntime;
 
 #if !NET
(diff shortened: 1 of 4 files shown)


chamons (approved):
I think the ambiguous case is worth the UX benefit. 

I wonder if this is "technically" and API break since code that worked could fail to compile now.

spouliot:
> I wonder if this is "technically" and API break since code that worked could fail to compile now.

@chamons yes, but it's not a binary breaking change, which is the one we really cared about.

FWIW similar source _breaks_ happens frequently when overloads are added. Most of the time we notice them with tests (failing compilation) or not at all.

src/Foundation/NSMutableDictionary.cs
```diff
@@ -335,13 +336,20 @@ public void LowlevelSetObject (IntPtr obj, IntPtr key)
...
-#else
-			ObjCRuntime.Messaging.void_objc_msgSend_IntPtr_IntPtr (this.Handle, Selector.GetHandle ("setObject:forKey:"), obj.Handle, key);
-#endif
+			if (obj is null)
+				ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (obj));
+
+			LowlevelSetObject (obj.Handle, key);
+		}
+
+		public void LowlevelSetObject (string obj, IntPtr key)
```
dalexsoto, line 345: So named parameters are easier

```suggestion
		public void LowlevelSetObject (string str, IntPtr key)
```

dalexsoto (approved):
Jus a nit

rolfbjarne (approved):

mandel-macaque (approved):

rolfbjarne:
Red test is unrelated (https://github.com/xamarin/maccore/issues/2414).

PR #13881, opened 2022-01-25
[net][foundation] Simplify `NSObjectFlag`

spouliot:
Use an enum (valuetype) instead of a class.

### DIFF
--- a/src/Foundation/NSObject2.cs
+++ b/src/Foundation/NSObject2.cs
@@ -51,11 +51,18 @@ using NativeHandle = System.IntPtr;
 #endif
 
 namespace Foundation {
+
+#if NET
+	public enum NSObjectFlag {
+		Empty,
+	}
+#else
 	public class NSObjectFlag {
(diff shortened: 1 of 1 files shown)


mandel-macaque (approved):

src/Foundation/NSObject2.cs
```diff
@@ -51,11 +51,18 @@
 #endif
 
 namespace Foundation {
+
+#if NET
+	public enum NSObjectFlag {
```
tj-devel709, line 56: Just based on the name, should this be a flag enum?

tj-devel709 (approved):

src/Foundation/NSObject2.cs
spouliot: (by design) it only has one value... 😄 
TBH I would have changed the name too - if it had not been a much larger change (and the fact I did not think of anything else)

rolfbjarne (approved):

PR #13893, opened 2022-01-26
[ModelIO] Fix the MDLVoxelIndexExtent name for .NET.

rolfbjarne:


chamons (approved):
Seems reasonable, but given all of that runtime magic I'd love a test.

rolfbjarne:
> Seems reasonable, but given all of that runtime magic I'd love a test.

We already have tests, and they pass 😄

Issue #13748, opened 2022-01-16
[net][proposal] Rename platform assemblies to Microsoft.*.dll ?


spouliot:
The breaking changes between _legacy_ and _dotnet_ profiles will lead to confusion about which version/profile is being used when looking at project sources (e.g. on GitHub) or when issues are reported.

The later (issues) are often incomplete and simply seeing `Xamarin.iOS.dll` will immediately lead to the question of "Is this a  legacy or dotnet app ?". Such questions always delay bug resolutions.

Updating the name of the assemblies to something else, like a `Microsoft` prefix, would automatically answer this question for future issues - even if it makes me sad to see Xamarin disappear even more from the code base :(

This is probably not a trivial change wrt msbuild... but now would be the best time to do it

rolfbjarne:
@emaf @mauroa would this affect the IDEs in any way?

PR #13902, opened 2022-01-26
Update assembly info for our platform assemblies. Fixes #13739.

### LINKED ISSUE #13739: [net][proposal] Add assembly versions (matching associated .net versions) into product assemblies
The use of version `0.0.0.0` for `Xamarin.iOS.dll` was to ease back-n-forth compatibility when building bindings (or libraries) as it was not easy to re-install (at least side-by-side) older versions of the products. IOW using XI 7.0 to create/maintain bindings working on 5.x was simplified (a lot).

With .net 6+ this should not be an issue anymore, each project can use it's own version of the SDK. IOW bindings could be built with an older version (to maximize compatibility across projects) while an application could target the latest version. All without having tweak (like re-install, symlinks...) the developers computers.

So it would seems to the the _right_ time to include a version number into all (net-profile) `Xamarin.*.dll` product assemblies.

rolfbjarne:
* Set the AssemblyVersion attribute for .NET (to the NuGet version).
* Use the NuGet version number for the AssemblyInformationalVersion attribute
  in .NET.
* Use 'Microsoft.*' as the assembly title and assembly product in .NET (ref:
  #13748).
* Remove the InternalsVisibleTo attribute to System.Net.Http in .NET, it's no
  longer needed.

Fixes https://github.com/xamarin/xamarin-macios/issues/13739.

### DIFF
--- a/src/AssemblyInfo.cs.in
+++ b/src/AssemblyInfo.cs.in
@@ -9,13 +9,25 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 
+#if NET
+[assembly: AssemblyInformationalVersion ("@NUGET_VERSION_NO_METADATA@; git-rev-head:@PACKAGE_HEAD_REV@; git-branch:@PACKAGE_HEAD_BRANCH@")]
+[assembly: AssemblyTitle ("Microsoft.@DOTNET_PLATFORM@")]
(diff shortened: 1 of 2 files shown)


spouliot (approved):
❤️

dalexsoto (approved):

rolfbjarne:
Red test is unrelated (https://github.com/xamarin/maccore/issues/2443).

PR #13915, opened 2022-01-27
[Foundation/ObjCRuntime] Use 'ObjCException' as the native exception type name for all platforms in .NET. Fixes #13855.

### LINKED ISSUE #13855: [net][proposal] Rename `MonoTouchException` to `NativeException`
The name is two generations old :-)

It's not very exposed so it an easy change.

```
src/Foundation/MonoTouchException.cs:   public class MonoTouchException : Exception {
src/Foundation/MonoTouchException.cs:           public MonoTouchException () : base () {
src/Foundation/MonoTouchException.cs:           public MonoTouchException (NSException exc) : base () {
src/ObjCRuntime/Runtime.cs:                     throw new MonoTouchException (new NSException (ns_exception));
src/ObjCRuntime/Runtime.cs:                     ex = new MonoTouchException (Runtime.GetNSObject<NSException> (ns_exception)!);
src/ObjCRuntime/Runtime.cs:                     var exc = obj as MonoTouchException;
src/frameworks.sources: Foundation/MonoTouchException.cs \
```

It might be a good time to move it to `ObjCRuntime` namespace too.

rolfbjarne:
* Use 'ObjCException' instead of 'MonoTouchException' as the managed exception
  type wrapping an NSException for all platforms in .NET (that was already the
  case for macOS, so no change there).
* Make the ObjCException class behave like the MonoTouchException class does.
* Move the ObjCException type to the ObjCRuntime namespace in .NET.

Fixes https://github.com/xamarin/xamarin-macios/issues/13855.

### DIFF
--- a/runtime/product.h
+++ b/runtime/product.h
@@ -31,7 +31,11 @@
   #else
     #error Unknown MONOTOUCH product for dual assembly
   #endif
-	#define PRODUCT_EXCEPTION_TYPE   "MonoTouchException"
+  #if DOTNET
+    #define PRODUCT_EXCEPTION_TYPE   "ObjCException"
+  #else
+    #define PRODUCT_EXCEPTION_TYPE   "MonoTouchException"
+  #endif
 	#ifdef __LP64__
 		#define ARCH_SUBDIR ".monotouch-64"
 	#else
--- a/tests/monotouch-test/Foundation/BundleTest.cs
+++ b/tests/monotouch-test/Foundation/BundleTest.cs
@@ -107,19 +107,31 @@ namespace MonoTouchFixtures.Foundation {
 		[Test]
 		public void PathForImageResource ()
 		{
+#if NET
+			Assert.Throws<ObjCException> (() => main.PathForImageResource ("basn3p08.png"));
+#else
 			Assert.Throws<MonoTouchException> (() => main.PathForImageResource ("basn3p08.png"));
+#endif
 		}
 
 		[Test]
 		public void PathForSoundResource ()
 		{
+#if NET
+			Assert.Throws<ObjCException> (() => main.PathForSoundResource ("basn3p08.png"));
+#else
 			Assert.Throws<MonoTouchException> (() => main.PathForSoundResource ("basn3p08.png"));
+#endif
 		}
 
 		[Test]
 		public void LoadNib ()
 		{
+#if NET
+			Assert.Throws<ObjCException> (() => NSBundle.LoadNib (String.Empty, main));
+#else
 			Assert.Throws<MonoTouchException> (() => NSBundle.LoadNib (String.Empty, main));
+#endif
 		}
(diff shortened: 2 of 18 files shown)


tests/monotouch-test/Foundation/CalendarTest.cs
```diff
@@ -8,11 +8,15 @@
...
 
+#if NET
+using PlatformException=ObjCRuntime.ObjCException;
+#else
 #if MONOMAC
 using PlatformException = Foundation.ObjCException;
 #else
 using PlatformException = Foundation.MonoTouchException;
 #endif
+#endif
```
tj-devel709, line 19:
```suggestion
#if NET
using PlatformException=ObjCRuntime.ObjCException;
#elif MONOMAC
using PlatformException = Foundation.ObjCException;
#else
using PlatformException = Foundation.MonoTouchException;
#endif
```
Just my opinion, but I think this is easier to read

tests/test-libraries/testgenerator.cs
```diff
@@ -791,11 +791,15 @@ static void WriteRegistrarTests ()
...
 using MonoTouchException=ObjCRuntime.RuntimeException;
+#if NET
+using NativeException=ObjCRuntime.ObjCException;
+#else
 #if __MACOS__
 using NativeException=Foundation.ObjCException;
 #else
 using NativeException=Foundation.MonoTouchException;
 #endif
+#endif
```
tj-devel709, line 802: Same here

tj-devel709 (approved):
Small suggestion, but looks good to me

chamons (approved):

runtime/product.h
```diff
@@ -31,7 +31,11 @@
   #else
     #error Unknown MONOTOUCH product for dual assembly
   #endif
-	#define PRODUCT_EXCEPTION_TYPE   "MonoTouchException"
+  #if DOTNET
+    #define PRODUCT_EXCEPTION_TYPE   "ObjCException"
+  #else
+    #define PRODUCT_EXCEPTION_TYPE   "MonoTouchException"
+  #endif
```
spouliot, line 38: indentation seems off here

spouliot (approved):

tests/monotouch-test/Foundation/BundleTest.cs
```diff
@@ -107,19 +107,31 @@ public void LoadNibWithOptions ()
 		[Test]
 		public void PathForImageResource ()
 		{
+#if NET
+			Assert.Throws<ObjCException> (() => main.PathForImageResource ("basn3p08.png"));
```
Therzok, line 111: `using ObjCException = MonoTouchException` at the top would reduce duplication.

rolfbjarne: Fixed

PR #13956, opened 2022-01-28
[msbuild] Rework code signing to sign everything at the end of the build.

rolfbjarne:
* Add support for specifying metadata on items that are to be codesigned to
  override any general codesign setting.

* Make the Codesign task able to sign files and directories that may depend on
  eachother.

  Implement support for ordering signing so that directories containing files
  that also must be signed are signed after those files.

  This is implemented by:

  1. Normalize all input (resolve symlinks, create full path, etc.)

  2. Sort by path length (longest to shortest paths). This way we're certain
     that if we find a directory, we'll know that we won't find any files
     later in the list from inside that directory.

  3. Group into an ordered list of buckets, where each bucket contains files
     and directories that don't depend on eachother (i.e. they can all be
     signed in parallel).

  This makes it possible to call Codesign once, listing both the app bundle
  itself, and all the individual files or directories inside that need
  signing, and the Codesign task will sign the items in an order that ensures
  parent directories are always signed after any files or directories inside.

* Finally rework code signing to sign everything with a single call to the
  Codesign task in the _CodesignAppBundle target for the executable project,
  instead of having multiple calls to the Codesign task from multiple targets
  (and projects). This makes it easier to reason about what's being signed,
  and it also makes it easier to add files to the signing process.

* Also make the Codesign task able to figure out if something needs to be
  signed, and change the _CodesignAppBundle target to not keep track of
  inputs/outputs, because it becomes quite complex (for directories, it needs
  to keep track of all the files inside that directory, and also if there's
  anything in the directory that's also being signed).


chamons (approved):
Love the work, think it will bring needed sanity to code signing nested groups.

What kind of test coverage do we have covering this? Should we get it tested on Windows before landing?

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
```diff
@@ -194,15 +241,107 @@ public override bool Execute ()
 				return true;
 
 			var codesignedFiles = new List<ITaskItem> ();
-			var resourcesToSign = Resources.Where (v => NeedsCodesign (v));
+			var resourcesToSign = Resources.Where (v => NeedsCodesign (v)).ToArray ();
```
mandel-macaque, line 244: Do we really need the array conversion? IEnumerable of the foreach is good enough.

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
```diff
@@ -194,15 +241,107 @@ public override bool Execute ()
...
+			resourcesToSign = resourcesToSign.OrderBy (v => v.ItemSpec.Length).Reverse ().ToArray ();
+
+			// Then we need to split the input into buckets, where everything in a bucket can be signed in parallel
+			// (i.e. no item in a bucket depends on any other item in the bucket being signed first).
+			// any such items must go into a different bucket. The bucket themselves are also sorted, where
+			// we have to sign the first bucket first, and so on.
+			// Since we've sorted by path length, we know that if we find a directory, we won't find any containing
+			// files from that directory later.
+			var buckets = new List<List<ITaskItem>> ();
+			for (var i = 0; i < resourcesToSign.Length; i++) {
```
mandel-macaque, line 278: foreach and we get the need of the toArray?

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
```diff
@@ -194,15 +241,107 @@ public override bool Execute ()
...
+			for (var b = 0; b < buckets.Count; b++) {
+				var bucket = buckets [b];
+				Log.LogWarning ($"    Bucket #{b + 1} contains {bucket.Count} items:");
+				foreach (var item in bucket) {
+					Log.LogWarning ($"        {item.ItemSpec}");
+				}
+			}
+#endif
+
+			for (var b = 0; b < buckets.Count; b++) {
```
mandel-macaque, line 335: Same, lets get index out, not really used elsewhere, right?

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
```diff
@@ -194,15 +241,107 @@ public override bool Execute ()
...
+					path = parent;
+
+				path = PathUtils.ResolveSymbolicLinks (path);
+				path = Path.GetFullPath (path);
+
+				res.ItemSpec = path;
+			}
+
+			// first sort all the items by path length, longest path first.
+			resourcesToSign = resourcesToSign.OrderBy (v => v.ItemSpec.Length).Reverse ().ToArray ();
```
mandel-macaque, line 269: Do we need toArray?

mandel-macaque (approved):
Some nits

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
rolfbjarne: Yes, because we need to ensure the order. An IEnumerable doesn't have a specific order.

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
rolfbjarne: The order is important, and an IEnumerable isn't guaranteed to be sorted.

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/CodesignTaskBase.cs
rolfbjarne: Same reason: the order is important, and an enumerable doesn't have a guaranteed order.

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -1869,31 +1868,20 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
-			Resources="@(_Frameworks)"
-			SigningKey="$(_FrameworkCodeSigningKey)"
-			ExtraArgs="$(CodesignExtraArgs)"
-			UseSecureTimestamp="$(UseHardenedRuntime)"
-			>
-		</Codesign>
-
-		<MakeDir SessionId="$(BuildSessionId)" Condition="'$(IsMacEnabled)' == 'true'" Directories="$(DeviceSpecificIntermediateOutputPath)codesign" />
-
-		<Touch
```
emaf, line 1889: @rolfbjarne This looks like a hack to create output files on Windows, but I see the Codesign task has an output property. Does that property contain these files?

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -1919,47 +1907,37 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
+				<CodesignKeychain>%(_AppExtensionCodesignProperties.Keychain)</CodesignKeychain>
+				<!-- CodesignResourceRules -->
+				<CodesignSigningKey>%(_AppExtensionCodesignProperties.LibrarySigningKey)</CodesignSigningKey>
+				<CodesignStampPath>$(DeviceSpecificIntermediateOutputPath)codesign\PlugIns\%(_AppExtensionCodesignProperties.Identity)</CodesignStampPath>
+				<!-- CodesignUseHardenedRuntime -->
+				<!-- CodesignUseSecureTimestamp -->
+			</_CodesignItems>
+		</ItemGroup>
 
-		<Touch
```
emaf, line 1953: Same here hack here

msbuild/Xamarin.Shared/Xamarin.Shared.targets
```diff
@@ -1919,47 +1907,37 @@ Copyright (C) 2018 Microsoft. All rights reserved.
...
-			SessionId="$(BuildSessionId)"
-			Condition="'$(IsMacEnabled)' == 'true' And ('$(_CanOutputAppBundle)' == 'true' And '%(_AppExtensionCodesignProperties.SigningKey)' != '') And Exists ('$(AppBundleDir)\..\%(_AppExtensionCodesignProperties.Identity).dSYM\Contents\Info.plist')"
-			Files="$(AppBundleDir)\..\%(_AppExtensionCodesignProperties.Identity).dSYM\Contents\Info.plist"
-		/>
-		<Touch
```
emaf, line 1958: And here

dalexsoto (approved):

msbuild/Xamarin.Shared/Xamarin.Shared.targets
rolfbjarne: This is on purpose: https://github.com/xamarin/xamarin-macios/commit/c8e3b94f069c12fc5f9cef6f82aff439917afef0

msbuild/Xamarin.Shared/Xamarin.Shared.targets
rolfbjarne: This started here: https://github.com/xamarin/xamarin-macios/commit/711bc9ab9edf157e37575e2fe2e7d2c66ea6e0be, and it doesn't look windows-related (although the commit doesn't explain anything about the Touch task).

msbuild/Xamarin.Shared/Xamarin.Shared.targets
rolfbjarne: This started here: https://github.com/xamarin/xamarin-macios/commit/b192a9c078b31e9b5899eb927ff93243467c8b6f, and it's not to support windows.

rolfbjarne (commented):
@emaf I tracked down those three Touch tasks, and none are to support Windows, they're all for timestamp tracking on the mac.

emaf (approved):

rolfbjarne:

emaf (approved):

PR #14010, opened 2022-02-01
[NET 6 Attribute Conversion] OpenGLES

tj-devel709:


src/OpenGLES/EAGLContext.cs
```diff
@@ -22,7 +22,17 @@ public EAGLContext ()
 		}
 #endif
 
-#if !NET
```
tj-devel709, line 25: Looks like `PresentRenderBuffer (nuint target, double presentationTime)` used to only be available for !NET, should this change be reverted?

mandel-macaque: /cc @rolfbjarne

rolfbjarne: What you did is correct!

rolfbjarne (approved):

chamons (approved):

tj-devel709:
**Unrelated Test Failure**
- framework-test/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414

PR #14011, opened 2022-02-01
[dotnet] Ask the linker to not mark copied assemblies if we're in a 'don't link' scenario.

rolfbjarne:
This speeds up builds significantly when the linker is disabled.

Test case: building tests/dotnet/MySimpleApp for macOS.

* Before: 37s
* After: 9s
* Difference: 26s (4x faster)

Test case: run the .NET tests

* Before: 2h55
* After: 1h43
* Difference: 1h12 (1.7x faster)

Contributes towards https://github.com/xamarin/xamarin-macios/issues/10251.
Ref: https://github.com/dotnet/linker/issues/2089


dalexsoto (approved):
wohoo!

mandel-macaque (approved):

PR #14020, opened 2022-02-02
[generator] Rework how we implement [Abstract] classes in .NET. Fixes #4969.

### LINKED ISSUE #4969: [RFC] Change what [Abstract] means when binding classes

rolfbjarne:
We have a problem that's shown up a few times, where we're given an instance
of a native type, where the closest bound managed type is a type declared as
[Abstract] in the api definition. In this case, we want to create an instance
of the [Abstract] type to wrap the native instance, but that hasn't been
possible because the managed type is abstract.

Note: this is totally fine from the OS perspective: it might have created an
instance of a private subclass we haven't bound (or it might even be a public
subclass we just haven't bound yet).

The fix is to:

* Stop making [Abstract] classes in the api definition abstract classes in the generated code.
* Make the default constructor default to "protected" visibility for [Abstract] classes.

This way we can still create instances of these types at runtime when we need
them, but they must be subclassed in order to create a managed instance of
them.

Additionally I also had to make abstract members virtual for such types
(because otherwise the type would have to be abstract as well), and instead
throw a "You_Should_Not_Call_base_In_This_MethodException" in the
corresponding method implementations.

Fixes https://github.com/xamarin/xamarin-macios/issues/4969.

### DIFF
--- a/src/generator.cs
+++ b/src/generator.cs
@@ -651,7 +651,11 @@ public class MemberInformation
 		} else if (is_static || is_category_extension || is_extension_method) {
 			mods += "static ";
 		} else if (is_abstract) {
+#if NET
+			mods += "virtual ";
+#else
 			mods += "abstract ";
+#endif
 		} else if (is_virtual_method && !is_type_sealed) {
 			mods += is_override ? "override " : "virtual ";
 		}
@@ -2477,7 +2481,11 @@ public partial class Generator : IMemberGatherer {
 				if (AttributeManager.HasAttribute<StaticAttribute> (pi))
 					need_static [t] = true;
 
+#if NET
+				var is_abstract = false;
+#else
 				bool is_abstract = AttributeManager.HasAttribute<AbstractAttribute> (pi) && pi.DeclaringType == t;
+#endif
 				
 				if (pi.CanRead){
 					MethodInfo getter = pi.GetGetMethod ();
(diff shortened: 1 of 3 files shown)


dalexsoto (approved):
Awesome ❤️

rolfbjarne:

PR #14034, opened 2022-02-03
Link against the Compression framework.

imhameed:
Mono will eventually use functions from the Compression framework to
decompress ICU data files during the runtime's initialization. Prepare
for this by linking against the compression framework.

Also see https://developer.apple.com/documentation/compression?language=objc.


dalexsoto:
Compression framework is iOS 9+ and our min version for net6 is iOS 10 so we are good there

rolfbjarne (approved):

PR #14040, opened 2022-02-03
[Foundation] Add strongly typed version of NSUndoManager.RunLoopModes for .NET.

rolfbjarne:


tj-devel709 (approved):

chamons (approved):

dalexsoto (approved):

src/Foundation/NSUndoManager.cs
```diff
@@ -17,5 +19,23 @@ public partial class NSUndoManager {
 		public virtual void SetActionName (string actionName) {
 			SetActionname (actionName);
 		}
+
+#if NET
+		public NSRunLoopMode []? RunLoopModes {
+			get {
+				var modes = WeakRunLoopModes;
+				if (modes is null)
+					return null;
```
spouliot, line 28: You can use `return Array.Empty<NSRunLoop> ();` and remove the `?` in the signature (and the need for the caller to deal with a possibly `null` return value).

src/Foundation/NSRunLoop.cs
```diff
@@ -42,12 +45,12 @@ static public partial class NSRunLoopModeExtensions {
 		// this is a less common pattern so it's not automatically generated
 		public static NSString[] GetConstants (this NSRunLoopMode[] self)
 		{
-			if (self == null)
+			if (self is null)
 				throw new ArgumentNullException (nameof (self));
```
spouliot, line 49: `ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (self));`

PR #14095, opened 2022-02-07
[UIKit] Implement the UIResponderStandardEditActions protocol.

rolfbjarne:
Implement the UIResponderStandardEditActions protocol, move the
corresponding API from UIResponder there, and make sure UIResponder implements
the new UIResponderStandardEditActions protocol (which should make this move a
non-breaking change, since the protocol is inlined in UIResponder).

### DIFF
--- a/src/uikit.cs
+++ b/src/uikit.cs
@@ -10452,7 +10452,7 @@ namespace UIKit {
 	}
 
 	[BaseType (typeof (NSObject))]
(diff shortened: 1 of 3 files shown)


chamons (approved):
I _think_ this is API compatible, I'd love @dalexsoto 👍  (which he did after I opened the tab apparently).

mandel-macaque (approved):

rolfbjarne:
I've confirmed this is api-compatible according to api-diff.

PR #14110, opened 2022-02-09
[src] Fix dependencies in makefile to not contain a trailing slash.

rolfbjarne:
This fixes a random build error, because there's no rule that knows how create
these dependencies with a trailing slash.


rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/1067).

mandel-macaque (approved):

tj-devel709 (approved):

PR #14130, opened 2022-02-11
Default VerifyDependencyInjectionOpenGenericServiceTrimmability

eerhardt:
We need to use better default values for `$(VerifyDependencyInjectionOpenGenericServiceTrimmability)`.

This feature switch ensures that DynamicallyAccessedMembers are applied correctly to open generic types used in Dependency Injection.

In the base SDK, this switch is enabled when `PublishTrimmed=true`. However, iOS apps don't set `PublishTrimmed=true` until a Target. So devs are missing out on this validation.

Conversely, in published apps, we don't want to pay the cost of doing this validation. It was showing up in startup profiles on Android. So turning the feature switch off in Release builds (builds without debugging enabled).

Port of https://github.com/xamarin/xamarin-android/commit/90f546cacc15ca6693020ac2e92aa737315ea0a4

cc @rolfbjarne @jonathanpeppers 

Note, in Android we used `PublishTrimmed` to set this property, but in iOS `PublishTrimmed` isn't set yet. So I followed the same pattern as `UseSystemResourceKeys`, which is the same scenario - when a dev is developing the app vs. a production build.

### DIFF
--- a/dotnet/targets/Xamarin.Shared.Sdk.targets
+++ b/dotnet/targets/Xamarin.Shared.Sdk.targets
@@ -138,6 +138,9 @@
 		<_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>		
 		<NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
 		<BuiltInComInteropSupport Condition="'$(BuiltInComInteropSupport)' == ''">false</BuiltInComInteropSupport>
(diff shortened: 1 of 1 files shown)


rolfbjarne (approved):

dalexsoto (approved):

eerhardt:
xamarin-macios (Test run) says it failed, but looking at https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5749521&view=results it shows:

image

mandel-macaque:
@eerhardt trust the comment not the output of the pipeline. The tests failed on watchOS. It might be a flaky test, I'll trigger a rerun.

PR #14134, opened 2022-02-14
[avfoundation] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to avfoundation.


src/AVFoundation/AVAudioFormat.cs
```diff
@@ -25,7 +27,7 @@ public partial class AVAudioFormat {
 				return true;
 			if ((object) a == null ^ (object) b == null)
 				return false;
-			return a.Equals (b);
+			return a!.Equals (b);
```
mandel-macaque, line 30: Do we really need that !you should be checking again null and get that warning out. Move the == to use is

src/AVFoundation/AVAudioSession.cs
```diff
@@ -60,16 +62,16 @@ public NSError SetActive (bool active, AVAudioSessionSetActiveOptions options)
 		}
 #endif
 
-		public NSError SetCategory (NSString theCategory)
+		public NSError? SetCategory (NSString? theCategory)
 		{
 			NSError outError;
 			
-			if (SetCategory (theCategory, out outError))
+			if (theCategory is null || SetCategory (theCategory, out outError))
```
mandel-macaque, line 69: Does the SetCategory with the out param take a nullable string? Should it?

mandel-macaque (changes requested):
minor comments.

src/AVFoundation/AVAudioFormat.cs
tj-devel709: The compiler complained here when using `is null`. I think it may have had something to do with the `^`

I decided to use the exclamation point since we first check if they are equal, then we check if one of the two are null, which I believe hits all the possibilities of it being null? @mandel-macaque

tj-devel709: @mandel-macaque  Edit: I can do this: `if (a is null ^ b is null)` and remove the object cast
When it had the (object) cast, it gave this warning/erorr: "Converting null literal or possible null value to non-nullable type."

What would an appropriate null check for this return if the objects are null and not null?

src/AVFoundation/AVAssetDownloadStorageManagementPolicy.cs
```diff
@@ -18,7 +20,12 @@ public partial class AVMutableAssetDownloadStorageManagementPolicy {
 		
 		public override AVAssetDownloadedAssetEvictionPriority Priority {
 			get { return AVAssetDownloadedAssetEvictionPriorityExtensions.GetValue (_Priority); }
-			set { _Priority = value.GetConstant (); }
+			set
+			{
+				var val = value.GetConstant ();
+				if (val is not null)
+					_Priority = val;
+			}
```
rolfbjarne, line 28: This is a change in behavior (which we don't want to do); in other places we've done this:

```suggestion
			set { _Priority = value.GetConstant ()!; }
```

src/AVFoundation/AVAudioFormat.cs
rolfbjarne: Changing the previous condition to: `if (a is null ^ b is null)` (or `if (a is null || b is null)`) looks correct.

src/AVFoundation/AVCaptureDeviceDiscoverySession.cs
```diff
@@ -19,7 +21,8 @@ public static AVCaptureDeviceDiscoverySession Create (AVCaptureDeviceType [] dev
 		{
 			var arr = new NSMutableArray ();
 			foreach (var device in deviceTypes)
-				arr.Add (device.GetConstant ());
+				if (device.GetConstant () is NSObject deviceType)
+					arr.Add (deviceType);
```
rolfbjarne, line 25: This is also a change in behavior:

```suggestion
				arr.Add (device.GetConstant ()!);
```

src/AVFoundation/AVLayerVideoGravity.cs
```diff
@@ -104,7 +108,8 @@ partial class AVPlayer
 	{
 		public AVLayerVideoGravity? ExternalPlaybackVideoGravity {
 			set {
-				WeakExternalPlaybackVideoGravity = value.HasValue ? AVPlayerLayer.EnumToKey (value.Value) : null;
+				if (value.HasValue && AVPlayerLayer.EnumToKey (value.Value) is NSString val)
+					WeakExternalPlaybackVideoGravity = val;
```
rolfbjarne, line 112:
```suggestion
				WeakExternalPlaybackVideoGravity = value.HasValue ? AVPlayerLayer.EnumToKey (value.Value) : null!;
```

src/AVFoundation/AVVideoSettings.cs
```diff
@@ -98,11 +100,12 @@ public AVVideoSettingsUncompressed (NSDictionary dictionary)
 
 		public AVVideoScalingMode? ScalingMode {
 			get {
-				return ScalingModeFromNSString (GetNSStringValue (AVVideo.ScalingModeKey));
+				var val = GetNSStringValue (AVVideo.ScalingModeKey);
+				return val is not null ? ScalingModeFromNSString (val) : null;
```
rolfbjarne, line 104: I think it's simpler to change `ScalingModeFromNSString` to take a nullable string (in which case this change would not be needed at all).

src/AVFoundation/AVVideoSettings.cs
```diff
@@ -307,11 +310,12 @@ public AVVideoSettingsCompressed (NSDictionary dictionary)
 
 		public AVVideoScalingMode? ScalingMode {
 			get {
-				return AVVideoSettingsUncompressed.ScalingModeFromNSString (GetNSStringValue (AVVideo.ScalingModeKey));
+				var val = GetNSStringValue (AVVideo.ScalingModeKey);
+				return val is not null ? AVVideoSettingsUncompressed.ScalingModeFromNSString (val) : null;
```
rolfbjarne, line 314: Same: I think it's simpler to change `ScalingModeFromNSString` to take a nullable string (in which case this change would not be needed at all).

src/AVKit/AVPlayerViewController.cs
```diff
@@ -22,7 +22,9 @@ partial class AVPlayerViewController {
 				return AVPlayerLayer.KeyToEnum (WeakVideoGravity);
 			}
 			set {
-				WeakVideoGravity = AVPlayerLayer.EnumToKey (value);
+				var val = AVPlayerLayer.EnumToKey (value);
+				if (val is not null)
+					WeakVideoGravity = val;
```
rolfbjarne, line 27:
```suggestion
				WeakVideoGravity = AVPlayerLayer.EnumToKey (value)!;
```

rolfbjarne (changes requested):

src/AVFoundation/AVAudioSession.cs
tj-devel709: After talking with you yesterday though, I see why you would say to pass it to the SetCategory method. I guess we could try adding the null arg to the selector and add in the ignore files but the headers do not say it takes a null value @mandel-macaque

tj-devel709: If I just remove the nullable parameter for 'theCategory' like Rolf suggested below then this is not an issue anymore!

src/AVFoundation/AVAudioSession.cs
```diff
@@ -60,16 +62,16 @@ public NSError SetActive (bool active, AVAudioSessionSetActiveOptions options)
 		}
 #endif
 
-		public NSError SetCategory (NSString theCategory)
+		public NSError? SetCategory (NSString? theCategory)
```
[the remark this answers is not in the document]
tj-devel709: So for this one, the `SetCategory  (AVAudioSessionCategory category)` that calls the above SetCategory method passes in the return value from CategoryToToken which can be null. Should I instead check that the return value from CategoryToToken is not null and pass that in to the above SetCategory like this?

```
public NSError? SetCategory (AVAudioSessionCategory category)
{
	var token = CategoryToToken (category);
	return token is not null ? SetCategory (token) : null;
}
```

src/AVFoundation/AVPlayerLayer.cs
```diff
@@ -22,16 +24,16 @@ public partial class AVPlayerLayer {
...
-				if (WeakPixelBufferAttributes != null) {
+				if (WeakPixelBufferAttributes is not null) {
 					var strongDict = new CVPixelBufferAttributes (WeakPixelBufferAttributes);
 					return strongDict;
 				}
 				return null;
 			}
 			set {
-				WeakPixelBufferAttributes = value != null ? value.Dictionary : null;
+				WeakPixelBufferAttributes = value is not null ? value.Dictionary : null;
```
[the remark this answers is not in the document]
tj-devel709: Much better, thank you!

src/AVFoundation/AVAudioSessionDataSourceDescription.cs
```diff
@@ -86,6 +89,9 @@ static internal NSString ToToken (AVAudioDataSourcePolarPattern p)
 		public AVAudioDataSourcePolarPattern []SupportedPolarPatterns {
 			get {
 				var x = SupportedPolarPatterns_;
+				if (x is null)
+					ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (x));
+
```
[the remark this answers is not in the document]
tj-devel709: Makes sense!

src/AVFoundation/AVAssetDownloadStorageManagementPolicy.cs
tj-devel709: @rolfbjarne  Question about this one, in this example, `_Priority` cannot be null and we are setting it equal to something that could possibly be null. Since we know that the GetConstant was not returning null in the past (because _Priority cannot be null) then it is okay that we tell the compiler we know that this value won't be null?

src/AVFoundation/AVLayerVideoGravity.cs
tj-devel709: How about `WeakExternalPlaybackVideoGravity = value.HasValue ? AVPlayerLayer.EnumToKey (value.Value)! : null!;`

src/AVFoundation/AVVideoSettings.cs
```diff
@@ -332,23 +335,23 @@ public AVVideoSettingsCompressed (NSDictionary dictionary)
...
 				var dict = GetNSDictionary (AVVideo.CompressionPropertiesKey);
-				if (dict == null)
+				if (dict is null)
 					return null;
 				return new AVVideoCodecSettings (dict);
 			}
 
 			set {
-				SetNativeValue (AVVideo.CompressionPropertiesKey, value == null ? null : value.Dictionary);
+				SetNativeValue (AVVideo.CompressionPropertiesKey, value is null ? null : value.Dictionary);
```
rolfbjarne, line 354: This is simpler:

```suggestion
				SetNativeValue (AVVideo.CompressionPropertiesKey, value?.Dictionary);
```

There are a few more below as well.

src/AVFoundation/AVAudioSession.cs
```diff
@@ -92,12 +94,13 @@ internal NSString CategoryToToken (AVAudioSessionCategory category)
...
-			return SetCategory (CategoryToToken (category));
+			var token = CategoryToToken (category);
+			return token is not null ? SetCategory (token) : null;
 		}
 
-		public NSError SetCategory (AVAudioSessionCategory category, AVAudioSessionCategoryOptions options)
+		public NSError? SetCategory (AVAudioSessionCategory category, AVAudioSessionCategoryOptions options)
 		{
 			NSError error;
 			if (SetCategory (CategoryToToken (category), options, out error))
```
rolfbjarne, line 106: If `CategoryToToken` returns a nullable value, the first argument to SetCategory must be nullable, but it isn't, so how does this compile without a warning?

src/AVFoundation/AVLayerVideoGravity.cs
rolfbjarne: Yeah, that works too.

src/AVFoundation/AVAssetDownloadStorageManagementPolicy.cs
rolfbjarne: It's ok to tell the compiler we know it won't be null, because that's how it behaved in the past, and we don't want to change the behavior.

src/AVFoundation/AVAudioSession.cs
tj-devel709: @rolfbjarne that is a great question... Yeah this is the generated method signature:
`public virtual bool SetCategory (string category, AVAudioSessionCategoryOptions options, out NSError outError)`
And CategoryToToken says it returns an `NSString?`
Although there are no compiler warnings/errors regarding this..

rolfbjarne: Actually your comment made me come up with a theory :) 

There's an implicit NSString -> string operator: 

https://github.com/xamarin/xamarin-macios/blob/1273190e5fdcea446ce4664031048961a3235e73/src/Foundation/NSString.cs#L158

and this method doesn't have nullability information (yet), which means that the compiler stops complaining about nullability problems.

I think that both here and the previous call to `SetCategory (CategoryToToken (category))`, the best is to tell the compiler to be silent: `SetCategory (CategoryToToken (category)!)` - which is also the current behavior (which is good for this PR).

rolfbjarne (approved):

tj-devel709:
After I get the green light from @mandel-macaque, I'll hit merge on this one!

src/AVFoundation/AVCaptureDeviceDiscoverySession.cs
```diff
@@ -19,7 +21,7 @@ public static AVCaptureDeviceDiscoverySession Create (AVCaptureDeviceType [] dev
 		{
 			var arr = new NSMutableArray ();
 			foreach (var device in deviceTypes)
-				arr.Add (device.GetConstant ());
+				arr.Add (device.GetConstant ()!);
```
mandel-macaque, line 24: If the contestant is null? Shall we just not add it? What is the reasoning behind the use of "!"

src/AVFoundation/AVCompat.cs
```diff
@@ -128,7 +130,7 @@ public virtual AVMetadataItem[] ChapterMetadataGroups (NSLocale forLocale, AVMet
 	partial class AVAssetTrack {
 
 		[Obsolete ("Use 'GetAssociatedTracks'.")]
```
mandel-macaque, line 132: @rolfbjarne  @tj-devel709  all this method that have not implemented exception, should we just wrap them around a !NET?

src/AVKit/AVPlayerViewController.cs
```diff
@@ -22,7 +22,7 @@ partial class AVPlayerViewController {
 				return AVPlayerLayer.KeyToEnum (WeakVideoGravity);
 			}
 			set {
-				WeakVideoGravity = AVPlayerLayer.EnumToKey (value);
+				WeakVideoGravity = AVPlayerLayer.EnumToKey (value)!;
```
mandel-macaque, line 25: Again, I dislike the use of ! here, can WeakBideoGreavity be nullable?

mandel-macaque (changes requested):

src/AVFoundation/AVAudioPlayer.cs
```diff
@@ -26,27 +26,29 @@
...
+		public static AVAudioPlayer? FromUrl (NSUrl url, out NSError? error)
 		{
 			unsafe {
 				IntPtr errhandle;
 				IntPtr ptrtohandle = (IntPtr) (&errhandle);
 
 				var ap = new AVAudioPlayer (url, ptrtohandle);
 				if (ap.Handle == IntPtr.Zero){
-					error = (NSError) Runtime.GetNSObject (errhandle);
+					error = (NSError?) Runtime.GetNSObject (errhandle);
```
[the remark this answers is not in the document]
tj-devel709: Ah didn't think about that, but this is the compiler error I get when I do this:
 `The type 'Foundation.NSError?' cannot be used as type parameter 'T' in the generic type or method 'Runtime.GetNSObject<T>(IntPtr)'. Nullability of type argument 'Foundation.NSError?' doesn't match constraint type 'Foundation.NSObject'.`

src/AVFoundation/AVLayerVideoGravity.cs
tj-devel709: @mandel-macaque This is the comment I am referring to below

src/AVFoundation/AVLayerVideoGravity.cs
```diff
@@ -104,7 +106,7 @@ partial class AVPlayer
 	{
 		public AVLayerVideoGravity? ExternalPlaybackVideoGravity {
 			set {
-				WeakExternalPlaybackVideoGravity = value.HasValue ? AVPlayerLayer.EnumToKey (value.Value) : null;
+				WeakExternalPlaybackVideoGravity = value.HasValue ? AVPlayerLayer.EnumToKey (value.Value)! : null!;
```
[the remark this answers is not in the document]
tj-devel709: The AVPlayerLayer.EnumToKey can return a nullable, but the problem is that WeakExternalPlaybackVideoGravity is coming from avfoundation.cs and cannot be nullable. Rolf mentioned using the '!' above as well that I just tagged you in

src/AVKit/AVPlayerViewController.cs
tj-devel709: Similar story as the WeakExternalPlaybackVideoGravity where it comes from avfoundation.cs

src/AVFoundation/AVCaptureDeviceDiscoverySession.cs
tj-devel709: I just tried to avoid introducing any new changes, but I don't see an issue with skipping if it is null!

src/AVFoundation/AVVideoSettings.cs
tj-devel709: Missed this change, fixing in this next commit!

tj-devel709:
I thought this was compiling locally before, but I guess I could have missed it! It definitely compiles now locally after last commit!

tj-devel709:
@mandel-macaque @rolfbjarne the [Use default value from apple docs for KeyToEnum method and correct earlier Rolf suggestion](https://github.com/xamarin/xamarin-macios/pull/14134/commits/55112b31d1d60edfb84fc25802c6ba3c05a1777d) commit adds some changes that use the default value according to Apple docs. I could have changed it around and removed the `AVLayerVideoGravity.ResizeAspect` case since it will still fall under default, but I think this was clearer for the future in case things get changed around.

src/AVFoundation/AVLayerVideoGravity.cs
```diff
@@ -67,7 +69,9 @@ static internal NSString EnumToKey (AVLayerVideoGravity vg)
 			case AVLayerVideoGravity.Resize:
 				return AVPlayerLayer.GravityResize;
 			default:
-				return null;
+				// The default value according to docs is AVLayerVideoGravity.ResizeAspect so return AVPlayerLayer.GravityResizeAspect;
```
rolfbjarne, line 72: The default value _for the VideoGravity property_ is GravityResizeAspect, but this method (EnumToKey) is being called from other places as well, so you can't assume that GravityResizeAspect will be the default value for all callers.

I think I would just change this default case to:

```cs
return null!:
```

rolfbjarne (changes requested):

tj-devel709:
Actually I'm going to close this branch and open up a new PR here as that may be nicer to not mess up previous suggestions and comments! https://github.com/xamarin/xamarin-macios/pull/14346

src/AVFoundation/AVCompat.cs
tj-devel709: Add xamcore5

PR #14135, opened 2022-02-14
[accelerate] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to accelerate.


src/Accelerate/vImageTypes.cs
```diff
@@ -272,11 +272,11 @@ public static vImageError ConvolveWithBiasARGBFFFF (ref vImageBuffer src, ref vI
 									vImageFlags flags)
 		{
 			if (kernels == null)
```
spouliot, line 274: you can also update the `== null` for `is null`
and `!= null` for `is not null`

That was not available when I wrote https://github.com/xamarin/xamarin-macios/wiki/Nullability but it should be added :)

spouliot: > Changing any == null or != null to is null and is not null

well well well :D

mandel-macaque (changes requested):
Please make the requested change from @spouliot

src/Accelerate/vImageTypes.cs
tj-devel709: I wrote down that this PR does the 'is null' and 'is not null' but I guess I lied haha. Thank you for catching my bluff

tj-devel709:
@spouliot and @mandel-macaque I did the 'is null' and 'is not null' changes as well as updated that wiki page to include it!

dalexsoto (approved):

src/Accelerate/vImageTypes.cs
```diff
@@ -271,12 +271,12 @@ public static vImageError ConvolveWithBiasARGBFFFF (ref vImageBuffer src, ref vI
 									Pixel8888 backgroundColor,
```
rolfbjarne, line 271: This file isn't in `API_SOURCES`, so according to your PR description it should get `#nullable enable` added

rolfbjarne (changes requested):

src/Accelerate/vImageTypes.cs
tj-devel709: Ah good catch, thank you!

mandel-macaque (changes requested):
Don't land this, I want confirmation that we are ok by throwing new diff exceptions when null in certain cases.

src/Accelerate/vImageTypes.cs
```diff
@@ -437,12 +439,19 @@ public static vImageError TentConvolveARGB8888 (ref vImageBuffer src, ref vImage
...
+			if (pre_bias is null)
+				ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (pre_bias));
+			else if (pre_bias.Length != 4)
 				throw new ArgumentException ("Must have four elements", "pre_bias");
-			if (post_bias != null && post_bias.Length != 4)
+
+			if (post_bias is null)
+				ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (post_bias));
+			else if (post_bias.Length != 4)
 				throw new ArgumentException ("Must have four elements", "post_bias");
```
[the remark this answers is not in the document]
chamons: We should not change behavior in a PR like this. If we decide we want to change behavior, it should be either in another PR or at least in a separate commit in this asking for special review. 

These PRs are mechanical enough, don't let non-mechanical changes hide in the diff if possible.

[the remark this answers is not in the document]
rolfbjarne: In any case this change is wrong, the P/Invoke accepts NULL for the parameter in question (which means that the managed parameter should become nullable instead).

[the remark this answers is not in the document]
tj-devel709: @mandel-macaque @chamons @rolfbjarne Reverting these changes here! Going to hop on a little call later with Manuel to make sure I understand what I did wrong up there and hopefully don't do in future PRs. Thanks for the review!

rolfbjarne (approved):

mandel-macaque (approved):

PR #14136, opened 2022-02-14
[runtime] Remove `NSLog (string format, params object?[] args)` overload

spouliot:
Instead have the callers uses string interpolation. The compiler (current
and future) is able to optimize this or, at worse, will end up with a
call to `String.Format`.


chamons (approved):

dalexsoto (approved):
❤️

spouliot:
@chamons `NSLog` is `internal` so it's not really worth mentioning in the release notes - unless it's for mentioning the contribution itself ;-)

rolfbjarne (approved):

rolfbjarne:

PR #13281, opened 2021-11-04
[Class] Enable nullability + fix a few code updates.

rolfbjarne:
* Enable nullability and fix code accordingly.
* Use 'is' and 'is not' instead of '==' and '!=' for object identity.
* Use 'nameof (parameter)' instead of string constants.
* Make the 'handle' field private.
* Remove the (IntPtr) constructor for .NET
* Make the (IntPtr, bool) constructor internal for .NET
* Remove unused 'Construct' method.

PR #14138, opened 2022-02-15
[objcruntime] Simplified earlier split of internal `Lookup` method

spouliot:
The split [1] was done to ease nullability annotations since there's no
syntax to cover this specific case with custom attributes.

However we don't need three methods for the split. Two are enough :)

[1] https://github.com/xamarin/xamarin-macios/pull/13281


dalexsoto (approved):

rolfbjarne (approved):

rolfbjarne:

PR #14140, opened 2022-02-15
[net][objcruntime] Remove `Class (IntPtr)` for .net profile

spouliot:
That was mentioned

> Remove the (IntPtr) constructor for .NET

inside https://github.com/xamarin/xamarin-macios/pull/13281
but not actually done inside that PR


rolfbjarne:
I had removed it, but then ended up putting it back in again, although right now I don't remember why. Maybe the tests will help :)

rolfbjarne (approved):

PR #14145, opened 2022-02-15
[ObjCRuntime] Keep accepting IntPtr constructors in .NET as an alternative to NativeHandle constructors. Fixes #14046.

### LINKED ISSUE #14046: .NET: Accept (IntPtr) constructors when (NativeHandle) constructors aren't found
As a part of the breaking changes in .NET, we introduced a new type, `ObjCRuntime.NativeHandle`, to represent native handles.

This meant that constructors taking taking `IntPtr handle`:

```cs
public class MyUIViewController : UIViewController {
    protected MyUIViewController (IntPtr handle)
        : base (handle)
    {
    }
}
```

will have to be ported to take `NativeHandle handle`:

```cs
public class MyUIViewController : UIViewController {
    protected MyUIViewController (NativeHandle handle)
        : base (handle)
    {
    }
}
```

The unfortunate part is that there will be no compiler warnings or errors flagging this, so users won't know to do it unless they either read the documentation (🤣) or run into the problem, googles for a while, runs into someone else who had the same problem, and applies their (probably broken) fix.

So the suggestion would be to

* Look for the `(IntPtr)` (or `(IntPtr, bool)`) constructor in .NET if the  `NativeHandle` version wasn't found.
* Show a warning
* Some time in the future maybe remove this hack/workaround.

Ref: https://github.com/xamarin/xamarin-macios/issues/13979#issuecomment-1026687922
Ref: https://github.com/xamarin/xamarin-macios/pull/14044

rolfbjarne:
As a part of the breaking changes in .NET, we introduced a new type,
`ObjCRuntime.NativeHandle`, to represent native handles.

This meant that constructors taking taking `IntPtr handle`:

```cs
public class MyUIViewController : UIViewController {
    protected MyUIViewController (IntPtr handle)
        : base (handle)
    {
    }
}
```

would have to be ported to take `NativeHandle handle`:

```cs
public class MyUIViewController : UIViewController {
    protected MyUIViewController (NativeHandle handle)
        : base (handle)
    {
    }
}
```

The unfortunate part is that there will be no compiler warnings or errors
flagging this, so users won't know to do it unless they either read the
documentation (🤣) or run into the problem, googles for a while, runs into
someone else who had the same problem, and applies their (probably broken)
fix.

So we change our logic to:

1. Look for and use an `(IntPtr)` (or `(IntPtr, bool)`) constructor in .NET if
   the `NativeHandle` version isn't found.
2. Show a warning.
3. Some time in the future maybe remove this hack/workaround.

Fixes https://github.com/xamarin/xamarin-macios/issues/14046.

### DIFF
--- a/src/ObjCRuntime/Runtime.cs
+++ b/src/ObjCRuntime/Runtime.cs
@@ -1206,7 +1206,11 @@ namespace ObjCRuntime {
 
 			var ctorArguments = new object [1];
 #if NET
-			ctorArguments [0] = new NativeHandle (ptr);
+			if (ctor.GetParameters () [0].ParameterType == typeof (IntPtr)) {
+				ctorArguments [0] = ptr;
+			} else {
+				ctorArguments [0] = new NativeHandle (ptr);
+			}
 #else
 			ctorArguments [0] = ptr;
 #endif
@@ -1232,7 +1236,11 @@ namespace ObjCRuntime {
 
 			var ctorArguments = new object [2];
 #if NET
-			ctorArguments [0] = new NativeHandle (ptr);
+			if (ctor.GetParameters () [0].ParameterType == typeof (IntPtr)) {
+				ctorArguments [0] = ptr;
+			} else {
+				ctorArguments [0] = new NativeHandle (ptr);
+			}
 #else
 			ctorArguments [0] = ptr;
 #endif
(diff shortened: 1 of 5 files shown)


dalexsoto (approved):

rolfbjarne:

PR #14149, opened 2022-02-15
[SceneKit] Remove dead code from SCNMatrix4.cs.

rolfbjarne:
This entire file is already inside a !NET condition, so no need to have
additional conditions inside the file.


dalexsoto (approved):

rolfbjarne:

PR #14150, opened 2022-02-15
[objcruntime] Inline the use of the `Class.LookupFullName` method

spouliot:
inside it's _only_ caller and remove the API.

Remove old test that was already not useful (since the method could not
be removed anyway for quite a while).


mandel-macaque (approved):

dalexsoto (approved):

mandel-macaque:
failed, but i is unrelated AFAIK

rolfbjarne:

PR #14161, opened 2022-02-16
[net][objcruntime] Remove `DisposableObject.ClearHandle`

spouliot:
It is not really needed. The only caller `SKIndex` was already clearing
the handle by calling `Dispose(bool)` which called `ClearHandle`.

If it really becomes needed then it can be reintroduced in the future.

### DIFF
--- a/src/ObjCRuntime/DisposableObject.cs
+++ b/src/ObjCRuntime/DisposableObject.cs
@@ -65,14 +65,15 @@ namespace ObjCRuntime {
 
 		protected virtual void Dispose (bool disposing)
 		{
-			ClearHandle ();
+			handle = NativeHandle.Zero;
 		}
 
+#if !NET
 		protected void ClearHandle ()
 		{
 			handle = NativeHandle.Zero;
 		}
-
+#endif
 		void InitializeHandle (NativeHandle handle, bool verify)
 		{
 #if !COREBUILD
--- a/src/SearchKit/SearchKit.cs
+++ b/src/SearchKit/SearchKit.cs
@@ -305,9 +305,8 @@ namespace SearchKit
 
 		protected override void Dispose (bool disposing)
 		{
-			if (Handle != IntPtr.Zero) {
+			if (Handle != NativeHandle.Zero) {
 				SKIndexClose (Handle);
-				ClearHandle ();
 			}
 			base.Dispose (disposing);
 		}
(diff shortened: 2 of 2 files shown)


dalexsoto (approved):

rolfbjarne (approved):

rolfbjarne:
This makes the xammac tests crash with:

```
Unhandled exception. System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at SearchKit.SKIndex.Release()
   at CoreFoundation.NativeObject.Dispose(Boolean disposing)
   at SearchKit.SKIndex.Dispose(Boolean disposing)
   at ObjCRuntime.DisposableObject.Finalize()
```

[execute-Mac-20220215_231706.txt](https://github.com/xamarin/xamarin-macios/files/8077644/execute-Mac-20220215_231706.txt)

spouliot:
@rolfbjarne so I see 3 options to solve this

1. Maybe `SKIndex` should have implemented `INativeObject` and not a subclass of `NSObject`. However

2. `ClearHandle` could stay (for `SKIndex` today) but not be used inside `NativeObject`. That way the linker can eliminate the method in all applications (except those using `SKIndex`).

2. Throwing an exception in `Retain` and `Release` is not needed (nor helpful). Why ? Both are `protected` so only a subclass of `SKIndex` (I seriously doubt it's common) or the Xamarin runtime can invoke them. Having this behaviour actually hurts the runtime and `ClearHandle` was a _fix_ for this. Any code that accept an `NativeObject` subclass now has to deal with potential exceptions on those call. Overriding the methods as "no-op" is a simpler fix.

Considering it's too late to change (at least for legacy) for option 1 then it's either 2 or 3. I much prefer the last one (3).

rolfbjarne:
@spouliot 

1. is too late for legacy, but it's not too late for .NET - we can make the SKIndex class subclass DisposableObject, and then just remove the Retain and Release overrides completely.

chamons (approved):

rolfbjarne:
Test failures are unrelated:

* introspection/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414
* link sdk/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2494
* trimmode link/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2494

PR #14162, opened 2022-02-16
[dotnet] Add support for the PublishFolderType metadata on Content and BundleResource items.

rolfbjarne:
Add support for the PublishFolderType metadata on Content and BundleResource
items, which makes it possible to change the location in the app bundle for
these items (this was possible to do before with the Link metadata), but most
importantly it also makes it possible to choose to *not* bundle these items in
the app bundle (which was not possible before with the Link metadata, nor any
other means).

At first I thought setting CopyToPublishDirectory /
CopyToOutputDirectory=Never would accomplish that, but it turns out we don't
honor those, and since we already have this behavior of ignoring
CopyToPublishDirectory / CopyToOutputDirectory in legacy Xamarin, I didn't
want to change it in .NET.

So I added support for honoring the PublishFolderType metadata instead, which
is new and we already support for other item groups. This is accomplished by
adding all Content and BundleResource items with PublishFolderType set to the
ResolvedFileToPublish item group (where we already handle any
PublishFolderType values), and then we ignore such Content and BundleResource
items in our CollectBundleResources task.

Also update the documentation and add tests.

### DIFF
--- a/dotnet/BundleContents.md
+++ b/dotnet/BundleContents.md
@@ -3,11 +3,13 @@
 There are a few items that are automatically included during the build, and
 that we're supposed to copy to the app bundle somehow:
 
-* `@(None)`, `@(Content)` and `@(EmbeddedResource)` items with the
-  `CopyToOutputDirectory` or the `CopyToPublishDirectory` metadata set (to
-  either `Always` or `PreserveNewest`).
+* `@(None)` and `@(EmbeddedResource)` items with the `CopyToOutputDirectory` or the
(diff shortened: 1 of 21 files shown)


emaf (approved):

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

* xammac tests/Mac Modern/Release (all optimizations): https://github.com/xamarin/maccore/issues/2139

dalexsoto (approved):

PR #14164, opened 2022-02-16
[nfloat] Obsolete nfloat.CopyArray in legacy Xamarin and stop using it.

rolfbjarne:
There's no corresponding System.Runtime.InteropServices.NFloat.CopyArray method in .NET.

It turned out that the API where we used CopyArray don't need to use CopyArray at all, the same can be accomplished faster and simpler by using unsafe code.

### DIFF
--- a/src/NativeTypes/Primitives.tt
+++ b/src/NativeTypes/Primitives.tt
@@ -38,6 +38,7 @@
 #endif
 
 using System;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


dalexsoto (approved):

mandel-macaque (approved):

tj-devel709 (approved):

src/NativeTypes/Primitives.tt
```diff
@@ -408,6 +409,8 @@ namespace System
 			return ((IConvertible)v).ToType (targetType, provider);
 		}
 
+		[Obsolete ("This API is not available in .NET, use Buffer.MemoryCopy instead.")]
+		[EditorBrowsable (EditorBrowsableState.Never)]
```
spouliot, line 413: Is there a `#if !NET` somewhere that makes them not available in the .net profile ?

rolfbjarne: @spouliot We're going to remove ObjCRuntime.nfloat and use System.Runtime.InteropServices.NFloat instead, and that type doesn't have these APIs.

rolfbjarne:

Issue #14157, opened 2022-02-15
CreateBindingResourcePackage crashing (task failed unexpectedly) 


dk-jlew:
This is a little hard to characterize, but it's acting like part of `CreateBindingResourcePackage` is not thread-safe.  I have a large Xamarin solution with a binding project with multiple NativeReferences to iOS frameworks.  This has been working fine, but I am trying to make the switch to xcframework, and persistently having the `CreateBindingResourcePackage` error out.  It happens consistently (80% of the time) with multi-threaded builds (`msbuild /m`) turned on, and rarely (never?) with maxCpuCount = 1.  I am using the following properties:

```
 <NoBindingEmbedding>true</NoBindingEmbedding>
<NativeReference Include="Resources\GeoComplySDK.xcframework">
      <Kind>Framework</Kind>
      <SmartLink>False</SmartLink>
</NativeReference>
```
...plus 3 other references to related xcframeworks with the same `Kind` and `SmartLink` arguments.

Building this from Visual Studio, or the command line with /m, results in a build errors with a stack trace like so:
```
/build/MyApp.sln" (DK_Gaming_iOS target) (1) ->
       "/build/MyApp/MyApp.iOS/MyApp.iOS.csproj" (default target) (2) ->
       "/build/DK.Ident.GeoComply.iOS/DK.Ident.GeoComply.iOS.csproj" (default target) (5:2) ->
       (_CreateBindingResourcePackage target) ->
         /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/iOS/Xamarin.Shared.targets(181,3): error MSB4018: The "CreateBindingResourcePackage" task failed unexpectedly. [/build/DK.Ident.GeoComply.iOS/DK.Ident.GeoComply.iOS.csproj]
       /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/iOS/Xamarin.Shared.targets(181,3): error MSB4018: System.IO.IOException: Write fault on path /build/DK.Ident.GeoComply.iOS/[Unknown] [/build/DK.Ident.GeoComply.iOS/DK.Ident.GeoComply.iOS.csproj]
[...]

Issue #12422, opened 2021-08-12
Build stuck when merging app bundles into universal app


filipnavara:
### Steps to Reproduce

1. Create a large project targeting `net6.0-macos` with `<RuntimeIdentifiers>osx-x64;osx-arm64</RuntimeIdentifiers>`
2. Run `dotnet build`

### Expected Behavior

Build finishes.

### Actual Behavior

Build gets stuck:
```
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  Detected signing identity:
          
    Bundle Id: com.emclient.mail.client
    App Id: com.emclient.mail.client
/usr/local/share/dotnet/packs/Microsoft.macOS.Sdk/12.0.100-ci.main.458/targets/Xamarin.Shared.Sdk.targets(318,3): error : Unable to merge the file 'Contents/Resources/Assets.car', it's different between the input app bundles. [/Users/filipnavara/Projects/emclient/MailClient/MailClient.csproj]
/usr/local/share/dotnet/packs/Microsoft.macOS.Sdk/12.0.100-ci.main.458/targets/Xamarin.Shared.Sdk.targets(318,3): error : App bundle file #1: /Users/filipnavara/Projects/emclient/MailClient/bin/Debug/net6.0-macos/osx-x64/MailClient.app/Contents/Resources/Assets.car [/Users/filipnavara/Projects/emclient/MailClient/MailClient.csproj]
/usr/local/share/dotnet/packs/Microsoft.macOS.Sdk/12.0.100-ci.main.458/targets/Xamarin.Shared.Sdk.targets(318,3): error : App bundle file #2: /Users/filipnavara/Projects/emclient/MailClient/bin/Debug/net6.0-macos/osx-arm64/MailClient.app/Contents/Resources/Assets.car [/Users/filipnavara/Projects/emclient/MailClient/MailClient.csproj]

^CAttempting to cancel the build...
^C
/usr/local/share/dotnet/packs/Microsoft.macOS.Sdk/12.0.100-ci.main.458/targets/Xamarin.Shared.Sdk.targets(318,3): warning MSB4220: Waiting for the currently executing task "MergeAppBundles" to cancel. [/Users/filipnavara/Projects/emclient/MailClient/MailClient.csproj]
[...]

PR #14176, opened 2022-02-17
[msbuild] Implement proper MSBuild logging in FileCopier.

rolfbjarne:
Fixes https://github.com/xamarin/xamarin-macios/issues/12422.
Fixes https://github.com/xamarin/xamarin-macios/issues/13697.
Fixes https://github.com/xamarin/xamarin-macios/issues/14157.

### DIFF
--- a/tools/common/FileCopier.cs
+++ b/tools/common/FileCopier.cs
@@ -65,17 +65,62 @@ namespace Xamarin.Bundler {
 		static extern int copyfile (string @from, string @to, IntPtr state, CopyFileFlags flags);
 
 		// This code is shared between our packaging tools (mmp\mtouch) and msbuild tasks
-#if MMP || MTOUCH
-		public static void Log (int min_verbosity, string format, params object[] args) => Driver.Log (min_verbosity, format, args);
-		public static Exception CreateError (int code, string message, params object[] args) => ErrorHelper.CreateError (code, message, args);
-#else
-		// LogMessage and LogError are instance objects on the tasks themselves and bubbling an event up is not ideal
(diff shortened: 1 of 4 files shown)


tools/common/FileCopier.cs
```diff
@@ -149,15 +194,31 @@ static CopyFileResult CopyFileCallback (CopyFileWhat what, CopyFileStep stage, I
 			}
 		}
 
+		public static bool IsUptodate (string source, string target, ReportErrorCallback reportErrorCallback, LogCallback logCallback, bool check_contents = false, bool check_stamp = true)
```
tj-devel709, line 197:
```suggestion
		public static bool IsUpToDate (string source, string target, ReportErrorCallback reportErrorCallback, LogCallback logCallback, bool check_contents = false, bool check_stamp = true)
```

tj-devel709: I think maybe this one and the ones below should have this casing, but I do see that there are already existing overloads with that casing so maybe not

tj-devel709 (approved):

tools/common/FileCopier.cs
rolfbjarne: Yeah, I wondered about this too, but decided to not change it to not make the PR bigger than it needs to be.

mandel-macaque: If you apple that, make sure all instances are done too. I have made the mistake of using git commit suggestion to have later a build error :/

mandel-macaque (approved):

mauroa (approved):

emaf (approved):

PR #14187, opened 2022-02-17
[ApiDiffs] More renaming to enable ApiDiffs

tj-devel709:
There was a large change to rename a lot of our Xamarin assemblies to Microsoft
ie) Xamarin.iOS -> Microsoft.iOS

There is a mismatch with some of the prerequisites in our tools/apidiff/Makefile where dependencies 
are looking for ...Microsoft.iOS... but they are still named ...Xamarin.iOS...

This PR takes any remaining "Xamarin" names and changes them to "Microsoft" for all dotnet related rules.
We will also change other dotnet rules to use the new naming convention of "macOS" and "tvOS"

The only exception is to the Xamarin.PLATFORM.dll's coming from the zip - those remain as Xamarin.iOS.dll

We should expect to see the gists showing up in ApiDiffs from this PR!

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -38,10 +38,10 @@ include $(TOP)/Make.versions
 
 APIDIFF_REFERENCES_iOS=https://bosstoragemirror.blob.core.windows.net/wrench/xcode13.1/8fd9e62891f8e4fbaa040cdfbf8b96467060c85c/5371501/package/bundle.zip
 APIDIFF_REFERENCES_Mac=https://bosstoragemirror.blob.core.windows.net/wrench/xcode13.1/8fd9e62891f8e4fbaa040cdfbf8b96467060c85c/5371501/package/bundle.zip
-APIDIFF_REFERENCES_DOTNET_iOS=https://bosstoragemirror.blob.core.windows.net/wrench/main/3c2c3d062daedd66ccd06ecb68192767ddf8e3c4/5315390/package/bundle.zip
-APIDIFF_REFERENCES_DOTNET_tvOS=https://bosstoragemirror.blob.core.windows.net/wrench/main/3c2c3d062daedd66ccd06ecb68192767ddf8e3c4/5315390/package/bundle.zip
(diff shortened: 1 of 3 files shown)


mandel-macaque (approved):
Lets wait to see the diff. I wonder if the links are correct in the comment. I suspect they won't (but I hope I'm wrong)

rolfbjarne (approved):
Thanks for taking care of this!

> The only exception is to the Xamarin.PLATFORM.dll's coming from the zip - those remain as Xamarin.iOS.dll

Maybe it would make sense to bump the zip to a version that has the Microsoft.*.dll version, and fix everything now? We're probably going to forget if we do the bumping later, and then it will break again...

tj-devel709:
> Thanks for taking care of this!
> 
> > The only exception is to the Xamarin.PLATFORM.dll's coming from the zip - those remain as Xamarin.iOS.dll
> 
> Maybe it would make sense to bump the zip to a version that has the Microsoft.*.dll version, and fix everything now? We're probably going to forget if we do the bumping later, and then it will break again...

@rolfbjarne I can change the dll names! I tried to only touch the dotnet items, but would you like me to change the legacy rules and things as well to use Microsoft?

mandel-macaque:
@tj-devel709  I think that the regex works and does the match, but lets double check  the output :)

rolfbjarne:
> > Thanks for taking care of this!
> > > The only exception is to the Xamarin.PLATFORM.dll's coming from the zip - those remain as Xamarin.iOS.dll
> > 
> > 
> > Maybe it would make sense to bump the zip to a version that has the Microsoft.*.dll version, and fix everything now? We're probably going to forget if we do the bumping later, and then it will break again...
> 
> @rolfbjarne I can change the dll names! I tried to only touch the dotnet items, but would you like me to change the legacy rules and things as well to use Microsoft?

The filenames didn't change for legacy Xamarin assemblies, so nothing should change there (except the comparisons between legacy and .NET assemblies).

rolfbjarne (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2443), and I'm merging this since having the apidiff working would be very helpful for a few upcoming PRs I have

PR #14190, opened 2022-02-18
[appcompare] Remove old/local copy of the tool and update comparison documentation

spouliot:
There's a newer version of the tool availble as a (tool)
[nuget](https://www.nuget.org/packages/appcompare/)

The nicest part, for this use case, is that it can map renamed files
between the two app bundles being compared.
[Example](https://gist.github.com/spouliot/68a43a4f514315d52b35446016ba0d2e)


rolfbjarne:

Issue #14182, opened 2022-02-17
[net] MySingleView (minimal) app just added a dependency on `System.Console.dll`


spouliot:
That's quite recent (likely last .net bump) but I have not looked for the specific commit.

That adds more than 20kb to the app size (beside the native AOT'ed code).

https://gist.github.com/spouliot/2448f92a45f960c17fcada82c6dd65da

spouliot:
That's the (hopefully only) culprit.

```
				Console.WriteLine ("The type {0} does not have a constructor that takes an ObjCRuntime.NativeHandle parameter, but a constructor that takes an System.IntPtr parameter was found instead. It's highly recommended to change the signature of the System.IntPtr constructor to be ObjCRuntime.NativeHandle instead.", type.FullName);
```

I'll fix it (to use `NSLog`) while trying to reduce it's size/duplication.

Issue #14188, opened 2022-02-18
Reducing strings size in assemblies


spouliot:
## Notes
 
* This is a meta bug that will require several PR to fix / optimize.
* The macOS version of `strings` does not show unicode / utf16 strings. The [workaround](https://github.com/spouliot/appcompare/wiki/DiffCustomToolOutput#strings) is to install the GNU version.

## Generating the data files

```bash
$ strings size-comparison/MySingleView/oldnet/bin/iPhone/Release/MySingleView.app/Xamarin.iOS.dll | sort > xu8.txt

$ /opt/homebrew/opt/binutils/bin/strings -e b size-comparison/MySingleView/oldnet/bin/iPhone/Release/MySingleView.app/Xamarin.iOS.dll | sort > xu16.txt

$ strings size-comparison/MySingleView/dotnet/bin/iPhone/Release/net6.0-ios/ios-arm64/MySingleView.app/Microsoft.iOS.dll | sort > mu8.txt

$ /opt/homebrew/opt/binutils/bin/strings -e b size-comparison/MySingleView/dotnet/bin/iPhone/Release/net6.0-ios/ios-arm64/MySingleView.app/Microsoft.iOS.dll | sort > mu16.txt

$ ls -l *.txt

-rw-r--r--  1 poupou  staff   6315 17 Feb 18:21 mu16.txt
-rw-r--r--  1 poupou  staff  16873 17 Feb 18:21 mu8.txt
-rw-r--r--  1 poupou  staff   6496 17 Feb 18:20 xu16.txt
-rw-r--r--  1 poupou  staff  16276 17 Feb 18:17 xu8.txt
```

Considering the _linked_ **and** stripped platform assemblies are around 50kb (for the MySingleView.app tracking app) we can see that the strings represent about 50% of what's being shipped (utf16 being twice has large as the files lengths).

```bash
-rw-r--r--  1 poupou  staff  56832 17 Feb 17:38 Xamarin.iOS.dll

-rw-r--r--  1 poupou  staff  53760 17 Feb 17:38 Microsoft.iOS.dll
```

There's a lot of required strings (e.g. symbols) that cannot be removed. However there are also many duplications (some partials) of error messages in the unicode strings (the `*u16.txt` files).

The good news is that using string interpolation (instead of `String.Format`) has the nice side effect of breaking the strings, between `{ }`, which makes them reusable in many more places.

## Files

Generated with commit 1273190e5fdcea446ce4664031048961a3235e73

[...]

PR #14191, opened 2022-02-18
Reduce string duplication in the platform assemblies (part 1)

spouliot:
The size of unicode strings for the minimal MySingleView.app (.net
version) goes from 6315 to 5171 characters (each being 2 bytes).

Part of the work to solve/minimize https://github.com/xamarin/xamarin-macios/issues/14188

This also removes the recent `Console.WriteLine` and replace it
with a call to `NSLog` so `System.Console.dll` is not part of the
final app bundle (size regression).

Fixes https://github.com/xamarin/xamarin-macios/issues/14182


rolfbjarne (approved):

rolfbjarne:
I think the conflict has to be resolved before CI will do anything here :/

dalexsoto (approved):

PR #14192, opened 2022-02-18
[msbuild/generator] Pass .NET's C# compiler to bgen to use. Fixes #12922.

rolfbjarne:
We can't execute mono's C# compiler when using .NET, so we need to tell bgen
where csc is in that case.

Fixes https://github.com/xamarin/xamarin-macios/issues/12922.

PR #14197, opened 2022-02-18
[runtime] Remove ObjCRuntime.nfloat in favor of System.Runtime.InteropServices.NFloat.

rolfbjarne:
⚠️ Draft PR ⚠️

* This contains the changes from #14192 as well, those will be removed once that PR is merged *

---

* Remove ObjCRuntime.nfloat (in favor of
  System.Runtime.InteropServices.NFloat).
* Automatically add a reference to the System.Runtime.InteropServices.Internal
  package, so that developers get the new NFloat API (with operators) we've
  added post .NET 6 (but pre .NET 7).
* Automatically add a global using alias for
  System.Runtime.InteropServices.NFloat -> nfloat. This is not behind the
  usual `ImplicitUsings` condition our other implicit usings are, because
  they're off by default for existing projects, and the main target for the
  global using alias for nfloat is upgraded projects.
* Automatically generate a global using alias (like above) in the generator
  for all code the generator compiles.
* Update xtro entries to reference System.Runtime.InteropServices.NFloat
  instead of ObjCRuntime.nfloat.
* Add a workaround for a hopefully temporary issue with .NET/CoreCLR where the
  wrong runtime pack is selected otherwise (without the new NFloat API, so
  nothing works at runtime).

Ref: https://github.com/xamarin/xamarin-macios/issues/13087


tests/introspection/ApiSignatureTest.cs
```diff
@@ -38,7 +38,7 @@ namespace Introspection {
 
 	public abstract class ApiSignatureTest : ApiBaseTest {
 #if NET
-		const string NFloatTypeName = "ObjCRuntime.nfloat";
+		const string NFloatTypeName = "System.Runtime.InteropServices.nfloat";
```
spouliot, line 41: `NFloat` ?

spouliot (commented):
❤️

tests/introspection/ApiSignatureTest.cs
rolfbjarne: Good catch!

dalexsoto (approved):
Looking good!

mauroa (approved):

chamons (approved):

rolfbjarne:
Test failure is unrelated (https://github.com/xamarin/maccore/issues/2542)

mandel-macaque (approved):

emaf (approved):

PR #14202, opened 2022-02-18
Bump minimum Visual Studio for Mac version to 17.0 series.

Therzok:
It currently still uses a preview version, this should be tracked on stable release.
Another option would be to query for the latest version and install that.

The usage of SYSTEM_MONO seems to be needed still.

### DIFF
--- a/Make.config
+++ b/Make.config
@@ -144,6 +144,13 @@ MACCATALYST_NUGET_VERSION_PATCH=$(word 3, $(subst ., ,$(MACCATALYST_NUGET_VERSIO
 MACCATALYST_NUGET_VERSION_NO_METADATA=$(MACCATALYST_NUGET_VERSION)-$(NUGET_PRERELEASE_IDENTIFIER)$(MACCATALYST_NUGET_COMMIT_DISTANCE)
 MACCATALYST_NUGET_VERSION_FULL=$(MACCATALYST_NUGET_VERSION_NO_METADATA)+$(NUGET_BUILD_METADATA)
 
+# Compute the architecture for the Visual Studio for Mac dmg
+ifeq ($(shell arch),arm64)
+VSMAC_ARCH=arm64
+else
+VSMAC_ARCH=x64
+endif
+
 # Xcode version should have both a major and a minor version (even if the minor version is 0)
 XCODE_VERSION=13.2
 XCODE_URL=https://dl.internalx.com/internal-files/xcodes/Xcode_13.2.xip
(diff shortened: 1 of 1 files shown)


dalexsoto:
We still need to support legacy so if we want to bump probably latest 8.10 is the one we want until you are stable

Therzok:
From the looks of it, VSMac is only used for:

Building Touch.Unit, which I suppose can be migrated to msbuild/dotnet?
```
external/Touch.Unit/Makefile
1:MDTOOL="/Applications/Xamarin Studio.app/Contents/MacOS/mdtool"
13:	$(MDTOOL) -v build -t:Build "-c:Debug|iPhoneSimulator" Touch.Unit.sln
21:	$(MDTOOL) -v build -t:Build "-c:Release|iPhone" Touch.Unit.sln
```

Running TextTransform:
```
Make.config
475:TT = $(SYSTEM_MONO) "/Applications/Visual Studio.app/Contents/MonoBundle/AddIns/MonoDevelop.TextTemplating/TextTransform.exe"
```

The latter is also available as dotnet tool CLI:
https://www.nuget.org/packages/Mono.TextTemplating/

I think effort is better invested in removing these dependencies rather than keeping them up to date. :D

dalexsoto:
> From the looks of it, VSMac is only used for:
> 
> Building Touch.Unit, which I suppose can be migrated to msbuild/dotnet?
> 
> ```
> external/Touch.Unit/Makefile
> 1:MDTOOL="/Applications/Xamarin Studio.app/Contents/MacOS/mdtool"
> 13:	$(MDTOOL) -v build -t:Build "-c:Debug|iPhoneSimulator" Touch.Unit.sln
> 21:	$(MDTOOL) -v build -t:Build "-c:Release|iPhone" Touch.Unit.sln
> ```
> 
> Running TextTransform:
> 
> ```
> Make.config
> 475:TT = $(SYSTEM_MONO) "/Applications/Visual Studio.app/Contents/MonoBundle/AddIns/MonoDevelop.TextTemplating/TextTransform.exe"
> ```
> 
> The latter is also available as dotnet tool CLI: https://www.nuget.org/packages/Mono.TextTemplating/
> 
> I think effort is better invested in removing these dependencies rather than keeping them up to date. :D

Yup totally agree

mandel-macaque:
+1 on removing the need of VS, but legacy might be an issue.

mandel-macaque (approved):

rolfbjarne:
I've created a PR to remove the VS dependency completely: #15102, so I'm closing this.

PR #14208, opened 2022-02-21
[CoreGraphics] Improve matrix types for .NET. Fixes #14125.

### LINKED ISSUE #14125: [iOS/macOS][net6] Member renames to `CGAffineTransform` are error prone when upgrading from Xamarin.iOS to net6 preview 13.2
### Steps to Reproduce

Upgrade a piece of code which uses any member of `CGAffineTransform` to net6.

### Expected Behavior

Get indications on what to upgrade to, or have no breaking change.

### Actual Behavior

Members are missing as they have been renamed in https://github.com/xamarin/xamarin-macios/commit/209eca1fb7e6326f4a2782b30dd4539512f8e8a5.

### Environment

15.2.301-preview.13.2

rolfbjarne:
Hoping that one day we'll have a System.Numerics.Matrix3x3 type we can replace our
RMatrix3 type with:

* Add all API in OpenTK.Matrix3 that also exists in equivalent form in System.Numerics.Matrix4x4.
* Remove all API that doesn't exist in equivalent form in System.Numerics.Matrix4x4.

For NMatrix2 and NMatrix3:

* Change the fields to be M## instead of R#C# (this is how System.Numerics does
  it, and also how we do it in other matrix types).
* Add obsolete and hidden R#C# versions of the fields to ease migrating existing
  code to .NET (but let's try to remove these in xamcore 5).
* Add properties in legacy Xamarin recommending users to use the new naming.
* A few other API additions to add equivalent API to the matrix types in System.Numerics.

For CGAffineMatrix:

* Add obsolete and hidden versions of the legacy field names to .NET to ease
  migrating existing (but let's try to remove these in xamcore 5).

Fixes https://github.com/xamarin/xamarin-macios/issues/14125.

### DIFF
--- a/src/Simd/MatrixFloat3x3RM.cs
+++ b/src/Simd/MatrixFloat3x3RM.cs
@@ -4,15 +4,21 @@
 //
 // Copyright (c) 2022 Microsoft Inc
 //
-
 //
 // This represents the native GLKMatrix3 type, which has a row-major layout
 // (same as OpenTK.Matrix3).
+//
+// One day we might be able to remove this type in favor of a System.Numerics.Matrix3x3 type (which doesn't exist yet),
+// so the only API available here is the one that I _think_ System.Numerics.Matrix3x3 will have (based on the existing
+// System.Numerics.Matrix4x4 type). In particular it must not have any API that *doesn't* exist in a potential
+// System.Numerics.Matrix3x3 type (🔮).
 //
 
 #nullable enable
 
 using System;
+using System.ComponentModel;
+using System.Numerics;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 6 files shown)


src/Simd/MatrixFloat3x3RM.cs
```diff
@@ -4,15 +4,21 @@
 //
 // Copyright (c) 2022 Microsoft Inc
 //
-
```
tj-devel709, line 7: nit: in case this was an accident

chamons (approved):
I'm going to approve this, with an admission that I did not do a hard core review. 

There are a _lot_ of variable changes, and they blur together.

PR #14213, opened 2022-02-22
Turn on template localization on dotnet CLI

bekir-ozturk:
Template engine (`dotnet new`) added support for template localizations in .NET 6.0. This new system replaces the old way of localizing templates that only worked on Visual Studio and works on both .NET CLI as well as VS.

This PR introduces changes to switch to the new template localization system.

The PR is currently draft with remaining work:
- [x] Restore tasks package
- [x] Configure root templates path
- [x] Initial loc files need to be generated
- [x] Update `LocProject.json` (automate, if possible)
- [x] Test-run the pipeline

### DIFF
--- a/dotnet/package/microsoft.templates.csproj
+++ b/dotnet/package/microsoft.templates.csproj
@@ -4,10 +4,10 @@
     <PackageType>Template</PackageType>
     <Description>Templates for $(_PlatformName) platforms</Description>
     <_packagePath>$(MSBuildThisFileDirectory)..\Templates\Microsoft.$(_PlatformName).Templates\</_packagePath>
-    <LocalizeTemplates Condition=" '$(ContinuousIntegrationBuild)' != 'true' ">true</LocalizeTemplates>
+    <LocalizeTemplates Condition=" '$(CIBuild)' != 'true' ">true</LocalizeTemplates>
     <LocalizableTemplatesPath>$(_packagePath)</LocalizableTemplatesPath>
   </PropertyGroup>
-  <ItemGroup Condition=" '$(ContinuousIntegrationBuild)' != 'true' ">
+  <ItemGroup Condition=" '$(CIBuild)' != 'true' ">
     <PackageReference Include="Microsoft.TemplateEngine.Tasks" Version="$(MicrosoftTemplateEngineTasksVersion)" PrivateAssets="all" IsImplicitlyDefined="true" />
   </ItemGroup>
   <Import Project="common.csproj" />
--- a/dotnet/package/microsoft.templates.csproj
+++ b/dotnet/package/microsoft.templates.csproj
@@ -4,6 +4,10 @@
     <PackageType>Template</PackageType>
     <Description>Templates for $(_PlatformName) platforms</Description>
     <_packagePath>$(MSBuildThisFileDirectory)..\Templates\Microsoft.$(_PlatformName).Templates\</_packagePath>
+    <LocalizeTemplates>true</LocalizeTemplates>
   </PropertyGroup>
+  <ItemGroup Condition=" '$(OS)' == 'Windows_NT' ">
+    <PackageReference Include="Microsoft.TemplateEngine.Tasks" Version="$(MicrosoftTemplateEngineTasksVersion)" PrivateAssets="all" IsImplicitlyDefined="true" />
+  </ItemGroup>
   <Import Project="common.csproj" />
 </Project>
--- a/dotnet/package/microsoft.templates.csproj
+++ b/dotnet/package/microsoft.templates.csproj
@@ -4,9 +4,10 @@
     <PackageType>Template</PackageType>
     <Description>Templates for $(_PlatformName) platforms</Description>
     <_packagePath>$(MSBuildThisFileDirectory)..\Templates\Microsoft.$(_PlatformName).Templates\</_packagePath>
-    <LocalizeTemplates>true</LocalizeTemplates>
+    <LocalizeTemplates Condition=" '$(ContinuousIntegrationBuild)' != 'true' ">true</LocalizeTemplates>
+    <LocalizableTemplatesPath>$(_packagePath)</LocalizableTemplatesPath>
   </PropertyGroup>
-  <ItemGroup Condition=" '$(OS)' == 'Windows_NT' ">
+  <ItemGroup Condition=" '$(ContinuousIntegrationBuild)' != 'true' ">
     <PackageReference Include="Microsoft.TemplateEngine.Tasks" Version="$(MicrosoftTemplateEngineTasksVersion)" PrivateAssets="all" IsImplicitlyDefined="true" />
   </ItemGroup>
   <Import Project="common.csproj" />
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -27,5 +27,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>db49d790a4bfa977a9ed7436bf2aa242cefae45e</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.TemplateEngine.Tasks" Version="7.0.100-alpha.1.21601.1">
+      <Uri>https://github.com/dotnet/templating</Uri>
+      <Sha />
+    </Dependency>
   </ToolsetDependencies>
 </Dependencies>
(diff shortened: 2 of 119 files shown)


dotnet/package/microsoft.templates.csproj
```diff
@@ -4,6 +4,10 @@
     <PackageType>Template</PackageType>
     <Description>Templates for $(_PlatformName) platforms</Description>
     <_packagePath>$(MSBuildThisFileDirectory)..\Templates\Microsoft.$(_PlatformName).Templates\</_packagePath>
+    <LocalizeTemplates>true</LocalizeTemplates>
   </PropertyGroup>
+  <ItemGroup Condition=" '$(OS)' == 'Windows_NT' ">
+    <PackageReference Include="Microsoft.TemplateEngine.Tasks" Version="$(MicrosoftTemplateEngineTasksVersion)" PrivateAssets="all" IsImplicitlyDefined="true" />
+  </ItemGroup>
```
rolfbjarne, line 12: I don't think this will have any effect, because we build on macOS.

bekir-ozturk: Updated, thanks!

bekir-ozturk:
@rolfbjarne Can you approve the CI runs?
I would also appreciate another review.

I made some changes to configure.yml file which seems to contain OneLocBuild task. Does this one have a special pipeline? Is there a way I can make some test runs?

rolfbjarne:
Azure DevOps isn't happy:

> /tools/devops/automation/templates/build/configure.yml (Line: 171, Col: 5): A sequence was not expected

rolfbjarne:
@bekir-ozturk I'm not sure I know what to look for to verify CI did as expected?

bekir-ozturk:
@rolfbjarne There is 2 parts to this PR:
1. Generating the loc files for templates
2. Passing the generated files to OneLocBuild process.

If I were to explain briefly:
1. The first part is mostly done by MSBuild. We pull the package from a nuget feed and invoke the task in it. The task scans the repo and generates some loc files. This only happens in development machines and not in CI. So the best testing would be to see if the solution builds locally.
2. The second part is about adding the generated files into LocProject.json file. This part happens in CI, through `configure.yml` file. Right after the initial `LocProject.json` file is generated, a new script is invoked to patch the file to include the new changes. This happens in "Configure Build" step of the pipeline. If it is green, we should be good. One trick here is that the consumer of `LocProject.json` file does not run unless the CI is triggered from the main branch. Therefore, CI won't tell us if the consumer is happy with the state of the file we left it in. Therefore, I temporarily forced it to run [here](https://github.com/xamarin/xamarin-macios/pull/14213/commits/525d7705f7c04930853f350ffd95db884cc2b668) to see if it was happy with the changes. I suggest we trigger at least one CI run to see the results before reverting this particular PR.

Feel free to ping me in Teams to discuss the details (though I will be unavailable temporarily for the next few hours).

bekir-ozturk:
Having temporarily turned on OneLocBuild task in CI, we seem to have template strings picked up correctly.

image
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5801482&view=logs&j=d4d86769-bae0-5955-544a-0740a81efd3c&t=b076b8fd-1a89-5fdb-2a15-f02f9055d0fd&l=189

image
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5801482&view=artifacts&pathAsName=false&type=publishedArtifacts

I will revert OneLocBuild task execution condition so that it only runs on 'main' again. Then this will be ready for review.

eng/Version.Details.xml
```diff
@@ -27,5 +27,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>db49d790a4bfa977a9ed7436bf2aa242cefae45e</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.TemplateEngine.Tasks" Version="7.0.100-alpha.1.21601.1">
+      <Uri>https://github.com/dotnet/templating</Uri>
+      <Sha />
```
rolfbjarne, line 32: Is the lack of hash here expected?

rolfbjarne (approved):

rolfbjarne:
@mandel-macaque could you have a look too?

tools/devops/azure-pipelines.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     name: 'Azure Pipelines'
     vmImage: 'internal-macos-10.15'
 
+  variables:
+  - name: ContinuousIntegrationBuild
+    value: true
+
```
mandel-macaque, line 16: This is not the correct pipeline for CI, it is tools/devops/automation/build-pipeline.yml

I probably need to find what this pipeline is used, or even if it is used.

mandel-macaque (changes requested):

tools/devops/azure-pipelines.yml
mandel-macaque: I am removing this file here: https://github.com/xamarin/xamarin-macios/pull/14274 I understand is very misleading. You can add the variable at the top of  tools/devops/automation/build-pipeline.yml

tj-devel709:
Looks like the OneLocBuild step is acknowledging the changes you made well! We have only used resx files for OneLocBuild and just want to confirm that that can translate the json files to lcl files as well? I see in the LocProject.json that the paths are to the json files though so I assume that that is sufficient? If so, this looks good to me!

Also just so you know, OneLocBuild requires that we have a green (warnings are okay though) build for them to translate our lcl files so we do a "dry-run" on Sunday nights through a cron job that do not run our device tests to get them a green build. We also ask the Loc team to only give us the updated lcl files once a week to prevent overlapping PRs from them. Therefore, after getting this PR in, there may be a delay in processing the localization.

eng/Version.Details.xml
bekir-ozturk: I have taken this part from the earlier PR in dotnet/maui repo. https://github.com/dotnet/maui/pull/4439/files
Seems to be OK not to have a hash.

tools/devops/azure-pipelines.yml
bekir-ozturk: Thanks! I moved the variable declaration to the correct file.

bekir-ozturk:
Hi @tj-devel709,
 Yes, OneLocBuild can work with json files as well. We have gotten approval for this file format right before we started adding localization support to `dotnet new`.
 Other repos such as dotnet/aspnetcore, dotnet/winforms and dotnet/wpf have already been using this for some time to localize their templates. No issues so far 🤞

bekir-ozturk:
@mandel-macaque Could you review again? Triggering the CI would also be appreciated :)

rolfbjarne:
The build fails with:

```
[...]
/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/15.0/bin/MSBuild.dll /nologo /bl:bin/Release/mmp.exe.binlog /m /p:Features=strict /p:Configuration=Release /t:mmp /verbosity:quiet ../../Xamarin.Mac.sln
/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Roslyn/Microsoft.Managed.Core.targets(191,5): error : SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true [/Users/builder/azdo/_work/1/s/xamarin-macios/tools/mmp/mmp.csproj]
make[2]: *** [bin/Release/mmp.exe] Error 1
```

which does not look related, but on the other hand it only happens in this PR (and I've never seen this before).

But I can't reproduce it locally, so I don't really know what's going on here.

rolfbjarne:
Something is setting `DeterministicSourcePaths=true` in the environment in CI.

@bekir-ozturk: can any of the changes you made cause that?

bekir-ozturk:
@rolfbjarne it appears that the variable name `ContiuousIntegrationBuild` I used has a special meaning for roslyn which is causing the problem.
I have created an issue here: https://github.com/dotnet/sdk/issues/24232
To get us unblocked, I will rename the variable.

rolfbjarne:
Test failures are unrelated (https://github.com/xamarin/xamarin-macios/pull/14334)

rolfbjarne:
@mandel-macaque Could you re-review this PR?

mandel-macaque (approved):

bekir-ozturk:
@rolfbjarne Looks like we have the approvals 🎊  I rebased to fix the conflicts with the main. Can you trigger the CI?

bekir-ozturk:
@rolfbjarne Do the errors seem related to this PR or is this the result of a known flakiness?

mandel-macaque:
@bekir-ozturk  those are tests that we know have to be fixed, one flaky one broken due to some CI needed work, we can land this :)

PR #14217, opened 2022-02-22
[UIKit] Hide the obsolete UIApplication.Main overload.

rolfbjarne:
Hide the obsolete UIApplication.Main overload that takes string parameters
from intellisense. This overload is used *a lot*, so it's not worth it to
remove it, since it would break a lot of user code.


tj-devel709 (approved):

rolfbjarne:

PR #14219, opened 2022-02-22
[CoreGraphics] Remove a CGPDFOperatorTable.SetCallback overload in .NET.

rolfbjarne:
Remove the CGPDFOperatorTable.SetCallback overload that takes a normal managed
delegate, because on platforms that are AOT'ed, this delegate must point to a
static function with the MonoPInvokeCallback attribute, and if you don't
follow this requirement, you'll either get an exception at runtime (which is
not very nice to the app developer) or make the AOT compiler crash (which is a
completely different level of not being nice to developers).

In .NET, we already have an overload that takes an unmanaged function pointer,
where these requirements are enforced by the C# compiler, so just use that
instead.


tj-devel709 (approved):

chamons (approved):

rolfbjarne:
Test failures are unrelated:

* xcframework-test/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414

PR #14221, opened 2022-02-22
[addressbookui] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to AddressBookUI.

*Note, I'd like extra focus on the DisplayedPropertiesCollection method changes

Inside AddressBookUI, I removed 
`[Obsolete ("Use SetHighlightedItemForProperty(ABPersonProperty,int?).")] `
since it is already marked as Internal thus not for public use. The issue with leaving it in is that with '#nullable enable`, the compiler tells us that we can't use this method since it is obsoleted, but we use this method in many places.

### DIFF
--- a/src/AddressBookUI/ABPersonViewController.cs
+++ b/src/AddressBookUI/ABPersonViewController.cs
@@ -6,6 +6,8 @@
 // Copyright (C) 2009 Novell, Inc
 //
 
+#nullable enable
+
 using System;
 
 using AddressBook;
@@ -39,7 +41,7 @@ namespace AddressBookUI {
 
 	class InternalABPersonViewControllerDelegate : ABPersonViewControllerDelegate {
 
-		internal EventHandler<ABPersonViewPerformDefaultActionEventArgs> performDefaultAction;
+		internal EventHandler<ABPersonViewPerformDefaultActionEventArgs>? performDefaultAction;
 
 		public InternalABPersonViewControllerDelegate ()
 		{
@@ -108,14 +110,14 @@ namespace AddressBookUI {
 		public void SetHighlightedItemForProperty (ABPersonProperty property, int? identifier)
 		{
 			SetHighlightedItemForProperty (
-					ABPersonPropertyId.ToId (property),
-					identifier.HasValue ? identifier.Value : ABRecord.InvalidPropertyId);
+					property,
+					identifier ?? ABRecord.InvalidPropertyId);
 		}
 
 		public void SetHighlightedProperty (ABPersonProperty property)
 		{
 			SetHighlightedItemForProperty (
-					ABPersonPropertyId.ToId (property),
+					property,
 					ABRecord.InvalidPropertyId);
 		}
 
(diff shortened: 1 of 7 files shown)


src/AddressBookUI/ABPersonViewController.cs
```diff
@@ -108,21 +110,21 @@ partial class ABPersonViewController {
 		public void SetHighlightedItemForProperty (ABPersonProperty property, int? identifier)
 		{
 			SetHighlightedItemForProperty (
-					ABPersonPropertyId.ToId (property),
-					identifier.HasValue ? identifier.Value : ABRecord.InvalidPropertyId);
+					property,
```
rolfbjarne, line 113: It looks like this is a difference in behavior: this will run into infinite recursion, because you removed the call to ABPersonPropertyId.ToId. That line should be reverted:

```suggestion
					ABPersonPropertyId.ToId (property),
```

rolfbjarne (changes requested):

src/AddressBookUI/ABPersonViewController.cs
tj-devel709: @rolfbjarne  Ah good catch! I didn't realize I was calling the same method... The reason I made this change is that the SetHighlightedItemForProperty (int property, int identifier) is Obsolete and suggests that I use this one - which in this case does not make sense. Should I in fact continue to use the obsoleted one?

rolfbjarne: The `SetHighlightedItemForProperty (int, int)` overload is both internal and obsolete - which makes the obsolete redundant because nobody will see it, and you can just remove the obsolete attribute:

https://github.com/xamarin/xamarin-macios/blob/5cc92ac564ece03efb79c9b271beafc8faf905c0/src/addressbookui.cs#L164-L169

The comment is also out of date, so it can be removed too.

src/AddressBookUI/DisplayedPropertiesCollection.cs
```diff
@@ -117,7 +119,7 @@ IEnumerator IEnumerable.GetEnumerator ()
 
 		public IEnumerator<ABPersonProperty> GetEnumerator ()
 		{
-			var values = g ();
+			var values = g ()!;
```
mandel-macaque, line 122: I think you can do better than using a ! since we are doing an IEnumerator, can we check if g is null and then throw the appropriate exception, for example:

```csharp
if (g is null) {
    yield break;
}
```

mandel-macaque (changes requested):

rolfbjarne (approved):

mandel-macaque (changes requested):
Why are we removing an Obsolete attr?

src/AddressBookUI/ABPersonViewController.cs
tj-devel709: @mandel-macaque This should answer your question below, however, I can add this reason in the PR description

src/addressbookui.cs
```diff
@@ -164,7 +164,6 @@ interface ABPersonViewController : UIViewControllerRestoration {
 		// Obsolete for public use; we should "remove" this member by making
 		// it [Internal] in some future release, as it's needed internally.
 		[Internal]
-		[Obsolete ("Use SetHighlightedItemForProperty(ABPersonProperty,int?).")]
```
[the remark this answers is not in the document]
tj-devel709: @mandel-macaque Tagged you in the above convo with Rolf!

PR #14232, opened 2022-02-23
[audiotoolbox] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to <FRAMEWORK>.

### DIFF
--- a/src/AudioToolbox/AudioFileStream.cs
+++ b/src/AudioToolbox/AudioFileStream.cs
@@ -188,7 +188,7 @@ namespace AudioToolbox {
 		protected virtual void OnPacketDecoded (int numberOfBytes, IntPtr inputData, AudioStreamPacketDescription []? packetDescriptions)
 		{
 			var p = PacketDecoded;
-			if (p != null)
+			if (p is not null)
 				p (this, new PacketReceivedEventArgs (numberOfBytes, inputData, packetDescriptions));
 		}
 
@@ -196,7 +196,7 @@ namespace AudioToolbox {
 		protected virtual void OnPropertyFound (AudioFileStreamProperty propertyID, ref AudioFileStreamPropertyFlag ioFlags)
 		{
 			var p = PropertyFound;
-			if (p != null){
+			if (p is not null){
 				var pf = new PropertyFoundEventArgs (propertyID, ioFlags);
 				p (this, pf);
 				ioFlags = pf.Flags; 
@@ -240,7 +240,7 @@ namespace AudioToolbox {
 
 		public AudioFileStreamStatus ParseBytes (byte [] bytes, bool discontinuity)
 		{
-			if (bytes == null)
+			if (bytes is null)
 				ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (bytes));
 			unsafe {
 				fixed (byte *bp = &bytes[0]){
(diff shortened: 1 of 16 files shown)


src/AudioToolbox/AudioFileStream.cs
```diff
@@ -102,19 +104,19 @@ public override string ToString ()
...
 		}
 		public int Bytes { get; private set; }
 		public IntPtr InputData { get; private set; }
-		public AudioStreamPacketDescription [] PacketDescriptions { get; private set;}
+		public AudioStreamPacketDescription []? PacketDescriptions { get; private set;}
 
 		public override string ToString ()
 		{
-			return String.Format ("Packet (Bytes={0} InputData={1} PacketDescriptions={2}", Bytes, InputData, PacketDescriptions.Length);
+			return String.Format ("Packet (Bytes={0} InputData={1} PacketDescriptions={2}", Bytes, InputData, PacketDescriptions!.Length);
```
rolfbjarne, line 119: This might be better with something like this:

```suggestion
			return String.Format ("Packet (Bytes={0} InputData={1} PacketDescriptions={2}", Bytes, InputData, PacketDescriptions?.Length ?? -1);
```

1. It will never throw a NullReferenceException (which ToString shouldn't do)
2. You can deduce from the output that PacketDescriptions is null (if "-1" is printed as the length).

src/AudioToolbox/AudioSession.cs
```diff
@@ -281,16 +283,16 @@ public static void Initialize ()
...
-			CFString s = runMode == null ? null : new CFString (runMode);
+			CFString? s = runMode is null ? null : new CFString (runMode);
 #if NET
 			int k;
 			unsafe {
 				k = AudioSessionInitialize (runLoop.GetHandle (), s.GetHandle (), &Interruption, IntPtr.Zero);
 			}
 #else
-			int k = AudioSessionInitialize (runLoop == null ? IntPtr.Zero : runLoop.Handle, s == null ? IntPtr.Zero : s.Handle, Interruption, IntPtr.Zero);
+			int k = AudioSessionInitialize (runLoop is null ? IntPtr.Zero : runLoop.Handle, s is null ? IntPtr.Zero : s.Handle, Interruption, IntPtr.Zero);
```
rolfbjarne, line 295: You can call GetHandle to simplify the code:

```suggestion
			int k = AudioSessionInitialize (runLoop.GetHandle (), s.GetHandle (), Interruption, IntPtr.Zero);
```

GetHandle is an extension method that's safe to call on null:

https://github.com/xamarin/xamarin-macios/blob/3b617adb0f0ad68d95129f550a7672629d46a2c5/src/ObjCRuntime/INativeObject.cs#L25-L28

src/AudioToolbox/AudioQueue.cs
```diff
@@ -1275,35 +1277,35 @@ static void output_callback (IntPtr userData, IntPtr AQ, IntPtr audioQueueBuffer
...
+		public OutputAudioQueue (AudioStreamBasicDescription desc, CFRunLoop? runLoop, CFString? runMode)
 		{
 			IntPtr h;
 			GCHandle gch = GCHandle.Alloc (this);
 
 			var code = AudioQueueNewOutput (ref desc, dOutputCallback, GCHandle.ToIntPtr (gch),
-							runLoop == null ? IntPtr.Zero : runLoop.Handle,
-							runMode == null ? IntPtr.Zero : runMode.Handle, 0, out h);
+							runLoop is null ? IntPtr.Zero : runLoop.Handle,
+							runMode is null ? IntPtr.Zero : runMode.Handle, 0, out h);
```
rolfbjarne, line 1308:
```suggestion
							runLoop.GetHandle (),
							runMode.GetHandle (), 0, out h);
```

rolfbjarne (approved):

tj-devel709:
**Unrelated Test Failure**
- trimmode copy/Mac Catalyst [dotnet]/Release [dotnet]: https://github.com/xamarin/maccore/issues/2414

dalexsoto (approved):

PR #14243, opened 2022-02-24
[monotouch-test] Fix the BaseEffectTest and EffectPropertyMaterialTest tests.

rolfbjarne:
The [Culture ("en")] attribute means: only run this test if the culture is
"en". This usually meant not running this test (apparently we don't run often
with culture = "en"), leading to outdated tests that happened to fail when
actually run under culture = "en" (such as on older macOS bots).

So change these tests to actually change the culture to "en" (by using the
SetCulture attribute), and also fix them.


tj-devel709:
Interesting, what culture are they defaulting to?

tj-devel709 (approved):

dalexsoto (approved):

rolfbjarne:

PR #14251, opened 2022-02-25
[src] Remove .NET-style availability attributes from API definitions.

rolfbjarne:
It's not necessary, the generator will generate the .NET-style attributes.


chamons (approved):
I hope you didn't do this by hand, but by revert. If so, let me know next time, mel can do this.

PR #14254, opened 2022-02-25
[nfloat] Fix conversion to and from [U]IntPtr.

rolfbjarne:
Fix conversion between nfloat and [U]IntPtr to use a numeric conversion
instead of copying memory.

Otherwise things like this:

    var number = (nint) (nfloat) 10;
    Console.WriteLine (number);

yields:

    0x4024000000000000

instead of:

    0xa

### DIFF
--- a/src/AppKit/NSColor.cs
+++ b/src/AppKit/NSColor.cs
@@ -188,30 +188,19 @@ namespace AppKit {
 			if (components == null)
 				throw new ArgumentNullException ("components");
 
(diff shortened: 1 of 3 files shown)


chamons (approved):

dalexsoto (approved):

mandel-macaque (approved):
Got tests for this change?

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated, and fixed in main.

PR #14258, opened 2022-02-25
[cecil-tests] Improve the obsolete member test. Fixes #13621.

### LINKED ISSUE #13621: Verify obsolete selectors to be removed
This issue addresses the last item on the checklist at https://github.com/xamarin/xamarin-macios/issues/10216.

https://github.com/xamarin/xamarin-macios/pull/13590 adds a cecil test as a starting point for verifying the obsolete selectors to be removed. We will use the switch case in the test to filter the selectors by namespace.

Currently, the test passes by default for all namespaces. Once all the selectors for all namespaces are verified, this should be switched to `return false` by default, and only `return true` for namespaces that have been resolved.

### Expected Behavior
The test switch case should `return false` by default. The test should pass not because it returns true, but because all the obsolete selectors have been resolved.

### Actual Behavior
Currently, the test passes by default for all namespaces.

rolfbjarne:
* Check implementation assemblies instead of reference assemblies.
* Try to print the source code location for failing API (this required processing
  the implementation assemblies, because the reference assemblies don't have debug
  information where the source code location is stored).
* Don't report API that has [EditorBrowsable (None)] attributes, presumably we
  decided to keep these for compatibility, while highly discouraging their
  continued use. Also stop doing this for the next time we can do a breaking
  change, maybe we can remove these APIs then.
* Don't report API that has [UnsupportedOSPlatform ("...#.#") attributes (with
  a version number), presumably this is API that is still valid for some OS
  versions.
* Enable the test for all APIs (no ignores anymore). It's ✅ due to numerous
  PRs this week fixing issues!

Fixes https://github.com/xamarin/xamarin-macios/issues/13621.

### DIFF
--- a/tests/cecil-tests/ObsoleteTest.cs
+++ b/tests/cecil-tests/ObsoleteTest.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 
 using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


dalexsoto (approved):

mandel-macaque (approved):

tests/cecil-tests/ObsoleteTest.cs
```diff
@@ -13,40 +14,60 @@ namespace Cecil.Tests {
...
+				return false;
+
+#if !XAMCORE_5_0
+			// If we've hidden an API from the IDE, assume we've decided to keep the API for binary compatibility
+			// At least until the next time we can do breaking changes.
+			if (HasEditorBrowseableNeverAttribute (provider))
+				return false;
+#endif
+
+			// I'm bad!
```
tj-devel709, line 69: lol

tj-devel709 (approved):

rolfbjarne:
Test failures are unrelated:

* link sdk/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2494
* [NUnit] Mono Mac OS X BCL tests group 2/Mac Full/Debug: https://github.com/xamarin/maccore/issues/2542

PR #14261, opened 2022-02-25
[audiounit] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to AudioUnit.

### DIFF
--- a/src/AudioUnit/AUParameter.cs
+++ b/src/AudioUnit/AUParameter.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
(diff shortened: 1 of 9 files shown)


src/AudioUnit/AUParameter.cs
```diff
@@ -17,7 +19,7 @@ public partial class AUParameter
 		public string GetString (float? value)
 		{
 			unsafe {
-				if (value != null && value.HasValue) {
+				if (value is not null && value.HasValue) {
```
rolfbjarne, line 22: NBD, but this is doing the same thing twice, but in two different ways. This can be simplified as:

```suggestion
				if (value is not null) {
```

rolfbjarne (approved):

PR #14264, opened 2022-02-26
[net][linker] Remove _empty_ `Dispose` on leafs types

spouliot:
This removes the _nearly_ `Dispose` methods when
* no field needs to be set to `null`; and
* no subclass for the type exists

In such case the code would only call `base.Dispose (bool)` which is the
same as not having the override (since it's virtual).

```diff
--- /var/folders/w_/zfvfxj_x3nbd8wlj9p4ck3xc0000gn/T/appcompare/diffcustomtooloutput/Microsoft.iOS.dll.a.text	2022-02-26 11:52:31.000000000 -0500
+++ /var/folders/w_/zfvfxj_x3nbd8wlj9p4ck3xc0000gn/T/appcompare/diffcustomtooloutput/Microsoft.iOS.dll.b.text	2022-02-26 11:52:31.000000000 -0500
@@ -990,11 +990,6 @@
 		protected internal UIApplication(NativeHandle P_0)
 		{
 		}
-
-		[MethodImpl(8)]
-		protected sealed override void Dispose(bool P_0)
-		{
-		}
 	}
 	[Register("UIButton", true)]
 	public sealed class UIButton : UIControl, INativeObject, IDisposable
@@ -1106,11 +1101,6 @@
 		protected internal UIScreen(NativeHandle P_0)
 		{
 		}
-
-		[MethodImpl(8)]
-		protected sealed override void Dispose(bool P_0)
-		{
-		}
 	}
 	[Register("UIView", true)]
 	public class UIView : UIResponder, INativeObject, IDisposable
@@ -1196,11 +1186,6 @@
 		protected internal UIViewController(NativeHandle P_0)
 		{
 		}
-
-		[MethodImpl(8)]
-		protected sealed override void Dispose(bool P_0)
-		{
-		}
 	}
 	[Register("UIWindow", true)]
 	public sealed class UIWindow : UIView
@@ -1241,11 +1226,6 @@
 		public void MakeKeyAndVisible()
 		{
 		}
-
-		[MethodImpl(8)]
-		protected sealed override void Dispose(bool P_0)
-		{
-		}
 	}
 	[Register("Microsoft_iOS__UIKit_UIApplicationDelegate", false)]
 	[Model]
```


tools/dotnet-linker/BackingFieldDelayHandler.cs
```diff
@@ -75,31 +84,55 @@ public static void ReapplyDisposedFields (DerivedLinkContext context, string ope
...
-								OptimizeGeneratedCodeHandler.ValidateInstruction (method, load_this, operation, Code.Ldarg_0)) {
+							if (OptimizeGeneratedCodeHandler.ValidateInstruction (method, store_field, Name, Code.Stfld) &&
+								OptimizeGeneratedCodeHandler.ValidateInstruction (method, load_null, Name, Code.Ldnull) &&
+								OptimizeGeneratedCodeHandler.ValidateInstruction (method, load_this, Name, Code.Ldarg_0)) {
 								store_field.OpCode = OpCodes.Nop;
 								load_null.OpCode = OpCodes.Nop;
 								load_this.OpCode = OpCodes.Nop;
 							}
+						} else if (field.DeclaringType.FullName != "System.IntPtr") {
+							its_a_keeper = true;
```
rolfbjarne, line 95: I'm not sure I'm reading this right, but with this change, we'll only keep Dispose methods that do something with fields whose field types aren't "System.IntPtr".

So for instance a Dispose method that calls another method (and doesn't touch any fields), will be removed.

Is that the intention?

spouliot: Almost. This only process a list of `Dispose` methods that have the `[BindingImpl (OptimizableCode)]` attribute...

_code earlier in the file_

```csharp
// only process methods that are marked as optimizable
if (!method.IsBindingImplOptimizableCode (LinkContext))
   return
```

and their pattern is always the same. E.g. `UIWindow`

```csharp
		protected override void Dispose (bool disposing)
		{
			base.Dispose (disposing);
			if (Handle == IntPtr.Zero) {
				__mt_WindowScene_var = null;
			}
		}
```

It's also only applied on types that are not subclassed. E.g. `UIView` would not be removed, even if empty, because a lot more logic would be needed to replace the calls (like `UIWindow::Dispose` back to its base class).

```cil
	IL_0000: ldarg.0
	IL_0001: ldarg.1
	IL_0002: call instance void UIKit.UIView::Dispose(bool)
```

Would be easier if IL has a `base` concept, but it's actually a `call` to a specific type :|

rolfbjarne: > This only process a list of Dispose methods that have the [BindingImpl (OptimizableCode)] attribute... and their pattern is always the same

That's true... _today_.

I think it's entirely possible that we can end up changing the Dispose implementation, and then this optimization kicks in when it shouldn't, which would only happen on Release builds, and it might be quite baffling to whomever has to diagnose it.

Is it possible to make the logic that decides which methods should be removed more specific? What's the IL of these methods?

spouliot: > I think it's entirely possible that we can end up changing the Dispose 

True, but that same argument apply to any generated code and optimization that are done. That includes the ones that are already present in that step: if you modify the generated `Dispose` then that code must be updated (or at least verified) to match.

A test can be added to make failure more likely but still **any** generated code change must be reflected in **all** the optimizations, that's the contract for any code decorated with `[BindingImpl]`.

> which would only happen on Release builds

I might have missed something but I don't know why this would be limited to release builds ? 

```
<_TrimmerCustomSteps Include="$(_AdditionalTaskAssembly)" Condition="'$(_AreAnyAssembliesTrimmed)' == 'true'" Type="Xamarin.Linker.BackingFieldDelayHandler" />
```

IIRC only the sealer/de-virtualizer step (and IL stripping later) are release only.

> Is it possible to make the logic that decides which methods should be removed more specific? 

It's already limited to `[BindingImpl]` decorated methods.

> What's the IL of these methods?

Before or after (below) the existing processing ?

```cil
.method family hidebysig virtual 
	instance void Dispose (
		bool ''
	) cil managed 
{
	// Method begins at RVA 0x37d4
	// Code size 32 (0x20)
	.maxstack 8

	IL_0000: ldarg.0
	IL_0001: ldarg.1
	IL_0002: call instance void Foundation.NSObject::Dispose(bool)
	IL_0007: ldarg.0
	IL_0008: call instance valuetype ObjCRuntime.NativeHandle Foundation.NSObject::get_Handle()
	IL_000d: ldsfld native int [System.Private.CoreLib]System.IntPtr::Zero
	IL_0012: call bool ObjCRuntime.NativeHandle::op_Equality(valuetype ObjCRuntime.NativeHandle, native int)
	IL_0017: brfalse.s IL_001f

	IL_0019: nop
	IL_001a: nop
	IL_001b: nop
	IL_001c: nop
	IL_001d: nop
	IL_001e: nop

	IL_001f: ret
} // end of method UIView::Dispose
```

The code could reject any other instructions, but that might just hide anything that breaks the optimization.

spouliot: > Removing an override is not a binary breaking change:

Earlier (but years ago) experiments did not work - but it might have been the linker that was unhappy to save back the assembly that did not include the method referenced. Worth checking again...

> to remove the Handle == IntPtr.Zero check as well.

It can be done but I'm not sure it's worth it. It only removes a few IL instructions that are stripped.

> dispose method to check for NativeHandle.Zero instead of IntPtr.Zero, 

Funny, I have a local change that includes this :-)

Still the optimization would only be disabled - which puts it back to today's situation.

So I can add some canary tests (to make sure it does not get disabled) but if you feel it's not safe enough I'll simply close the PR.

rolfbjarne: > It can be done but I'm not sure it's worth it. It only removes a few IL instructions that are stripped.

The idea was that both these optimizations would add up to be something greater, not that each would be particularly significant on its own :)

> but if you feel it's not safe enough I'll simply close the PR.

Yes, I feel this is too risky for the limited gain without additional work.

PR #14275, opened 2022-02-28
[callkit] Add nullability to (generated and manual) bindings

tj-devel709:
I am adding nullable enable to all manual files that are not "API_SOURCES" in src/frameworks.sources and making the required nullability changes. In this case, only one file does not have the header.


rolfbjarne (approved):

mandel-macaque (approved):

tj-devel709:
Unrelated Build Error: https://github.com/xamarin/maccore/issues/2545

PR #14280, opened 2022-03-01
[classkit] Add nullability to (generated and manual) bindings

tj-devel709:
I am adding nullable enable to all manual files that are not "API_SOURCES" in src/frameworks.sources and making the required nullability changes


rolfbjarne (approved):

src/ClassKit/CLSContext.cs
```diff
@@ -15,7 +17,7 @@ namespace ClassKit {
 	public partial class CLSContext {
 
 		public CLSContextTopic Topic {
-			get => CLSContextTopicExtensions.GetValue (WeakTopic);
+			get => CLSContextTopicExtensions.GetValue (WeakTopic!);
```
mandel-macaque, line 20: WeakTopic CAN be null and therefore will be null at some point. The GetValue should check if WeakTopic is null and return CLSContextType.None in the case when it is.
```suggestion
			get => WeakTopic is null ? CLSContextTopic.None :  CLSContextTopicExtensions.GetValue (WeakTopic!);
```

mandel-macaque (changes requested):

src/ClassKit/CLSContext.cs
tj-devel709: Looking at the CLSContextTopicExtensions.GetValue method that is generated:
```
		public static CLSContextTopic GetValue (NSString constant)
		{
			if (constant is null)
				throw new ArgumentNullException (nameof (constant));
			if (constant.IsEqualTo (CLSContextTopicMath))
				return CLSContextTopic.Math;
			if (constant.IsEqualTo (CLSContextTopicScience))
				return CLSContextTopic.Science;
			if (constant.IsEqualTo (CLSContextTopicLiteracyAndWriting))
				return CLSContextTopic.LiteracyAndWriting;
			if (constant.IsEqualTo (CLSContextTopicWorldLanguage))
				return CLSContextTopic.WorldLanguage;
			if (constant.IsEqualTo (CLSContextTopicSocialScience))
				return CLSContextTopic.SocialScience;
			if (constant.IsEqualTo (CLSContextTopicComputerScienceAndEngineering))
				return CLSContextTopic.ComputerScienceAndEngineering;
			if (constant.IsEqualTo (CLSContextTopicArtsAndMusic))
				return CLSContextTopic.ArtsAndMusic;
			if (constant.IsEqualTo (CLSContextTopicHealthAndFitness))
				return CLSContextTopic.HealthAndFitness;
			throw new NotSupportedException ($"{constant} has no associated enum value on this platform.");
		}
```
It does not take a null NSString, but checks for one and throws an exception. Should I perhaps throw an exception then instead since there is no case where it returns CLSContextType.None?

tj-devel709: @mandel-macaque So there is not a CLSContextTopic.None and the documentation for the enum does not have a default value. Would you recommend just using the first enum value for the null value?

mandel-macaque: My mistake, as per the docs, the property takes a nullable value: https://developer.apple.com/documentation/classkit/clscontext/2953026-topic?language=objc to chante it, I wonder if that is present in the xtro files.

tj-devel709: @mandel-macaque Hmm now I am confused.. So there is no ignore files for ClassKit which is not the part I am confused about. I am confused that I cannot find where this method is located: 
`CLSContextTopicExtensions.GetValue (NSString constant)`

I see the generated method inside CLSContextTopic.g.cs but by grepping, I do not see this method or the CLSContextTopicExtensions class.

tj-devel709:
@mandel-macaque The enum now takes a nullable value and I think is sufficient regarding your request!

tj-devel709:
**Unrelated Test Failures**
- link sdk/Mac Catalyst [dotnet]/Debug [dotnet]: https://github.com/xamarin/maccore/issues/2414
- trimmode copy/Mac Catalyst [dotnet]/Release [dotnet]: https://github.com/xamarin/maccore/issues/2414

**Unrelated Build Failure**
https://github.com/xamarin/maccore/issues/2545

mandel-macaque (approved):
Wait for the tests to be completed.

tj-devel709:
After this PR lands: https://github.com/xamarin/xamarin-macios/pull/14334
I will update this branch and those tests should pass!

PR #14281, opened 2022-03-01
[ARKit] Add ARKit to Mac Catalyst.

rolfbjarne:


spouliot (commented):
Not sure it's worth shipping bindings that would only call stubs.

> This framework ignores calls from Mac apps built with Mac Catalyst.

https://developer.apple.com/documentation/arkit/

rolfbjarne:
> Not sure it's worth shipping bindings that would only call stubs.
> 
> > This framework ignores calls from Mac apps built with Mac Catalyst.
> 
> https://developer.apple.com/documentation/arkit/

That's a good point 😐

PR #14297, opened 2022-03-03
Remove reference to SocketsHttpHandler

eerhardt:
ObjCRuntime.RuntimeOptions.GetHttpMessageHandler has code in it to create a SocketsHttpHandler, which in turn is preventing all of the HTTP stack in System.Net.Http.dll to be preserved in an iOS application. This is not necessary because System.Net.Http will create a SocketsHttpHandler itself and not call RuntimeOptions.GetHttpMessageHandler when the application is configured for System.Net.Http.UseNativeHttpHandler=false.

Removing this reference from RuntimeOptions will allow all the System.Net.Http internal logic (like Http2/Http3 handling), and other dependencies like System.Net.Security, Crypto, etc to be trimmed in a typical iOS application.

cc @rolfbjarne


rolfbjarne (approved):

dalexsoto (approved):

rolfbjarne:

Issue #12007, opened 2021-06-23
Mac app crashes while using FSEventStreamCreateFlags.UseExtendedData


anoopvaidya:
### Steps to Reproduce

Create a Mac project.
Use the following code :
_fsEventStream = new FSEventStream(new[] { path },
                TimeSpan.FromSeconds(0),
                FSEventStreamCreateFlags.UseExtendedData
                );
Build and run the app, in the observed folder, rename or delete any file. The app crashes and gives the attached Crash report

Note: For all other Flags, the client works.

### Expected Behavior
The app should work and give the callbacks.

### Actual Behavior
The app crashes.

### Environment

MacOS11.4
=== Visual Studio Community 2019 for Mac ===

Version 8.10.3 (build 15)
Installation UUID: c0925b23-6e9d-4b80-a12d-fddf5ddbeb32
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000140

=== Mono Framework MDK ===

Runtime:
	Mono 6.12.0.140 (2020-02/51d876a041e) (64-bit)
	Package version: 612000140

=== Roslyn (Language Service) ===

3.10.0-4.21269.26+029847714208ebe49668667c60ea5b0a294e0fcb

=== NuGet ===

Version: 5.9.0.7134

=== .NET Core SDK ===

SDK: /usr/local/share/dotnet/sdk/5.0.301/Sdks
SDK Versions:
	5.0.301
	5.0.203
	5.0.202
	5.0.100
	3.1.410
	3.1.409
	3.1.408
	3.1.404
MSBuild SDKs: /Applications/Visual Studio.app/Contents/Resources/lib/monodevelop/bin/MSBuild/Current/bin/Sdks

=== .NET Core Runtime ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	5.0.7
	5.0.6
	5.0.5
	5.0.0
	3.1.16
	3.1.15
	3.1.14
	3.1.10

=== .NET Core 3.1 SDK ===

SDK: 3.1.410

=== Xamarin.Profiler ===

Version: 1.6.15.68
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Updater ===

Version: 11

=== Xamarin.Android ===

Not Installed

=== Microsoft OpenJDK for Mobile ===

Java SDK: Not Found

[...]

PR #14318, opened 2022-03-04
FSEventStream: fix crash when asking for extended data; support dispatch queues (#12007)

abock:
Implements support for `FSEventStreamCreateFlags.UseExtendedData`, fixing #12007.

When `.UseExtendedData` is specified, the event data type changes from `CFString` to `CFDictionary`; the dictionary contains the path (`path` key) and inode (`fileID` key) information for the file and may be extended in the future with other fields. Previously this was crashing because we assumed `CFString` always.

Further add a convenience constructor for monitoring a single path, add the missing `UnscheduleFromRunLoop` APIs, and add `SetDispatchQueue` to allow using dispatch queues directly instead of run loops.

Finally, this PR adds a fairly exhaustive file system test which covers the existing (non-extended) and fixed (extended) creation modes, along with using a dispatch queue instead of run loop.

Fixes https://github.com/xamarin/xamarin-macios/issues/12007

### DIFF
--- a/src/CoreServices/FSEvents.cs
+++ b/src/CoreServices/FSEvents.cs
@@ -354,6 +354,36 @@ namespace CoreServices
 		}
 
 		[DllImport (Constants.CoreServicesLibrary)]
+		static extern void FSEventStreamUnscheduleFromRunLoop (IntPtr handle,
+			IntPtr runLoop, IntPtr runLoopMode);
+
+		public void UnscheduleFromRunLoop (CFRunLoop runLoop, NSString runLoopMode)
+		{
+			FSEventStreamScheduleWithRunLoop (GetCheckedHandle (), runLoop.Handle, runLoopMode.Handle);
+		}
+
+		public void UnscheduleFromRunLoop (CFRunLoop runLoop)
+		{
+			UnscheduleFromRunLoop (runLoop, CFRunLoop.ModeDefault);
+		}
+
+		public void UnscheduleFromRunLoop (NSRunLoop runLoop, NSString runLoopMode)
+		{
+			UnscheduleFromRunLoop (runLoop.GetCFRunLoop (), runLoopMode);
+		}
+
+		public void UnscheduleFromRunLoop (NSRunLoop runLoop)
+		{
+			UnscheduleFromRunLoop (runLoop.GetCFRunLoop (), CFRunLoop.ModeDefault);
+		}
+
+		[DllImport (Constants.CoreServicesLibrary)]
+		static extern void FSEventStreamSetDispatchQueue (IntPtr handle, IntPtr dispatchQueue);
+
+		public void SetDispatchQueue (DispatchQueue? dispatchQueue)
+			=> FSEventStreamSetDispatchQueue (GetCheckedHandle (), dispatchQueue.GetHandle ());
+
+		[DllImport (Constants.CoreServicesLibrary)]
 		static extern IntPtr FSEventStreamCopyPathsBeingWatched (IntPtr handle);
 
 		public string? []? PathsBeingWatched {
--- a/src/CoreServices/FSEvents.cs
+++ b/src/CoreServices/FSEvents.cs
@@ -88,10 +88,11 @@ namespace CoreServices
 		public ulong Id { get; internal set; }
 		public string? Path { get; internal set; }
 		public FSEventStreamEventFlags Flags { get; internal set; }
+		public ulong FileId { get; internal set; }
 
 		public override string ToString ()
 		{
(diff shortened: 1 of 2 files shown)


rolfbjarne (changes requested):

src/CoreServices/FSEvents.cs
```diff
@@ -259,12 +288,37 @@ static void FreeGCHandle (IntPtr gchandle)
...
+					var fileIdHandle = CFDictionaryGetValue (
+						eventDataHandle,
+						kFSEventStreamEventExtendedFileIDKey.Handle);
+					if (fileIdHandle != IntPtr.Zero)
+						CFNumberGetValue (fileIdHandle, 4 /*kCFNumberSInt64Type*/, out fileId);
+				}
+
+				events[i] = new FSEvent
+				{
+					Id = (ulong)Marshal.ReadInt64 (eventIds, i * 8),
```
Therzok, line 317: This looks like something that could be wrapped into a `ReadOnlySpan<ulong>()`. I think it's been done already in a similar fashion over at dotnet/runtime in FileSystemWatcher.OSX.cs

src/CoreServices/FSEvents.cs
```diff
@@ -248,6 +254,29 @@ static void FreeGCHandle (IntPtr gchandle)
 			GCHandle.FromIntPtr (gchandle).Free ();
 		}
 
+		[DllImport (Constants.CoreFoundationLibrary)]
+		static extern IntPtr CFArrayGetValueAtIndex (IntPtr array, int index);
```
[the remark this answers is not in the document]
abock: Yeah, I forgot to backpedal the duplicate P/Invokes when integrating to the build. The issue here is that it takes hours to provision a machine with the dependencies to actually build all of Xamarin.Mac, and a lot of the low level CF APIs are internal. I added extended data support out-of-band to iterate quickly on it. Really wish all the CF P/Invokes were cleanly organized and public 😄

tests/monotouch-test/CoreServices/FSEventStreamTest.cs
```diff
@@ -5,24 +5,226 @@
...
 
 	[TestFixture]
 	[Preserve (AllMembers = true)]
-	public class FSEventStreamTest {
+	public sealed class FSEventStreamTest {
+		const string TestUid = "com.microsoft.xamarin-macios.fs-events-tests";
 
 		[Test]
-		public void Create ()
+		public void TestFileEvents ()
```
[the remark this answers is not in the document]
abock: I backed the tests off to just create 256 files on disk regardless of rlmits. Bot IO is too slow and the test times out. The important part is that the test creates files, observes them through FSEventsStream, deletes files, and observes those deletions as well, for both extended-data and non-extended-data paths.

src/CoreServices/FSEvents.cs
abock: Sure, but this code was just moved from a few lines above. This isn't new code, so I didn't go around changing things.

abock: Er, no, that would mean the underlying native alloc'd memory would have to stick around. It's releated once the callback returns.

abock:
I forgot to mention, the reason I added the `SetDispatchQueue` API _in this PR_ is to enable the new tests. The test needs to block the main thread for the sake of NUnit, which means we can't easily use the main run loop for dispatching FS events, and it's far simpler _and more appropriate_ to create a private dispatch queue than it is a non-main-run-loop. I plan on an additional PR to bring the binding up to parity - there are missing `dev_t` based APIs that were added at the time extended data events were added (which this PR fixes) that are necessary for root-path-agnostic file watches. Depending on the velocity of this PR though I might tack the additions on to this PR as additional commits. YOLO. ❤️

abock:
@rolfbjarne I have another patch incoming for the two missing `dev_t` based APIs as well, which become more useful with extended data events (path agnostic device ID and inode-based watching). Would you prefer I layer it into this PR or post a subsequent PR?

rolfbjarne:
@abock new PR is best I think, doing it in chunks usually ends up easier :)

abock:
@rolfbjarne okay, looking at the `dev_t` stuff, it's going to introduce some combinatorial .ctor explosion, and there's a chance to clean up the `sinceWhenId` parameter as well to make `FSEventStream` creation cleaner and more fluent. With that coming work in mind, I added commit 5086857 to drop the new `string pathToWatch` .ctor override to ensure that convenience API addition isn't released and frozen in time (yet). I can rebase if you'd like but figured a separate commit would be easier review. Apologies for the noise :-/

dalexsoto (approved):
nice ❤️

PR #14333, opened 2022-03-08
[tests] Don't use reflection to create HttpClient handlers.

rolfbjarne:
Fixes two test failures on tvOS with all optimizations enabled:

    MonoTests.System.Net.Http.MessageHandlerTest.DnsFailure
        [FAIL] DnsFailure(System.Net.Http.SocketsHttpHandler) :   Exception
            Expected: instance of <System.Net.Http.HttpRequestException>
            But was:  <System.MissingMethodException: No parameterless constructor defined for type 'System.Net.Http.SocketsHttpHandler'.
                at System.RuntimeType.CreateInstanceMono(Boolean , Boolean )
                at System.RuntimeType.CreateInstanceDefaultCtor(Boolean , Boolean )
                at System.Activator.CreateInstance(Type , Boolean , Boolean )
                at System.Activator.CreateInstance(Type , Boolean )
                at System.Activator.CreateInstance(Type )
                at MonoTests.System.Net.Http.MessageHandlerTest.GetHandler(Type ) in /Users/builder/azdo/_work/1/s/xamarin-macios/tests/monotouch-test/System.Net.Http/MessageHandlers.cs:line 50
                at MonoTests.System.Net.Http.MessageHandlerTest.<>c__DisplayClass3_0.<<DnsFailure>b__0>d.MoveNext() in /Users/builder/azdo/_work/1/s/xamarin-macios/tests/monotouch-test/System.Net.Http/MessageHandlers.cs:line 76>

    MonoTests.System.Net.Http.MessageHandlerTest.RejectSslCertificatesServicePointManager
        [FAIL] RejectSslCertificatesServicePointManager(System.Net.Http.SocketsHttpHandler) : System.MissingMethodException : No parameterless constructor defined for type 'System.Net.Http.SocketsHttpHandler'.
            at System.RuntimeType.CreateInstanceMono(Boolean , Boolean )
            at System.RuntimeType.CreateInstanceDefaultCtor(Boolean , Boolean )
            at System.Activator.CreateInstance(Type , Boolean , Boolean )
            at System.Activator.CreateInstance(Type , Boolean )
            at System.Activator.CreateInstance(Type )
            at MonoTests.System.Net.Http.MessageHandlerTest.GetHandler(Type ) in /Users/builder/azdo/_work/1/s/xamarin-macios/tests/monotouch-test/System.Net.Http/MessageHandlers.cs:line 50
[...]

PR #14337, opened 2022-03-08
[msbuild] Make the default value for NoDSymUtil the same for Mac Catalyst as it is for macOS.

rolfbjarne:
I think this is what makes most sense, since they're both desktop platforms.


chamons (approved):

emaf (approved):

dalexsoto (approved):

rolfbjarne:
Test failures are unrelated (#14333, #14334)

PR #14341, opened 2022-03-08
[cfnetwork] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to CFNetwork.

**********
Something interesting about this framework is that the following files are not inside the src/frameworks.sources, but when removed or `#nullable enable`'d, it did not compile:
src/CFNetwork/Content.cs
src/CFNetwork/MessageHandler.cs
src/CFNetwork/WebRequestStream.cs
src/CFNetwork/WebResponseStream.cs
src/CFNetwork/WorkerThread.cs

Extra attention on the async/await changes would be great! These changes were brought by the compiler complaining about awaitable methods not using await so for most of them I added async and await but for two that were returning Tasks, I added a progma ignore warning. If desired, I can ignore the others as well! - These changes are in the `Enable Nullability and fix errors including async and await changes` commit!

### DIFF
--- a/src/CFNetwork/MessageHandler.cs
+++ b/src/CFNetwork/MessageHandler.cs
@@ -105,7 +105,7 @@ namespace CFNetwork {
 
 			SetupRequest (request, message, isFirstRequest);
 
-			if ((auth == null) || (Credentials == null) || !PreAuthenticate)
+			if ((auth is null) || (Credentials is null) || !PreAuthenticate)
 				return message;
 
 			if (auth is not null && !auth.AppliesToRequest (message))
@@ -113,7 +113,7 @@ namespace CFNetwork {
 
 			var method = auth?.GetMethod ();
 			var credential = Credentials.GetCredential (request.RequestUri, method);
-			if (credential == null)
+			if (credential is null)
 				return message;
 
 			if (isFirstRequest)
@@ -127,11 +127,11 @@ namespace CFNetwork {
 			if ((AutomaticDecompression & DecompressionMethods.GZip) != 0)
 				accept_encoding = "gzip";
 			if ((AutomaticDecompression & DecompressionMethods.Deflate) != 0)
-				accept_encoding = accept_encoding != null ? "gzip, deflate" : "deflate";
-			if (accept_encoding != null)
+				accept_encoding = accept_encoding is not null ? "gzip, deflate" : "deflate";
+			if (accept_encoding is not null)
 				message.SetHeaderFieldValue ("Accept-Encoding", accept_encoding);
 
-			if (request.Content != null) {
+			if (request.Content is not null) {
 				foreach (var header in request.Content.Headers) {
 					if (!isFirstRequest && header.Key == "Authorization")
 						continue;
@@ -141,13 +141,13 @@ namespace CFNetwork {
 			}
 
 			foreach (var header in request.Headers) {
-				if ((accept_encoding != null) && header.Key.Equals ("Accept-Encoding"))
+				if ((accept_encoding is not null) && header.Key.Equals ("Accept-Encoding"))
 					continue;
 				var value = string.Join (",", header.Value);
 				message.SetHeaderFieldValue (header.Key, value);
 			}
 
-			if (UseCookies && (CookieContainer != null)) {
+			if (UseCookies && (CookieContainer is not null)) {
 				string cookieHeader = CookieContainer.GetCookieHeader (request.RequestUri);
 				if (cookieHeader != "")
 					message.SetHeaderFieldValue ("Cookie", cookieHeader);
@@ -157,7 +157,7 @@ namespace CFNetwork {
 		async Task<WebRequestStream?> CreateBody (HttpRequestMessage request, CFHTTPMessage message,
 		                                         CancellationToken? cancellationToken)
 		{
-			if (request.Content == null)
+			if (request.Content is null)
 				return null;
 
 			/*
@@ -176,7 +176,7 @@ namespace CFNetwork {
 			 *
 			 */
 			var length = request.Content.Headers.ContentLength;
-			if (((request.Content is StreamContent) || (length == null))) {
+			if (((request.Content is StreamContent) || (length is null))) {
 				var stream = await request.Content.ReadAsStreamAsync ().ConfigureAwait (false);
 				return new WebRequestStream (stream, cancellationToken ?? CancellationToken.None);
 			}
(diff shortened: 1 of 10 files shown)


src/CFNetwork/MessageHandler.cs
```diff
@@ -103,15 +105,15 @@ CFHTTPMessage CreateRequest (HttpRequestMessage request, bool isFirstRequest)
...
+			if ((auth is null) || (Credentials is null) || !PreAuthenticate)
 				return message;
 
-			if (!auth.AppliesToRequest (message))
+			if (auth is not null && !auth.AppliesToRequest (message))
 				return message;
 
-			var method = auth.GetMethod ();
+			var method = auth?.GetMethod ();
 			var credential = Credentials.GetCredential (request.RequestUri, method);
```
mandel-macaque, line 115: Does the Credentials.GetCredential take a null value in the method? That is something to look at before we land this since we now know that method can be null.

src/CFNetwork/MessageHandler.cs
```diff
@@ -174,9 +176,9 @@ void SetupRequest (HttpRequestMessage request, CFHTTPMessage message, bool isFir
 			 *
 			 */
 			var length = request.Content.Headers.ContentLength;
-			if ((request.Content is StreamContent) || (length == null)) {
+			if (((request.Content is StreamContent) || (length is null))) {
 				var stream = await request.Content.ReadAsStreamAsync ().ConfigureAwait (false);
-				return new WebRequestStream (stream, cancellationToken);
+				return new WebRequestStream (stream, cancellationToken ?? CancellationToken.None);
```
mandel-macaque, line 181: :+1:

src/CFNetwork/WebRequestStream.cs
```diff
@@ -74,18 +76,18 @@ public WebRequestStream (Stream stream, CancellationToken cancellationToken)
...
+					await Write ();
 			};
-			writeStream.CanAcceptBytesEvent += (sender, e) => {
+			writeStream.CanAcceptBytesEvent += async (sender, e) => {
 				if (!open) {
 					canWrite = true;
 					return;
 				}
-				Write ();
+				await Write ();
```
mandel-macaque, line 90: I really like the change

mandel-macaque (approved):
Olease verify that those methods that are not for sure getting a null will not throw a NRE, simply confirming that is good enough,

src/CFNetwork/MessageHandler.cs
tj-devel709: Haha I looked that one up!

src/CFNetwork/MessageHandler.cs
tj-devel709: That is a good point.
Because these files are not in the xamios.csproj (or frameworks.sources), I added them to the xamios.csproj just to handle the nullability since visual studio underlines them, but there are a lot of things that Visual Studio can't tell me and one of those things is what the heck the Credentials variable is. Thus I cannot find the method for Credentials.GetCredential... I still don't know where these files are being used, but building main does complain if I don't fix all the nullability
image

src/CFNetwork/WebRequestStream.cs
rolfbjarne: I like it too, but it's also a behavior change in a PR with a lot of unrelated changes, so I think it might be best to have these changes in a different PR, especially since networking is code where we have limited test coverage.

src/CFNetwork/WebResponseStream.cs
```diff
@@ -595,28 +600,33 @@ void DoOpen ()
...
 			stream.ClosedEvent += (sender, e) => {
 				if (!open) {
 					open = true;
-					openTcs.SetResult (stream.GetResponseHeader ());
+					openTcs?.SetResult (stream.GetResponseHeader ());
 				}
 				OnCompleted ();
 			};
-			stream.HasBytesAvailableEvent += (sender, e) => {
+			stream.HasBytesAvailableEvent += async (sender, e) => {
```
rolfbjarne, line 618: Same goes for these async/await changes.

rolfbjarne (changes requested):

src/CFNetwork/WebRequestStream.cs
tj-devel709: @rolfbjarne so would this look like just adding the ignore warning progma thing in this PR?

tj-devel709: @rolfbjarne Yes, setting `#nullable enable` enables 'Warnings as Errors' according to Sebastien's docs, so I cannot compile without setting ignores to any warnings / fixing them

src/CFNetwork/MessageHandler.cs
```diff
@@ -27,6 +27,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+#nullable enable
```
dalexsoto, line 30: I would say that we should just rollback all the changes in this file since it is deprecated and since enabling nullability here would need us to look at some behavior changes 

```csharp
[Obsolete ("Use 'System.Net.Http.CFNetworkHandler' or the more recent 'Foundation.NSUrlSessionHandler' instead.")]
	public class MessageHandler : HttpClientHandler {
```

dalexsoto (changes requested):

src/CFNetwork/MessageHandler.cs
mandel-macaque: There was a reason why I asked ;) 

This class is extending the HttpClientHandler, that is a System.Net class, so you will not find the source in our tree, you can use the reference code here: https://github.com/microsoft/referencesource/blob/master/System/net/System/Net/Http/HttpClientHandler.cs

Credentials is a ICrendential implementation, which is using https://docs.microsoft.com/en-us/dotnet/api/system.net.icredentials.getcredential?view=net-6.0 The issue here is that, the signature of the interface does not let us know if it is nullable, in that case, I wold not consider it safe to send a null parameter to a contract (interface) that does not explicitly say that it takes null. Also, the "method" parameter is https://docs.microsoft.com/en-us/dotnet/api/system.net.iauthenticationmodule.authenticationtype?view=net-6.0#system-net-iauthenticationmodule-authenticationtype which is the http auth type values, which AFAIK does not take null (I can only remember a few: Basic, Digest, Ntml, Kerberos and Anonymous, maybe null == anonymous? ).

I would not do this changes since it looks like you added a number of possible NRE based in the auth? being possibly null, in this particular case, we need to check if null and do the correct thing rather than use ? in the method calls.

mandel-macaque (changes requested):
See my comment in our thread.

src/CFNetwork/WebRequestStream.cs
rolfbjarne: Ah ok, in that case yes, please add #pragma's to ignore the warnings

tj-devel709:
Closing this PR in favor of this one: https://github.com/xamarin/xamarin-macios/pull/14362
Just carrying over changes in the files that are not !Net

Issue #3931, opened 2018-04-13
unable to find utility *, not a developer tool or in PATH


rolfbjarne:
### Steps to Reproduce

1. Ensure the system Xcode (as defined by `xcode-select -p`) points to a different Xcode than Visual Studio for Mac.
2. Execute `xcrun -k`
3. Create an iOS project, makes sure it has multiple architectures selected (and device-specific builds disabled), and build it.

Result:

> lipo : error : unable to find utility "lipo", not a developer tool or in PATH

The error can be reproduced directly on the command line as well:

```shell
xcrun -k
$ DEVELOPER_DIR=/Applications/Xcode92.app XCODE_DEVELOPER_DIR_PATH=/Applications/Xcode93.app/Contents/Developer xcrun lipo
xcrun: error: unable to find utility "lipo", not a developer tool or in PATH
```

The problem is the `XCODE_DEVELOPER_DIR_PATH` environment variable, which confuses Apple's tooling (when building an XI or XM project the variable is inherited from Visual Studio for Mac, which sets it).

### Workarounds

* Make sure the system Xcode and the VSfM Xcode is the same.
* Run `xcrun` once without the variable (per broken tool) so that the tool's path is cached:

    ```shell
    $ xcrun -k
    $ xcrun clang --version
    Apple LLVM version 9.0.0 (clang-900.0.39.2)
    [...]
    $ DEVELOPER_DIR=/Applications/Xcode92.app XCODE_DEVELOPER_DIR_PATH=/Applications/Xcode93.app/Contents/Developer xcrun clang --version
    Apple LLVM version 9.0.0 (clang-900.0.39.2)
    [...]
    ```

### Potential fixes

1. Make mtouch and mmp unset `XCODE_DEVELOPER_DIR_PATH` unconditionally.
2. Make VSfM unset `XCODE_DEVELOPER_DIR_PATH`. Presumably there's a reason it's set in the first place, so this might be non-trivial.

I think making mtouch and mmp unset the variable is the simplest and fastest to implement.

madhurig:
Is there a way to change "VS for Mac" Xcode settings without going through the UI? We have a customer running into this on the [VSTS Hosted macOS VMs](https://docs.microsoft.com/en-us/vsts/build-release/concepts/agents/hosted?view=vsts) and we are looking for a way to switch the "VS for mac" default via a script. 

Thanks,
Madhuri

rolfbjarne:
[...]

PR #14342, opened 2022-03-08
[msbuild] Improve the Strip task.

rolfbjarne:
* Augment it to be able to take multiple files to strip at the same time.
* Strip in parallel.
* Execute using xcrun (ref: #3931)
* Pass the full path to the executable file to strip, to make command lines
  easier to copy-paste.
* Remove test that is now outdated. We have other tests that run strip
  anyways, so this shouldn't be a problem.


chamons (approved):

mandel-macaque (approved):

msbuild/Xamarin.MacDev.Tasks.Core/Tasks/SymbolStripTaskBase.cs
```diff
@@ -1,64 +1,67 @@
...
 namespace Xamarin.MacDev.Tasks
 {
-	public abstract class SymbolStripTaskBase : XamarinToolTask
+	public abstract class SymbolStripTaskBase : XamarinTask
 	{
 		#region Inputs
 
 		[Required]
-		public string Executable { get; set; }
+		public ITaskItem[] Executable { get; set; } = Array.Empty<ITaskItem> ();
```
emaf, line 19: When building from VS, making this an ITaskItem[] will make XMA copy files from Windows. 
I'm not sure why this wasn't an ITaskItem and how this change will impact the build (do the files we need exist on Windows or those were generated by a previous task on the Mac (in that case the Windows file will be empty)?

We can also avoid copying files to the Mac by ShouldCopyToBuildServer on the SymbolStrip task.

rolfbjarne: @emaf I implemented `ShouldCopyToBuildServer` to get the old behavior.

emaf (approved):

PR #14345, opened 2022-03-08
[coreanimation] Add nullability to (generated and manual) bindings

tj-devel709:
This PR aims to bring nullability changes to CoreAnimation.

Changing any == null or != null to is null and is not null


src/CoreAnimation/CALayer.cs
```diff
@@ -122,7 +122,7 @@ void OnDispose ()
 
 		public void SetContents (NSObject value)
 		{
-			_Contents = value == null ? IntPtr.Zero : value.Handle;
+			_Contents = value?.Handle ?? IntPtr.Zero;
```
rolfbjarne, line 125: Even simpler:

```suggestion
			_Contents = value.GetHandle ();
```

rolfbjarne (approved):

dalexsoto (approved):

PR #14346, opened 2022-03-08
[avfoundation] Add nullability to (generated and manual) bindings - Fixed Branch

tj-devel709:
APOLOGIES! I messed up this branch here: https://github.com/xamarin/xamarin-macios/pull/14134
So this is the fixed branch with the changes cherry-picked over here!

This PR aims to bring nullability changes to avfoundation.

I am looking for any !missing-null-allowed! in this frameworks ignore files and applying them
I am adding nullable enable to all manual files that are not "API_SOURCES" in src/frameworks.sources and making the required nullability changes
Changing all throw new ArgumentNullException ("object")); to ObjCRuntime.ThrowHelper.ThrowArgumentNullException (nameof (object)); for size saving optimization as well to mark that this framework contains nullability changes
Changing any == null or != null to is null and is not null


tj-devel709:
@rolfbjarne I addressed the latest comment you left on the old PR on this one
@mandel-macaque There were a few questions on the old PR that I had responded to that you may want to check over.

Old PR link: https://github.com/xamarin/xamarin-macios/pull/14134#pullrequestreview-895294053

src/AVFoundation/Events.cs
```diff
@@ -69,30 +71,30 @@ public InternalAVAudioPlayerDelegate ()
...
 				throw new ObjectDisposedException ("player", "the player object was Dispose()d during the callback, this has corrupted the state of the program");
 		}
 	
 		[Preserve (Conditional = true)]
-		public override void DecoderError (AVAudioPlayer player, NSError  error)
+		public override void DecoderError (AVAudioPlayer player, NSError?  error)
 		{
-			if (cbDecoderError != null)
+			if (cbDecoderError is not null && error is not null)
 				cbDecoderError (player, new AVErrorEventArgs (error));
```
rolfbjarne, line 84: I think this is better, that way we won't swallow errors, but let the app developer handle them even if we're not told what's going wrong:

```suggestion
			if (cbDecoderError is not null)
				cbDecoderError (player, error is not null ? new AVErrorEventArgs (error) : null);
```

src/AVFoundation/Events.cs
```diff
@@ -165,28 +171,28 @@ public InternalAVAudioRecorderDelegate ()
...
+			if (cbFinishedRecording is not null)
 				cbFinishedRecording (recorder, new AVStatusEventArgs (flag));
 		}
 	
 		[Preserve (Conditional = true)]
-		public override void EncoderError (AVAudioRecorder recorder, NSError error)
+		public override void EncoderError (AVAudioRecorder recorder, NSError? error)
 		{
-			if (cbEncoderError != null)
+			if (cbEncoderError is not null && error is not null)
```
rolfbjarne, line 181: Same

rolfbjarne (changes requested):

tj-devel709:
@rolfbjarne Just a ping to let you know the previous suggestions were updated!

rolfbjarne (approved):

src/AVFoundation/AVAudioPlayer.cs
```diff
@@ -26,27 +26,29 @@
...
+		public static AVAudioPlayer? FromUrl (NSUrl url, out NSError? error)
 		{
 			unsafe {
 				IntPtr errhandle;
 				IntPtr ptrtohandle = (IntPtr) (&errhandle);
 
 				var ap = new AVAudioPlayer (url, ptrtohandle);
 				if (ap.Handle == IntPtr.Zero){
-					error = (NSError) Runtime.GetNSObject (errhandle);
+					error = Runtime.GetNSObject<NSError> (errhandle);
```
mandel-macaque, line 43: 👍

src/AVFoundation/AVLayerVideoGravity.cs
```diff
@@ -57,7 +59,7 @@ static internal AVLayerVideoGravity KeyToEnum (NSString key, bool clamp = true)
 			return (AVLayerVideoGravity) (-1);
 		}
 
-		static internal NSString EnumToKey (AVLayerVideoGravity vg)
+		static internal NSString EnumToKey (AVLayerVideoGravity? vg)
```
mandel-macaque, line 62: Why can't you return NSString? and remove the null!

mandel-macaque (changes requested):

src/AVFoundation/AVLayerVideoGravity.cs
tj-devel709: @mandel-macaque The problem with adding the nullable return is that properties coming from src/avfoundation.cs that are not nullable are being assigned with EnumToKey. We looked at adding the default value for GravityResizeAspect, but @rolfbjarne pointed out [here](https://github.com/xamarin/xamarin-macios/pull/14134#pullrequestreview-901359406) that VideoGravity-type objects are not the only assignments made by EnumToKey and that it is just better to leave as is.

src/AVFoundation/AVAssetDownloadStorageManagementPolicy.cs
```diff
@@ -18,7 +20,7 @@ public partial class AVMutableAssetDownloadStorageManagementPolicy {
 		
 		public override AVAssetDownloadedAssetEvictionPriority Priority {
 			get { return AVAssetDownloadedAssetEvictionPriorityExtensions.GetValue (_Priority); }
-			set { _Priority = value.GetConstant (); }
+			set { _Priority = value.GetConstant ()!; }
```
mandel-macaque, line 23: I would throw an exception since if GetConstant returns null it means that AVAssetDownloadedAssetEvictionPriority is out of range. Setters can throw exceptions.

mandel-macaque (approved):
Please throw the exception in the setter.

PR #14347, opened 2022-03-08
[CI] Fix Variable for Halted Jobs

tj-devel709:
This PR fixes a variable name and makes sure it gets passed into both jobs that require it! From the [last pipeline run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5844899&view=logs&j=704af8c3-a112-593d-cee6-636e1ea22709), and the future one here, we should be able to see that the `Upload test results to VSDrops` and `Publish Html report in VSDrops` jobs run all the way through and do not fail!

Also sorry for all the commits - I tried to rebase but I guess merging main made things not as easy as it usually is for me


mandel-macaque:
It looks like there is  branch in origin with the same name, are this related? We could have done with a single branch in origin which would trigger a CI build for testing and then create a PR from that one to get a PR run in theCI.

mandel-macaque (approved):

dalexsoto (approved):

rolfbjarne (approved):

tj-devel709:
> It looks like there is branch in origin with the same name, are this related? We could have done with a single branch in origin which would trigger a CI build for testing and then create a PR from that one to get a PR run in theCI.

Yeah I was testing this branch on origin, but thought a PR would be cleaner coming from my fork. Next time I'll just leave it on origin and create a PR from there!

tj-devel709:
Actually, I think I may have fixed that variable, but it doesn't look like the steps that are surrounded by the conditional are running. I'm going to look at this one more!

chamons (approved):

mandel-macaque:
I'm closing the PR in favour of some work that is coming after the EO work landed.

